PR #4656, opened 2019-07-05
Make sstable reader fail on unknown column names

### LINKED ISSUE #2598: Sstable reader should fail on unknown column names
Currently it drops them, which is not safe. We risk data loss due to a bug. Refs #2597. Dropped columns would still be recognized, because they're recorded in schema.

haaawk:
Fixes #2598.

Tests: unit(dev)

### DIFF
--- a/sstables/mp_row_consumer.hh
+++ b/sstables/mp_row_consumer.hh
@@ -225,6 +225,17 @@ public:
         }
     };
 
+    void check_column_missing_in_current_schema(const column& col, int64_t timestamp) {
+        if (__builtin_expect(col.cell.empty() || col.cdef, true)) {
+            return;
+        }
+        auto cell_str = sstring(to_sstring_view(col.cell));
+        auto dc = _schema->dropped_columns().find(cell_str);
+        if (dc == _schema->dropped_columns().end() || dc->second.timestamp < timestamp) {
+            throw malformed_sstable_exception(format("Column {} missing in current schema", cell_str));
+        }
+    }
+
 private:
     // Notes for collection mutation:
     //
@@ -509,6 +520,7 @@ public:
     //}
     proceed do_consume_cell(bytes_view col_name, int64_t timestamp, int64_t ttl, int64_t expiration, CreateCell&& create_cell) {
         struct column col(*_schema, col_name, timestamp);
+        check_column_missing_in_current_schema(col, timestamp);
 
         auto ret = flush_if_needed(col.is_static, col.clustering);
         if (_skip_in_progress) {
@@ -576,6 +588,7 @@ public:
     virtual proceed consume_deleted_cell(bytes_view col_name, sstables::deletion_time deltime) override {
         auto timestamp = deltime.marked_for_delete_at;
         struct column col(*_schema, col_name, timestamp);
+        check_column_missing_in_current_schema(col, timestamp);
         gc_clock::duration secs(deltime.local_deletion_time);
 
         return consume_deleted_cell(col, timestamp, gc_clock::time_point(secs));
--- a/tests/sstable_3_x_test.cc
+++ b/tests/sstable_3_x_test.cc
@@ -4969,66 +4969,6 @@ SEASTAR_THREAD_TEST_CASE(test_read_missing_summary) {
     sst.load();
 }
 
-SEASTAR_THREAD_TEST_CASE(test_sstable_reader_on_unknown_column) {
-    api::timestamp_type write_timestamp = 1525385507816568;
-    auto wait_bg = seastar::defer([] { sstables::await_background_jobs().get(); });
-    storage_service_for_tests ssft;
-    auto get_builder = [&] (bool has_missing_column) {
-        auto builder = schema_builder("ks", "cf")
-            .with_column("pk", int32_type, column_kind::partition_key)
-            .with_column("ck", int32_type, column_kind::clustering_key);
-        if (!has_missing_column) {
-            builder.with_column("val1", int32_type);
-        }
-        builder.with_column("val2", int32_type);
-        return builder;
-    };
-    schema_ptr write_schema = get_builder(false).build();
-    schema_ptr read_schema = get_builder(true).build();
-    auto val2_cdef = read_schema->get_column_definition(to_bytes("val2"));
-    auto to_ck = [write_schema] (int ck) {
-        return clustering_key::from_single_value(*write_schema, int32_type->decompose(ck));
-    };
-    auto bytes = int32_type->decompose(int32_t(0));
-    auto pk = partition_key::from_single_value(*write_schema, bytes);
-    auto dk = dht::global_partitioner().decorate_key(*write_schema, pk);
-    mutation partition(write_schema, pk);
-    for (int i = 0; i < 3; ++i) {
-        clustering_key ckey = to_ck(i);
-        partition.partition().apply_insert(*write_schema, ckey, write_timestamp);
-        partition.set_cell(ckey, "val1", data_value{100 + i}, write_timestamp);
-        partition.set_cell(ckey, "val2", data_value{200 + i}, write_timestamp);
-    };
-    auto mt = make_lw_shared<memtable>(write_schema);
-    mt->apply(partition);
-    for (const auto version : all_sstable_versions) {
-        for (auto index_block_size : {1, 128, 64*1024}) {
-            tmpdir dir;
-            sstable_writer_config cfg;
-            cfg.promoted_index_block_size = index_block_size;
-            test_env env;
-            auto sst = env.make_sstable(write_schema,
-                dir.path().string(),
-                1 /* generation */,
-                version,
-                sstables::sstable::format_types::big);
-            sst->write_components(mt->make_flat_reader(write_schema), 1, write_schema, cfg, mt->get_encoding_stats()).get();
-            sst->load().get();
-
-            BOOST_REQUIRE_EXCEPTION(
-                assert_that(sst->read_rows_flat(read_schema))
-                    .produces_partition_start(dk)
-                    .produces_row(to_ck(0), {{val2_cdef, int32_type->decompose(int32_t(200))}})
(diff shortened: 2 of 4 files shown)


haaawk:
@avikivity @tgrabiec @bhalevy please have a look

sstables/mp_row_consumer.hh
```diff
@@ -509,6 +509,11 @@ public:
     //}
     proceed do_consume_cell(bytes_view col_name, int64_t timestamp, int64_t ttl, int64_t expiration, CreateCell&& create_cell) {
         struct column col(*_schema, col_name, timestamp);
+        auto cell_str = sstring(to_sstring_view(col.cell));
+        auto dc = _schema->dropped_columns().find(cell_str);
+        if (!col.cdef && col.cell.size() > 0 && (dc == _schema->dropped_columns().end() || dc->second.timestamp < timestamp)) {
```
bhalevy: How about evaluating dc only `if (!col.cdef && col.cell.size() > 0)`?

sstables/mp_row_consumer.hh
```diff
@@ -576,6 +581,11 @@ public:
     virtual proceed consume_deleted_cell(bytes_view col_name, sstables::deletion_time deltime) override {
         auto timestamp = deltime.marked_for_delete_at;
         struct column col(*_schema, col_name, timestamp);
+        auto cell_str = sstring(to_sstring_view(col.cell));
+        auto dc = _schema->dropped_columns().find(cell_str);
+        if (!col.cdef && col.cell.size() > 0 && (dc == _schema->dropped_columns().end() || dc->second.timestamp < timestamp)) {
```
bhalevy: ditto

sstables/mp_row_consumer.hh
haaawk: I optimised the condition and pulled the code to a single helper function.

sstables/mp_row_consumer.hh
haaawk: done

avikivity:
Why should we fail on unknown column names? The cover letter should explain this.

IMO this is wrong:
1. create table on cluster A
2. write some data
3. delete column from table
4. recreate schema (after column deletion) on cluster B
5. import sstable from cluster A
6. import fails

haaawk:
It's a fair point @avikivity but you can always use initial schema for B population and then remove the column again.

@tgrabiec: you were the author of the issue so maybe you could comment :)

Maybe printing warning to logs instead of an exception would be a better solution?

bhalevy:
> Maybe printing warning to logs instead of an exception would be a better solution?

If we consider this, I think it merits a configuration option to allow the user to decide about the semantics.
If possible, we should set the default to the safe behavior (throwing an exception) at least for new installations (in `conf/scylla.yaml`) and to the current behavior (ignoring / warning) on upgrades (as the hard coded default if missing from `scylla.conf`) for backward compatibility.

avikivity:
> It's a fair point @avikivity but you can always use initial schema for B population and then remove the column again.

You could, if you anticipated the problem. Most people will not.

> 
> Maybe printing warning to logs instead of an exception would be a better solution?

We can do that for mc sstables where the column names are known in advance. We can ignore ka/la, they are obsolete.

haaawk:
> > Maybe printing warning to logs instead of an exception would be a better solution?
> 
> We can do that for mc sstables where the column names are known in advance. We can ignore ka/la, they are obsolete.

At the moment we throw exception in mc already.

avikivity:
Then we should change it to log, to prevent the scenario from above failing.

tgrabiec:
@avikivity In general your scenario won't work, because the information about dropped columns is needed to properly interpret the sstables. If some column was dropped in cluster A and then created again, you need the information about dropped columns to filter out writes in sstables.

If we import data by copying sstables, we should copy the schema sstables as well. Or have a tool for exporting and importing schema properly.

Dropping writes to unknown tables seems very unsafe. A bug could cause compaction to cause data loss. Refs https://github.com/scylladb/scylla/issues/2597.

avikivity:
@glommer you're a copy-sstables-to-migrate kind of guy, what's your take?

glommer:
It's hard. From the user level, we can't really match a schema to an SSTable.

Let's say you delete a column. The existing SSTables are not aware of that, as far as I understand. The user will import the schema through a `DESC` command, which will not contain the updated schema only.

tgrabiec:
@glommer What do you suggest as a way forward? 

Do we agree that importing data by copying the sstables and at the same time importing schema by using `DESC` is in general incorrect?

I think we either need to import data using sstableloader or have a way to migrate full schema definition to another cluster. There already is a way if you want to migrate all tables (copy the schema tables). We could have a command for duping a slice of schema tables for a particular table.

glommer:
I don't agree that using DESC and importing data is incorrect - there is no other way around it and we've done it many times.

Even for backups: Let's say you are taking backups from the cluster and the cluster dies. Our documented procedure is to restore the schema and then load the SSTables.

How is that supposed to work ?

haaawk:
@glommer The is a way around poor work that DESC does. It's enough to read full schema description from system tables with CQL and then write it to the new/restored cluster. We can have a simple tool for that. This way all information about deleted columns would be preserved. It would work with backups too.

tgrabiec:
On Wed, Jul 10, 2019 at 11:38 PM Glauber Costa <notifications@github.com>
wrote:

> I don't agree that using DESC and importing data is incorrect - there is
> no other way around it and we've done it many times.
>

The fact that we've done it many times doesn't make it correct. I gave an
example of when it breaks, and proposed a solution.


> Even for backups: Let's say you are taking backups from the cluster and
> the cluster dies. Our documented procedure is to restore the schema and
> then load the SSTables.
>
> How is that supposed to work ?
>
This procedure is wrong, for the same reasons. Schema should be backed up
by snapshotting schema tables.

>

avikivity:
We could have a version of DESCRIBE that generates the schema with deleted/renamed columns documented, and CQL extensions for deleted/renamed columns:

```
DESCRIBE my_ks.my_tab FULL

CREATE TABLE my_ks.my_tab (
    id INT PRIMARY KEY
    c1 INT,
    c2 INT DELETED AT 01-APR-2019,
    c3 INT RENAMED FROM c4
)
```

That would require migrations to be using our cqlsh against Cassandra, at least when there have been column deletions. Implementing the cql extensions should not be hard.

haaawk:
This won't work with Cassandra clusters because Cassandra won't understand the extension. Did you think about cqlsh extension instead of CQL extension?.

avikivity:
DESCRIBE is implemented in cqlsh so no Cassandra changes are needed, just cqlsh.

The cql changes are needed to import that schema.

avikivity:
to be clear, both cqlsh and cql extensions are needed, the former for DESCRIBE, the latter for CREATE TABLE.

haaawk:
Wouldn't it be easier to just provide a tool that reads schema description using CQL and then uploads it to a different cluster using CQL as well? All by simply modifying system tables.

avikivity:
The tool to read the schema tables exists, it is called cqlsh. It also has the ability to create a schema on a different cluster by executing those commands. 

Modifying system tables doesn't create new tables, because the cluster doesn't perform the side effects. You also have to make the changes on every node.

haaawk:
So it seems that this change is appropriate and we should additionally modify cqlsh and CQL to make it possible for proper schema export/import, right?

avikivity:
I'm not sure we should touch la/ka, since we won't create new sstables with those formats.

glommer:
I am not saying that we need to keep doing things a certain way just because we're already doing it, but I am saying that we need a user-friendly way of doing it.

`DESC SCHEMA` is such a way, system tables is not. For starters, it doesn't work cross versions (while importing SSTables cross versions is fine as long as token ownership is the same), and we have a variety of users that tried to restore system table schema on slack and made an almost unrecoverable mess out of their clusters.

Also once the procedure is out in the open, people keep using it.

I like Avi's idea of enriching `DESC schema`.

tgrabiec:
On Thu, Jul 11, 2019 at 3:50 PM Glauber Costa <notifications@github.com>
wrote:

> I am not saying that we need to keep doing things a certain way just
> because we're already doing it, but I am saying that we need a
> user-friendly way of doing it.
>
> DESC SCHEMA is such a way, system tables is not. For starters, it doesn't
> work cross versions (while importing SSTables cross versions is fine as
> long as token ownership is the same),
>
Can you say more on why doesn't it work? Importing schema tales from older
to a newer version should work, like during regular upgrade.

Importing to a lower version may not, but so is true for data sstables, and
for the new "DESC schema" syntax.

and we have a variety of users that tried to restore system table schema on
> slack and made an almost unrecoverable mess out of their clusters.
>
> What was the cause?

> Also once the procedure is out in the open, people keep using it.
>
Are you saying this as an argument against throwing an exception, because
people using the current procedure would get an exception in case they have
data in sstables which was dropped? It doesn't seem like a big problem,
because if we have a new procedure in place, we can tell them to use the
new procedure, and their problem will be solved.


> I like Avi's idea of enriching DESC schema.
>

glommer:
I have no opinion on whether throwing an exception is the right way to handle this. I am focusing my argument around whether or not it is reasonable to forego `DESC schema`, which I think it's not.

It's a user friendly process that works for both Cassandra and Scylla. Would importing system tables from Cassandra work in Scylla ?

bhalevy:
@haaawk said:
> So it seems that this change is appropriate and we should additionally modify cqlsh and CQL to make it possible for proper schema export/import, right?

@avikivity replied:
> I'm not sure we should touch la/ka, since we won't create new sstables with those formats.

@haaawk change LGTM modulo exception controversy.  IMO it's safer to throw an exception in this case. The suggested follow up project at cqlsh/CQL level to allow migrating dropped columns between clusters will prevent these errors, and until then I assume that a user could resort to sstableloader or to modifying the schema on the destination cluster to create and drop the missing columns as needed.

PR #4658, opened 2019-07-07
init: do not allow replace-address-first-boot for seeds

### LINKED ISSUE #3889: Replace dead node: node fails to join cluster if it's defined as a seed

- [] I have read the disclaimer above, and I am reporting a suspected malfunction in Scylla.

*Installation details*
Cluster size:

Disks: (SSD/HDD, count)

amnonh:
If a node is a seed node, it can not be started with
replace-address-first-boot flag.

This patch will throw a band_configuration_error exception
in this case.

Fixes #3889

Scylla doesn't use pull-requests, please send a patch to the [mailing list](mailto:scylladb-dev@googlegroups.com) instead.
See our [contributing guidelines](../CONTRIBUTING.md) and our [Scylla development guidelines](../HACKING.md) for more information.

If you have any questions please don't hesitate to send a mail to the [dev list](mailto:scylladb-dev@googlegroups.com).

### DIFF
--- a/init.cc
+++ b/init.cc
@@ -155,6 +155,10 @@ void init_ms_fd_gossiper(sharded<gms::gossiper>& gossiper
                 to_string(seeds), listen_address_in, broadcast_address);
         throw std::runtime_error("Use broadcast_address for seeds list");
     }
+    if ((!cfg.replace_address_first_boot().empty() || !cfg.replace_address().empty()) && seeds.count(broadcast_address)) {
+        startlog.error("Bad configuration: replace-address and replace-address-first-boot are not allowed for seed nodes");
+        throw bad_configuration_error();
+    }
     gossiper.local().set_seeds(seeds);
     gossiper.invoke_on_all([cluster_name](gms::gossiper& g) {
         g.set_cluster_name(cluster_name);


asias:
Can you explain in the commit log why "If a node is a seed node, it can not be started with
replace-address-first-boot flag"?

In storage_service::prepare_replacement_info, when the replacing node talks to seeds to get the tokens and host id of the node being replaced, the replacing node will exclude itself from the seed nodes to talk with. It should be able to get the tokens and host id info needed for replace operation. 

There is another option called --replace--address. If we block --replace-address-first-boot, we should --replace-address as well.

slivne:
@asias
- AFAIK nodes that are not seeds will fail to boot if they can't connect to
a seed
- I think that in Larisa's case  the node (that was a seed and we tried to
boot it) was not able to connect to any seeds - since it was a seed it
booted - this is wrong.

asias:
> - AFAIK nodes that are not seeds will fail to boot if they can't connect to a seed

Yes, but in this case, replacing node is one of the seed node. It should be able to talk to other seed nodes to get tokens for the replacing node.

> - I think that in Larisa's case the node (that was a seed and we tried to boot it) was not able to connect to any seeds - since it was a seed it booted - this is wrong.

The replacing node is the only seed node?

Let me explain my view of this problem:

n1, n2, n3
n3 is dead
n4 is replacing n3 with the same ip address of n3, say 127.0.0.3. n4 sets seed node as "127.0.0.1 127.0.0.3".

When n4 boots, it will be able to get the tokens for n3 with no problem.

But the real problem is that, in storage_service::join_token_ring, should_bootstrap will return false because it is a seed node. As a result, it will not use n3's tokens but generates new ones, which of course is wrong.

Just as we should not bootstrap a node as seed, we should not replace a node as seed.

init.cc
```diff
@@ -155,6 +155,11 @@ void init_ms_fd_gossiper(sharded<gms::gossiper>& gossiper
                 to_string(seeds), listen_address_in, broadcast_address);
         throw std::runtime_error("Use broadcast_address for seeds list");
     }
+    if (!cfg.replace_address_first_boot().empty() && seeds.count(broadcast_address)) {
+        startlog.error("Bad configuration: replace-address-first-boot is not allowed for seed nodes");
+        throw bad_configuration_error();
+
```
avikivity: gratuituous extra line

avikivity:
@asias to be even more clear, are you okay with this patch?

asias:
> @asias to be even more clear, are you okay with this patch?

As I mentioned: 

There is another option called --replace--address. If we block --replace-address-first-boot, we should --replace-address as well.

slivne:
@amnonh - please fix, please apply this to --replace-address as well.

amnonh:
I have forced-push V2 changes:
* don't allow replace-address as well
* remove redundant new line

asias:
Can you explain in the commit log why "If a node is a seed node, it can not be started with
replace-address-first-boot flag"?

amnonh:
@asias forced-push a commit message that I hope explains it

asias:
Thanks. Looks good to me.

slivne:
@avikivity - asias approved the pull request can you please re-review

avikivity:
Queued as 399d79fc6f

Issue #4642, opened 2019-07-03
Fix possible stalls in row level repair


asias:
After switching to rpc stream interface, we increased the row buffer
size. Code works on the buffer that do not yield can stall the reactor.
We should futurize or run the code in thread and yield.

avikivity:
In 3.2. Will see how it fares in production first.

avikivity:
Will wait one more week with 3.2.

avikivity:
Aged out.

PR #4664, opened 2019-07-08
Fix possible stalls in row level repair

asias:
After switching to rpc stream interface, we increased the row buffer
size. Code works on the buffer that do not yield can stall the reactor.

This series fixes the issue by futurizing or running the code in thread
and yield.

Fixes: #4642

### DIFF
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -1016,7 +1016,7 @@ private:
         _last_sync_boundary = _current_sync_boundary;
         size_t current_size = row_buf_size();
         return read_rows_from_disk(current_size).then([this, skipped_sync_boundary = std::move(skipped_sync_boundary)] (std::list<repair_row> new_rows) mutable {
-            size_t new_rows_size = boost::accumulate(new_rows, size_t(0), [] (size_t x, const repair_row& r) { return x + r.size(); });
+            size_t new_rows_size = get_repair_rows_size(new_rows);
             size_t new_rows_nr = new_rows.size();
             _row_buf.splice(_row_buf.end(), new_rows);
             std::optional<repair_sync_boundary> sb_max;
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -168,6 +168,7 @@ static thread_local row_level_repair_metrics _metrics;
 static const std::vector<row_level_diff_detect_algorithm>& suportted_diff_detect_algorithms() {
     static std::vector<row_level_diff_detect_algorithm> _algorithms = {
         row_level_diff_detect_algorithm::send_full_set,
+        row_level_diff_detect_algorithm::send_full_set_rpc_stream,
     };
     return _algorithms;
 };
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -938,10 +938,14 @@ private:
     }
 
     // return the combined checksum of rows in _row_buf
-    repair_hash row_buf_csum() {
-        repair_hash combined;
-        boost::for_each(_row_buf, [&combined] (repair_row& r) { combined.add(r.hash()); });
-        return combined;
+    future<repair_hash> row_buf_csum() {
+        return do_with(repair_hash(), [this] (repair_hash& combined) {
+            return do_for_each(_row_buf, [&combined] (repair_row& r) mutable {
+                combined.add(r.hash());
+            }).then([&combined] {
+                return combined;
+            });
+        });
     }
 
     repair_hash do_hash_for_mf(const decorated_key_with_hash& dk_with_hash, const mutation_fragment& mf) {
@@ -1020,13 +1024,15 @@ private:
             size_t new_rows_size = get_repair_rows_size(new_rows);
             size_t new_rows_nr = new_rows.size();
             _row_buf.splice(_row_buf.end(), new_rows);
-            std::optional<repair_sync_boundary> sb_max;
-            if (!_row_buf.empty()) {
-                sb_max = _row_buf.back().boundary();
-            }
-            rlogger.debug("get_sync_boundary: Got nr={} rows, sb_max={}, row_buf_size={}, repair_hash={}, skipped_sync_boundary={}",
-                    new_rows.size(), sb_max, row_buf_size(), row_buf_csum(), skipped_sync_boundary);
-            return get_sync_boundary_response{sb_max, row_buf_csum(), row_buf_size(), new_rows_size, new_rows_nr};
+            return row_buf_csum().then([this, new_rows_size, new_rows_nr, skipped_sync_boundary = std::move(skipped_sync_boundary)] (repair_hash row_buf_combined_hash) {
+                std::optional<repair_sync_boundary> sb_max;
+                if (!_row_buf.empty()) {
+                    sb_max = _row_buf.back().boundary();
+                }
+                rlogger.debug("get_sync_boundary: Got nr={} rows, sb_max={}, row_buf_size={}, repair_hash={}, skipped_sync_boundary={}",
+                        new_rows_nr, sb_max, row_buf_size(), row_buf_combined_hash, skipped_sync_boundary);
+                return get_sync_boundary_response{sb_max, row_buf_combined_hash, row_buf_size(), new_rows_size, new_rows_nr};
+            });
         });
     }
 
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -956,7 +956,7 @@ private:
         return repair_hash(h.finalize_uint64());
     }
 
-    stop_iteration handle_mutation_fragment(mutation_fragment_opt mfopt, size_t& cur_size, std::list<repair_row>& cur_rows) {
+    stop_iteration handle_mutation_fragment(mutation_fragment_opt mfopt, size_t& cur_size, size_t& new_rows_size, std::list<repair_row>& cur_rows) {
         if (!mfopt) {
             return stop_iteration::yes;
         }
@@ -978,6 +978,7 @@ private:
         _metrics.row_from_disk_nr++;
         _metrics.row_from_disk_bytes += r.size();
         cur_size += r.size();
+        new_rows_size += r.size();
         cur_rows.push_back(std::move(r));
         return stop_iteration::no;
     }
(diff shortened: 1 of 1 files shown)


repair/row_level.cc
```diff
@@ -1085,23 +1085,45 @@ class repair_meta {
...
     // Give a set of row hashes, return the corresponding rows
     // If needs_all_rows is set, return all the rows in _working_row_buf, ignore the set_diff
-    std::list<repair_row>
-    get_row_diff(const std::unordered_set<repair_hash>& set_diff, needs_all_rows_t needs_all_rows = needs_all_rows_t::no) {
-        std::list<repair_row> rows;
+    future<std::list<repair_row>>
+    get_row_diff(std::unordered_set<repair_hash> set_diff, needs_all_rows_t needs_all_rows = needs_all_rows_t::no) {
         if (needs_all_rows) {
+            std::list<repair_row> rows;
             if (!_repair_master || _nr_peer_nodes == 1) {
```
avikivity: unused?

repair/row_level.cc
```diff
@@ -1995,7 +2022,7 @@ future<> repair_init_messaging_service_handler(repair_service& rs, distributed<d
             auto from = cinfo.retrieve_auxiliary<gms::inet_address>("baddr");
             return smp::submit_to(src_cpu_id % smp::count, [from, repair_meta_id, set_diff = std::move(set_diff), needs_all_rows] () mutable {
                 auto rm = repair_meta::get_repair_meta(from, repair_meta_id);
-                return rm->get_row_diff_handler(set_diff, repair_meta::needs_all_rows_t(needs_all_rows));
+                return rm->get_row_diff_handler(std::move(set_diff), repair_meta::needs_all_rows_t(needs_all_rows));
             });
```
avikivity: This will cause a lot of cross-cpu frees, because the moved set_diff will be freed on the target cpu, not the source.

Maybe we should wrap set_diff in a foreign_ptr.

repair/row_level.cc
```diff
@@ -832,10 +832,15 @@ class repair_meta {
...
 
     // Get a list of row hashes in _working_row_buf
-    std::unordered_set<repair_hash>
+    future<std::unordered_set<repair_hash>>
     working_row_hashes() {
-        return boost::copy_range<std::unordered_set<repair_hash>>(_working_row_buf |
-                boost::adaptors::transformed([] (repair_row& r) { return r.hash(); }));
+        return do_with(std::unordered_set<repair_hash>(), [this] (std::unordered_set<repair_hash>& hashes) {
+            return do_for_each(_working_row_buf, [&hashes] (repair_row& r) {
+                hashes.emplace(r.hash());
```
avikivity: Note, it can still cause a rare stall when enlarging the hash table. We'll need to work on a non-stalling hash table (will be useful in other places too).

repair/row_level.cc
```diff
@@ -875,8 +875,13 @@ class repair_meta {
 
     future<uint64_t> do_estimate_partitions_on_local_shard() {
         auto sstables = _cf.get_sstables();
-        uint64_t partition_count = boost::accumulate(*sstables, uint64_t(0), [this] (uint64_t x, auto&& sst) { return x + sst->estimated_keys_for_range(_range); });
-        return make_ready_future<uint64_t>(partition_count);
```
avikivity: Even if you have 1000 sstables this shouldn't take a long time. Assuming 30ns per sstable it's just 30us.

repair/row_level.cc
```diff
@@ -928,12 +928,18 @@ class repair_meta {
                && engine().cpu_id() == _master_node_shard_config.shard;
     }
 
-    size_t get_repair_rows_size(const std::list<repair_row>& rows) const {
-        return boost::accumulate(rows, size_t(0), [] (size_t x, const repair_row& r) { return x + r.size(); });
+    future<size_t> get_repair_rows_size(const std::list<repair_row>& rows) const {
+        return do_with(size_t(0), [&rows] (size_t& sz) {
+            return do_for_each(rows, [&sz] (const repair_row& r) mutable {
+                sz += r.size();
```
avikivity: I guess we need a futurized version of accumulate. But of course not in this series.

repair/row_level.cc
asias: Yes, I will remove it. It is not used after switching to the helpers.

repair/row_level.cc
asias: You are right when the number if sstables is small. But if there are 1m sstables, it would be 30ms. I think it does not hurt to futurize this function.

repair/row_level.cc
asias: I agree.

repair/row_level.cc
asias: Do you mean something like this:
```
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -1760,9 +1760,10 @@ static future<stop_iteration> repair_get_row_diff_with_rpc_stream_process_op(
             return make_exception_future<stop_iteration>(std::runtime_error("get_row_diff_with_rpc_stream: Inject error in handler loop"));
         }
         bool needs_all_rows = hash_cmd.cmd == repair_stream_cmd::needs_all_rows;
-        return smp::submit_to(src_cpu_id % smp::count, [from, repair_meta_id, needs_all_rows, set_diff = std::move(current_set_diff)] {
+        auto fp = make_foreign(std::make_unique<std::unordered_set<repair_hash>>(std::move(current_set_diff)));
+        return smp::submit_to(src_cpu_id % smp::count, [from, repair_meta_id, needs_all_rows, fp = std::move(fp)] {
             auto rm = repair_meta::get_repair_meta(from, repair_meta_id);
-            return rm->get_row_diff_handler(std::move(set_diff), repair_meta::needs_all_rows_t(needs_all_rows));
+            return rm->get_row_diff_handler(std::move(*fp), repair_meta::needs_all_rows_t(needs_all_rows));
         }).then([sink] (repair_rows_on_wire rows_on_wire) mutable {
             if (rows_on_wire.empty()) {
                 return sink(repair_row_on_wire_with_cmd{repair_stream_cmd::end_of_current_rows, repair_row_on_wire()});

```

repair/row_level.cc
asias: I created https://github.com/scylladb/scylla/issues/4688 to track this issue.

repair/row_level.cc
asias: @avikivity

repair/row_level.cc
avikivity: If there are 1M sstables in a shard, the node will OOM.

repair/row_level.cc
avikivity: If you move *fp then you get the same delete-on-wrong-shard problem. foreign_ptr only takes care of data it owns.

repair/row_level.cc
asias: It might, but a shard can have a lot of memory. Anyway, do you insist we should not furturize do_estimate_partitions_on_local_shard?

repair/row_level.cc
asias: Added "repair: Wrap with foreign_ptr to avoid cross cpu free" in new version.

asias:
New version added. Please review.

asias:
@avikivity Do you have any more comments?  Patch "repair: Wrap with foreign_ptr to avoid cross cpu free" has the foreign_ptr stuff.

repair/row_level.cc
avikivity: no

asias: Ok. I will mark it resolved.

avikivity:
Any changes apart from that patch?

asias:
one small change: removing the unused std::list<repair_row> rows.

asias:
@avikivity ping

asias:
@avikivity Anything I can do to get this merged?

avikivity:
@asias I'll review the updated patch.

repair/row_level.cc
```diff
@@ -1799,9 +1804,14 @@ static future<stop_iteration> repair_put_row_diff_with_rpc_stream_process_op(
         return make_ready_future<stop_iteration>(stop_iteration::no);
     } else if (row.cmd == repair_stream_cmd::end_of_current_rows) {
         rlogger.trace("Got repair_rows_on_wire from peer={}, got end_of_current_rows", from);
-        return smp::submit_to(src_cpu_id % smp::count, [from, repair_meta_id, rows = std::move(current_rows)] () mutable {
+        auto fp = make_foreign(std::make_unique<repair_rows_on_wire>(std::move(current_rows)));
+        return smp::submit_to(src_cpu_id % smp::count, [from, repair_meta_id, fp = std::move(fp)] () mutable {
             auto rm = repair_meta::get_repair_meta(from, repair_meta_id);
-            return rm->put_row_diff_handler(std::move(rows), from);
+            if (fp.get_owner_shard() == engine().cpu_id()) {
+                return rm->put_row_diff_handler(std::move(*fp), from);
```
avikivity: This doesn't fix anything. *fp will be destroyed on its owner shard, but it's empty after you moved away from it. The contents will be destroyed on the current shard, after put_row_diff_handler() returns (unless it also moves it).

asias: See the else branch which does not move *fp. The if branch is an optimization where the submit_to submits to the *same* shard.

asias: @avikivity ?

asias: No, they are not. See: 

```
1616     // RPC handler 
1617     future<repair_rows_on_wire> get_row_diff_handler(std::unordered_set<repair_hash> set_diff, needs_all_rows_t needs_all_rows) {
1618         return with_gate(_gate, [this, set_diff = std::move(set_diff), needs_all_rows] () mutable {
1619             return get_row_diff(std::move(set_diff), needs_all_rows).then([this] (std::list<repair_row> row_diff) {
1620                 return to_repair_rows_on_wire(std::move(row_diff));
1621             });
1622         });
1623     }
```

```
1736     // RPC handler 
1737     future<> put_row_diff_handler(repair_rows_on_wire rows, gms::inet_address from) {
1738         return with_gate(_gate, [this, rows = std::move(rows)] () mutable {
1739             return apply_rows_on_follower(std::move(rows));
1740         });     
1741     }    
```

They are changed in this series. E.g., in "repair: Futurize get_row_diff to avoid stall".

asias: Ping.

avikivity: Queued

PR #4667, opened 2019-07-08
Auto-expand replication_factor for NetworkTopologyStrategy

### LINKED ISSUE #4210: Auto-expand replication_factor for NetworkTopologyStrategy
Scylla version (or git commit hash): 3.0

Apache Cassandra add an option for using `replication_factor` with `NetworkTopologyStrategy` so, **each** DC have the same RF (for example 3):

```
 CREATE KEYSPACE test WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 3}
```
https://issues.apache.org/jira/browse/CASSANDRA-14303

Pros:
* decoupled DC names from the schema. 
* no need to update RF when adding new DCs
* C* comptability

Cons:
* easy to miss the RF will apply to all clusters.

kbr-:
If the user supplies the 'replication_factor' to the 'NetworkTopologyStrategy' class,
it will expand into a replication factor for each existing DC for their convenience.
Resolves #4210.

### DIFF
--- a/cql3/statements/ks_prop_defs.cc
+++ b/cql3/statements/ks_prop_defs.cc
@@ -46,6 +46,53 @@ namespace cql3 {
 
 namespace statements {
 
+static std::map<sstring, sstring> prepare_options(
+        const sstring& strategy_class,
+        const locator::token_metadata& tm,
+        std::map<sstring, sstring> options,
+        const std::map<sstring, sstring>& old_options = {}) {
+    options.erase(ks_prop_defs::REPLICATION_STRATEGY_CLASS_KEY);
+
+    if (strategy_class != "NetworkTopologyStrategy") {
+        return options;
+    }
+
+    // For users' convenience, expand the 'replication_factor' option into a replication factor for each DC.
+    // If the user simply switches from another strategy without providing any options,
+    // but the other strategy used the 'replication_factor' option, it will also be expanded.
+    // See issue CASSANDRA-14303.
+
+    sstring rf;
+    auto it = options.find(ks_prop_defs::REPLICATION_FACTOR_KEY);
+    if (it != options.end()) {
+        // Expand: the user explicitly provided a 'replication_factor'.
+        rf = it->second;
+        options.erase(it);
+    } else if (options.empty()) {
+        auto it = old_options.find(ks_prop_defs::REPLICATION_FACTOR_KEY);
+        if (it != old_options.end()) {
+            // Expand: the user switched from another strategy that specified a 'replication_factor'
+            // and didn't provide any additional options.
+            rf = it->second;
+        }
+    }
+
+    if (!rf.empty()) {
+        // We keep previously specified DC factors for safety.
+        for (const auto& opt : old_options) {
+            if (opt.first != ks_prop_defs::REPLICATION_FACTOR_KEY) {
+                options.insert(opt);
+            }
+        }
+
+        for (const auto& dc : tm.get_topology().get_datacenter_endpoints()) {
+            options.emplace(dc.first, rf);
+        }
+    }
+
+    return options;
+}
+
 void ks_prop_defs::validate() {
     // Skip validation if the strategy class is already set as it means we've alreayd
     // prepared (and redoing it would set strategyClass back to null, which we don't want)
@@ -74,20 +121,23 @@ std::optional<sstring> ks_prop_defs::get_replication_strategy_class() const {
     return _strategy_class;
 }
 
-lw_shared_ptr<keyspace_metadata> ks_prop_defs::as_ks_metadata(sstring ks_name) {
-    auto options = get_replication_options();
-    options.erase(REPLICATION_STRATEGY_CLASS_KEY);
-    return keyspace_metadata::new_keyspace(ks_name, get_replication_strategy_class().value(), options, get_boolean(KW_DURABLE_WRITES, true));
+lw_shared_ptr<keyspace_metadata> ks_prop_defs::as_ks_metadata(sstring ks_name, const locator::token_metadata& tm) {
+    auto sc = get_replication_strategy_class().value();
+    return keyspace_metadata::new_keyspace(ks_name, sc,
+            prepare_options(sc, tm, get_replication_options()), get_boolean(KW_DURABLE_WRITES, true));
 }
 
-lw_shared_ptr<keyspace_metadata> ks_prop_defs::as_ks_metadata_update(lw_shared_ptr<keyspace_metadata> old) {
-    auto options = get_replication_options();
-    options.erase(REPLICATION_STRATEGY_CLASS_KEY);
+lw_shared_ptr<keyspace_metadata> ks_prop_defs::as_ks_metadata_update(lw_shared_ptr<keyspace_metadata> old, const locator::token_metadata& tm) {
+    std::map<sstring, sstring> options;
+    const auto& old_options = old->strategy_options();
     auto sc = get_replication_strategy_class();
-    if (!sc) {
+    if (sc) {
+        options = prepare_options(*sc, tm, get_replication_options(), old_options);
+    } else {
         sc = old->strategy_name();
-        options = old->strategy_options();
+        options = old_options;
     }
+
     return keyspace_metadata::new_keyspace(old->name(), *sc, options, get_boolean(KW_DURABLE_WRITES, true));
 }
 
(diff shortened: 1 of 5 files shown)


kbr-:
@penberg @avikivity  could you review please?

cql3/statements/ks_prop_defs.cc
```diff
@@ -74,20 +75,64 @@ std::optional<sstring> ks_prop_defs::get_replication_strategy_class() const {
...
-lw_shared_ptr<keyspace_metadata> ks_prop_defs::as_ks_metadata(sstring ks_name) {
+std::map<sstring, sstring> ks_prop_defs::prepare_options(const sstring& strategy_class, const std::map<sstring, sstring>& old_options = {}) {
     auto options = get_replication_options();
     options.erase(REPLICATION_STRATEGY_CLASS_KEY);
-    return keyspace_metadata::new_keyspace(ks_name, get_replication_strategy_class().value(), options, get_boolean(KW_DURABLE_WRITES, true));
+
+    if (strategy_class != "NetworkTopologyStrategy") {
+        return options;
+    }
+
```
avikivity: Seems wrong to add strategy-specific code here. Maybe add a virtual abstract_replication_strategy::adjust_options()?

cql3/statements/ks_prop_defs.cc
```diff
@@ -74,20 +75,64 @@ std::optional<sstring> ks_prop_defs::get_replication_strategy_class() const {
...
+    auto it = options.find(REPLICATION_FACTOR_KEY);
+    if (it != options.end()) {
+        rf = it->second;
+        options.erase(it);
+    } else if (options.empty()) {
+        auto it = old_options.find(REPLICATION_FACTOR_KEY);
+        if (it != old_options.end()) {
+            rf = it->second;
+        }
+    }
```
avikivity: Why do you touch old_options here? Wouldn't the have their REPLICATION_FACTOR_KEY erased?

cql3/statements/ks_prop_defs.cc
```diff
@@ -74,20 +75,64 @@ std::optional<sstring> ks_prop_defs::get_replication_strategy_class() const {
...
+
+    if (!rf.empty()) {
+        // We keep previously specified DC factors for safety.
+        for (const auto& opt: old_options) {
+            if (opt.first != REPLICATION_FACTOR_KEY) {
+                options.insert(opt);
+            }
+        }
+
+        for (const auto& dc: service::get_local_storage_service().get_token_metadata().
```
avikivity: :(

cql3/statements/ks_prop_defs.cc
kbr-: Thanks for the review!

Adding a virtual method to `abstract_replication_strategy` was my initial idea too. Unfortunately, that would probably require a lot of refactoring. Let me explain.

Currently, to create an `abstract_replication_strategy` (abbreviated as `a_r_s` from now on), we need to pass the replication strategy options.
1. each time `a_r_s` is instantiated in our code, options are passed down from some `keyspace_metadata` instance.
2. `keyspace_metadata` is created in many places. The options used to instantiate a `keyspace_metadata` can come from:
- a user performing a CQL query, in `ks_prop_defs::as_ks_metadata` and `ks_prop_defs::as_ks_metadata_update`, in cql3/statements/ks_prop_defs.cc
- a user accessing the DB through Thrift, in `thrift_handler::keyspace_from_thrift`, in thrift/handler.cc
- some old keyspace_metadata, e.g. `database::update_keyspace` in database.cc
- a developer, e.g. `service::create_keyspace_if_missing`, in auth/service.cc
3. there is only one case when the options should be expanded/adjusted: the CQL case. This is when Cassandra does it. Remember, expanding the replication_factor is a UX/UI convenience thing for the CQL API.
4. since `a_r_s` is sometimes created with the options expanded and sometimes without, the decision to adjust the options must be made before `a_r_s` is created.
5. therefore adjusting the options can't be made in a `a_r_s` (non-static) method, as it requires an `a_r_s` instance, which we cannot yet create. QED.

I could try to refactor the code somehow so that a virtual method in `a_r_s` could be used, or come up with a different "generic"/"modular" solution. But is it worth it? To answer this question, I looked into how Cassandra does it:
https://github.com/apache/cassandra/commit/12aa45434a6d5913a2a3db91a26a8d2f7081042a
it obtains a `Class<? extends AbstractReplicationStrategy>` from the given strategy string and then uses reflection to check if it has a `prepareOptions` static method.
Great, so it's a nice generic solution: every ReplicationStrategy can have its own `prepareOptions` static method; if it doesn't then the options stay unchanged. However, the patch was made a year ago, and the only class which has a `prepareOptions` method to this day is, of course, `NetworkTopologyStrategy`, which does the replication_factor expansion. For me it's a great example of over-engineering, or [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). I prefer the simple approach.

cql3/statements/ks_prop_defs.cc
kbr-: Would `const auto& datacenters = service::stuff()` before the loop and `for (const auto& dc: datacenters)` change the ":(" into a ":)"?

cql3/statements/ks_prop_defs.cc
kbr-: Only in the case when the strategy class was previously NetworkTopologyStrategy would the old options have had their REPLICATION_FACTOR_KEY erased.

This comment explains it:
```
// If the user simply switches from another strategy without providing any options,
// but the other strategy used the 'replication_factor' option, it will also be expanded.
```
So if, for example, the previous strategy was SimpleStrategy, and we switch directly to NetworkTopologyStrategy using an ALTER statement without any other options than the class (`else if (options.empty())`, 'class' was erased), replication_factor will be expanded. The tests check for this case.

There was a discussion in the Cassandra issue whether this is the correct approach, and they decided to go with it:
https://issues.apache.org/jira/browse/CASSANDRA-14303?focusedCommentId=16713793&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16713793

cql3/statements/ks_prop_defs.cc
avikivity: We could replace the convoluted way we create the replication strategy (which in turn was derived from Java) by registering a factory class with two virtual methods, one to massage the parameters and one to create the class. But I agree it's overkill.

cql3/statements/ks_prop_defs.cc
avikivity: The :( refers to accessing a global object (get_local_storage_service()). Is there away to get the token_metadata without a global reference? We declared a war on globals some time ago as they hide dependencies and make unit testing difficult. If not, we'll keep the :( and pass it to whoever removes that global eventually.

btw, our coding style puts spaces before and after ":" in range for.

kbr-:
v2:
- rebase
- make prepare_options into a free-standing function (not a method of ks_prop_defs), local to the ks_prop_defs.cc unit: makes it clearer that it's a pure function, easier to reason about, and doesn't pollute the header
- add some comments in prepare_options
- call get_local_storage_service() in alter_keyspace_statement::announce_migration instead of prepare_options; pass token_metadata down to as_ks_metadata[_update], then prepare_options
- spaces around : in range for

cql3/statements/ks_prop_defs.cc
kbr-: >  Is there a way to get the token_metadata without a global reference?

There is, but not without some "hard" work. Here's what I'd do:

1. provide a reference to _the_storage_service in qp.start in main.cc
2. store service::storage_service& _service in query_processor, similarly to service::storage_proxy& _proxy
3. pass _service to cql_statement::execute in query_processor::process and friends
4. pass the service further down to schema_altering_statement::announce_migration in execute

Should I open an issue? I don't think those refactors belong in this pull request.
For now, I moved the get_local_storage_service() call higher up in the call stack, to announce_migration. This is where other globals are accessed and should make future refactoring easier.

By the way: a reference to storage_proxy is already being stored in query_processor and passed down to execute, announce_migration, and validate in schema_altering_statement(s). But sometimes it gets ignored and get_local_storage_proxy() gets called nevertheless... (see, for example, alter_keyspace_statement::announce_migration). I didn't change it in this PR so to not mix things up.

cql3/statements/ks_prop_defs.cc
kbr-: Added more comments to explain.

kbr-:
Rebased.

PR #4682, opened 2019-07-09
Updatable timeout config

### LINKED ISSUE #2199: nodetool gettimeout / settimeout
*Installation details*
Scylla version (or git commit hash): 1.7

support `nodetool gettimeout` and `nodetool settimeout`
timeout types:

- read
- range
- write
- counterwrite
- cascontention
- truncate
- streamingsocket
- misc, such as general rpc_timeout_in_ms

examples
```
nodetool settimeout write 100
nodetool gettimeout write
```

https://issues.apache.org/jira/browse/CASSANDRA-10953

amnonh:
This series adds allows changing the different timeout configuration during runtime.
The following configuration:
range_request_timeout_in_ms
read_request_timeout_in_ms
counter_write_request_timeout_in_ms
cas_contention_timeout_in_ms
truncate_request_timeout_in_ms
write_request_timeout_in_ms
request_timeout_in_ms to be updatable.

Can now be updated in the configuration file and using SIGHUP will be updated.

Fixes #2199


denesb:
Please replace the pull request template text with whatever you would put in the cover letter.

service/storage_service.cc
```diff
@@ -118,14 +118,15 @@ distributed<storage_service> _the_storage_service;
...
-    tc.read_timeout = cfg.read_request_timeout_in_ms() * 1ms;
-    tc.write_timeout = cfg.write_request_timeout_in_ms() * 1ms;
-    tc.range_read_timeout = cfg.range_request_timeout_in_ms() * 1ms;
-    tc.counter_write_timeout = cfg.counter_write_request_timeout_in_ms() * 1ms;
-    tc.truncate_timeout = cfg.truncate_request_timeout_in_ms() * 1ms;
-    tc.cas_timeout = cfg.cas_contention_timeout_in_ms() * 1ms;
-    tc.other_timeout = cfg.request_timeout_in_ms() * 1ms;
+    timeout_config tc{
+        .read_timeout = utils::updateable_value<db::timeout_clock::duration>(cfg.read_request_timeout_in_ms() * 1ms),
+        .write_timeout = utils::updateable_value<db::timeout_clock::duration>(cfg.write_request_timeout_in_ms() * 1ms),
```
avikivity: Designated initializers are not in C++17 (they're a gcc extension), let's only use them with C++20.


How is this updateable? by doing the multipication, you broke the link with the updateable_value_source.

auth/common.cc
```diff
@@ -103,7 +103,7 @@ future<> wait_for_schema_agreement(::service::migration_manager& mm, const datab
 }
 
 const timeout_config& internal_distributed_timeout_config() noexcept {
-    static const auto t = 5s;
+    static const auto t = utils::updateable_value<db::timeout_clock::duration>(5s);
```
denesb: Shouldn't this change go together with the previous one? Doesn't changing the values to be updateable require changes to the users to compile?

service/storage_service.cc
amnonh: I'll remove the designated initialization, I tried to keep the code as close as possible to the original and was happy that it compiled for me...

I'll have  second look to be sure or update the implementation

auth/common.cc
amnonh: it does

amnonh:
I have forced-push V2.
Changes:
* Remove Designated initializers.
* The timeout_config struct holds now uint32_t instead of duration and the typecast is done just before used. This is so the link to the configuration will remain.

auth/common.cc
amnonh: If you mean the configuration change, then no, if a configuration is updatable you can still use it as before.

service/storage_service.cc
```diff
@@ -119,13 +119,14 @@ distributed<storage_service> _the_storage_service;
...
     timeout_config tc;
-    tc.read_timeout = cfg.read_request_timeout_in_ms() * 1ms;
-    tc.write_timeout = cfg.write_request_timeout_in_ms() * 1ms;
-    tc.range_read_timeout = cfg.range_request_timeout_in_ms() * 1ms;
-    tc.counter_write_timeout = cfg.counter_write_request_timeout_in_ms() * 1ms;
-    tc.truncate_timeout = cfg.truncate_request_timeout_in_ms() * 1ms;
-    tc.cas_timeout = cfg.cas_contention_timeout_in_ms() * 1ms;
-    tc.other_timeout = cfg.request_timeout_in_ms() * 1ms;
+    tc.read_timeout = cfg.read_request_timeout_in_ms;
+    tc.write_timeout = cfg.write_request_timeout_in_ms;
```
avikivity: Since those values are no longer durations, their names should have _ms suffixes to indicate the units.

Even better, please see if you can make the configuration item store a duration instead of an integer. Maybe more glue logic is needed, but in the end it will save a lot of unit mistakes.

tracing/trace_keyspace_helper.cc
```diff
@@ -60,7 +60,13 @@ const sstring trace_keyspace_helper::NODE_SLOW_QUERY_LOG("node_slow_log");
 const sstring trace_keyspace_helper::NODE_SLOW_QUERY_LOG_TIME_IDX("node_slow_log_time_idx");
 
 timeout_config tracing_db_timeout_config {
-    5s, 5s, 5s, 5s, 5s, 5s, 5s,
+    utils::updateable_value<uint32_t>(5 * 1000),
+    utils::updateable_value<uint32_t>(5 * 1000),
```
avikivity: Again we lost the units, would be better to write 5s / 1ms to make it clear.

amnonh:
@avikivity I forced-pushed a newer version.
I've used a new struct that can read from the configuration uint and return duration.
It's constructor uses duration so the different places that create their own timeout duration would work with duration objects and not uint

db/config.cc
```diff
@@ -153,6 +166,18 @@ struct convert<db::config::seed_provider_type> {
     }
 };
 
+template<>
+struct convert<db::duration_in_ms_type> {
+    static bool decode(const Node& node, db::duration_in_ms_type& v) {
+        uint32_t tmp;
+        if (!convert<uint32_t>::decode(node, tmp)) {
+            return false;
```
avikivity: It will be nice (one day) do add support for unit suffixes so the user can write 2s or 2000ms and get the same result.

db/config.cc
```diff
@@ -714,6 +739,20 @@ void config_file::named_value<db::config::seed_provider_type>::add_command_line_
...
+                const std::string_view& name, const std::string_view& desc) {
+    init((hyphenate(name)).data(),
+                    value_ex<uint32_t>()->notifier(
+                                    [this](uint32_t new_duration) {
+                                        auto old_duration = operator()();
+                                        old_duration._duration = new_duration;
+                                        set(std::move(old_duration), config_source::CommandLine);
+                                    }),
+                    desc.data());
+}
```
avikivity: wild indentation

db/config.cc
```diff
@@ -817,3 +860,20 @@ sstring config_value_as_json(const std::unordered_map<sstring, log_level>& v) {
 }
 
 }
+
+db::duration_in_ms_type::duration_in_ms_type(const timeout_clock::duration &d) :
+        _duration(std::chrono::duration_cast<std::chrono::milliseconds>(d).count()) {
```
avikivity: Why a new type and not std::chrono::milliseonds?

db/config.hh
```diff
@@ -61,6 +62,29 @@ struct seed_provider_type {
     }
 };
 
+/*
+ * This type is a glue between configuration in ms
+ * and duration
+ */
+struct duration_in_ms_type {
```
avikivity: Why not use duration directly?

api/storage_proxy.cc
```diff
@@ -143,7 +143,7 @@ void set_storage_proxy(http_context& ctx, routes& r) {
     });
 
     sp::get_rpc_timeout.set(r, [&ctx](const_req req)  {
-        return ctx.db.local().get_config().request_timeout_in_ms()/1000.0;
+        return ctx.db.local().get_config().request_timeout_in_ms().count()/1000.0;
     });
```
avikivity: A nicer way to write it: request_timeout_in_ms() / 1ms.

avikivity: (safer too in case the duration ratio<> changes)

amnonh: ok

db/config.hh
amnonh: The issue is that the configuration is in ms, so the config needs to create some kind of glue between that (number of ms) and duration.

I think that translating int directly to duration is confusing. when you will look at the code, you'll see `duration` and you need to guess what are the units.

Using this kind of type makes it clearer.

db/config.cc
amnonh: I thought about it and if we would have broken backward compatibility I would use duration is type (vs adding a new type).
I decided against it because in yaml, 1000 and 1s are not the same.

But it's definitely an option.

amnonh:
@avikivity I have push force a newer version. I kept the dedicated structure and add an explenation to the commit message.
I fixed the indentation mark and used duration arithmetics when applicable to make the code more robust to changes.

db/config.hh
```diff
@@ -61,6 +62,32 @@ struct seed_provider_type {
...
+    duration_in_ms_type() = default;
+    duration_in_ms_type(const duration_in_ms_type&) = default;
+    duration_in_ms_type(uint32_t n)
+            : _duration(n) {
+    }
+    duration_in_ms_type(const timeout_clock::duration &d);
+    duration_in_ms_type& operator=(const timeout_clock::duration &d);
+    uint32_t _duration = 0;
```
denesb: We usually list data members on the top, not intermixed with member methods.

amnonh: I've forced-push a version that changes the order

slivne:
@avikivity ping

slivne:
@avikivity ping

slivne:
@avikivity ping

slivne:
@avikivity ping

db/config.cc
avikivity: Why?

db/config.hh
avikivity: std::chrono::milliseconds names the units directly (and duration has an std::ratio parameter that defines them)

avikivity:
@slivne there are unaddressed comments.

amnonh:
@avikivity I have forced push a new version, it now uses std::chrono::milliseconds instead of the new type.

I have also rebased to fix some conflicts.

slivne:
@avikivity ping - lets try to get this in for 3.2

slivne:
@avikivity ping this has a blessing from botond - we want to allow timer config changes without restarts

slivne:
@avikivity ping this has a blessing from botond - we want to allow timer config changes without restarts

glommer:
@amnonh there are conflicts in this branch. Please let's get this moving.

amnonh:
I have pull-rebased, solved the conflicts made it compile again and forced pushed the result

slivne:
@avikivity ping this has a blessing from botond - we want to allow timer config changes without restarts

slivne:
@avikivity ping this has a blessing from botond - we want to allow timer config changes without restarts, can you please provide info what you are missing (rebasing over and over with no feedback doesn't help)

slivne:
@avikivity ping

slivne:
@avikivity ping

slivne:
@avikivity ping

nyh:
@amnonh I found this abandoned PR. I think it's no longer relevant, e.g., I randomly looked at range_request_timeout_in_ms and it seems that sometime in last 5 years it was already made live-updatable. Can you please check and if no longer relevant - close this PR? Thanks.

PR #4687, opened 2019-07-10
lsa: make sure align_up_for_asan() doesn't cause reads past end of segment

### LINKED ISSUE #4653: lsa: invalid memory accesses caused by `align_up_for_asan()`
Since e071445373f3587d4f4656eb7d1fa66985b4616d (Merge "More precise poisoning in logalloc" from Rafael) LSA, in debug mode, requires objects to be aligned at 8-byte boundary so that it can maximise the coverage that AddressSanitizer provides. LSA, being an allocator is doing some pointer arithmetic, and where there's pointer arithmetic there are bugs.

In particular, `align_up_for_asan()` allowed access past the end of a range in loops like this one:

```c++
auto pos = seg->at<const char>(0);
while (pos < seg->at<const char>(segment::size)) {
    align_up_for_asan(pos);
    /* ... something that dereferences `pos` ... */
}
```

Only debug build is susceptible.

pdziepak:
In debug mode the LSA needs objects to be 8-byte aligned in order to
maximise coverage from the AddressSanitizer.

Usually `close_active()` creates a dummy objects that covers the end of
the segment being closed. However, it the last real objects ends in the
last eight bytes of the segment then that dummy won't be created because
of the alignment requirements. This broke exit conditions on loops
trying to read all objects in the segment and caused them to attempt to
dereference address at the end of the segment. This patch fixes that.

Fixes #4653.

### DIFF
--- a/utils/logalloc.cc
+++ b/utils/logalloc.cc
@@ -60,8 +60,8 @@ namespace debug {
 constexpr size_t logalloc_alignment = 8;
 }
 template<typename T>
-static void align_up_for_asan(T& val) {
-    val = align_up(val, size_t(8));
+[[nodiscard]] static T align_up_for_asan(T val) {
+    return align_up(val, size_t(8));
 }
 template<typename T>
 void poison(const T* addr, size_t size) {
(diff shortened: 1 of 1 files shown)


espindola:
LGTM

Thanks,
Rafael

Issue #4693, opened 2019-07-11
Multiple flat mutation reader can cause deadlock


asias:
Start n1, n2
Create ks with rf = 2
Run repair on n2
Stop n2 in the middle of repair
n1 will notice n2 is DOWN, gossip handler will remove repair instance with n2 which calls remove_repair_meta().

Inside remove_repair_meta(), we have:

```
1        return parallel_for_each(*repair_metas, [repair_metas] (auto& rm) {
2            return rm->stop();
3        }).then([repair_metas, from] {
4            rlogger.debug("Removed all repair_meta for single node {}", from);
5        });
```

Since 3.1, we start 16 repair instances in parallel which will create 16 readers.The reader semaphore is 10. 

At line 2, it calls 

```
6    future<> stop() {
7       auto gate_future = _gate.close();
8       auto writer_future = _repair_writer.wait_for_writer_done();
9       return when_all_succeed(std::move(gate_future), std::move(writer_future));
10    }
```

The gate protects the reader to read data from disk:

```
11 with_gate(_gate, [] {
12   read_rows_from_disk
13        return _repair_reader.read_mutation_fragment() --> calls reader() to read data
14 })
```

So line 7 won't return until all the 16 readers return from the call of reader().

The problem is, the reader won't release the reader semaphore until the reader is destroyed!
So, even if 10 out of the 16 readers have finished reading, they won't  release the semaphore. As a result, the stop() hangs forever. 

To fix in short term, we can delete the reader, aka, drop the the repair_meta object once it is stopped. 

To fix in long term, we should make the flat mutation reader release the semaphore when it is done reading.

asias:
@denesb FYI

denesb:
Related: https://github.com/scylladb/scylla/issues/3866. Though we can maybe have a simpler solution here.

slivne:
patch merged to next

asias:
@slivne Please note, the fix in repair code is a workaround. To fix long term, we need to fix the flat mutation reader. Please assign this to Botond.

slivne:
@asias it would have been better to create a separate issue for this specific case for repair and leave this generic issue open ... for botond to tackle :)

asias:
> @asias it would have been better to create a separate issue for this specific case for repair and leave this generic issue open ... for botond to tackle :)

As you can see, my repair patch Refs this issue instead of closing it.

denesb:
[...]

PR #4694, opened 2019-07-11
repair: Avoid deadlock in remove_repair_meta

asias:
Start n1, n2
Create ks with rf = 2
Run repair on n2
Stop n2 in the middle of repair
n1 will notice n2 is DOWN, gossip handler will remove repair instance
with n2 which calls remove_repair_meta().

Inside remove_repair_meta(), we have:

```
1        return parallel_for_each(*repair_metas, [repair_metas] (auto& rm) {
2            return rm->stop();
3        }).then([repair_metas, from] {
4            rlogger.debug("Removed all repair_meta for single node {}", from);
5        });
```

Since 3.1, we start 16 repair instances in parallel which will create 16
readers.The reader semaphore is 10.

At line 2, it calls

```
6    future<> stop() {
7       auto gate_future = _gate.close();
8       auto writer_future = _repair_writer.wait_for_writer_done();
9       return when_all_succeed(std::move(gate_future), std::move(writer_future));
10    }
```

The gate protects the reader to read data from disk:

```
11 with_gate(_gate, [] {
12   read_rows_from_disk
13        return _repair_reader.read_mutation_fragment() --> calls reader() to read data
14 })
```

So line 7 won't return until all the 16 readers return from the call of
reader().

The problem is, the reader won't release the reader semaphore until the
reader is destroyed!
So, even if 10 out of the 16 readers have finished reading, they won't
release the semaphore. As a result, the stop() hangs forever.

To fix in short term, we can delete the reader, aka, drop the the
repair_meta object once it is stopped.

Refs: #4693

### DIFF
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -780,7 +780,9 @@ public:
             }
         }
         return parallel_for_each(*repair_metas, [repair_metas] (auto& rm) {
-            return rm->stop();
+            return rm->stop().then([&rm] {
+                rm = {};
+            });
         }).then([repair_metas, from] {
             rlogger.debug("Removed all repair_meta for single node {}", from);
         });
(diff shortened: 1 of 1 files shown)


denesb:
Reviewed-by: Botond Dénes <bdenes@scylladb.com>

slivne:
@tgrabiec / @avikivity can one of you please review its for 3.1

PR #4684, opened 2019-07-09
Handle termination signals during streaming

avikivity:
In b19792405faee91e, we changed the shutdown process not to rely on the global reactor-defined exit, but instead added a local variable to hold the shutdown state. However, we did not propagate that state everywhere, and now streaming processes are not able to abort.

Fix that by enhancing stop_signal with a sharded<abort_source> member that can be propagated to services. Propagate it to storage_service and thence to boot_strapper and range_streamer so that streaming processes can be aborted.

Fixes #4674
Fixes #4501 

Tests: unit (dev), manual bootstrap test

PR #4697, opened 2019-07-11
storage_service: Use sleep_abortable instead of sleep

asias:
Make the sleep abortable so that it is able to break the loop during
shutdown.

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -635,13 +635,13 @@ void storage_service::join_token_ring(int delay) {
         auto& gossiper = gms::get_gossiper().local();
         // first sleep the delay to make sure we see *at least* one other node
         for (int i = 0; i < delay && gossiper.get_live_members().size() < 2; i += 1000) {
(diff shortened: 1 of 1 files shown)


avikivity:
I have almost the same patch (prepared as part of https://github.com/scylladb/scylla/pull/4684) but I pass the abort_source instead of relying on a global termination signal.

asias:
Nice timing ;-)  But I saw you did not modify the code where calls sleep(). Is this patch still needed?

avikivity:
I have an extra patch that I did not yet push that adds more sleep_abortable()s, but decided to fix the critical stuff first.

asias:
Ok. I will handle off the patch to you if you need any of the changes in this patch.

avikivity:
Actually it is better that you do it because you know better which sleeps can be aborted and which do not. But it should be done after #4684 and use the abort_source from there.

asias:
OK.

asias:
Updated:
- Use _abort_source inside storage_service:

PR #4698, opened 2019-07-11
repair: Log repair id in check_failed_ranges

asias:
Add the word `id` before the repair id in the log. It makes the log
easier to figure out what the number stands for.


avikivity:
Queued as 722ab3bb65b78912b779a7a78fc90488beee705a

PR #4710, opened 2019-07-15
storage_service: Replicate and advertise tokens early in the boot up process

### LINKED ISSUE #4709: [shard 15] storage_proxy - Failed to apply mutation from : std::runtime_error (sorted_tokens is empty in first_token_index!) error message
Scylla version (or git commit hash): 3.0.6

```
[shard 2] token_metadata - sorted_tokens is empty in first_token_index! 
[shard 5] token_metadata - sorted_tokens is empty in first_token_index! 
[shard 8] token_metadata - sorted_tokens is empty in first_token_index! 
[shard 1] token_metadata - sorted_tokens is empty in first_token_index! 
[shard 17] token_metadata - sorted_tokens is empty in first_token_index! 
[shard 17] token_metadata - sorted_tokens is empty in first_token_index! 
[shard15] storage_proxy - Failed to apply mutation from : std::runtime_error (sorted_tokens is empty in first_token_index!) 
```

@asias 

> I think the problem is:
> 
> When a node is down, proxy_service will remove it from the nodes node when sending mutations. Once a node is restarted, there is a race between gossip starts (other nodes will mark this node up again) and the tokens are replicated to other shards.

``` 
 stoarge_service::init_server
    prepare_to_join()
          adds application state of NET_VERSION, SCHEMA and so on.
         _gossiper.start_gossiping().get();
    join_token_ring()
           _token_metadata.update_normal_tokens(tokens, get_broadcast_address());
           replicate_to_all_cores().get();
           storage_service::set_gossip_tokens() which adds the gossip application state of TOKENS and STATUS
```

asias:
When a node is restarted, there is a race between gossip starts (other
nodes will mark this node up again and send requests) and the tokens are
replicated to other shards. Here is an example:

- n1, n2
- n2 is down, n1 think n2 is down
- n2 starts again, n2 starts gossip service, n1 thinks n2 is up and sends
  reads/writes to n2, but n2 hasn't replicated the token_metadata to all
  the shards.
- n2 complains:
  token_metadata - sorted_tokens is empty in first_token_index!
  token_metadata - sorted_tokens is empty in first_token_index!
  token_metadata - sorted_tokens is empty in first_token_index!
  token_metadata - sorted_tokens is empty in first_token_index!
  token_metadata - sorted_tokens is empty in first_token_index!
  token_metadata - sorted_tokens is empty in first_token_index!
  storage_proxy - Failed to apply mutation from $ip#4: std::runtime_error
  (sorted_tokens is empty in first_token_index!)

The code path looks like below:
```
0 stoarge_service::init_server
1    prepare_to_join()
2          add gossip application state of NET_VERSION, SCHEMA and so on.
3         _gossiper.start_gossiping().get()
4    join_token_ring()
5           _token_metadata.update_normal_tokens(tokens, get_broadcast_address());
6           replicate_to_all_cores().get()
7           storage_service::set_gossip_tokens() which adds the gossip application state of TOKENS and STATUS
```
The race talked above is at line 3 and line 6.

To fix, we can replicate the token_metadata early after it is filled
with the tokens read from system table before gossip starts. So that
when other nodes think this restarting node is up, the tokens are
already replicated to all the shards.

In addition, this patch also fixes the issue that other nodes might see
a node miss the TOKENS and STATUS application state in gossip if that
node failed in the middle of a restarting process, i.e., it is killed
after line 3 and before line 7. As a result we could not replace the
node.

Tests: update_cluster_layout_tests.py
Fixes: #4709
Fixes: #4723

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -490,6 +490,14 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
         }
     }
 
+    // If this is a restarting node, we should update tokens before gossip starts
+    auto my_tokens = db::system_keyspace::get_saved_tokens().get0();
+    bool restarting_normal_node = db::system_keyspace::bootstrap_complete() && !db().local().is_replacing() && !my_tokens.empty();
+    if (restarting_normal_node) {
+        slogger.info("Restarting a node in NORMAL status");
+        _token_metadata.update_normal_tokens(my_tokens, get_broadcast_address());
+    }
+
     // have to start the gossip service before we can see any info on other nodes.  this is necessary
     // for bootstrap to get the load info it needs.
     // (we won't be part of the storage ring though until we add a counterId to our state, below.)
@@ -500,6 +508,12 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
     }).get();
     auto features = get_config_supported_features();
     _token_metadata.update_host_id(local_host_id, get_broadcast_address());
+
+    // Replicate the tokens early because once gossip runs other nodes
+    // might send reads/writes to this node. Replicate it early to make
+    // sure the tokens are valid on all the shards.
+    replicate_to_all_cores().get();
+
     auto broadcast_rpc_address = utils::fb_utilities::get_broadcast_rpc_address();
     auto& proxy = service::get_storage_proxy();
     // Ensure we know our own actual Schema UUID in preparation for updates
@@ -514,6 +528,10 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
     app_states.emplace(gms::application_state::RPC_READY, value_factory.cql_ready(false));
     app_states.emplace(gms::application_state::VIEW_BACKLOG, versioned_value(""));
     app_states.emplace(gms::application_state::SCHEMA, value_factory.schema(schema_version));
+    if (restarting_normal_node) {
+        app_states.emplace(gms::application_state::TOKENS, value_factory.tokens(my_tokens));
+        app_states.emplace(gms::application_state::STATUS, value_factory.normal(my_tokens));
+    }
     slogger.info("Starting up server gossip");
 
     _gossiper.register_(this->shared_from_this());
@@ -822,6 +840,7 @@ void storage_service::bootstrap(std::unordered_set<token> tokens) {
     } else {
         // Dont set any state for the node which is bootstrapping the existing token...
         _token_metadata.update_normal_tokens(tokens, get_broadcast_address());
+        replicate_to_all_cores().get();
         auto replace_addr = db().local().get_replace_address();
         if (replace_addr) {
             slogger.debug("Removing replaced endpoint {} from system.peers", *replace_addr);
@@ -1592,6 +1611,7 @@ future<> storage_service::init_server(int delay, bind_messaging_port do_bind) {
             auto tokens = db::system_keyspace::get_saved_tokens().get0();
             if (!tokens.empty()) {
                 _token_metadata.update_normal_tokens(tokens, get_broadcast_address());
+                replicate_to_all_cores().get();
                 // order is important here, the gossiper can fire in between adding these two states.  It's ok to send TOKENS without STATUS, but *not* vice versa.
                 _gossiper.add_local_application_state({
                     { gms::application_state::TOKENS, value_factory.tokens(tokens) },


tgrabiec:
I don't understand this explanation. What's `proxy_service`? Why a race between gossip and token replication is causing problems? Replica side is not supposed to need token metadata. Maybe the problem is that there's a race between token metadata replication and CQL server init?

asias:
> I don't understand this explanation. What's `proxy_service`? Why a race between gossip and token replication is causing problems? Replica side is not supposed to need token metadata. Maybe the problem is that there's a race between token metadata replication and CQL server init?

Sorry, I meant storage_proxy.cc. 

I will give you an example of the race:

n1, n2
n2 is down, n1 think n2 is down
n2 starts again, n2 starts gossip service, n1 thinks n2 is up and sends reads/writes to n2, but n2 hasn't replicated the token_metadata to all the shards.

n2 complains:

```
[shard 52] token_metadata - sorted_tokens is empty in first_token_index! 
[shard 45] token_metadata - sorted_tokens is empty in first_token_index! 
[shard 38] token_metadata - sorted_tokens is empty in first_token_index! 
[shard 41] token_metadata - sorted_tokens is empty in first_token_index! 
[shard 7] token_metadata - sorted_tokens is empty in first_token_index! 
[shard 7] token_metadata - sorted_tokens is empty in first_token_index! 
[shard 47] storage_proxy - Failed to apply mutation from $ip#4: std::runtime_error (sorted_tokens is empty in first_token_index!)
```

gleb-cloudius:
On Mon, Jul 15, 2019 at 07:20:43PM -0700, Asias He wrote:
> > I don't understand this explanation. What's `proxy_service`? Why a race between gossip and token replication is causing problems? Replica side is not supposed to need token metadata. Maybe the problem is that there's a race between token metadata replication and CQL server init?
> 
> Sorry, I meant storage_proxy.cc. 
> 
> I will give you an example of the race:
> 
> n1, n2
> n2 is down, n1 think n2 is down
> n2 starts again, n2 starts gossip service, n1 thinks n2 is up and sends reads/writes to n2, but n2 hasn't replicated the token_metadata to all the shards.
> 
Regular read writes do not consult token_metadata on a replica. Counters
and MV updates do though.

> n2 complains:
> 
> ```
> [shard 52] token_metadata - sorted_tokens is empty in first_token_index! 
> [shard 45] token_metadata - sorted_tokens is empty in first_token_index! 
> [shard 38] token_metadata - sorted_tokens is empty in first_token_index! 
> [shard 41] token_metadata - sorted_tokens is empty in first_token_index! 
> [shard 7] token_metadata - sorted_tokens is empty in first_token_index! 
> [shard 7] token_metadata - sorted_tokens is empty in first_token_index! 
> [shard 47] storage_proxy - Failed to apply mutation from $ip#4: std::runtime_error (sorted_tokens is empty in first_token_index!)
> ```
> 
> 
> 
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/scylladb/scylla/pull/4710#issuecomment-511637907

--
			Gleb.

asias:
I pushed a new branch. 
Changes:
- Fixes both #4723 and #4709
- Add local tokens in token_metadata early if it is a restarting of node in NORMAL status
- Advertise TOKENS and STATUS application state early if it is a restarting of node in NORMAL status

bhalevy:
@tgrabiec please re-review

tgrabiec:
The fix for #4723 looks good. 

I am not convinced that #4709 is fixed though. What prevents requests from other nodes to arrive to the node before you load and replica the tokens? You do it earlier, but the server is already up, right? The other nodes may have not notice that the node is down before they sent the requests.

asias:
> The fix for #4723 looks good.
> 
> I am not convinced that #4709 is fixed though. What prevents requests from other nodes to arrive to the node before you load and replica the tokens? You do it earlier, but the server is already up, right? The other nodes may have not noticed that the node is down before they sent the requests.

If other nodes do not notice this node is down, they can send requests at only time during this node restarts and this node might not be fully initialized. It won't be a big problem because the request will fail since the node is not fully initialized. 

This patch completely eliminates the race if other nodes have noticed the node is down which is very likely because restarts will send SHUTDOWN message. 

Even if they do not notice, this patch reduces the  big race window to a tiny window. 

gossip starts
wait for gossip to settle  (multiple seconds)
_token_metadata.update_normal_tokens(tokens, get_broadcast_address())
replicate_to_all_cores().get()

tgrabiec:
> > The fix for #4723 looks good.
> > I am not convinced that #4709 is fixed though. What prevents requests from other nodes to arrive to the node before you load and replica the tokens? You do it earlier, but the server is already up, right? The other nodes may have not noticed that the node is down before they sent the requests.
> 
> If other nodes do not notice this node is down, they can send requests at only time during this node restarts and this node might not be fully initialized. It won't be a big problem because the request will fail since the node is not fully initialized.

Is it guaranteed to fail? Or is it undefined behavior? 

Isn't the fact that the request fails the problem which this pull request is trying to fix?

> 
> This patch completely eliminates the race if other nodes have noticed the node is down which is very likely because restarts will send SHUTDOWN message.
> 
> Even if they do not notice, this patch reduces the big race window to a tiny window.
> 
> gossip starts
> wait for gossip to settle (multiple seconds)
> _token_metadata.update_normal_tokens(tokens, get_broadcast_address())
> replicate_to_all_cores().get()

Well, I think we should eliminate the race completely rather than add complexity to reduce the likelihood of failure.

asias:
> > > The fix for #4723 looks good.
> > > I am not convinced that #4709 is fixed though. What prevents requests from other nodes to arrive to the node before you load and replica the tokens? You do it earlier, but the server is already up, right? The other nodes may have not noticed that the node is down before they sent the requests.
> > 
> > 
> > If other nodes do not notice this node is down, they can send requests at only time during this node restarts and this node might not be fully initialized. It won't be a big problem because the request will fail since the node is not fully initialized.
> 
> Is it guaranteed to fail? Or is it undefined behavior?

If the request queries the token_metadata, it is guaranteed to fail because we throw if ti is not initialized. 

> 
> Isn't the fact that the request fails the problem which this pull request is trying to fix?

It tries to fix the restart node case. 

> 
> > This patch completely eliminates the race if other nodes have noticed the node is down which is very likely because restarts will send SHUTDOWN message.
> > Even if they do not notice, this patch reduces the big race window to a tiny window.
> > gossip starts
> > wait for gossip to settle (multiple seconds)
> > _token_metadata.update_normal_tokens(tokens, get_broadcast_address())
> > replicate_to_all_cores().get()
> 
> Well, I think we should eliminate the race completely rather than add complexity to reduce the likelihood of failure.

This patch tries to fix the restart node case and it fixes it 100%. 

For the case a node does not notice the node is down, it also helped but not 100%. We can add a new issue to track.

tgrabiec:
On Fri, Jul 26, 2019 at 6:06 PM Asias He <notifications@github.com> wrote:

> The fix for #4723 <https://github.com/scylladb/scylla/issues/4723> looks
> good.
> I am not convinced that #4709
> <https://github.com/scylladb/scylla/issues/4709> is fixed though. What
> prevents requests from other nodes to arrive to the node before you load
> and replica the tokens? You do it earlier, but the server is already up,
> right? The other nodes may have not noticed that the node is down before
> they sent the requests.
>
> If other nodes do not notice this node is down, they can send requests at
> only time during this node restarts and this node might not be fully
> initialized. It won't be a big problem because the request will fail since
> the node is not fully initialized.
>
> Is it guaranteed to fail? Or is it undefined behavior?
>
> If the request queries the token_metadata, it is guaranteed to fail
> because we throw if ti is not initialized.
>
Yes. But it may rely on something else which may result in more subtle
errors. In general, the system may not behave correctly if you allow
requests to act on an uninitialized system. The proper fix is to not let
requests in when the system is in such a state.

Initializing things sooner, as this PR does, may reduce the likelihood of
error but doesn't eliminate it. We should solve the problem completely now,
when we identified it, and not cover up some cases and wait until a
disaster happens.

asias:
On Mon, Jul 29, 2019, 18:36 Tomasz Grabiec <notifications@github.com> wrote:

> On Fri, Jul 26, 2019 at 6:06 PM Asias He <notifications@github.com> wrote:
>
> > The fix for #4723 <https://github.com/scylladb/scylla/issues/4723> looks
> > good.
> > I am not convinced that #4709
> > <https://github.com/scylladb/scylla/issues/4709> is fixed though. What
> > prevents requests from other nodes to arrive to the node before you load
> > and replica the tokens? You do it earlier, but the server is already up,
> > right? The other nodes may have not noticed that the node is down before
> > they sent the requests.
> >
> > If other nodes do not notice this node is down, they can send requests at
> > only time during this node restarts and this node might not be fully
> > initialized. It won't be a big problem because the request will fail
> since
> > the node is not fully initialized.
> >
> > Is it guaranteed to fail? Or is it undefined behavior?
> >
> > If the request queries the token_metadata, it is guaranteed to fail
> > because we throw if ti is not initialized.
> >
> Yes. But it may rely on something else which may result in more subtle
> errors. In general, the system may not behave correctly if you allow
> requests to act on an uninitialized system. The proper fix is to not let
> requests in when the system is in such a state.
>
> Initializing things sooner, as this PR does, may reduce the likelihood of
> error but doesn't eliminate it. We should solve the problem completely now,
> when we identified it, and not cover up some cases and wait until a
> disaster happens.
>

As I mentioned, we should create a new issue to track proxy service startup
and early requests race in general. This patch fixes the original issue
already.

asias:
I created https://github.com/scylladb/scylla/issues/4771 to track storage proxy initialization issue.

asias:
@tgrabiec @slivne ping

Issue #4695, opened 2019-07-11
mutation_reader_test fails some of the time


penberg:
I see `mutation_reader_test` fail 2 times out of 100 runs as follows:

```
The following test(s) have failed:
  build/release/tests/mutation_reader_test -- -c3 -m2G --overprovisioned --unsafe-bypass-fsync 1 --blocked-reactor-notify-ms 2000000 --collectd 0
  build/release/tests/mutation_reader_test -- -c3 -m2G --overprovisioned --unsafe-bypass-fsync 1 --blocked-reactor-notify-ms 2000000 --collectd 0
Summary: 2 of the total 10500 tests failed
```

```
random_mutation_generator seed: 489635627
*** 1 failure is detected in the test module "tests/mutation_reader_test.cc"
```
```
random_mutation_generator seed: 4251321561
*** 1 failure is detected in the test module "tests/mutation_reader_test.cc"
```

I manage to reproduce _something_ if I use the `--random-seed` option:

```
./tools/toolchain/dbuild build/release/tests/mutation_reader_test --random-seed=489635627 -- -c3 -m2G --overprovisioned --unsafe-bypass-fsync 1 --blocked-reactor-notify-ms 2000000 --collectd 0
```

The failure seems to be different across runs. For example, I saw this:

```
unknown location(0): fatal error: in "reader_selector_gap_between_readers_test": memory access violation at address: 0x7ffec2d9d1c0: invalid permissions
```

and this:

```
unknown location(0): fatal error: in "test_combined_reader_slicing_with_overlapping_range_tombstones": signal: SIGSEGV, si_code: 0 (memory access violation at address: 0x00000080)
```

You can use the `--repeat N` command line option that was added to `test.py` to find more corner cases that fail:

```
./tools/toolchain/dbuild ./test.py --repeat 100 --name mutation_reader_test
```

bhalevy:
Seen in [next/985](https://jenkins.scylladb.com/view/master/job/scylla-master/job/next/985/consoleText) scylla version 67c624d96752482f6b63e451fdf3533d757b2e2b:
```
[2019-07-12T20:00:57.191Z] Output of the failed tests:
[...]

PR #4715, opened 2019-07-15
tests/mutation_reader_test stabilize test_multishard_combining_reader_non_strictly_monotonic_positions

denesb:
Currently the
`test_multishard_combining_reader_non_strictly_monotonic_positions` is
flaky. The test is somewhat unconventional, in that it doesn't use the
same instance of data as the input to the test and as it's expected
output, instead it invokes the method which generates this data
(`make_fragments_with_non_monotonic_positions()`) twice, first to
generate the input, and a secondly to generate the expected output. This
means that the test is prone to any deviation in the data generated by
said method. One such deviation, discovered recently, is that the method
doesn't explicitly specify the deletion time of the generated range
tombstones. This results in this deletion time sometimes differing
between the test input and the expected output. Solve by explicitly
passing the same deletion time to all created range tombstones.

Refs: #4695

### DIFF
--- a/tests/mutation_reader_test.cc
+++ b/tests/mutation_reader_test.cc
@@ -2035,11 +2035,12 @@ std::deque<mutation_fragment> make_fragments_with_non_monotonic_positions(simple
     std::deque<mutation_fragment> fragments;
 
     fragments.emplace_back(partition_start{std::move(pkey), {}});
+    const auto tombstone_deletion_time = gc_clock::now();
 
     int i = 0;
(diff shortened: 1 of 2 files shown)


bhalevy:
for the record, the test keeps failing in the same way even with this fix.
For example [next/1006](https://jenkins.scylladb.com/view/master/job/scylla-master/job/next/1006/console) , [next/1009](https://jenkins.scylladb.com/view/master/job/scylla-master/job/next/1009/console)

denesb:
Yes, actually the fix is wrong, will send a follow-up fix soon.

PR #4720, opened 2019-07-16
Fix storage service for tests

denesb:
Fix another source of flakyness in `mutation_reader_test`. This one is caused by `storage_service_for_tests` lacking a `config::broadcast_to_all_shards()` call, triggering an invalid memory access (or SEGFAULT) when run on more than one shards.

Refs: #4695


bhalevy:
LGTM

avikivity:
Queued

Issue #4708, opened 2019-07-15
Repair of a single shard range opens rpc connections for streaming on all shards


slivne:
Scylla Version: 3.0.X

I ran nodetool repair from the command line using the ranges from the trace

nodetool repair -st 8811292670950375424 -et 8812418570857218048 keyspace1 

```
Jul 14 22:30:35 ip-10-0-0-95.ec2.internal scylla[7188]:  [shard 0] repair - starting user-requested repair for keyspace keyspace1, repair id 40428, options {{ trace -> false}, { ranges -> 8811292670950375424:881
Jul 14 22:30:35 ip-10-0-0-95.ec2.internal scylla[7188]:  [shard 0] repair - Repair 1 out of 1 ranges, id=40428, shard=0, keyspace=keyspace1, table={standard1}, range=(8811292670950375424, 8812418570857218048]
Jul 14 22:30:35 ip-10-0-0-95.ec2.internal scylla[7188]:  [shard 0] repair - repair 40428 on shard 0 completed successfully
Jul 14 22:30:35 ip-10-0-0-95.ec2.internal scylla[7188]:  [shard 2] repair - Repair 1 out of 1 ranges, id=40428, shard=2, keyspace=keyspace1, table={standard1}, range=(8811292670950375424, 8812418570857218048]
Jul 14 22:30:35 ip-10-0-0-95.ec2.internal scylla[7188]:  [shard 3] repair - Repair 1 out of 1 ranges, id=40428, shard=3, keyspace=keyspace1, table={standard1}, range=(8811292670950375424, 8812418570857218048]
Jul 14 22:30:35 ip-10-0-0-95.ec2.internal scylla[7188]:  [shard 1] repair - Repair 1 out of 1 ranges, id=40428, shard=1, keyspace=keyspace1, table={standard1}, range=(8811292670950375424, 8812418570857218048]
Jul 14 22:30:35 ip-10-0-0-95.ec2.internal scylla[7188]:  [shard 1] repair - repair 40428 on shard 1 completed successfully
Jul 14 22:30:35 ip-10-0-0-95.ec2.internal scylla[7188]:  [shard 3] repair - repair 40428 on shard 3 completed successfully
Jul 14 22:30:35 ip-10-0-0-95.ec2.internal scylla[7188]:  [shard 2] repair - Start streaming for repair id=40428, shard=2, index=0, ranges_in=1, ranges_out=2
[...]

PR #4721, opened 2019-07-16
streaming: Do not open rpc stream connection if ranges are not relevant to a shard

asias:
Given a list of ranges to stream, stream_transfer_task will create an
reader with the ranges and create a rpc stream connection on all the shards.

When user provides ranges to repair with -st -et options, e.g.,
using scylla-manger, such ranges can belong to only one shard, repair
will pass such ranges to streaming.

As a result, only one shard will have data to send while the rpc stream
connections are created on all the shards, which can cause the kernel
run out of ports in some systems.

To mitigate the problem, do not open the connection if the ranges do not
belong to the shard at all.

Refs: #4708

### DIFF
--- a/streaming/stream_transfer_task.cc
+++ b/streaming/stream_transfer_task.cc
@@ -105,6 +105,21 @@ struct send_info {
         , prs(to_partition_ranges(ranges))
         , reader(cf.make_streaming_reader(cf.schema(), prs)) {
     }
+    future<bool> has_relevant_range_on_this_shard() {
+        return do_with(false, [this] (bool& found_relevant_range) {
+            return do_for_each(ranges, [this, &found_relevant_range] (dht::token_range range) {
+                if (!found_relevant_range) {
+                    auto sharder = dht::selective_token_range_sharder(range, engine().cpu_id());
+                    auto range_shard = sharder.next();
+                    if (range_shard) {
+                        found_relevant_range = true;
+                    }
+                }
+            }).then([&found_relevant_range] {
+                return found_relevant_range;
+            });
+        });
+    }
     future<size_t> estimate_partitions() {
         return do_with(cf.get_sstables(), size_t(0), [this] (auto& sstables, size_t& partition_count) {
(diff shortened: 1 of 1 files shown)


asias:
Changes in:
- Use do_with for found_relevant_range
- Drop the repeat because we only need to call it once
- Drop the usage with do_with for sharder

slivne:
@gleb-cloudius please review we need this for customers

streaming/stream_transfer_task.cc
```diff
@@ -105,6 +105,23 @@ struct send_info {
         , prs(to_partition_ranges(ranges))
         , reader(cf.make_streaming_reader(cf.schema(), prs)) {
     }
+    future<bool> has_relevant_range_on_this_shard() {
+        return do_with(false, [this] (bool& found_relevant_range) {
+            return do_for_each(ranges, [this, &found_relevant_range] (dht::token_range range) {
+                if (found_relevant_range) {
```
gleb-cloudius: Nitpick: you can put the logic below in if (!found_relevant_range) and drop both returns. Two lines less.
Although it is a pity we need to iterate over all ranges even if a first one belongs to the shard especially as you claim
there can be a lot of them.

gleb-cloudius:
Looks fine (only a nit). I still think find_if is enough here though.

asias:
Updated:
- Put logic under !found_relevant_range branch
- Drop two return make_ready_future<>()

streaming/stream_transfer_task.cc
asias: If we add "break" support to do_for_each, we can avoid the further loops.

Since this code is on a slow path, I do not worry about the extra loop a lot.

gleb-cloudius: We can add do_for_each version that takes a function that returns  future<stop_iteration>

slivne:
@avikivity ping we have customers waiting for this

avikivity:
Queued as 64a4c0ede2

Issue #4726, opened 2019-07-16
Implement "sstable description" api command


elcallio:
In general, and for certain enterprise features in particular, it would be useful to have a "describe sstables" command that list all/some sstables properties and attributes, such as size, generation, timestamp, version and compression and other "extended" attributes that affect its use and behaviour.

slivne:
@elcallio whats missing to close this ?

elcallio:
It should be backported everywhere now, so nothing.

PR #4727, opened 2019-07-16
Implement `sstable_info` API command (info on sstables)

elcallio:
Refs #4726

Implement the api portion of a "describe sstables" command. 

Adds rest types for collecting both fixed and dynamic attributes, some grouped. Allows extensions to add attributes as well. (Hint hint)

v2:
* Use (not really legal) nested "type" in json
* Rename "table" param to "cf" for consistency
* Some comments on data sizes
* Stream result to avoid huge string allocations on final json

### DIFF
--- a/api/api-doc/storage_service.json
+++ b/api/api-doc/storage_service.json
@@ -2164,7 +2164,42 @@
                ]
             }
          ]
-      }
+      },
+      {
+         "path":"/storage_service/sstable_info",
+         "operations":[
+            {
+               "method":"GET",
+               "summary":"SSTable information",
+               "type":"array",
+               "items":{
+                  "type":"table_sstables"
+               },
+               "nickname":"sstable_info",
+               "produces":[
+                  "application/json"
+               ],
+               "parameters":[
+                  {
+                     "name":"keyspace",
+                     "description":"The keyspace",
+                     "required":false,
+                     "allowMultiple":false,
+                     "type":"string",
+                     "paramType":"query"
+                  },
+                  {
+                     "name":"cf",
+                     "description":"column family name",
+                     "required":false,
+                     "allowMultiple":false,
+                     "type":"string",
+                     "paramType":"query"
+                  }
+               ]
+            }
+         ]
+      }      
    ],
    "models":{
       "mapper":{
@@ -2324,6 +2359,92 @@
                "description":"The endpoint details"
             }
          }
+      },
+      "named_maps":{
+        "id":"named_maps",
+        "properties":{
+            "group":{
+                "type":"string"
+            },
+            "attributes":{
+                "type":"array",
+                "items":{
+                    "type":"mapper"
+                }
+            }
+        }
+      },
+      "sstable":{
+        "id":"sstable",
+        "properties":{
+            "size":{
+               "type":"long",
+               "description":"Total size in bytes of sstable"
+            },
+            "data_size":{
+                "type":"long",
+                "description":"The size in bytes on disk of data"
+            },
+            "index_size":{
+               "type":"long",
+               "description":"The size in bytes on disk of index"
+            },
+            "filter_size":{
+               "type":"long",
+               "description":"The size in bytes on disk of filter"
+            },
+            "timestamp":{
+                "type":"datetime",
+                "description":"File creation time"
+            },
+            "generation":{
+                "type":"long",
+                "description":"SSTable generation"
+            },
+            "level":{
+               "type":"long",
+               "description":"SSTable level"
+            },
+            "version":{
+               "type":"string",
+               "enum":[
+                  "ka", "la", "mc"
+               ],
+               "description":"SSTable version"
+            },
+            "properties":{
+                "type":"array",
+                "description":"SSTable attributes",
+                "items":{
+                    "type":"mapper"
+                }
+            },
+            "extended_properties":{
+                "type":"array",
+                "description":"SSTable extended attributes",
+                "items":{
+                    "type":"named_maps"
+                }
+            }
+        }
+      },
+      "table_sstables":{
+        "id":"table_sstables",
+        "description":"Per-table SSTable info and attributes",
+        "properties":{
+            "keyspace":{
+                "type":"string"
+            },
+            "table":{
+                "type":"string"
+            },
+            "sstables":{
+                "type":"array",
+                "items":{
+                    "$ref":"sstable"
+                }
+            }
+        }
       }
    }
 }
--- a/api/storage_service.cc
+++ b/api/storage_service.cc
@@ -22,13 +22,15 @@
 #include "storage_service.hh"
 #include "api/api-doc/storage_service.json.hh"
 #include "db/config.hh"
+#include <optional>
+#include <time.h>
 #include <boost/range/adaptor/map.hpp>
 #include <boost/range/adaptor/filtered.hpp>
-#include <service/storage_service.hh>
-#include <db/commitlog/commitlog.hh>
-#include <gms/gossiper.hh>
-#include <db/system_keyspace.hh>
-#include <seastar/http/exception.hh>
+#include "service/storage_service.hh"
+#include "db/commitlog/commitlog.hh"
+#include "gms/gossiper.hh"
+#include "db/system_keyspace.hh"
+#include "seastar/http/exception.hh"
 #include "repair/repair.hh"
 #include "locator/snitch_base.hh"
 #include "column_family.hh"
@@ -37,6 +39,7 @@
 #include "sstables/compaction_manager.hh"
 #include "sstables/sstables.hh"
 #include "database.hh"
+#include "db/extensions.hh"
 
 sstables::sstable::version_types get_highest_supported_format();
 
@@ -51,7 +54,6 @@ static sstring validate_keyspace(http_context& ctx, const parameters& param) {
     }
     throw bad_param_exception("Keyspace " + param["keyspace"] + " Does not exist");
 }
-
 
 static std::vector<ss::token_range> describe_ring(const sstring& keyspace) {
     std::vector<ss::token_range> res;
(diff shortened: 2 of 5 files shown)


sstables/sstables.hh
```diff
@@ -948,6 +949,14 @@ class file_io_extension {
 public:
     virtual ~file_io_extension() {}
     virtual future<file> wrap_file(sstable&, component_type, file, open_flags flags) = 0;
+    // optionally populate a map of attributes for a given sstable,
+    // suitable for "describe".
+    // This would preferably be interesting info on what/why the extension did
+    // to this table.
+    using attr_value_type = std::variant<sstring, std::map<sstring, sstring>>;
+    using attr_value_map = std::map<sstring, attr_value_type>;
+    virtual void get_attributes(const sstable&, attr_value_map& atts) const
```
denesb: Why not return the attribute map?

sstables/compress.hh
```diff
@@ -367,6 +367,10 @@ public:
     friend class sstable;
 };
 
+// for API query only. Free function just to distinguish it from an accessor in compression
+// accepts nullptr as arg. 
+compressor_ptr get_sstable_compressor(const compression&);
```
denesb: How does it accept a nullptr when the arg is a reference?

api/storage_service.cc
```diff
@@ -900,6 +902,129 @@ void set_storage_service(http_context& ctx, routes& r) {
...
+                                    e.value = cp->name();
+                                    nm.attributes.push(std::move(e));
+                                }
+                                info.extended_properties.push(std::move(nm));
+                            }
+
+                            sstables::file_io_extension::attr_value_map map;
+
+                            for (auto* ep : ext.sstable_file_io_extensions()) {
+                                ep->get_attributes(*sstable, map);
```
denesb: I see now why (the map is not returned). Well there is `std::map::merge()` now but it will be slightly less efficient than this (keys will be looked up twice). IMO the code would be more natural, but no strong opinion.

api/storage_service.cc
```diff
@@ -900,6 +902,129 @@ void set_storage_service(http_context& ctx, routes& r) {
...
+                                        }
+                                        info.extended_properties.push(std::move(nm));
+                                    }
+                                    void operator()(const sstring& value) const {
+                                        ss::map_entry e;
+                                        e.key = key;
+                                        e.value = value;
+                                        info.properties.push(std::move(e));                                        
+                                    }
+                                } v{p.first, info};
```
denesb: Nice visitor trick.

sstables/sstables.hh
elcallio: Because I want to allow an extension to _add_ either single attributes, or _modify_ them. Obviously, merging of maps can be done on the higher level, but here I opted to add-to-or-replace-in-set functionality direct.

sstables/compress.hh
elcallio: Sorry, the comment is stale. First iteration was thusly, but I modified it. Will reword the comment.

elcallio: Removed.

api/storage_service.cc
elcallio: I sort of agree, but it would also (for better or worse) not allow for example overriding an existing attribute (since merge does not). Not sure if there is any market for this particular feature really, but I sort of wanted it just in case. Of course it also means extensions have potentially to much control, which is a problem in the other direction.

elcallio: Actually, I'm gonna go with your objection. I've changed the API. It is cleaner. New version pushed.

sstables/sstables.hh
elcallio: Api changed.

api/storage_service.cc
elcallio: We should add "make_override" to some util header...

avikivity:
Please use standard English in subject lines, using pull requests is not an invitation for uninfomative descriptions.

elcallio:
Subject modified.

api/api-doc/storage_service.json
```diff
@@ -2164,7 +2164,42 @@
...
+                  {
+                     "name":"keyspace",
+                     "description":"The keyspace",
+                     "required":false,
+                     "allowMultiple":false,
+                     "type":"string",
+                     "paramType":"query"
+                  },
+                  {
+                     "name":"table",
```
amnonh: while it makes sense using table (as we are following Cassandra's naming) everything else is CF and even your description says Column Family.

I would go with using `cf` as the name, but even with `table` the description should match

api/api-doc/storage_service.json
```diff
@@ -2324,6 +2359,103 @@
                "description":"The endpoint details"
             }
          }
+      },
+      "map_entry":{
```
amnonh: we already have a model call `mapper` that has the exact same key/value attributes

api/api-doc/storage_service.json
```diff
@@ -2324,6 +2359,103 @@
...
+      "named_maps":{
+        "id":"named_maps",
+        "properties":{
+            "group":{
+                "type":"string"
+            },
+            "attributes":{
+                "type":"array",
+                "items":{
+                    "$ref":"map_entry"
```
amnonh: if you use type instead of $ref you'll not need to change the json2code

api/api-doc/storage_service.json
```diff
@@ -2324,6 +2359,103 @@
...
+                }
+            }
+        }
+      },
+      "sstable":{
+        "id":"sstable",
+        "properties":{
+            "size":{
+               "type":"long",
+               "description":"Total size in bytes on"
```
amnonh: didn't understand the description

api/api-doc/storage_service.json
```diff
@@ -2324,6 +2359,103 @@
...
+               "enum":[
+                  "ka", "la", "mc"
+               ],
+               "description":"SSTable version"
+            },
+            "properties":{
+                "type":"array",
+                "description":"SSTable attributes",
+                "items":{
+                    "$ref":"map_entry"
```
amnonh: again about the ref and object

api/storage_service.cc
```diff
@@ -900,6 +902,129 @@ void set_storage_service(http_context& ctx, routes& r) {
             return make_ready_future<json::json_return_type>(map_to_key_value(std::move(status), res));
         });
     });
+
+    ss::sstable_info.set(r, [&ctx] (std::unique_ptr<request> req) {
+        auto ks = api::req_param<sstring>(*req, "keyspace", {}).value;
+        auto cf = api::req_param<sstring>(*req, "table", {}).value;
+
+        using table_sstables_list = std::vector<ss::table_sstables>;
```
amnonh: how big can this vector get?
There are two potential concerns with the API, stalls and large allocations.
I'm not sure it is an issue, but if the vector is big it can cause large allocation and if the structures are big, streaming it to JSON can cause a large allocation.
If it's not an issue we can ignore it

api/storage_service.cc
```diff
@@ -900,6 +902,129 @@ void set_storage_service(http_context& ctx, routes& r) {
...
+
+        return do_with(table_sstables_list{}, [ks, cf, &ctx](table_sstables_list& dst) {
+            return service::get_local_storage_service().db().map_reduce([&dst](table_sstables_list&& res) {
+                for (auto&& t : res) {
```
amnonh: The same issue from before, it's an iteration without preemption if it's not too much work, we can keep it as is, if it can potentially take a few ms we should make it safer to prevent stalls.

api/api-doc/storage_service.json
elcallio: Honestly, I would rather clean up the description. We should really migrate to "table" (and potentially allow using either name for the parameter in all places).

api/api-doc/storage_service.json
elcallio: Missed that.

api/api-doc/storage_service.json
elcallio: Missed that.

api/api-doc/storage_service.json
elcallio: Must have slipped with clipboard or something. Should read "of sstable".

api/api-doc/storage_service.json
elcallio: Same answer as above.

api/storage_service.cc
elcallio: It can absolutely get big if you ask for all sstables in the system. But "big" here is probably the order of 100/1000 entries, or whatever number of sstables in flight a node might have.

elcallio: Using a list will help with large alloc, but json serialization is another matter. Should that not be streaming?

api/storage_service.cc
elcallio: What do you mean "safer"? Explicit yields? The map-reduce op should be yielding in itself really...

api/storage_service.cc
amnonh: It all depends on the size, using stream (see compaction history for example) makes the code a bit more complex but is safer.
If it can grow big, you can use: `stream_range_as_array` (you can see examples in the api/storage_servie.cc file)

api/api-doc/storage_service.json
amnonh: are you sure it's not legal in 1.2?
anyhow, we do use it in other places, so you can use it as well.

api/storage_service.cc
amnonh: right, but the for loop will not, so that's the question if the for loop can take anything more than a 1ms we should probably do something else

avikivity:
@elcallio were the changes requested by @amnonh done? what were the changes?

elcallio:
Yes, sorry, maybe should have copied the "v2" list from patch comment. 
v2:
* Use (not really legal) nested "type" in json
* Rename "table" param to "cf" for consistency
* Some comments on data sizes
* Stream result to avoid huge string allocations on final json

api/storage_service.cc
elcallio: I think it should be fairly safe. It will be a lot of string copying, but nothing really heavy. Plus, it is not a command you should run willy-nilly polling the server. If we find a case where it causes a stall, I can add yielding behaviour then. Hopefully it will be when we have coroutines. :-)

api/storage_service.cc
elcallio: Right. I've modified the final transform to use the streaming version.

api/api-doc/storage_service.json
elcallio: Ok, why not. Still think it is an easy fix to the generator code, since it already handles $ref for actual code gen (only dependency checking broke). But I've changed it to "type".

api/api-doc/storage_service.json
elcallio: fixed

api/api-doc/storage_service.json
elcallio: Changed to cf

api/storage_service.cc
```diff
@@ -900,6 +902,133 @@ void set_storage_service(http_context& ctx, routes& r) {
...
+
+                                std::visit(v, p.second);
+                            }
+
+                            tst.sstables.push(std::move(info));
+                        }
+                        res.emplace_back(std::move(tst));
+                    }
+                }
+                std::sort(res.begin(), res.end(), [](const ss::table_sstables& t1, const ss::table_sstables& t2) {
```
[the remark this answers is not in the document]
elcallio: It is needed in the "reduce" part, to make resolving duplicates faster.

api/storage_service.cc
```diff
@@ -900,6 +902,133 @@ void set_storage_service(http_context& ctx, routes& r) {
...
+                for (auto& t : db.get_column_families() | boost::adaptors::map_values) {
+                    auto& schema = t->schema();
+                    if ((ks.empty() || ks == schema->ks_name()) && (cf.empty() || cf == schema->cf_name())) {
+                        // at most Nsstables long
+                        ss::table_sstables tst;
+                        tst.keyspace = schema->ks_name();
+                        tst.table = schema->cf_name();
+
+                        for (auto sstable : *t->get_sstables_including_compacted_undeleted()) {
+                            auto ts = db_clock::to_time_t(sstable->data_file_write_time());
```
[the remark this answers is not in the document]
elcallio: Indeed. I am hoping not, but I could be wrong.

sstables/sstables.hh
```diff
@@ -948,6 +949,15 @@ class file_io_extension {
...
+    // optionally return a map of attributes for a given sstable,
+    // suitable for "describe".
+    // This would preferably be interesting info on what/why the extension did
+    // to this table.
+    using attr_value_type = std::variant<sstring, std::map<sstring, sstring>>;
+    using attr_value_map = std::map<sstring, attr_value_type>;
+    virtual attr_value_map get_attributes(const sstable&) const {
+        return {};
+    }
 };
```
avikivity: I think it's nicer to return C++ structs here and not unstructured data, then do the conversion at the API level.

Probably not worth the effort.

elcallio: How do you create a struct to describe a property map that is not, you know, a ... map...?
The point is to return arbitrary key-values (with value optionally being another key-value set).

avikivity: They're not arbitrary.

```c++
struct sstable_attributes {
    bool is_compressed;
    sstring creator_name;
};
```

it's kind of pointless in this case because the only consumer will mangle them to a map, but that's how I would have done it, I like delaying the destruction of structure as far as I can.

elcallio: Those attributes are not handled by this. This is only for extensions to provide info not known to the collecting API call. 
All "known" quantities are handled structured. 
This is basically for E@R and other extensions to be able to provide a set of properties.

avikivity: they could provide proper methods too (but exposing complex attributes is hard)

avikivity:
Queued as 0d0ee20f76

Issue #4733, opened 2019-07-17
multishard_combining_reader may be missing fragments if recreated right after prefix range tombstone


tgrabiec:
When `shard_reader::remote_reader` stops, it saves the key of the last emitted fragment. Then if the reader needs to be recreated, it adjusts the slice so that it will emit elements right after that fragment. However, it works with `clustering_key_prefixes`. If the last fragment was a range tombstone starting at a prefix, then recreation will create a slice which skips the whole prefix, but there may be elements after that tombstone falling into the prefix which were not emitted yet.

Reproducer:
```diff
diff --git a/mutation_reader.cc b/mutation_reader.cc
index 1d5e3cb..a78a879 100644
--- a/mutation_reader.cc
+++ b/mutation_reader.cc
@@ -1109,9 +1109,9 @@ flat_mutation_reader shard_reader::remote_reader::resume_or_create_reader() {
         _reader_created = true;
         return reader;
     }
-    if (auto reader_opt = _lifecycle_policy.try_resume(std::move(_irh))) {
-        return std::move(*reader_opt);
-    }
+    //if (auto reader_opt = _lifecycle_policy.try_resume(std::move(_irh))) {
+    //    return std::move(*reader_opt);
+    //}
     return recreate_reader();
 }
 
@@ -1203,7 +1214,7 @@ future<shard_reader::fill_buffer_result> shard_reader::remote_reader::fill_buffe
         if (pending_next_partition) {
             reader.next_partition();
         }
-
+        reader.set_max_buffer_size(1);
         return do_fill_buffer(reader, timeout).then([this, &reader, &buffer, timeout] {
             buffer = reader.detach_buffer();
             // When the reader is recreated (after having been evicted) we
diff --git a/tests/mutation_source_test.cc b/tests/mutation_source_test.cc
index 58a0802..dd5516f 100644
--- a/tests/mutation_source_test.cc
+++ b/tests/mutation_source_test.cc
@@ -1505,6 +1538,14 @@ static mutation_sets generate_mutation_sets() {
             result.unequal.emplace_back(mutations{m1, m2});
             m2.partition().apply_row_tombstone(*s1, key, tomb);
             result.equal.emplace_back(mutations{m1, m2});
+
+            // Add a row which falls under the tombstone prefix.
+            auto ck2 = clustering_key_prefix::from_deeply_exploded(*s1, {data_value(bytes("ck2_0")), data_value(bytes("ck1_1")), });
+            auto ts = new_timestamp();
[...]

PR #4740, opened 2019-07-18
Multishard combining reader more robust reader recreation

denesb:
Make the reader recreation logic more robust, by moving away from deciding which fragments have to be dropped based on a bunch of special cases, instead replacing this with a general logic which just drops all already seen fragments (based on their position).
Special handling is added for the case when the last position is a range tombstone with a non full prefix starting position.
Reproducer unit tests are added for both cases.

Refs: #4695
Fixes: #4733

### DIFF
--- a/mutation_reader.cc
+++ b/mutation_reader.cc
@@ -1096,24 +1096,22 @@ flat_mutation_reader shard_reader::remote_reader::recreate_reader() {
     if (_last_pkey) {
         bool partition_range_is_inclusive = true;
 
-        {
-            switch (_next_position_in_partition.region()) {
-            case partition_region::partition_start:
-                partition_range_is_inclusive = false;
-                break;
-            case partition_region::static_row:
-                _drop_partition_start = true;
-                break;
-            case partition_region::clustered:
-                _drop_partition_start = true;
-                _drop_static_row = true;
-                adjust_partition_slice();
-                slice = &*_slice_override;
-                break;
-            case partition_region::partition_end:
-                partition_range_is_inclusive = false;
-                break;
-            }
+        switch (_next_position_in_partition.region()) {
+        case partition_region::partition_start:
+            partition_range_is_inclusive = false;
+            break;
+        case partition_region::static_row:
+            _drop_partition_start = true;
+            break;
+        case partition_region::clustered:
+            _drop_partition_start = true;
+            _drop_static_row = true;
+            adjust_partition_slice();
+            slice = &*_slice_override;
+            break;
+        case partition_region::partition_end:
+            partition_range_is_inclusive = false;
+            break;
         }
 
         // The original range contained a single partition and we've read it
--- a/mutation_reader.cc
+++ b/mutation_reader.cc
@@ -924,9 +924,10 @@ class shard_reader : public enable_lw_shared_from_this<shard_reader>, public fla
         bool _reader_created = false;
         bool _drop_partition_start = false;
         bool _drop_static_row = false;
+        position_in_partition::tri_compare _tri_cmp;
 
         std::optional<dht::decorated_key> _last_pkey;
-        std::optional<position_in_partition> _last_position_in_partition;
+        position_in_partition _next_position_in_partition = position_in_partition::for_partition_start();
         // These are used when the reader has to be recreated (after having been
         // evicted while paused) and the range and/or slice it is recreated with
         // differs from the original ones.
@@ -934,14 +935,13 @@ class shard_reader : public enable_lw_shared_from_this<shard_reader>, public fla
         std::optional<query::partition_slice> _slice_override;
 
     private:
-        void update_last_position(const circular_buffer<mutation_fragment>& buffer);
+        void update_next_position(flat_mutation_reader& reader, circular_buffer<mutation_fragment>& buffer);
         void adjust_partition_slice();
         flat_mutation_reader recreate_reader();
         flat_mutation_reader resume_or_create_reader();
         bool should_drop_fragment(const mutation_fragment& mf);
         future<> do_fill_buffer(flat_mutation_reader& reader, db::timeout_clock::time_point timeout);
-        future<> ensure_buffer_contains_all_fragments_for_last_pos(flat_mutation_reader& reader, circular_buffer<mutation_fragment>& buffer,
-                db::timeout_clock::time_point timeout);
+        future<> fill_buffer(flat_mutation_reader& reader, circular_buffer<mutation_fragment>& buffer, db::timeout_clock::time_point timeout);
 
     public:
         remote_reader(
@@ -1039,7 +1039,7 @@ void shard_reader::stop() noexcept {
     }).finally([zis = shared_from_this()] {}));
 }
 
-void shard_reader::remote_reader::update_last_position(const circular_buffer<mutation_fragment>& buffer) {
+void shard_reader::remote_reader::update_next_position(flat_mutation_reader& reader, circular_buffer<mutation_fragment>& buffer) {
     if (buffer.empty()) {
         return;
     }
@@ -1050,7 +1050,31 @@ void shard_reader::remote_reader::update_last_position(const circular_buffer<mut
         _last_pkey = pk_it->as_partition_start().key();
     }
 
-    _last_position_in_partition.emplace(buffer.back().position());
+    const auto last_pos = buffer.back().position();
+    switch (last_pos.region()) {
+        case partition_region::partition_start:
+            _next_position_in_partition = position_in_partition::for_static_row();
+            break;
+        case partition_region::static_row:
+            _next_position_in_partition = position_in_partition::before_all_clustered_rows();
+            break;
+        case partition_region::clustered:
+            if (reader.is_buffer_empty()) {
+                _next_position_in_partition = position_in_partition::after_key(last_pos);
+            } else {
+               const auto& next_frag = reader.peek_buffer();
+               if (next_frag.is_end_of_partition()) {
+                   buffer.emplace_back(reader.pop_mutation_fragment());
+                   _next_position_in_partition = position_in_partition::for_partition_start();
+               } else {
+                   _next_position_in_partition = position_in_partition(next_frag.position());
+               }
+            }
+            break;
+        case partition_region::partition_end:
+           _next_position_in_partition = position_in_partition::for_partition_start();
+           break;
+    }
 }
 
 void shard_reader::remote_reader::adjust_partition_slice() {
@@ -1058,9 +1082,8 @@ void shard_reader::remote_reader::adjust_partition_slice() {
         _slice_override = _ps;
     }
 
-    auto& last_ckey = _last_position_in_partition->key();
     auto ranges = _slice_override->default_row_ranges();
-    query::trim_clustering_row_ranges_to(*_schema, ranges, last_ckey);
+    query::trim_clustering_row_ranges_to(*_schema, ranges, _next_position_in_partition);
 
     _slice_override->clear_ranges();
     _slice_override->set_range(*_schema, _last_pkey->key(), std::move(ranges));
@@ -1073,14 +1096,13 @@ flat_mutation_reader shard_reader::remote_reader::recreate_reader() {
     if (_last_pkey) {
         bool partition_range_is_inclusive = true;
 
-        if (_last_position_in_partition) {
-            switch (_last_position_in_partition->region()) {
+        {
+            switch (_next_position_in_partition.region()) {
             case partition_region::partition_start:
-                _drop_partition_start = true;
+                partition_range_is_inclusive = false;
                 break;
             case partition_region::static_row:
                 _drop_partition_start = true;
-                _drop_static_row = true;
                 break;
             case partition_region::clustered:
                 _drop_partition_start = true;
@@ -1156,29 +1178,57 @@ future<> shard_reader::remote_reader::do_fill_buffer(flat_mutation_reader& reade
     });
 }
 
-future<> shard_reader::remote_reader::ensure_buffer_contains_all_fragments_for_last_pos(flat_mutation_reader& reader,
-        circular_buffer<mutation_fragment>& buffer, db::timeout_clock::time_point timeout) {
-    if (buffer.empty() || !buffer.back().is_range_tombstone()) {
-        return make_ready_future<>();
-    }
-
-    auto stop = [this, &reader, &buffer] {
+future<> shard_reader::remote_reader::fill_buffer(flat_mutation_reader& reader, circular_buffer<mutation_fragment>& buffer,
+        db::timeout_clock::time_point timeout) {
+    return do_fill_buffer(reader, timeout).then([this, &reader, &buffer, timeout] {
         if (reader.is_buffer_empty()) {
-            return reader.is_end_of_stream();
+            return make_ready_future<>();
         }
-        const auto& next_pos = reader.peek_buffer().position();
-        if (next_pos.region() != partition_region::clustered) {
-            return true;
-        }
-        return !next_pos.key().equal(*_schema, buffer.back().position().key());
-    };
-
-    return do_until(stop, [this, &reader, &buffer, timeout] {
-        if (reader.is_buffer_empty()) {
-            return do_fill_buffer(reader, timeout);
-        }
-        buffer.emplace_back(reader.pop_mutation_fragment());
-        return make_ready_future<>();
+        buffer = reader.detach_buffer();
+        auto stop = [this, &reader, &buffer] {
+            // The only problematic fragment kind is the range tombstone.
+            // All other fragment kinds are safe to end the buffer on, and
+            // are guaranteed to represent progress vs. the last buffer fill.
+            if (!buffer.back().is_range_tombstone()) {
+                return true;
+            }
+            if (reader.is_buffer_empty()) {
+                return reader.is_end_of_stream();
+            }
+            const auto& next_pos = reader.peek_buffer().position();
+            // To ensure safe progress we have to ensure the following:
+            //
+            // _next_position_in_partition < buffer.back().position() < next_pos
+            //
+            // * The first condition is to ensure we made progress since the
+            // last buffer fill. Otherwise we might get into an endless loop if
+            // the reader is recreated after each `fill_buffer()` call.
+            // * The second condition is to ensure we have seen all fragments
+            // with the same position. Otherwise we might jump over those
+            // remaining fragments with the same position as the last
+            // fragment's in the buffer when the reader is recreated.
+            return _tri_cmp(_next_position_in_partition, buffer.back().position()) < 0 && _tri_cmp(buffer.back().position(), next_pos) < 0;
+        };
+        // Read additional fragments until it is safe to stop, if needed.
+        // We have to ensure we stop at a fragment such that if the reader is
+        // evicted and recreated later, we won't be skipping any fragments.
+        // Practically, range tombstones are the only ones that are
+        // problematic to end the buffer on. This is due to the fact range
+        // tombstones can have the same position that multiple following range
+        // tombstones, or a single following clustering row in the stream has.
+        // When a range tombstone is the last in the buffer, we have to continue
+        // to read until we are sure we've read all fragments sharing the same
+        // position, so that we can safely continue reading from after said
+        // position.
+        return do_until(stop, [this, &reader, &buffer, timeout] {
+            if (reader.is_buffer_empty()) {
+                return do_fill_buffer(reader, timeout);
+            }
+            buffer.emplace_back(reader.pop_mutation_fragment());
+            return make_ready_future<>();
+        });
+    }).then([this, &reader, &buffer] {
+        update_next_position(reader, buffer);
     });
 }
 
@@ -1196,7 +1246,8 @@ shard_reader::remote_reader::remote_reader(
     , _ps(ps)
     , _pc(pc)
     , _trace_state(std::move(trace_state))
-    , _fwd_mr(fwd_mr) {
+    , _fwd_mr(fwd_mr)
+    , _tri_cmp(*_schema) {
 }
 
 future<shard_reader::fill_buffer_result> shard_reader::remote_reader::fill_buffer(const dht::partition_range& pr, bool pending_next_partition,
@@ -1204,7 +1255,7 @@ future<shard_reader::fill_buffer_result> shard_reader::remote_reader::fill_buffe
     // We could have missed a `fast_forward_to()` if the reader wasn't created yet.
     _pr = &pr;
     if (pending_next_partition) {
-        _last_position_in_partition = position_in_partition(position_in_partition::end_of_partition_tag_t{});
+        _next_position_in_partition = position_in_partition::for_partition_start();
     }
     return do_with(resume_or_create_reader(), circular_buffer<mutation_fragment>{},
             [this, pending_next_partition, timeout] (flat_mutation_reader& reader, circular_buffer<mutation_fragment>& buffer) mutable {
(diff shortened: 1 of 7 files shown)


denesb:
Hold on merging, the fix might not be complete (just produced a test failure).

mutation_reader.cc
```diff
@@ -1048,6 +1050,20 @@ void shard_reader::remote_reader::adjust_partition_slice() {
     auto ranges = _slice_override->default_row_ranges();
     query::trim_clustering_row_ranges_to(*_schema, ranges, last_ckey);
 
+    if (_last_position_in_partition_is_prefix) {
```
tgrabiec: If we stopped at a clustering row, we don't need to filter, because slicing will make sure we won't see any old fragments. In that case we could set `drop_needed` to false.

But I wonder if we ever need to filter. We could do this instead:

1) change `trim_clustering_row_ranges_to()` to work with `position_in_partition`, and make the resulting range include the position rather than exclude it like currently. We can always include a position_in_partition, but we cannot always exclude it (we cannot exclude the start of a prefix, which is the problem we have here). We can avoid the ugly `find_if` this way, and also this will allow us to avoid filtering by also doing this change:

We should set `_last_position_in_partition` (which now would be included rather than excluded) not from `buffer.back()` in `update_last_position()`, but in `ensure_buffer_contains_all_fragments_for_last_pos()` in the following way:
  - if we stopped at a clustering row with key `k`, set `_last_position_in_partition` to `after(key)`.
  - if we had to consume tombstones and stopped by peeking a fragment with key `k2`, set `_last_position_in_partition` to `k2`.

tgrabiec: Thinking about it, we should also always trim range tombstones to _last_position_in_partition, since after recreation we may get range tombstones starting before the start of the slice, right?

denesb:
I found another edge-case around non-full prefix positions.

mutation_reader.cc
```diff
@@ -1115,35 +1114,27 @@ flat_mutation_reader shard_reader::remote_reader::resume_or_create_reader() {
...
             }
-            if (_drop_static_row) {
-                _drop_static_row = false;
-                if (reader.peek_buffer().is_static_row()) {
-                    reader.pop_mutation_fragment();
-                }
-            }
-
-            return stop_iteration(reader.is_buffer_full() || eos);
+            return stop_iteration(reader.is_buffer_full() || reader.is_end_of_stream());
```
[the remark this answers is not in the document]
denesb: My original line of thinking for this exit condition is that since all this is going to be transferred to another shard, we better not make the trip with just a few fragments.

mutation_reader.cc
denesb: > If we stopped at a clustering row, we don't need to filter, because slicing will make sure we won't see any old fragments. In that case we could set drop_needed to false.

We can't because the `partition_start` and the `static_row` will be re-emitted again anyway.

> Thinking about it, we should also always trim range tombstones to _last_position_in_partition, since after recreation we may get range tombstones starting before the start of the slice, right?

Possibly, but with the current code those will not be emitted again (`should_drop_fragment()` will flag them up).

I like the `position_in_partition` idea. I'm already in the process of refactoring `trim_clustering_ranges_to()` because of the other issue I found after submitting this PR. The issue is, just FYI, that `trim_clustering_ranges_to()` is not prepared for a key that is a prefix, and in this condition it will eliminate ranges, whoose end bound key is a prefix of. For example, for the inputs:
```
range=[{0, 0}, {0, 99999})
key={0}
```
The range will be dropped.

In general prefixes are confusing because they do not represent a point in the key-space, rather a range in themselves, breaking code that is not prepared for this. The world would be a better place if we had prefixes replaced with full compounds with unset components pre-filled with min/max values for the type, accordingly.

tgrabiec: For `partition_start` and the `static_row` we need the old filtering, yes.

About dropping the tombstones, I think it is not correct to completely drop them, because after recreation their `end` may have been extended due to merging with later fragments.

yes, a prefix inside `clustering_range` has a different meaning than a prefix as a standalone key. The latter is positioned as if it was null-extended. While a prefix in the range means that the range either excludes or includes the whole prefix.

`position_in_partition` was supposed to solve this problem. You can always treat a range as a pair of p_i_p, and assume the range includes all p_i_p such that start < p <= end.

denesb: I don't understand how the `end` could be extended after recreating the reader. We create the reader with monotonically narrowing `range` (moving only `start`, never `end`).

tgrabiec: A range deletion [1, 5] can be also represented as [1, 3) and [3, 5]. Before recreation you could have get [1, 3) then pause, and have [3, 5] yet to consume. But after recreation you may get [1, 5]. If you ignore it, the upper consumer will only know about [1, 3) and won't know that [3, 5] is also deleted.

denesb: Right, unfortunately many (most) of our tests expect the exact representation of writes back that they provided and will fail for such a re-emitted range-tombstone.

Also, I found another problem. Consider reading a range bounded by two non-full prefixes (`-` denotes missing component):
```
[{0, -}, {2, -}]
```
Eventually we will arrive at a singular-looking range:
```
[{2, -}, {2, -}] == [{2, -}]
```
Now if the reader is recreated after having read some clustering row, the range remaining to read will look like this:
```
({2, 1}, {2, -}]
```
Depending on how you look this is an invalid or wrapping range, as `end` includes positions that are before `start`, although in practice it might just work without problems.

tgrabiec: It may look confusing, yes, becasue the meaning of the prefix is overloaded when put in a range. So you cannot just take clustering_row's key and put it in a range bound and hope it will mean that the range will start right after the key. Prefix keys have to be null-extended when put into a range.

Currently, the correct interpretation of a prefix inside a range is using prefix equality comparison. Every key which is prefixed by a prefix in the range's bound is considered equal to the bound. Then exclusiveness is used to decide whether equal things are in or out.

With position_in_partition there is arguably less confusion, since bound equality is the same as p_i_p equality.

denesb: > Thinking about it, we should also always trim range tombstones to _last_position_in_partition, since after recreation we may get range tombstones starting before the start of the slice, right?

This should already be done by upstream readers. It is illegal for readers to emit fragments outside of the query range and slice. A violation of this was fixed not long ago, see https://github.com/scylladb/scylla/issues/4104.

denesb: > We should set `_last_position_in_partition` (which now would be included rather than excluded) not from `buffer.back()` in `update_last_position()`, but in `ensure_buffer_contains_all_fragments_for_last_pos()` in the following way:
> 
>     * if we stopped at a clustering row with key `k`, set `_last_position_in_partition` to `after(key)`.
> 
>     * if we had to consume tombstones and stopped by peeking a fragment with key `k2`, set `_last_position_in_partition` to `k2`.

`ensure_buffer_contains_all_fragments_for_last_pos()` is not a reliable place to do this at, it is only invoked when needed (we return early when we are good), but we always need a `_last_position_in_partition`. I found another solution. v2 is coming soon.

denesb:
v2 pushed, changes:
* Rebased on latest master
* Roll back the drop logic generalization. We only really need to ever drop `partition_start` and `static_row`. Having a general position comparison based drop logic is misleading and unnecessarily expensive. Keep the nice `should_drop()` method though, wrapping the drop logic.
* Extend the logic which is meant to ensure that the read makes a steady progress without dropping any data to (hopefully) include all corner cases.
* Refactor `trim_clustering_row_ranges_to()` to allow passing `position_in_partition_view` as `pos` parameter. Use this to better implement the workaround for `prefix` last positions.

denesb:
The commits, as shown on the PR page, are not ordered according to their natural order, as ordered in git. Instead, they are ordered according to some random ordering I couldn't yet figure out. This will make review harder as dependent commits are ordered after the commit they provide infrastructure for.

bhalevy:
@denesb I think the order in which they are displayed here might be by author date which doesn't change when you rebase/amend the patches...

denesb:
> @denesb I think the order in which they are displayed here might be by author date which doesn't change when you rebase/amend the patches...

Yes, I suspect the same. This is unfortunately very unhelpful as it makes impossible to review the bisectability of the series.

denesb:
@tgrabiec please take another look.

mutation_reader.cc
tgrabiec: I can see that `ensure_buffer_contains_all_fragments_for_last_pos()` is always invoked by `fill_buffer()`.

clustering_bounds_comparator.hh
```diff
@@ -149,6 +149,10 @@ public:
         bool inclusive = bv._kind != bound_kind::excl_end && bv._kind != bound_kind::excl_start;
         return {typename R<clustering_key_prefix_view>::bound(bv._prefix.get().view(), inclusive)};
     }
+    static bool ranges_overlap(const schema& s, bound_view s1, bound_view e1, bound_view s2, bound_view e2) {
```
tgrabiec: There's `position_range::overlaps()` which you could reuse. We should move away from bound_view.

I didn't check if this implementation is correct.

mutation_reader.cc
```diff
@@ -1044,9 +1044,12 @@ void shard_reader::remote_reader::adjust_partition_slice() {
         _slice_override = _ps;
     }
 
-    auto& last_ckey = _last_position_in_partition->key();
+    const auto pos = _last_position_in_partition->key().is_full(*_schema) ?
```
tgrabiec: I think that if you changed the semantics of `_last_position_in_partition` to mean the first position yet to be produced, you wouldn't need this transformation, and wouldn't need as much as 3 checks in `ensure_safe_progress()`. What prevents us from doing that?

denesb: Indeed the code would be much cleaner. The problem is that the only time we actually get to peek at what is the next fragment is when the progress assurance logic kicks in, that is when the last fragment in the buffer is a `range_tombstone`. When this is not the case, `ensure_safe_progress()` returns early and we don't know what the next position is. To be able to peek the next fragment, we'd have to call `fill_buffer()` on the underlying reader, then leave that buffer there, possibly wasting the effort if the reader is evicted in the meanwhile.

If we consider reader eviction to be very reader (as it should be) then we can maybe justify always reading ahead one buffer worth of fragments, just to be able to have `_next_position_in_partition` instead of a `_last_position_in_partition`. However this would interact poorly with paging, as in the "worst" case we'd have a full buffer worth of data on each shard when the combining reader would be stopped. This is somewhat mitigated by the fact that we save readers between pages now.

tgrabiec: I think we can always calculate `_next_position_in_partition`. 

If the buffer is empty, don't change anything.

When the last fragment is not a range tombstone:
  - if it's a clustering row, use `after_row()`
  - if it's a partition start, use `for_static_row()`
  - if it's a static row, use `before_all_clustered_rows()`

If the last fragment is a range tombstone, set `_next_position_in_partition` to the first peeked position != last position in the buffer.

tgrabiec: Well, since we're only concerned with clustering positions, the partition start and static row are handled by filtering and `_next_position_in_partition` is left unchanged.

bhalevy:
Cc @avikivity

avikivity:
@bhalevy is my input needed?

bhalevy:
> @bhalevy is my input needed?

No, just fyi (I thought this was the issue @slivne referred to in our status meeting).

denesb:
v3 pushed, changes:
* Use next position, instead of the last position to calculate the starting position of the reader when recreated.
* Move away from `bound_view` (in https://github.com/scylladb/scylla/pull/4740/commits/0ebbb0e9aab116cc339234156eda86cf0de3bfd9), work in terms of `position_in_partition`. Since not needed anymore,  `bound_view::ranges_overlap()` was dropped.
* Change `as_range_end()` to create an inclusive bound if `_bound_weight == bound_weight::after_all_prefixed` (instead of if `_bound_weight != bound_weight::before_all_prefixed`).

As I heard that's how force pushed should be done, I pushed a tag for v2 ([multishard_combining_reader_more_robust_reader_recreation/v2](https://github.com/denesb/scylla/commit/9cf6f9828043f7f14c9c8b07d89cdecb637b8705)) and another one for v3 (([multishard_combining_reader_more_robust_reader_recreation/v3](https://github.com/denesb/scylla/commit/090adc44124ff985a2c0ec40ef4b6eedd60d75f1)), but they don't show up here in the PR, nor have I found a good way to link to them via the github interface, other than linking to the commit they point to.

denesb:
@tgrabiec pleas take another look.

denesb:
(sneakily pushed a commit I forgot to include, just an indentation restoration patch, nothing to see there)

mutation_reader.cc
```diff
@@ -1216,11 +1219,25 @@ future<const mutation_fragment*> shard_reader::remote_reader::ensure_buffer_cont
...
+        const auto res = _tri_cmp(buffer.back().position(), _position_in_partition);
+        // To ensure progress we have to ensure three tings:
+        // * We have seen a fragment with position that is *greater* the last
+        //   seen one from the last `fill_buffer()` call.
+        // * The first unconsumed fragment in the reader has a position that is
+        //   *greater* than the last fragment in the buffer. This means that we
+        //   are guaranteed to have seen all fragment for the position of the
+        //   last fragment in the buffer and therefore it will be safe to
+        //   continue *after* that position in the next `fill_buffer()` call
+        //   (if the reader is evicted in the meanwhile).
```
tgrabiec: If the second point is satisfied, do we still need to check the first? Isn't it also guaranteed?

mutation_reader.cc
```diff
@@ -1216,11 +1219,25 @@ future<const mutation_fragment*> shard_reader::remote_reader::ensure_buffer_cont
...
+        return (_position_kind == position_kind::previous ? res > 0 : res >= 0) &&
+                _tri_cmp(next_pos, position_in_partition_view::after_key(buffer.back().key())) > 0;
```
tgrabiec: I don't understand why you use `after_key()` here. Why not this:

```
_tri_cmp(next_pos, buffer.back().position())) > 0
```

slivne:
@denesb  ^^

mutation_reader.cc
denesb: We have to ensure:
```
_next_position_in_partition < buffer.back_position() < reader.peek_buffer().position()
```
Ensuring either condition won't imply the other.

denesb:
v4 pushed, changes:
* Simplify the calculation of the position where the reader has to be recreated. We now always use the next position. This is calculated in `update_next_position()`.

denesb:
Commit level changelog:
* [CHANGED] https://github.com/scylladb/scylla/pull/4740/commits/074be7faec7297742424d0390d6f3e03ff5d7292. This is the merge of two v3 commits, containing the simplifications on top.
* [NEW] https://github.com/scylladb/scylla/pull/4740/commits/46e7ad00f940b5cadfccea9a9e56ea67b03bfe67.

I added this because at this point even I can't really untangle the mess that is this PR. This is horrible! I wonder how it would look at v10.

mutation_reader.cc
```diff
@@ -1156,29 +1178,57 @@ future<> shard_reader::remote_reader::do_fill_buffer(flat_mutation_reader& reade
...
+                return reader.is_end_of_stream();
+            }
+            const auto& next_pos = reader.peek_buffer().position();
+            // To ensure safe progress we have to ensure the following:
+            //
+            // _next_position_in_partition < buffer.back().position() < next_pos
+            //
+            // * The first condition is to ensure we made progress since the
+            // last buffer fill. Otherwise we might get into an endless loop if
+            // the reader is recreated after each `fill_buffer()` call.
```
tgrabiec: Isn't `_next_position_in_partition`, at the time of recreation, guaranteed to be always greater than `buffer.back().position`? 

We ensure that by the second check: `_tri_cmp(buffer.back().position(), next_pos) < 0`

tgrabiec:
On Fri, Aug 9, 2019 at 10:11 AM Botond Dénes <notifications@github.com>
wrote:

> *@denesb* commented on this pull request.
>
> I will attempt to simplify the logic using your suggestions. Indeed, this
> logic is way more complicated than I would like already.
>
> In query.cc
> <https://github.com/scylladb/scylla/pull/4740#discussion_r312363077>:
>
> >              it = ranges.erase(it);
>              continue;
> -        } else if (cmp(start_bound(range), key)) {
> -            assert(cmp(key, end_bound(range)));
> -            auto r = reversed ? clustering_range(it->start(), clustering_range::bound { key, false })
> -                : clustering_range(clustering_range::bound { key, false }, it->end());
> +        } else if (cmp(start_bound(*it), pos) <= 0) {
> +            assert(cmp(pos, end_bound(*it)) < 0);
> +            auto r = reversed ? clustering_range(it->start(), pos.as_range_end()) : clustering_range(pos.as_range_start(), it->end());
>
> The behaviour has now been changed such that we trim to (-int, pos}, } ->
> whatever the inclusiveness of pos is. So now the caller controls whether
> they wish to trim to (-int, pos] or (-int, pos]. That was my intent
> anyway.
>

You can have a position which is before or after a key, but that's a
separate thing. If you pass "pos" which is before a key, you can still
create ranges which either include or exclude that position. If your range
includes "before(x)", you won't see a range tombstone which starts
before(x). If you include it, you will see it.

Maybe it's a documentation issue, I'm not sure, but currently, the doc
contradicts the implementation.

>

position_in_partition.hh
```diff
@@ -194,6 +194,10 @@ public:
         return bound_view(*_ck, _bound_weight == bound_weight::before_all_prefixed ? bound_kind::excl_end : bound_kind::incl_end);
     }
 
+    bool adjacent(const schema& s, position_in_partition_view o) const {
+        return as_end_bound_view().adjacent(s, o.as_start_bound_view());
```
tgrabiec: Note that for position ranges, adjacent() is the same as equality. [x, y) is adjacent to [z, w) if y == z.

mutation_reader.cc
tgrabiec: Can you give an example sequence of events which leads to an endless loop without the first check?

tgrabiec: Is it because we may emit fragments which start before the requested ranges start?

denesb: It sorta is. But not necessarily. Consider the following:
```
read_buffer={..., {rt, pos=[1, 100)}, {cr, pos=1}}
```
When we recreate the reader with pos `(1, +inf)` clustering range, the underlying reader might re-emit all those range tombstones with `pos=[1, 100)`, with the position trimmed to the clustering range: `pos=(1, 100)`. If we again stop at this position we will be unable to make progress if the reader is evicted after each fill buffer call. This scenario happened in the tests.

position_in_partition.hh
denesb: I'm confused. Are position ranges always `[)` (closed start bound, open end bound)?

mutation_reader.cc
denesb: On second though, this might not be the case now that we use the next position. Let me try.

denesb:
> You can have a position which is before or after a key, but that's a separate thing. If you pass "pos" which is before a key, you can still create ranges which either include or exclude that position. If your range includes "before(x)", you won't see a range tombstone which starts before(x). If you include it, you will see it. Maybe it's a documentation issue, I'm not sure, but currently, the doc contradicts the implementation.

Since clustering ranges don't include a weight, `before(x)` translates to `{x, inclusive}` (for a start bound) and `{x, exclusive}` if an end bound. No?

mutation_reader.cc
denesb: I removed this condition and the test hangs.

tgrabiec: I would like to understand what exactly happens.

position_in_partition.hh
tgrabiec: Yes, in the domain of `position_in_partition`, they're always `[a, b)`. The end p_i_p (`b`) can be still either before or after a key, in which case the range either includes the key or it excludes the key. If you have ranges `[a, before(x))` and `[before(x), c)`, such ranges are adjacent, because all elements in the range `[a, c)` belong to either one or the other.

mutation_reader.cc
denesb: Consider the following stream:
`f0=rt{k1, -1}, f1=cr{k1, 0}`
If `f0` is the last element in the buffer, we recreate the reader with `[k1, +inf)` clustering range, however the underlying reader will re-emit the range tombstone with `rt{k1, -1}`, which will again become the last element in the buffer (buffer_size=1). We have an infinite loop. This is a contrived test scenario which stresses the reader recreation as much as possible, however a situation like this can theoretically happen in a real server, given enough range tombstones and a sufficient amount of bad luck.

tgrabiec:
> > You can have a position which is before or after a key, but that's a separate thing. If you pass "pos" which is before a key, you can still create ranges which either include or exclude that position. If your range includes "before(x)", you won't see a range tombstone which starts before(x). If you include it, you will see it. Maybe it's a documentation issue, I'm not sure, but currently, the doc contradicts the implementation.
> 
> Since clustering ranges don't include a weight, `before(x)` translates to `{x, inclusive}` (for a start bound) and `{x, exclusive}` if an end bound. No?

Yes, but how this fact relates to the problem at hand?

I'll reiterate. The doc says that the function returns `(-int, pos]`. Which is a position range, since pos is a position_in_partition. But it in fact returns a clustering range, which corresponds to a position range of `(-inf, pos)`.

denesb:
> I'll reiterate. The doc says that the function returns `(-int, pos]`. Which is a position range, since pos is a position_in_partition. But it in fact returns a clustering range, which corresponds to a position range of `(-inf, pos)`.

The doc was written in terms of clustering ranges and I simply translate position in partition to a clustering range bound internally.

tgrabiec:
> > I'll reiterate. The doc says that the function returns `(-int, pos]`. Which is a position range, since pos is a position_in_partition. But it in fact returns a clustering range, which corresponds to a position range of `(-inf, pos)`.
> 
> The doc was written in terms of clustering ranges and I simply translate position in partition to a clustering range bound internally.

The doc is still confusing, I think. When I read a notation like `(-inf, pos]`, I assume that pos is included. But your transformation excludes it. 

It's fine to return a `clustering_range`, but it should be equivalent to a p_i_p range that we state in the doc.

denesb:
> > > I'll reiterate. The doc says that the function returns `(-int, pos]`. Which is a position range, since pos is a position_in_partition. But it in fact returns a clustering range, which corresponds to a position range of `(-inf, pos)`.
> > 
> > 
> > The doc was written in terms of clustering ranges and I simply translate position in partition to a clustering range bound internally.
> 
> The doc is still confusing, I think. When I read a notation like `(-inf, pos]`, I assume that pos is included. But your transformation excludes it.
> 
> It's fine to return a `clustering_range`, but it should be equivalent to a p_i_p range that we state in the doc.

Ok, I will clarify this in the doc, and will also make sure the code does what is expected.

denesb:
v5 pushed, changes:
* Drop patch that adds `position_in_partition_view::adjacent()`, instead just compare the positions for equality.
* Drop patch that adds `position_in_partition_view::as_range_start()` and `position_in_partition_view::as_range_end()`.
* Fix `trim_clustering_ranges_to()` implementation to match its documentation.

query.cc
```diff
@@ -71,34 +71,38 @@ std::ostream& operator<<(std::ostream& out, const specific_ranges& s) {
...
+                clustering_range(clustering_range::bound(pos.key(), pos.get_bound_weight() != bound_weight::after_all_prefixed), it->end());
             *it = std::move(r);
         }
         ++it;
     }
 }
 
+void trim_clustering_row_ranges_to(const schema& s, clustering_row_ranges& ranges, const clustering_key& key, bool reversed) {
+    return trim_clustering_row_ranges_to(s, ranges,
+            position_in_partition_view(key, reversed ? bound_weight::before_all_prefixed : bound_weight::after_all_prefixed), reversed);
```
tgrabiec: If the key is a non-full key, using `after_all_prefixed` may cause us to skip over some rows, right?

I think we should use `position_in_partition::after_key()` here, and fix it to work with prefix keys. 

We do this kind of trimming in sstable reader, which extends non-full keys:
```
            auto ck = key;
            bool was_non_full_key = clustering_key::make_full(*_schema, ck);
            auto end_kind = was_non_full_key ? bound_kind::excl_end : bound_kind::incl_end;
            assert(!_stored_tombstone);
            auto rt = range_tombstone(std::move(_opened_range_tombstone->ck),
                _opened_range_tombstone->kind,
                ck,
                end_kind,
                _opened_range_tombstone->tomb);
            sstlog.trace("mp_row_consumer_m {}: push({})", this, rt);
            _opened_range_tombstone->ck = std::move(ck);
            _opened_range_tombstone->kind = was_non_full_key ? bound_kind::incl_start : bound_kind::excl_start;

```

tgrabiec: It is a preexisting problem though. We can fix incrementally.

tgrabiec: https://github.com/scylladb/scylla/issues/4839

PR #4742, opened 2019-07-19
Add ZStandard compression.

### LINKED ISSUE #2613: Can we add ZSTD compression?
https://github.com/MatejTymes/cassandra-zstd

Cassandra supports ZSTD compression by including the jar files built from the above source.

kbr-:
This adds the option to compress sstables using the Zstandard algorithm
(https://facebook.github.io/zstd/).
To use, pass 'sstable_compression': 'org.apache.cassandra.io.compress.ZstdCompressor'
to the 'compression' argument when creating a table.
You can also specify a 'compression_level' (default is 3). See Zstd documentation for the available
compression levels.
Resolves #2613.

This PR also fixes a bug in sstables/compress.cc, where chunk length in bytes
was passed to the compressor as chunk length in kilobytes. Fortunately,
none of the compressors implemented until now used this parameter.

Example usage (assuming there exists a keyspace 'a'):
```
create table a.a (a text primary key, b int) with compression = {'sstable_compression': 'org.apache.cassandra.io.compress.ZstdCompressor', 'compression_level': 1, 'chunk_length_in_kb': '64'};
```
Notes:
1. The code uses an external dependency: https://github.com/facebook/zstd. Since I'm using "experimental" features of the library (using my own allocated memory to store the compression/decompression contexts), according to the library's documentation we need to link it statically (https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L63). I added a git submodule.
2. The compressor performs some dynamic allocations. Depending on the specified chunk length and/or compression level the allocations might be big and seastar throws warnings. But with reasonable chunk length sizes it should be OK.
3. It doesn't yet provide an option to train it with dictionaries, but that should be easy to add in another commit.


dorlaor:
Cheers Kamil, it will be interesting to compare the compression speed/size

avikivity:
Please try to upstream the zstd changes (not a blocker, but better to start early).

avikivity:
Can it read Cassandra sstables written compressed with zstd?

configure.py
```diff
@@ -1278,7 +1280,8 @@ def query_seastar_flags(seastar_pc_file, link_static_cxx=False):
                 f.write('build $builddir/{}/{}: ar.{} {}\n'.format(mode, binary, mode, str.join(' ', objs)))
             else:
                 objs.extend(['$builddir/' + mode + '/' + artifact for artifact in [
-                    'libdeflate/libdeflate.a'
+                    'libdeflate/libdeflate.a',
+                    'zstd/libzstd.a'
```
denesb: Add a trailing comma so we don't have to touch the existing line next time.

configure.py
```diff
@@ -1392,6 +1395,9 @@ def query_seastar_flags(seastar_pc_file, link_static_cxx=False):
         f.write('rule libdeflate.{mode}\n'.format(**locals()))
         f.write('    command = make -C libdeflate BUILD_DIR=../build/{mode}/libdeflate/ CFLAGS="{libdeflate_cflags}" CC={args.cc} ../build/{mode}/libdeflate//libdeflate.a\n'.format(**locals()))
         f.write('build build/{mode}/libdeflate/libdeflate.a: libdeflate.{mode}\n'.format(**locals()))
+        f.write('rule zstd.{mode}\n'.format(**locals()))
+        f.write('    command = make -C zstd/lib BUILD_DIR=../../build/{mode}/zstd/ CFLAGS="{zstd_cflags}" CC={args.cc} ../../build/{mode}/zstd//libzstd.a\n'.format(**locals()))
```
denesb: What's this `../../build`? Isn't that the same dir we are in now?

sstables/compress.cc
```diff
@@ -264,7 +264,7 @@ local_compression::local_compression(const compression& c)
         sstring n(c.name.value.begin(), c.name.value.end());
         return compressor::create(n, [&c, &n](const sstring& key) -> compressor::opt_string {
             if (key == compression_parameters::CHUNK_LENGTH_KB || key == compression_parameters::CHUNK_LENGTH_KB_ERR) {
-                return to_sstring(c.chunk_len);
+                return to_sstring(c.chunk_len / 1024);
```
denesb: Should've gone into a separate commit.

zstd.cc
```diff
@@ -0,0 +1,134 @@
...
+    std::unique_ptr<char[], free_deleter> _cctx_raw;
+    ZSTD_CCtx* _cctx;
+
+    // Decompression context.
+    std::unique_ptr<char[], free_deleter> _dctx_raw;
+    ZSTD_DCtx* _dctx;
+public:
+    zstd_processor(const opt_getter&);
+
+    size_t uncompress(const char* input, size_t input_len, char* output,
```
denesb: Prefix overriden virtuals with `virtual`.

zstd.cc
```diff
@@ -0,0 +1,134 @@
...
+class zstd_processor : public compressor {
+    int _compression_level = 3;
+
+    // Compression context.
+    std::unique_ptr<char[], free_deleter> _cctx_raw;
+    ZSTD_CCtx* _cctx;
+
+    // Decompression context.
+    std::unique_ptr<char[], free_deleter> _dctx_raw;
+    ZSTD_DCtx* _dctx;
```
denesb: I see initialization code for this pointer, but no de-initialization (the class doesn't even have a destructor).

kbr-: The default destructor does the right thing (i.e., calling the destructor of these unique_ptrs, which free the memory allocated for the contexts)
https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L1272

denesb:
> Question: how do I test it (requires memtable flushes)? DTest?

Can be tested using unit tests too. Dtests in general are required when the test involves multiple nodes, otherwise it is generally better to use unit tests as they are way faster.

configure.py
kbr-: Suppose the working directory is Scylla's main directory.
Zstd's Makefile lies in ./zstd/lib.
We tell the Makefile to build the "../../build/dev/zstd//libzstd.a" target (when mode=dev), which is specified here:
https://github.com/kbr-/zstd/blob/1.4.0/lib/Makefile#L152
in this case $(BUILD_DIR) = "../../build/dev/zstd/".
"../../build/dev/zstd//libzstd.a" where PWD="./zstd/lib" resolves to to ./build/dev/zstd/ibzstd.a.

zstd.cc
kbr-: That would be inconsistent with other compressors. See compress.cc.

configure.py
denesb: I see.

zstd.cc
denesb: I see. Maybe `_dctx_storage` would be a better name then (and consequently `_cctx_storage`).

zstd.cc
denesb: I'm looking but I'm seeing `virtual` prepended to such lines: https://github.com/scylladb/scylla/blob/7604980d638ef83cf8f8d8083fa1a82c00572ba4/sstables/compress.cc#L386

Yes, we have many undocumented de-facto coding standards that give way to pointless discussions around what is the *true* standard, as you can pretty much find precedence for anything in our code-base.

kbr-: https://github.com/scylladb/scylla/blob/7604980d638ef83cf8f8d8083fa1a82c00572ba4/compress.cc#L31

denesb: Yeah, like I said, there can be found precedents for pretty-much any code-style.
BTW why was this not added to `compress.cc` like the other compressors?

zstd.cc
kbr-: What about `_dctx_bytes`?

denesb: `bytes` sounds more like a unit of measurement to me (for size), than a storage place. I guess raw is fine too, maybe add a comment.

zstd.cc
kbr-: Separate compilation units = code can be compiled in parallel, which speeds up the compilation times. Plus it's more readable IMO (you don't spend time wondering which code belongs to which compressor). I'd even refactor what we have and put every compressor to a separate file in a directory like compressor/.

kbr-:
@avikivity 
1. Copying tables from https://github.com/apache/cassandra/tree/cassandra-3.9 with plugin https://github.com/MatejTymes/cassandra-zstd to Scylla: works
2. Copying tables from Scylla to cassandra-3.9 with plugin as above: works
3. Copying tables from Scylla to current https://github.com/apache/cassandra/tree/trunk (cassandra-4.0 dev branch, supports Zstd natively): works
4. Copying from current Cassandra 4.0 to Scylla: doesn't work since new Cassandra uses a new sstable format ('na') which we don't support.

kbr-:
@denesb ok, I will write unit tests.

kbr-:
@avikivity the out of dir build feature won't be necessary for us in the long run --- when the advanced features stop being "experimental", we will be able to link the library dynamically and use the versions from standard Linux distribution repositories (e.g. Fedora has a libzstd-devel package).
Should I try to upstream those changes nevertheless?

zstd.cc
denesb: To a certain point yes, but it also has a link-time overhead which hurts edit-compile-run cycles.
Also this is inconsistent with the other compressors. :)

kbr-: Linking is super fast though compared to compilation (especially if you use lld).
I can move the other compressors to separate files :P

haaawk: I'd keep the compressor in a separate file since it's not dependant on any of the other compressors. I don't think bundling multiple unrelated classes in a single file makes sense. The fact that they implement the same interface does not make them related.

kbr-:
@dorlaor thanks. I'll try to run some benchmarks later.

kbr-:
v2:
- Took out the chunk_length bugfix into a separate commit.
- Added a comma after 'zstd/libzstd.a' in configure.py.
- Added comments describing _{c,d}ctx_raw.
- Enabled a forgotten test in tests/sstable_3_x_test.cc and does some minor refactoring.
- Added some unit tests for Zstd compression.

configure.py
kbr-: Done

sstables/compress.cc
kbr-: Done

zstd.cc
kbr-: Added a comment

tests/sstable_3_x_test.cc
```diff
@@ -1732,7 +1732,13 @@ static void test_partition_key_with_values_of_different_types_read(const sstring
...
     auto abj = defer([] { await_background_jobs().get(); });
     test_partition_key_with_values_of_different_types_read(
-        UNCOMPRESSED_PARTITION_KEY_WITH_VALUES_OF_DIFFERENT_TYPES_PATH);
+        UNCOMPRESSED_PARTITION_KEY_WITH_VALUES_OF_DIFFERENT_TYPES_PATH, compression_parameters::no_compression());
+}
+
+SEASTAR_THREAD_TEST_CASE(test_compressed_partition_key_with_values_of_different_types_read) {
+    auto abj = defer([] { await_background_jobs().get(); });
+    test_partition_key_with_values_of_different_types_read(
+        COMPRESSED_PARTITION_KEY_WITH_VALUES_OF_DIFFERENT_TYPES_PATH, compressor::lz4);
```
denesb: Should do a round with each available compressor.

tests/sstable_3_x_test.cc
```diff
@@ -1741,6 +1741,52 @@ SEASTAR_THREAD_TEST_CASE(test_compressed_partition_key_with_values_of_different_
...
+//      WITH compression = {'sstable_compression': 'org.apache.cassandra.io.compress.ZstdCompressor',
+//                          'compression_level': 5,
+//                          'chunk_length_in_kb': 4};
+//
+// And for each `i` in the range [1, 2000],
+// INSERT INTO test_ks.test_table (val1, val2) VALUES (0, i);
+
+static thread_local const sstring ZSTD_MULTIPLE_CHUNKS_PATH =
+    "tests/sstables/3.x/compressed/zstd_multiple_chunks";
+static thread_local const schema_ptr ZSTD_MULTIPLE_CHUNKS_SCHEMA =
```
denesb: Please no more `thread_local`. We want to get rid of these.

kbr-:
I did some compression benchmarks.
I took a reasonably structured file (specifically, a list of actors from IMDB: actors.list.gz from http://ftp.sunet.se/mirror/archive/ftp.sunet.se/pub/tv+movies/imdb/), inserted it into the database under 10 different partitions in chunks of 1KB under increasing clustering keys, and measured the time it takes to perform a memtable flush and space taken by the sstables.
I did it for 3 different chunk lengths (4, 16, 64 KB), using first 100M or first 10M of the file, with different compressors, including different compression levels for Zstd.
The results showed that the chunk length doesn't really affect the results (I didn't perform good statistical analysis - just looking at the plots). Also, the relative differences between compressors are similar for 100M and 10M.

Here are some plots. All for chunk length 64kb - other plots really look the same.
100M, flush times:
image
100M, space:
image

10M, flush times:
image
10M, space:
image

I also did the same with a file consisting of zeros only, for 10M:
zeros 10M, times:
image
zeros 10M, space:
image

My opinionated conclusions:
1. zstd with compression levels > 9 is garbage (for the data I used at least),
2. zstd 1 works really well. The space used is not much worse than deflate, much better than snappy and lz4, and time is better than deflate, close to lz4 and snappy, sometimes even better.

glommer:
I would also like to check that the new algorithm is at least not more
prone to stalls than the existing ones.

Checking one of your large flushes in a singlr cpu at a 1ms report
threshold should be enough

kbr-:
Here are some plots. Measured with 10M of  the actors dataset, chunk length 64KB:
number of stalls during flush:
image

means and standard deviations:
image

Notes:
- duration of each stall shouldn't depend on on how much data needs to be flushed, since it's flushed in chunks and each chunk is flushed in a different task. It does depend on the chunk length though, but even with 64 KB it's mostly in the 1-1.5ms range.
- number of stalls does depend on the amount of data. Should scale linearly.

As we can see, it's not bad (unless you count a task calculating for 2ms a stall).

glommer:
On Wed, Jul 24, 2019 at 12:02 PM kbr- <notifications@github.com> wrote:

> Here are some plots. Measured with 10M of the actors dataset, chunk length
> 64KB:
> number of stalls during flush:
> [image: image]
> <https://user-images.githubusercontent.com/6516501/61808776-6195f880-ae3c-11e9-9505-f28e3eb6ac73.png>
>
> means and standard deviations:
> [image: image]
> <https://user-images.githubusercontent.com/6516501/61809160-0fa1a280-ae3d-11e9-82dc-b2cc8b0a0959.png>
>
> Notes:
>
>    - duration of each stall shouldn't depend on on how much data needs to
>    be flushed, since it's flushed in chunks and each chunk is flushed in a
>    different task. It does depend on the chunk length though, but even with 64
>    KB it's mostly in the 1-1.5ms range.
>    - number of stalls does depend on the amount of data. Should scale
>    linearly.
>
> As we can see, it's not bad (unless you count a task calculating for 2ms a
> stall).
>

I do.


We can still merge this, specially  if this not the default compressor, but
to make this useful we should decode these stalls, make sure they are
really coming from the library and not a coincidence, and then see what we
can do to make them disappear.

kbr-:
```
void seastar::backtrace<seastar::backtrace_buffer::append_backtrace()::{lambda(seastar::frame)#1}>(seastar::backtrace_buffer::append_backtrace()::{lambda(seastar::frame)#1}&&) at crtstuff.c:?
seastar::print_with_backtrace(seastar::backtrace_buffer&) at reactor.cc:?
seastar::internal::cpu_stall_detector::generate_trace() at crtstuff.c:?
seastar::internal::cpu_stall_detector::maybe_report() at crtstuff.c:?
seastar::internal::cpu_stall_detector::on_signal() at crtstuff.c:?
seastar::reactor::block_notifier(int) at crtstuff.c:?
?? ??:0
ZSTD_HcFindBestMatch_selectMLS at zstd_lazy.c:?
ZSTD_compressBlock_lazy2 at ??:?
ZSTD_compressBlock_internal at zstd_compress.c:?
ZSTD_compress_frameChunk at zstd_compress.c:?
ZSTD_compressContinue_internal at zstd_compress.c:?
ZSTD_compressEnd at ??:?
ZSTD_compress_advanced_internal at ??:?
ZSTD_compress_usingDict at ??:?
ZSTD_compressCCtx at ??:?
zstd_processor::compress(char const*, unsigned long, char*, unsigned long) const at crtstuff.c:?
sstables::local_compression::compress(char const*, unsigned long, char*, unsigned long) const at crtstuff.c:?
compressed_file_data_sink_impl<crc32_utils, (compressed_checksum_mode)1>::put(seastar::temporary_buffer<char>) at crtstuff.c:?
seastar::output_stream<char>::put(seastar::temporary_buffer<char>) at crtstuff.c:?
seastar::output_stream<char>::slow_write(char const*, unsigned long) at crtstuff.c:?
seastar::output_stream<char>::write(char const*, unsigned long) at crtstuff.c:?
sstables::file_writer::write(std::basic_string_view<signed char, std::char_traits<signed char> >) at crtstuff.c:?
sstables::mc::writer::write_clustered(clustering_row const&, unsigned long) at crtstuff.c:?
sstables::mc::writer::consume(clustering_row&&) at crtstuff.c:?
sstables::sstable_writer::consume(clustering_row&&) at crtstuff.c:?
void flat_mutation_reader::impl::consume_pausable_in_thread<std::reference_wrapper<flat_mutation_reader::impl::consumer_adapter<sstables::sstable_writer> >, flat_mutation_reader::consume_in_thread<sstables::sstable_writer>(sstables::sstable_writer, std::chrono::time_point<seastar::lowres_clock, std::chrono::duration<long, std::ratio<1l, 1000l> > >)::{lambda(dht::decorated_key const&)#1}>(std::reference_wrapper<flat_mutation_reader::impl::consumer_adapter<sstables::sstable_writer> >, flat_mutation_reader::consume_in_thread<sstables::sstable_writer>(sstables::sstable_writer, std::chrono::time_point<seastar::lowres_clock, std::chrono::duration<long, std::ratio<1l, 1000l> > >)::{lambda(dht::decorated_key const&)#1}, std::chrono::time_point<seastar::lowres_clock, std::chrono::duration<long, std::ratio<1l, 1000l> > >) at crtstuff.c:?
seastar::noncopyable_function<void ()>::direct_vtable_for<seastar::async<sstables::sstable::write_components(flat_mutation_reader, unsigned long, seastar::lw_shared_ptr<schema const>, sstables::sstable_writer_config const&, encoding_stats, seastar::io_priority_class const&)::{lambda()#1}>(seastar::thread_attributes, std::decay&&, (std::decay<sstables::sstable::write_components(flat_mutation_reader, unsigned long, seastar::lw_shared_ptr<schema const>, sstables::sstable_writer_config const&, encoding_stats, seastar::io_priority_class const&)::{lambda()#1}>::type&&)...)::{lambda(seastar::async<{lambda()#1}>(seastar::futurize<std::result_of<std::decay<auto:1>::type ()>::type>::type, seastar::thread_attributes, std::decay<auto:1>::type&&)::work&)#1}::operator()(seastar::async<{lambda()#1}>(seastar::futurize<std::result_of<std::decay<{lambda()#1}>::type ()>::type>::type, seastar::thread_attributes, std::decay<{lambda()#1}>::type&&)::work)::{lambda()#1}>::call(seastar::noncopyable_function<void ()> const*) at sstables.cc:?
seastar::thread_context::main() at crtstuff.c:?
```

Looks like the compression algorithm needs some time.
I don't immediately see a way to make this dissapear: the library needs the CPU for the whole duration of compression.

Also notice that Zstd 1 is better than Deflate, which we already use.

glommer:
On Wed, Jul 24, 2019 at 12:15 PM kbr- <notifications@github.com> wrote:

> void seastar::backtrace<seastar::backtrace_buffer::append_backtrace()::{lambda(seastar::frame)#1}>(seastar::backtrace_buffer::append_backtrace()::{lambda(seastar::frame)#1}&&) at crtstuff.c:?
> seastar::print_with_backtrace(seastar::backtrace_buffer&) at reactor.cc:?
> seastar::internal::cpu_stall_detector::generate_trace() at crtstuff.c:?
> seastar::internal::cpu_stall_detector::maybe_report() at crtstuff.c:?
> seastar::internal::cpu_stall_detector::on_signal() at crtstuff.c:?
> seastar::reactor::block_notifier(int) at crtstuff.c:?
> ?? ??:0
> ZSTD_HcFindBestMatch_selectMLS at zstd_lazy.c:?
> ZSTD_compressBlock_lazy2 at ??:?
> ZSTD_compressBlock_internal at zstd_compress.c:?
> ZSTD_compress_frameChunk at zstd_compress.c:?
> ZSTD_compressContinue_internal at zstd_compress.c:?
> ZSTD_compressEnd at ??:?
> ZSTD_compress_advanced_internal at ??:?
> ZSTD_compress_usingDict at ??:?
> ZSTD_compressCCtx at ??:?
> zstd_processor::compress(char const*, unsigned long, char*, unsigned long) const at crtstuff.c:?
> sstables::local_compression::compress(char const*, unsigned long, char*, unsigned long) const at crtstuff.c:?
> compressed_file_data_sink_impl<crc32_utils, (compressed_checksum_mode)1>::put(seastar::temporary_buffer<char>) at crtstuff.c:?
> seastar::output_stream<char>::put(seastar::temporary_buffer<char>) at crtstuff.c:?
> seastar::output_stream<char>::slow_write(char const*, unsigned long) at crtstuff.c:?
> seastar::output_stream<char>::write(char const*, unsigned long) at crtstuff.c:?
> sstables::file_writer::write(std::basic_string_view<signed char, std::char_traits<signed char> >) at crtstuff.c:?
> sstables::mc::writer::write_clustered(clustering_row const&, unsigned long) at crtstuff.c:?
> sstables::mc::writer::consume(clustering_row&&) at crtstuff.c:?
> sstables::sstable_writer::consume(clustering_row&&) at crtstuff.c:?
> void flat_mutation_reader::impl::consume_pausable_in_thread<std::reference_wrapper<flat_mutation_reader::impl::consumer_adapter<sstables::sstable_writer> >, flat_mutation_reader::consume_in_thread<sstables::sstable_writer>(sstables::sstable_writer, std::chrono::time_point<seastar::lowres_clock, std::chrono::duration<long, std::ratio<1l, 1000l> > >)::{lambda(dht::decorated_key const&)#1}>(std::reference_wrapper<flat_mutation_reader::impl::consumer_adapter<sstables::sstable_writer> >, flat_mutation_reader::consume_in_thread<sstables::sstable_writer>(sstables::sstable_writer, std::chrono::time_point<seastar::lowres_clock, std::chrono::duration<long, std::ratio<1l, 1000l> > >)::{lambda(dht::decorated_key const&)#1}, std::chrono::time_point<seastar::lowres_clock, std::chrono::duration<long, std::ratio<1l, 1000l> > >) at crtstuff.c:?
> seastar::noncopyable_function<void ()>::direct_vtable_for<seastar::async<sstables::sstable::write_components(flat_mutation_reader, unsigned long, seastar::lw_shared_ptr<schema const>, sstables::sstable_writer_config const&, encoding_stats, seastar::io_priority_class const&)::{lambda()#1}>(seastar::thread_attributes, std::decay&&, (std::decay<sstables::sstable::write_components(flat_mutation_reader, unsigned long, seastar::lw_shared_ptr<schema const>, sstables::sstable_writer_config const&, encoding_stats, seastar::io_priority_class const&)::{lambda()#1}>::type&&)...)::{lambda(seastar::async<{lambda()#1}>(seastar::futurize<std::result_of<std::decay<auto:1>::type ()>::type>::type, seastar::thread_attributes, std::decay<auto:1>::type&&)::work&)#1}::operator()(seastar::async<{lambda()#1}>(seastar::futurize<std::result_of<std::decay<{lambda()#1}>::type ()>::type>::type, seastar::thread_attributes, std::decay<{lambda()#1}>::type&&)::work)::{lambda()#1}>::call(seastar::noncopyable_function<void ()> const*) at sstables.cc:?
> seastar::thread_context::main() at crtstuff.c:?
>
> Looks like the compression algorithm needs some time.
> I don't immediately see a way to make this dissapear: the library needs
> the CPU for the whole duration of compression.
>
> Also notice that Zstd 1 is better than Deflate, which we already use.
>

We don't use, we list. I've never seen anyone actually using it.

I don't understand what is the motivation for the new compressor, so we can
probably merge it and open a bug for a future investigation.
We may have to try to understand how exactly the algorithm works in the
library and if we can yield work.

Another option would be to each compression chunk in a seastar::alien()
context and measure how much slower it gets. I understand that we're adding
this compressor for data density, not speed reasons, right?

I'd prefer something 10% slower that doesn't stall than something faster
that does.

kbr-:
1. Motivation is: users requested it. See the linked issue.
2. Zstd 1 has significantly better compression ratios than lz4, with similar speed.
3. The charts above come from a single run. I should've run each compressor multiple times and merge the results. I'll do that tommorow. Then you'll see that even no compression and lz4 scenarios produce 1ms stalls.

kbr-:
Motivation is: users requested it. See the linked issue. Also, Zstd 1 has significantly better compression ratios than lz4, with similar speeds.

The plots above come from single runs. I should've ran each compression algorithm multiple times and merged the results. I'll do that tomorrow. Then you'll see that lz4, or even no compression scenarios also produce 1ms stalls (on my computer).

tzach:
Equivalent Apache Cassandra ticket https://issues.apache.org/jira/browse/CASSANDRA-14482

glommer:
I am well aware we have 1ms events, and they are not coming from the current compression algorithms (at least not the ones we actually use).

Each one of those events maps directly to a tear shed by a Scylla user, so we should invest what we can not to make it worse.

Maybe to avoid investing too much in understanding what the library is doing we can explore the performance of using alien threads here.

tests/sstable_3_x_test.cc
kbr-: I removed thread_local from the path, but I can't remove it from the schema builder, because I get errors like this one:
```
unknown location(0): fatal error: in "test_uncompressed_partition_key_with_values_of_different_types_read": sstables::malformed_sstable_exception: bool_val definition in serialization header does not match schema. Expected org.apache.cassandra.db.marshal.BooleanType but got org.apa
che.cassandra.db.marshal.BooleanType in sstable tests/sstables/3.x/uncompressed/partition_key_with_values_of_different_types/mc-1-big-Data.db                                                                                                                                             
```
most probably the problem is that these datatype singletons are thread local. Comparing the type is implemented by comparing the addresses, which are different. There should be a separate issue and pull request that fixes that.
Or should there? It's good that these singletons are thread local. We don't want shards accessing memory of other shards too often.

kbr-:
v3:
- extended the "PARTITION_KEY_WITH_VALUES_OF_DIFFERENT_TYPES_PATH" test to other compressors
- renamed some paths
- removed a thread_local

tests/sstable_3_x_test.cc
kbr-: Done

kbr-:
Here are some better plots for the 10M data, 64 KB chunk case. I ran 10 iterations for each compressor.
10M_64_spaces
10M_64_stall_nums
10M_64_stall_times
10M_64_times

kbr-:
v3:
- use zstd's cmake build option to build the project out of tree. Turns out there's no need to introduce any modifications to zstd's build system.
- move to new zstd version (1.4.2)

The change is now ready.

configure.py
```diff
@@ -1117,6 +1119,25 @@ def query_seastar_flags(seastar_pc_file, link_static_cxx=False):
...
+        '-DCMAKE_BUILD_TYPE={}'.format(MODE_TO_CMAKE_BUILD_TYPE[mode]),
+        '-DCMAKE_C_COMPILER={}'.format(args.cc),
+        '-DCMAKE_CXX_COMPILER={}'.format(args.cxx),
+        '-DCMAKE_C_FLAGS={}'.format(zstd_cflags),
+        '-DZSTD_BUILD_PROGRAMS=OFF'
+    ]
+
+    zstd_cmd = ['cmake', '-G', 'Ninja', os.path.relpath('zstd/build/cmake', zstd_build_dir)] + zstd_cmake_args
+
+    print(zstd_cmd)
```
denesb: Debug leftover?

tests/sstable_3_x_test.cc
denesb: The real solution would be to move the schema building into a function, and have each test build it, instead of using a shared instance. This is perpendicular to this issue, so you can leave it as-is, but let's not make the problem worse.

configure.py
kbr-: Actually, no: I followed Seastar's behavior, which prints the executed cmake command (see configure.py in Seastar repo, end of configure_mode function).

denesb: Fair enough. Just wanted to check. :)

kbr-:
> Some changes in the change-log I can't find in the code itself. After a few update cycles pull request become completely unreadable to me, I can't tell anymore which code belongs to which version, or how to find a specific version. Even the order of commits is a complete mess.

I agree :( That's why moving to Gerrit, which is a great code collaboration tool, would be nice. For example, in Gerrit you can compare between any two patch versions you like.

kbr-:
By the way, when I was doing the changes in configure.py an idea occurred to me: we could easily make the compilation of zstd (and other parts of Scylla) be optional.
So, for example, if you wanted zstd support in Scylla you would say e.g.
`./configure.py --mode=dev --enable-zstd`
and if the user doesn't provide "--enable-zstd", no zstd-related files would compile -- faster compilation and linking!
This would require moving the zstd-related tests to a separate file only. Zstd.cc is already a separate compilation unit.
With some minor refactoring this could be done with other compressors too (snappy, deflate). And we could try to find other features of Scylla that could be made optional.
What do you think?

penberg:
@kbr- Please fix the target branch of this pull request to `next`.

zstd.cc
avikivity: The problem with separate compilation units is that each compiles all its headers. This is less of a problem with compressors because they don't include so much (and on the other hand, they also don't benefit so buch because their dependencies rarely change).

avikivity:
> @avikivity
> 
> 1. Copying tables from https://github.com/apache/cassandra/tree/cassandra-3.9 with plugin https://github.com/MatejTymes/cassandra-zstd to Scylla: works
> 2. Copying tables from Scylla to cassandra-3.9 with plugin as above: works
> 3. Copying tables from Scylla to current https://github.com/apache/cassandra/tree/trunk (cassandra-4.0 dev branch, supports Zstd natively): works
> 4. Copying from current Cassandra 4.0 to Scylla: doesn't work since new Cassandra uses a new sstable format ('na') which we don't support.

Please add some Cassandra-generated sstables to the unit tests.

avikivity:
> By the way, when I was doing the changes in configure.py an idea occurred to me: we could easily make the compilation of zstd (and other parts of Scylla) be optional.
> So, for example, if you wanted zstd support in Scylla you would say e.g.
> `./configure.py --mode=dev --enable-zstd`
> and if the user doesn't provide "--enable-zstd", no zstd-related files would compile -- faster compilation and linking!
> This would require moving the zstd-related tests to a separate file only. Zstd.cc is already a separate compilation unit.
> With some minor refactoring this could be done with other compressors too (snappy, deflate). And we could try to find other features of Scylla that could be made optional.
> What do you think?

We already do this for dpdk. But I think here the benefit/cost ratio is low, since zstd would be compiled once and then never changed, and the compressor itself is not a big deal.

sstables/compress.cc
```diff
@@ -264,7 +264,7 @@ local_compression::local_compression(const compression& c)
...
             }
```
avikivity: What do you mean that no compressors implement it? I think all compressors respect chunk_length_in_kb, no?

.gitmodules
```diff
@@ -12,3 +12,6 @@
 [submodule "libdeflate"]
 	path = libdeflate
 	url = ../libdeflate
+[submodule "zstd"]
+	path = zstd
+	url = git@github.com:facebook/zstd.git
```
avikivity: Can't we use the distribution-packaged zstd? I can add it to the frozen toolchain.

tests/sstable_3_x_test.cc
avikivity: Those global thread_locals (the problem isn't that they're thread_local, but global) are not too bad in leaf files like unit tests. They cause harm when they're in infrastructure, where you can't use one thing without the other because everything depends indirectly on every other thing in a tangled mess.

.gitmodules
avikivity: Ah, you explain why we can't. Please point the submodule at ../zstd, this way we are insulated from changes in facebook's repository. I'll fork it into scylladb/zstd.

avikivity: Forked.

configure.py
```diff
@@ -1117,6 +1119,25 @@ def query_seastar_flags(seastar_pc_file, link_static_cxx=False):
     modes[mode]['seastar_cflags'] = seastar_cflags
     modes[mode]['seastar_libs'] = seastar_libs
 
+MODE_TO_CMAKE_BUILD_TYPE = {'release' : 'RelWithDebInfo', 'debug' : 'Debug', 'dev' : 'Dev', 'sanitize' : 'Sanitize' }
+
+def configure_zstd(build_dir, mode):
+    zstd_build_dir = os.path.join(build_dir, mode, 'zstd')
```
avikivity: Please add a comment explaining that we use features not available in the dynamically linked version of zstd (and which features, too) so people don't try to optimize this away.

configure.py
```diff
@@ -1392,6 +1417,9 @@ def query_seastar_flags(seastar_pc_file, link_static_cxx=False):
         f.write('rule libdeflate.{mode}\n'.format(**locals()))
         f.write('    command = make -C libdeflate BUILD_DIR=../build/{mode}/libdeflate/ CFLAGS="{libdeflate_cflags}" CC={args.cc} ../build/{mode}/libdeflate//libdeflate.a\n'.format(**locals()))
         f.write('build build/{mode}/libdeflate/libdeflate.a: libdeflate.{mode}\n'.format(**locals()))
+        f.write('build build/{mode}/zstd/lib/libzstd.a: ninja\n'.format(**locals()))
+        f.write('   subdir = build/{mode}/zstd\n'.format(**locals()))
+        f.write('   target = libzstd.a\n'.format(**locals()))
```
avikivity: We should rename seastar_pool to submodule_pool and build all the submodules in that pool. Out of scope for this patchset, but would be nice to follow up.

avikivity:
@glommer some workloads may want to trade off latency for better compression. For the latency senstive workloads we can recommend lz4 or no compression.

@kbr- the traditional response tp someone going to the trouble of generating such good data is to punish them with the task of writing a blog post

avikivity:
@kbr- I had some ultra-minor comments but otherwise this is good to merge

sstables/compress.cc
kbr-: chunk_length_in_kb is indeed always used in the higher layer - the code which uses the chosen compressor to compress chunks of sstable data. But none of the compressors in their code cared about chunk_length_in_kb until now. The other compressors don't assume anything about the length of the input they get each time `compress` is called. In Zstd on the other hand, we can optimize the amount of memory allocated for compression given an upper bound on the input length given each time the compressor is used - which is chunk_len_in_kb in our case.

kbr-:
> Please add some Cassandra-generated sstables to the unit tests.

All of the tests added in the third commit are Cassandra-generated.

configure.py
kbr-: I added a commit in this PR, I think it won't hurt.

configure.py
kbr-: Done

.gitmodules
kbr-: Done

kbr-:
v4:
- changed the submodule path to ../zstd
- added a comment in configure.py about static linking
- new commit: rename seastar_pool to submodule_pool, compile all submodules there

avikivity:
Dequeued - #include <zstd.h> resolves to /usr/include instead of the submodule (I found out because I don't have zstd.h installed).

kbr-:
Good catch... will fix and make another PR.

PR #4743, opened 2019-07-21
systemd: put scylla processes in systemd slices.

glommer:
It is well known that seastar applications, like Scylla, do not play
well with external processes: CPU usage from external processes may
confuse the I/O and CPU schedulers and create stalls.

We have also recently seen that memory usage from other application's
anonymous and page cache memory can bring the system to OOM.

Linux has a very good infrastructure for resource control contributed by
amazingly bright engineers in the form of cgroup controllers. This
infrastructure is exposed by SystemD in the form of slices: a
hierarchical structure to which controllers can be attached.

In true systemd way, the hierarchy is implicit in the filenames of the
slice files. a "-" symbol defines the hierarchy, so the files that this
patch presents, scylla-server and scylla-helper, essentially create a
"scylla" cgroup at the top level with "server" and "helper" children.

Later we mark the Services needed to run scylla as belonging to one
or the other through the Slice= directive.

Scylla DBAs can benefit from this setup by using the systemd-run
utility to fire ad-hoc commands.

Let's say for example that someone wants to hypothetically run a backup
and transfer files to an external object store like S3, making sure that
the amount of page cache used won't create swap pressure leading to
database timeouts.

One can then run something like:

   sudo systemd-run --uid=`id -u scylla` --gid=`id -g scylla` -t --slice=scylla-helper.slice /path/to/my/magical_backup_tool

(or even better, the backup tool can itself be a systemd timer)

Scylla doesn't use pull-requests, please send a patch to the [mailing list](mailto:scylladb-dev@googlegroups.com) instead.
See our [contributing guidelines](../CONTRIBUTING.md) and our [Scylla development guidelines](../HACKING.md) for more information.

If you have any questions please don't hesitate to send a mail to the [dev list](mailto:scylladb-dev@googlegroups.com).


tzach:
@mmatczuk @ajankovic @rjeczalik FYI

dist/common/systemd/scylla-helper.slice
```diff
@@ -0,0 +1,35 @@
...
+CPUShares=10
+BlockIOWeight=10
+
+# Run for at most 25% of the time per each 1ms increment per CPU. That does cap
+# the maximum CPU usage at 25% which is less ideal, but that is a requirement to
+# use CPUQuotaPeriodSec, which we want to make sure that the process does not consume
+# an unreasonable amount of the task quota.
+CPUQuota=25%
+CPUQuotaPeriodSec=1ms
+
```
avikivity: This needs significant testing. These low numbers can increase scheduling overhead. Also, a quota means that idle time cannot be used, and it applies to non-Scylla cores too.

Perhaps using priority is better. We can both increase Scylla's priority and reduce helper priority. Low priority can still take unused bandwidth on both Scylla and non-Scylla cores, and likely has fewer interactions with the scheduler, since it doesn't need to set additional timers.

dist/common/systemd/scylla-server.slice
```diff
@@ -0,0 +1,22 @@
...
+# Systemd deprecated settings BlockIOWeight and CPUShares. But they are still the ones used in RHEL7
+# Newer SystemD wants IOWeight and CPUWeight instead. Luckily both newer and older SystemD seem to
+# ignore the unwanted option so safest to get both. Using just the old versions would work too but
+# seems less future proof. Using just the new versions does not work at all for RHEL7/
+BlockIOWeight=1000
+IOWeight=1000
+# Should be zero, but work around https://github.com/systemd/systemd/issues/8363 for older systemd
+MemorySwapMax=1
+CPUShares=1000
+CPUweight=1000
```
avikivity: s/w/W/

dist/common/systemd/scylla-helper.slice
```diff
@@ -0,0 +1,35 @@
...
+CPUWeight=10
+IOWeight=10
+
+# MemoryHigh is the throttle point
+MemoryHigh=4%
+# MemoryMax is the OOM point. As Scylla reserves 7% by default, the other two percent goes to
+# the kernel and other non contained processes
+MemoryMax=5%
+# Systemd deprecated settings BlockIOWeight, MemoryLimit and CPUShares. But they are still the ones used in RHEL7
```
avikivity: Seastar has the ability to use an absolute reserve:

```
struct configuration {
    optional<size_t> total_memory;
    optional<size_t> reserve_memory;  // if total_memory not specified
    optional<size_t> cpus;
    optional<cpuset> cpu_set;
    optional<unsigned> io_queues;
};
```

which unfortunately we don't use. But the Seastar's default is max(1.5GB, 7%), so the percentage calculation can be too limiting in small instances.

Perhaps the postinstall script can calculate and write an override to /etc/systemd.

dist/common/systemd/scylla-helper.slice
glommer: The whole thing needs significant testing, and I was wondering if we can run a preview in QA so we can get more coverage than what I can locally.

I agree with what you say about the disadvantage of the quota, I said as much in the commit message. I ended up deciding to use it just so we can try to guarantee the task quota is not violated too.

However to make this move I'd propose we remove this altogether in favor of the Weight system only, and as we gain more experience with it if we still feel the need to add Quotas we can revisit the decision.

dist/common/systemd/scylla-server.slice
glommer: Thanks

dist/common/systemd/scylla-helper.slice
glommer: The problem with postinstall is that it is still wrong when people change instance sizes.
It's a weird operation if you think in terms of real hardware, but in the cloud it happens all the time...

But perhaps you are right and the solution for that problem is a different one: so many things change when you move between instance sizes, that the best solution is probably to have a script to re-configure.

If installing the rpm again would re-do the post install part, that's probably a good solution.

glommer: I really wish systemd could read this from an environment variable, then we could put the reservation there. It seems like they only support environment variables in key specific directives, like `ExecStart`, but not in the slice configuration. Then I guess an override it is

glommer:
@avikivity force pushed - It was still pointing to master, since the original PR was before the next discussion

glommer:
@avikivity Please notice that because of the next / master confusion mentioned above this shows as two commits, but this PR is really a single commit. Would it be easier for me to close it and open a new one (losing history) ?

avikivity:
If it's one commit it should show as one commit. What branch did you use as a base for this?

glommer:
It shows as two because I ended up using `next` as a base - seeing new discussions about it - but the PR was already opened against master and I don't know how to change that.

I will force push again now using master as a base.

avikivity:
I'm still waiting for testing results.

glommer:
@avikivity I have added extensive benchmarks in the commit message description.

Did github hide them , maybe? (I can just resend this if it's getting too weird).

Is there any more tests not in my commit message that you'd like to see ?

avikivity:
I didn't know that you added them. I don't re-read commit messages without some trigger.

dist/common/systemd/scylla-helper.slice
```diff
@@ -0,0 +1,27 @@
+[Unit]
+Description=Slice used to run companion programs to Scylla. Memory, CPU and IO restricted
+DefaultDependencies=no
+Before=slices.target
```
avikivity: Not recommended according to the documentation of slices.target.

dist/common/systemd/scylla-helper.slice
```diff
@@ -0,0 +1,27 @@
...
+Wants=-.slice
+After=-.slice
```
avikivity: Are these two needed?

dist/redhat/scylla.spec.mustache
```diff
@@ -93,6 +93,21 @@ AmbientCapabilities=CAP_SYS_NICE
...
+MEMTOTAL_BYTES=$(cat /proc/meminfo | grep MemTotal | awk '{print $2 * 1024}')
+if [ $MEMTOTAL_BYTES -lt 23008753371 ]; then
+    mkdir -p /etc/systemd/system/scylla-helper.slice.d/
+    cat << EOS > /etc/systemd/system/scylla-helper.slice.d/memory.conf
+[Slice]
+MemoryHigh=1200M
+MemoryMax=1400M
+MemoryLimit=1400M
+EOS
+fi
```
avikivity: Let's avoid this duplication.

avikivity:
Looks good, I had some cosmetic comments.

Did you verify that this doesn't trigger the quota system? If it does we'll need to benchmark effects on latency too.

dist/common/systemd/scylla-helper.slice
glommer: Sorry, what exactly is not recommended ? (default dependencies or before)

Can you point me to the docs ?

dist/common/systemd/scylla-helper.slice
glommer: It is my understanding that it is, to guarantee that this is initialized after the root slice.

The scylla slice is not top level, so it has to fall within the root slice.

dist/redhat/scylla.spec.mustache
glommer: through a script ? Or any other idea?

glommer:
Avi, which quota system?

I got rid of the quotas in this patch.

dist/common/systemd/scylla-helper.slice
avikivity: It's odd that it would be needed, since it's kind of implied.

dist/redhat/scylla.spec.mustache
avikivity: Script. In fact you need a script, to call it from install.sh. We should be thinking in terms of install.sh, with distribution packages just a way to defer the parts of install.sh that cannot be called at packaging time.

avikivity:
I saw, I just want to ensure systemd isn't adding them as some kind of default when you add CPUWeight.

dist/common/systemd/scylla-helper.slice
glommer: Found it -> it is DefaultDependencies that is not recommended. I will remove.

dist/common/systemd/scylla-helper.slice
glommer: Man page states that `After` and `Requires` are implicitly added. 
`Wants` is not, but `Wants` is weaker than `Requires`, so I think we can safely delete this.

dist/redhat/scylla.spec.mustache
glommer: There was already code to handle capabilities that was not shared. My new version will include a patch that comes before this one and moves that code to a shared file first. Thanks

glommer:
> I saw, I just want to ensure systemd isn't adding them as some kind of default when you add CPUWeight.

The documentation claims that " Setting CPUQuotaPeriodSec= to an empty value resets it to the default." So I doubt that not specifying it would pose any harm.

glommer:
@avikivity forced pushed a new branch. Note that there are two commits now.

scylla_post_install.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if [ ! -d /run/systemd/system ]; then
+    exit 0
+fi
+
+# Install capabilities.conf when AmbientCapabilities supported
+. /etc/os-release
+
+# the command below will still work in systems without system (like docker) despite versioning
+SYSTEMD_VER=`( systemctl --version || echo 0 0) | head -n1 | awk '{print $2}'`
```
avikivity: s/without system/without systemctl/

What do you mean "despite versioning"?

Please use the more modern $(...) instead of backtick...backtick.

scylla_post_install.sh
```diff
@@ -0,0 +1,37 @@
...
+ISRHEL=`echo $ID_LIKE | grep -oi rhel`
+SYSTEMD_REL=0
```
avikivity: RHEL

dist/debian/debian/scylla-server.postinst
```diff
@@ -36,19 +36,6 @@ echo "api_ui_dir: /opt/scylladb/swagger-ui/dist/" >> /tmp/scylla.yaml
...
-    cat << EOS > /etc/systemd/system/scylla-server.service.d/capabilities.conf
-[Service]
-AmbientCapabilities=CAP_SYS_NICE
-EOS
-fi
-
-if [ -d /run/systemd/system ]; then
-    systemctl --system daemon-reload >/dev/null || true
-fi
+/opt/scylladb/scripts/scylla_post_install.sh
```
avikivity: You can also remove the sourcing of /etc/os-release.

scylla_post_install.sh
```diff
@@ -0,0 +1,37 @@
...
+            mkdir -p /etc/systemd/system/scylla-server.service.d/
+            cat << EOS > /etc/systemd/system/scylla-server.service.d/capabilities.conf
+[Service]
+AmbientCapabilities=CAP_SYS_NICE
+EOS
+        fi
+    fi
+
+    systemctl --system daemon-reload >/dev/null || true
+fi
```
avikivity: Need to call the new script from install.sh (but different pull request)

dist/common/systemd/scylla-server.slice
```diff
@@ -0,0 +1,19 @@
...
+# Newer SystemD wants IOWeight and CPUWeight instead. Luckily both newer and older SystemD seem to
+# ignore the unwanted option so safest to get both. Using just the old versions would work too but
+# seems less future proof. Using just the new versions does not work at all for RHEL7/
+BlockIOWeight=1000
+IOWeight=1000
+# Should be zero, but work around https://github.com/systemd/systemd/issues/8363 for older systemd
+MemorySwapMax=1
+CPUShares=1000
+CPUWeight=1000
+
```
avikivity: Please check (by inspection is sufficient) that systemd does not create the quota stuff by itself.

glommer: I have verified by looking at the cgroups hierarchy that the property is not set, if it is not mentioned by the slice:

```
$ cat cpu.cfs_quota_us 
-1
$ pwd
/sys/fs/cgroup/cpu,cpuacct/scylla.slice
cd scylla-server.slice/
$ cat cpu.cfs_quota_us 
-1
```

After setting `CPUQuota` to something (not part of the patch, just making sure):

```
$ cat cpu.cfs_quota_us 
1000
```

Therefore, as expected, systemd is not messing with it.

scylla_post_install.sh
glommer: I will postpone this since Takuya has code in flight to change install.sh for various modes.
Should be done after that.

glommer:
@avikivity force pushed.

scripts/create-relocatable-package.py
```diff
@@ -159,6 +159,7 @@ def ldd(executable):
 ar.add('seastar/scripts')
 ar.add('seastar/dpdk/usertools')
 ar.add('install.sh')
+ar.add('scylla_post_install.sh', arcname="dist/common/scripts/scylla_post_install.sh")
 ar.add('scripts/relocate_python_scripts.py', arcname='relocate_python_scripts.py')
```
avikivity: Why are you copying it twice? It's already copied as part of dist/.

scylla_post_install.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
```
avikivity: License/copyright blurb.

scylla_post_install.sh
```diff
@@ -35,5 +35,19 @@ EOS
...
+    if [ $MEMTOTAL_BYTES -lt 23008753371 ]; then
+        mkdir -p /etc/systemd/system/scylla-helper.slice.d/
+        cat << EOS > /etc/systemd/system/scylla-helper.slice.d/memory.conf
+[Slice]
+MemoryHigh=1200M
+MemoryMax=1400M
+MemoryLimit=1400M
+EOS
+    fi
+
```
avikivity: Please add this as %ghost file in the rpm (and debian equivalent), so it is removed after uninstall. We already have a few like that.

scripts/create-relocatable-package.py
glommer: It doesn't live inside dist/, however, but at the top level.

I can move it inside dist/ if you prefer, but my reasoning is that this file would better be laid in the source tree together with install.sh.

However once it is installed into the system, my reasoning went, it would be better off at /usr/lib/scylla together with the other scripts

scylla_post_install.sh
glommer: Done

scylla_post_install.sh
glommer: Done

scripts/create-relocatable-package.py
glommer: For now I will just add a comment for this explaining why this is the case so I can force push the branch.

If you really want this moved I will move tomorrow.

glommer:
@avikivity post-refresh re-ping

scylla_post_install.sh
avikivity: I don't see it.

scylla_post_install.sh
avikivity: I don't see it.

glommer:
On Mon, Aug 26, 2019 at 11:55 AM Avi Kivity <notifications@github.com>
wrote:

>
> In scylla_post_install.sh
> <https://github.com/scylladb/scylla/pull/4743#discussion_r317670852>:
>
> > @@ -35,5 +35,19 @@ EOS
>          fi
>      fi
>
> +    # For systems with not a lot of memory, override default reservations for the slices
> +    # seastar has a minimum reservation of 1.5GB that kicks in, and 21GB * 0.07 = 1.5GB.
> +    # So for anything smaller than that we will not use percentages in the helper slice
> +    MEMTOTAL_BYTES=$(cat /proc/meminfo | grep MemTotal | awk '{print $2 * 1024}')
> +    if [ $MEMTOTAL_BYTES -lt 23008753371 ]; then
> +        mkdir -p /etc/systemd/system/scylla-helper.slice.d/
> +        cat << EOS > /etc/systemd/system/scylla-helper.slice.d/memory.conf
> +[Slice]
> +MemoryHigh=1200M
> +MemoryMax=1400M
> +MemoryLimit=1400M
> +EOS
> +    fi
> +
>
> I don't see it.
>

What's "it" ?
I got lost.

avikivity:
Do not read github emails, that is not "the way". They are context-free. 
Click the link as github orders. All hail github.

PR #4753, opened 2019-07-23
Catch unclosed partition during sstable write

tgrabiec:
Not emitting partition_end for a partition is incorrect. SStable
writer assumes that it is emitted. If it's not, the sstable will not
be written correctly. The partition index entry for the last partition
will be left partially written, which will result in errors during
reads. Also, statistics and sstable key ranges will not include the
last partition.

It's better to catch this problem at the time of writing, and not
generate bad sstables.

Another way of handling this would be to implicitly generate a
partition_end, but I don't think that we should do this. We cannot
trust the mutation stream when invariants are violated, we don't know
if this was really the last partition which was supposed to be
written. So it's safer to fail the write.

Enabled for both mc and la/ka.

Passing --abort-on-internal-error on the command line will switch to
aborting instead of throwing an exception.

The reason we don't abort by default is that it may bring the whole
cluster down and cause unavailability, while it may not be necessary
to do so. It's safer to fail just the affected operation,
e.g. repair. However, failing the operation with an exception leaves
little information for debugging the root cause. So the idea is that the
user would enable aborts on only one of the nodes in the cluster to
get a core dump and not bring the whole cluster down.

### DIFF
--- a/db/config.cc
+++ b/db/config.cc
@@ -673,6 +673,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
         " It is not enough to have ever since upgraded to newer versions of Cassandra. If you EVER used a version earlier than 2.1 in the cluster where these SSTables come from, DO NOT TURN ON THIS OPTION! You will corrupt your data. You have been warned.")
     , enable_shard_aware_drivers(this, "enable_shard_aware_drivers", value_status::Used, true, "Enable native transport drivers to use connection-per-shard for better performance")
     , enable_ipv6_dns_lookup(this, "enable_ipv6_lookup", value_status::Used, false, "Use IPv6 address resolution")
+    , abort_on_internal_error(this, "abort_on_internal_error", liveness::LiveUpdate, value_status::Used, false, "Abort the server instead of throwing exception when internal invariants are violated")
 
     , default_log_level(this, "default_log_level", value_status::Used)
     , logger_log_level(this, "logger_log_level", value_status::Used)
--- a/main.cc
+++ b/main.cc
@@ -629,6 +629,12 @@ int main(int ac, char** av) {
             if (opts.count("developer-mode")) {
                 smp::invoke_on_all([] { engine().set_strict_dma(false); }).get();
             }
+
+            auto abort_on_internal_error_observer = cfg->abort_on_internal_error.observe([] (auto val) {
+                set_abort_on_internal_error(val);
+            });
+            set_abort_on_internal_error(cfg->abort_on_internal_error());
+
             supervisor::notify("creating tracing");
             tracing::backend_registry tracing_backend_registry;
             tracing::register_tracing_keyspace_backend(tracing_backend_registry);
--- a/sstables/mc/writer.cc
+++ b/sstables/mc/writer.cc
@@ -29,6 +29,7 @@
 #include "sstables/mc/types.hh"
 #include "db/config.hh"
 #include "atomic_cell.hh"
+#include "utils/exceptions.hh"
 
 #include <functional>
 #include <boost/iterator/iterator_facade.hpp>
@@ -1382,10 +1383,15 @@ stop_iteration writer::consume_end_of_partition() {
         _first_key = *_partition_key;
     }
     _last_key = std::move(*_partition_key);
+    _partition_key = std::nullopt;
     return get_data_offset() < _cfg.max_sstable_size ? stop_iteration::no : stop_iteration::yes;
 }
 
 void writer::consume_end_of_stream() {
+    if (_partition_key) {
+        on_internal_error(sstlog, "Mutation stream ends with unclosed partition during write");
+    }
+
     _cfg.monitor->on_data_write_completed();
 
     seal_summary(_sst._components->summary, std::move(_first_key), std::move(_last_key), _index_sampling_state);
--- a/utils/exceptions.cc
+++ b/utils/exceptions.cc
@@ -18,8 +18,13 @@
  * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <seastar/core/print.hh>
+#include <seastar/util/log.hh>
+#include <seastar/util/backtrace.hh>
+
 #include <exception>
 #include <system_error>
+#include <atomic>
 #include "exceptions.hh"
 
 #include <iostream>
@@ -62,4 +67,19 @@ bool should_stop_on_system_error(const std::system_error& e) {
         }
     }
     return true;
+}
+
+std::atomic<bool> abort_on_internal_error{false};
+
+void set_abort_on_internal_error(bool do_abort) {
+    abort_on_internal_error.store(do_abort);
+}
+
+void on_internal_error(seastar::logger& logger, const seastar::sstring& msg) {
+    if (abort_on_internal_error.load()) {
+        logger.error("{}, at: {}", msg.c_str(), seastar::current_backtrace());
+        abort();
+    } else {
+        seastar::throw_with_backtrace<std::runtime_error>(msg.c_str());
+    }
 }
--- a/utils/exceptions.hh
+++ b/utils/exceptions.hh
@@ -21,8 +21,12 @@
 
 #pragma once
 
+#include <seastar/core/sstring.hh>
+
 #include <functional>
 #include <system_error>
+
+namespace seastar { class logger; }
 
 typedef std::function<bool (const std::system_error &)> system_error_lambda_t;
 
@@ -45,3 +49,10 @@ public:
 
     const std::error_code& code() const { return _code; }
 };
+
+// Controls whether on_internal_error() aborts or throws.
+void set_abort_on_internal_error(bool do_abort);
+
+// Handles reporting of violation of internal invariants.
+// Callers can assume that it does not return. May throw.
+[[noreturn]] void on_internal_error(seastar::logger&, const seastar::sstring& reason);
--- a/db/config.hh
+++ b/db/config.hh
@@ -279,6 +279,7 @@ public:
     named_value<bool> enable_dangerous_direct_import_of_cassandra_counters;
     named_value<bool> enable_shard_aware_drivers;
     named_value<bool> enable_ipv6_dns_lookup;
+    named_value<bool> abort_on_internal_error;
 
     seastar::logging_settings logging_settings(const boost::program_options::variables_map&) const;
 
--- a/sstables/sstables.cc
+++ b/sstables/sstables.cc
@@ -2112,11 +2112,15 @@ stop_iteration components_writer::consume_end_of_partition() {
         _first_key = *_partition_key;
     }
     _last_key = std::move(*_partition_key);
+    _partition_key = std::nullopt;
 
     return get_offset() < _max_sstable_size ? stop_iteration::no : stop_iteration::yes;
 }
 
 void components_writer::consume_end_of_stream() {
+    if (_partition_key) {
+        on_internal_error(sstlog, "Mutation stream ends with unclosed partition during write");
+    }
     // what if there is only one partition? what if it is empty?
     seal_summary(_sst._components->summary, std::move(_first_key), std::move(_last_key), _index_sampling_state);
 
--- a/tests/cql_test_env.cc
+++ b/tests/cql_test_env.cc
@@ -360,6 +360,7 @@ public:
                 create_directories((cfg->view_hints_directory() + "/" + std::to_string(i)).c_str());
             }
 
+            set_abort_on_internal_error(true);
             const gms::inet_address listen("127.0.0.1");
             auto& ms = netw::get_messaging_service();
             // don't start listening so tests can be run in parallel


utils/exceptions.hh
```diff
@@ -45,3 +48,10 @@ public:
 
     const std::error_code& code() const { return _code; }
 };
+
+// Controls whether on_internal_error() aborts or throws.
+void set_abort_on_internal_error(bool do_abort);
+
+// Handles reporting of violation of internal invariants.
+// Callers can assume that it does not return. May throw.
+void on_internal_error(seastar::logger&, const seastar::sstring& reason);
```
denesb: `[[noreturn]]`?

sstables/mc/writer.cc
```diff
@@ -1382,10 +1383,15 @@ stop_iteration writer::consume_end_of_partition() {
         _first_key = *_partition_key;
     }
     _last_key = std::move(*_partition_key);
+    _partition_key = std::nullopt;
```
denesb: Can we also (cheaply) validate other invariants, like partition start emitted before clustering rows are, static row is emitted right after partition start, etc? Not proposing doing these in this PR.

tgrabiec:
> LGTM. Do we know why the stream ends without returning partition_end?

We don't.

utils/exceptions.hh
tgrabiec: Ok

bhalevy:
@avikivity ping review/merge

utils/exceptions.hh
bhalevy: We actually only need to pass a `const char*` rather than sstring& this will reduce header dependencies a bit.

utils/exceptions.cc
```diff
@@ -63,3 +66,18 @@ bool should_stop_on_system_error(const std::system_error& e) {
     }
     return true;
 }
+
+std::atomic<bool> abort_on_internal_error{false};
```
bhalevy: does that have to be atomic?
`volatile bool` should work too, no?

haaawk: I think volatile would work here but would be super fragile. If two threads access volatile bool at the same time and at least one is a write then this is undefined behaviour. This won't happen in the current state of the code but would be a very hard to find bug in the future if the startup code gets rearranged. For example, when we start reloading configuration without restarts.

haaawk: BTW - I checked the following program:
#if 0
#include <atomic>
std::atomic<bool> stop_(false);
#else
volatile bool stop_ = false;
#endif

int main(int argc, char **argv)
{
    long n = 1000000000;
    while (!stop_) {
        if (--n < 0)
            stop_ = true;
    }

    return 0;
}
result for atomic:
real	0m0.322s
user	0m0.319s
sys	0m0.003s
result for volatile:
real	0m0.662s
user	0m0.660s
sys	0m0.001s

but it can be wrong because there's only one thread so -O3 may optimize out synchronization.

utils/exceptions.hh
tgrabiec: Ok.

utils/exceptions.cc
tgrabiec: @bhalevy AFAIK `volatile` doesn't provide guarantees for atomicity and multi-threaded access. What's wrong with `std::atomic`?

haaawk: volatile will work here because there's no race on this variable at the moment. We write to it before we start the concurrent reads and concurrent reads are ok because they don't change the value. Reads will see the write done at the beginning because volatile guarantees that they always go to the memory for the value of this variable.

I still think that using volatile instead of atomic here is dangerous in the long run and I'm not sure whether there's any performance benefit of doing it at all.

tgrabiec: Performance of internal error handling doesn't matter.
If it was, we could use `std::atomic` with relaxed ordering, which adds no overhead.

slivne:
@avikivity this is the patch I want to get in and also backport to 3.1 - with additional checks

utils/exceptions.cc
bhalevy: @tgrabiec I just wondered if there's a requirement to use std::atomic that I missed.  I'm ok with using it.

db/config.cc
```diff
@@ -673,6 +673,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
         " It is not enough to have ever since upgraded to newer versions of Cassandra. If you EVER used a version earlier than 2.1 in the cluster where these SSTables come from, DO NOT TURN ON THIS OPTION! You will corrupt your data. You have been warned.")
     , enable_shard_aware_drivers(this, "enable_shard_aware_drivers", value_status::Used, true, "Enable native transport drivers to use connection-per-shard for better performance")
     , enable_ipv6_dns_lookup(this, "enable_ipv6_lookup", value_status::Used, false, "Use IPv6 address resolution")
+    , abort_on_internal_error(this, "abort_on_internal_error", value_status::Used, false, "Abort the server instead of throwing exception when internal invariants are violated")
```
avikivity: Please make this LiveUpdate. Reasoning: we see errors, user is prepared for a core dump, enables setting, captures core dump, disables setting, ships core dump for analisys

main.cc
```diff
@@ -629,6 +629,7 @@ int main(int ac, char** av) {
             if (opts.count("developer-mode")) {
                 smp::invoke_on_all([] { engine().set_strict_dma(false); }).get();
             }
+            set_abort_on_internal_error(cfg->abort_on_internal_error());
             supervisor::notify("creating tracing");
```
avikivity: And here you'd have to set up a listener to update the value.

utils/exceptions.cc
avikivity: Anything that is used cross-thread without extra synchronization should be atomic<>.

utils/exceptions.hh
```diff
@@ -21,6 +21,9 @@
 
 #pragma once
 
+#include <seastar/core/sstring.hh>
+#include <seastar/util/log.hh>
+
```
avikivity: Please use a forward declaration for seastar::logger to avoid compile-time bloat.

utils/exceptions.cc
```diff
@@ -63,3 +66,18 @@ bool should_stop_on_system_error(const std::system_error& e) {
...
+void set_abort_on_internal_error(bool do_abort) {
+    abort_on_internal_error.store(do_abort);
+}
+
+void on_internal_error(seastar::logger& logger, const seastar::sstring& msg) {
+    if (abort_on_internal_error.load()) {
+        logger.error(msg.c_str());
+        abort();
+    } else {
+        throw std::runtime_error(msg.c_str());
```
bhalevy: I think it would be helpful to print the backtrace here to help identify the code path.

slivne:
If an exception is thrown:
- Is the sstable file still created ? 
- Is the operation that triggered the write aborted (repair or other ?)

What will happen in case we are attempting to flush a memtable and failing ? will that not cause more issues - e.g. running out of memory and causing that node/shard to be slower - which can cause the whole cluster to be unresponsive. 

Its clear that we should not write corrupted sstables - I am questioning the decision to not enable abort (at least by default) on all cases of detecting something incorrect in the write path. If the user wants to disable it he can - but by default in my view corruption on the write path should abort the node.

tgrabiec:
On Mon, Jul 29, 2019 at 1:01 PM Shlomi Livne <notifications@github.com>
wrote:

> If an exception is thrown:
>
>    - Is the sstable file still created ?
>
> It will be in an unsealed state. The one who initiated the write marks it
for removal when failure propagates to it. It would be also removed on boot.

>
>    - Is the operation that triggered the write aborted (repair or other ?)
>
>
Yes.


>
>
> What will happen in case we are attempting to flush a memtable and failing
> ? will that not cause more issues - e.g. running out of memory and causing
> that node/shard to be slower - which can cause the whole cluster to be
> unresponsive.
>
Failed memtable flushes are currently retired after 10 seconds. If the
memtable cannot be flushed consistently, eventually writes will block on
dirty.

Reads will still be able to proceed as usual, so the node will not be
totally unavailable.

I think the decision of how we handle the failure of memtable flush is
orthogonal to this issue. That failure could happen for a number of reasons
even before this change. The sstable layer should not decide about that.

For repair, for example, throwing an exception and failing just the repair
seems the safest option.

> Its clear that we should not write corrupted sstables - I am questioning
> the decision to not enable abort (at least by default) on all cases of
> detecting something incorrect in the write path. If the user wants to
> disable it he can - but by default in my view corruption on the write path
> should abort the node.
>
By "on the write path", do you mean memtable flush?

>

tgrabiec:
On Tue, Jul 30, 2019 at 10:04 AM Tomasz Grabiec <tgrabiec@gmail.com> wrote:

>
>
> On Mon, Jul 29, 2019 at 1:01 PM Shlomi Livne <notifications@github.com>
> wrote:
>
>> If an exception is thrown:
>>
>>    - Is the sstable file still created ?
>>
>> It will be in an unsealed state. The one who initiated the write marks it
> for removal when failure propagates to it. It would be also removed on boot.
>
>>
>>    - Is the operation that triggered the write aborted (repair or other
>>    ?)
>>
>>
> Yes.
>
>
>>
>>
>> What will happen in case we are attempting to flush a memtable and
>> failing ? will that not cause more issues - e.g. running out of memory and
>> causing that node/shard to be slower - which can cause the whole cluster to
>> be unresponsive.
>>
> Failed memtable flushes are currently retired after 10 seconds. If the
> memtable cannot be flushed consistently, eventually writes will block on
> dirty.
>

s/retired/retried/

slivne:
On Tue, Jul 30, 2019 at 11:04 AM Tomasz Grabiec <notifications@github.com>
wrote:

> On Mon, Jul 29, 2019 at 1:01 PM Shlomi Livne <notifications@github.com>
> wrote:
>
> > If an exception is thrown:
> >
> > - Is the sstable file still created ?
> >
> > It will be in an unsealed state. The one who initiated the write marks it
> for removal when failure propagates to it. It would be also removed on
> boot.
>
> >
> > - Is the operation that triggered the write aborted (repair or other ?)
> >
> >
> Yes.
>
>
> >
> >
> > What will happen in case we are attempting to flush a memtable and
> failing
> > ? will that not cause more issues - e.g. running out of memory and
> causing
> > that node/shard to be slower - which can cause the whole cluster to be
> > unresponsive.
> >
> Failed memtable flushes are currently retired after 10 seconds. If the
> memtable cannot be flushed consistently, eventually writes will block on
> dirty.
>
> Reads will still be able to proceed as usual, so the node will not be
> totally unavailable.
>
> I think the decision of how we handle the failure of memtable flush is
> orthogonal to this issue. That failure could happen for a number of reasons
> even before this change. The sstable layer should not decide about that.
>
> For repair, for example, throwing an exception and failing just the repair
> seems the safest option.
>
> > Its clear that we should not write corrupted sstables - I am questioning
> > the decision to not enable abort (at least by default) on all cases of
> > detecting something incorrect in the write path. If the user wants to
> > disable it he can - but by default in my view corruption on the write
> path
> > should abort the node.
> >
> By "on the write path", do you mean memtable flush?
>

Yes

utils/exceptions.cc
tgrabiec: Yes.

sstables/mc/writer.cc
tgrabiec: Yes, will do as a follow up

db/config.cc
tgrabiec: Ok.

tgrabiec:
@haaawk @avikivity @bhalevy I pushed a new version with all comments addressed.

avikivity:
Looks good to me, will wait for a while to let others review too.

haaawk:
LGTM

avikivity:
Dequeued, fails compile in main.cc:

```
14:20:32                           ^~~~~~
14:20:32  main.cc:633:82: error: 'struct utils::config_file::named_value<bool>' has no member named 'observe'
14:20:32               auto abort_on_internal_error_observer = cfg->abort_on_internal_error.observe([] (auto val) {
14:20:32                                                                                    ^~~~~~~
```


btw generic lambdas should be avoided.

tgrabiec:
@avikivity The following works:

```diff
diff --git a/main.cc b/main.cc
index d8f4f2e..1d2b873 100644
--- a/main.cc
+++ b/main.cc
@@ -630,7 +630,7 @@ int main(int ac, char** av) {
                 smp::invoke_on_all([] { engine().set_strict_dma(false); }).get();
             }
 
-            auto abort_on_internal_error_observer = cfg->abort_on_internal_error.observe([] (auto val) {
+            auto abort_on_internal_error_observer = static_cast<utils::updateable_value<bool>>(cfg->abort_on_internal_error).observe([] (bool val) {
                 set_abort_on_internal_error(val);
             });
             set_abort_on_internal_error(cfg->abort_on_internal_error());
```

But it's a bit ugly. Is there a better way? Should we add `observe()` to `named_value`?

avikivity:
Agree about adding named_value::observe().

tgrabiec:
I opened https://github.com/scylladb/scylla/pull/4794 with the fix:
  - added observe() to named_value
  - avoided auto capture in the lambda

PR #4754, opened 2019-07-23
Add option to create a XUnit output file for non-boost tests in test.py.

### LINKED ISSUE #4680: test.py should provide an option to create xml file in xunit format
To help track the unit tests results we should integrate the unit test results with Jenkins.
We would benefit greatly by tracking unit test failures in the next-gating jobs to quickly see which unit tests failed and analyze their history to find out when they became flaky, for example.

It seems like the easiest way to achieve that is to have an option in `test.py` to create a xml file in the [xunit format](https://nose.readthedocs.io/en/latest/plugins/xunit.html), like we do for dtests and collect it to the test artifacts in Jenkins.

kbr-:
If the user specifies an output file name using "--xunit=<filename>",
test.py will write the test results of non-boost tests to the file in the XUnit XML format.
Every boost test creates its own results file already.
Resolves #4680.

### DIFF
--- a/test.py
+++ b/test.py
@@ -27,6 +27,7 @@ import subprocess
 import concurrent.futures
 import io
 import multiprocessing
+import xml.etree.ElementTree as ET
 
 boost_tests = [
     'bytes_ostream_test',
@@ -211,6 +212,8 @@ if __name__ == "__main__":
                         help='Verbose reporting')
     parser.add_argument('--jobs', '-j', action="store", default=default_num_jobs, type=int,
                         help="Number of jobs to use for running the tests")
+    parser.add_argument('--xunit', action="store",
+                        help="Name of a file to write results of non-boost tests to in xunit format")
     args = parser.parse_args()
 
     print_progress = print_status_verbose if args.verbose else print_progress_succint
@@ -299,7 +302,7 @@ if __name__ == "__main__":
             def report_subcause(e):
                 print('  with error {e}\n'.format(e=e), file=file)
             report_error(e, e, report_subcause=report_subcause)
-        return (path, boost_args + exec_args, success, file.getvalue())
+        return (path, boost_args + exec_args, type, success, file.getvalue())
 
     executor = concurrent.futures.ThreadPoolExecutor(max_workers=args.jobs)
     futures = []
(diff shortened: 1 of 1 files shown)


kbr-:
@bhalevy @penberg could you review please?

test.py
```diff
@@ -299,7 +302,7 @@ def report_subcause(e):
             def report_subcause(e):
                 print('  with error {e}\n'.format(e=e), file=file)
             report_error(e, e, report_subcause=report_subcause)
-        return (path, boost_args + exec_args, success, file.getvalue())
+        return (path, boost_args + exec_args, type, success, file.getvalue())
```
bhalevy: is `type` used?
It's unused on line 321
and then it's retrieved on line 344, but not used.

test.py
```diff
@@ -310,9 +313,12 @@ def report_subcause(e):
         for _ in range(args.repeat):
             futures.append(executor.submit(run_test, path, test_type, exec_args))
 
+    results = []
     cookie = len(futures)
     for future in concurrent.futures.as_completed(futures):
-        test_path, test_args, success, out = future.result()
+        result = future.result()
+        results.append(result)
+        test_path, test_args, _, success, out  = result
```
bhalevy: nit: extra space before `=`

test.py
kbr-: Indeed, this `type` in line 305 is not used.
Type is only used on line 338 (`if r[2] != 'boost'`).
Changed `type` to `_`.

test.py
kbr-: fixed

kbr-:
v2:
- rebased
- changed `type` to `_` at line 344
- removed extra space

avikivity:
Dequeued, since the From: header was bad (did not include author name).

penberg:
@avikivity The _commit_ looks fine:

```
commit d1f84a85dc789e4802588f56bcdf3e51697b7b8e
Author:     Kamil Braun <kbraun@scylladb.com>
AuthorDate: Thu Jul 18 10:31:58 2019 +0200
Commit:     Kamil Braun <kbraun@scylladb.com>
CommitDate: Wed Jul 24 18:08:07 2019 +0200

    Add option to create a XUnit output file for non-boost tests in test.py.

    If the user specifies an output file name using "--xunit=<filename>",
    test.py will write the test results of non-boost tests to the file in the XUnit XML format.
    Every boost test creates its own results file already.
    Resolves #4680.

    Signed-off-by: Kamil Braun <kbraun@scylladb.com>
```

So it has to be Github doing something stupid for squash and merge... Perhaps it picks up the profile name, which I think was "incorrect" for Kamil when I merged...

kbr-:
New PR: #4757

penberg:
Yup, changing the profile name seems to have fixed the issue for us.

avikivity:
hurray!

PR #4760, opened 2019-07-25
Move reader_concurrency_semaphore in front of row_cache

### LINKED ISSUE #4758: cache-mostly read workloads can overload memory
Requests served from disk have a queue that limits memory use an expire timed-out requests, but requests served from cache do not. This can cause those requests to consume too much memory, and the reactor to process expired requests rather than timing them out and returning an error.

avikivity:
Currently, all requests accepted by the replica are passed to row_cache for processing. Misses are placed on a queue managed by reader_concurrency_semaphore where they can be admitted when resources allow, expired and rejected if the spend too long on the queue, or rejected outright if the queue is too long.

This approach has a problem when reads hit the cache and are hard to process (for example, due to reading many small rows per query). Each query will consume a relatively large of CPU time, and so it is easy to queue new requests faster than old requests can be retired. This results in memory consumption increasing without bounds.

This patchset changes the ordering so that the reader_concurrency_semaphore is in front of the cache. Each request is first queued on the reader_concurrency_semaphore so it has a chance to be rejected before expensive processing, fixing the problem.

Later, we might also add accounting for resources consumed by the cache.

Fixes #4758.

Tests: unit (dev, debug)

### DIFF
--- a/cql3/statements/select_statement.cc
+++ b/cql3/statements/select_statement.cc
@@ -526,6 +526,7 @@ indexed_table_select_statement::do_execute_base_query(
     auto timeout = db::timeout_clock::now() + options.get_timeout_config().*get_timeout_config_selector();
 
     struct base_query_state {
+        semaphore concurrency_limiter{100};
         query::result_merger merger;
         std::vector<primary_key> primary_keys;
         std::vector<primary_key>::iterator current_primary_key;
@@ -543,14 +544,16 @@ indexed_table_select_statement::do_execute_base_query(
         auto &merger = query_state.merger;
         auto &keys = query_state.primary_keys;
         auto &key_it = query_state.current_primary_key;
-        return repeat([this, &keys, &key_it, &merger, &proxy, &state, &options, cmd, timeout]() {
+        return repeat([this, &keys, &key_it, &merger, &proxy, &state, &options, cmd, timeout, &query_state]() {
             // Starting with 1 key, we check if the result was a short read, and if not,
             // we continue exponentially, asking for 2x more key than before
             auto key_it_end = std::min(key_it + std::distance(keys.begin(), key_it) + 1, keys.end());
             auto command = ::make_lw_shared<query::read_command>(*cmd);
 
             query::result_merger oneshot_merger(cmd->row_limit, query::max_partitions);
-            return map_reduce(key_it, key_it_end, [this, &proxy, &state, &options, cmd, timeout] (auto& key) {
+            return map_reduce(key_it, key_it_end, [this, &proxy, &state, &options, cmd, timeout, &query_state] (auto& key) {
+              // Limit concurrency here to avoid flooding the reader_concurrency_semaphore (and in general generating huge loads). See #3807.
+              return with_semaphore(query_state.concurrency_limiter, 1, [this, &proxy, &state, &options, cmd, timeout, key] {
                 auto command = ::make_lw_shared<query::read_command>(*cmd);
                 // for each partition, read just one clustering row (TODO: can
                 // get all needed rows of one partition at once.)
@@ -562,6 +565,7 @@ indexed_table_select_statement::do_execute_base_query(
                 .then([] (service::storage_proxy::coordinator_query_result qr) {
                     return std::move(qr.query_result);
                 });
+              });
             }, std::move(oneshot_merger)).then([&key_it, key_it_end = std::move(key_it_end), &keys, &merger] (foreign_ptr<lw_shared_ptr<query::result>> result) {
                 bool is_short_read = result->is_short_read();
                 merger(std::move(result));
--- a/row_cache.hh
+++ b/row_cache.hh
@@ -40,6 +40,7 @@
 #include <seastar/core/metrics_registration.hh>
 #include "flat_mutation_reader.hh"
 #include "mutation_cleaner.hh"
+#include "reader_concurrency_semaphore.hh"
 
 namespace bi = boost::intrusive;
 
@@ -495,7 +496,8 @@ public:
                                      const io_priority_class& = default_priority_class(),
                                      tracing::trace_state_ptr trace_state = nullptr,
                                      streamed_mutation::forwarding fwd = streamed_mutation::forwarding::no,
-                                     mutation_reader::forwarding fwd_mr = mutation_reader::forwarding::no);
+                                     mutation_reader::forwarding fwd_mr = mutation_reader::forwarding::no,
+                                     reader_resource_tracker = no_resource_tracking());
 
     flat_mutation_reader make_reader(schema_ptr s, const dht::partition_range& range = query::full_partition_range) {
         auto& full_slice = s->full_slice();
--- a/table.cc
+++ b/table.cc
@@ -330,11 +330,8 @@ table::make_sstable_reader(schema_ptr s,
                                    const io_priority_class& pc,
                                    tracing::trace_state_ptr trace_state,
                                    streamed_mutation::forwarding fwd,
-                                   mutation_reader::forwarding fwd_mr) const {
-    auto* semaphore = service::get_local_streaming_read_priority().id() == pc.id()
-        ? _config.streaming_read_concurrency_semaphore
-        : _config.read_concurrency_semaphore;
-
+                                   mutation_reader::forwarding fwd_mr,
+                                   reader_resource_tracker tracker) const {
     // CAVEAT: if make_sstable_reader() is called on a single partition
     // we want to optimize and read exactly this partition. As a
     // consequence, fast_forward_to() will *NOT* work on the result,
@@ -384,11 +381,7 @@ table::make_sstable_reader(schema_ptr s,
         }
     }();
 
-    if (semaphore) {
-        return make_restricted_flat_reader(*semaphore, std::move(ms), std::move(s), pr, slice, pc, std::move(trace_state), fwd, fwd_mr);
-    } else {
-        return ms.make_reader(std::move(s), pr, slice, pc, std::move(trace_state), fwd, fwd_mr);
-    }
+    return ms.make_reader(std::move(s), pr, slice, pc, std::move(trace_state), fwd, fwd_mr, tracker);
 }
 
 // Exposed for testing, not performance critical.
@@ -428,13 +421,14 @@ table::find_row(schema_ptr s, const dht::decorated_key& partition_key, clusterin
 }
 
 flat_mutation_reader
-table::make_reader(schema_ptr s,
+table::do_make_reader(schema_ptr s,
                            const dht::partition_range& range,
                            const query::partition_slice& slice,
                            const io_priority_class& pc,
                            tracing::trace_state_ptr trace_state,
                            streamed_mutation::forwarding fwd,
-                           mutation_reader::forwarding fwd_mr) const {
+                           mutation_reader::forwarding fwd_mr,
+                           reader_resource_tracker tracker) const {
     if (_virtual_reader) {
         return (*_virtual_reader).make_reader(s, range, slice, pc, trace_state, fwd, fwd_mr);
     }
@@ -467,9 +461,9 @@ table::make_reader(schema_ptr s,
     }
 
     if (_config.enable_cache && !slice.options.contains(query::partition_slice::option::bypass_cache)) {
-        readers.emplace_back(_cache.make_reader(s, range, slice, pc, std::move(trace_state), fwd, fwd_mr));
+        readers.emplace_back(_cache.make_reader(s, range, slice, pc, std::move(trace_state), fwd, fwd_mr, tracker));
     } else {
-        readers.emplace_back(make_sstable_reader(s, _sstables, range, slice, pc, std::move(trace_state), fwd, fwd_mr));
+        readers.emplace_back(make_sstable_reader(s, _sstables, range, slice, pc, std::move(trace_state), fwd, fwd_mr, tracker));
     }
 
     auto comb_reader = make_combined_reader(s, std::move(readers), fwd, fwd_mr);
@@ -477,6 +471,48 @@ table::make_reader(schema_ptr s,
         return _config.data_listeners->on_read(s, range, slice, std::move(comb_reader));
     } else {
         return comb_reader;
+    }
+}
+
+flat_mutation_reader
+table::make_reader(schema_ptr s,
+                           const dht::partition_range& range,
+                           const query::partition_slice& slice,
+                           const io_priority_class& pc,
+                           tracing::trace_state_ptr trace_state,
+                           streamed_mutation::forwarding fwd,
+                           mutation_reader::forwarding fwd_mr) const {
+    auto* semaphore = _config.read_concurrency_semaphore;
+
+    auto ms = mutation_source([this] (
+            schema_ptr s,
+            const dht::partition_range& pr,
+            const query::partition_slice& slice,
+            const io_priority_class& pc,
+            tracing::trace_state_ptr trace_state,
+            streamed_mutation::forwarding fwd,
+            mutation_reader::forwarding fwd_mr,
+            reader_resource_tracker tracker) {
+        return do_make_reader(std::move(s), pr, slice, pc, std::move(trace_state), fwd, fwd_mr, tracker);
+    });
+    return maybe_restrict_reader(semaphore, std::move(s), range, slice, pc, std::move(trace_state),
+            fwd, fwd_mr, std::move(ms));
+}
+
+flat_mutation_reader
+table::maybe_restrict_reader(reader_concurrency_semaphore* semaphore,
+        schema_ptr schema,
+        const dht::partition_range& range,
+        const query::partition_slice& slice,
+        const io_priority_class& pc,
+        tracing::trace_state_ptr trace_state,
+        streamed_mutation::forwarding fwd,
+        mutation_reader::forwarding fwd_mr,
+        mutation_source ms) {
+    if (semaphore) {
+        return make_restricted_flat_reader(*semaphore, std::move(ms), std::move(schema), range, slice, pc, std::move(trace_state), fwd, fwd_mr);
+    } else {
+        return ms.make_reader(std::move(schema), range, slice, pc, std::move(trace_state), fwd, fwd_mr);
     }
 }
 
@@ -504,8 +540,8 @@ table::make_streaming_reader(schema_ptr s,
     return make_flat_multi_range_reader(s, std::move(source), ranges, slice, pc, nullptr, mutation_reader::forwarding::no);
 }
 
-flat_mutation_reader table::make_streaming_reader(schema_ptr schema, const dht::partition_range& range,
-        const query::partition_slice& slice, mutation_reader::forwarding fwd_mr) const {
+flat_mutation_reader table::do_make_streaming_reader(schema_ptr schema, const dht::partition_range& range,
+        const query::partition_slice& slice, mutation_reader::forwarding fwd_mr, reader_resource_tracker tracker) const {
     const auto& pc = service::get_local_streaming_read_priority();
     auto trace_state = tracing::trace_state_ptr();
     const auto fwd = streamed_mutation::forwarding::no;
@@ -515,8 +551,24 @@ flat_mutation_reader table::make_streaming_reader(schema_ptr schema, const dht::
     for (auto&& mt : *_memtables) {
         readers.emplace_back(mt->make_flat_reader(schema, range, slice, pc, trace_state, fwd, fwd_mr));
     }
-    readers.emplace_back(make_sstable_reader(schema, _sstables, range, slice, pc, std::move(trace_state), fwd, fwd_mr));
+    readers.emplace_back(make_sstable_reader(schema, _sstables, range, slice, pc, std::move(trace_state), fwd, fwd_mr, std::move(tracker)));
     return make_combined_reader(std::move(schema), std::move(readers), fwd, fwd_mr);
+}
+
+flat_mutation_reader table::make_streaming_reader(schema_ptr schema, const dht::partition_range& range,
+        const query::partition_slice& slice, mutation_reader::forwarding fwd_mr) const {
+    const auto& pc = service::get_local_streaming_read_priority();
+    auto trace_state = tracing::trace_state_ptr();
+    const auto fwd = streamed_mutation::forwarding::no;
+
+    auto ms = mutation_source([this] (schema_ptr s, const dht::partition_range& pr, const query::partition_slice& slice,
+                    const io_priority_class& pc, tracing::trace_state_ptr trace, streamed_mutation::forwarding fwd_sm,
+                    mutation_reader::forwarding fwd_mr, reader_resource_tracker rt) {
+        return do_make_streaming_reader(std::move(s), pr, slice, fwd_mr, std::move(rt));
+    });
+
+    return maybe_restrict_reader(_config.streaming_read_concurrency_semaphore, std::move(schema), range, slice,
+            pc, std::move(trace_state), fwd, fwd_mr, std::move(ms));
 }
 
 future<std::vector<locked_cell>> table::lock_counter_cells(const mutation& m, db::timeout_clock::time_point timeout) {
@@ -1590,8 +1642,9 @@ table::sstables_as_snapshot_source() {
                 const io_priority_class& pc,
                 tracing::trace_state_ptr trace_state,
                 streamed_mutation::forwarding fwd,
-                mutation_reader::forwarding fwd_mr) {
-            return make_sstable_reader(std::move(s), sst_set, r, slice, pc, std::move(trace_state), fwd, fwd_mr);
+                mutation_reader::forwarding fwd_mr,
+                reader_resource_tracker tracker) {
+            return make_sstable_reader(std::move(s), sst_set, r, slice, pc, std::move(trace_state), fwd, fwd_mr, std::move(tracker));
         }, [this, sst_set] {
             return make_partition_presence_checker(sst_set);
         });
@@ -2441,7 +2494,8 @@ table::make_reader_excluding_sstable(schema_ptr s,
     auto effective_sstables = ::make_lw_shared<sstables::sstable_set>(*_sstables);
     effective_sstables->erase(sst);
 
-    readers.emplace_back(make_sstable_reader(s, std::move(effective_sstables), range, slice, pc, std::move(trace_state), fwd, fwd_mr));
+    // no resource tracking as this there is just one reader like this
+    readers.emplace_back(make_sstable_reader(s, std::move(effective_sstables), range, slice, pc, std::move(trace_state), fwd, fwd_mr, no_resource_tracking()));
     return make_combined_reader(s, std::move(readers), fwd, fwd_mr);
 }
 
--- a/table.cc
+++ b/table.cc
@@ -336,25 +336,16 @@ table::make_sstable_reader(schema_ptr s,
     // we want to optimize and read exactly this partition. As a
     // consequence, fast_forward_to() will *NOT* work on the result,
     // regardless of what the fwd_mr parameter says.
-    {
-        if (pr.is_singular() && pr.start()->value().has_key()) {
-            const dht::ring_position& pos = pr.start()->value();
-            if (dht::shard_of(pos.token()) != engine().cpu_id()) {
-                {
-                    return make_empty_flat_reader(s); // range doesn't belong to this shard
-                }
-            }
-
-            {
-                return create_single_key_sstable_reader(const_cast<column_family*>(this), std::move(s), std::move(sstables),
-                        _stats.estimated_sstable_per_read, pr, slice, pc, tracker, std::move(trace_state), fwd, fwd_mr);
-            }
-        } else {
-            {
-                return make_local_shard_sstable_reader(std::move(s), std::move(sstables), pr, slice, pc,
-                        tracker, std::move(trace_state), fwd, fwd_mr);
-            }
+    if (pr.is_singular() && pr.start()->value().has_key()) {
+        const dht::ring_position& pos = pr.start()->value();
+        if (dht::shard_of(pos.token()) != engine().cpu_id()) {
+            return make_empty_flat_reader(s); // range doesn't belong to this shard
         }
+        return create_single_key_sstable_reader(const_cast<column_family*>(this), std::move(s), std::move(sstables),
+                _stats.estimated_sstable_per_read, pr, slice, pc, tracker, std::move(trace_state), fwd, fwd_mr);
+    } else {
+        return make_local_shard_sstable_reader(std::move(s), std::move(sstables), pr, slice, pc,
+                tracker, std::move(trace_state), fwd, fwd_mr);
     }
 }
 
--- a/table.cc
+++ b/table.cc
@@ -336,52 +336,26 @@ table::make_sstable_reader(schema_ptr s,
     // we want to optimize and read exactly this partition. As a
     // consequence, fast_forward_to() will *NOT* work on the result,
     // regardless of what the fwd_mr parameter says.
-    auto ms = [&] () -> mutation_source {
+    {
         if (pr.is_singular() && pr.start()->value().has_key()) {
             const dht::ring_position& pos = pr.start()->value();
             if (dht::shard_of(pos.token()) != engine().cpu_id()) {
-                return mutation_source([] (
-                        schema_ptr s,
-                        const dht::partition_range& pr,
-                        const query::partition_slice& slice,
-                        const io_priority_class& pc,
-                        tracing::trace_state_ptr trace_state,
-                        streamed_mutation::forwarding fwd,
-                        mutation_reader::forwarding fwd_mr,
-                        reader_resource_tracker tracker) {
+                {
                     return make_empty_flat_reader(s); // range doesn't belong to this shard
-                });
+                }
             }
 
-            return mutation_source([this, sstables=std::move(sstables)] (
-                    schema_ptr s,
-                    const dht::partition_range& pr,
-                    const query::partition_slice& slice,
-                    const io_priority_class& pc,
-                    tracing::trace_state_ptr trace_state,
-                    streamed_mutation::forwarding fwd,
-                    mutation_reader::forwarding fwd_mr,
-                    reader_resource_tracker tracker) {
+            {
                 return create_single_key_sstable_reader(const_cast<column_family*>(this), std::move(s), std::move(sstables),
                         _stats.estimated_sstable_per_read, pr, slice, pc, tracker, std::move(trace_state), fwd, fwd_mr);
-            });
+            }
         } else {
-            return mutation_source([sstables=std::move(sstables)] (
-                    schema_ptr s,
-                    const dht::partition_range& pr,
-                    const query::partition_slice& slice,
-                    const io_priority_class& pc,
-                    tracing::trace_state_ptr trace_state,
-                    streamed_mutation::forwarding fwd,
-                    mutation_reader::forwarding fwd_mr,
-                    reader_resource_tracker tracker) {
+            {
                 return make_local_shard_sstable_reader(std::move(s), std::move(sstables), pr, slice, pc,
                         tracker, std::move(trace_state), fwd, fwd_mr);
-            });
+            }
         }
-    }();
-
-    return ms.make_reader(std::move(s), pr, slice, pc, std::move(trace_state), fwd, fwd_mr, tracker);
+    }
 }
 
 // Exposed for testing, not performance critical.
--- a/table.cc
+++ b/table.cc
@@ -498,13 +498,7 @@ table::make_streaming_reader(schema_ptr s,
 
     auto source = mutation_source([this] (schema_ptr s, const dht::partition_range& range, const query::partition_slice& slice,
                                       const io_priority_class& pc, tracing::trace_state_ptr trace_state, streamed_mutation::forwarding fwd, mutation_reader::forwarding fwd_mr) {
-        std::vector<flat_mutation_reader> readers;
-        readers.reserve(_memtables->size() + 1);
-        for (auto&& mt : *_memtables) {
-            readers.emplace_back(mt->make_flat_reader(s, range, slice, pc, trace_state, fwd, fwd_mr));
-        }
-        readers.emplace_back(make_sstable_reader(s, _sstables, range, slice, pc, std::move(trace_state), fwd, fwd_mr));
-        return make_combined_reader(s, std::move(readers), fwd, fwd_mr);
+        return make_streaming_reader(std::move(s), range);
     });
 
     return make_flat_multi_range_reader(s, std::move(source), ranges, slice, pc, nullptr, mutation_reader::forwarding::no);
--- a/table.cc
+++ b/table.cc
@@ -356,7 +356,7 @@ table::make_sstable_reader(schema_ptr s,
                 });
             }
 
-            return mutation_source([semaphore, this, sstables=std::move(sstables)] (
+            return mutation_source([this, sstables=std::move(sstables)] (
                     schema_ptr s,
                     const dht::partition_range& pr,
                     const query::partition_slice& slice,
@@ -369,7 +369,7 @@ table::make_sstable_reader(schema_ptr s,
                         _stats.estimated_sstable_per_read, pr, slice, pc, tracker, std::move(trace_state), fwd, fwd_mr);
             });
         } else {
-            return mutation_source([semaphore, sstables=std::move(sstables)] (
+            return mutation_source([sstables=std::move(sstables)] (
                     schema_ptr s,
                     const dht::partition_range& pr,
                     const query::partition_slice& slice,
--- a/database.hh
+++ b/database.hh
@@ -592,12 +592,36 @@ private:
                                         const io_priority_class& pc,
                                         tracing::trace_state_ptr trace_state,
                                         streamed_mutation::forwarding fwd,
-                                        mutation_reader::forwarding fwd_mr) const;
+                                        mutation_reader::forwarding fwd_mr,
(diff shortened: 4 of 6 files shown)


avikivity:
@denesb @tgrabiec please review carefully
@psarna you are also mentioned

avikivity:
I did some smoke tests with cassandra-stress, and verified that database_active_reads is <= 100.

glommer:
Wait, glommer didn't approve anything. I meant to approve a single patch!
Is that even doable ?

cql3/statements/select_statement.cc
```diff
@@ -543,14 +544,16 @@ indexed_table_select_statement::do_execute_base_query(
...
             // Starting with 1 key, we check if the result was a short read, and if not,
             // we continue exponentially, asking for 2x more key than before
             auto key_it_end = std::min(key_it + std::distance(keys.begin(), key_it) + 1, keys.end());
             auto command = ::make_lw_shared<query::read_command>(*cmd);
 
             query::result_merger oneshot_merger(cmd->row_limit, query::max_partitions);
-            return map_reduce(key_it, key_it_end, [this, &proxy, &state, &options, cmd, timeout] (auto& key) {
+            return map_reduce(key_it, key_it_end, [this, &proxy, &state, &options, cmd, timeout, &query_state] (auto& key) {
+              // Limit concurrency here to avoid flooding the reader_concurrency_semaphore (and in general generating huge loads). See #3807.
+              return with_semaphore(query_state.concurrency_limiter, 1, [this, &proxy, &state, &options, cmd, timeout, key] {
```
glommer: if I understand correctly what this patch does, it seems good even outside of the context of the reader_concurrency_semaphore.

glommer:
I am okay with the general strategy of having the resource tracker moved to where it is now, and the patch to limit concurrency at the statement level seems good regardless of the rest.

Nothing that caught my eye on the negative side and this will increase our protection and ability to expire requests early.

cql3/statements/select_statement.cc
```diff
@@ -543,14 +544,16 @@ indexed_table_select_statement::do_execute_base_query(
         auto &merger = query_state.merger;
         auto &keys = query_state.primary_keys;
         auto &key_it = query_state.current_primary_key;
-        return repeat([this, &keys, &key_it, &merger, &proxy, &state, &options, cmd, timeout]() {
+        return repeat([this, &keys, &key_it, &merger, &proxy, &state, &options, cmd, timeout, &query_state]() {
             // Starting with 1 key, we check if the result was a short read, and if not,
             // we continue exponentially, asking for 2x more key than before
```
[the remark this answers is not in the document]
psarna: *** I also edited the example implementation by changing `keys.begin() + 100` to `key_it + 100`, I'm writing it just in case GitHub doesn't notify when a comment is edited.

avikivity:
> I'm marking this review as `request changes`, because I think there's a simpler and cleaner way to make this change (see my comment), but do notice that I also think that this patch **works**, so if it's an emergency, then I'm fine with merging it as is and I can provide follow-up patches later.

Someone with an emergency can always apply the patch on a private branch, and patches should be reviewed objectively and not given special consideration for being in a hurry. Of course it's okay to defer major work sometimes, especially if the patch is not the direct cause but just a victim.

table.cc
```diff
@@ -480,6 +474,48 @@ table::make_reader(schema_ptr s,
     }
 }
 
+flat_mutation_reader
+table::make_reader(schema_ptr s,
```
denesb: We have a method `table::as_mutation_source()` which calls this `table::make_reader()` to make its reader, which now in turn makes another internal `mutation_source` which it optionally restricts.
Perhaps it would be cleaner to have an private version of `as_mutation_source_internal()`, which wraps `table::do_make_reader()` and have all other method create their `mutation_source`:s with this.
The public `as_mutation_source()` would look like this:
```
auto ms = as_mutation_source_internal();
if (needs_restricting()) {
    return restrict_mutation_source(ms);
}
return ms;
```
Just a suggestion.

table.cc
```diff
@@ -480,6 +474,48 @@ table::make_reader(schema_ptr s,
...
+                           const dht::partition_range& range,
+                           const query::partition_slice& slice,
+                           const io_priority_class& pc,
+                           tracing::trace_state_ptr trace_state,
+                           streamed_mutation::forwarding fwd,
+                           mutation_reader::forwarding fwd_mr) const {
+    auto* semaphore = _config.read_concurrency_semaphore;
```
denesb: So happy to see that fragile heuristics on the `io_priority_class` gone!

table.cc
```diff
@@ -480,6 +474,48 @@ table::make_reader(schema_ptr s,
...
+            const dht::partition_range& pr,
+            const query::partition_slice& slice,
+            const io_priority_class& pc,
+            tracing::trace_state_ptr trace_state,
+            streamed_mutation::forwarding fwd,
+            mutation_reader::forwarding fwd_mr,
+            reader_resource_tracker tracker) {
+        return do_make_reader(std::move(s), pr, slice, pc, std::move(trace_state), fwd, fwd_mr, tracker);
+    });
+    return maybe_restrict_reader(semaphore, std::move(s), range, slice, pc, std::move(trace_state),
```
denesb: nit: you could pass `_config.read_concurrency_semaphore` here directly.

table.cc
```diff
@@ -2441,7 +2494,8 @@ table::make_reader_excluding_sstable(schema_ptr s,
     auto effective_sstables = ::make_lw_shared<sstables::sstable_set>(*_sstables);
     effective_sstables->erase(sst);
 
-    readers.emplace_back(make_sstable_reader(s, std::move(effective_sstables), range, slice, pc, std::move(trace_state), fwd, fwd_mr));
+    // no resource tracking as this there is just one reader like this
```
tgrabiec: Not sure why this is a reason not to track it. This reader still consumes resources. If we have many sstables, it can consume a lot of memory.

tgrabiec:
I think that this change may hurt performance in case there is a mix of hits and misses. 

Before, hits would consistently have low latency (and thus high throughput) regardless of the number of active reads which missed. Now they share the semaphore, so if the number of active misses is larger than 100, cache-hitting reads will start to observe latency of a missing read. This will hurt throughput, the CPU will be underutilized.

Perhaps we should use separate concurrency semaphores for different resources (disk vs CPU) in order to achieve better utilization.

If we had 2 disks, it would also make sense to use different semaphore for each disk.

This is orthogonal to memory accounting which prevents OOM.

cql3/statements/select_statement.cc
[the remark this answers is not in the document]
avikivity: Yes, it looks like the author of the patch can't read.

We have exponentials without cutoff elsewhere too, you aren't the only one guilty of it. And in fact it's hard to think of a good cutoff, low ones will result in slow processing on large clusters that _can_ handle the concurrency.

I'll adjust the code to use std::min() as you suggest.

row_cache.hh
```diff
@@ -495,7 +496,8 @@ public:
                                      const io_priority_class& = default_priority_class(),
                                      tracing::trace_state_ptr trace_state = nullptr,
                                      streamed_mutation::forwarding fwd = streamed_mutation::forwarding::no,
-                                     mutation_reader::forwarding fwd_mr = mutation_reader::forwarding::no);
+                                     mutation_reader::forwarding fwd_mr = mutation_reader::forwarding::no,
+                                     reader_resource_tracker = no_resource_tracking());
```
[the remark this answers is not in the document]
avikivity: Looking at mutation_source::make_reader(), it already does, no?

table.cc
avikivity: Yes, I can call it as_unrestricted_mutation_source().

table.cc
avikivity: yes

table.cc
avikivity: Agree it's safer. @psarna opinion?

avikivity:
> I think that this change may hurt performance in case there is a mix of hits and misses.
> 
> Before, hits would consistently have low latency (and thus high throughput) regardless of the number of active reads which missed. Now they share the semaphore, so if the number of active misses is larger than 100, cache-hitting reads will start to observe latency of a missing read. This will hurt throughput, the CPU will be underutilized.

If the number of active reads is larger, and they are all misses, then the cache hit rate is <1% (of course fluctuations mean that it can happen with a cache hit rate larger than 1%, but not by much).

If some of the active reads are hits, then they will complete and make room for another hit (or miss, taking the hit rate below 1% again). The cache concurrency is 1, it doesn't need concurrency to run at full speed.

In practical implementations, if the cache hit rate is below 1% then we will max out on CPU as miss processing is much more expensive than hit processing. We'll only have spare CPU if we're talking to an extermely slow disk (consider an array of 100+ spinning disks connected to one shard, I think that's the only case we regress with this patch).

> 
> Perhaps we should use separate concurrency semaphores for different resources (disk vs CPU) in order to achieve better utilization.
> 
> If we had 2 disks, it would also make sense to use different semaphore for each disk.
> 

That only works if we know in advance which disk a query is going to use.

We'll have a similar problem with workload prioritization and users, but in that case, the user is known in advance.

> This is orthogonal to memory accounting which prevents OOM.

cql3/statements/select_statement.cc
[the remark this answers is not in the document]
avikivity: The author of the comment above can't read either, you also told me about the other places. :man_facepalming:

avikivity:
Updated:
 - use a named constant to limit indexed select concurrency
 - use existing indexed select concurrency control code instead of adding a semaphore
 - spin off a helper table::as_unrestricted_mutation_source() from table::make_reader()
 - avoid use-once "semaphore" local variable and use the class member directly
 - do not exclude make_reader_excluding_sstable() from going through make_restricting_reader()

No change:
 - @tgrabiec's comment about hits being starved when there are 100 concurrent misses, I think we are okay here (see my reply to the comment)

gleb-cloudius:
On Sat, Jul 27, 2019 at 10:26:36PM -0700, Avi Kivity wrote:
> > I think that this change may hurt performance in case there is a mix of hits and misses.
> > 
> > Before, hits would consistently have low latency (and thus high throughput) regardless of the number of active reads which missed. Now they share the semaphore, so if the number of active misses is larger than 100, cache-hitting reads will start to observe latency of a missing read. This will hurt throughput, the CPU will be underutilized.
> 
> If the number of active reads is larger, and they are all misses, then the cache hit rate is <1% (of course fluctuations mean that it can happen with a cache hit rate larger than 1%, but not by much).
> 
> If some of the active reads are hits, then they will complete and make room for another hit (or miss, taking the hit rate below 1% again). The cache concurrency is 1, it doesn't need concurrency to run at full speed.
> 
Don't we moving the semaphore before the cache exactly because we
observed that with large reads from the cache concurrency stops
been 1.

--
			Gleb.

avikivity:
> On Sat, Jul 27, 2019 at 10:26:36PM -0700, Avi Kivity wrote: > I think that this change may hurt performance in case there is a mix of hits and misses. > > Before, hits would consistently have low latency (and thus high throughput) regardless of the number of active reads which missed. Now they share the semaphore, so if the number of active misses is larger than 100, cache-hitting reads will start to observe latency of a missing read. This will hurt throughput, the CPU will be underutilized. If the number of active reads is larger, and they are all misses, then the cache hit rate is <1% (of course fluctuations mean that it can happen with a cache hit rate larger than 1%, but not by much). If some of the active reads are hits, then they will complete and make room for another hit (or miss, taking the hit rate below 1% again). The cache concurrency is 1, it doesn't need concurrency to run at full speed.
> Don't we moving the semaphore before the cache exactly because we observed that with large reads from the cache concurrency stops been 1.

The cache needs a concurrency of 1 for maximum throughput in hits (unlike disks). Scheduling can make it run with higher concurrency, but it doesn't need it.

gleb-cloudius:
On Sun, Jul 28, 2019 at 02:44:43AM -0700, Avi Kivity wrote:
> > On Sat, Jul 27, 2019 at 10:26:36PM -0700, Avi Kivity wrote: > I think that this change may hurt performance in case there is a mix of hits and misses. > > Before, hits would consistently have low latency (and thus high throughput) regardless of the number of active reads which missed. Now they share the semaphore, so if the number of active misses is larger than 100, cache-hitting reads will start to observe latency of a missing read. This will hurt throughput, the CPU will be underutilized. If the number of active reads is larger, and they are all misses, then the cache hit rate is <1% (of course fluctuations mean that it can happen with a cache hit rate larger than 1%, but not by much). If some of the active reads are hits, then they will complete and make room for another hit (or miss, taking the hit rate below 1% again). The cache concurrency is 1, it doesn't need concurrency to run at full speed.
> > Don't we moving the semaphore before the cache exactly because we observed that with large reads from the cache concurrency stops been 1.
> 
> The cache needs a concurrency of 1 for maximum throughput in hits (unlike disks). Scheduling can make it run with higher concurrency, but it doesn't need it.
> 
With multiple scheduling groups though letting one hit run to completion
may cause runtime imbalance, no?

--
			Gleb.

avikivity:
master has only one scheduling group for cache. For multiple users we'll need multiple reader_concurrency_semaphores (for memory accounting too).

table.cc
psarna: I'm not 100% sure what "resource tracking" is, but if it's taking the resources into account in order to limit resource usage, then I agree it should be taken into account. Currently the view update generator serves the sstables one by one (per shard) and limits its max queue size to `5` with a semaphore, but it sounds more future-proof to add tracking now.

tgrabiec:
> > I think that this change may hurt performance in case there is a mix of hits and misses.
> > Before, hits would consistently have low latency (and thus high throughput) regardless of the number of active reads which missed. Now they share the semaphore, so if the number of active misses is larger than 100, cache-hitting reads will start to observe latency of a missing read. This will hurt throughput, the CPU will be underutilized.
> 
> If the number of active reads is larger than 100, and they are all misses, then the cache hit rate is <1% (of course fluctuations mean that it can happen with a cache hit rate larger than 1%, but not by much).

Why does it have to be the case that hit rate is <1%?

I think it doesn't have to be. For instance, if there are two groups of processes making read requests. One group has a high cache hit ratio and then another group starts which does something miss-intensive. Since the second group has lower throughput due to disk latency, even if it needs the same amount of CPU per request, the first group will dominate.

> 
> If some of the active reads are hits, then they will complete and make room for another hit (or miss, taking the hit rate below 1% again). The cache concurrency is 1, it doesn't need concurrency to run at full speed.

Yes, but if there is more than 100 active misses then hits will experience an increase in latency due to queueing. How often depends on the concurrency of hits. In particular, if there is one process which constantly hits, and there are more than 100 active misses, then that process will always experience the latency of the queued and active misses. Whereas before the patch it wouldn't. This will reduce the throughput of such a process (assuming there is spare CPU).

> 
> In practical implementations, if the cache hit rate is below 1% then we will max out on CPU as miss processing is much more expensive than hit processing. We'll only have spare CPU if we're talking to an extermely slow disk (consider an array of 100+ spinning disks connected to one shard, I think that's the only case we regress with this patch).

If we max out on the CPU in a full-miss case, it means we can't fully utilize the disk. That sounds like a bug in the design or an unbalanced system. Disk's bandwidth is partitioned among many shards, and also severely reduced by ongoing write IOs. I recall seeing IO-bound workloads in the field.

> 
> > Perhaps we should use separate concurrency semaphores for different resources (disk vs CPU) in order to achieve better utilization.
> > If we had 2 disks, it would also make sense to use different semaphore for each disk.
> 
> That only works if we know in advance which disk a query is going to use.
> 
> We'll have a similar problem with workload prioritization and users, but in that case, the user is known in advance.
> 
> > This is orthogonal to memory accounting which prevents OOM.

avikivity:
> > > I think that this change may hurt performance in case there is a mix of hits and misses.
> > > Before, hits would consistently have low latency (and thus high throughput) regardless of the number of active reads which missed. Now they share the semaphore, so if the number of active misses is larger than 100, cache-hitting reads will start to observe latency of a missing read. This will hurt throughput, the CPU will be underutilized.
> > 
> > 
> > If the number of active reads is larger than 100, and they are all misses, then the cache hit rate is <1% (of course fluctuations mean that it can happen with a cache hit rate larger than 1%, but not by much).
> 
> Why does it have to be the case that hit rate is <1%?
> 
> I think it doesn't have to be. For instance, if there are two groups of processes making read requests. One group has a high cache hit ratio and then another group starts which does something miss-intensive. Since the second group has lower throughput due to disk latency, even if it needs the same amount of CPU per request, the first group will dominate.


If those two groups need the same amount of CPU per request, then they have similar throughput in requests per second and the hit rate is 50%. Can you provide a concrete scenario where we regress?


> 
> > If some of the active reads are hits, then they will complete and make room for another hit (or miss, taking the hit rate below 1% again). The cache concurrency is 1, it doesn't need concurrency to run at full speed.
> 
> Yes, but if there is more than 100 active misses then hits will experience an increase in latency due to queueing. How often depends on the concurrency of hits. In particular, if there is one process which constantly hits, and there are more than 100 active misses, then that process will always experience the latency of the queued and active misses. Whereas before the patch it wouldn't. This will reduce the throughput of such a process (assuming there is spare CPU).

Having 100 active misses AND spare CPU requires that misses have very long disk latency (my spinning disk example). With SSD, latency is O(task quota) so 0.5ms. Since we are able to serve 20-50 kops/sec/core (usually less due to hyperthreading and due to misses being more expensive), Little's Law gives us effective concurrency of 10-25 (and again usually less).  Above that concurrency we are CPU bound.

> 
> > In practical implementations, if the cache hit rate is below 1% then we will max out on CPU as miss processing is much more expensive than hit processing. We'll only have spare CPU if we're talking to an extermely slow disk (consider an array of 100+ spinning disks connected to one shard, I think that's the only case we regress with this patch).
> 
> If we max out on the CPU in a full-miss case, it means we can't fully utilize the disk. That sounds like a bug in the design or an unbalanced system. Disk's bandwidth is partitioned among many shards, and also severely reduced by ongoing write IOs. I recall seeing IO-bound workloads in the field.

If a workload is I/O bound, then running requests through the cache outside the semaphore won't help it. Yes, that request will complete quickly if it hits, but the next one that misses will just have to wait longer for the next miss to complete.

> 
> > > Perhaps we should use separate concurrency semaphores for different resources (disk vs CPU) in order to achieve better utilization.
> > > If we had 2 disks, it would also make sense to use different semaphore for each disk.
> > 
> > 
> > That only works if we know in advance which disk a query is going to use.
> > We'll have a similar problem with workload prioritization and users, but in that case, the user is known in advance.
> > > This is orthogonal to memory accounting which prevents OOM.

glommer:
On Mon, Jul 29, 2019 at 10:48 AM Avi Kivity <notifications@github.com>
wrote:

> I think that this change may hurt performance in case there is a mix of
> hits and misses.
> Before, hits would consistently have low latency (and thus high
> throughput) regardless of the number of active reads which missed. Now they
> share the semaphore, so if the number of active misses is larger than 100,
> cache-hitting reads will start to observe latency of a missing read. This
> will hurt throughput, the CPU will be underutilized.
>
> If the number of active reads is larger than 100, and they are all misses,
> then the cache hit rate is <1% (of course fluctuations mean that it can
> happen with a cache hit rate larger than 1%, but not by much).
>
> Why does it have to be the case that hit rate is <1%?
>
> I think it doesn't have to be. For instance, if there are two groups of
> processes making read requests. One group has a high cache hit ratio and
> then another group starts which does something miss-intensive. Since the
> second group has lower throughput due to disk latency, even if it needs the
> same amount of CPU per request, the first group will dominate.
>
> If those two groups need the same amount of CPU per request, then they
> have similar throughput in requests per second and the hit rate is 50%. Can
> you provide a concrete scenario where we regress?
>
> If some of the active reads are hits, then they will complete and make
> room for another hit (or miss, taking the hit rate below 1% again). The
> cache concurrency is 1, it doesn't need concurrency to run at full speed.
>
> Yes, but if there is more than 100 active misses then hits will experience
> an increase in latency due to queueing. How often depends on the
> concurrency of hits. In particular, if there is one process which
> constantly hits, and there are more than 100 active misses, then that
> process will always experience the latency of the queued and active misses.
> Whereas before the patch it wouldn't. This will reduce the throughput of
> such a process (assuming there is spare CPU).
>
> Having 100 active misses AND spare CPU requires that misses have very long
> disk latency (my spinning disk example). With SSD, latency is O(task quota)
> so 0.5ms. Since we are able to serve 20-50 kops/sec/core (usually less due
> to hyperthreading and due to misses being more expensive), Little's Law
> gives us effective concurrency of 10-25 (and again usually less). Above
> that concurrency we are CPU bound.
>

If the disk is under heavy writes, the latency can easily get to 2 - 4ms in
common SSDs.
Sure, a small read will cost less if that's all we do, but it may face that
queueing on its way to the device. If we multiply your number by a factor
of 4 to adjust for that, we're easily hitting 100.

table.cc
avikivity: I did so in the latest version. Yes, resource tracking tracks how much memory is used by the reader and is used to queue new readers until the older ones retire and release their memory,

avikivity:
> If the disk is under heavy writes, the latency can easily get to 2 - 4ms in common SSDs. Sure, a small read will cost less if that's all we do, but it may face that queueing on its way to the device. If we multiply your number by a factor of 4 to adjust for that, we're easily hitting 100.

If we're doing heavy writes then we won't be doing 50k ops/sec, some of the CPU will go towards the writes (and 50k/sec/core is optimistic for full miss workload; we should benchmark that).

avikivity:
I did not address @tgrabiec's point about asymmetric processes. Yes, those can now behave differently. But I don't see a choice, we can't just admit a read to the cache and hope it's a hit, if it turns out to be a partial miss it will hold a lot of memory in its buffers and we'll just run out.

denesb:
> I did not address @tgrabiec's point about asymmetric processes. Yes, those can now behave differently. But I don't see a choice, we can't just admit a read to the cache and hope it's a hit, if it turns out to be a partial miss it will hold a lot of memory in its buffers and we'll just run out.

If we had a cheap and quick way to determine whether a read will fully hit or miss we could solve this by queueing reads that are known/suspected to hit and allowing reads that we know will hit. Even something like a bloom filter (allowing false positives but never false negatives) would be enough. Alas I don't know if that is possible/easy to do. I seem to recall something like that being discussed recently.
Something like:
```
if (auto res_opt = cache.try_read(cmd)) {
    return make_ready_future<...>(std::move(*res_opt));
}
return enqueue(cmd);
```

avikivity:
> LGTM (mostly based on the changelog).
> 
> Due to github's messed up commit order it's not possible to properly review anything but the very first version.
> 

I can send as patches too if it helps.

avikivity:
> > I did not address @tgrabiec's point about asymmetric processes. Yes, those can now behave differently. But I don't see a choice, we can't just admit a read to the cache and hope it's a hit, if it turns out to be a partial miss it will hold a lot of memory in its buffers and we'll just run out.
> 
> If we had a cheap and quick way to determine whether a read will fully hit or miss we could solve this by queueing reads that are known/suspected to hit and allowing reads that we know will hit. Even something like a bloom filter (allowing false positives but never false negatives) would be enough. Alas I don't know if that is possible/easy to do. I seem to recall something like that being discussed recently.
> Something like:
> 
> ```
> if (auto res_opt = cache.try_read(cmd)) {
>     return make_ready_future<...>(std::move(*res_opt));
> }
> return enqueue(cmd);
> ```

I do think that try_read() can simplify the cache significantly by avoiding weaving, but IIRC @tgrabiec disagreed and in any case it cannot block this fix. Of course we have to make sure we don't regress in performance here too.

try_read() does address @tgrabiec's concern by making cache hits always have a concurrency of 1 and making them not compete with misses, so I think it complements this change nicely.

denesb:
> > > I did not address @tgrabiec's point about asymmetric processes. Yes, those can now behave differently. But I don't see a choice, we can't just admit a read to the cache and hope it's a hit, if it turns out to be a partial miss it will hold a lot of memory in its buffers and we'll just run out.
> > 
> > 
> > If we had a cheap and quick way to determine whether a read will fully hit or miss we could solve this by queueing reads that are known/suspected to hit and allowing reads that we know will hit. Even something like a bloom filter (allowing false positives but never false negatives) would be enough. Alas I don't know if that is possible/easy to do. I seem to recall something like that being discussed recently.
> > Something like:
> > ```
> > if (auto res_opt = cache.try_read(cmd)) {
> >     return make_ready_future<...>(std::move(*res_opt));
> > }
> > return enqueue(cmd);
> > ```
> 
> I do think that try_read() can simplify the cache significantly by avoiding weaving, but IIRC @tgrabiec disagreed and in any case it cannot block this fix. Of course we have to make sure we don't regress in performance here too.
> 
> try_read() does address @tgrabiec's concern by making cache hits always have a concurrency of 1 and making them not compete with misses, so I think it complements this change nicely.

Preventing weaving in cache is an entirely different discussion. I was thinking about `try_read()` purely in the context of queueing reads. I agree about it not blocking this fix. It can be a way out later when we do find workloads that show a regression in performance due to this fix.

denesb:
> > LGTM (mostly based on the changelog).
> > Due to github's messed up commit order it's not possible to properly review anything but the very first version.
> 
> I can send as patches too if it helps.

That would potentially cause comments being split between github and the email thread, which I'm sure you agree is very undesirable. I'll learn to live with github and will stop complaining. :)

avikivity:
> > > > I did not address @tgrabiec's point about asymmetric processes. Yes, those can now behave differently. But I don't see a choice, we can't just admit a read to the cache and hope it's a hit, if it turns out to be a partial miss it will hold a lot of memory in its buffers and we'll just run out.
> > > 
> > > 
> > > If we had a cheap and quick way to determine whether a read will fully hit or miss we could solve this by queueing reads that are known/suspected to hit and allowing reads that we know will hit. Even something like a bloom filter (allowing false positives but never false negatives) would be enough. Alas I don't know if that is possible/easy to do. I seem to recall something like that being discussed recently.
> > > Something like:
> > > ```
> > > if (auto res_opt = cache.try_read(cmd)) {
> > >     return make_ready_future<...>(std::move(*res_opt));
> > > }
> > > return enqueue(cmd);
> > > ```
> > 
> > 
> > I do think that try_read() can simplify the cache significantly by avoiding weaving, but IIRC @tgrabiec disagreed and in any case it cannot block this fix. Of course we have to make sure we don't regress in performance here too.
> > try_read() does address @tgrabiec's concern by making cache hits always have a concurrency of 1 and making them not compete with misses, so I think it complements this change nicely.
> 
> Preventing weaving in cache is an entirely different discussion. I was thinking about `try_read()` purely in the context of queueing reads. I agree about it not blocking this fix. It can be a way out later when we do find workloads that show a regression in performance due to this fix.

try_read() not only checks, but also reads, so it avoids weaving too and that was its main intent. Just checking would increase the cost of both hits and misses, and is not accurate since the cache can change between the check and the read.

denesb:
> try_read() not only checks, but also reads, so it avoids weaving too and that was its main intent. Just checking would increase the cost of both hits and misses, and is not accurate since the cache can change between the check and the read.

If I remember correctly the previous discussion, weaving is avoided, if after discovering that the read will not be able to be served from cache in its entirety, we read the whole thing from disk, then update cache. Trying again, after some sitting in a queue, can still be a weaving read in that it uses the current mechanisms (populating readers) to perform the read.

glommer:
> > > LGTM (mostly based on the changelog).
> > > Due to github's messed up commit order it's not possible to properly review anything but the very first version.
> > 
> > 
> > I can send as patches too if it helps.
> 
> That would potentially cause comments being split between github and the email thread, which I'm sure you agree is very undesirable. I'll learn to live with github and will stop complaining. :)

We should tag Dejan (@dekimir) as a reviewer in every single github PR

avikivity:
@denesb are you implying that logical ordering of commits is more important than syntax highlighting?!

tgrabiec:
> > > > I think that this change may hurt performance in case there is a mix of hits and misses.
> > > > Before, hits would consistently have low latency (and thus high throughput) regardless of the number of active reads which missed. Now they share the semaphore, so if the number of active misses is larger than 100, cache-hitting reads will start to observe latency of a missing read. This will hurt throughput, the CPU will be underutilized.
> > > 
> > > 
> > > If the number of active reads is larger than 100, and they are all misses, then the cache hit rate is <1% (of course fluctuations mean that it can happen with a cache hit rate larger than 1%, but not by much).
> > 
> > 
> > Why does it have to be the case that hit rate is <1%?
> > I think it doesn't have to be. For instance, if there are two groups of processes making read requests. One group has a high cache hit ratio and then another group starts which does something miss-intensive. Since the second group has lower throughput due to disk latency, even if it needs the same amount of CPU per request, the first group will dominate.
> 
> If those two groups need the same amount of CPU per request, then they have similar throughput in requests per second and the hit rate is 50%. 

I think that doesn't have to be 50%. If the miss workload leaves more than 50% of the CPU to the hit workload, then the hit workload will have higher throughput.

> Can you provide a concrete scenario where we regress?

Let's start with just the miss workload running. It's I/O bound, doesn't saturate the CPU, 100 active reads on average. The throughput is 25k reads/s, so average read latency is 4ms.

Add the hit workload, which has concurrency of 1. Before the patch it would have average latency of a task quota (0.5ms) so throughput of 2k rps. After the patch it will have the latency of 4ms so throughput of 250 rps.

There is also a problem latency-wise. The first workload may be analytics, for which latency is unimportant. The second workload may be latency-sensitive, for which low latency in high percentiles is important. But it will be severely affected by the first workload.

> 
> > > If some of the active reads are hits, then they will complete and make room for another hit (or miss, taking the hit rate below 1% again). The cache concurrency is 1, it doesn't need concurrency to run at full speed.
> > 
> > 
> > Yes, but if there is more than 100 active misses then hits will experience an increase in latency due to queueing. How often depends on the concurrency of hits. In particular, if there is one process which constantly hits, and there are more than 100 active misses, then that process will always experience the latency of the queued and active misses. Whereas before the patch it wouldn't. This will reduce the throughput of such a process (assuming there is spare CPU).
> 
> Having 100 active misses AND spare CPU requires that misses have very long disk latency (my spinning disk example). With SSD, latency is O(task quota) so 0.5ms. Since we are able to serve 20-50 kops/sec/core (usually less due to hyperthreading and due to misses being more expensive), Little's Law gives us effective concurrency of 10-25 (and again usually less). Above that concurrency we are CPU bound.

Reads are composed of many requests, with dependencies between them. We first do I/O for the index, then for the data file, which has to queue again in our IO queues. Reads may also touch many sstables, requiring many I/Os to complete, which adds to the load on the disk and latency of other reads. So even if a single I/O has low latency, the whole read may have higher latency. You need to take into account the latency of the whole read in those calculations for the average number of active reads.

It could be that a single scanning read puts so much I/O pressure on the disk (it can issue pretty much arbitrary amount of I/O), that a queue will form behind it equal in size to the external read concurrency.

What we could do to avoid this problem is to keep the old queue for the disk as it was, and add another one for the cache, with the concurrency of 1. When the cache needs to go to disk, it releases the cache queue permit back. So each read is either limited by the cache queue, or by the sstable queue. I think the goal of this PR is still achieved, which is to have inactive reads queued somewhere and expirable. If execution stage's queue could expire items, it could be that queue.

> 
> > > In practical implementations, if the cache hit rate is below 1% then we will max out on CPU as miss processing is much more expensive than hit processing. We'll only have spare CPU if we're talking to an extermely slow disk (consider an array of 100+ spinning disks connected to one shard, I think that's the only case we regress with this patch).
> > 
> > 
> > If we max out on the CPU in a full-miss case, it means we can't fully utilize the disk. That sounds like a bug in the design or an unbalanced system. Disk's bandwidth is partitioned among many shards, and also severely reduced by ongoing write IOs. I recall seeing IO-bound workloads in the field.
> 
> If a workload is I/O bound, then running requests through the cache outside the semaphore won't help it. Yes, that request will complete quickly if it hits, but the next one that misses will just have to wait longer for the next miss to complete.

Even if the hit workload is independent of the miss workload?

> 
> > > > Perhaps we should use separate concurrency semaphores for different resources (disk vs CPU) in order to achieve better utilization.
> > > > If we had 2 disks, it would also make sense to use different semaphore for each disk.
> > > 
> > > 
> > > That only works if we know in advance which disk a query is going to use.
> > > We'll have a similar problem with workload prioritization and users, but in that case, the user is known in advance.
> > > > This is orthogonal to memory accounting which prevents OOM.

denesb:
> @denesb are you implying that logical ordering of commits is more important than syntax highlighting?!

Indeed I am. For me, syntax highlighting is a nice-to-have, but the logical commit ordering is pretty essential.

denesb:
> What we could do to avoid this problem is to keep the old queue for the disk as it was, and add another one for the cache, with the concurrency of 1. When the cache needs to go to disk, it releases the cache queue permit back. So each read is either limited by the cache queue, or by the sstable queue. I think the goal of this PR is still achieved, which is to have inactive reads queued somewhere and expirable. If execution stage's queue could expire items, it could be that queue.

I think making the execution stage's queue able to expire items makes sense regardless of how we go about making sure hits don't suffer. Currently, if I'm not mistaken, only the reader concurrency semaphore times out reads on the replica side. The more places we check for expiry the less chance we are doing waste work.

gleb-cloudius:
On Mon, Jul 29, 2019 at 11:18:48PM -0700, Botond Dénes wrote:
> > What we could do to avoid this problem is to keep the old queue for the disk as it was, and add another one for the cache, with the concurrency of 1. When the cache needs to go to disk, it releases the cache queue permit back. So each read is either limited by the cache queue, or by the sstable queue. I think the goal of this PR is still achieved, which is to have inactive reads queued somewhere and expirable. If execution stage's queue could expire items, it could be that queue.
> 
> I think making the execution stage's queue able to expire items makes sense regardless of how we go about making sure hits don't suffer. Currently, if I'm not mistaken, only the reader concurrency semaphore times out reads on the replica side. The more places we check for expiry the less chance we are doing waste work.
> 
The execution stage queue is bounded after previous Avi's patch. SMP
queue (which is before execution stage and before reader concurrency
semaphore) is problematic one now.

--
			Gleb.

denesb:
> On Mon, Jul 29, 2019 at 11:18:48PM -0700, Botond Dénes wrote: > What we could do to avoid this problem is to keep the old queue for the disk as it was, and add another one for the cache, with the concurrency of 1. When the cache needs to go to disk, it releases the cache queue permit back. So each read is either limited by the cache queue, or by the sstable queue. I think the goal of this PR is still achieved, which is to have inactive reads queued somewhere and expirable. If execution stage's queue could expire items, it could be that queue. I think making the execution stage's queue able to expire items makes sense regardless of how we go about making sure hits don't suffer. Currently, if I'm not mistaken, only the reader concurrency semaphore times out reads on the replica side. The more places we check for expiry the less chance we are doing waste work.
> The execution stage queue is bounded after previous Avi's patch. SMP queue (which is before execution stage and before reader concurrency semaphore) is problematic one now.

Wasn't that addressed by https://github.com/scylladb/seastar/commit/7f705a76aa265d34d3abf4c8a51fc025af078195?

gleb-cloudius:
On Tue, Jul 30, 2019 at 12:48:41AM -0700, Botond Dénes wrote:
> > On Mon, Jul 29, 2019 at 11:18:48PM -0700, Botond Dénes wrote: > What we could do to avoid this problem is to keep the old queue for the disk as it was, and add another one for the cache, with the concurrency of 1. When the cache needs to go to disk, it releases the cache queue permit back. So each read is either limited by the cache queue, or by the sstable queue. I think the goal of this PR is still achieved, which is to have inactive reads queued somewhere and expirable. If execution stage's queue could expire items, it could be that queue. I think making the execution stage's queue able to expire items makes sense regardless of how we go about making sure hits don't suffer. Currently, if I'm not mistaken, only the reader concurrency semaphore times out reads on the replica side. The more places we check for expiry the less chance we are doing waste work.
> > The execution stage queue is bounded after previous Avi's patch. SMP queue (which is before execution stage and before reader concurrency semaphore) is problematic one now.
> 
> Wasn't that addressed by https://github.com/scylladb/seastar/commit/7f705a76aa265d34d3abf4c8a51fc025af078195?
> 
That one only makes things worse. The point of the commit it to protect
receiving shard, not sending one. Without it a shard may send requests
to another one until receiver chokes and stop polling at which point
sending shard will accumulate a queue. With the patch the sender shard
will start to accumulate the queue much earlier.

--
			Gleb.

avikivity:
> > @denesb are you implying that logical ordering of commits is more important than syntax highlighting?!
> 
> Indeed I am. For me, syntax highlighting is a nice-to-have, but the logical commit ordering is pretty essential.

It was a joke, of course syntax highlighing is by far less important than viewing the commits in the right order.

avikivity:
@tgrabiec I think you're right and this patch is penalizing independent workloads for the same user that have different hit/miss characteristics.

What we can do is require that such independent workloads use workload prioritization (it makes sense anyway, because the relative throughput is at the mercy of the concurrency of each workload) but that doesn't help open-source users.

dorlaor:
On Tue, Jul 30, 2019, 11:55 Avi Kivity <notifications@github.com> wrote:

> @denesb <https://github.com/denesb> are you implying that logical
> ordering of commits is more important than syntax highlighting?!
>
> Indeed I am. For me, syntax highlighting is a nice-to-have, but the
> logical commit ordering is pretty essential.
>
> It was a joke, of course syntax highlighing is by far less important than
> viewing the commits in the right order.
>
Can we syntax highlight the jokes?

dekimir:
> We should tag Dejan (@dekimir) as a reviewer in every single github PR

Or we could, you know, um, try to keep PR discussions on topic?

tgrabiec:
On Tue, Jul 30, 2019 at 11:52 AM Avi Kivity <notifications@github.com>
wrote:

> @tgrabiec <https://github.com/tgrabiec> I think you're right and this
> patch is penalizing independent workloads for the same user that have
> different hit/miss characteristics.
>
> What we can do is require that such independent workloads use workload
> prioritization (it makes sense anyway, because the relative throughput is
> at the mercy of the concurrency of each workload) but that doesn't help
> open-source users.
>

What about the following suggestion:


> What we could do to avoid this problem is to keep the old queue for the
> disk as it was, and add another one for the cache, with the concurrency of
> 1. When the cache needs to go to disk, it releases the cache queue permit
> back. So each read is either limited by the cache queue, or by the sstable
> queue. I think the goal of this PR is still achieved, which is to have
> inactive reads queued somewhere and expirable. If execution stage's queue
> could expire items, it could be that queue.

avikivity:
> What about the following suggestion:
> […](#)
> What we could do to avoid this problem is to keep the old queue for the disk as it was, and add another one for the cache, with the concurrency of 1. When the cache needs to go to disk, it releases the cache queue permit back. So each read is either limited by the cache queue, or by the sstable queue. I think the goal of this PR is still achieved, which is to have inactive reads queued somewhere and expirable. If execution stage's queue could expire items, it could be that queue.

The cache reader will have consumed some resources by filling buffers, so it cannot just go into the disk inactive queue. If it does, then we're exposed to running out of resources by the cache serving partial hits and moving them out of the way.

We could make inactive reads also consume resources, but that would quickly limit the inactive queue size and would cause requests to bounce.

tgrabiec:
On Mon, Aug 5, 2019 at 10:58 AM Avi Kivity <notifications@github.com> wrote:

> What about the following suggestion:
> … <#m_781992061380929358_>
> What we could do to avoid this problem is to keep the old queue for the
> disk as it was, and add another one for the cache, with the concurrency of
> 1. When the cache needs to go to disk, it releases the cache queue permit
> back. So each read is either limited by the cache queue, or by the sstable
> queue. I think the goal of this PR is still achieved, which is to have
> inactive reads queued somewhere and expirable. If execution stage's queue
> could expire items, it could be that queue.
>
> The cache reader will have consumed some resources by filling buffers, so
> it cannot just go into the disk inactive queue. If it does, then we're
> exposed to running out of resources by the cache serving partial hits and
> moving them out of the way.
>

We can make sure that reader's buffer is empty when it goes to sstables by
yielding from fill_buffer() to let the user consume what is ready. Then the
resource usage of inactive readers waiting for the sstable layer permit
would be the same as if there was no cache.

> We could make inactive reads also consume resources, but that would
> quickly limit the inactive queue size and would cause requests to bounce.
>
Why would this limit the inactive queue size?

>

avikivity:
> On Mon, Aug 5, 2019 at 10:58 AM Avi Kivity ***@***.***> wrote: What about the following suggestion: … <#m_781992061380929358_> What we could do to avoid this problem is to keep the old queue for the disk as it was, and add another one for the cache, with the concurrency of 1. When the cache needs to go to disk, it releases the cache queue permit back. So each read is either limited by the cache queue, or by the sstable queue. I think the goal of this PR is still achieved, which is to have inactive reads queued somewhere and expirable. If execution stage's queue could expire items, it could be that queue. The cache reader will have consumed some resources by filling buffers, so it cannot just go into the disk inactive queue. If it does, then we're exposed to running out of resources by the cache serving partial hits and moving them out of the way.
> We can make sure that reader's buffer is empty when it goes to sstables by yielding from fill_buffer() to let the user consume what is ready. Then the resource usage of inactive readers waiting for the sstable layer permit would be the same as if there was no cache.

Let's say the caller is filling a result set and waiting for either the reader to end or for the page to fill up. Then yes our reader isn't consuming resources, but active operations in the system tied to the reader are.

> We could make inactive reads also consume resources, but that would quickly limit the inactive queue size and would cause requests to bounce.
> Why would this limit the inactive queue size?

Right now the inactive queue size is computed as 2% of shard memory divided by an estimate of the cost of an inactive read (10k, perhaps we should reduce this). If an inactive read can have partial results, and we start accounting for them, this can drop the inactive queue length by a factor of 50 (or even more, since individual rows can be multiple megabytes in size and we don't do partial rows).

nyh:
Guys, is this PR still relevant? Any chance we can decide if it should go in or should be closed?

avikivity:
It's relevant in the sense that we need something equivalent, but not relevant in that we're not going to merge this one.

PR #4763, opened 2019-07-27
Simplify types v3

espindola:
This is hopefully the last large refactoring on the way of UDF.                                                                                                                                                                               
                                                                                                                                                                                                                                              
In UDF we have to convert internal types to Lua and back. Currently                                                                                                                                                                           
almost all our types and hidden in types.cc and expose functionality                                                                                                                                                                          
via virtual functions. While it should be possible to add a                                                                                                                                                                                   
convert_{to|from}_lua virtual functions, that seems like a bad design.                                                                                                                                                                        
                                                                                                                                                                                                                                              
In compilers, the type definition is normally public and different                                                                                                                                                                            
passes know how to reason about each type. The alias analysis knows                                                                                                                                                                           
about int and floats, not the other way around.                                                                                                                                                                                               
                                                                                                                                                                                                                                              
This patch series is inspired by both the LLVM RTTI                                                                                                                                                                                           
(https://www.llvm.org/docs/HowToSetUpLLVMStyleRTTI.html) and                                                                                                                                                                                  
std::variant.                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                              
The series makes the types public, adds a visit function and converts                                                                                                                                                                         
the various virtual methods to just use visit. As a small example of                                                                                                                                                                          
why this is useful, it then moves a bit of cql3 and json specific                                                                                                                                                                             
logic out of types.cc and types.hh. In a similar way, the UDF code                                                                                                                                                                            
will be able to used visit to convert objects to Lua.                                                                                                                                                                                         
    
In comparison with the previous versions, this series doesn't require the intermediate step of converting void* to data_value& in a few member functions.

This version also has fewer double dispatches I a am fairly confident has all the tools for avoiding all double dispatches.

Sending it as is since I am going on vacation.


types.cc
```diff
@@ -2278,7 +2212,34 @@ shared_ptr<const abstract_type> abstract_type::underlying_type() const {
 }
 
 abstract_type::cql3_kind abstract_type::get_cql3_kind_impl() const {
-    assert(0 && "no kind for this type");
```
avikivity: This should be moved (one day) to cql3/.

types.cc
```diff
@@ -2254,9 +2188,50 @@ sstring abstract_type::quote_json_string(const sstring& s) {
     return json::value_to_quoted_string(s);
 }
 
+static sstring cql3_type_name_impl(const abstract_type& t) {
```
avikivity: This also should be moved to cql3/, not now.

types.cc
```diff
@@ -314,9 +314,6 @@ struct string_type_impl : public concrete_type<sstring> {
     virtual bool is_byte_order_equal() const override {
         return true;
     }
-    virtual bool is_byte_order_comparable() const override {
-        return true;
-    }
     virtual bool is_string() const override {
```
avikivity: A lot of functions like this one can be converted to bool : 1 fields stored in the base class. Types are immutable so we don't need to recalculate this over and over.

types.cc
```diff
@@ -2011,6 +2011,13 @@ struct is_byte_order_equal_visitor {
 
 bool abstract_type::is_byte_order_equal() const { return visit(*this, is_byte_order_equal_visitor{}); }
 
+bool abstract_type::equal(bytes_view v1, bytes_view v2) const {
+    if (is_byte_order_equal()) {
+        return compare_unsigned(v1, v2) == 0;
```
avikivity: This adds a virtual call, since in final types this would be inlined, and here it cannot be.

types.cc
```diff
@@ -2013,6 +1976,12 @@ struct compare_visitor {
 }
 
 int32_t abstract_type::compare(bytes_view v1, bytes_view v2) const {
+    if (v1.empty()) {
+        return v2.empty() ? 0 : -1;
+    }
+    if (v2.empty()) {
+        return 1;
+    }
     return visit(*this, compare_visitor{v1, v2});
```
avikivity: This slows down types that don't have special empty values, like text and blob.

avikivity: I think this commit should just be dropped.

types.cc
```diff
@@ -3333,24 +3158,176 @@ static bytes concat_fields(const std::vector<bytes>& fields, const std::vector<i
...
+        bool v;
+        if (s.empty() || s_lower == "false") {
+            v = false;
+        } else if (s_lower == "true") {
+            v = true;
+        } else {
+            throw marshal_exception(format("unable to make boolean from '{}'", s));
+        }
+        return t.make_value(v).serialize();
+    }
```
avikivity: Doesn't this (and similar) add a virtual call?  The original knows the type, but make_value() forgets the type and now we have a virtual call, unless the compiler is clever enough to inline everything.

types.cc
```diff
@@ -2185,6 +2076,125 @@ void tuple_type_impl::validate(bytes_view v, cql_serialization_format sf) const
...
+    void operator()(const map_type_impl& t) { validate_aux(t, v, sf); }
+    void operator()(const set_type_impl& t) { validate_aux(t, v, sf); }
+    void operator()(const list_type_impl& t) { validate_aux(t, v, sf); }
+    void operator()(const tuple_type_impl& t) { validate_aux(t, v, sf); }
+};
+}
+
+void abstract_type::validate(bytes_view v, cql_serialization_format sf) const {
+    visit(*this, validate_visitor{v, sf});
```
avikivity: One thing we can do later, is to have a validate function for each cql_serialization_format (essentially making it a template parameter) and store the pointer to the function in the connection. This eliminates moving the format around.

avikivity:
@bhalevy can you adopt this patchset? Most of my comments are for the future, but a few point to regressions that need to be addressed.

bhalevy:
> @bhalevy can you adopt this patchset? Most of my comments are for the future, but a few point to regressions that need to be addressed.

ok, will give a stab

kbr-:
This is great and makes the code much more maintainable. Got some questions/suggestions though:
- I'm curious about `_kind` + `get_kind()` + the implementation of `visit`. Wouldn't it be possible (and desirable) to get rid of the `abstract_type` hierarchy entirely and instead use something like
`using abstract_type = std::variant<concrete types...>`? Now we're kind of reinventing the wheel. I don't know about the performance, whether one solution (hand-crafted enum + switch) is better than the other (`std::variant` + `std::visit`). However, when using `std::variant`, we could perform a dispatch on multiple variants at once (`std::visit(visitor, type1, type2)`, which would, for example, allow more elegant implementation of `is_compatible_with`. Now using `get_kind()` in the implementation of `is_compatible_with` seems like a workaround for the lack of power of the current solution.
- `_name` depends only on the concrete type and doesn't change in most cases. Therefore it could (should?) be changed into a visitor-powered function too. And only in the types where `_name` is not known at compile type should it be kept in the type (so, UDTs).
- Same for `value_length_if_fixed`. Most of the types, if not all, know it during compilation time (`data::type_info::make_variable_size()` is also a "constant"). In general, when I see constructors like `long_type_impl::long_type_impl() : integer_type_impl{kind::long_kind, long_type_name, 8} {}` it seems that the two constants should be refactored into a visitor.
- those `get_kind()`s in function.cc and operation.cc could probably be refactored to use a visitor.

types.cc
espindola: In fact, most of them can be a bitfield in reversed_type_impl since that is the only case where it is not known by just looking a _kind.

Can this be in a followup series?

types.cc
espindola: This one is harder, since we cache the result. I will mark as resolved since it is for another patch series.

espindola:
> This is great and makes the code much more maintainable. Got some questions/suggestions though:
> 
>     * I'm curious about `_kind` + `get_kind()` + the implementation of `visit`. Wouldn't it be possible (and desirable) to get rid of the `abstract_type` hierarchy entirely and instead use something like
>       `using abstract_type = std::variant<concrete types...>`? Now we're kind of reinventing the wheel. I don't know about the performance, whether one solution (hand-crafted enum + switch) is better than the other (`std::variant` + `std::visit`). However, when using `std::variant`, we could perform a dispatch on multiple variants at once (`std::visit(visitor, type1, type2)`, which would, for example, allow more elegant implementation of `is_compatible_with`. Now using `get_kind()` in the implementation of `is_compatible_with` seems like a workaround for the lack of power of the current solution.

We do use inheritance, so I think the current solution is a bit more direct that a std::variant.

>     * `_name` depends only on the concrete type and doesn't change in most cases. Therefore it could (should?) be changed into a visitor-powered function too. And only in the types where `_name` is not known at compile type should it be kept in the type (so, UDTs).
>       EDIT: or maybe it should be left as it is, changing it to use a visitor would probably impact performance negatively...

It might be possible, but this patch series is already fairly large. I would like to keep it just about exposing simple concrete types in concrete_types.hh.

>     * those `get_kind()`s in function.cc and operation.cc could probably be refactored to use a visitor.

Why? For simple predicates just checking the enum is simpler.

types.cc
espindola: fixed

types.cc
espindola: fixed in the new version.

types.cc
espindola: dropped it.

types.cc
espindola: resolving since it is for another series.

espindola:
Don't merge it just yet. asan found something.

espindola:
Updated again. This version passes all tests in dev and debug mode. The only difference is making sure that decompose uses the "this" type, not the type of the data_value. They should be the same, but avoiding changing that means we don't have to fix all cases just yet.

espindola:
test @bhalevy

avikivity:
@espindola what was changed? so I can focus my re-review.

espindola:
The double dispatches you pointed out and creation of temporary data_values
are gone.

Other than that I removed some of the cleanup and bug fixes patches by
making the new code bug by bug compatibility when there are two abstract
types present.


Cheers,
Rafael

>

types.cc
avikivity: Yes of course. This one is long enough.

types.cc
```diff
@@ -2165,6 +2150,24 @@ struct is_byte_order_equal_visitor {
...
+struct is_byte_order_comparable_visitor {
+    bool operator()(const abstract_type&) { return false; }
+    bool operator()(const reversed_type_impl& r) { return r.underlying_type()->is_byte_order_comparable(); }
+    bool operator()(const string_type_impl&) { return true; }
+    bool operator()(const bytes_type_impl&) { return true; }
+    bool operator()(const date_type_impl&) { return true; }
+    bool operator()(const inet_addr_type_impl&) { return true; }
+    bool operator()(const duration_type_impl&) { return true; }
+    // FIXME: origin returns false for list.  Why?
+    bool operator()(const set_type_impl& s) { return s.get_elements_type()->is_byte_order_comparable(); }
```
avikivity: I think it's because timeuuids are not byte order comparable.

avikivity: Oh, you just moved the comment.

types.cc
```diff
@@ -3234,6 +3234,15 @@ int32_t abstract_type::compare(bytes_view v1, bytes_view v2) const {
     return visit(*this, compare_visitor{v1, v2});
 }
 
+bool abstract_type::equal(bytes_view v1, bytes_view v2) const {
+    return ::visit(*this, [&](const auto& t) {
+        if (is_byte_order_equal_visitor{}(t)) {
+            return compare_unsigned(v1, v2) == 0;
```
avikivity: This can be optimized later. There's no reason to make two runtime checks here.

avikivity:
Thanks @github for randomizing the commit order in review.

avikivity:
Being a little evil, I benchmarked with perf_simple_query --smp 1. Post-series is faster by 1-2%.

avikivity:
Queued as 20aed3398de28a38c79287145af6d79a9861b6eb

kbr-:
> Why? For simple predicates just checking the enum is simpler.

Maybe it's just me philosophising too much, but we check the enum to make a runtime dispatch on the type --- and that's exactly what the `visit` interface is for. The enum, IMHO, is just an implementation detail for how to make the dispatch. We could even remove get_kind() from the public interface. And if for some reason we decided to switch to `std::variant` or anything else in the future, only `visit` would need to be changed. Now we have at least two "correct" ways to make a dispatch: `visit` or manual switch on `get_kind`.

One downside of using `visit` is the boilerplate. It can however be reduced greatly using `seastar::make_visitor`, and when P1260R0 or similar proposal gets merged to the standard, using the dedicated pattern matching syntax.

> We do use inheritance, so I think the current solution is a bit more direct that a std::variant.

Could you elaborate on being "more direct"? My idea was to get rid of inheritance in types completely and switch to composition whenever we want to share code, or inheriting from small abstract pure classes created for specific purposes when we want common interfaces.

espindola:

> +namespace {
> +struct is_byte_order_comparable_visitor {
> +    bool operator()(const abstract_type&) { return false; }
> +    bool operator()(const reversed_type_impl& r) { return r.underlying_type()->is_byte_order_comparable(); }
> +    bool operator()(const string_type_impl&) { return true; }
> +    bool operator()(const bytes_type_impl&) { return true; }
> +    bool operator()(const date_type_impl&) { return true; }
> +    bool operator()(const inet_addr_type_impl&) { return true; }
> +    bool operator()(const duration_type_impl&) { return true; }
> +    // FIXME: origin returns false for list.  Why?
> +    bool operator()(const set_type_impl& s) { return s.get_elements_type()->is_byte_order_comparable(); }
>
> I think it's because timeuuids are not byte order comparable.

Then set_type_impl is wrong, no?

> +bool abstract_type::equal(bytes_view v1, bytes_view v2) const {
> +    return ::visit(*this, [&](const auto& t) {
> +        if (is_byte_order_equal_visitor{}(t)) {
> +            return compare_unsigned(v1, v2) == 0;
>
> This can be optimized later. There's no reason to make two runtime checks here.

is_byte_order_equal_visitor is not a runtime check. In here t has the
same dynamic and static type.

Cheers,
Rafael

espindola:
Kamil Braun <notifications@github.com> writes:

>> Why? For simple predicates just checking the enum is simpler.
>
> Maybe it's just me philosophising too much, but we check the enum to make a runtime dispatch on the type --- and that's exactly what the `visit` interface is for. The enum, IMHO, is just an implementation detail for how to make the dispatch. We could even remove get_kind() from the public interface. And if for some reason we decided to switch to `std::variant` or anything else in the future, only `visit` would need to be changed. Now we have at least two "correct" ways to make a dispatch: `visit` or manual switch on `get_kind`.

Sometimes we don't really want to dispatch anything:

bool is_foo() { return get_kind() == foo; }


> One downside of using `visit` is the boilerplate. It can however be reduced greatly using `seastar::make_visitor`, and when P1260R0 or similar proposal gets merged to the standard, using the dedicated pattern matching syntax.

The first version actually used make_visitor, but templated lambdas are c++2x.

>> We do use inheritance, so I think the current solution is a bit more direct that a std::variant.
>
> Could you elaborate on being "more direct"? My idea was to get rid of inheritance in types completely and switch to composition whenever we want to share code, or inheriting from small abstract pure classes created for specific purposes when we want common interfaces.

In general I agree with having less inheritance, but that would be an
independent change.

Cheers,
Rafael

PR #4779, opened 2019-07-30
Improve compaction logging

### LINKED ISSUE #4778: Improve compaction log printouts
We need to include more information in the compaction printouts like:
- a unique compaction identifier that will help match start and finish pairs.
- sstables run identifier

bhalevy:
In this series:
1. Generate and a unique random identifier per compaction and print it on report_start/finish
2. Cleanup printout of sstable lists
3. Print also the run_identofier for each sstable to the log

Fixes #4778

Test: unit (dev), compaction_test.py:TestCompaction_with_SizeTieredCompactionStrategy.sstable_deletion_test (dev)

### DIFF
--- a/sstables/compaction.cc
+++ b/sstables/compaction.cc
@@ -390,7 +390,8 @@ private:
             _ancestors.push_back(sst->generation());
             _info->start_size += sst->bytes_on_disk();
             _info->total_partitions += sst->get_estimated_key_count();
-            formatted_msg += format("{}{}:level={:d}", delim, sst->component_basename(component_type::Data), sst->get_sstable_level());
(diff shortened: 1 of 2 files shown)


avikivity:
> Not sure we should have another identifier which could add more confusion. What about using the output run identifier as the compaction id?

Agree that the output run identifier should be sufficient. Of course there may be multiple output runs (LCS, or using @denesb TWCS splitter)

avikivity:
But what about moving all of these logs to the compaction_history table? Compaction is too noisy for the log.

bhalevy:
> But what about moving all of these logs to the compaction_history table? Compaction is too noisy for the log.

I'm for it.  Let's withdraw this pr then and work on moving the logs with the additional information into the compaction_history table.

PR #4787, opened 2019-08-01
sstables: ka/la: reader: Make sure push_ready_fragments() does not mi…

### LINKED ISSUE #4786: KA/LA sstable reader doesn't emit partition_end and partition_start
mp_row_consumer_k_l::push_ready_fragments has a bug that when _ready is set it will only call mp_row_consumer_k_l::push_ready_fragments_with_ready_set and no mp_row_consumer_k_l::push_ready_fragments_out_of_range.

That leads to sstable_mutation_reader::fill_buffer reading the content of the next partition without closing the previous one.

haaawk:
…ss to emit partition_end

Currently, if there is a fragment in _ready and _out_of_range was set
after row end was consumer, push_ready_fragments() would return
without emitting partition_end.

This is problematic once we make consume_row_start() emit
partiton_start directly, because we will want to assume that all
fragments for the previous partition are emitted by then. If they're
not, then we'd emit partition_start before partition_end for the
previous partition. The fix is to make sure that
push_ready_fragments() emits everything.

Fixes #4786

(cherry picked from commit 9b8ac5ecbc66de574b8ce885de809cbe08d7c16c)

### DIFF
--- a/sstables/mp_row_consumer.hh
+++ b/sstables/mp_row_consumer.hh
@@ -702,9 +702,12 @@ public:
     // Sets streamed_mutation::_end_of_range when there are no more fragments for the query range.
     // Returns information whether the parser should continue to parse more
     // input and produce more fragments or we have collected enough and should yield.
+    // Returns proceed:yes only when all pending fragments have been pushed.
     proceed push_ready_fragments() {
         if (_ready) {
-            return push_ready_fragments_with_ready_set();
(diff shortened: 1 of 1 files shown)


haaawk:
@avikivity @glommer @tgrabiec Please have a look

avikivity:
Queued as d77aaada86885b1a08b90a9a6cda3379cd0735a9

PR #4788, opened 2019-08-01
sstables: ka/la: reader: Make sure push_ready_fragments() does not mi…

### LINKED ISSUE #4786: KA/LA sstable reader doesn't emit partition_end and partition_start

haaawk:
…ss to emit partition_end

Currently, if there is a fragment in _ready and _out_of_range was set
after row end was consumer, push_ready_fragments() would return
without emitting partition_end.

This is problematic once we make consume_row_start() emit
partiton_start directly, because we will want to assume that all
fragments for the previous partition are emitted by then. If they're
not, then we'd emit partition_start before partition_end for the
previous partition. The fix is to make sure that
push_ready_fragments() emits everything.

Fixes #4786

(cherry picked from commit 9b8ac5ecbc66de574b8ce885de809cbe08d7c16c)

### DIFF
--- a/sstables/mp_row_consumer.hh
+++ b/sstables/mp_row_consumer.hh
@@ -703,9 +703,12 @@ public:
     // Sets streamed_mutation::_end_of_range when there are no more fragments for the query range.
     // Returns information whether the parser should continue to parse more
     // input and produce more fragments or we have collected enough and should yield.
+    // Returns proceed:yes only when all pending fragments have been pushed.
     proceed push_ready_fragments() {
         if (_ready) {
-            return push_ready_fragments_with_ready_set();
+            if (push_ready_fragments_with_ready_set() == proceed::no) {
+                return proceed::no;
+            }
         }
 
         if (_out_of_range) {


haaawk:
@avikivity @glommer @tgrabiec Please have a look

avikivity:
Queued as b1e2842c8c450edd2ed6543f052a625e8b5cd802

tgrabiec:
One problematic thing around this area is that the result of `push_ready_fragment()` is actually ignored. I think this is currently not a problem because whenever it returns `proceed::no` (buffer full), we will also break inside `sstable_mutation_reader::fill_buffer()` because the check is duplicated:

```
                    _consumer.push_ready_fragments();
                    if (is_buffer_full() || _partition_finished || _end_of_stream) {
                        return make_ready_future<>();
                    }
```

But it seems fragile to rely on that.

haaawk:
@tgrabiec I agree with your deduction. That's my understanding too. Moreover, it is possible that _mutation_end is set to true but _partition_finished is not yet set to true and then it is theoretically possible that we will continue reading.

avikivity:
Committed as b1e2842c8c450edd2ed6543f052a625e8b5cd802

Issue #4789, opened 2019-08-01
Receiving side of streaming may abort or produce incorrect sstables when failure happens on the sender side


tgrabiec:
Scylla version (or git commit hash): 3.0

We've seen a node abort inside sstable writer:
```
sstables/sstables.cc:1748: void sstables::seal_summary(sstables::summary&, std::experimental::fundamentals_v1::optional<sstables::key>&&, std::experimental::fundamentals_v1::optional<sstables::key>&&, const sstables::index_sampling_state&): Assertion `first_key' failed. 
```

The cause was that active partition was not closed in the input mutation fragment stream (missing `partition_end` fragment). Because that was the only partition, the sstable writer appeared to have no partitions, which triggers the asserts.

If there were more partitions, the writer would not fail, but sstable index and partition range metadata would be incorrect, which would manifest as errors during read, or data apparently missing. Recompacting such sstable fixes the problem without loosing any data.

The writer belonged to the streaming receiver and it looked like the generating reader reached end-of-stream after emitting `clustering_row`. This is incorrect since there should be a `partition_end` fragment before the stream ends.

It does so when the mutation fragment generating functor returns a disengaged optional. The streaming functor does so when it gets a disengaged optional from the RPC stream's source:

```
auto get_next_mutation_fragment = [source, plan_id, from, s] () mutable {
                        return source().then([plan_id, from, s] (stdx::optional<std::tuple<frozen_mutation_fragment>> fmf_opt) mutable {
                            if (fmf_opt) {
                                frozen_mutation_fragment& fmf = std::get<0>(fmf_opt.value());
                                auto sz = fmf.representation().size();
                                auto mf = fmf.unfreeze(*s);
                                streaming::get_local_stream_manager().update_progress(plan_id, from.addr, progress_info::direction::IN, sz);
                                return make_ready_future<mutation_fragment_opt>(std::move(mf));
                            } else {
                                return make_ready_future<mutation_fragment_opt>();
                            }
                        });
                    };
```

The matching sink on the sender side looks like this:

```
auto sink_op = [sink, si, got_error_from_peer] () mutable -> future<> {
            return do_with(std::move(sink), [si, got_error_from_peer] (rpc::sink<frozen_mutation_fragment>& sink) {
                return repeat([&sink, si, got_error_from_peer] () mutable {
                    return si->reader(db::no_timeout).then([&sink, si, s = si->reader.schema(), got_error_from_peer] (mutation_fragment_opt mf) mutable {
[...]

PR #4792, opened 2019-08-01
streaming: Send error code from the sender to receiver

asias:
In case of error on the sender side, the sender does not propagate the
error to the receiver. The sender will close the stream. As a result,
the receiver will get nullopt from the source in
get_next_mutation_fragment and pass mutation_fragment_opt with no value
to the generating_reader. In turn, the generating_reader generates end
of stream. However, the last element that the generating_reader has
generated can be any type of mutation_fragment. This makes the sstable
that consumes the generating_reader violates the mutation_fragment
stream rule.

To fix, we need to propagate the error. However RPC streaming does not
support propagate the error in the framework. User has to send an error
code explicitly.

Fixes: #4789

### DIFF
--- a/streaming/stream_session.cc
+++ b/streaming/stream_session.cc
@@ -161,7 +161,7 @@ void stream_session::init_messaging_service_handler() {
             });
         });
     });
-    ms().register_stream_mutation_fragments([] (const rpc::client_info& cinfo, UUID plan_id, UUID schema_id, UUID cf_id, uint64_t estimated_partitions, rpc::optional<stream_reason> reason_opt, rpc::source<frozen_mutation_fragment> source) {
+    ms().register_stream_mutation_fragments([] (const rpc::client_info& cinfo, UUID plan_id, UUID schema_id, UUID cf_id, uint64_t estimated_partitions, rpc::optional<stream_reason> reason_opt, rpc::source<frozen_mutation_fragment, rpc::optional<stream_mutation_fragments_cmd>> source) {
         auto from = netw::messaging_service::get_source(cinfo);
         auto reason = reason_opt ? *reason_opt: stream_reason::unspecified;
         sslog.trace("Got stream_mutation_fragments from {} reason {}", from, int(reason));
@@ -173,15 +173,41 @@ void stream_session::init_messaging_service_handler() {
         return with_scheduling_group(service::get_local_storage_service().db().local().get_streaming_scheduling_group(), [from, estimated_partitions, plan_id, schema_id, &cf, source, reason] () mutable {
                 return service::get_schema_for_write(schema_id, from).then([from, estimated_partitions, plan_id, schema_id, &cf, source, reason] (schema_ptr s) mutable {
                     auto sink = ms().make_sink_for_stream_mutation_fragments(source);
-                    auto get_next_mutation_fragment = [source, plan_id, from, s] () mutable {
-                        return source().then([plan_id, from, s] (std::optional<std::tuple<frozen_mutation_fragment>> fmf_opt) mutable {
-                            if (fmf_opt) {
-                                frozen_mutation_fragment& fmf = std::get<0>(fmf_opt.value());
+                    struct stream_mutation_fragments_cmd_status {
+                        bool got_cmd = false;
+                        bool got_end_of_stream = false;
+                    };
+                    auto cmd_status = make_lw_shared<stream_mutation_fragments_cmd_status>();
+                    auto get_next_mutation_fragment = [source, plan_id, from, s, cmd_status] () mutable {
+                        return source().then([plan_id, from, s, cmd_status] (std::optional<std::tuple<frozen_mutation_fragment, rpc::optional<stream_mutation_fragments_cmd>>> opt) mutable {
+                            if (opt) {
+                                auto cmd = std::get<1>(*opt);
+                                if (cmd) {
+                                    cmd_status->got_cmd = true;
+                                    switch (*cmd) {
+                                    case stream_mutation_fragments_cmd::mutation_fragment_data:
+                                        break;
+                                    case stream_mutation_fragments_cmd::error:
+                                        return make_exception_future<mutation_fragment_opt>(std::runtime_error("Sender failed"));
+                                    case stream_mutation_fragments_cmd::end_of_stream:
+                                        cmd_status->got_end_of_stream = true;
+                                        return make_ready_future<mutation_fragment_opt>();
+                                    default:
+                                        return make_exception_future<mutation_fragment_opt>(std::runtime_error("Sender sent wrong cmd"));
+                                    }
+                                }
+                                frozen_mutation_fragment& fmf = std::get<0>(*opt);
                                 auto sz = fmf.representation().size();
                                 auto mf = fmf.unfreeze(*s);
                                 streaming::get_local_stream_manager().update_progress(plan_id, from.addr, progress_info::direction::IN, sz);
                                 return make_ready_future<mutation_fragment_opt>(std::move(mf));
                             } else {
+                                // If the sender has sent stream_mutation_fragments_cmd it means it is
+                                // a node that understands the new protocol. It must send end_of_stream
+                                // before close the stream.
+                                if (cmd_status->got_cmd && !cmd_status->got_end_of_stream) {
+                                    return make_exception_future<mutation_fragment_opt>(std::runtime_error("Sender did not sent end_of_stream"));
+                                }
                                 return make_ready_future<mutation_fragment_opt>();
                             }
                         });
--- a/streaming/stream_transfer_task.cc
+++ b/streaming/stream_transfer_task.cc
@@ -176,7 +176,7 @@ future<> send_mutations(lw_shared_ptr<send_info> si) {
 future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
   return si->estimate_partitions().then([si] (size_t estimated_partitions) {
     sslog.info("[Stream #{}] Start sending ks={}, cf={}, estimated_partitions={}, with new rpc streaming", si->plan_id, si->cf.schema()->ks_name(), si->cf.schema()->cf_name(), estimated_partitions);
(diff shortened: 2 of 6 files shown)


streaming/stream_transfer_task.cc
```diff
@@ -199,18 +199,25 @@ future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
...
+                            bool error = false;
+                            return sink(fmf, error).then([] { return stop_iteration::no; });
                         } else {
                             return make_ready_future<stop_iteration>(stop_iteration::yes);
                         }
                     });
+                }).handle_exception([&sink] (std::exception_ptr ep) mutable {
+                    // Notify the receiver the sender has failed
+                    bool error = true;
+                    return sink(frozen_mutation_fragment(bytes_ostream()), error).then([ep = std::move(ep)] () mutable {
```
tgrabiec: What if this `sink()` call fails, wouldn't we have the same problem?

streaming/stream_session.cc
```diff
@@ -174,9 +174,13 @@ void stream_session::init_messaging_service_handler() {
                 return service::get_schema_for_write(schema_id, from).then([from, estimated_partitions, plan_id, schema_id, &cf, source, reason] (schema_ptr s) mutable {
                     auto sink = ms().make_sink_for_stream_mutation_fragments(source);
                     auto get_next_mutation_fragment = [source, plan_id, from, s] () mutable {
-                        return source().then([plan_id, from, s] (std::optional<std::tuple<frozen_mutation_fragment>> fmf_opt) mutable {
-                            if (fmf_opt) {
-                                frozen_mutation_fragment& fmf = std::get<0>(fmf_opt.value());
+                        return source().then([plan_id, from, s] (std::optional<std::tuple<frozen_mutation_fragment, rpc::optional<bool>>> opt) mutable {
```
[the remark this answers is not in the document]
asias: Yes. The flag is rpc::optional. New node sends to old node: the flag will be ignored.  Old node sends to new node: the flag will be nullopt.

streaming/stream_transfer_task.cc
asias: If sink() call fails, the close() call will fail as well (@gleb-cloudius can you confirm?), so the end of stream will not be generated.

asias:
Updated:
- Introduce stream_mutation_fragments_cmd
- Send this cmd instead of a bool
- The new node must send a end_of_stream before close the stream. So that if the sending of the error has failed in handle_exception, when we call sink.close() in finally, the receiver fail the stream as well.

streaming/stream_transfer_task.cc
asias: I have a better version to fix this issue. See the updates below.

streaming/stream_transfer_task.cc
```diff
@@ -199,18 +199,27 @@ future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
...
-                            return sink(fmf).then([] { return stop_iteration::no; });
+                            return sink(fmf, stream_mutation_fragments_cmd::mutation_fragment_data).then([] { return stop_iteration::no; });
                         } else {
                             return make_ready_future<stop_iteration>(stop_iteration::yes);
                         }
                     });
+                }).then([&sink] () mutable {
+                    return sink(frozen_mutation_fragment(bytes_ostream()), stream_mutation_fragments_cmd::end_of_stream);
+                }).then([&sink] () mutable {
+                    return sink.flush();
```
tgrabiec: Why do we need to `flush()`?

streaming/stream_transfer_task.cc
```diff
@@ -199,18 +199,27 @@ future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
...
+                }).handle_exception([&sink] (std::exception_ptr ep) mutable {
+                    // Notify the receiver the sender has failed
+                    return sink(frozen_mutation_fragment(bytes_ostream()), stream_mutation_fragments_cmd::error).then([ep = std::move(ep)] () mutable {
```
tgrabiec: Sending `error` doesn't seem very useful. If we already sent end_of_stream, we can as well let the stream succeed. If we didn't send end_of_stream, the receiver will deduce there was an error.

streaming/stream_session.cc
```diff
@@ -217,7 +243,7 @@ void stream_session::init_messaging_service_handler() {
                             });
                         },
                         cf.stream_in_progress()
-                    ).then_wrapped([s, plan_id, from, sink, estimated_partitions] (future<uint64_t> f) mutable {
+                    ).then_wrapped([s, plan_id, from, sink, estimated_partitions, cmd_status] (future<uint64_t> f) mutable {
```
tgrabiec: Is this capture (and the one below) needed?

streaming/stream_transfer_task.cc
tgrabiec: I don't think that has to be the case:
 1) `sink()` may fail on memory allocation, but `close()` may succeed.
 2) doesn't receiver close the stream when connection is dropped regardless of what state the sender is in? \cc @gleb-cloudius

gleb-cloudius: If sink() fails (for any reason) it suppose to close the underlying connection which will be seen as an error by a receiver.

streaming/stream_transfer_task.cc
asias: Why deducing instead explicitly telling the peer there is an error? 

If we sent the error explicitly the peer can lean the error faster, for example:

Sender:
1. error happened
2. send error
3. send close

Skipping 2), the receiver only knows there is error when the stream is closed.  There can be long delay between 1) and 3), e.g., sudden lost of the node.

end_of_stream is sent lastly only if there is nothing to send. If the receiver receives end_of_stream, it means the stream is end successfully. It will not care about  whatever was sent latter.

streaming/stream_session.cc
asias: cmd_status must live until mutation_writer::distribute_reader_and_consume_on_shards is finished. The capture extends the lifetime.

streaming/stream_transfer_task.cc
asias: Make sure the sender does not cache and flush the sending queue because there is nothing to send. It is currently a no-op but it might change in the future.

streaming/stream_session.cc
gleb-cloudius: `do_with()` would make it clear what `cmd_status` lifetime should be.

streaming/stream_transfer_task.cc
tgrabiec: I still don't understand why we need it. What bad would happen without it?

streaming/stream_session.cc
tgrabiec: @asias Why does it have to live until distribute_reader_and_consume_on_shards is finished? The continuations which use it already capture it.

asias: Because the get_next_mutation_fragment which uses the cmd_status is called again and again within distribute_reader_and_consume_on_shards. 

A second thought: since get_next_mutation_fragment is stored inside multishard_writer and we have do_with to keep the lifetime of multishard_writer. I think capture cmd_status in the get_next_mutation_fragment alone is enough.

```
191 future<uint64_t> distribute_reader_and_consume_on_shards(schema_ptr s,
192     dht::i_partitioner& partitioner,
193     flat_mutation_reader producer,
194     std::function<future<> (flat_mutation_reader)> consumer,
195     utils::phased_barrier::operation&& op) {
196     return do_with(multishard_writer(std::move(s), partitioner, std::move(producer), std::move(consumer)), std::move(op), [] (multishard_writer& writer, utils::phased_barrier::operation&) {
197         return writer();
198     });
199 }
```

asias:
Updated: 
- Capture cmd_status in get_next_mutation_fragment only.

slivne:
@tgrabiec / @asias ping

streaming/stream_transfer_task.cc
gleb-cloudius: Data may be buffered indefinitely, so if sending new data depends on a reply from a receiver and the receiver replies only when all data is received (ping-pong pattern) there will be deadlock.

streaming/stream_transfer_task.cc
tgrabiec: We already deduce the error when the stream is closed. Sending `error` is only extra complexity and extra code.

sending close is the same as sending error. Sending close is not delayed more than sending error is.

streaming/stream_session.cc
tgrabiec: As long as `get_next_mutation_fragment` is called again and again, it must be alive, and as long as its alive, its captures are alive as well.

streaming/stream_transfer_task.cc
tgrabiec: @gleb-cloudius We don't wait for anything after `flush()`here, we call `close()` right after calling `flush()`. Is the flush still needed in such a case?

gleb-cloudius: No need to flush before close, only if we are going to reuse the connection to stream more data later.

streaming/stream_transfer_task.cc
gleb-cloudius: I do not understand what was wrong with the version that had an error only. Having explicit error notification allows passing assitional info about an error an (error code, or an error message). The patch does not do it, but it can be extended if needed.

tgrabiec: Passed additional info would be useful indeed. But in this patch the error message is just an enum value, so it cannot do that nor can be extended later. As such, it's a dead end and unnecessary.

glommer:
needs 3.0 backports once merged to master.

streaming/stream_transfer_task.cc
tgrabiec: It's not that of a big deal though. It may be useful for distinguishing sender failure from other possible (?) reasons of premature end of stream.

streaming/stream_transfer_task.cc
tgrabiec: So @asias, please remove the flush() as a follow up

asias: I wanted to make it future-proof if we use reuse the connection inside the rpc framework in the future. 
But since we call close(), even if we reuse the connection, the rpc should flush the data before reusing the connection.

I will remove the flush().

It is better to squash the change in this commit because we want to backport to 3.0 and 3.0 does not have flush in seastar. 

But it is too late, I saw it is merged in master.

So, let's commit the patch to remove flush and backport two of the commit in a single commit to 3.0.

PR #4809, opened 2019-08-07
streaming: Do not call rpc stream flush in send_mutation_fragments

asias:
The stream close() guarantees the data sent will be flushed. No need to
call the stream flush() since the stream is not reused.

Follow up fix for commit bac987e32ada (streaming: Send error code from
the sender to receiver).

Fixes: #4789


asias:
@slivne @tgrabiec Ping. I am waiting for this for the 3.0 backport requested by Glauber.

PR #4812, opened 2019-08-07
streaming: Move stream_mutation_fragments_cmd to a new file

asias:
Avoid including the lengthy stream_session.hh in messaging_service.

More importantly, fix the build because currently messaging_service.cc
and messaging_service.hh does not include stream_mutation_fragments_cmd.
I am not sure why it builds on my machine. Spotted this when backporting
the "streaming: Send error code from the sender to receiver" to 3.0
branch.

Refs: #4789


asias:
@slivne @tgrabiec Ping. I am waiting for this for the 3.0 backport requested by Glauber.

PR #4817, opened 2019-08-07
utils: loading_cache: prevent bogus discarded future warning

avikivity:
The future is safe to discard, so cast it to const and add a
comment explaining why it is safe.

It would be better done with a new API for gate, but we can
change that when that new API is available.

### DIFF
--- a/utils/loading_cache.hh
+++ b/utils/loading_cache.hh
@@ -580,7 +580,9 @@ private:
         }
 
         // Reload all those which value needs to be reloaded.
-        with_gate(_timer_reads_gate, [this] {
+        // We can discard the return value of with_gate() since stop() will wait
+        // got the gate to be closed (and thus for these continuations to complete).
+        (void)with_gate(_timer_reads_gate, [this] {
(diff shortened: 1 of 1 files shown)


utils/loading_cache.hh
```diff
@@ -580,7 +580,9 @@ private:
         }
 
         // Reload all those which value needs to be reloaded.
-        with_gate(_timer_reads_gate, [this] {
+        // We can discard the return value of with_gate() since stop() will wait
+        // got the gate to be closed (and thus for these continuations to complete).
```
tgrabiec: "got" looks like a typo.

denesb:
LGTM. BTW I have a patch that fixes all warnings in the Scylla codebase. I'm just waiting for the gate discussion to settle before sending it.
Maybe I should send it with all gates handled like this one (return value cast to void) and we can revisit them again after the API change lands.

avikivity:
Withdrawing this patch in favor of @denesb more complete work.

PR #4821, opened 2019-08-09
cql3: Handle empty LIKE pattern

dekimir:
Match SQL's LIKE in allowing an empty pattern, which matches only an empty text field.

Tests: unit (dev)

### DIFF
--- a/tests/cql_query_test.cc
+++ b/tests/cql_query_test.cc
@@ -3937,15 +3937,15 @@ SEASTAR_TEST_CASE(test_like_operator_bind_marker) {
     });
 }
 
-#if 0 // TODO: Enable when like_matcher handles blank pattern.
 SEASTAR_TEST_CASE(test_like_operator_blank_pattern) {
     return do_with_cql_env_thread([] (cql_test_env& e) {
-        cquery_nofail(e, "create table t (s text primary key, )");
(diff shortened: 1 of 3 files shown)


tests/cql_query_test.cc
```diff
@@ -3937,15 +3937,13 @@ SEASTAR_TEST_CASE(test_like_operator_bind_marker) {
     });
 }
 
-#if 0 // TODO: Enable when like_matcher handles blank pattern.
 SEASTAR_TEST_CASE(test_like_operator_blank_pattern) {
     return do_with_cql_env_thread([] (cql_test_env& e) {
         cquery_nofail(e, "create table t (s text primary key, )");
         cquery_nofail(e, "insert into t (s) values ('abc')");
         require_rows(e, "select s from t where s like '' allow filtering", {});
     });
```
avikivity: How about an extra row with an empty key to test that it is matched by a blank pattern?

dekimir:
Feedback addressed, PTAL.

avikivity:
Dequeued, got the noreply.github.com email address. Will merge manually.

avikivity:
Queued as 8be147d0693d7db2b85dfbf8e2b3edc43adc5585

PR #4827, opened 2019-08-11
Scyllatop: one pass update of multiple metrics

bhalevy:
Update previous results dictionary using the update_metrics method.
It calls metric_source.query_list to get a list of results (similar to discover()) then for each line in the response it updates results dictionary.

New results may be appeneded depending on the do_append parameter (True by default).
    
Previously, with prometheous, each metric.update called query_list resulting in O(n^2) when all metric were updated, like in the scylla_top dtest - causing test timeout when testing debug build.
(E.g. [dtest-debug/216/testReport/scyllatop_test/TestScyllaTop/default_start_test/](http://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug/216/testReport/scyllatop_test/TestScyllaTop/default_start_test/))

### DIFF
--- a/tools/scyllatop/dumptostdout.py
+++ b/tools/scyllatop/dumptostdout.py
@@ -1,5 +1,6 @@
 import livedata
 import views.stdout
+import logging
 
 
 class _FakeLoop:
@@ -8,6 +9,7 @@ class _FakeLoop:
         self._liveData = liveData
 
     def draw_screen(self):
+        logging.debug("iterations={}".format(self._iterations))
         if self._iterations is None:
             return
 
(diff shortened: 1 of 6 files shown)


slivne:
@amnonh I don't understand your comment - do you need benny to fix something ?

bhalevy:
@slivne we reviewed the change offline and there are changes to be made.

bhalevy:
In v2:
- fixed marking metrics as absent
- add ttl for expiring absent metrics (60 seconds by default)

slivne:
@amnonh ping

amnonh:
LGTM

bhalevy:
Maintainers: this test is flaky in next (e.g. [next/1272/testReport/scyllatop_test/TestScyllaTop/default_start_test](https://jenkins.scylladb.com/view/master/job/scylla-master/job/next/1272/testReport/scyllatop_test/TestScyllaTop/default_start_test))
and constantly failing in dtest-debug.

PR #4828, opened 2019-08-11
Clean estimated histogram

amnonh:
These series clean the estimated histogram implementation by removing FIXME and if0 from the code and allow finer and customized histogram report.

It uses that implementation for the read and write latency histogram report.

### DIFF
--- a/utils/estimated_histogram.hh
+++ b/utils/estimated_histogram.hh
@@ -114,15 +114,6 @@ struct estimated_histogram {
     }
 
 
-    // FIXME: convert Java code below.
-#if 0
-    public EstimatedHistogram(long[] offsets, long[] bucketData)
-    {
-        assert bucketData.length == offsets.length +1;
-        bucketOffsets = offsets;
-        buckets = new AtomicLongArray(bucketData);
-    }
-#endif
 private:
     void new_offsets(int size) {
         bucket_offsets.resize(size);
@@ -251,66 +242,12 @@ public:
 
     friend estimated_histogram merge(estimated_histogram a, const estimated_histogram& b);
 
-    // FIXME: convert Java code below.
-#if 0
     /**
      * @return the count in the given bucket
      */
-    long get(int bucket)
-    {
-        return buckets.get(bucket);
+    int64_t get(int bucket) {
+        return buckets[bucket];
     }
-
-    /**
-     * @param reset zero out buckets afterwards if true
-     * @return a long[] containing the current histogram buckets
-     */
-    public long[] getBuckets(boolean reset)
-    {
-        final int len = buckets.length();
-        long[] rv = new long[len];
-
-        if (reset)
-            for (int i = 0; i < len; i++)
-                rv[i] = buckets.getAndSet(i, 0L);
-        else
-            for (int i = 0; i < len; i++)
-                rv[i] = buckets.get(i);
-
-        return rv;
-    }
-
-    /**
-     * @return the smallest value that could have been added to this histogram
-     */
-    public long min()
-    {
-        for (int i = 0; i < buckets.length(); i++)
-        {
-            if (buckets.get(i) > 0)
-                return i == 0 ? 0 : 1 + bucketOffsets[i - 1];
-        }
-        return 0;
-    }
-
-    /**
-     * @return the largest value that could have been added to this histogram.  If the histogram
-     * overflowed, returns Long.MAX_VALUE.
-     */
-    public long max()
-    {
-        int lastBucket = buckets.length() - 1;
-        if (buckets.get(lastBucket) > 0)
-            return Long.MAX_VALUE;
-
-        for (int i = lastBucket - 1; i >= 0; i--)
-        {
-            if (buckets.get(i) > 0)
-                return bucketOffsets[i];
-        }
-        return 0;
-    }
-#endif
 
     /**
      * @param percentile
@@ -372,15 +309,6 @@ public:
         }
         return *this;
     }
-#if 0
-    /**
-     * @return true if this histogram has overflowed -- that is, a value larger than our largest bucket could bound was added
-     */
-    public boolean isOverflowed()
-    {
-        return buckets.get(buckets.length() - 1) > 0;
-    }
-#endif
 
     friend std::ostream& operator<<(std::ostream& out, const estimated_histogram& h) {
         // only print overflow if there is any
@@ -436,71 +364,6 @@ public:
         s += "]";
         return s;
     }
-
-#if 0
-    @Override
-    public boolean equals(Object o)
-    {
-        if (this == o)
-            return true;
-
-        if (!(o instanceof EstimatedHistogram))
-            return false;
-
-        EstimatedHistogram that = (EstimatedHistogram) o;
-        return Arrays.equals(getBucketOffsets(), that.getBucketOffsets()) &&
-               Arrays.equals(getBuckets(false), that.getBuckets(false));
-    }
-
-    @Override
-    public int hashCode()
-    {
-        return Objects.hashCode(getBucketOffsets(), getBuckets(false));
-    }
-
-    public static class EstimatedHistogramSerializer implements ISerializer<EstimatedHistogram>
-    {
-        public void serialize(EstimatedHistogram eh, DataOutputPlus out) throws IOException
-        {
-            long[] offsets = eh.getBucketOffsets();
-            long[] buckets = eh.getBuckets(false);
-            out.writeInt(buckets.length);
-            for (int i = 0; i < buckets.length; i++)
-            {
-                out.writeLong(offsets[i == 0 ? 0 : i - 1]);
-                out.writeLong(buckets[i]);
-            }
-        }
-
-        public EstimatedHistogram deserialize(DataInput in) throws IOException
-        {
-            int size = in.readInt();
-            long[] offsets = new long[size - 1];
-            long[] buckets = new long[size];
-
-            for (int i = 0; i < size; i++) {
-                offsets[i == 0 ? 0 : i - 1] = in.readLong();
-                buckets[i] = in.readLong();
-            }
-            return new EstimatedHistogram(offsets, buckets);
-        }
-
-        public long serializedSize(EstimatedHistogram eh, TypeSizes typeSizes)
-        {
-            int size = 0;
-
-            long[] offsets = eh.getBucketOffsets();
-            long[] buckets = eh.getBuckets(false);
-            size += typeSizes.sizeof(buckets.length);
-            for (int i = 0; i < buckets.length; i++)
-            {
-                size += typeSizes.sizeof(offsets[i == 0 ? 0 : i - 1]);
-                size += typeSizes.sizeof(buckets[i]);
-            }
-            return size;
-        }
-    }
-#endif
 };
 
 inline estimated_histogram estimated_histogram_merge(estimated_histogram a, const estimated_histogram& b) {


amnonh:
I would like to add additional functionality to estimated_histogram, it will be best if we can merge the clean up patch first

slivne:
@glommer can you please bless so I can get avi to review and merge

glommer:
It works as a cleanup patch, and I see that amnon reimplements min() later.
From my PoV this is okay, Avi may want a different patch split

utils/estimated_histogram.hh
```diff
@@ -247,66 +238,12 @@ public:
...
-            for (int i = 0; i < len; i++)
-                rv[i] = buckets.get(i);
-
-        return rv;
-    }
-
-    /**
-     * @return the smallest value that could have been added to this histogram
-     */
-    public long min()
```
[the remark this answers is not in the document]
amnonh: it's implemented, but we didn't remove the java code, so I did

slivne:
@avikivity ping

slivne:
@avikivity ping we have globbers blessing on this

utils/estimated_histogram.hh
```diff
@@ -339,16 +276,7 @@ public:
...
-        int64_t elements = 0;
-        int64_t sum = 0;
-        for (size_t i = 0; i < lastBucket; i++) {
-            long bCount = buckets[i];
-            elements += bCount;
-            sum += bCount * bucket_offsets[i];
-        }
-
-        return ((double) (sum + elements -1)/ elements);
+        return (_count) ? ((double)_sample_sum)/_count : 0;
```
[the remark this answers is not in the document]
avikivity: Yes, @amnonh please fix.

utils/estimated_histogram.hh
```diff
@@ -73,13 +73,59 @@ struct estimated_histogram {
...
+     * \brief return a subset of the histogram according to given buckets
+     *
+     * While it is useful to reduce the number of buckets, it's not always clear what
+     * is the best bucket list the caller wants.
+     *
+     * This function, gets a vector of buckets, buckets can be double or duration depends
+     * on the histogram usage and return a minimized histogram
+     */
+    template<class T>
+    seastar::metrics::histogram get_histogram(const std::vector<T>& buckets_limit) const {
```
avikivity: Why not pass base, ratio, and count? It looks too generic.

Also, this shouldn't be a template is there are just two possible Ts. You can have a private implenentation for two non-template public funcitons.

utils/estimated_histogram.hh
```diff
@@ -352,6 +388,43 @@ public:
...
 
+    /*!
+     * \brief duration vector bucket
+     *
+     * As a default, it is more or less make sense.
+     */
+    static const std::vector<duration>& get_duration_vector() {
+        static std::vector<duration> res {
+            std::chrono::microseconds(16),
+            std::chrono::microseconds(32),
```
avikivity: Where does this come from?

service/storage_proxy.cc
```diff
@@ -809,8 +809,8 @@ storage_proxy::storage_proxy(distributed<database>& db, storage_proxy::config cf
     , _view_update_handlers_list(std::make_unique<view_update_handlers_list>()) {
     namespace sm = seastar::metrics;
```
avikivity: The distribution above is exponential, so the patch description is incorrect.

utils/estimated_histogram.hh
amnonh: the incentive for the change is that exponential is good for lower values, but not so good for higher values.

I'll change to non-template implmentation

service/storage_proxy.cc
amnonh: not entirely, it replaces the exponential distribution we had with something more forgiven near the 1s buckets

amnonh:
@avikivity I've forced push a new version with the changes that are not addressed in the review

service/storage_proxy.cc
amnonh: right, but everything on the way supports it, we can get finer resolution when we need it

amnonh: in that case we can keep the 1.2 distribution that is currently reported, we want to lower the number of buckets from the number that is used to collect the data (90) to a lower number

amnonh:
@avikivity I have forced push v3, I use a 1.3 ratio, it uses a bit more buckets but I guess it's not the end of the world.

We should expect lower latencies following this patch

slivne:
@amnonh lease rebase so we can ping avi to rereview

amnonh:
I have pull rebase to master

slivne:
@avikivity ping

slivne:
@avikivity ping

slivne:
@avikivity ping

bhalevy:
target branch should be `next`
and in addition to this patch, need to fix https://github.com/scylladb/scylla/issues/5577

slivne:
@avikivity whats blocking this fix - we also have a customer issue that needs to be fixed ontop of this.

service/storage_proxy.cc
```diff
@@ -1382,10 +1382,10 @@ storage_proxy::storage_proxy(distributed<database>& db, storage_proxy::config cf
...
-        sm::make_histogram("cas_read_latency", sm::description("Transactional read latency histogram"), [this]{ return _stats.estimated_cas_read.get_histogram(16, 20);}),
-        sm::make_histogram("cas_write_latency", sm::description("Transactional write latency histogram"), [this]{return _stats.estimated_cas_write.get_histogram(16, 20);}),
+        sm::make_histogram("read_latency", sm::description("The general read latency histogram"), [this]{ return _stats.estimated_read.get_histogram(std::chrono::microseconds(500), 33, 1.3);}),
+        sm::make_histogram("write_latency", sm::description("The general write latency histogram"), [this]{return _stats.estimated_write.get_histogram(std::chrono::microseconds(500), 33, 1.3);}),
+        sm::make_histogram("cas_read_latency", sm::description("Transactional read latency histogram"), [this]{ return _stats.estimated_cas_read.get_histogram(std::chrono::microseconds(500), 33, 1.3);}),
+        sm::make_histogram("cas_write_latency", sm::description("Transactional write latency histogram"), [this]{return _stats.estimated_cas_write.get_histogram(std::chrono::microseconds(500), 33, 1.3);}),
```
avikivity: I don't see a 3-argument version of get_historgram. Did the patches get reordered by github?

utils/estimated_histogram.hh
```diff
@@ -73,40 +73,59 @@ struct estimated_histogram {
...
 
-    seastar::metrics::histogram get_histogram(size_t lower_bucket = 1, size_t max_buckets = 16) const {
+    /*
+     * \brief get an accumulative histogram
+     * When getting an accumulative histogram, you can set what is the first bucket to report
+     * what is the number of buckets
+     * and what is the ratio between bucket limits.
+     * Setting the bucket_ratio for 1 or less, will use the buckets as-is
+     */
+    seastar::metrics::histogram get_histogram(size_t lower_bucket = 1, size_t max_buckets = 20, double bucket_ratio = 1.44) const {
```
avikivity: I don't understand how 1.3 gives more accurate results than the original 1.2.

avikivity: And why we use 1.44 for the default

utils/estimated_histogram.hh
```diff
@@ -73,40 +73,59 @@ struct estimated_histogram {
...
                 cummulative_count += buckets[pos];
+                current_bound = bucket_offsets[pos];
                 pos++;
             }
-
+            if (bucket_ratio <= 1) {
+                last_bound = bucket_offsets[pos];
+            } else {
+                last_bound *= bucket_ratio;
+            }
```
avikivity: I don't understand this bucket_ratio < 1 thing. Isn't there a bucket_ratio that corresponds to the current code? If so, it can be the default.

service/storage_proxy.cc
amnonh: yes

utils/estimated_histogram.hh
amnonh: It eliminates all the ratio calculation, so it allows the caller to say: give me the maximal precision available.
If we know that we want the maximal precision, why do the calculation? instead, just take the value.

utils/estimated_histogram.hh
amnonh: We never use the original 1.2, up until now we use a factor of 2, which appears to be too high.
I pick sqrt(2) it seems to make sense, I can pick anything else including the maximum available.

utils/estimated_histogram.hh
```diff
@@ -360,8 +360,6 @@ public:
         s += "]";
         return s;
     }
-
-    }
```
avikivity: There is another, more efficient way to do this, and that is by using pseudo floating point. For example we can have 4 significant bits and then a value of 195 = 0xc3 = 0y1100'0011 = (128) * (1 + (0y100'0011 / 128)) =~ 128 * (1 + 0y100 / 8) = (1<<7)*(1 + 4/8)

The bucket number is therefore 7*8+4 = 60. To recover the value again divide by 8 (we get 7 with a reminder of 4), and write (1<<7) + (4 << (7 - 3)) = 128 + 64 = 192.

utils/estimated_histogram.hh
avikivity: I don't understand what's the maximal precision, since we create buckets in any case.

amnonh: We start with the base estimated histogram (typically, uses a ratio of 1.2 and 92 bucket but can use anything else).
When we report it, we use an accumulative histogram with an (optional) different ratio (i.e. different precision).
So the user of the accumulative histogram (Prometheus) can choose to limit the size of the histogram it will get, or ask for the highest precision histogram available.
The idea is to allow requesting the histogram without reducing the number of buckets.

A valid alternative is to report it by default.
If we are reporting all the buckets, there is no point in calculating the buckets limits, instead, we can just copy the values.

utils/estimated_histogram.hh
amnonh: I'll have a look

slivne:
@amnonh ping

utils/estimated_histogram.hh
amnonh: I've openned https://github.com/scylladb/scylla/issues/5815

amnonh:
@avikivity I have forced push a new version (also cleared the conflicts)
I have addressed your comments here, but if there is something else you want to change I can do it.

Main changes, the default generated histogram now uses 1.3 ratio and reduce the number of buckets to 33 (about 1/3 of the original histogram)
I've based it on your PR comment that looks reasonable.
Additional changes to the histogram will be based on the cleaner version

slivne:
@avikivity can you please review this

slivne:
@avikivity ping

amnonh:
@avikivity I would love to move to the next step (after this initial clean up is completed) for an optimized histogram that will be widely used.
Is there's anything else that stops this PR from getting in?

slivne:
@avikivity ping

utils/estimated_histogram.hh
```diff
@@ -73,40 +73,58 @@ struct estimated_histogram {
 
     int64_t _count = 0;
```
avikivity, line 74: This isn't "Clean estimated histogram".

utils/estimated_histogram.hh
```diff
@@ -73,40 +73,58 @@ struct estimated_histogram {
...
 
     estimated_histogram(int bucket_count = 90) {
 
         new_offsets(bucket_count);
         buckets.resize(bucket_offsets.size() + 1, 0);
     }
 
-    seastar::metrics::histogram get_histogram(size_t lower_bucket = 1, size_t max_buckets = 16) const {
+    /*
+     * \brief get an accumulative histogram
```
avikivity, line 86: s/accumulative/cumulative/

utils/estimated_histogram.hh
```diff
@@ -73,40 +73,58 @@ struct estimated_histogram {
...
+     * When getting an accumulative histogram, you can set what is the first bucket to report
+     * what is the number of buckets and what is the ratio between bucket limits.
+     * Setting the bucket_ratio to 1 or less, will return all the buckets.
+     */
+    seastar::metrics::histogram get_histogram(size_t lower_bucket = 1, size_t max_buckets = 33, double bucket_ratio = 1.3) const {
```
avikivity, line 91: Why the change to 33? Not that I think there should be a default here (except maybe all).

utils/estimated_histogram.hh
avikivity: I still don't understand what bucket_ratio <= 1 means. Is it a way to say "change the method of calculating ratios to something else"? If so it is a bad method because no one will be able to guess it.

avikivity: I see there's a comment that says so. But it's a bad API. Better is to use a function with a different name (internally they can share implementations).

utils/estimated_histogram.hh
```diff
@@ -174,6 +192,12 @@ public:
         if (new_count <= _count) {
             return;
         }
+        if (!_min || n < _min) {
+            _min = n;
+        }
+        if (!_max || n > _max) {
+            _max = n;
+        }
```
avikivity, line 200: Please use min_tracker/max_tracker. Using 0 as a flag value is dangerous.

utils/estimated_histogram.hh
```diff
@@ -192,31 +216,15 @@ public:
...
     int64_t min() const {
-        size_t i = 0;
-        for (auto b : buckets) {
-            if (b > 0) {
-                return i == 0 ? 0 : 1 + bucket_offsets[i - 1];
-            }
-            i++;
-        }
-        return 0;
+        return _min;
```
avikivity, line 219: What's the motivation? we get better min/max, but also do more work per operation. Is there a problem with the current min/max?

utils/estimated_histogram.hh
```diff
@@ -233,6 +241,8 @@ public:
         }
         _count += b._count;
         _sample_sum += b._sample_sum;
+        _min = (b._min < _min) ? b._min : _min;
+        _max = (b._max > _max) ? b._max : _max;
```
avikivity, line 245: std::min/std::max.

What if one of them had _min = 0 because there were no values?

utils/estimated_histogram.hh
```diff
@@ -276,16 +286,7 @@ public:
...
-        int64_t elements = 0;
-        int64_t sum = 0;
-        for (size_t i = 0; i < lastBucket; i++) {
-            long bCount = buckets[i];
-            elements += bCount;
-            sum += bCount * bucket_offsets[i];
-        }
-
-        return ((double) (sum + elements -1)/ elements);
+        return (_count) ? ((double)_sample_sum)/_count : 0;
```
avikivity, line 289: Again we get better accuracy, but at the cost of doing more work. It's negligible compared to the work done by searching for the right bucket (which will be eliminated by using fixed-point pseudo-logarithms), but still should be justified.

service/storage_proxy.cc
```diff
@@ -1349,7 +1349,7 @@ void storage_proxy_stats::write_stats::register_stats() {
     _metrics.add_group(COORDINATOR_STATS_CATEGORY, {
```
avikivity, line 1349: Please add the word "change" to the subject so it becomes a sentence.

avikivity: The proemtheus text format is so :facepalm:

service/storage_proxy.cc
```diff
@@ -1444,7 +1444,7 @@ void storage_proxy_stats::stats::register_stats() {
...
         sm::make_histogram("read_latency", sm::description("The general read latency histogram"),
                 {storage_proxy_stats::current_scheduling_group_label()},
-                [this]{ return estimated_read.get_histogram(16, 20);}),
+                [this]{ return estimated_read.get_histogram(std::chrono::microseconds(500), 33, 1.3);}),
```
avikivity, line 1447: I think we need a few more buckets, so a 4-second query isn't shown as +inf.

avikivity: maybe a better ratio is sqrt(2) or cuberoot(2), so the powers of two are still represented. It's not really important, but it's nicer to see numbers such as 1ms or 16ms. Really a cosmetic issue.

utils/estimated_histogram.hh
amnonh: it's not a critical problem, but there is no other way of knowing what is the real min and max.

service/storage_proxy.cc
amnonh: I'm going to drop all that ratio as soon as I'll move to the new implementation based on your comments on the previous review.

utils/estimated_histogram.hh
amnonh: I'll replace it all anyhow, this is a step that keeps the current API.

utils/estimated_histogram.hh
amnonh: Are you talking about the PR header?
The patch is about getting better estimation

avikivity: Yes. The subject should summarize what the patches are doing.

amnonh:
@avikivity without the min/max optimization and sense it will not go into 4.1, there is no point to add the part of a better resolution and I'll delay it to a different PR based on pseudo floating point.

This lives what I hope is noncontroversial, the cleaning part of the patch.

This is a pull rebase of that part

avikivity:
Queued as 515c77e881

PR #4830, opened 2019-08-11
build: build seastar in build directory

avikivity:
Currently, seastar is built in seastar/build/{mode}. This means we have two build
directories: build/{mode} and seastar/build/{mode}.

This patch changes that to have only a single build directory (build/{mode}). It
does that by calling Seastar's cmake directly instead of through Seastar's
./configure.py.  However, to support dpdk, if that is enabled it calls cmake
through Seastar's ./cooking.sh (similar to what Seastar's ./configure.py does).

All ./configure.py flags are translated to cmake variables, in the same way that
Seastar does.

Tests: unit (dev), build with and without dpdk, peek at compiler flags while the
       compiler is running.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1060,17 +1060,6 @@ scylla_release = file.read().strip()
 
 extra_cxxflags["release.cc"] = "-DSCYLLA_VERSION=\"\\\"" + scylla_version + "\\\"\" -DSCYLLA_RELEASE=\"\\\"" + scylla_release + "\\\"\""
 
-seastar_flags = []
-if args.dpdk:
-    # fake dependencies on dpdk, so that it is built before anything else
-    seastar_flags += ['--enable-dpdk']
-if args.gcc6_concepts:
-    seastar_flags += ['--enable-gcc6-concepts']
-if args.alloc_failure_injector:
-    seastar_flags += ['--enable-alloc-failure-injector']
-if args.split_dwarf:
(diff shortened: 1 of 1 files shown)


configure.py
```diff
@@ -1438,7 +1453,7 @@ def configure_zstd(build_dir, mode):
         rule configure
           command = {python} configure.py $configure_args
           generator = 1
-        build build.ninja: configure | configure.py seastar/configure.py
+        build build.ninja: configure | configure.py seastar/CMakeLists.txt
```
denesb: I think the `seastar/*` dependency is unnecessary in this rule. We already unconditionally invoke `ninja .../libseastar.a` regardless of the target built. CMake will detect if any of its `CMakeLists.txt` files is dirty and will re-run itself.

avikivity: ok, will drop (it's very partial in any case)

avikivity:
updated: dropped dependency on seastar/CMakeLists.txt according to @denesb recommendation.

configure.py
denesb: Yes, there is a CMakeLists.txt in every directory, not just the top-level one.

avikivity:
@scylladb/scylla-maint ping

avikivity:
Dequeued since it fails next checks. I see @espindola already sent a fix, I'll merge it into the original patch and resubmit.

PR #4840, opened 2019-08-13
docker: relax permission checks

### LINKED ISSUE #4823: 3.1.0.rc3 fails to start on volume mount permission errors
scylla --version 3.1.0.rc3-0.20190725.bebfd7b26

Running inside a docker container and mounting a volume at /var/lib/scylla fails with below error. Previous version 3.0.9 worked fine.

`
running: (['/usr/lib/scylla/scylla_dev_mode_setup', '--developer-mode', '0'],)
running: (['/usr/lib/scylla/scylla_io_setup'],)
/usr/lib/scylla/libexec/scylla_io_setup:88: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  cfg = yaml.load(open(os.path.join(conf_dir, "scylla.yaml")))
ERROR:root:/var/lib/scylla/commitlog was not found. Please check the configuration and run scylla_io_setup again.

tuning /sys/devices/pci0000:00/0000:00:1f.0/nvme/nvme1/nvme1n1
tuning: /sys/devices/pci0000:00/0000:00:1f.0/nvme/nvme1/nvme1n1/queue/nomerges 2
warning: unable to tune /sys/devices/pci0000:00/0000:00:1f.0/nvme/nvme1/nvme1n1/queue/nomerges to 2

failed!
  File "/docker-entrypoint.py", line 17, in <module>
    setup.io()
  File "/scyllasetup.py", line 51, in io
    self._run(['/usr/lib/scylla/scylla_io_setup'])
  File "/scyllasetup.py", line 30, in _run
    subprocess.check_call(*args, **kwargs)
  File "/opt/scylladb/python3/lib64/python3.7/subprocess.py", line 347, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/usr/lib/scylla/scylla_io_setup']' returned non-zero exit status 1.
/scyllasetup.py:42: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  cfg = yaml.load(open(os.path.join(conf_dir, "scylla.yaml")))
`

My current workaround is to use a custom docker_entrypoint.py  that creates /var/lib/scylla/commitlog and chown -R scylla:scylla /var/lib/scylla before supervisord spins up processes.

avikivity:
Commit e3f7fe44c04006bfe601713dde97a0bd1436d499 added file owner validation to prevent Scylla from crashing when it tries to touch a file it doesn't own. However, under docker, we cannot expect to pass this check since user IDs are from different namespaces: the process runs in a container namespace, but the data files usually come from a mounted volume, and so their uids are from the host namespace.

So we need to relax the check. We do this by reverting b1226fb15a5a8f0ac4d37733df3c5ba12d84644c, which causes Scylla to run as euid 0 in docker, and by special-casing euid 0 in the ownership verification step.

Fixes #4823.

### DIFF
--- a/dist/docker/redhat/etc/supervisord.conf.d/scylla-housekeeping.conf
+++ b/dist/docker/redhat/etc/supervisord.conf.d/scylla-housekeeping.conf
@@ -4,4 +4,3 @@ stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
 stderr_logfile=/dev/stderr
 stderr_logfile_maxbytes=0
-user=scylla
--- a/dist/docker/redhat/etc/supervisord.conf.d/scylla-jmx.conf
+++ b/dist/docker/redhat/etc/supervisord.conf.d/scylla-jmx.conf
@@ -4,4 +4,3 @@ stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
 stderr_logfile=/dev/stderr
 stderr_logfile_maxbytes=0
-user=scylla
--- a/dist/docker/redhat/etc/supervisord.conf.d/scylla-server.conf
+++ b/dist/docker/redhat/etc/supervisord.conf.d/scylla-server.conf
@@ -4,4 +4,3 @@ stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
 stderr_logfile=/dev/stderr
 stderr_logfile_maxbytes=0
-user=scylla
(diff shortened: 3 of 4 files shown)


penberg:
@avikivity I switched merge target branch to `next` from `master`.

penberg:
Looks good to me. @bhalevy?

PR #4846, opened 2019-08-13
Validate position in partition monotonicity

### LINKED ISSUE #4803: Expand sstable writer validation check for partition fragments monotonicity
Based on @tgrabiec internal presentation:

Patch 95c08047318ba819cdd74ac7b1aabcd35f086a0b added validation of partition_end being  emitted 

Expand the validation to cover monotonicity of all partition fragments parts:

* partition_start
* static_row
* clustering row
* partition_end

such that they adhere partition_start, [static_row], {clustering_row}* parition_end

bhalevy:
Introduce `mutation_fragment_stream_validator` class and use it as a Filter to `flat_mutation_reader::consume_in_thread` from `sstable::write_components` to validate partition region and optionally clustering key monotonicity.

Fixes #4803

### DIFF
--- a/db/config.cc
+++ b/db/config.cc
@@ -155,6 +155,12 @@ struct convert<db::config::seed_provider_type> {
 
 }
 
+#if defined(DEBUG)
+#define ENABLE_SSTABLE_KEY_VALIDATION true
+#else
+#define ENABLE_SSTABLE_KEY_VALIDATION false
+#endif
+
 #define str(x)  #x
 #define _mk_init(name, type, deflt, status, desc, ...)  , name(this, str(name), value_status::status, type(deflt), desc)
 
@@ -678,7 +684,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
     , enable_keyspace_column_family_metrics(this, "enable_keyspace_column_family_metrics", value_status::Used, false, "Enable per keyspace and per column family metrics reporting")
     , enable_sstable_data_integrity_check(this, "enable_sstable_data_integrity_check", value_status::Used, false, "Enable interposer which checks for integrity of every sstable write."
         " Performance is affected to some extent as a result. Useful to help debugging problems that may arise at another layers.")
-    , enable_sstable_key_validation(this, "enable_sstable_key_validation", value_status::Used, false, "Enable validation of partition and clustering keys monotonicity"
+    , enable_sstable_key_validation(this, "enable_sstable_key_validation", value_status::Used, ENABLE_SSTABLE_KEY_VALIDATION, "Enable validation of partition and clustering keys monotonicity"
         " Performance is affected to some extent as a result. Useful to help debugging problems that may arise at another layers.")
     , cpu_scheduler(this, "cpu_scheduler", value_status::Used, true, "Enable cpu scheduling")
     , view_building(this, "view_building", value_status::Used, true, "Enable view building; should only be set to false when the node is experience issues due to view building")
--- a/db/config.cc
+++ b/db/config.cc
@@ -678,6 +678,8 @@ db::config::config(std::shared_ptr<db::extensions> exts)
     , enable_keyspace_column_family_metrics(this, "enable_keyspace_column_family_metrics", value_status::Used, false, "Enable per keyspace and per column family metrics reporting")
     , enable_sstable_data_integrity_check(this, "enable_sstable_data_integrity_check", value_status::Used, false, "Enable interposer which checks for integrity of every sstable write."
         " Performance is affected to some extent as a result. Useful to help debugging problems that may arise at another layers.")
+    , enable_sstable_key_validation(this, "enable_sstable_key_validation", value_status::Used, false, "Enable validation of partition and clustering keys monotonicity"
+        " Performance is affected to some extent as a result. Useful to help debugging problems that may arise at another layers.")
     , cpu_scheduler(this, "cpu_scheduler", value_status::Used, true, "Enable cpu scheduling")
     , view_building(this, "view_building", value_status::Used, true, "Enable view building; should only be set to false when the node is experience issues due to view building")
     , enable_sstables_mc_format(this, "enable_sstables_mc_format", value_status::Used, true, "Enable SSTables 'mc' format to be used as the default file format")
--- a/flat_mutation_reader.cc
+++ b/flat_mutation_reader.cc
@@ -922,11 +922,15 @@ bool mutation_fragment_stream_validator::operator()(const dht::decorated_key& dk
     return true;
 }
 
-mutation_fragment_stream_validator::mutation_fragment_stream_validator(const schema& s)
+mutation_fragment_stream_validator::mutation_fragment_stream_validator(const schema& s, bool compare_keys)
     : _schema(s)
     , _prev_kind(mutation_fragment::kind::partition_end)
     , _prev_pos(position_in_partition::end_of_partition_tag_t{})
-{ }
+    , _prev_region(partition_region::partition_end)
+    , _compare_keys(compare_keys)
+{
+    fmr_logger.debug("[validator {}] Will validate {} monotonicity.", static_cast<void*>(this), compare_keys ? "keys" : "only partition regions");
+}
 
 mutation_fragment_stream_validator::~mutation_fragment_stream_validator() {
     if (_prev_kind != mutation_fragment::kind::partition_end) {
@@ -937,26 +941,39 @@ mutation_fragment_stream_validator::~mutation_fragment_stream_validator() {
 bool mutation_fragment_stream_validator::operator()(const mutation_fragment& mv) {
     auto kind = mv.mutation_fragment_kind();
     auto pos = mv.position();
+    auto region = pos.region();
     bool valid = false;
 
     fmr_logger.debug("[validator {}] {}:{}", static_cast<void*>(this), kind, pos);
 
     if (mv.is_partition_start()) {
         valid = (_prev_kind == mutation_fragment::kind::partition_end);
-    } else {
+    } else if (_compare_keys) {
         auto less = position_in_partition::less_compare(_schema);
         if (_prev_kind != mutation_fragment::kind::range_tombstone) {
             valid = less(_prev_pos, pos);
         } else {
             valid = !less(pos, _prev_pos);
         }
+    } else if (region != partition_region::clustered) {
+        valid = (_prev_region < region);
+    } else {
+        valid = (_prev_region <= region);
     }
 
     if (__builtin_expect(!valid, false)) {
-        on_internal_error(fmr_logger, format("[validator {}] Unexpected mutation fragment: previous {}:{}, current {}:{}", static_cast<void*>(this), _prev_kind, _prev_pos, kind, pos));
+        auto fmt = "[validator {}] Unexpected mutation fragment: previous {}:{}, current {}:{}";
+        sstring msg = _compare_keys ?
+                format(fmt, static_cast<void*>(this), _prev_kind, _prev_pos, kind, pos) :
+                format(fmt, static_cast<void*>(this), _prev_kind, _prev_region, kind, pos);
+        on_internal_error(fmr_logger, msg);
     }
 
     _prev_kind = mv.mutation_fragment_kind();
-    _prev_pos = pos;
+    if (_compare_keys) {
+        _prev_pos = pos;
+    } else {
+        _prev_region = region;
+    }
     return true;
 }
--- a/flat_mutation_reader.cc
+++ b/flat_mutation_reader.cc
@@ -27,6 +27,9 @@
 
 #include <boost/range/adaptor/transformed.hpp>
 #include <seastar/util/defer.hh>
+#include "utils/exceptions.hh"
+
+logging::logger fmr_logger("flat_mutation_reader");
 
 static size_t compute_buffer_size(const schema& s, circular_buffer<mutation_fragment>& buffer)
 {
@@ -911,4 +914,49 @@ public:
 
 flat_mutation_reader make_generating_reader(schema_ptr s, std::function<future<mutation_fragment_opt> ()> get_next_fragment) {
     return make_flat_mutation_reader<generating_reader>(std::move(s), std::move(get_next_fragment));
+}
+
+bool mutation_fragment_stream_validator::operator()(const dht::decorated_key& dk) {
+    // FIXME: Validate partition key monotonicity
+    //        https://github.com/scylladb/scylla/issues/4804
+    return true;
+}
+
+mutation_fragment_stream_validator::mutation_fragment_stream_validator(const schema& s)
+    : _schema(s)
+    , _prev_kind(mutation_fragment::kind::partition_end)
+    , _prev_pos(position_in_partition::end_of_partition_tag_t{})
+{ }
+
+mutation_fragment_stream_validator::~mutation_fragment_stream_validator() {
+    if (_prev_kind != mutation_fragment::kind::partition_end) {
+        on_internal_error(fmr_logger, format("[validator {}] Stream ended with unclosed partition: {}", static_cast<void*>(this), _prev_kind));
+    }
+}
+
+bool mutation_fragment_stream_validator::operator()(const mutation_fragment& mv) {
+    auto kind = mv.mutation_fragment_kind();
+    auto pos = mv.position();
+    bool valid = false;
+
+    fmr_logger.debug("[validator {}] {}:{}", static_cast<void*>(this), kind, pos);
+
+    if (mv.is_partition_start()) {
+        valid = (_prev_kind == mutation_fragment::kind::partition_end);
+    } else {
+        auto less = position_in_partition::less_compare(_schema);
+        if (_prev_kind != mutation_fragment::kind::range_tombstone) {
+            valid = less(_prev_pos, pos);
+        } else {
+            valid = !less(pos, _prev_pos);
+        }
+    }
+
+    if (__builtin_expect(!valid, false)) {
+        on_internal_error(fmr_logger, format("[validator {}] Unexpected mutation fragment: previous {}:{}, current {}:{}", static_cast<void*>(this), _prev_kind, _prev_pos, kind, pos));
+    }
+
+    _prev_kind = mv.mutation_fragment_kind();
+    _prev_pos = pos;
+    return true;
 }
--- a/flat_mutation_reader.hh
+++ b/flat_mutation_reader.hh
@@ -61,6 +61,11 @@ GCC6_CONCEPT(
     concept bool PartitionFilter = requires(T filter, const dht::decorated_key& dk) {
         { filter(dk) } -> bool;
     };
+
+    template<typename T>
+    concept bool MutationFragmentFilter = requires(T filter, const mutation_fragment& mf) {
+        { filter(mf) } -> bool;
+    };
 )
 
 /*
@@ -177,7 +182,7 @@ public:
 
         template<typename Consumer, typename Filter>
         GCC6_CONCEPT(
-            requires FlatMutationReaderConsumer<Consumer>() && PartitionFilter<Filter>
+            requires FlatMutationReaderConsumer<Consumer>() && PartitionFilter<Filter> && MutationFragmentFilter<Filter>
         )
         // A variant of consume_pausable() that expects to be run in
         // a seastar::thread.
@@ -200,7 +205,7 @@ public:
                     next_partition();
                     continue;
                 }
-                if (consumer(std::move(mf)) == stop_iteration::yes) {
+                if (filter(mf) && (consumer(std::move(mf)) == stop_iteration::yes)) {
                     return;
                 }
             }
@@ -277,7 +282,7 @@ public:
 
         template<typename Consumer, typename Filter>
         GCC6_CONCEPT(
-            requires FlattenedConsumer<Consumer>() && PartitionFilter<Filter>
+            requires FlattenedConsumer<Consumer>() && PartitionFilter<Filter> && MutationFragmentFilter<Filter>
         )
         // A variant of consumee() that expects to be run in a seastar::thread.
         // Partitions for which filter(decorated_key) returns false are skipped
@@ -360,9 +365,50 @@ public:
         return _impl->consume(std::move(consumer), timeout);
     }
 
+    class filter {
+    private:
+        std::function<bool (const dht::decorated_key&)> _partition_filter = [] (const dht::decorated_key&) { return true; };
+        std::function<bool (const mutation_fragment&)> _mutation_fragment_filter = [] (const mutation_fragment&) { return true; };
+    public:
+        filter() = default;
+
+        filter(std::function<bool (const dht::decorated_key&)>&& pf)
+            : _partition_filter(std::move(pf))
+        { }
+
+        filter(std::function<bool (const dht::decorated_key&)>&& pf,
+               std::function<bool (const mutation_fragment&)>&& mf)
+            : _partition_filter(std::move(pf))
+            , _mutation_fragment_filter(std::move(mf))
+        { }
+
+        template <typename Functor>
+        filter(Functor&& f)
+            : _partition_filter(std::forward<Functor>(f))
+        { }
+
+        bool operator()(const dht::decorated_key& dk) const {
+            return _partition_filter(dk);
+        }
+
+        bool operator()(const mutation_fragment& mf) const {
+            return _mutation_fragment_filter(mf);
+        }
+    };
+
+    struct no_filter {
+        bool operator()(const dht::decorated_key& dk) const {
+            return true;
+        }
+
+        bool operator()(const mutation_fragment& mf) const {
+            return true;
+        }
+    };
+
     template<typename Consumer, typename Filter>
     GCC6_CONCEPT(
-        requires FlattenedConsumer<Consumer>() && PartitionFilter<Filter>
+        requires FlattenedConsumer<Consumer>() && PartitionFilter<Filter> && MutationFragmentFilter<Filter>
     )
     auto consume_in_thread(Consumer consumer, Filter filter, db::timeout_clock::time_point timeout) {
         return _impl->consume_in_thread(std::move(consumer), std::move(filter), timeout);
@@ -373,7 +419,7 @@ public:
         requires FlattenedConsumer<Consumer>()
     )
     auto consume_in_thread(Consumer consumer, db::timeout_clock::time_point timeout) {
-        return consume_in_thread(std::move(consumer), [] (const dht::decorated_key&) { return true; }, timeout);
+        return consume_in_thread(std::move(consumer), no_filter{}, timeout);
     }
 
     // Skips to the next partition.
--- a/flat_mutation_reader.hh
+++ b/flat_mutation_reader.hh
@@ -731,3 +731,16 @@ future<> consume_partitions(flat_mutation_reader& reader, Consumer consumer, db:
 
 flat_mutation_reader
 make_generating_reader(schema_ptr s, std::function<future<mutation_fragment_opt> ()> get_next_fragment);
+
+// Track position_in_partition transitions and validate monotonicity.
+class mutation_fragment_stream_validator {
+    const schema& _schema;
+    mutation_fragment::kind _prev_kind;
+    position_in_partition _prev_pos;
+public:
+    mutation_fragment_stream_validator(const schema& s);
+    ~mutation_fragment_stream_validator();
+
+    bool operator()(const dht::decorated_key& dk);
+    bool operator()(const mutation_fragment& mv);
+};
--- a/flat_mutation_reader.hh
+++ b/flat_mutation_reader.hh
@@ -737,8 +737,10 @@ class mutation_fragment_stream_validator {
     const schema& _schema;
     mutation_fragment::kind _prev_kind;
     position_in_partition _prev_pos;
+    partition_region _prev_region;
+    bool _compare_keys;
 public:
-    mutation_fragment_stream_validator(const schema& s);
+    mutation_fragment_stream_validator(const schema& s, bool compare_keys = false);
     ~mutation_fragment_stream_validator();
 
     bool operator()(const dht::decorated_key& dk);
--- a/position_in_partition.hh
+++ b/position_in_partition.hh
@@ -251,6 +251,17 @@ public:
             }
         }
 
+    position_in_partition& operator=(position_in_partition_view view) {
+        _type = view._type;
+        _bound_weight = view._bound_weight;
+        if (view._ck) {
+            _ck = *view._ck;
+        } else {
+            _ck.reset();
+        }
+        return *this;
+    }
+
     static position_in_partition before_all_clustered_rows() {
         return {position_in_partition::range_tag_t(), bound_view::bottom()};
     }
--- a/sstables/compaction.cc
+++ b/sstables/compaction.cc
@@ -461,7 +461,7 @@ private:
         };
     }
 
-    virtual flat_mutation_reader::filter filter_func() const {
+    virtual flat_mutation_reader::filter make_partition_filter() const {
         return [] (const dht::decorated_key&) {
             return true;
         };
@@ -585,7 +585,7 @@ public:
         };
     }
 
-    virtual flat_mutation_reader::filter filter_func() const override {
+    virtual flat_mutation_reader::filter make_partition_filter() const override {
         return [] (const dht::decorated_key& dk){
             return dht::shard_of(dk.token()) == engine().cpu_id();
         };
@@ -724,7 +724,7 @@ public:
         clogger.info("Cleaned {}", formatted_msg);
     }
 
-    flat_mutation_reader::filter filter_func() const override {
+    flat_mutation_reader::filter make_partition_filter() const override {
         dht::token_range_vector owned_ranges = service::get_local_storage_service().get_local_ranges(_schema->ks_name());
 
         return [this, owned_ranges = std::move(owned_ranges)] (const dht::decorated_key& dk) {
@@ -876,7 +876,7 @@ future<compaction_info> compaction::run(std::unique_ptr<compaction> c) {
             // leave this block either successfully or exceptionally with the reader object
             // destroyed.
             auto r = std::move(reader);
-            r.consume_in_thread(std::move(cfc), c->filter_func(), db::no_timeout);
+            r.consume_in_thread(std::move(cfc), c->make_partition_filter(), db::no_timeout);
         } catch (...) {
             c->delete_sstables_for_interrupted_compaction();
             c = nullptr; // make sure writers are stopped while running in thread context
--- a/sstables/compaction.cc
+++ b/sstables/compaction.cc
@@ -461,7 +461,7 @@ private:
         };
     }
 
-    virtual std::function<bool(const dht::decorated_key&)> filter_func() const {
+    virtual flat_mutation_reader::filter filter_func() const {
         return [] (const dht::decorated_key&) {
             return true;
         };
@@ -585,7 +585,7 @@ public:
         };
     }
 
-    virtual std::function<bool(const dht::decorated_key&)> filter_func() const override {
+    virtual flat_mutation_reader::filter filter_func() const override {
         return [] (const dht::decorated_key& dk){
             return dht::shard_of(dk.token()) == engine().cpu_id();
         };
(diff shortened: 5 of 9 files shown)


position_in_partition.cc
```diff
@@ -0,0 +1,98 @@
...
+#include "position_in_partition.hh"
+#include "utils/exceptions.hh"
+#include "log.hh"
+
+logging::logger plogger("position_in_partition");
+
+const sstring position_in_partition_tracker::current_position_string() {
+    if (is_beginning_of_stream()) {
+        return "beginning of stream";
+    }
```
denesb: A switch would be nicer.

position_in_partition.cc
```diff
@@ -0,0 +1,98 @@
...
+        return "end of stream";
+    }
+    return "unrecognized position_in_partition";
+}
+
+void position_in_partition_tracker::on_partition_start() {
+    plogger.trace("partition start");
+    if (is_partition_end() || is_beginning_of_stream()) {
+        _pos = partition_region::partition_start;
+        _in_stream = true;
```
denesb: Maybe it would be better to encode this additional bit by making `_pos` an `std::optional`.

position_in_partition.cc
```diff
@@ -0,0 +1,98 @@
...
+    } else {
+        on_internal_error(plogger, format("New partition must not start after {}", current_position_string()));
```
denesb: Better as: "Unexpected partition-region: expected {}, got {} instead".
Same for the other messages below.

position_in_partition.cc
```diff
@@ -0,0 +1,98 @@
...
+    }
+    if (is_end_of_stream()) {
+        return "end of stream";
+    }
+    return "unrecognized position_in_partition";
+}
+
+void position_in_partition_tracker::on_partition_start() {
+    plogger.trace("partition start");
+    if (is_partition_end() || is_beginning_of_stream()) {
```
denesb: Maybe we should annotate these conditions with `__builtin_likely()`

position_in_partition.hh
```diff
@@ -676,3 +676,41 @@ bool position_range::overlaps(const schema& s, position_in_partition_view start,
...
+    bool _in_stream = false;
+    partition_region _pos = partition_region::partition_start;
+public:
+    void on_partition_start();
+    void on_static_row();
+    void on_clustering_row();
+    void on_partition_end();
+    void on_end_of_stream();
+
+    bool in_stream() {
```
denesb: const

The same for the other accessors below.

sstables/writer_impl.hh
```diff
@@ -32,12 +33,14 @@ struct sstable_writer::writer_impl {
...
     const io_priority_class& _pc;
     const sstable_writer_config _cfg;
+    position_in_partition_tracker _pip_tracker;
 
     writer_impl(sstable& sst, const schema& schema, const io_priority_class& pc, const sstable_writer_config& cfg)
         : _sst(sst)
         , _schema(schema)
         , _pc(pc)
         , _cfg(cfg)
+        , _pip_tracker()
```
denesb: Not needed, the compiler will call this automatically.

position_in_partition.hh
```diff
@@ -676,3 +676,41 @@ bool position_range::overlaps(const schema& s, position_in_partition_view start,
     position_in_partition::less_compare less(s);
     return !less(end, _start) && less(start, _end);
 }
+
+// Track position_in_partittion transitions
+// and validate monotonicity.
+class position_in_partition_tracker {
```
denesb: Perhaps `_tracker` is not the right name. Don't have a better one though.

position_in_partition.cc
bhalevy: You mean a switch on the `partition_region`?

position_in_partition.cc
bhalevy: I tried an early version with this method.  The problem was that there is no distinction between beginning vs. end of stream this way and there is a need for another bool for end_of_stream so the end result is not much different.

position_in_partition.cc
bhalevy: Ok, that seems useful.

position_in_partition.cc
denesb: Yes.

position_in_partition.hh
tgrabiec: I suggest to make it a `Filter` (as defined by flat_mutation_reader.hh).

This way you only need to implement:

```
bool operator()(const mutation_fragment& mf);
bool operator()(const dht::decorated_key& dk);
```

And work in terms of `mf.position()`. No need to even know what are the possible fragment types (see https://github.com/tgrabiec/scylla/commit/fa53994fb04697b3e1e60060fe62122fc577b106).

In sstables.cc, call `consume_in_thread()` variant which accepts a `Filter`.

position_in_partition.cc
bhalevy: It's possible, though I saw no noticeable impact on performance.
Still it won't hurt (except some reduced readability of the code)

position_in_partition.hh
bhalevy: Right.  Thanks.

sstables/writer_impl.hh
bhalevy: It's a placeholder for the follow up patchset :)
But you're right, it could be dropped here and added later when needed.

position_in_partition.hh
denesb: Isn't this a lot more heavy-weight (also compares clustering keys)? Or is this something we want to do eventually anyway? In that case it is indeed better to just compare positions.

Even if so, I think this is stretching the filter concept a bit far.

position_in_partition.cc
```diff
@@ -0,0 +1,98 @@
...
+#include <seastar/core/print.hh>
+
+#include "position_in_partition.hh"
+#include "utils/exceptions.hh"
+#include "log.hh"
+
+logging::logger plogger("position_in_partition");
+
+const sstring position_in_partition_tracker::current_position_string() {
+    if (is_beginning_of_stream()) {
```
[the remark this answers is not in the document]
bhalevy: I contemplated to extend `enum partition_region` by adding `stream_begin` and `end_of_stream`.  Does that what you mean?

position_in_partition.hh
bhalevy: Hmm, sounds intriguing...

tgrabiec: @denesb You can compare just the `region()` of a position. You still don't need to care about fragment types.

What's the problem with using a filter? It would be a filter, one which doesn't filter anything, but still, it's a filter. The advantage is that we reuse the existing infrastructure. We avoid the risk of missing a callback, plus the patch is smaller.

position_in_partition.cc
denesb: Yes, it might be premature optimization.

position_in_partition.hh
denesb: Right. It would indeed be much simpler.

position_in_partition.cc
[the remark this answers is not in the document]
tgrabiec: No, stream begin is not a partition region, it's a stream region. So for stream begin and end of stream we need separate flags. `stream_begin` can be encoded in `std::optional` in `std::optional<region>`.

It's still good to reuse the region, you don't need to worry about what kind of fragments are there within the partition.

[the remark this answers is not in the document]
bhalevy: > you don't need to worry about what kind of fragments are there within the partition.

But we would like to enforce singularity of static rows and allow plurality of clustering rows.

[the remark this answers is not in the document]
tgrabiec: That is already enforced by requiring strict ordering on positions. Duplicated static rows would violate "<". We want the same for clustering rows. Range tombstones are exceptions to the rule, we allow them to have equal positions one after the other.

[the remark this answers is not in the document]
tgrabiec: But if we're only looking at region then you're right, we need to distinguish them.

position_in_partition.cc
bhalevy: change in v2 to use operator<

position_in_partition.cc
bhalevy: dropped in v2

position_in_partition.cc
bhalevy: Rephrased in v2.

position_in_partition.hh
bhalevy: Dropped in v2.

sstables/writer_impl.hh
bhalevy: Dropped in v2.

position_in_partition.hh
bhalevy: Renamed in v2.

position_in_partition.cc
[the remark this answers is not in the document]
bhalevy: Reimplemented in v2.  Supports both region based comparison and full position_in_partition, including clustering key.

position_in_partition.hh
bhalevy: Using a Filter class in v2 as @tgrabiec suggested.

bhalevy:
In v2.:
1. added "on_internal_error: flush standard output before aborting"
2. flat_mutation_reader: define filter class and use in compaction
3. tracker renamed to `mutation_fragment_stream_validator`
    and moved from position_in_partition to flat_mutation_reader.{cc,hh}
4. compare_keys flag to compare either only partition_region or also clustering keys monotonicity (by comparing completely constructed position_in_partition)

TODO: validate partition keys monotonicity.

Test: unit(dev)

utils/exceptions.cc
```diff
@@ -78,6 +78,8 @@ void set_abort_on_internal_error(bool do_abort) {
 void on_internal_error(seastar::logger& logger, const seastar::sstring& msg) {
     if (abort_on_internal_error.load()) {
         logger.error("{}, at: {}", msg.c_str(), seastar::current_backtrace());
+        std::cout.flush();
+        std::cerr.flush();
```
denesb: How about when we use the syslog backend? Or does that also use `std::cout/std::cerr`? Maybe we should have a `logger.flush()` or something similar.

sstables/compaction.cc
```diff
@@ -725,21 +727,19 @@ class cleanup_compaction final : public regular_compaction {
...
 
-        return [this, owned_ranges = std::move(owned_ranges)] (const dht::decorated_key& dk) {
-            if (dht::shard_of(dk.token()) != engine().cpu_id()) {
-                clogger.trace("Token {} does not belong to CPU {}, skipping", dk.token(), engine().cpu_id());
-                return false;
-            }
+        if (dht::shard_of(dk.token()) != engine().cpu_id()) {
+            clogger.trace("Token {} does not belong to CPU {}, skipping", dk.token(), engine().cpu_id());
+            return false;
+        }
```
denesb: This code reindenting should have been done in a separate patch.

flat_mutation_reader.hh
```diff
@@ -360,6 +360,19 @@ public:
         return _impl->consume(std::move(consumer), timeout);
     }
 
+    class filter {
```
denesb: What is the advantage of this class? It looks like a mere wrapper over `std::function`.

flat_mutation_reader.hh
```diff
@@ -200,7 +205,7 @@ public:
                     next_partition();
                     continue;
                 }
-                if (consumer(std::move(mf)) == stop_iteration::yes) {
+                if (filter(mf) && (consumer(std::move(mf)) == stop_iteration::yes)) {
```
denesb: This is not a filter, but a stop condition.

flat_mutation_reader.hh
```diff
@@ -61,6 +61,11 @@ GCC6_CONCEPT(
     concept bool PartitionFilter = requires(T filter, const dht::decorated_key& dk) {
         { filter(dk) } -> bool;
     };
+
+    template<typename T>
+    concept bool MutationFilter = requires(T filter, const mutation_fragment& mf) {
```
tgrabiec: `MutationFilter` is already defined in mutation_reader.hh as a different thing.

position_in_partition.hh
```diff
@@ -251,6 +251,17 @@ public:
             }
         }
 
+    position_in_partition& operator=(position_in_partition_view view) {
```
tgrabiec: Should be in a separate patch

flat_mutation_reader.cc
```diff
@@ -912,3 +915,47 @@ class generating_reader final : public flat_mutation_reader::impl {
...
+
+bool mutation_fragment_stream_validator::operator()(const dht::decorated_key& dk) {
+    // FIXME: Validate fragment position monotonicity
+    //        https://github.com/scylladb/scylla/issues/4804
+    return true;
+}
+
+template<typename T>
+sstring position_desc(mutation_fragment::kind kind, const T& pos) {
+    return format("{}{}", kind == mutation_fragment::kind::range_tombstone ? "range_tombstone: " : "", pos);
```
tgrabiec: Why not print the `kind` unconditionally, using its natural `operator<<`?

It's not clear what T is. I think you want to accept `const position_in_partition_view&`.

flat_mutation_reader.cc
```diff
@@ -912,3 +915,47 @@ class generating_reader final : public flat_mutation_reader::impl {
...
+        on_internal_error(fmr_logger, format("Stream ended with unclosed partition: {}", position_desc(_prev_kind, _prev_pos)));
+    }
+}
+
+bool mutation_fragment_stream_validator::operator()(const mutation_fragment& mv) {
+    auto kind = mv.mutation_fragment_kind();
+    auto pos = mv.position();
+    bool valid;
+
+    fmr_logger.debug("{} -> {}", position_desc(_prev_kind, _prev_pos), position_desc(kind, pos));
```
tgrabiec: `position_desc()` will be invoked even if the log level is disabled, which adds performance overhead. We should only pass printable objects here.

flat_mutation_reader.cc
```diff
@@ -936,26 +936,35 @@ mutation_fragment_stream_validator::~mutation_fragment_stream_validator() {
...
     if (__builtin_expect(!valid, false)) {
         on_internal_error(fmr_logger, format("Unexpected mutation fragment: previous: {}, got: {}", position_desc(_prev_kind, _prev_pos), position_desc(kind, pos)));
     }
 
     _prev_kind = mv.mutation_fragment_kind();
-    _prev_pos = pos;
+    if (_compare_keys) {
+        _prev_pos = pos;
+    } else {
+        _prev_pos = position_in_partition(region, pos.get_bound_weight(), std::optional<clustering_key_prefix>());
```
tgrabiec: This is a bit hacky since `_prev_pos` created this way is not really valid. Better to store the region in a separate member when tracking just the region.

flat_mutation_reader.cc
denesb: An alternative is to use `seastar::value_of()` which is evaluted only if the log line is actually printed.

utils/exceptions.cc
bhalevy: Sent scylladb/seastar#675 to define logger::flush and call it in the seastar abort signal handler path so this won't be needed.

sstables/compaction.cc
bhalevy: Will fix in v3.

flat_mutation_reader.hh
bhalevy: As @tgrabiec suggested, it eventually defines to predicates, one for `dht::decorated_key` and the other for `mutation_fragment`.  This way we can still pass one filter to `consume_in_thread` and friends rather than two.

flat_mutation_reader.hh
bhalevy: Hmm, the stop condition was there before.  With the filter in the expression, `consumer(std::move(mf))` will not be called if the filter returns `false` and the mutation fragment will not be consumed at all.
That said, it is true though that at this time we have no use for this method as a conditional filter and it could return void.

flat_mutation_reader.hh
bhalevy: Moved to `flat_mutation_reader.hh` and used by both in next version.

position_in_partition.hh
bhalevy: fixed in next version

flat_mutation_reader.hh
denesb: More than that, stopping in the middle of a partition is not legal, so we cannot stop here on any fragment.

flat_mutation_reader.cc
bhalevy: I'll try printing kind unconditionally as you suggested.
T could be either `position_in_partition` or position_in_partition_view.
The former can be converted to a view of course.

flat_mutation_reader.cc
bhalevy: Or maybe surround with `if (fmr_logger.is_enabled(log_level::debug))` but that might be too low level.

bhalevy: And as @tgrabiec suggested if we pass printable objects then their operator<<  will be called only if the log line is being printed.

flat_mutation_reader.cc
bhalevy: OK, no problem.  A bit of waste of memory, but a small one.

flat_mutation_reader.hh
bhalevy: @denesb The consumer can stop after `partition_end`, no?

denesb: Yes, *only* after `partition_end`.

BTW my top comment here refers to the fact that this filter is now is in very not filter-y roles. It now determines when consuming stops, as opposed to allowing/dropping fragments or partitions.

flat_mutation_reader.cc
bhalevy: @tgrabiec also, if it's illegal to have a clustered position_in_partition{,_view} with no clustering key (well, it doesn't make much sense), it might be a good idea to assert that upon construction.

bhalevy:
In v3:
- extracted "compaction: fix indentation of filter_func methods"
- "mutation_reader: reuse flat_mutation_reader PartitionFilter concept" +
  "mutation_reader: s/MutationFilter/Filter/"
- extracted "position_in_partition: define operator=(position_in_partition_view)"
- mutation_fragment_stream_validator:
  - moved constructor to `flat_mutation_reader.cc`
  - unconditionally print `kind` in debug and error messages.
  - pass printable objects to `logger.debug` to reduce overhead when default-log-level is higher than debug. 
  - use `_prev_kind` to validate partition_start following partition_end so it could be used
    also when validating `partition_region` only
- "mutation_fragment: declare partition_region operator<< in header file"
  - to be used when `compare_keys` is false
- "mutation_fragment_stream_validator: add compare_keys flag":
  - keep region in `_prev_region` member rather than in (invalid) `_prev_pos`
- Added "config: enable_sstable_key_validation by default in debug build"
  - To increase coverage in debug testing.

flat_mutation_reader.hh
bhalevy: Previously the `consumer` could return `stop_iteration::yes` on any `mutation_fragment`.  Doing so after mutation_fragments that aren't `partition_end` is a bug indeed but it was possible before, and will be detected in `~mutation_fragment_stream_validator()`.
Now,  stopping the stream would require both returning `true` by the filter and `stop_iteration::yes` from the `consumer`.  If the mutation filter returns `false`, the mutation_fragment will be skipped (dangerous indeed) and the stream will not stop.

The mutation_fragment can't be both skipped and consumed (for stopping consumption), that doesn't make sense at all.  As I said, we can have the mutation filter return `void` rather than `bool` and consume all mutation_fragments with no change with that regards..

sstables/compaction.cc
```diff
@@ -462,10 +462,14 @@ class compaction {
         };
     }
 
-    virtual std::function<bool(const dht::decorated_key&)> filter_func() const {
-        return [] (const dht::decorated_key&) {
+    virtual bool filter_func(const dht::decorated_key&) const {
```
denesb: The name isn't really meaningful. Maybe 'consider_partition()`, or something similar that expresses the purpose of this function, which is whether the compaction even considers that partition for the output or not.

flat_mutation_reader.hh
```diff
@@ -360,6 +360,19 @@ public:
         return _impl->consume(std::move(consumer), timeout);
     }
 
+    class filter {
+    private:
+        std::function<bool (const dht::decorated_key&)> partition_filter;
+    public:
+        filter(std::function<bool (const dht::decorated_key&)>&& pf = [] (const dht::decorated_key&) { return true; })
```
denesb: Maybe it should also have a forwarding constructor:
```
template <typename Functor>
filter(Functor&& f) : partition_filter(std::forward<Functor>(f)) { }
```
This way existing code would not have to be changed, the lambdas they pass in would implicitly be converted to this filter.

flat_mutation_reader.hh
denesb: This looks like a `MutationFragmentFilter`. `Mutation` represents a partition.

flat_mutation_reader.hh
denesb: I just realized this is `consume_pausable()`, where you can stop anywhere you want, the assumption is that you are not stopping for good, just "pausing". Calling this method again, you will continue where you left off previously.

So when the filter rejects a fragment, it just throws out the consumer from the consumption loop. The consumer then might call `consume_pausable()` again, essentially just skipping a fragment. Or it might consider the consumption to be over. In both cases it can end up with an invalid fragment stream without even knowing it.

I don't know what to think about this. On one hand its great that we are reusing the existing filter and consume infrastructure for validation. On the other hand we are just about to build into it a huge gun aimed at our feet.

flat_mutation_reader.hh
```diff
@@ -363,19 +368,26 @@ public:
     class filter {
     private:
         std::function<bool (const dht::decorated_key&)> partition_filter;
+        std::function<bool (const mutation_fragment&)> mutation_filter;
     public:
-        filter(std::function<bool (const dht::decorated_key&)>&& pf = [] (const dht::decorated_key&) { return true; })
+        filter(std::function<bool (const dht::decorated_key&)>&& pf = [] (const dht::decorated_key&) { return true; },
+               std::function<bool (const mutation_fragment&)>&& mf = [] (const mutation_fragment&) { return true; })
             : partition_filter(std::move(pf))
+            , mutation_filter(std::move(mf))
```
denesb: Again, this term tells me this is filtering partitions also. `mutation` == `partition`.

flat_mutation_reader.cc
```diff
@@ -912,3 +915,48 @@ class generating_reader final : public flat_mutation_reader::impl {
...
+mutation_fragment_stream_validator::~mutation_fragment_stream_validator() {
+    if (_prev_kind != mutation_fragment::kind::partition_end) {
+        on_internal_error(fmr_logger, format("Stream ended with unclosed partition: {}", _prev_kind));
+    }
+}
+
+bool mutation_fragment_stream_validator::operator()(const mutation_fragment& mv) {
+    auto kind = mv.mutation_fragment_kind();
+    auto pos = mv.position();
+    bool valid;
```
denesb: I think it would be best to initialize this, to avoid UB due to mistakes.

flat_mutation_reader.hh
tgrabiec: @denesb When the filter rejects a fragment the consumption loop is not terminated, it continues. It's equivalent to consuming a filtered stream:

`(read() | filter()).consume_pausable_in_thread()`

I don't understand what's the objection for this.

denesb: It is terminated:
```
if (filter(mf) && (consumer(std::move(mf)) == stop_iteration::yes))
```

For the consumption loop to continue the filter has to return true and the consumer has to opt to continue. If the filter rejects the fragment the consumer will not even see it and the consumption loop will terminate.

denesb: I was wrong, its the other way around. Both have to opt in for the consumption to terminate. I guess that is fine then.

denesb: Wait, shouldn't this be
```
if (!filter(mf) && (consumer(std::move(mf)) == stop_iteration::yes))
```
then? `true` means the fragment passed the filter, no? In the current form the consumption will break when the fragment passes the filter *and* the consumer wants to stop.

flat_mutation_reader.cc
```diff
@@ -912,3 +915,65 @@ class generating_reader final : public flat_mutation_reader::impl {
 flat_mutation_reader make_generating_reader(schema_ptr s, std::function<future<mutation_fragment_opt> ()> get_next_fragment) {
     return make_flat_mutation_reader<generating_reader>(std::move(s), std::move(get_next_fragment));
 }
+
+bool mutation_fragment_stream_validator::operator()(const dht::decorated_key& dk) {
+    // FIXME: Validate fragment position monotonicity
+    //        https://github.com/scylladb/scylla/issues/4804
```
denesb: Why is this FIXME here? Isn't this done below already?

flat_mutation_reader.hh
bhalevy: Right. If the filter rejects the fragment, then it won't be passed to the consumer and the loop will continue. What's wrong with that?

sstables/compaction.cc
bhalevy: I agree. It is a bit out of scope of this patchset but I can squeeze that in the next version while at it.

flat_mutation_reader.hh
bhalevy: Good idea, thanks. Will add it.

flat_mutation_reader.hh
bhalevy: Ok. Will change.

flat_mutation_reader.hh
bhalevy: Will rename to `mutation_fragment_filter`

flat_mutation_reader.cc
bhalevy: I beg to differ. We actually want the compiler to warn us if there's a path that doesn't set it.  Initializing it will prevent that. Ideally it could be initialized to `false` only for non-debug builds.

flat_mutation_reader.hh
tgrabiec: I think Benny's code is correct, if `!filter(mf)` then the fragment is treated as if it didn't exist, so we don't call the consumer, and just keep looping. If `filter(mf) == true`, we execute whatever logic we had before for that fragment, cause it exists in the consumer's input stream.

flat_mutation_reader.cc
bhalevy: The issue also calls for partition key monotonicity validation (taking into account the partitioner of course).  The comment is wrong (copy-pasted). Will fix.

sstables/compaction.cc
bhalevy: @denesb now with the forwarding constructor there is no change in this code so it's hard to justify this rename as part of this patchset.  For the sake of backportability, let's keep it minimal and I'll send a followup PR with the cosmetic improvements (for master, not to be backported).

flat_mutation_reader.hh
denesb: Ok. Fair enough.

flat_mutation_reader.cc
denesb: Does it warn for that? I though these bugs can only be caught at runtime.
If there is a warning for that then sure, no need to initialize it defensively.

bhalevy: Well, `-Wmaybe-uninitialized` does that but we disable it.  IIRC it produces too many false warnings.  I'll add the initialization then.

sstables/compaction.cc
bhalevy: hmm, unfortunately the `filter_func` signature does need to change apparently.  So I'll include the renaming patch after all.

flat_mutation_reader.cc
denesb: I think it was disabled due to bad interactions with `std::optional`. I'm not sure anymore. I think, from time-to-time we should go through our disabled warning list and see if they can be re-enabled.

bhalevy:
In v4:
- Added forwarding constructor to class filter so that existing call sites won't have to be changed.
  - Dropped respective change to class compaction and in flat_mutation_reader_test.
  - Dropped following indentation fixes to compaction.
  - Instead, compaction `filter_func` methods need to return a filter.
  - "compaction: s/filter_func/consider_partition/" as per Botond's suggestion.
- Renamed (new) `MutationFilter` to `MutationFragmentFilter` (and `mutation_filter` to `mutation_fragment_filter`, respectively)
  - Dropped "mutation_reader: reuse flat_mutation_reader PartitionFilter concept" and "mutation_reader: s/MutationFilter/Filter/" as they are not required now (will send as follow up)
- fixed FIXME comment in `mutation_fragment_stream_validator::operator()(const dht::decorated_key& dk)`
- Added initialization for `valid` in mutation_fragment_stream_validator::operator()(const mutation_fragment& mv)

db/config.cc
```diff
@@ -155,6 +155,12 @@ struct convert<db::config::seed_provider_type> {
 
 }
 
+#if defined(CONFIG_ENABLE_SSTABLE_KEY_VALIDATION) || defined(DEBUG)
```
denesb: What is this `CONFIG_ENABLE_SSTABLE_KEY_VALIDATION`? Is this a way to force validation at build time?

bhalevy:
On Fri, Aug 23, 2019, 17:50 Botond Dénes <notifications@github.com> wrote:

> *@denesb* commented on this pull request.
>
> In db/config.cc
> <https://github.com/scylladb/scylla/pull/4846#discussion_r317168260>:
>
> > @@ -155,6 +155,12 @@ struct convert<db::config::seed_provider_type> {
>
>  }
>
> +#if defined(CONFIG_ENABLE_SSTABLE_KEY_VALIDATION) || defined(DEBUG)
>
> What is this CONFIG_ENABLE_SSTABLE_KEY_VALIDATION? Is this a way to force
> validation at build time?
>

Yeah, we may want to do that for a subset of the unit tests also in
non-debug modes.

bhalevy:
@tgrabiec ping

flat_mutation_reader.hh
```diff
@@ -373,7 +388,7 @@ public:
         requires FlattenedConsumer<Consumer>()
     )
     auto consume_in_thread(Consumer consumer, db::timeout_clock::time_point timeout) {
-        return consume_in_thread(std::move(consumer), [] (const dht::decorated_key&) { return true; }, timeout);
+        return consume_in_thread(std::move(consumer), filter(), timeout);
```
tgrabiec: Would this mean we'll be paying for virtual dispatch inside `filter::partition_filer::operator()`? Perhaps we have a `no_filter` object with non-virtual methods which can be fully inlined.

flat_mutation_reader.cc
```diff
@@ -912,3 +915,48 @@ class generating_reader final : public flat_mutation_reader::impl {
...
+        on_internal_error(fmr_logger, format("Stream ended with unclosed partition: {}", _prev_kind));
+    }
+}
+
+bool mutation_fragment_stream_validator::operator()(const mutation_fragment& mv) {
+    auto kind = mv.mutation_fragment_kind();
+    auto pos = mv.position();
+    bool valid = false;
+
+    fmr_logger.debug("{}:{}", kind, pos);
```
tgrabiec: Since there could be many validators active at any time, you should distinguish them. Typically we prefix the line with the value of `this`.

sstables/compaction.cc
```diff
@@ -462,7 +462,7 @@ class compaction {
         };
     }
 
-    virtual flat_mutation_reader::filter filter_func() const {
+    virtual flat_mutation_reader::filter consider_partition() const {
```
tgrabiec: To me, the new name seems confusing because it's a verb which describes an action which the returned filter does, not what this method does. `make_partion_filter()` / `partition_filter()` seems better.

It's like having `stop()` which doesn't stop but returns a functor which stops.

flat_mutation_reader.hh
```diff
@@ -363,21 +368,37 @@ public:
     class filter {
     private:
         std::function<bool (const dht::decorated_key&)> partition_filter = [] (const dht::decorated_key&) { return true; };
+        std::function<bool (const mutation_fragment&)> mutation_fragment_filter = [] (const mutation_fragment&) { return true; };
```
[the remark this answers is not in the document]
bhalevy: Sure. will fix in v5.

flat_mutation_reader.hh
bhalevy: Not exactly, just calling indirectly via a pointer.  But we avoid this as well with a `no_filter` object as you suggested.

flat_mutation_reader.cc
bhalevy: ok, will fix in v5.

sstables/compaction.cc
bhalevy: makes sense.  I'll go with `make_partition_filter`.

bhalevy:
In v5:
- class filter: prefix members with `_`
- Define flat_mutation_reader::no_filter that can be
  fully inlined.
- Squashed first 2 patches together:
  
- compaction: rename filter_func to make_partition_filter
  rather than `consider_partition`. 
- Do not use CONFIG_ENABLE_SSTABLE_KEY_VALIDATION yet.
  - Should be defined in the future if and when used.
- Print value of `this` in validator debug message.

sstables/sstables.cc
```diff
@@ -2412,7 +2412,8 @@ future<> sstable::write_components(
     }
     return seastar::async([this, mr = std::move(mr), estimated_partitions, schema = std::move(schema), cfg, stats, &pc] () mutable {
         auto wr = get_writer(*schema, estimated_partitions, cfg, stats, pc);
-        mr.consume_in_thread(std::move(wr), db::no_timeout);
```
tgrabiec: I think the most important part of the change is the fact that the sstable writer now validates the input mutation fragment stream. This cannot be deduced from the commit message ("flat_mutation_reader: mutation_fragment_stream_validator"). I would suggest "sstables: writer: Validate input mutation fragment stream". The addition of the validator could be a separate patch, but I am not going to insist on that.

flat_mutation_reader.cc
```diff
@@ -922,11 +922,15 @@ bool mutation_fragment_stream_validator::operator()(const dht::decorated_key& dk
...
-mutation_fragment_stream_validator::mutation_fragment_stream_validator(const schema& s)
+mutation_fragment_stream_validator::mutation_fragment_stream_validator(const schema& s, bool compare_keys)
     : _schema(s)
     , _prev_kind(mutation_fragment::kind::partition_end)
     , _prev_pos(position_in_partition::end_of_partition_tag_t{})
-{ }
+    , _prev_region(partition_region::partition_end)
+    , _compare_keys(compare_keys)
+{
+    fmr_logger.debug("mutation_fragment_stream_validator: will validate {} monotonicity.", compare_keys ? "keys" : "only partition regions");
```
tgrabiec: Missing `this` prefix.

flat_mutation_reader.cc
```diff
@@ -937,26 +941,39 @@ mutation_fragment_stream_validator::~mutation_fragment_stream_validator() {
...
         }
+    } else if (region != partition_region::clustered) {
+        valid = (_prev_region < region);
+    } else {
+        valid = (_prev_region <= region);
     }
 
     if (__builtin_expect(!valid, false)) {
-        on_internal_error(fmr_logger, format("Unexpected mutation fragment: previous {}:{}, current {}:{}", _prev_kind, _prev_pos, kind, pos));
+        auto fmt = "Unexpected mutation fragment: previous {}:{}, current {}:{}";
```
tgrabiec: Missing `this` prefix for correlation with earlier log messages.

sstables/sstables.cc
bhalevy: sure, makes sense.

flat_mutation_reader.cc
bhalevy: ok, will fix in next version.

flat_mutation_reader.cc
bhalevy: ok, will fix in next version.

bhalevy:
In v6:
- renamed "flat_mutation_reader: mutation_fragment_stream_validator" patch to "sstables: writer: Validate input mutation fragment stream"
- Added "[validator {}] ... this" prefix to all `mutation_fragment_stream_validator` debug and error messages.

tgrabiec:
The title of the commit "compaction: s/filter_func/consider_partition/" is no longer valid.

tgrabiec:
Otherwise looks good.

bhalevy:
In v7:
- updated "compaction: s/filter_func/consider_partition/" to "compaction: s/filter_func/make_partition_filter/"

bhalevy:
@tgrabiec please re-review

PR #4848, opened 2019-08-14
repair: Add ranges_parallelism option

### LINKED ISSUE #4847: repair: Add ranges_parallelism option
In commit 131acc09cc1d568f994059736c8500adf5d05d88 (repair: Adjust
parallelism according to memory size), we choose the ranges to repair in
parallel according to the memory size automatically.

However, this automatic number might not the optimal one. For example,
user wants repair to have minimal impact on the user cql reads and
writes, or user wants to repair at full speed allowing more cpu and
memory resources.

We can add an option that controls number of ranges that we repair 
in parallel, so that the advanced user can control the parallelism.

asias:
In commit 131acc09cc1d568f994059736c8500adf5d05d88 (repair: Adjust
parallelism according to memory size), we choose the ranges to repair in
parallel according to the memory size automatically.

However, this automatic number might not the optimal one. For example,
user wants repair to have minimal impact on the user cql reads and
writes, or user wants to repair at full speed allowing more cpu and
memory resources.

This patch introduces a ranges_parallelism option that controls number
of ranges that we repair in parallel, so that the advanced
user can control the parallelism.

Here is an example:

In a cluster with high network latency, e.g., multiple DCs with high
latency link, user can increase ranges_parallelism to compensate.

For example in two DCs cluster with 7 nodes + 60ms round trip time + RF
(dc1:3, dc2:4) + a keyspace with 3 empty tables:

2019-04-10 11:00:03.303549 Repair with ks ranges_parallelism=1 on node 1 started ...
2019-04-10 11:32:32.673043 Repair with ks on node 1 finished ...

2019-04-10 11:32:32.673292 Repair with ks ranges_parallelism=16 on node 1 started ...
2019-04-10 11:36:49.475977 Repair with ks on node 1 finished ...

2019-04-10 11:36:49.476145 Repair with ks ranges_parallelism=256 on node 1 started ...
2019-04-10 11:38:04.242553 Repair with ks on node 1 finished ...

That is 1949s vs 257s vs 75s to complete repair respectively, which
gives 7X and 25X difference.

Fixes #4847

### DIFF
--- a/api/api-doc/storage_service.json
+++ b/api/api-doc/storage_service.json
@@ -917,6 +917,14 @@
                      "allowMultiple":false,
                      "type":"string",
                      "paramType":"query"
+                  },
+                  {
+                     "name":"ranges_parallelism",
+                     "description":"An integer specifying the number of ranges to repair in parallel.",
+                     "required":false,
+                     "allowMultiple":false,
+                     "type":"string",
+                     "paramType":"query"
                   }
                ]
             },
--- a/api/storage_service.cc
+++ b/api/storage_service.cc
@@ -376,7 +376,7 @@ void set_storage_service(http_context& ctx, routes& r) {
     ss::repair_async.set(r, [&ctx](std::unique_ptr<request> req) {
         static std::vector<sstring> options = {"primaryRange", "parallelism", "incremental",
                 "jobThreads", "ranges", "columnFamilies", "dataCenters", "hosts", "trace",
-                "startToken", "endToken" };
+                "startToken", "endToken", "ranges_parallelism" };
         std::unordered_map<sstring, sstring> options_map;
         for (auto o : options) {
             auto s = req->get_query_param(o);
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -642,7 +642,8 @@ repair_info::repair_info(seastar::sharded<database>& db_,
     const std::vector<sstring>& cfs_,
     int id_,
     const std::vector<sstring>& data_centers_,
-    const std::vector<sstring>& hosts_)
+    const std::vector<sstring>& hosts_,
+    int ranges_parallelism_)
     : db(db_)
     , keyspace(keyspace_)
     , ranges(ranges_)
@@ -651,6 +652,8 @@ repair_info::repair_info(seastar::sharded<database>& db_,
     , shard(engine().cpu_id())
     , data_centers(data_centers_)
     , hosts(hosts_)
+    , ranges_parallelism(ranges_parallelism_)
+    , ranges_parallelism_semaphore(ranges_parallelism == -1 ? std::nullopt : std::optional<semaphore>(ranges_parallelism))
     , _row_level_repair(service::get_local_storage_service().cluster_supports_row_level_repair()) {
 }
 
@@ -1112,6 +1115,8 @@ struct repair_options {
     // repair to a data center other than the named one returns an error.
     std::vector<sstring> data_centers;
 
+    int ranges_parallelism = -1;
+
     repair_options(std::unordered_map<sstring, sstring> options) {
         bool_opt(primary_range, options, PRIMARY_RANGE_KEY);
         ranges_opt(ranges, options, RANGES_KEY);
@@ -1145,6 +1150,7 @@ struct repair_options {
         // Consume, ignore.
         int job_threads;
         int_opt(job_threads, options, JOB_THREADS_KEY);
+        int_opt(ranges_parallelism, options, RANGES_PARALLELISM_KEY);
 
         // The parsing code above removed from the map options we have parsed.
         // If anything is left there in the end, it's an unsupported option.
@@ -1165,6 +1171,7 @@ struct repair_options {
     static constexpr const char* TRACE_KEY = "trace";
     static constexpr const char* START_TOKEN = "startToken";
     static constexpr const char* END_TOKEN = "endToken";
+    static constexpr const char* RANGES_PARALLELISM_KEY = "ranges_parallelism";
 
     // Settings of "parallelism" option. Numbers must match Cassandra's
     // RepairParallelism enum, which is used by the caller.
@@ -1261,7 +1268,11 @@ static future<> do_repair_ranges(lw_shared_ptr<repair_info> ri) {
     if (ri->row_level_repair()) {
         // repair all the ranges in limited parallelism
         return parallel_for_each(ri->ranges, [ri] (auto&& range) {
-            return with_semaphore(repair_tracker().range_parallelism_semaphore(), 1, [ri, &range] {
+            // If user specified the ranges to repair in parallel respect it.
+            semaphore& sem = ri->ranges_parallelism_semaphore
+                    ? *ri->ranges_parallelism_semaphore
+                    : repair_tracker().range_parallelism_semaphore();
+            return with_semaphore(sem, 1, [ri, &range] {
                 check_in_shutdown();
                 ri->check_in_abort();
                 ri->ranges_index++;
@@ -1414,10 +1425,10 @@ static int do_repair_start(seastar::sharded<database>& db, sstring keyspace,
         repair_results.reserve(smp::count);
         for (auto shard : boost::irange(unsigned(0), smp::count)) {
             auto f = db.invoke_on(shard, [keyspace, cfs, id, ranges,
-                    data_centers = options.data_centers, hosts = options.hosts] (database& localdb) mutable {
+                    data_centers = options.data_centers, hosts = options.hosts, ranges_parallelism = options.ranges_parallelism] (database& localdb) mutable {
                 auto ri = make_lw_shared<repair_info>(service::get_local_storage_service().db(),
                         std::move(keyspace), std::move(ranges), std::move(cfs),
-                        id, std::move(data_centers), std::move(hosts));
+                        id, std::move(data_centers), std::move(hosts), ranges_parallelism);
                 return repair_ranges(ri);
             });
             repair_results.push_back(std::move(f));
--- a/repair/repair.hh
+++ b/repair/repair.hh
@@ -159,6 +159,8 @@ public:
     shard_id shard;
     std::vector<sstring> data_centers;
     std::vector<sstring> hosts;
+    int ranges_parallelism;
+    std::optional<semaphore> ranges_parallelism_semaphore;
     size_t nr_failed_ranges = 0;
     bool aborted = false;
     // Map of peer -> <cf, ranges>
@@ -188,7 +190,8 @@ public:
             const std::vector<sstring>& cfs_,
             int id_,
             const std::vector<sstring>& data_centers_,
-            const std::vector<sstring>& hosts_);
+            const std::vector<sstring>& hosts_,
+            int ranges_parallelism);
     future<> do_streaming();
     void check_failed_ranges();
     future<> request_transfer_ranges(const sstring& cf,


slivne:
@avikivity this is one of the changes we agreed for row level repair to make it work with high latency links

slivne:
@avikivity this is one of the changes we agreed for row level repair to make it work with high latency links

asias:
ping.

asias:
Rebased to latest master.

asias:
@avikivity Ping. This is needed by 3.2.

slivne:
@avikivity Ping. This is needed by 3.2.

glommer:
Last ping on this was 21 days ago.
There is no maintainer action: no request for changes, no merge.

Patches need to either move or be closed.

I am tagging all 4 people with  maintainer status according to the maintainers file: @avikivity @nyh @tgrabiec , @penberg 

This is an unrelated discussion, but I think we need more maintainers that can commit code.

We can't have a pull request with no action sitting here for 3 weeks.

api/api-doc/storage_service.json
```diff
@@ -917,6 +917,14 @@
                      "allowMultiple":false,
                      "type":"string",
                      "paramType":"query"
+                  },
+                  {
+                     "name":"ranges_parallelism",
+                     "description":"An integer specifying the number of ranges to repair in parallel.",
```
nyh: This number is so opaque... What integer should a user use here? 3? 100? 1000000? What is the default if I don't specify this? As a user, I wouldn't know how to use this option (although I guess it might come handy for our field engineers?).

nyh:
@glommer I left a lengthy review of this patch. I don't like this approach myself, but I also don't see anything specifically wrong with this code, so if you or @slivne believe it is important for users right now and **request** it in - despite not being the ideal solution - I can commit it. Please let me know.

glommer:
Thanks a lot Nadav.

I am not aware of any pressing need for a hack here.
This needs to be solved for repairs to work better in high latency links, but if the approach is not good @asias should engage in the search for a new one.

glommer:
FWIW I also favor a solution without this tunable.

asias:
> I don't see anything wrong with this patch's code, but I don't like the approach:
> 
> Exactly as demonstrated int he commit message, the meaning of the "repair parallelism" is very different on high-latency vs low-latency links. While on low-latency links a low parallelism can nevertheless take a big chunk of the machine's CPU, disk and network capacity, on a high-latency link low parallelism can result in underused network, disk and CPU. Moreover, what does the "range_paralellism" number actually mean? What is "a range"? Are there more layers of parallelism inside each range (including read-ahead, socket buffers, etc.) or not? Is this a master-side, slave-side or what parameter? Besides changing the parallelism (and noticably - the speed) of the repair, what does it do to the CPU or disk percentage used up by repair? As a user I wouldn't have a clue how to use this parameter. Although if @slivne or @glommer request this specific implementation as an emergency tweak for problems in the field, I'll oblige and commit this patch as-is.
> 
> If I think about how to solve this problem ideally, not in the way done here, I think that if the user is to configure anything - it should be the fraction of CPU, disk bandwidth and/or network bandwidth that they want to devote to repair (on _all_ machines involved!). That is our traditional "boring database"/"zero configuration" approach. We already support CPU and disk schedulers (and usually assumed the network split is derived from the CPU/disk available), but need to make it repair actually uses them. In other words, what I envision was repair adaptively controlling its own concurrency, increasing it while the CPU/disk quotas of the involved machines aren't fully used (I think it is quite fine to overshoot a bit, and use a slightly too high concurrency). This probably won't be easy to do, so maybe we need this patch in the interim?
> 
> Moreover, orthogonally to this specific patch there is another big problem in our repair approach on high-latency high-throughput links. Consider that we have a very high latency. You'll need very high parallelism to compensate, but this - with the current code which keeps in memory every row read for this entire latency - will need huge amounts of memory. A really good solution for this problem will use **less memory** in this case. Perhaps not caching all the rows in memory and hoping that only a few end up needing to be reread. This is why in the past I asked you to benchmark also the case of high latency with almost no inconsistencies in the data - this sort of repair should ideally be able to proceed at disk and network bandwidth, unrelated to the amount of memory or network latency.

Nadav, thanks for the lengthy review. 

First of all, I do not like user to configure anything neither. We already have the ranges parallelism auto-configured according to the memory size. See 131acc09cc1d568f994059736c8500adf5d05d88 (repair: Adjust parallelism according to memory size (#4696)). If user does not specify the "advanced" option, everything is zero configuration. The parameter introduced is for the case where the auto-configured parameter does not work. For example, why limit repair to 10% of memory if all I need to do is repair the cluster at the moment. The point is the auto-configured parameter is not optimal for all the use cases. 

What does ranges_parallelism option actually control? It controls the number of repair instances per shard in parallel. Each repair instance runs a full row level repair algorithm. It is a repair master side parameter. The follower side will follow whatever the repair master chooses.  

Regarding not caching all rows in memory and reread, what if the rows are from peers, rereading means retransmitting from the peers. I do not think this something we want to pursuit.

api/api-doc/storage_service.json
asias: The default is used to be 16. Now it is changed to be configured according to the number of memory allowed for repair, see 131acc09cc1d568f994059736c8500adf5d05d88.

asias:
> Thanks a lot Nadav.
> 
> I am not aware of any pressing need for a hack here.

Because nobody is running the repair in 3.2 release which is not released yet.

> This needs to be solved for repairs to work better in high latency links, but if the approach is not good @asias should engage in the search for a new one.

Increasing parallelism is the common way to compensate the latency. The 25X improvement mentioned in the commit log is good example it works well. I am open for new ideas but let's proceed and gain more actual experience before arguing if this this works or not.

glommer:
Hi @asias 

I was mostly referring to the tunable.
I think we can do away with it.

On/off tunables are okay for features we are afraid may have issues - like repair based streaming, but these kinds of tunables that you are introducing tend to become hard to remove later.

asias:
> Hi @asias
> 
> I was mostly referring to the tunable.
> I think we can do away with it.

You are welcome to give suggestions how to get rid of. No matter how smart we are, the auto-configured number could go wrong in some cases. 

> 
> On/off tunables are okay for features we are afraid may have issues - like repair based streaming, but these kinds of tunables that you are introducing tend to become hard to remove later.

Yes, it is hard to remove but why not just stopping using it. The option is present does not mean we have to use it.

glommer:
Users start depending on it, Asias.

My suggestion on how to remove it is by not adding it.

asias:
> Users start depending on it, Asias.

The option is not wired to nodetool repair, so normal user won't use it. We can easily make Scylla manager stop using it if we want.  In case we really want to remove the option we can return an error in the restful API when the option is provided. 

> My suggestion on how to remove it is by not adding it.

In my view, doing nothing is not better than providing more sophisticated control. Also, offering such control does not prevent us looking for better automatic method.

slivne:
@nyh - asias provided answers to your questions, are we missing anything - can this go in ?

nyh:
@slivne as I said, it *can* go in, there is nothing really "wrong" with it. Whether it should go in, however, is a different question. This solution goes against our usual "zero configuration" philosophy. So I think **you** should say if this should go in - considering what you know about users or field-engineers or whatever who want such a feature. I don't think there's any risk with this patch - it just that it feels like taking a step in the wrong direction.

slivne:
@nyh thanks - this was agreed by avi in the review process @avikivity - any input from you - about this added configuration - its not great, but we agreed that we will add it.

penberg:
As discussed here, it's highly unusual that we add such tunables, because they violate the "no configuration" rule and are hard to get rid of later. However, if @avikivity approved the approach, he can certainly merge the pull request.

avikivity:
I think we need to make this automatic. Allocate 10% of memory for repair and run as many ranges as will fit there.

Of course, running many ranges requires more compaction work later (or some other work to unify many small sstables).

Too bad we didn't go for rolling repair which could do this in a single stream.

asias:
> I think we need to make this automatic. Allocate 10% of memory for repair and run as many ranges as will fit there.

As I mentioned in the commit log, we already have the 10% memory thing to choose the ranges in parallel automatically.  This patch is for the case where such automation does not work. 

> Of course, running many ranges requires more compaction work later (or some other work to unify many small sstables).

Running ranges in parallel or not does not change the number of sstables repair will generate.

> 
> Too bad we didn't go for rolling repair which could do this in a single stream.

I am not sure what you meant here. Do you mean a single repair instance repair all the ranges?

avikivity:
> > I think we need to make this automatic. Allocate 10% of memory for repair and run as many ranges as will fit there.
> 
> As I mentioned in the commit log, we already have the 10% memory thing to choose the ranges in parallel automatically. This patch is for the case where such automation does not work.

How can you use it? If repair is slow, increase the parallel ranges until either it is fast or the node crashes due to OOM?

> 
> > Of course, running many ranges requires more compaction work later (or some other work to unify many small sstables).
> 
> Running ranges in parallel or not does not change the number of sstables repair will generate.

How is that possible? If the ranges are parallel, aren't they also written in parallel?

> 
> > Too bad we didn't go for rolling repair which could do this in a single stream.
> 
> I am not sure what you meant here. Do you mean a single repair instance repair all the ranges?

I meant that instead of reading a buffer and synchonizing it, have a queue where each time a row is added to the front and rows are popped from the back as they are transferred. What we talked about when we discussed the rsync algorithm.

asias:
> > > I think we need to make this automatic. Allocate 10% of memory for repair and run as many ranges as will fit there.
> > 
> > 
> > As I mentioned in the commit log, we already have the 10% memory thing to choose the ranges in parallel automatically. This patch is for the case where such automation does not work.
> 
> How can you use it? If repair is slow, increase the parallel ranges until either it is fast or the node crashes due to OOM?

If we know the table has little data, e.g., the system tables, we can increase the parallelism, like the one I mentioned in the commit log where 25x improvement is observed. If the auto configured parallelism which uses allows 10% of memory gives 10, we know if we increase it 2X for instances it will consume 20% of memory at most which will not not lead to OOM. 

Above is increasing case, we can have cases where we actually want to decrease. For example if the auto configured 10 parallelism has more cql impact than the user can bear, we can decrease it.  Without such tunable, the end result is probably the user can not run repair. IMHO, making things work is more important than the zero config philosophy. 

> 
> > > Of course, running many ranges requires more compaction work later (or some other work to unify many small sstables).
> > 
> > 
> > Running ranges in parallel or not does not change the number of sstables repair will generate.
> 
> How is that possible? If the ranges are parallel, aren't they also written in parallel?

I meant the total number of sstables are the same.

> 
> > > Too bad we didn't go for rolling repair which could do this in a single stream.
> > 
> > 
> > I am not sure what you meant here. Do you mean a single repair instance repair all the ranges?
> 
> I meant that instead of reading a buffer and synchonizing it, have a queue where each time a row is added to the front and rows are popped from the back as they are transferred. What we talked about when we discussed the rsync algorithm.

I spent a lot of time explaining why the rsync algorithm does not fit row level repair ;-)

penberg:
@asias is this pull request still relevant or should we just close it?

asias:
I will close it for now.

vladzcloudius:
> Thanks a lot Nadav.
> 
> I am not aware of any pressing need for a hack here. This needs to be solved for repairs to work better in high latency links, but if the approach is not good @asias should engage in the search for a new one.

One "pressing need" is our control of a repair parallelism (which is effectively defined by a number of ranges being repaired at the same time) from tools like `nodetool`.

I suggest reviving this PR and linking the corresponding API to one of the `nodetool` 's options, e.g. `-j`.

benipeled:
> CI state **`FAILURE`** - https://jenkins.scylladb.com/job/scylla-master/job/scylla-ci/2699/

The PR missing the `merge` branch - we should try to rebase and fix conflicts before retriggering the CI
```
benipeled@fedora ~/repos/scylla-4848 $ git ls-remote git@github.com:scylladb/scylla.git 'refs/pull/4848/*'
bc605cf4227422783f52bc0d10c5dd9d78fa522c        refs/pull/4848/head
```
For comparison, common PR contains two branches - head * merge - and the CI uses the `merge` branch

```
benipeled@fedora ~/repos/scylla-4848 $ git ls-remote git@github.com:scylladb/scylla.git 'refs/pull/14835/*'
dc3475901608ca12f61e871e98048ed32dcee85f        refs/pull/14835/head
5b31c3396cf2c7e6cee9e26e78134609d91bc6da        refs/pull/14835/merge
```

asias:
Replaced by https://github.com/scylladb/scylladb/pull/14886

PR #4863, opened 2019-08-18
relocatable: switch from run-time relocation to install-time relocation

### LINKED ISSUE #4673: Thread debugging is not available for relocatable scylla package
Scylla version: 3.1

Without thread debugging most of the `scylla-gdb.py` commands won't work. Thread-locals can't be read.

I guess this is due to `/usr/lib/debug/opt/scylladb/libreloc/libthread_db.so.1`, which matches the `/opt/scylladb/bin/../libreloc/libpthread.so.0`, being missing.

GDB log with `set debug libthread-db 1`:
```

Trying host libthread_db library: libthread_db.so.1.
Host libthread_db.so.1 resolved to: /lib64/libthread_db.so.1.
td_ta_new failed: versions of libpthread and libthread_db do not match
Trying host libthread_db library: /usr/lib/debug/opt/scylladb/libreloc/libthread_db.so.1.
dlopen failed: /usr/lib/debug/opt/scylladb/libreloc/libthread_db.so.1: cannot open shared object file: No such file or directory.
Trying host libthread_db library: /opt/scylladb/bin/../libreloc/libthread_db.so.1.
dlopen failed: /opt/scylladb/bin/../libreloc/libthread_db.so.1: cannot open shared object file: No such file or directory.
thread_db_load_search returning 0
warning: Unable to find libthread_db matching inferior's thread library, thread debugging will not be available.
```

avikivity:
Our current relocation works by invoking the dynamic linker with the
executable as an argument. This confuses gdb since the kernel records
the dynamic linker as the executable, not the real executable.

Switch to install-time relocation with patchelf: when installing the
executable and libraries, all paths are known, and we can update the
path to the dynamic loader and to the dynamic libraries.

Since patchelf itself is dynamically linked, we have to relocate it
dynamically (with the old method of invoking it via the dynamic linker).
This is okay since it's a one-time operation and since we don't expect
to debug core dumps of patchelf crashes.

We lose the ability to run scylla directly from the uninstalled
tarball, but since the nonroot installer is already moving in the
direction of requiring install.sh, that is not a great loss, and
certainly the ability to debug is more important.

dh_strip barfs on some binaries which were treated with patchelf,
so exclude them from dh_strip. This doesn't lose any functionality,
since these binaries didn't have debug information to begin with
(they are already-stripped Fedora executables).

Fixes #4673.

### DIFF
--- a/dist/debian/rules.mustache
+++ b/dist/debian/rules.mustache
@@ -34,7 +34,9 @@ override_dh_installinit:
 	dh_installinit --no-start --name node-exporter
 
 override_dh_strip:
-	dh_strip -Xlibprotobuf.so.15 -Xld.so --dbg-package={{product}}-server-dbg
+	# The binaries (ethtool...patchelf) don't pass dh_strip after going through patchelf. Since they are
+	# already stripped, nothing is lost if we exclude them, so that's what we do.
+	dh_strip -Xlibprotobuf.so.15 -Xld.so -Xethtool -Xgawk -Xgzip -Xhwloc-calc -Xhwloc-distrib -Xifconfig -Xlscpu -Xnetstat -Xpatchelf --dbg-package={{product}}-server-dbg
 
 override_dh_makeshlibs:
 
--- a/install.sh
+++ b/install.sh
@@ -89,6 +89,29 @@ if [ -n "$pkg" ] && [ "$pkg" != "server" -a "$pkg" != "conf" -a "$pkg" != "kerne
     exit 1
 fi
 
+patchelf() {
+    # patchelf comes from the build system, so it needs the build system's ld.so and
+    # shared libraries. We can't use patchelf on patchelf itself, so invoke it via
+    # ld.so.
+    LD_LIBRARY_PATH="$PWD/libreloc" libreloc/ld.so libexec/patchelf "$@"
+}
+
+adjust_bin() {
+    local bin="$1"
+    # We could add --set-rpath too, but then debugedit (called by rpmbuild) barfs
+    # on the result. So use LD_LIBRARY_PATH in the thunk, below.
+    patchelf \
+	--set-interpreter "$prefix/libreloc/ld.so" \
+	"$root/$prefix/libexec/$bin"
+    cat > "$root/$prefix/bin/$bin" <<EOF
+#!/bin/bash -e
+export GNUTLS_SYSTEM_PRIORITY_FILE="\${GNUTLS_SYSTEM_PRIORITY_FILE-$prefix/libreloc/gnutls.config}"
+export LD_LIBRARY_PATH="$prefix/libreloc"
+exec -a "\$0" "$prefix/libexec/$bin" "\$@"
+EOF
+    chmod +x "$root/$prefix/bin/$bin"
+}
+
 rprefix="$root/$prefix"
 retc="$root/etc"
 rusr="$root/usr"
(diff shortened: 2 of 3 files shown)


avikivity:
Updated: some hacks to make .deb generation works (documented in patch).

install.sh
```diff
@@ -89,6 +89,29 @@ if [ -n "$pkg" ] && [ "$pkg" != "server" -a "$pkg" != "conf" -a "$pkg" != "kerne
...
+    # We could add --set-rpath too, but then debugedit (called by rpmbuild) barfs
+    # on the result. So use LD_LIBRARY_PATH in the thunk, below.
+    patchelf \
+	--set-interpreter "$prefix/libreloc/ld.so" \
+	"$root/$prefix/libexec/$bin"
+    cat > "$root/$prefix/bin/$bin" <<EOF
+#!/bin/bash -e
+export GNUTLS_SYSTEM_PRIORITY_FILE="\${GNUTLS_SYSTEM_PRIORITY_FILE-$prefix/libreloc/gnutls.config}"
+export LD_LIBRARY_PATH="$prefix/libreloc"
+exec -a "\$0" "$prefix/libexec/$bin" "\$@"
```
espindola: Since we are creating a script anyway, it seems better to use a CWD relative path for the interpreter. That way the script can be similar to

cd $prefix/libexec
exec -a "\$0" ./$bin "\$@"

And we don't need to patch the binary.

dist/debian/rules.mustache
```diff
@@ -34,7 +34,9 @@ override_dh_installinit:
 	dh_installinit --no-start --name node-exporter
 
 override_dh_strip:
-	dh_strip -Xlibprotobuf.so.15 -Xld.so --dbg-package={{product}}-server-dbg
+	# The binaries (ethtool...patchelf) don't pass dh_strip after going through patchelf. Since they are
+	# already stripped, nothing is lost if we exclude them, so that's what we do.
```
espindola: Do we know what is special about these?

install.sh
```diff
@@ -89,6 +89,29 @@ if [ -n "$pkg" ] && [ "$pkg" != "server" -a "$pkg" != "conf" -a "$pkg" != "kerne
...
+    # patchelf comes from the build system, so it needs the build system's ld.so and
+    # shared libraries. We can't use patchelf on patchelf itself, so invoke it via
+    # ld.so.
+    LD_LIBRARY_PATH="$PWD/libreloc" libreloc/ld.so libexec/patchelf "$@"
+}
+
+adjust_bin() {
+    local bin="$1"
+    # We could add --set-rpath too, but then debugedit (called by rpmbuild) barfs
+    # on the result. So use LD_LIBRARY_PATH in the thunk, below.
```
espindola: Do we know why? What is the error?

install.sh
avikivity: We can't change CWD from under the executable's feet.

dist/debian/rules.mustache
avikivity: No.

install.sh
avikivity: The error manifests in debugedit complaining that DWARF version 0 is unsupported, followed by another failure. Random googling shows a bug in patchelf where the linker used the rpath string as a string pool and pointed other strings into it (not 100% sure), and since patchelf nuked it those other strings got corrupted. Maybe it's related and maybe not.

bhalevy:
@fruch, we probably need to adjust `scylla-ccm/ccmlib/scylla_node.py` the same way to get debuggable coredumps.

https://github.com/scylladb/scylla-ccm/blob/468f559e248fb0144496b6807baacf4404b4a10d/ccmlib/scylla_node.py#L197-L201

avikivity:
@bhalevy ccm should use install.sh instead of duplicating it

avikivity:
@espindola is you are satisfied, please mark it on github, or @tgrabiec will never merge it

install.sh
espindola: Fair point. LGTM

slivne:
@tgrabiec if all is ok please merge

PR #4865, opened 2019-08-19
Add python tests using tox

ultrabug:
These commits allows us to test our python code base using tox and (finally) leverage our tox.ini configuration to integrate this automated testing in the test suite.

- adds `python3-tox` installation dependency
- modifies `tox.ini` to test for the black python code style + flake8
- adds a non-default `python` test mode to the `tests.py` for automation

These commits are meant to be as readable as possible and not to break the current tests while allowing us to acknowledge all the code style and modifications to be done on the python code base.

If this is accepted, this will be used as the source of other proposals to fix all the python code base.

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -61,6 +61,7 @@ fedora_packages=(
     python3-magic
     python3-psutil
     python3-cassandra-driver
+    python3-tox
     dnf-utils
     pigz
     net-tools
--- a/test.py
+++ b/test.py
@@ -186,7 +186,7 @@ def alarm_handler(signum, frame):
 
 
 if __name__ == "__main__":
-    all_modes = ['debug', 'release', 'dev', 'sanitize']
+    all_modes = ['debug', 'release', 'dev', 'sanitize', 'python']
 
     sysmem = os.sysconf('SC_PAGE_SIZE') * os.sysconf('SC_PHYS_PAGES')
     testmem = 2e9
@@ -228,6 +228,9 @@ if __name__ == "__main__":
     test_to_run = []
     modes_to_run =  ['debug', 'release', 'dev'] if not args.modes else args.modes
     for mode in modes_to_run:
+        # python has its own 'run_python_test' function, skip it
+        if mode == "python":
+            continue
         prefix = os.path.join('build', mode, 'tests')
         standard_args = '--overprovisioned --unsafe-bypass-fsync 1 --blocked-reactor-notify-ms 2000000'.split()
         seastar_args = '-c2 -m2G'.split()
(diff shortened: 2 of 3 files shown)


avikivity:
@ultrabug because all of our tests run in dbuild, I'll have to update the docker container to include tox, and then ask you to rebase on top.

avikivity:
err, I see that installing python3-tox also pulls in python3{4,5,6,8}. That's a bit much. Is it truly necessary? That's a lot of baggage for a style check.

ultrabug:
@avikivity if our long term intention is to only run style checks, I agree that it sounds a bit too much but this is not my only motivation; it's just a start ;)

That being said I'm surprised that your current dbuild system does not provide any python3 version?

ultrabug:
> That being said I'm surprised that your current dbuild system does not provide any python3 version?

Sorry just saw that python3.7 was indeed not in your list. I have no clue if Fedora allows you to specify dependencies for specific interpreters?

Since for now, of course python3.7 would be enough. Is there a trick you know on this distro?

avikivity:
It looks like those are weak dependencies, and the manual suggests adding `--setopt=install_weak_deps=False`

avikivity:
What's the rest?

nyh:
I'd like to hear a bit about the motivation of this change - is your motivation to check that Scylla's Python scripts work on multiple versions of Python? Or something else?

ultrabug:
> It looks like those are weak dependencies, and the manual suggests adding --setopt=install_weak_deps=False

Should I modify it somewhere then?

> What's the rest?

The rest is that since we'll have a nice automated test framework in place we will also be able to add python code tests to the code base and run them as well seamlessly

So we start with enforcing a code style on the python code base but this will allow us to run tests on the python code as well shall we want to.

>  is your motivation to check that Scylla's Python scripts work on multiple versions of Python? Or something else?

No. First motivation is to make sure the code base respects a code style and enforce some basic QA checks (such as unused imports, variable shadowing, etc). So I'm happy with having only the default python 3.7 version for this.

ultrabug:
Hey guys, is there anything else I can do here? 

Cheers

ultrabug:
@nyh @avikivity friendly ping please; I've a lot more to come after this one.

kostja:
I am doing some heavy lifting with test.py to add LWT tests. I'm afraid I will be unable to integrate my changes with this commit.

ultrabug:
@kostja do you mean that I'm implementing this in a way that will prevent your additions to work?

Or do you simply mean that I need to wait until you're done and rebase my PR on you?

kostja:
I'm not sure you will be able to rebase. Could you show what else you have in the pipeline so we align the plans?

I am currently planning the following:
- automatic mode selection based on ./configure.py. This is the latest patch: https://groups.google.com/forum/#!topic/scylladb-dev/VvzriCsh-tg
- a configuration file searched in project root and then in ~ 
- more concise screen output
- automatic search for tests
- support multiple directories for tests
- move test options to test suite (test directory) configuration file
- make sure ctrl-c kills outstanding tests
- add a log file for the harness
- add support for tests against scylla binary (start/stop of scylla binary for tests).

ultrabug:
@kostja FYI I pinged you on Slack in case you're up to discuss it live.

What I have in the pipeline is that those python tests if you run them outputs a lot of things to be fixed. So the rest of the pipeline contributions is basically to have to python tests pass :)

ultrabug:
So after discussing it with @kostja I'll wait for the current refactoring of test.py to be done and then update this PR to switch to the new way of modular testing!

Meanwhile, I'll be happy to help @kostja in his Python endeavor :+1:

fruch:
this commit doesn't include the output of the black changes yet, it would be in a separate PR ?

ultrabug:
> this commit doesn't include the output of the black changes yet, it would be in a separate PR ?

Planned as separate PR since the changes will be huge

nyh:
@ultrabug unfortunately I can't review this patch because I don't understand its motivation.

If I understand correctly, you are adding new a new "test" which tests the style of Python code. Beyond the question of why do we need to enforce strict style rules, and why this strict enforcement is so important that we now need not just one but two separate tools to check it (pep8 and black), I also don't understand why this integrates in test.py as a new "mode" and not as a test that can run in any mode (since, after all, the Python style can be checked regardless of which mode you used to compile Scylla).

But most importantly, it seems that both you and @kostja are busy rewriting test.py which - I think you should remember - is a small 300-line Python program half of which is nothing more than a list of tests - but I didn't see anywhere any design document or mail or anything on what it is exactly that you're trying to achieve by these modifications.

While doing that, Scylla already has a bunch of test mechanisms not run by test.py, including alternator-test, redis-test, LWT's tests (I don't know where these are, I never saw them), and dtest. Is it your goal to integrate those inside test.py? Or do you have a different goal in mind? I didn't understand.

Another thing I completely failed to understand is how tox is related to this. Is this about testing our Python code, or testing **with** Python code?

I see that @avikivity already committed your previous contribution of **tox.ini** (commit 2022da2405) so hopefully he understands your direction and can properly review this patch. If not maybe you can give a longer explanation of what it is that you're trying to do.

PR #4873, opened 2019-08-20
cql3: Optimize LIKE regex construction

dekimir:
Currently we create a regex from the LIKE pattern for every row
considered during filtering, even though the pattern is always the
same.  This is wasteful, especially since we require costly
optimization in the regex compiler.  Fix it by reusing the regex
whenever the pattern is unchanged since the last call.

Tests: unit (dev)

### DIFF
--- a/cql3/restrictions/single_column_restriction.hh
+++ b/cql3/restrictions/single_column_restriction.hh
@@ -41,6 +41,8 @@
 
 #pragma once
 
+#include <optional>
+
 #include "cql3/restrictions/restriction.hh"
 #include "cql3/restrictions/term_slice.hh"
 #include "cql3/term.hh"
@@ -51,6 +53,7 @@
 #include "exceptions/exceptions.hh"
 #include "keys.hh"
 #include "mutation_partition.hh"
+#include "utils/like_matcher.hh"
 
 namespace cql3 {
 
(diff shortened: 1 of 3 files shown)


cql3/restrictions/single_column_restriction.hh
```diff
@@ -384,6 +387,11 @@ public:
 class single_column_restriction::LIKE final : public single_column_restriction {
 private:
     ::shared_ptr<term> _value;
+    /// Matches cell value against LIKE pattern.  Pointer because it cannot be initialized in the
+    /// constructor when the pattern is a bind marker.  Mutable because it is initialized on demand
+    /// in is_satisfied_by().
+    mutable std::unique_ptr<like_matcher> _matcher;
```
denesb: `std::optional` is another option here, we usually use `std::unique_ptr` when stable address is needed.

cql3/restrictions/statement_restrictions.cc
```diff
@@ -916,6 +915,18 @@ bool token_restriction::slice::is_satisfied_by(const schema& schema,
...
 }
 
+bool single_column_restriction::LIKE::init_matcher(const query_options& options) const {
+    auto pattern = to_bytes_opt(_value->bind_and_get(options));
+    if (!pattern) {
+        return false;
+    }
+    if (!_matcher || pattern != _last_pattern) {
+        _matcher.reset(new like_matcher(*pattern));
+        _last_pattern = std::move(pattern);
```
avikivity: std::make_unique<> is better practice than a naked new. std::optional<> would be even better.

tests/cql_query_test.cc
```diff
@@ -3722,6 +3722,21 @@ void require_rows(cql_test_env& e,
...
+void require_rows(cql_test_env& e,
+                  cql3::prepared_cache_key_type id,
+                  const std::vector<cql3::raw_value>& values,
+                  const std::vector<std::vector<bytes_opt>>& expected,
+                  const source_location& loc = source_location::current()) {
+    try {
+        assert_that(e.execute_prepared(id, values).get0()).is_rows().with_rows_ignore_order(expected);
+    }
+    catch (const std::exception& e) {
+        BOOST_FAIL(format("execute_prepared failed: {}\n{}:{}: originally from here",
```
avikivity: Our coding style is  `} catch (` in one line.

dekimir:
Feedback addressed, PTAL.

PR #4874, opened 2019-08-21
build: build seastar in build directory

avikivity:
Currently, seastar is built in seastar/build/{mode}. This means we have two build
directories: build/{mode} and seastar/build/{mode}.

This patch changes that to have only a single build directory (build/{mode}). It
does that by calling Seastar's cmake directly instead of through Seastar's
./configure.py.  However, to support dpdk, if that is enabled it calls cmake
through Seastar's ./cooking.sh (similar to what Seastar's ./configure.py does).

All ./configure.py flags are translated to cmake variables, in the same way that
Seastar does.

Contains fix from Rafael to pass the flags for the correct mode.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1060,17 +1060,6 @@ scylla_release = file.read().strip()
 
 extra_cxxflags["release.cc"] = "-DSCYLLA_VERSION=\"\\\"" + scylla_version + "\\\"\" -DSCYLLA_RELEASE=\"\\\"" + scylla_release + "\\\"\""
 
-seastar_flags = []
-if args.dpdk:
-    # fake dependencies on dpdk, so that it is built before anything else
-    seastar_flags += ['--enable-dpdk']
-if args.gcc6_concepts:
-    seastar_flags += ['--enable-gcc6-concepts']
-if args.alloc_failure_injector:
-    seastar_flags += ['--enable-alloc-failure-injector']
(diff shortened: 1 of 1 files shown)


avikivity:
This is a replacement for #4830 which failed next promotion. Change is 

```

diff --git a/configure.py b/configure.py
index c867e5ff6..4a5f59d43 100755
--- a/configure.py
+++ b/configure.py
@@ -1087,7 +1087,7 @@ def configure_seastar(build_dir, mode):
         '-DCMAKE_BUILD_TYPE={}'.format(MODE_TO_CMAKE_BUILD_TYPE[mode]),
         '-DCMAKE_C_COMPILER={}'.format(args.cc),
         '-DCMAKE_CXX_COMPILER={}'.format(args.cxx),
-        '-DSeastar_CXX_FLAGS={}'.format((seastar_cflags + ' ' + modes['release']['cxx_ld_flags']).replace(' ', ';')),
+        '-DSeastar_CXX_FLAGS={}'.format((seastar_cflags + ' ' + modes[mode]['cxx_ld_flags']).replace(' ', ';')),
         '-DSeastar_LD_FLAGS={}'.format(seastar_ldflags),
         '-DSeastar_CXX_DIALECT=gnu++17',
         '-DSeastar_STD_OPTIONAL_VARIANT_STRINGVIEW=ON',
```

PR #4889, opened 2019-08-23
use labels to represent protocol-level statistics

glommer:
We currently support the thrift and CQL protocols, and I see activity in
the mailing list being done in order to add Redis. Not that we plan to
add any other, but 3 already justifies some level of commonality. The
metrics for the thrift and CQL protocols predate our usage of prometheus
labels, so they use names like "thrift_served" and
"transport_requests_served", and same for connections and others.

This makes it harder to answer questions like "what is the total
throughput of my cluster across all protocols?", since we have to sum up
all possible protocols, and that's exactly what labels are for.

Not all concepts translates well from one protocol to another, but most
do, and to the extent that they do this change will make it easier for
our monitoring stack to display good metrics as new protocols are added.

### DIFF
--- a/thrift/server.cc
+++ b/thrift/server.cc
@@ -302,18 +302,21 @@ thrift_server::requests_served() const {
     return _requests_served;
 }
 
+static const metrics::label class_label("protocol");
+
 thrift_stats::thrift_stats(thrift_server& server) {
     namespace sm = seastar::metrics;
+    auto proto_label_instance = class_label("thrift");
 
-    _metrics.add_group("thrift", {
-        sm::make_derive("thrift-connections", [&server] { return server.total_connections(); },
-                        sm::description("Rate of creation of new Thrift connections.")),
+    _metrics.add_group("transport", {
+        sm::make_derive("connections", [&server] { return server.total_connections(); },
+                        sm::description("Rate of creation of new Thrift connections."), {proto_label_instance}),
 
         sm::make_gauge("current_connections", [&server] { return server.current_connections(); },
-                        sm::description("Holds a current number of opened Thrift connections.")),
+                        sm::description("Holds a current number of opened Thrift connections."), {proto_label_instance}),
 
-        sm::make_derive("served", [&server] { return server.requests_served(); },
-                        sm::description("Rate of serving Thrift requests.")),
+        sm::make_derive("requests_served", [&server] { return server.requests_served(); },
+                        sm::description("Rate of serving Thrift requests."), {proto_label_instance}),
     });
 }
 
--- a/transport/server.cc
+++ b/transport/server.cc
@@ -144,6 +144,8 @@ event::event_type parse_event_type(const sstring& value)
     }
 }
 
+static const metrics::label class_label("protocol");
+
 cql_server::cql_server(distributed<service::storage_proxy>& proxy, distributed<cql3::query_processor>& qp, auth::service& auth_service,
         const cql3::cql_config& cql_config, cql_server_config config)
     : _proxy(proxy)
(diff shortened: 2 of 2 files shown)


thrift/server.cc
```diff
@@ -302,18 +302,21 @@ thrift_server::requests_served() const {
...
+                        sm::description("Rate of creation of new Thrift connections."), {proto_label_instance}),
 
         sm::make_gauge("current_connections", [&server] { return server.current_connections(); },
-                        sm::description("Holds a current number of opened Thrift connections.")),
+                        sm::description("Holds a current number of opened Thrift connections."), {proto_label_instance}),
 
-        sm::make_derive("served", [&server] { return server.requests_served(); },
-                        sm::description("Rate of serving Thrift requests.")),
+        sm::make_derive("requests_served", [&server] { return server.requests_served(); },
+                        sm::description("Rate of serving Thrift requests."), {proto_label_instance}),
```
amnonh: The description should not be specific

transport/server.cc
```diff
@@ -157,31 +159,34 @@ cql_server::cql_server(distributed<service::storage_proxy>& proxy, distributed<c
...
-                                            "Non-zero value indicates that our bottleneck is memory and more specifically - the memory quota allocated for the \"CQL transport\" component.", _max_request_size))),
+                                            "Non-zero value indicates that our bottleneck is memory and more specifically - the memory quota allocated for the \"CQL transport\" component.", _max_request_size)), {proto_label_instance}),
         sm::make_derive("requests_blocked_memory", _requests_blocked_memory,
                         sm::description(
                             seastar::format("Holds an incrementing counter with the requests that ever blocked due to reaching the memory quota limit ({}B). "
-                                            "The first derivative of this value shows how often we block due to memory exhaustion in the \"CQL transport\" component.", _max_request_size))),
+                                            "The first derivative of this value shows how often we block due to memory exhaustion in the \"CQL transport\" component.", _max_request_size)), {proto_label_instance}),
```
amnonh: remvoe the cql from the description

transport/server.cc
```diff
@@ -157,31 +159,34 @@ cql_server::cql_server(distributed<service::storage_proxy>& proxy, distributed<c
...
        sm::make_gauge("requests_memory_available", [this] { return _memory_available.current(); },
                         sm::description(
                             seastar::format("Holds the amount of available memory for admitting new requests (max is {}B)."
-                                            "Zero value indicates that our bottleneck is memory and more specifically - the memory quota allocated for the \"CQL transport\" component.", _max_request_size))),
+                                            "Zero value indicates that our bottleneck is memory and more specifically - the memory quota allocated for the \"CQL transport\" component.", _max_request_size)), {proto_label_instance}),
```
amnonh: remove the CQL from the description

thrift/server.cc
```diff
@@ -302,18 +302,21 @@ thrift_server::requests_served() const {
...
-    _metrics.add_group("thrift", {
-        sm::make_derive("thrift-connections", [&server] { return server.total_connections(); },
-                        sm::description("Rate of creation of new Thrift connections.")),
+    _metrics.add_group("transport", {
+        sm::make_derive("connections", [&server] { return server.total_connections(); },
+                        sm::description("Rate of creation of new Thrift connections."), {proto_label_instance}),
 
         sm::make_gauge("current_connections", [&server] { return server.current_connections(); },
-                        sm::description("Holds a current number of opened Thrift connections.")),
+                        sm::description("Holds a current number of opened Thrift connections."), {proto_label_instance}),
```
amnonh: remove thrift from the description

amnonh:
On Mon, Aug 26, 2019 at 9:38 AM Piotr Jastrzębski <notifications@github.com>
wrote:

> *@haaawk* approved this pull request.
>
> Would probably by good @amnonh <https://github.com/amnonh> had a look on
> this.
>
Just did and had some comments

glommer:
Amnon, so the gist of your review is that we need the text strings to be the same across all protocols? That's asking for a mistake in the future. if that's the case Maybe we should just have those strings encoded somewhere else as a constexpr string.

What happens if some of them have no help text at all and some others do ?

thrift/server.cc
```diff
@@ -302,18 +302,21 @@ thrift_server::requests_served() const {
     return _requests_served;
 }
 
+static const metrics::label class_label("protocol");
```
avikivity: protocol_label?

thrift/server.cc
```diff
@@ -302,18 +302,21 @@ thrift_server::requests_served() const {
...
+
 thrift_stats::thrift_stats(thrift_server& server) {
     namespace sm = seastar::metrics;
+    auto proto_label_instance = class_label("thrift");
 
-    _metrics.add_group("thrift", {
-        sm::make_derive("thrift-connections", [&server] { return server.total_connections(); },
-                        sm::description("Rate of creation of new Thrift connections.")),
+    _metrics.add_group("transport", {
```
avikivity: I'm tempted to ask to rename this to "coordinator". "transport" doesn't say anythign, it's just the name of the directory where the native protocol lives (which isn't even our native protocol).

thrift/server.cc
avikivity: Yes, to make the common across all protocols, you can add a class protocol_statistics just to register and hold the metrics. It can accept the protocol label instance and lambdas for each metric in its constructor.

avikivity: (each protocol server would have its own instance of the protocol_statistics class).

transport/server.cc
```diff
@@ -157,31 +159,34 @@ cql_server::cql_server(distributed<service::storage_proxy>& proxy, distributed<c
 {
     namespace sm = seastar::metrics;
 
+    auto proto_label_instance = class_label("cql3");
+
```
avikivity: It's really sad that the protocol is named "cql", because that's really the name of the query language and causes confusion.

However, let's call it cql and not cql3 as this is the name used in Cassandra documentation. Or maybe even "cqlbinary" to distinguish it from the language (the docs reference it as "CQL BINARY PROTOCOL").

Note that the thrift protocol can process cql3 statements.

avikivity:
Good idea, had some comments on the implementation and even more on text strings.

thrift/server.cc
amnonh: Just to be clear, help is associated with a metric name, you got one help-line regardless of the labels.
This is a property of the Prometheus protocol, so you need to make the description (which will appear as HELP in the output) to match all metrics with that name.

avikivity: Yes, so you'll need to unify the semantics.

avikivity: If most of the metrics are common, that's okay.

Which metrics are not common to all protocols?

slivne:
@amnonh / @nyh I know we have some stats available for alternator - did we merge this in another form (so this pull requests simply should be closed) or are we waiting for a new version

amnonh:
@slivne the metrics in alternator, are alternator specific, I don't think they are related to this PR

glommer:
I am retiring this for now.

Despite the good direction, Alternator added a lot more metrics and we're not actively pursuing at the moment.

PR #4892, opened 2019-08-26
Minor cleanup patches for sstables

asias:
- sstables: Move leveled_compaction_strategy implementation to source file
- sstables: Include dht/i_partitioner.hh for dht::partition_range


asias:
Sorry, I could not request reviewers.

sstables/leveled_compaction_strategy.cc
```diff
@@ -0,0 +1,159 @@
+/*
+ * Copyright (C) 2019 ScyllaDB
```
bhalevy: nit: since this code was merely copied over it should carry the original year it was authored, not this year,.
Otherwise, LGTM

avikivity: It's okay to update the copyright year.

PR #4905, opened 2019-08-26
Optimize type find

espindola:
This avoids a double dispatch on _kind and also removes a few shared_ptr copies.

The extra work was a small regression from the recent types refactoring.

### DIFF
--- a/types.cc
+++ b/types.cc
@@ -713,12 +713,11 @@ bool abstract_type::is_string() const {
     return visit(*this, visitor{});
 }
 
-static bool find(const abstract_type& t, std::function<bool(const abstract_type&)> f) {
-    if (f(t)) {
-        return true;
-    }
+template<typename Predicate>
+GCC6_CONCEPT(requires CanHandleAllTypes<Predicate>)
+static bool find(const abstract_type& t, const Predicate& f) {
     struct visitor {
-        std::function<bool(const abstract_type&)>& f;
+        const Predicate& f;
         bool operator()(const abstract_type&) { return false; }
         bool operator()(const reversed_type_impl& r) { return find(*r.underlying_type(), f); }
         bool operator()(const tuple_type_impl& t) {
@@ -727,21 +726,30 @@ static bool find(const abstract_type& t, std::function<bool(const abstract_type&
         bool operator()(const map_type_impl& m) { return find(*m.get_keys_type(), f) || find(*m.get_values_type(), f); }
         bool operator()(const listlike_collection_type_impl& l) { return find(*l.get_elements_type(), f); }
     };
-    return visit(t, visitor{f});
+    return visit(t, [&](const auto& t) {
+        if (f(t)) {
+            return true;
+        }
+        return visitor{f}(t);
+    });
 }
 
 bool abstract_type::references_duration() const {
-    return find(*this, [] (const abstract_type& t) { return t.get_kind() == kind::duration; });
+    struct visitor {
+        bool operator()(const abstract_type&) const { return false; }
+        bool operator()(const duration_type_impl&) const { return true; }
+    };
+    return find(*this, visitor{});
 }
 
 bool abstract_type::references_user_type(const sstring& keyspace, const bytes& name) const {
     struct visitor {
         const sstring& keyspace;
         const bytes& name;
-        bool operator()(const abstract_type&) { return false; }
-        bool operator()(const user_type_impl& u) { return u._keyspace == keyspace && u._name == name; }
+        bool operator()(const abstract_type&) const { return false; }
+        bool operator()(const user_type_impl& u) const { return u._keyspace == keyspace && u._name == name; }
     };
-    return find(*this, [&](const abstract_type& t) { return visit(t, visitor{keyspace, name}); });
+    return find(*this, visitor{keyspace, name});
 }
 
 namespace {
--- a/types.hh
+++ b/types.hh
@@ -815,7 +815,7 @@ class reversed_type_impl : public abstract_type {
         , _underlying_type(t)
     {}
 public:
-    shared_ptr<const abstract_type> underlying_type() const {
+    const data_type& underlying_type() const {
         return _underlying_type;
     }
 
--- a/types/collection.hh
+++ b/types/collection.hh
@@ -104,7 +104,7 @@ protected:
     data_type _elements;
     explicit listlike_collection_type_impl(kind k, sstring name, data_type elements,bool is_multi_cell);
 public:
-    data_type get_elements_type() const { return _elements; }
+    const data_type& get_elements_type() const { return _elements; }
 };
 
 template <typename BytesViewIterator>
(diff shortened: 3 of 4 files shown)


types.cc
```diff
@@ -727,21 +725,30 @@ static bool find(const abstract_type& t, std::function<bool(const abstract_type&
...
         const sstring& keyspace;
         const bytes& name;
-        bool operator()(const abstract_type&) { return false; }
-        bool operator()(const user_type_impl& u) { return u._keyspace == keyspace && u._name == name; }
+        bool operator()(const abstract_type&) const { return false; }
+        bool operator()(const user_type_impl& u) const { return u._keyspace == keyspace && u._name == name; }
     };
-    return find(*this, [&](const abstract_type& t) { return visit(t, visitor{keyspace, name}); });
+    return find(*this, visitor{keyspace, name});
 }
```
avikivity: AFAIR, this is only used during DDL operations when considering whether we can remove a type or not. Is it worth optimizing?

espindola:
All tests compile and run in dev mode at least.  I think I tested in debug
mode too.

avikivity:
Sorry, I had a pending review comment but forgot to push the button to unleash it.

espindola:
So, is this OK?

Cheers,
Rafael

espindola:
Ping. I have rebased locally (no conflicts) and all tests pass in dev and debug mode.

espindola:
ping2. This is a pretty simple change and removes the double dispatch and std::function creation.

types.cc
```diff
@@ -713,12 +713,10 @@ bool abstract_type::is_string() const {
     return visit(*this, visitor{});
 }
 
-static bool find(const abstract_type& t, std::function<bool(const abstract_type&)> f) {
-    if (f(t)) {
-        return true;
-    }
+template<typename T>
+static bool find(const abstract_type& t, const T& f) {
```
avikivity: Please add a concept for T (and a better name) so people know what to expect.

avikivity: I don't understand why this is a good thing. It bloats the runtime and compile time, to provide a performance improvement where one isn't needed.

espindola:

> +template<typename T>
> +static bool find(const abstract_type& t, const T& f) {
>
> I don't understand why this is a good thing. It bloats the runtime and compile time, to provide a performance improvement where one isn't needed.

That is false.

The current code expansion includes shared_ptr and std::function
construction and destruction. The release and dev size of .text in
scylla in each commit is:

                dev         release
master:         22385412    40847348
first patch:    22384708    40847812
second patch:   22374356    40796500

So the only increase is by 464 bytes and on the first patch and only in
release mode. Given that dev mode decreases, this is most likely because
of the inliner being able to inline more now that there is less noise.

I also looked at the final assembly for references_duration, and it is
as good as if it had been written by hand.

I can go look at the assembly produced with and without the first patch
if you want, but it is a bit baffling why this simple patch series is so
controversial.

espindola:
I force pushed a version with a concept on the predicate of find.

avikivity:
I resist patches that add templates at wider scopes, and encourage reducing the scope of templates (usually by type erasing with std::function), whenever this isn't harming performance. This shouldn't surprise you, it's part of what made dev mode so much faster. If I saw the post-patch code I would be tempted to do this type erasure, and it is likely that as we add more calls to find() the bloat will get worse.

How much is the second patch alone? I don't want to make this a project, but sorry, it doesn't look like a win to me.

espindola:
Sorry,  but this is crazy inconsistent. When this code was first developed
you carefully looked for each case of double dispatch and asked me to fix
them. Now there is a battle to remove one.

The general complaint about templates over functions doesn't apply in this
case:

* With std::function we need a double dispatch since a std::function will
always take a abstract_type and have to switch on it.

* With a template what we pass is effectively an overload set. That is why
there is now a single switch.

So before the patch we had one find function with two switches and two
calls creating std::function.  Now there are two functions with a switch
each that are trivial to inline into their respective callers.

As the dev build size shows,  the amount of produced code was reduced!

espindola:
I pushed a new version where I reordered the patches.

In the second patch, which adds the template, I compared the sizes of find and the two callers before and after. The results are in the commit message for the second patch.

espindola:
Ping. Now that the patches are reordered and the measurements show that the templates are not an issue, what is missing?

kostja:
Avi, let's push this! It seems the arguments against are *very complex* (I am struggling trying to understand them) and not immediate. This is the main argument actually - that it is not worth arguing, so the contributor should be respected for his work.
The less major argument is that having studied the types API previously I like the new code as more consistent with the rest of the changes. Having both polymorphic and type-based dispatch will require an effort to underestand - when should I use which? Having a C background, I would actually use a simple switch statement for dispatch instead of @espindola approach, but in a C++ project his way of dispatching looks quite cool & fancy.

avikivity:
@kostja the arguments against are not complicated. A change to a non-hot code path should not cause code (and compile time) bloat, so I asked @espindola to measure it, because the code looked like it was going in the wrong direction. Of course I don't like rejecting work, but changes have to move the code in a positive direction and maintainers may request the contributor for additional data to justify the contribution.

avikivity:
@espindola we routinely type-erase in cold paths (like sharded::invoke_on_all) and avoid virtual functions in hot paths. There should be nothing surprising here.

The fact that it turned out that removing type-erasure actually helps things was a nice surprise, but not something that can be relied on.

espindola:

> @espindola we routinely type-erase in cold paths (like sharded::invoke_on_all) and avoid virtual functions in hot paths. There should be nothing surprising here.
>
> The fact that it turned out that removing type-erasure actually helps things was a nice surprise, but not something that can be relied on.

"type-erased good, type expanded bad", without looking at what the patch
is doing makes as much sense as "Four legs good, two legs bad."

Cheers,
Rafael

avikivity:
Of course I looked at the patch. And the motto does make sense, we used it extensively, from type-erasing future::then() to ugly type-erasing using std::any() in the api/ directory.

It happened to work for you, but in the general case, adding templates at the higher scope increases bloat.

nyh:
On Sun, Sep 8, 2019 at 6:03 AM Rafael Ávila de Espíndola <
notifications@github.com> wrote:

> "type-erased good, type expanded bad", without looking at what the patch
> is doing makes as much sense as "Four legs good, two legs bad."
>

All patches are equal, but some patches are more equal than others :-)

espindola:
nyh <notifications@github.com> writes:

> On Sun, Sep 8, 2019 at 6:03 AM Rafael Ávila de Espíndola <
> notifications@github.com> wrote:
>
>> "type-erased good, type expanded bad", without looking at what the patch
>> is doing makes as much sense as "Four legs good, two legs bad."
>>
>
> All patches are equal, but some patches are more equal than others :-)

Could be the project motto!

Cheers,
Rafael

kostja:
@espindola people are biased it is part of human nature; the good thing is that sum of all biases creates a better product.

PR #4913, opened 2019-08-27
remove date type

espindola:
This is split in two patches so that it is easy to drop the last one if it turns out we need to keep date_type.

### DIFF
--- a/concrete_types.hh
+++ b/concrete_types.hh
@@ -118,13 +118,6 @@ struct bytes_type_impl final : public concrete_type<bytes> {
     bytes_type_impl();
 };
 
-// This is the old version of timestamp_type_impl, but has been replaced as it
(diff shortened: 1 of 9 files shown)


denesb:
We should either commit both patches or none. If a type can theoretically can be created (even if not through CQL) it should be used by `random_schema`.
BTW have you checked that it is not possible to create the type though thrift either?

espindola:
I will move the tests removal to the second patch.

I never looked at thrift. What is a good point to take a look?

denesb:
I'm not too familiar with thrift, look in the `thrift` folder. I vaguely recall that in thrift you could specify the name of the type class you wanted to use, e.g. `org.apache.cassandra.db.marshal.DataType`, so you might be able to instantiate types not accessible from CQL.

espindola:
Abandoned in favour of https://github.com/scylladb/scylla/pull/4931

PR #4915, opened 2019-08-27
do not keep the scylla metadata in memory

glommer:
There is no reason to keep the Scylla Metadata component in memory
after it is read, parsed, and its information fed into the SSTable.

We have seen systems in which the Scylla metadata component is one
of the heaviest memory users, more than the Summary and Filter.

The run identifier is trivially small, so are the feature bits. The
shard information is stored post-parsed once we call
compute_shards_for_this_sstable().

This patch proposes to just get rid of the metadata component copy
in memory altogether once all of that information is already read.

Tests: unit (release), + manual scylla write load + reshard.

### DIFF
--- a/sstables/sstables.cc
+++ b/sstables/sstables.cc
@@ -1275,6 +1275,19 @@ future<> sstable::open_data() {
         if (_shards.empty()) {
             _shards = compute_shards_for_this_sstable();
         }
+        auto* sm = _components->scylla_metadata->data.get<scylla_metadata_type::Sharding, sharding_metadata>();
+        if (!sm) {
+            return make_ready_future<>();
+        }
+        auto c = &sm->token_ranges.elements;
+        // Sharding information uses a lot of memory and once we're doing with this computation we will no longer use it.
+        return do_until([c] { return c->empty(); }, [c] {
+            c->pop_back();
+            return make_ready_future<>();
+        }).then([this, c] () mutable {
+            c = {};
+            return make_ready_future<>();
+        });
     });
 }
 


avikivity:
This pull request has random commits in it.

distributed_loader.cc
```diff
@@ -330,9 +330,9 @@ future<> distributed_loader::open_sstable(distributed<database>& db, sstables::e
                     shards_interested_in_this_sstable.insert(shard_responsible_for_generation);
```
avikivity: subject should have susbsystem prefix, so people understand what scylla metadata is being not kept in memory.

sstables/sstables.cc
```diff
@@ -1258,6 +1258,24 @@ future<> sstable::open_data() {
...
+        return seastar::async([this] {
+            _features = has_scylla_component() ? _scylla_metadata->get_features() : sstable_enabled_features{};
+            auto* sm = has_scylla_component() ? _scylla_metadata->data.get<scylla_metadata_type::Sharding, sharding_metadata>() : nullptr;
+            // Destruction of large containers can cause stalls, so protect against it.
+            while (sm && !sm->token_ranges.elements.empty()) {
+                sm->token_ranges.elements.pop_back();
+                if (need_preempt()) {
+                    seastar::thread::yield();
+                }
+            }
```
avikivity: This is maybe_yield()

sstables/sstables.cc
```diff
@@ -1284,7 +1302,9 @@ future<> sstable::update_info_for_opened_data() {
     }).then([this] {
         this->set_clustering_components_ranges();
         this->set_first_and_last_keys();
-        _run_identifier = _components->scylla_metadata->get_optional_run_identifier().value_or(utils::make_random_uuid());
+        if (_scylla_metadata) {
+            _run_identifier = _scylla_metadata->get_optional_run_identifier().value_or(utils::make_random_uuid());
```
avikivity: Should we even have a _scylla_metadata member? if we can get rid of it we can be sure we don't have some code path that keeps is around.

distributed_loader.cc
avikivity: There is no way to attach a comment to the suject, so I attached it to the first line of the patch.

You can tell that I referred to the subject, by the fact that the word "subject" is used as the subject of the sentence.

sstables/sstables.cc
avikivity: Yes, agree that it's not easy to get rid of it for the write path. One day we'll move it to the sstable writer class.

sstables/sstables.cc
```diff
@@ -2149,18 +2166,14 @@ components_writer::~components_writer() {
...
-    return read_toc().then([this, &pc] {
-        _components->scylla_metadata.emplace();  // engaged optional means we won't try to re-read this again
-        if (!has_component(component_type::Scylla)) {
-            return make_ready_future<>();
-        }
-        return read_simple<component_type::Scylla>(*_components->scylla_metadata, pc);
-    });
+    sstables::scylla_metadata scylla_metadata;
+    read_simple<component_type::Scylla>(scylla_metadata, pc).get();
+    return make_ready_future<std::optional<sstables::scylla_metadata>>(scylla_metadata);
```
bhalevy: This is awkward and will only work when called within an async thread.
How about passing a reference to a `std::optional<sstables::scylla_metadata>` to the function and engage it here and then`return read_simple<component_type::Scylla>(*scylla_metadata, pc);`

glommer:
On Thu, Aug 29, 2019 at 2:08 AM Benny Halevy <notifications@github.com>
wrote:

> *@bhalevy* requested changes on this pull request.
>
> In sstables/sstables.cc
> <https://github.com/scylladb/scylla/pull/4915#discussion_r318898393>:
>
> >      }
> -    return read_toc().then([this, &pc] {
> -        _components->scylla_metadata.emplace();  // engaged optional means we won't try to re-read this again
> -        if (!has_component(component_type::Scylla)) {
> -            return make_ready_future<>();
> -        }
> -        return read_simple<component_type::Scylla>(*_components->scylla_metadata, pc);
> -    });
> +    sstables::scylla_metadata scylla_metadata;
> +    read_simple<component_type::Scylla>(scylla_metadata, pc).get();
> +    return make_ready_future<std::optional<sstables::scylla_metadata>>(scylla_metadata);
>
> This is awkward and will only work when called within an async thread.
> How about passing a reference to a
> std::optional<sstables::scylla_metadata> to the function and engage it
> here and thenreturn read_simple<component_type::Scylla>(*scylla_metadata,
> pc);
>

Alternatively I can add an assert for being in a thread.
In general all the SSTable code lives in threads already anyway, so I find
going out of our way to make it non-threaded doesn't have a lot of pay-off
(although this particular code is quite simple so it wouldn't be hard)


>

sstables/sstables.cc
avikivity: Isn't github wonderful.

glommer: I changed this, @bhalevy

glommer:
Force pushed. Changed read_scylla_metadata not to assume it's in a thread as per Benny's suggestion. Fixed the subject as per Avi's review.

bhalevy:
> Force pushed. Changed read_scylla_metadata not to assume it's in a thread as per Benny's suggestion. Fixed the subject as per Avi's review.

Thanks. Looks better now.

slivne:
what about the extension_attributes - they are used in enterprise AFAIK 

why can't we simply drop the parts that we do not need from the scylla_component, the sharding information and not drop everything else ?

elcallio:
This patch would completely break E@R, since this needs access to scylla_metadata, both for opening files, and for writing them. 
I agree with @bhalevy, if you need the component to take less space, why not just not store the parts that are not needed during the sstable lifetime (sharding?), but let the component exist?

glommer:
My original implementation did (almost) exactly that.
Avi requested that I just did away with the component altogether.

He probably was unaware of your usage - and so was I, otherwise I'd have pushed back.

I'll rework with this in mind, thanks.

glommer:
Force pushed.

@slivne , @elcallio let me know if this is more to your liking

sstables/sstables.cc
```diff
@@ -1275,6 +1275,17 @@ future<> sstable::open_data() {
         if (_shards.empty()) {
             _shards = compute_shards_for_this_sstable();
         }
+        // Sharding information uses a lot of memory and once we're doing with this computation we will no longer use it.
+        // Destruction of large containers can cause stalls, so protect against it.
+        // This is not perf critical and a thread looks better.
+        return seastar::async([this] {
+            auto* sm = _components->scylla_metadata->data.get<scylla_metadata_type::Sharding, sharding_metadata>();
+            while (sm && !sm->token_ranges.elements.empty()) {
```
avikivity: This check for sm in the loop can be moved outside.

sstables/sstables.cc
```diff
@@ -1275,6 +1275,17 @@ future<> sstable::open_data() {
...
+                sm->token_ranges.elements.pop_back();
+                seastar::thread::maybe_yield();
+            }
+        });
+        return make_ready_future<>();
     });
```
avikivity: Without freeing the container, the memory is still in use.

avikivity:
Fixes #4951

sstables/sstables.cc
avikivity: Can be simplified to

```c++
return do_until([c = &sm->token_ranges.elements] { return c->empty(); },
    c->pop_back();
}).then([this] {
    sm->_components.scylla_metadata = {}
});
```

But _components is a foreign_ptr, please verify this is never called after we shared it.

elcallio:
From a scylla_metadata user perspective I am fine with the modified PR (assuming avis comments are addressed)

sstables/sstables.cc
glommer: Indeed `_components` is a foreign_ptr, but it seems to be only invoked remotely by `get_open_info` that does this:

```
future<foreign_sstable_open_info> sstable::get_open_info() & {
    return _components.copy().then([this] (auto c) mutable {                                                                                                               
        return foreign_sstable_open_info{std::move(c), this->get_shards_for_this_sstable(), _data_file.dup(), _index_file.dup(),                                           
            _generation, _version, _format};                                                                                                                               
    });
}  
```

First weird thing here is that it actually creates a `.copy()`. What is the point of having shared components if we create a copy ?

But one way or another, the sharding information is already copied post-parsed. So we never access that particular container cross-CPU even if we do access the `_components` pointer

avikivity: It copy()s the pointer, not the value.

avikivity: Yes, cross-cpu loads don't use it:

```c++
future<> sstable::load(sstables::foreign_sstable_open_info info) {
    return read_toc().then([this, info = std::move(info)] () mutable {
        _components = std::move(info.components);
        _data_file = make_checked_file(_read_error_handler, info.data.to_file());
        _index_file = make_checked_file(_read_error_handler, info.index.to_file());
        _shards = std::move(info.owners);
        validate_min_max_metadata();
        validate_max_local_deletion_time();
        validate_partitioner();
        return update_info_for_opened_data();
    });
}
```

glommer: Yes, my understanding from the code above is that we're using the post-parsed information recorded in `owners`

avikivity: It is somewhat fragile since we keep the object in memory. It would be better to read that as a temporary and never store it in the sstable, but we can't, because it's part of a larger object that does need to be kept in the sstable.

PR #4931, opened 2019-08-29
simplify date type

espindola:
With this patch series one has to be explicit to create a date_type_impl and now there is only the one documented difference between date_type_impl and timestamp_type_impl.


cql3/untyped_result_set.hh
```diff
@@ -67,8 +77,8 @@ public:
         return *_data.at(name);
     }
     template<typename T>
-    T get_as(const sstring& name) const {
-        return value_cast<T>(data_type_for<T>()->deserialize(get_blob(name)));
+    typename cast_type_for<T>::type get_as(const sstring& name) const {
+        return value_cast<typename cast_type_for<T>::type>(data_type_for<T>()->deserialize(get_blob(name)));
```
denesb: I don't understand this change. What are you trying to achieve with this change?

types.cc
```diff
@@ -752,7 +752,7 @@ struct is_byte_order_equal_visitor {
     bool operator()(const reversed_type_impl& t) { return t.underlying_type()->is_byte_order_equal(); }
     bool operator()(const string_type_impl&) { return true; }
     bool operator()(const bytes_type_impl&) { return true; }
-    bool operator()(const date_type_impl&) { return true; }
+    bool operator()(const concrete_type<db_clock::time_point>&) { return true; }
```
denesb: At the very least we need an alias for this, with the old name and new implementation.

types.cc
```diff
@@ -1954,13 +1952,13 @@ struct serialize_visitor {
             *out++ = char(*v);
         }
     }
-    void operator()(const date_type_impl& t, const date_type_impl::native_type* v) {
-        if (v->empty()) {
+    void operator()(const concrete_type<db_clock::time_point>& t, const emptyable<db_clock::time_point>* v1) {
```
denesb: I don't understand the `emptyable`.

types.cc
```diff
@@ -843,7 +842,7 @@ abstract_type::cql3_kind abstract_type::get_cql3_kind_impl() const {
         cql3_kind operator()(const simple_date_type_impl&) { return cql3_kind::DATE; }
         cql3_kind operator()(const utf8_type_impl&) { return cql3_kind::TEXT; }
         cql3_kind operator()(const time_type_impl&) { return cql3_kind::TIME; }
-        cql3_kind operator()(const timestamp_type_impl&) { return cql3_kind::TIMESTAMP; }
+        cql3_kind operator()(const concrete_type<db_clock::time_point>&) { return cql3_kind::TIMESTAMP; }
```
denesb: Just above, you replaced `date_type_impl` with `concrete_type<db_clock::time_point>`, now you do the same with `timestamp_type_impl`. Are the two types essentially the same now? I'm confused.

cql3/untyped_result_set.hh
espindola: It is a temporary change to allow get_as<date_type_native_type>. It is removed in a followup patch.

The idea is to first require all code to use date_type_native_type or timestamp_type_native_type. That way we know we are not accidentally changing code that really wants to use date_type.

types.cc
espindola: Not sure I understand. The base class for date_type_impl and timestamp_type_impl is concrete_type<db_clock::time_point>. That is how this patch removes duplicated code.

I can add a

using timestamp_date_base_class = concrete_type<db_clock::time_point>;

Is that what you are suggesting?

types.cc
espindola: That is the type that is stored in the data_value.

types.cc
espindola: It is the common base class. The one difference between date_type and timestamp_type is how they are compared.

espindola:
I have pushed a version that uses an alias for the base class.

espindola:
Any comments now that this has an alias for the base class?

espindola:
merge ping

espindola:
merge ping 2

kostja:
Like the entire stack.

espindola:
pushed a rebase version

cql3/functions/time_uuid_fcts.hh
```diff
@@ -111,7 +111,7 @@ make_date_of_fct() {
             return {};
         }
         auto ts = db_clock::time_point(db_clock::duration(UUID_gen::unix_timestamp(UUID_gen::get_UUID(*bb))));
-        return {timestamp_type->decompose(date_type_native_type{ts})};
+        return {timestamp_type->decompose(ts)};
     });
```
avikivity: It's strange, but dateof() returns a timestamp instead of a date. Cassandra does the same, and did so since timestamp type was added.

espindola:

> avikivity commented on this pull request.
>
>
>
>> @@ -111,7 +111,7 @@ make_date_of_fct() {
>              return {};
>          }
>          auto ts = db_clock::time_point(db_clock::duration(UUID_gen::unix_timestamp(UUID_gen::get_UUID(*bb))));
> -        return {timestamp_type->decompose(date_type_native_type{ts})};
> +        return {timestamp_type->decompose(ts)};
>      });
>
> It's strange, but dateof() returns a timestamp instead of a date. Cassandra does the same, and did so since timestamp type was added.

Yes, my understanding is that there are only two ways to get a date
type:

* An old sstable
* thrift

PR #4934, opened 2019-08-30
Support non-frozen UDTs.

kbr-:
This change allows creating tables with non-frozen UDT columns. Such columns can then have single fields modified or deleted.

I had to do some refactoring first. Please read the initial commit messages, they are pretty descriptive of what happened (read the commits in the order they are listed on my branch: https://github.com/kbr-/scylla/commits/udt, starting from https://github.com/kbr-/scylla/commit/8eee36ef56df2782f413ce6d0dbb19cf5ff2d203, in order to understand them). I also wrote a bunch of documentation in the code.

@haaawk @psarna @avikivity please review.
@espindola maybe you'd be interested in this since you're working on UDFs. There is at least a small overlap (see the "generalize function_call" commit). May be trivial, but might be bigger from the UDF side.

Fixes https://github.com/scylladb/scylla/issues/2201.


cql3/functions/functions.cc
```diff
@@ -463,23 +465,32 @@ function_call::contains_bind_marker() const {
...
+        return fragmented_temporary_buffer::view{};
+    };
+
+    auto& ret_type = fun->return_type();
 
-    auto ctype = static_pointer_cast<const collection_type_impl>(fun->return_type());
-    fragmented_temporary_buffer::view res;
-    if (result) {
-        res = fragmented_temporary_buffer::view(bytes_view(*result));
+    if (ret_type->is_collection()) {
```
espindola: This might be better as a visit() now.

collection_mutation.hh
```diff
@@ -0,0 +1,121 @@
...
+//  for lists: the key is a timeuuid identifying the list entry, the cell contains the serialized value,
+//  for user types: the key is an index identifying the field, the cell contains the value of the field.
+//  The mutation may also contain a collection-wide tombstone.
+class collection_mutation {
+public:
+    using imr_object_type =  imr::utils::object<data::cell::structure>;
+    imr_object_type _data;
+
+    collection_mutation() {}
+    collection_mutation(const abstract_type&, collection_mutation_view);
```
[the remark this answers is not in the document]
haaawk: We discussed this with Kamil quite a lot during the development of this patch. If I remember correctly, initially, Kamil has changed the name but then it turned out that this object is a mutation of collection of cells. Both UDTs and tuples if are not frozen are a collection of cells and fits this as well. In the end, we were not able to come up with a better name to describe a mutation of collection of cells. Would you have any suggestions @espindola?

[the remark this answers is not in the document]
kbr-: Actually I'd be glad to give it a better name.
As @haaawk said it is a "mutation of a collection of cells", but it's still easy to get confused by someone who's reading some code involving collection_mutations for the first time and assume that they only deal with collection types (fortunately, the doc comment explains that it involves UDTs too).

We already have names for the set of non-frozen collections and non-frozen UDTs: "non-atomic" or "multi-cell". So I thought about renaming collection_mutation to multi_cell_mutation. But then I didn't know what to do with "atomic_cell_or_collection", which is a variant<atomic_cell, collection_mutation> :D 
(actually, atomic_cell_or_collection could probably be called column_mutation, because it represents a mutation of a single column entry in a single row, if I understand correctly).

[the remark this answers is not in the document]
kostja: Why call it a mutation in the first place? A mutation is Scylla's word for a database row - in the sense that it is a LSM representation of a database row. This class applies to one cell, not an entire row.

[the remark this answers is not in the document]
haaawk: @kostja I don't think mutation means database row in Scylla. It very often means the whole partition. In fact mutation class defined in mutation.hh represents the mutation of the whole partition.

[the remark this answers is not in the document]
kostja: You can see from the fact that we have this discussion the term is abused. Let's have partition_mutation, row_mutation and cell_mutation as base terms. Then we can have collection_cell_mutation as a class name. But that would require consistent application in other domains in the source code.

psarna:
Can somebody explain the order of commits appearing in a github pull request to me? `kbr-:udt` branch suggests that the first commit of this series is https://github.com/kbr-/scylla/commit/8eee36ef56df2782f413ce6d0dbb19cf5ff2d203 , yet in the pull request it shows somewhere in the middle of all the commits. How can I achieve seeing them in the chronological order?

collection_mutation.hh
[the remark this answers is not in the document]
kbr-: > collection_cell_mutation
more like collection_of_cells_mutation, or just cell_collection_mutation.

Here's how I understand all this stuff:

a cell mutation, or simply a cell, is the smallest (atomic) possible change (modification, or mutation) you can make to the database. It contains liveness information (TTL, timestamp, tombstone, etc.), and most of the time, additional data.

Mutations can be composed to obtain "bigger" mutations. The composition operation should be associative (`(a . b) . c == a . (b . c)`). Mutations should also be idempotent (`a . a = a`). Simple algebraic stuff.

Confusion arises when we try to mix the algebraic mutation stuff with the DB terminology (columns, rows, partitions, tables). A database is just a (possibly very big) mutation. Mutations can be applied to a different database (i.e., composed with) to obtain a modified (mutated) database.

In our code, the word "cell" is overloaded. One time it means the smallest possible mutation (see the `atomic_cell.hh` header), in other place it means `atomic_cell_or_collection`, or in other words, a mutation of a single column in a single row:
`void set_cell(const clustering_key_prefix& prefix, const column_definition& def, atomic_cell_or_collection&& value);` (from `mutation.hh`).
So a mutation of a column is either an `atomic_cell` or a collection of atomic cells, i.e. `collection_mutation`. This variant is represented as `atomic_cell_or_collection`. The first case is used for "atomic types" (int, text, frozen map, frozen UDT, etc.), and the other for "non-atomic" types (map, set, list, non-frozen UDT).

When `atomic_cell_or_collection`s are gathered to form a row, they are called `row` (see mutation_partition.hh). It could be called row_mutation. Then `row` is wrapped into a `deletable_row`, which is then wrapped into a `rows_entry`, and a bunch of such entries is called `mutation_partition` (should be `partition_mutation` IMO, because it's really a "mutation of a partition").

The codebase unfortunately often mixes interface with implementation: `atomic_cell_or_collection` is calling the class based on its implementation. The interfacey way to call it would probably be `column_mutation`, or just `column`.
On the other hand, `row` is calling the class based on interface. The implementation way would be calling it `cell_set` or `cell_sequence` or whatever.

mutation_writer/timestamp_based_splitting_writer.cc
```diff
@@ -219,12 +219,10 @@ std::optional<timestamp_based_splitting_mutation_writer::bucket_id> timestamp_ba
     if (cdef.is_atomic()) {
         return _classifier(cell.as_atomic_cell(cdef).timestamp());
     }
-    if (cdef.type->is_collection()) {
+    if (cdef.type->is_collection() || cdef.type->is_user_type()) {
```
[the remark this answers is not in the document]
kbr-: > Can you think of a name for "collection or user type"?

Probably `!cdef.type->is_atomic()`, or `cdef.type->is_multi_cell()`. I will change it.

> BTW, we implement user types on top of tuple. Do we want to support modifying a single element of a tuple too?

Not in the nearest future I think. Cassandra doesn't support it.

cql3/functions/functions.cc
kbr-: Thought so too, but then I saw `static_pointer_cast<const list_type_impl>(ret_type)`: `from_serialized` expects a `data_type`.
But I agree, so I'll add a commit that modifies `from_serialized` to take `const abstract_type&`, and then use `visit` here.

kbr-:
@psarna 
https://github.com/kbr-/scylla/commit/8eee36ef56df2782f413ce6d0dbb19cf5ff2d203 is indeed the first commit of the series. My branch is the source of truth. Github did something weird with the pull request.

cql3/selection/selectable.cc
```diff
@@ -142,20 +142,19 @@ shared_ptr<selector::factory>
 selectable::with_field_selection::new_selector_factory(database& db, schema_ptr s, std::vector<const column_definition*>& defs) {
     auto&& factory = _selected->new_selector_factory(db, s, defs);
     auto&& type = factory->new_instance()->get_type();
-    auto&& ut = dynamic_pointer_cast<const user_type_impl>(type->underlying_type());
```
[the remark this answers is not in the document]
kbr-: Philosophical reason: checking if the result of `dynamic_cast` is null is a very implementation-focused way to check if something is a user type, which I don't like. But then there's a `static_cast` further down below instead, so not really much of an improvement...

I did it for efficiency more than anything else, since `dynamic_cast`s are slow. It probably doesn't matter here, but the fewer dynamic casts we have, the better I feel :P

denesb:
@kbr- can you please force-push a branch that contains only the to-be reviewed commits?

> read the commits in chronological order to understand them

That is not the order in which commits are listed when I go to your [branch](https://github.com/kbr-/scylla/commits/udt). And I bet that the PR is listing them in yet another order.
I fail to understand why can't Github just list commits in their natural order.

kbr-:
@denesb sorry, I got confused. I didn't mean chronological order, I meant the order they are listed in my branch. I did a lot of rebasing and Git doesn't change the timestamps or whatever. I edited the first post.

> can you please force-push a branch that contains only the to-be reviewed commits?

https://github.com/kbr-/scylla/commits/udt is the branch, and all commits starting from "Remove mutation_partition_applier" are to be reviewed, in the listed order (bottom-to-top).

cql3/selection/selectable.cc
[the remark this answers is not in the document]
avikivity: This will be converted to a visitor One Day (tm) anyway.

types/user.hh
```diff
@@ -51,6 +51,7 @@ public:
     bytes_view field_name(size_t i) const { return _field_names[i]; }
     sstring field_name_as_string(size_t i) const { return _string_field_names[i]; }
     const std::vector<bytes>& field_names() const { return _field_names; }
+    std::optional<size_t> idx_of_field(const bytes& name) const;
     sstring get_name_as_string() const;
```
avikivity: Why size_t? This isn't defined by the size of a pointer on the machine it is running on, but is a fixed limit.

cql3/Cql.g
```diff
@@ -1396,8 +1396,9 @@ columnOperation[operations_type& operations]
 
 columnOperationDifferentiator[operations_type& operations, ::shared_ptr<cql3::column_identifier::raw> key]
     : '=' normalColumnOperation[operations, key]
-    | '[' k=term ']' specializedColumnOperation[operations, key, k, false]
-    | '[' K_SCYLLA_TIMEUUID_LIST_INDEX '(' k=term ')' ']' specializedColumnOperation[operations, key, k, true]
+    | '[' k=term ']' collectionColumnOperation[operations, key, k, false]
+    | '.' field=ident udtColumnOperation[operations, key, field]
+    | '[' K_SCYLLA_TIMEUUID_LIST_INDEX '(' k=term ')' ']' collectionColumnOperation[operations, key, k, true]
     ;
```
avikivity: This should be the last patch, no?

cql3/Cql.g
```diff
@@ -1396,8 +1396,9 @@ columnOperation[operations_type& operations]
 
 columnOperationDifferentiator[operations_type& operations, ::shared_ptr<cql3::column_identifier::raw> key]
     : '=' normalColumnOperation[operations, key]
-    | '[' k=term ']' specializedColumnOperation[operations, key, k, false]
-    | '[' K_SCYLLA_TIMEUUID_LIST_INDEX '(' k=term ')' ']' specializedColumnOperation[operations, key, k, true]
+    | '[' k=term ']' collectionColumnOperation[operations, key, k, false]
+    | '.' field=ident udtColumnOperation[operations, key, field]
```
avikivity: Why this rename?

denesb:
If I add comments to the commits in your branch, they are just added here without any context.... :man_facepalming:

kbr-:
Awesome :D just use my branch to see the order of commits then.

tests/random_schema.cc
```diff
@@ -263,6 +263,28 @@ std::vector<data_value> generate_frozen_tuple_values(std::mt19937& engine, value
...
 }
 
+data_model::mutation_description::collection generate_user_value(std::mt19937& engine, const user_type_impl& type,
+        value_generator& val_gen) {
+    using md = data_model::mutation_description;
+
+    auto key_dist = std::uniform_int_distribution<uint16_t>(0, type.size() - 1);
+
+    std::map<uint16_t, md::atomic_value> collection;
+    for (size_t i = 0; i < type.size(); ++i) {
```
denesb: Instead of type.size() we should generate a number in the range [min_valid_value, type.size()), where min_valid_value is 1, as I suppose empty mutations don't make sense (do they?).

tests/random_schema.cc
```diff
@@ -263,6 +263,28 @@ std::vector<data_value> generate_frozen_tuple_values(std::mt19937& engine, value
...
+        auto key = key_dist(engine);
+        auto member_type = type.type(key);
```
denesb: This can generate the same value multiple times. Can a single UDT mutations contain multiple mutations for the same cell?

denesb:
> Awesome :D just use my branch to see the order of commits then.

Awesome indeed. Now I get to to have a list of commits on one windows (where the commit order is correct), then copy-paste the current commit id into the find field on the other window (where I can comment). So much better then emails. :D

cql3/user_types.cc
```diff
@@ -295,4 +295,25 @@ void user_types::setter::execute(mutation& m, const clustering_key_prefix& row_k
     }
 }
 
+void user_types::setter_by_field::execute(mutation& m, const clustering_key_prefix& row_key, const update_parameters& params) {
+    assert(column.type->is_user_type() && column.type->is_multi_cell());
+
+    auto value = _t->bind(params._options);
+    if (value == constants::UNSET_VALUE) {
```
avikivity: Why not bind_and_get? It's faster.

I'm missing how we'll get UNSET_VALUE here, if _t is a marker.

cql3/user_types.cc
```diff
@@ -295,4 +295,25 @@ void user_types::setter::execute(mutation& m, const clustering_key_prefix& row_k
...
+        return;
+    }
+
+    auto& type = static_cast<const user_type_impl&>(*column.type);
+
+    auto idx = type.idx_of_field(_field_name);
+    assert(idx && *idx < type.size());
+
```
avikivity: Shoud we not have computed idx during prepare?

This assert() is dangeous, if we made a mistake then this can crash the server. As long as we capture the type during prepare everything should be self-consistent. Better to throw something.

cql3/user_types.cc
```diff
@@ -295,4 +295,25 @@ void user_types::setter::execute(mutation& m, const clustering_key_prefix& row_k
...
+    collection_mutation_description mut;
+    mut.cells.emplace_back(serialize_field_index(*idx), value
+                ? params.make_cell(*type.type(*idx), *value->get(params._options), atomic_cell::collection_member::yes)
+                : make_dead_cell(params));
+
+    m.set_cell(row_key, column, mut.serialize(type));
+}
```
avikivity: set_cell() is misnamed, it merges the input into m, or this would behave badly when we set two fields in the same column.

types.cc
```diff
@@ -1555,21 +1555,26 @@ void abstract_type::validate(bytes_view v, cql_serialization_format sf) const {
...
-static void serialize_aux(const tuple_type_impl& t, const tuple_type_impl::native_type* v, bytes::iterator& out) {
-    auto do_serialize = [&out] (auto&& t_v) {
-        const data_type& t = boost::get<0>(t_v);
-        const data_value& v = boost::get<1>(t_v);
+static void serialize_aux(const tuple_type_impl& type, const tuple_type_impl::native_type* val, bytes::iterator& out) {
+    assert(val);
+    auto& elems = *val;
+
+    assert(elems.size() <= type.size());
+
```
avikivity: Is illegal to remove fields from UDTs?

types.cc
```diff
@@ -1888,6 +1893,58 @@ data_value abstract_type::deserialize(bytes_view v) const {
...
+    // when the only difference between the tuples is that one of them has additional trailing nulls,
+    // we consider them equal. For example, in the following CQL scenario:
+    // 1. create type ut (a int);
+    // 2. create table cf (a int primary key, b frozen<ut>);
+    // 3. insert into cf (a, b) values (0, (0));
+    // 4. alter type ut add b int;
+    // 5. select * from cf where b = {a:0,b:null};
+    // the row with a = 0 should be returned, even though the value stored in the database is shorter
+    // (by one null) than the value given by the user.
+
```
avikivity: Maybe it's better to deserialize undersized UDTs to the full length. Then we kill the complexity then and there, and the rest of the code can deal with full-sized UDTs.

db/view/view.cc
```diff
@@ -70,6 +70,7 @@
 #include "service/storage_service.hh"
 #include "view_info.hh"
```
avikivity: @psarna / @nyh please review this patch.

collection_mutation.cc
```diff
@@ -23,10 +23,10 @@
 
 #include "collection_mutation.hh"
 
-collection_mutation::collection_mutation(const collection_type_impl& type, collection_mutation_view v)
+collection_mutation::collection_mutation(const abstract_type& type, collection_mutation_view v)
     : _data(imr_object_type::make(data::cell::make_collection(v.data), &type.imr_state().lsa_migrator())) {}
```
avikivity: It's dangerous to leave it with such a midleading name.

I suggest to rename it to "aggregate_cell_mutation" or maybe uglier but even less ambiguous "multi_cell_mutation". Let's agree on a name now, but only do it as a follow up to avoid disturbing the series.

avikivity: Ah, I see it is discussed at some length below. But github will show my comment first!

collection_mutation.cc
```diff
@@ -297,9 +299,6 @@ collection_mutation difference(const abstract_type& type, collection_mutation_vi
 
 collection_mutation_view_description
 deserialize_collection_mutation(const abstract_type& type, bytes_view in) {
-    assert(type.is_collection());
```
avikivity: Again we should update the name.

collection_mutation.cc
```diff
@@ -309,15 +308,35 @@ deserialize_collection_mutation(const abstract_type& type, bytes_view in) {
...
+    visit(type, make_visitor(
+        [&] (const collection_type_impl& ctype) {
+            // value_comparator(), ugh
+            auto& type_info = ctype.value_comparator()->imr_state().type_info();
+            for (uint32_t i = 0; i != nr; ++i) {
+                // FIXME: we could probably avoid the need for size
+                auto ksize = read_simple<uint32_t>(in);
+                auto key = read_simple_bytes(in, ksize);
+                auto vsize = read_simple<uint32_t>(in);
+                auto value = atomic_cell_view::from_bytes(type_info, read_simple_bytes(in, vsize));
```
avikivity: How do we maintain compatibility with Cassandra here?

cql3/user_types.hh
```diff
@@ -71,6 +71,19 @@ public:
         virtual sstring to_string() const override;
     };
 
+    class value : public multi_item_terminal {
+        std::vector<bytes_opt> _elements;
+    public:
+        value(std::vector<bytes_opt>);
+        value(std::vector<bytes_view_opt>);
+
```
avikivity: explicit, explicit

mutation_writer/timestamp_based_splitting_writer.cc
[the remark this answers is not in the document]
avikivity: collection or tuple or user type == aggregate (can be atomic or multi_cell)

cql3/user_types.cc
```diff
@@ -229,6 +230,17 @@ cql3::raw_value_view user_types::delayed_value::bind_and_get(const query_options
...
 }
 
+shared_ptr<terminal> user_types::marker::bind(const query_options& options) {
+    auto value = options.get_value_at(_bind_index);
+    if (value.is_null()) {
+        return nullptr;
+    }
+    if (value.is_unset_value()) {
+        return constants::UNSET_VALUE;
+    }
```
avikivity: Ah, I see an answer to an earlier question, so I guess github, in its wisdom, decided to reorder the patches so it is impossible for me to understand anything.

avikivity: So again, can we use bind_and_get() (in the call site) instead of bind() to avoid the extra allocation?

If the other setters suffer from the same problem then we can fix everything later.

types.cc
```diff
@@ -3042,6 +3042,48 @@ static bytes serialize_for_cql_aux(const list_type_impl&, collection_mutation_vi
...
+    for (auto&& e : mut.cells) {
+        auto field_pos = deserialize_field_index(e.first);
+        assert(field_pos < type.size());
+
+        // Some fields don't have corresponding cells -- these fields are null.
+        while (curr_field_pos < field_pos) {
+            tmp.push_back(std::nullopt);
+            ++curr_field_pos;
+        }
+
```
avikivity: How about size tmp to the ultimate size, then use operator[] to set fields? I wonder how the client deals with trailing nulls due to fields it doesn't know about yet.

types.cc
```diff
@@ -3042,6 +3042,48 @@ static bytes serialize_for_cql_aux(const list_type_impl&, collection_mutation_vi
...
+        if (e.second.is_live(mut.tomb, false)) {
+            auto value_view = e.second.value();
+            if (value_view.is_fragmented()) {
+                const auto& v = linearized.emplace_back(value_view.linearize());
+                tmp.emplace_back(v);
+            } else {
+                tmp.emplace_back(value_view.first_fragment());
+            }
```
avikivity: This snippet should be captured in a function, it is quite generic.

types.cc
```diff
@@ -3042,6 +3042,48 @@ static bytes serialize_for_cql_aux(const list_type_impl&, collection_mutation_vi
...
+        }
+
+        ++curr_field_pos;
+    }
+
+    // Trailing null fields
+    while (curr_field_pos++ < type.size()) {
+        tmp.push_back(std::nullopt);
+    }
+
```
avikivity: I see you are resizing it anyway, so operator[] would be a bit simpler.

types.cc
```diff
@@ -3057,6 +3057,17 @@ bytes serialize_for_cql(const abstract_type& type, collection_mutation_view v, c
     });
 }
 
+bytes serialize_field_index(uint16_t idx) {
+    bytes b(bytes::initialized_later(), sizeof(uint16_t));
```
avikivity: Let's add protection against that user that creates a UDT with 65536 fields and then complains that the wrong fields gets updated.

collection_mutation.cc
```diff
@@ -242,49 +238,56 @@ collection_type_impl::merge(collection_mutation_view a, collection_mutation_view
...
 collection_mutation
 collection_type_impl::difference(collection_mutation_view a, collection_mutation_view b) const
 {
- return a.data.with_linearized([&] (bytes_view a_in) {
-  return b.data.with_linearized([&] (bytes_view b_in) {
-    auto aa = deserialize_mutation_form(a_in);
-    auto bb = deserialize_mutation_form(b_in);
+    return a.with_deserialized(*this, [&] (collection_mutation_view_description a_view) {
+    return b.with_deserialized(*this, [&] (collection_mutation_view_description b_view) {
+
```
avikivity: Please unbreak the indent

collection_mutation.hh
```diff
@@ -50,12 +54,29 @@ struct collection_mutation_view_description {
...
+
+    // Packs the data to a serialized blob.
+    collection_mutation serialize(const abstract_type&) const;
 };
 
+// Given a linearized collection_mutation_view, returns an auxiliary struct allowing the inspection of each cell.
+// The struct is an observer of the data given by the collection_mutation_view and doesn't extend its lifetime.
+// The function needs to be given the type of stored data to reconstruct the structural information.
+collection_mutation_view_description deserialize_collection_mutation(const abstract_type&, bytes_view);
+
```
avikivity: Maybe a good way to encapsulate this is to provide a function to the deserializer to observe the data. Not suggesting to do it now, and in any case it needs to be done carefully so as not to create template bloat.

collection_mutation.hh
```diff
@@ -113,3 +114,6 @@ public:
 collection_mutation merge(const abstract_type&, collection_mutation_view, collection_mutation_view);
 
 collection_mutation difference(const abstract_type&, collection_mutation_view, collection_mutation_view);
+
+// Serializes the given collection of cells to a sequence of bytes ready to be sent over the CQL protocol.
+bytes serialize_for_cql(const abstract_type&, collection_mutation_view, cql_serialization_format);
```
avikivity: I like the fact that it's a free function, it can later be moved to the native transport since that's the only thing it affects.

mutation_partition.cc
```diff
@@ -645,14 +645,15 @@ void write_cell(RowWriter& w, const query::partition_slice& slice, ::atomic_cell
...
+    if (type->is_collection() && slice.options.contains<query::partition_slice::option::collections_as_maps>()) {
+        auto& ctype = static_cast<const collection_type_impl&>(*type);
+        type = map_type_impl::get_instance(ctype.name_comparator(), ctype.value_comparator(), true);
     }
+
     w.add().write().skip_timestamp()
         .skip_expiry()
-        .write_value(ctype->to_value(v, slice.cql_format()))
+        .write_value(serialize_for_cql(*type, std::move(v), slice.cql_format()))
         .skip_ttl()
```
avikivity: (and we'd move serialize_for_cql into RowWriter)

cql3/sets.cc
```diff
@@ -277,13 +277,13 @@ sets::adder::do_add(mutation& m, const clustering_key_prefix& row_key, const upd
...
     auto set_type = dynamic_pointer_cast<const set_type_impl>(column.type);
     if (column.type->is_multi_cell()) {
-        // FIXME: mutation_view? not compatible with params.make_cell().
-        collection_type_impl::mutation mut;
-
         if (!set_value || set_value->_elements.empty()) {
             return;
         }
 
+        // FIXME: mutation_view? not compatible with params.make_cell().
```
haaawk: s/mutation_view/collection_mutation_view_description/ ?

cql3/user_types.cc
```diff
@@ -162,21 +162,28 @@ void user_types::delayed_value::collect_marker_specification(shared_ptr<variable
     }
 }
```
avikivity: Please explain in the commit message how it will be useful.

cql3/user_types.cc
```diff
@@ -162,21 +162,28 @@ void user_types::delayed_value::collect_marker_specification(shared_ptr<variable
...
 
-std::vector<cql3::raw_value> user_types::delayed_value::bind_internal(const query_options& options) {
+std::vector<bytes_opt> user_types::delayed_value::bind_internal(const query_options& options) {
     auto sf = options.get_cql_serialization_format();
-    std::vector<cql3::raw_value> buffers;
+
+    // user_types::literal::prepare makes sure that every field gets a corresponding value.
+    // For missing fields the values become nullopts.
+    assert(_type->size() == _values.size());
+
```
avikivity: What happens if we have unset fields?

avikivity: Ah, I see it is already illegal (though I don't understand why). Does it still make sense to reject unset values with non-frozen UDT?

avikivity: e.g. {a=1,b=:b,c=:c} with :b=2 and :c=unset should bind_and_get to {a=1,b=2}

cql3/user_types.cc
```diff
@@ -229,4 +229,58 @@ cql3::raw_value_view user_types::delayed_value::bind_and_get(const query_options
...
+    auto& type = static_cast<const user_type_impl&>(*column.type);
+    if (type.is_multi_cell()) {
+        // Non-frozen user defined type.
+
+        collection_mutation_description mut;
+
+        // Setting a non-frozen (multi-cell) UDT means overwriting all cells.
+        // We start by deleting all existing cells.
+
+        // TODO? (kbraun): is this the desired behaviour?
```
avikivity: Yes, collections work the same way. I guess this nullifies my unset comment.

avikivity: For collections, you can use SET c = c + ? to merge into the current value (and not create a tombstone).

cql3/user_types.cc
```diff
@@ -229,4 +229,58 @@ cql3::raw_value_view user_types::delayed_value::bind_and_get(const query_options
...
+        // This is how C* does it, but consider the following scenario:
+        // create type ut (a int, b int);
+        // create table cf (a int primary key, b ut);
+        // insert into cf json '{"a": 1, "b":{"a":1, "b":2}}';
+        // insert into cf json '{"a": 1, "b":{"a":1}}' default unset;
+        // currently this would set b.a to null. However, by specifying 'default unset',
+        // perhaps the user intended to leave b.a unchanged.
```
avikivity: Good point, but better not to diverge from Cassandra here.

types.hh
```diff
@@ -1053,6 +1053,17 @@ to_bytes_opt(bytes_view_opt bv) {
     return std::nullopt;
 }
 
+inline
+std::vector<bytes_opt>
```
avikivity: The function is too heavy to be inlined.

collection_mutation.cc
```diff
@@ -249,22 +249,19 @@ collection_mutation collection_mutation_view_description::serialize(const abstra
...
 
-collection_mutation merge(const abstract_type& type, collection_mutation_view a, collection_mutation_view b) {
-    return a.with_deserialized(type, [&] (collection_mutation_view_description a_view) {
-    return b.with_deserialized(type, [&] (collection_mutation_view_description b_view) {
-
-    assert(type.is_collection());
-    auto& ctype = static_cast<const collection_type_impl&>(type);
-
+template <typename F>
+static collection_mutation_view_description merge(collection_mutation_view_description a, collection_mutation_view_description b, F key_compare) {
```
avikivity: Please add a concept for F as a way to document it (and to provide early type checking)

collection_mutation.cc
```diff
@@ -282,16 +279,37 @@ collection_mutation merge(const abstract_type& type, collection_mutation_view a,
...
+
+    return visit(type, make_visitor(
+        [&] (const collection_type_impl& ctype) {
+            return merge(std::move(a_view), std::move(b_view), [&key_type = *ctype.name_comparator()] (bytes_view k1, bytes_view k2) {
+                return key_type.less(k1, k2);
+            });
+        },
+        [&] (const user_type_impl& utype) {
+            return merge(std::move(a_view), std::move(b_view), [] (bytes_view k1, bytes_view k2) {
+                return deserialize_field_index(k1) < deserialize_field_index(k2);
```
avikivity: We could avoid this by using short_type, no? That would limit UDTs to 32768 fields, but that should be acceptable.

cql3/cql3_type.cc
```diff
@@ -207,9 +207,9 @@ class cql3_type::raw_ut : public raw {
             throw exceptions::invalid_request_exception(format("Unknown type {}", _name));
         }
         try {
-            auto&& type = user_types->get_type(_name.get_user_type_name());
-            if (!is_frozen()) {
-                throw exceptions::invalid_request_exception("Non-frozen User-Defined types are not supported, please use frozen<>");
+            data_type type = user_types->get_type(_name.get_user_type_name());
```
avikivity: This should be guarded with a cluster feature. Otherwise, a lucky client could connect to an new node, create a table with a thawed UDT, then try to access it via an old node and get crashes (at best).

avikivity:
Good work, I had some minor comments and questions.

avikivity:
> Can somebody explain the order of commits appearing in a github pull request to me? `kbr-:udt` branch suggests that the first commit of this series is [kbr-@8eee36e](https://github.com/kbr-/scylla/commit/8eee36ef56df2782f413ce6d0dbb19cf5ff2d203) , yet in the pull request it shows somewhere in the middle of all the commits. How can I achieve seeing them in the chronological order?

Actually, github shows them in chronological order (as defined by the Date: tag in the commit). What you want is the topological order, but alas that would make sense.

Do you know if Gerrit displays in topological order?

haaawk:
In gerrit you can group commits into topic barnches -> https://hyperledger-fabric.readthedocs.io/en/release-1.4/Gerrit/best-practices.html#topic-branches

They probably present commits in topology order but that would have to be tested.

BTW Gerrit now supports commenting from email -> https://gerrit-review.googlesource.com/Documentation/intro-user.html#reply-by-email

collection_mutation.hh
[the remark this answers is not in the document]
espindola: I am still fairly new, so I don't think I can suggest a good name, other than something super generic like change_list.
I am OK with any name really, sorry for the bikeshedding.

types.cc
```diff
@@ -2876,6 +2876,16 @@ bool abstract_type::is_value_compatible_with(const abstract_type& other) const {
     return is_value_compatible_with_internal(*this, *other.underlying_type());
 }
 
+std::optional<size_t>
+user_type_impl::idx_of_field(const bytes& name) const {
+    for (size_t i = 0; i < _field_names.size(); ++i) {
```
haaawk: Would it make sense to use std::find here?

avikivity: This was copied from a non-general implementation, so it makes sense to keep it like this (and maybe make it prettier in a follow up). Code movement + refactoring in the same patch is a recipe for a stealth bug.

mutation_partition.cc
```diff
@@ -645,14 +645,15 @@ void write_cell(RowWriter& w, const query::partition_slice& slice, ::atomic_cell
 }
 
 template<typename RowWriter>
-void write_cell(RowWriter& w, const query::partition_slice& slice, const data_type& type, collection_mutation_view v) {
-    auto ctype = static_pointer_cast<const collection_type_impl>(type);
-    if (slice.options.contains<query::partition_slice::option::collections_as_maps>()) {
-        ctype = map_type_impl::get_instance(ctype->name_comparator(), ctype->value_comparator(), true);
+void write_cell(RowWriter& w, const query::partition_slice& slice, data_type type, collection_mutation_view v) {
```
haaawk: Why did you remove const& on |type|? data_type is a shared_ptr so it's cheaper to take it by reference when possible.

types.cc
haaawk: Sure. Can be in another commit. Just pointing out it looks very much like a standard algorithm.

cql3/lists.cc
```diff
@@ -166,7 +166,7 @@ lists::value::equals(shared_ptr<list_type_impl> lt, const value& v) {
             [t = lt->get_elements_type()] (const bytes_opt& e1, const bytes_opt& e2) { return t->equal(*e1, *e2); });
 }
 
-std::vector<bytes_opt>
+const std::vector<bytes_opt>&
```
haaawk: This can potentially be a source of a very nasty bug. Did you make sure reference returned from get_elements() never outlives the value object? This comment relates to the whole patch not just to the lists::value. It's just impossible to comment on a commit in github.

collection_mutation.cc
```diff
@@ -309,15 +308,35 @@ deserialize_collection_mutation(const abstract_type& type, bytes_view in) {
...
-        // value_comparator(), ugh
-        auto value = atomic_cell_view::from_bytes(ctype.value_comparator()->imr_state().type_info(), read_simple_bytes(in, vsize));
-        ret.cells.emplace_back(key, value);
-    }
+
+    visit(type, make_visitor(
+        [&] (const collection_type_impl& ctype) {
+            // value_comparator(), ugh
+            auto& type_info = ctype.value_comparator()->imr_state().type_info();
+            for (uint32_t i = 0; i != nr; ++i) {
```
haaawk: This loop and the one for UDTs look very similar. It could make sense to generalize them and just pass lambda that reads key and value in collection/udt specific way.

collection_mutation.cc
```diff
@@ -62,27 +62,47 @@ bool collection_mutation_view::is_empty() const {
...
-        in.remove_prefix(ksize);
-        auto vsize = read_simple<uint32_t>(in);
-        auto value = atomic_cell_view::from_bytes(ctype.value_comparator()->imr_state().type_info(), read_simple_bytes(in, vsize));
-        if (value.is_live(tomb, now, false)) {
-            return true;
+
+    return visit(type, make_visitor(
+        [&] (const collection_type_impl& ctype) {
+            auto& type_info = ctype.value_comparator()->imr_state().type_info();
+            for (uint32_t i = 0; i != nr; ++i) {
```
haaawk: This loop and the one for UDTs look very similar. It could make sense to generalize them and just pass lambda that reads key and value in collection/udt specific way.

collection_mutation.cc
```diff
@@ -107,24 +107,42 @@ bool collection_mutation_view::is_any_live(const abstract_type& type, tombstone
...
-        in.remove_prefix(ksize);
-        auto vsize = read_simple<uint32_t>(in);
-        auto value = atomic_cell_view::from_bytes(ctype.value_comparator()->imr_state().type_info(), read_simple_bytes(in, vsize));
-        max = std::max(value.timestamp(), max);
-    }
+
+    visit(type, make_visitor(
+        [&] (const collection_type_impl& ctype) {
+            auto& type_info = ctype.value_comparator()->imr_state().type_info();
+            for (uint32_t i = 0; i != nr; ++i) {
```
haaawk: This loop and the one for UDTs look very similar. It could make sense to generalize them and just pass lambda that reads key and value in collection/udt specific way.

collection_mutation.cc
haaawk: Would it make sense to take key_compare by F&&?

collection_mutation.cc
```diff
@@ -282,16 +279,37 @@ collection_mutation merge(const abstract_type& type, collection_mutation_view a,
...
-    merged.tomb = std::max(a_view.tomb, b_view.tomb);
+    merged.tomb = std::max(a.tomb, b.tomb);
 
-    return merged.serialize(type);
+    return merged;
+}
+
+collection_mutation merge(const abstract_type& type, collection_mutation_view a, collection_mutation_view b) {
+    return a.with_deserialized(type, [&] (collection_mutation_view_description a_view) {
+    return b.with_deserialized(type, [&] (collection_mutation_view_description b_view) {
```
haaawk: wrong indent?

collection_mutation.cc
```diff
@@ -332,35 +332,52 @@ collection_mutation merge(const abstract_type& type, collection_mutation_view a,
...
     }
-    return diff.serialize(type);
+
+    return diff;
+}
+
+collection_mutation difference(const abstract_type& type, collection_mutation_view a, collection_mutation_view b)
+{
+    return a.with_deserialized(type, [&] (collection_mutation_view_description a_view) {
+    return b.with_deserialized(type, [&] (collection_mutation_view_description b_view) {
```
haaawk: wrong indent?

cql3/statements/create_table_statement.cc
```diff
@@ -205,18 +206,30 @@ std::unique_ptr<prepared_statement> create_table_statement::raw_statement::prepa
...
-        if (pt.is_collection() && pt.get_type()->is_multi_cell()) {
-            if (!defined_multi_cell_collections) {
-                defined_multi_cell_collections = std::map<bytes, data_type>{};
+        if (pt.get_type()->is_multi_cell()) {
+            // check for multi-cell types (non-frozen UDTs or collections) inside a non-frozen UDT
+            if (pt.get_type()->is_user_type()) {
+                auto type = static_cast<const user_type_impl*>(pt.get_type().get());
+                for (auto&& inner: type->all_types()) {
+                    if (inner->is_multi_cell()) {
+                        // a nested non-frozen UDT should have already been rejected when defining the type
```
haaawk: Why are non-frozen UDT rejected when defining the type but non-frozen collections are not? Shouldn't they be treated the same way?

cql3/user_types.cc
haaawk: s/b.a/b.b/ as Sarna pointed out

collection_mutation.hh
```diff
@@ -38,6 +38,7 @@ class collection_mutation;
 // (tombstone and cells) inside the mutation easily.
 struct collection_mutation_description {
     tombstone tomb;
+    // FIXME: use iterators?
```
haaawk: What's this about? Could you please elaborate?

haaawk:
LGTM over all but I'm not an expert in many of those layers.

tests/random_schema.cc
kbr-: Observe that I'm first generating a std::map which is then converted into a mutation. If there's a collision of keys during the generating phase (the for loop you are commenting), the old value will stay (the new value won't be inserted). That is equivalent to what you suggested in the previous comment: first generate a number of elements, then generate them without key collisions. I'm just doing it in a simpler way IMO and with a different distribution as a side effect (it's harder to have more non-null fields).

tests/random_schema.cc
kbr-: Read my answer to your comment below.

tests/random_schema.cc
denesb: Ok, makes sense.

cql3/user_types.cc
kbr-: I'm not sure about our policy with assertions versus throws for bugs.
1. An assertion is (?) better than segfaulting by dereferencing a null pointer or even worse, going into undefined behaviour by accessing an array out of bounds or something,
2. But then do we want to crash the server in case like that? Or maybe we should throw there saying "we no longer go into this territory due to a bug, but we can still run the rest of the code"?

Maybe it should depend on the layer? Like when something is wrong during the CQL phase we can throw so only some features will be disabled due to a bug, but if something is wrong deeper in the code then it just doesn't make sense to keep the DB running anymore?

In any case, I'll see if I can move idx to the prepare phase. Indeed it looks like I don't need `_field_name` anywhere else.

tests/random_schema.cc
denesb: What I wanted is to have UDT mutations with only some of the fields set. I guess you already achieve this to some extent by generating colliding keys (below). Still smaller UDT mutations that have just one or two fields set will be very unlikely to be generated.

kbr-: Ok, maybe it'll be better to make the number of non-null fields uniformly distributed. I'll change this.

types/user.hh
kbr-: What else should I use?
> This isn't defined by the size of a pointer on the machine it is running on

that's not how I look at `size_t`. For me it's just a type to represent sizes and often indexes. The functions above: `field_name`, `field_name_as_string` also take `size_t` as an index type.

cql3/Cql.g
kbr-: You mean the one which adds the feature of setting single fields?
Not necessarily. The deleter patch is independent from this patch and their relative order doesn't matter. Some fixes like the trailing nulls patch can also be after this one. Also there are patches that add tests, which I could merge with the other patches, but I was already too tired with all the rebasing to split the test patches in a way that would match the other patches.

cql3/Cql.g
kbr-: Because "collectionColumnOperation" explains more than "specializedColumnOperation" IMO. And now there are two types of "specialized" operations: for collections and for UDTs, and we have to distinguish between them somehow.

types.cc
kbr-: Yes, you can only add and rename.
https://docs.datastax.com/en/dse/5.1/cql/cql/cql_reference/cql_commands/cqlAlterType.html

types.cc
kbr-: By giving the tuple type to tuple_deserializing_iterator?

cql3/user_types.cc
kbr-: > I'm missing how we'll get UNSET_VALUE here, if _t is a marker.

Indeed, I didn't catch that. In `constants::marker::bind`:

```        virtual ::shared_ptr<terminal> bind(const query_options& options) override {
            auto bytes = bind_and_get(options);
            if (!bytes) {
                return ::shared_ptr<terminal>{};
            }
            return ::make_shared<constants::value>(std::move(cql3::raw_value::make_value(to_bytes(*bytes))));
        }
```

Instead of checking `!bytes`, shouldn't we check for `unset_value` vs `null_value` and return constants::UNSET_VALUE in the unset vase?

Because right now I see that e.g. maps have the same problem:
```
cqlsh:a> create table tm (a int primary key, b map<int, int>);
cqlsh:a> update tm set b[0] = 0 where a = 0;
cqlsh:a> select * from tm;

 a | b
---+--------
 0 | {0: 0}

```
then from python driver:
```
    session.execute(session.prepare('update tm set b[0] = ? where a = 0'))
```

and we're left with:
```
cqlsh:a> select * from tm;

 a | b
---+---
```

and this happens because:
```
void
maps::setter::execute(mutation& m, const clustering_key_prefix& row_key, const update_parameters& params, const column_definition& column, ::shared_ptr<terminal> value) {
    if (value == constants::UNSET_VALUE) {
        return;
    }
```

In Cassandra it works as expected (the python driver part doesn't have any effect).

kbr-: Actually, my mistake, I was looking at wrong code in maps. The problem is in `setter_by_key` of course:
```
void
maps::setter_by_key::execute(mutation& m, const clustering_key_prefix& prefix, const update_parameters& params) {
    using exceptions::invalid_request_exception;
    assert(column.type->is_multi_cell()); // "Attempted to set a value for a single key on a frozen map"m
    auto key = _k->bind_and_get(params._options);
    auto value = _t->bind_and_get(params._options);
    if (!key) {
        throw invalid_request_exception("Invalid null map key");
    }
    auto ctype = static_cast<const map_type_impl*>(column.type.get());
    auto avalue = value ? params.make_cell(*ctype->get_values_type(), *value, atomic_cell::collection_member::yes) : params.make_dead_cell();
```

again we don't differentiate between unset/null

kbr-: Found another related bug.
When I fixed `maps::setter_by_key::execute`:
```
void
maps::setter_by_key::execute(mutation& m, const clustering_key_prefix& prefix, const update_parameters& params) {
    using exceptions::invalid_request_exception;
    assert(column.type->is_multi_cell()); // "Attempted to set a value for a single key on a frozen map"m
    auto key = _k->bind_and_get(params._options);
    auto value = _t->bind_and_get(params._options);
    if (!key) {
        throw invalid_request_exception("Invalid null map key");
    }
    if (value.is_unset_value()) {
        return;
    }
    auto ctype = static_cast<const map_type_impl*>(column.type.get());
    auto avalue = value ? params.make_cell(*ctype->get_values_type(), *value, atomic_cell::collection_member::yes) : params.make_dead_cell();
```
and did:
```
cqlsh:a> create table tmut (a int primary key, b map<int, frozen<ut>>);
cqlsh:a> update tmut set b[0] = {a:0, b:0} where a = 0;
cqlsh:a> select * from tmut;

 a | b
---+-------------------
 0 | {0: {a: 0, b: 0}}
```
then, from python driver:
```
    session.execute(session.prepare('update tmut set b[0] = ? where a = 0'))
```
guess what happened:
```
cqlsh:a> select * from tmut;

 a | b
---+---
```

But `map<int, int>` works correctly.

This is because `user_type::marker::bind_and_get` is inherited from `non_terminal`: 
```
        virtual cql3::raw_value_view bind_and_get(const query_options& options) override {
            auto t = bind(options);
            if (t) {
                return options.make_temporary(t->get(options));
            }
            return cql3::raw_value_view::make_null();
        };
```

`user_type::marker::bind` returns `constants::UNSET_VALUE`, but `options.make_temporary` changes this to null:
```
        cql3::raw_value_view query_options::make_temporary(cql3::raw_value value) const
        {
            if (value) {
                auto value_view = *value;
                auto ptr = _temporaries.write_place_holder(value_view.size());
                std::copy_n(value_view.data(), value_view.size(), ptr);
                return cql3::raw_value_view::make_value(fragmented_temporary_buffer::view(bytes_view{ptr, value_view.size()}));
            }
            return cql3::raw_value_view::make_null();
        }
```

changing `make_temporary` fixes the issue:
```
cql3::raw_value_view query_options::make_temporary(cql3::raw_value value) const
{
    if (value.is_null()) {
        return cql3::raw_value_view::make_null();
    } else if (value.is_unset_value()) {
        return raw_value_view::make_unset_value();
    }

    const auto& bytes = *value;
    auto ptr = _temporaries.write_place_holder(bytes.size());
    std::copy_n(bytes.data(), bytes.size(), ptr);
    return cql3::raw_value_view::make_value(fragmented_temporary_buffer::view(bytes_view{ptr, bytes.size()}));
}
```

collection_mutation.cc
kbr-: By storing a non-frozen UDT as a collection of cells, where each cell's key is serialized as a 2-byte short. Then `write_collection` and `write_cell` in sstables/mc/writer.cc do the right thing.

Actually I made a mistake by using uint16_t instead of int16_t. Cassandra uses Java's short type for these cell keys. The difference would only show for values >= 2^16 though (we would correctly handle such long UDTs while Cassandra would start doing weird things).

cql3/user_types.cc
kbr-: `user_types::marker::bind_and_get` will call `bind`, and then `get` as it currently is implemented (inherited from `non_terminal`). See also my monologue about bugs below your other bind_and_get comment.

We could try to implement an optimized version of `bind_and_get`, but using it in `setter` would only make the `setter` more complicated: `bind_and_get` returns a `bytes` blob, while I want to have an easy access to the separate fields of my UDT (which `multi_item_terminal::get_elements` nicely provides).

collection_mutation.hh
kbr-: How'd that be different from `collection_mutation_view::with_deserialized`?

collection_mutation.cc
kbr-: I liked it that way actually, but OK

collection_mutation.cc
kbr-: True, using uint16_t was my mistake anyway.

collection_mutation.cc
kbr-: Will use short_type instead

mutation_partition.cc
kbr-: Because of this:
`        type = map_type_impl::get_instance(ctype.name_comparator(), ctype.value_comparator(), true);
`

I could use `const abstract_type*` instead. Even though `get_instance` returns a `shared_ptr`, I could call `get()` on its result, because `type_interning_helper` keeps up the type's lifetime. That's an implementation detail though so I don't know if it's safe in the long run.

cql3/lists.cc
kbr-: Yes

cql3/statements/create_table_statement.cc
kbr-: That's how Cassandra does it :(

collection_mutation.cc
kostja: What happens when I alter the UDT and add/remove the same column 32k times? Do you run out of keys, even if the UDT has the just a handful of fields, or you do not?

kbr-: You cannot remove a UDT field.
For adding too many fields or creating a type with too many fields I'm gonna add a check.

collection_mutation.cc
kbr-: maybe, maybe not

kbr-:
v2:
- add a commit that modifies from_serialized to take const abstract_type& instead of data_type, which simplifies code in places using from_serialized (after switching to visitors in these places), removing some pointer_casts.
- make tuple_type_impl(std::vector<data_type> types, bool freze_inner); protected
- set_field: move computation of field index to prepare phase, use bind_and_get instead of bind in setter_by_field::execute
- make constructors of user_types::value explicit
- serialize_for_cql_aux(user type): resize tmp, then use operator[] instead of push_back. Introduce a helper "linearized" function.
- reject user types with 2^16 or more fields (during type creation or field addition)
- indent code inside with_deserialized(...)
- better explain why make bind_internal return vector<bytes_opt> (in commit message)
- don't inline to_bytes_opt_vec
- pass short_type for the UDT case in collection_mutation.cc: merge and difference
- s/mutation_view/collection_mutation_view_description/ in a FIXME
- use int16_t for serialization of UDT field keys, like in C*
- add a cluster feature for non-frozen UDTs
- elaborate on "FIXME: use iterators?"
- make the number of non-null fields uniformly distributed in random_schema.cc:generate_user_value
- check that the only collection types are map, set, and list in places where this assumption is made: function_call::make_terminal, insert_prepared_json_statement::execute_set_value.
- refactor functions in collection_mutation.cc to reduce redundancy

click [here](https://github.com/scylladb/scylla/compare/674de8d92353c506d7d5238576f7039f1113dd7e..efb644af8d0c8a7a14eac9a6ceb943c53c596fc3) to see the diff.

collection_mutation.hh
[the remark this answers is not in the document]
kbr-: Let's leave any further name changes, if any, for the future.

types/collection.hh
```diff
@@ -41,55 +40,14 @@ protected:
     explicit collection_type_impl(kind k, sstring name, bool is_multi_cell)
             : abstract_type(k, std::move(name), {}, data::type_info::make_collection()), _is_multi_cell(is_multi_cell) {}
 public:
-    // representation of a collection mutation, key/value pairs, value is a mutation itself
```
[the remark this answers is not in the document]
kbr-: You're welcome

cql3/selection/selectable.cc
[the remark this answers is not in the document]
kbr-: Leaving it as it is

mutation_writer/timestamp_based_splitting_writer.cc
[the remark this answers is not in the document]
kbr-: Leaving it as it is. It's basically a runtime check that says "there are no multi-cell types other than collections or UDTs".

types/tuple.hh
```diff
@@ -93,8 +93,9 @@ protected:
     static boost::iterator_range<tuple_deserializing_iterator> make_range(bytes_view v) {
         return { tuple_deserializing_iterator::start(v), tuple_deserializing_iterator::finish(v) };
     }
-    tuple_type_impl(kind k, sstring name, std::vector<data_type> types);
+    tuple_type_impl(kind k, sstring name, std::vector<data_type> types, bool freeze_inner);
 public:
+    tuple_type_impl(std::vector<data_type> types, bool freze_inner);
```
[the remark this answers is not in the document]
kbr-: Done

cql3/functions/functions.cc
kbr-: Done

types/collection.hh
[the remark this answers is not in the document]
kbr-: You're welcome

tests/random_schema.cc
kbr-: Done

cql3/user_types.cc
kbr-: I switched to `bind_and_get` in `setter_by_field::execute`, but I didn't fix the bugs with other setters and make_temporary. Another PR is more suitable for that.

cql3/user_types.cc
kbr-: Moved `idx` to `prepare`.

collection_mutation.cc
kbr-: For this PR I'll leave the name as it is.

cql3/user_types.hh
kbr-: Done

types.cc
kbr-: Done

types.cc
kbr-: Done

types.cc
kbr-: Done

types.cc
kbr-: Done, but I should add a test for that.

kbr-:
v3: added a test for checking for too many fields in UDTs.

types.cc
kbr-: And done

haaawk:
Please add all relevant Cassandra dtests to our dtests and maybe write some new.

types/user.hh
avikivity: size_t is used for data structures that can grow as much as main memory can. They're a bad choice when something doesn't depend on main memory. Since the problem already exists, we can keep it.

A good choice here is uint32_t.

collection_mutation.hh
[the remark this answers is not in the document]
avikivity: It's somewhat problematic in that the usual collections are homogeneous while tuple/UDT are heterogeneous. Collections have unlimited size, while tuple/UDT have fixed size (that can be altered, but for a particular schema, it is fixed).

Collections are ordered by their values, tuples are ordered by position.

Right now it works because you view a tuple as a map<index, blob> but as soon as we want to make use of individual types, it will break apart.

I have a feeling this merging of tuples and collections is accidental and we'll need to split it in the future into two classes.

collection_mutation.cc
avikivity: has it changed?

collection_mutation.hh
avikivity: It wouldn't. I'm not sure I understand the motivation for with_deserialized().

collection_mutation.cc
```diff
@@ -215,17 +215,19 @@ collection_mutation collection_mutation_view_description::serialize(const abstra
...
 }
 
-collection_mutation
-collection_type_impl::merge(collection_mutation_view a, collection_mutation_view b) const {
-    return a.with_deserialized(*this, [&] (collection_mutation_view_description a_view) {
-    return b.with_deserialized(*this, [&] (collection_mutation_view_description b_view) {
+collection_mutation merge(const abstract_type& type, collection_mutation_view a, collection_mutation_view b) {
+    return a.with_deserialized(type, [&] (collection_mutation_view_description a_view) {
+    return b.with_deserialized(type, [&] (collection_mutation_view_description b_view) {
+
```
avikivity: Please use half/quarter indents here to make the scopes visible.

types.cc
```diff
@@ -3057,6 +3057,22 @@ bytes serialize_for_cql(const abstract_type& type, collection_mutation_view v, c
...
 }
 
+bytes serialize_field_index(size_t idx) {
+    if (idx >= size_t(std::numeric_limits<int16_t>::max())) {
+        // should've been rejected earlier, but just to be sure...
+        throw std::runtime_error(format("index for user type field too large: {}", idx));
+    }
+
+    bytes b(bytes::initialized_later(), sizeof(int16_t));
+    *reinterpret_cast<int16_t*>(b.begin()) = net::hton(static_cast<int16_t>(idx));
```
avikivity: This violates strict aliasing. You can use write_be() to do this safely.

types.cc
```diff
@@ -3057,6 +3057,22 @@ bytes serialize_for_cql(const abstract_type& type, collection_mutation_view v, c
...
+    return b;
+}
+
+size_t deserialize_field_index(const bytes_view& b) {
+    assert(b.size() == sizeof(int16_t));
+    return net::ntoh(*reinterpret_cast<const int16_t*>(b.begin()));
+}
```
avikivity: read_be()

cql3/sets.cc
```diff
@@ -121,7 +121,7 @@ sets::literal::to_string() const {
     return "{" + join(", ", _elements) + "}";
 }
 
-sets::value
+shared_ptr<sets::value>
 sets::value::from_serialized(const fragmented_temporary_buffer::view& val, set_type type, cql_serialization_format sf) {
     try {
```
avikivity: We're trying to move away from extensive use of shared_ptr, this is making it harder.

db/view/view.cc
avikivity: @psarna / @nyh please review this patch.

db/view/view.cc
```diff
@@ -454,7 +454,7 @@ void create_virtual_column(schema_builder& builder, const bytes& name, const dat
         // A map has keys and values. We don't need these values,
         // and can use empty values instead.
         auto mtype = dynamic_pointer_cast<const map_type_impl>(type);
-        builder.with_column(name, map_type_impl::get_instance(mtype->get_values_type(), empty_type, true), column_kind::regular_column, column_view_virtual::yes);
+        builder.with_column(name, map_type_impl::get_instance(mtype->get_keys_type(), empty_type, true), column_kind::regular_column, column_view_virtual::yes);
     } else if (ctype->is_set()) {
```
avikivity: Please send this as a separate patch with an accompanying issue.

avikivity:
I made another pass. Sorry for the huge delay!

I have deep reservations about merging collections and UDTs. Yes they have the same overall shape, but we may want to treat them differently in the future (store UDTs as real tuples instead of maps, with fixed offsets for fields). The fact that only a little reuse was achieved and a lot of templates added strengthens this opinion.

I don't want to rework this patchset, so this serves as a big I-told-you-so in case we have to separate the two again.

slivne:
@kbr- are you working on a new version ?

kbr-:
@slivne 
not right now, I'm going to get back to this as soon as I push my CDC change for review.
@avikivity 
you mean we shouldn't merge UDTs and collections on the "mutation level"?
We can't store non-frozen UDTs in the same way as tuples, because tuples are atomic types. In UDTs we can have a separate timestamp, tombstone etc. for each field (that's why we store fields as separate cells).
Unless we change tuples to be stored as a collection of cells, but then that would look the same as collections.
In any case, whatever way we choose to represent UDTs internally, we will have to translate them to/from the "map of int16_t to field" representation when writing to SSTables to stay compatible with Cassandra.
As for achieving reuse, I disagree, the "generalizing" commits are very small, I think that there's been a lot of reuse.
@kostja 
Can you elaborate? Don't bother with what? Are you referring to some particular commit in this patchset?

kbr-:
v3: rebased

collection_mutation.cc
kbr-: It is fixed in a later commit.

collection_mutation.cc
kbr-: Ok, one of the commits had it wrong, another commit changed it later.
I can change it in all the affected commits.

collection_mutation.hh
kbr-: It's just convenient.
I still need to keep `deserialize_collection_mutation` in the header though to implement the template (I really don't want to write the implementation of `deserialize_collection_mutation` in the header)

db/view/view.cc
kbr-: https://github.com/scylladb/scylla/pull/5166

collection_mutation.hh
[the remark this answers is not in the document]
kbr-: In that case it should be an easy refactor (much easier than what I did here).

kbr-:
v4:
- fixed indentation in collection_mutation.cc in earlier commits
- from_serialized doesn't return shared_ptr anymore, dropped this commit
- dropped the view bugfix, it's in a separate PR now: https://github.com/scylladb/scylla/pull/5166
- use read_be/write_be in (de)serialize_field_index

kbr-:
rebased

bhalevy:
@kbr- github says this pr needs to be rebased again.
Are all review comments addressed?
What's left to merge this pr?

cql3/user_types.cc
avikivity: btw, the whole unset thing is a huge mistake, it mixes semantics with the protocol.

It would have been much better to do it in the language itself:

    UPDATE tab
    SET col1 = ? CONDITIONAL ON ?, col2 = ? CONDIONAL ON ?
    WHERE pk = ? AND ck = ?

this allows the prepare phase to distinuish between a statement that can have unset values and a statement that cannot.

avikivity:
@kbr- please list the changes, this allows reviewers to focus their review on what has changed.

avikivity:
> @denesb sorry, I got confused. I didn't mean chronological order, I meant the order they are listed in my branch. I did a lot of rebasing and Git doesn't change the timestamps or whatever. I edited the first post.
> 
> > can you please force-push a branch that contains only the to-be reviewed commits?
> 
> https://github.com/kbr-/scylla/commits/udt is the branch, and all commits starting from "Remove mutation_partition_applier" are to be reviewed, in the listed order (bottom-to-top).

Please try `git rebase --ignore-date $(git merge-base HEAD origin/master)` to reset the author dates.

kbr-:
@avikivity 
The only real changes after your last review were:

> v4:
> 
>     fixed indentation in collection_mutation.cc in earlier commits
>     from_serialized doesn't return shared_ptr anymore, dropped this commit
>     dropped the view bugfix, it's in a separate PR now: #5166
>     use read_be/write_be in (de)serialize_field_index

The other pushes were just stupid rebases.

Lemme rebase again...

kbr-:
Rebased.

One "significant" change was removing an assertion that I added in `prefetch_data_builder` that it works with collection types only. `prefetch_data_builder` has been generalized by @kostja to work with any types.

I also reset the dates as requested.

kbr-:
@avikivity please consider merging this. It's tiresome to rebase it for the `n`th time.

cql3/Cql.g
avikivity: Maybe collectionElementOperation is more descriptive.

cql3/user_types.cc
avikivity: Yes, in the front-end throwing is often better than assersions. In the back end usually one can't recover so it's better to fail fast.

PR #4950, opened 2019-09-03
install.sh: be consistent about $root/$prefix

### LINKED ISSUE #4949: install.sh: be consistent about $root/$prefix
scylla version 6c4ad932969749ee132c3a13b3c79328a1c575b5

In most case the script uses `$root/$prefix` but in one occasion a path is constructed this way:
`--set-interpreter "$prefix/libreloc/ld.so"`
This needs to be `--set-interpreter "$root/$prefix/libreloc/ld.so"`

bhalevy:
Fix #4949

- install.sh: adjust_bin: be consistent about $root/$prefix
- install.sh: fix default substitution for GNUTLS_SYSTEM_PRIORITY_FILE
- install.sh: fix default prefix in usage message

Test:
```
$ ./tools/toolchain/dbuild ./reloc/build_reloc.sh --mode dev
$ mkdir /tmp/scylla-package
$ cd /tmp/scylla-package
$ tar xzf ~/dev/scylla/build/dev/scylla-package.tar.gz
$ ./install.sh --root /tmp/install --prefix node1 --target centos --disttype redhat
$ /tmp/install/node1/bin/scylla --help
```

Also, manually validated `/tmp/install/node1/libreloc/gnutls.config`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -101,13 +101,13 @@ adjust_bin() {
     # We could add --set-rpath too, but then debugedit (called by rpmbuild) barfs
     # on the result. So use LD_LIBRARY_PATH in the thunk, below.
     patchelf \
-	--set-interpreter "$prefix/libreloc/ld.so" \
+	--set-interpreter "$root/$prefix/libreloc/ld.so" \
 	"$root/$prefix/libexec/$bin"
     cat > "$root/$prefix/bin/$bin" <<EOF
 #!/bin/bash -e
-export GNUTLS_SYSTEM_PRIORITY_FILE="\${GNUTLS_SYSTEM_PRIORITY_FILE:-$prefix/libreloc/gnutls.config}"
-export LD_LIBRARY_PATH="$prefix/libreloc"
-exec -a "\$0" "$prefix/libexec/$bin" "\$@"
+export GNUTLS_SYSTEM_PRIORITY_FILE="\${GNUTLS_SYSTEM_PRIORITY_FILE:-$root/$prefix/libreloc/gnutls.config}"
+export LD_LIBRARY_PATH="$root/$prefix/libreloc"
+exec -a "\$0" "$root/$prefix/libexec/$bin" "\$@"
 EOF
     chmod +x "$root/$prefix/bin/$bin"
 }
--- a/install.sh
+++ b/install.sh
@@ -28,7 +28,7 @@ Usage: install.sh [options]
 
 Options:
   --root /path/to/root     alternative install root (default /)
-  --prefix /prefix         directory prefix (default /usr)
+  --prefix /prefix         directory prefix (default /opt/scylladb)
   --python3 /opt/python3   path of the python3 interpreter relative to install root (default /opt/scylladb/python3/bin/python3)
   --housekeeping           enable housekeeping service
   --target centos          specify target distribution
--- a/install.sh
+++ b/install.sh
@@ -105,7 +105,7 @@ adjust_bin() {
 	"$root/$prefix/libexec/$bin"
     cat > "$root/$prefix/bin/$bin" <<EOF
 #!/bin/bash -e
-export GNUTLS_SYSTEM_PRIORITY_FILE="\${GNUTLS_SYSTEM_PRIORITY_FILE-$prefix/libreloc/gnutls.config}"
+export GNUTLS_SYSTEM_PRIORITY_FILE="\${GNUTLS_SYSTEM_PRIORITY_FILE:-$prefix/libreloc/gnutls.config}"
 export LD_LIBRARY_PATH="$prefix/libreloc"
 exec -a "\$0" "$prefix/libexec/$bin" "\$@"
 EOF


install.sh
```diff
@@ -105,7 +105,7 @@ adjust_bin() {
 	"$root/$prefix/libexec/$bin"
     cat > "$root/$prefix/bin/$bin" <<EOF
 #!/bin/bash -e
-export GNUTLS_SYSTEM_PRIORITY_FILE="\${GNUTLS_SYSTEM_PRIORITY_FILE-$prefix/libreloc/gnutls.config}"
+export GNUTLS_SYSTEM_PRIORITY_FILE="\${GNUTLS_SYSTEM_PRIORITY_FILE:-$prefix/libreloc/gnutls.config}"
 export LD_LIBRARY_PATH="$prefix/libreloc"
 exec -a "\$0" "$prefix/libexec/$bin" "\$@"
```
avikivity: What does this fix?

install.sh
```diff
@@ -101,13 +101,13 @@ adjust_bin() {
     # We could add --set-rpath too, but then debugedit (called by rpmbuild) barfs
     # on the result. So use LD_LIBRARY_PATH in the thunk, below.
     patchelf \
-	--set-interpreter "$prefix/libreloc/ld.so" \
+	--set-interpreter "$root/$prefix/libreloc/ld.so" \
 	"$root/$prefix/libexec/$bin"
```
avikivity: It's more consistent but wrong. This won't work at all.

install.sh
```diff
@@ -101,13 +101,13 @@ adjust_bin() {
...
     cat > "$root/$prefix/bin/$bin" <<EOF
 #!/bin/bash -e
-export GNUTLS_SYSTEM_PRIORITY_FILE="\${GNUTLS_SYSTEM_PRIORITY_FILE:-$prefix/libreloc/gnutls.config}"
-export LD_LIBRARY_PATH="$prefix/libreloc"
-exec -a "\$0" "$prefix/libexec/$bin" "\$@"
+export GNUTLS_SYSTEM_PRIORITY_FILE="\${GNUTLS_SYSTEM_PRIORITY_FILE:-$root/$prefix/libreloc/gnutls.config}"
+export LD_LIBRARY_PATH="$root/$prefix/libreloc"
+exec -a "\$0" "$root/$prefix/libexec/$bin" "\$@"
 EOF
```
avikivity: Again, more consistent, but less working.

install.sh
bhalevy: A syntax error.  It was supposed to provide a default value for the parameter substitution.
```
man bash(1):

       ${parameter:-word}
              Use Default Values.  If parameter is unset or null, the expansion of word is substituted.  Otherwise, the value of parameter is substituted.
       ${parameter:=word}
```

install.sh
bhalevy: Why wouldn't it work?  Isn't `prefix` supposed to be a sub-directory under `root`?

install.sh
bhalevy: So how does https://github.com/scylladb/scylla/blob/6c4ad932969749ee132c3a13b3c79328a1c575b5/install.sh#L103-L105 work?
```
103     patchelf \
104         --set-interpreter "$prefix/libreloc/ld.so" \
105         "$root/$prefix/libexec/$bin"
```

install.sh
bhalevy: I'll send it as an independent patch...
It's an orthogonal fix

install.sh
bhalevy: Is the intention that `prefix` would be related to the runtime root and the `--root` directory be used only at install time?

install.sh
avikivity: Installation has a special mindset, related to chroot. $root is where the installed filesystem is located, and the expectation is that files from $root will be copied to / (or alternatively, we will chroot to $root). This is common in rpm/deb packaging.

The idea is that during packaging, we copy the interpreter to $root/opt/scylladb/libexec, but real installation (when rpm -i runs) copies the file to /opt/scylladb/libexec, so anything that embeds a full path needs to strip $root away from it. As a side effect you can't run until this copy is performed.

If you want a local installtion, use $root=/.

install.sh
avikivity: Yes, a two-step installation with $root being a stand-in for the final /.

install.sh
avikivity: Works for me:

```sh
$ x=aa
$ echo ${x-bb}
aa
$ echo ${y-bb}
bb
```

The bash manual has this hidden piece of extra info:

```
   When not performing substring expansion, using the form described
below (e.g., ':-'), Bash tests for a parameter that is unset or null.
Omitting the colon results in a test only for a parameter that is unset.
Put another way, if the colon is included, the operator tests for both
PARAMETER's existence and that its value is not null; if the colon is
omitted, the operator tests only for existence.
```

I used the colon-less form so that if the user explicitly set the variable to null, it is not changed. Only if the variable does not exist I want it to be set.

bhalevy: I wasn't aware of that feature... thanks

avikivity: Well, I had to search the manual for it, because I never remember all the variations.

avikivity:
No progress, closing. Please file a new pull request if you update this.

Issue #4960, opened 2019-09-04
cql_cast_test failures casting to smallint


bhalevy:
scylla version 6c4ad932969749ee132c3a13b3c79328a1c575b5
scylladb/scylla-dtest@2846bff39db09d2cecff0e8e76d1745696484887
Couple errors seen in [dtest-release/231](https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/231):
[cast_decimal_test](https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/231/testReport/cql_cast_test/CQLCastTest/cast_decimal_test/):
```
casting from type decimal to type smallint
```

[cast_varint_test](https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/231/testReport/cql_cast_test/CQLCastTest/cast_varint_test/):
```
casting from type varint to type smallint
```

Relevant debug printouts in [dtest.log](https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/231/artifact/logs-release.2/dtest.log):
```
2019-09-02 06:01:06,928 12426   dtest                          DEBUG    | cql_cast_test.CQLCastTest.cast_decimal_testcast_decimal_test - Test case: CAST(decimal as smallint). Run: casted value "76571.345" and query use ks; SELECT cast(avg(decimal_clmn) as smallint) FROM cast_decimal_test

2019-09-02 06:03:11,743 17924   dtest                          DEBUG    | cql_cast_test.CQLCastTest.cast_varint_testcast_varint_test - Test case: CAST(varint as smallint). Run: casted value "32767456456456456456545678943512357658768763546575675" and query use ks; SELECT cast(avg(varint_clmn) as smallint) FROM cast_varint_test
```

FWIW, I wasn't able to reproduce the errors when running locally.
I added more information to the assert in scylladb/scylla-dtest#1097 in case it is hit again.

PR #4967, opened 2019-09-04
types: Fix decimal to integer and varint to integer conversion

espindola:
The release notes for boost 1.67.0 includes:

Breaking Change: When converting a multiprecision integer to a narrower type, if the value is too large (or negative) to fit in the smaller type, then the result is either the maximum (or minimum) value of the target 

Since we just moved out of boost 1.66, we have to update our code.

This fixes issue #4960

### DIFF
--- a/types.cc
+++ b/types.cc
@@ -3544,12 +3544,11 @@ std::function<data_value(data_value)> make_castas_fctn_from_decimal_to_float() {
     };
 }
 
-template<typename ToType>
-std::function<data_value(data_value)> make_castas_fctn_from_decimal_to_integer() {
+std::function<data_value(data_value)> make_castas_fctn_from_decimal_to_varint() {
     return [](data_value from) -> data_value {
         auto val_from = value_cast<big_decimal>(from);
         boost::multiprecision::cpp_int ten(10);
-        return static_cast<ToType>(val_from.unscaled_value() / boost::multiprecision::pow(ten, val_from.scale()));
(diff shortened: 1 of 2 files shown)


denesb:
Please provide a subject and description. "fix issue #4960" should be a note after the subject and a description that describes what is being fixed and how, not the entire subject and description itself.

types.cc
```diff
@@ -3550,6 +3550,16 @@ static boost::multiprecision::cpp_int from_decimal_to_cppint(const data_value& f
     return val_from.unscaled_value() / boost::multiprecision::pow(ten, val_from.scale());
 }
 
+template<typename ToType>
+std::function<data_value(data_value)> make_castas_fctn_from_decimal_to_integer() {
+    return [](data_value from) -> data_value {
+        auto varint = from_decimal_to_cppint(from);
+        bool negative = varint < 0;
+        uint64_t v = negative ? static_cast<uint64_t>(-varint) : static_cast<uint64_t>(varint);
+        return static_cast<ToType>(negative ? -v : v);
```
avikivity: We should check for overflow here, no?

avikivity: Well, Cassandra doesn't:

```
cqlsh:ks1> select * from bigdec;

 k
-----------------
 100000000000000

(1 rows)
cqlsh:ks1> select cast(k as int) from bigdec;

 cast(k as int)
----------------
      276447232

(1 rows)
```

avikivity: Please add a comment (as a follow-up) that the overflow is intentional.

Issue #4943, opened 2019-09-02
longevity 5000 tables in on keyspace, streaming failed (Cannot assign requested address)


fruch:
This is Scylla's bug tracker, to be used for reporting bugs only.
If you have a question about Scylla, and not a bug, please ask it in
our mailing-list at scylladb-dev@googlegroups.com or in our slack channel.

- [x] I have read the disclaimer above, and I am reporting a suspected malfunction in Scylla.

## Installation details
Scylla version (or git commit hash): 3.1.0.rc4-0.20190830.d70c2db09
Cluster size: 2
OS (RHEL/CentOS/Ubuntu/AWS AMI):  ami-04b9560705a3975d1

## logs
gs://scratch.scylladb.com/ifruchte/5000ks_stream_error_node1.log
gs://scratch.scylladb.com/ifruchte/5000ks_stream_error_node2.log

## summery

after creating 5000 tables in one keyspace on node1, 
node2 two was added to the cluster and started to stream from node1.

node failed on stream in keyspace=feeds
**node1**:
```
2019-09-02T13:43:53+00:00  ip-10-0-136-244 !WARNING | scylla: [shard 0] stream_session - [Stream #aed42ad0-cd87-11e9-8cb8-000000000005] stream_transfer_task: Fail to send to 10.0.68.176:0: std::system_error (error system:99, connect: Cannot assign requested address)
2019-09-02T13:43:53+00:00  ip-10-0-136-244 !WARNING | scylla: [shard 0] stream_session - [Stream #aed42ad0-cd87-11e9-8cb8-000000000005] Failed to send: std::system_error (error system:99, connect: Cannot assign requested address)
2019-09-02T13:43:53+00:00  ip-10-0-136-244 !WARNING | scylla: [shard 0] stream_session - [Stream #aed42ad0-cd87-11e9-8cb8-000000000005] Streaming error occurred, peer=10.0.68.176
2019-09-02T13:43:53+00:00  ip-10-0-136-244 !WARNING | scylla: [shard 0] stream_session - [Stream #aed42ad0-cd87-11e9-8cb8-000000000005] Streaming plan for Bootstrap-feeds-index-0 failed, peers={10.0.68.176}, tx=0 KiB, 0.00 KiB/s, rx=0 KiB, 0.00 KiB/s
````
**node2**:
```
[...]

PR #4968, opened 2019-09-05
streaming:  Do not open rpc stream connection if reader has no data

asias:
We can use the reader::peek() to check if the reader contains any data.
If not, do not open the rpc stream connection. It helps to reduce the
port usage.

Refs: #4943

### DIFF
--- a/streaming/stream_transfer_task.cc
+++ b/streaming/stream_transfer_task.cc
@@ -175,6 +175,13 @@ future<> send_mutations(lw_shared_ptr<send_info> si) {
 }
 
 future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
+ return si->reader.peek(db::no_timeout).then([si] (mutation_fragment* mfp) {
+  if (!mfp) {
+    // The reader contains no data
+    sslog.info("[Stream #{}] Skip sending ks={}, cf={}, reader contains no data, with new rpc streaming",
+        si->plan_id, si->cf.schema()->ks_name(), si->cf.schema()->cf_name());
+    return make_ready_future<>();
+  }
   return si->estimate_partitions().then([si] (size_t estimated_partitions) {
     sslog.info("[Stream #{}] Start sending ks={}, cf={}, estimated_partitions={}, with new rpc streaming", si->plan_id, si->cf.schema()->ks_name(), si->cf.schema()->cf_name(), estimated_partitions);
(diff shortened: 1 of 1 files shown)


streaming/stream_transfer_task.cc
```diff
@@ -175,6 +175,13 @@ future<> send_mutations(lw_shared_ptr<send_info> si) {
 }
 
 future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
+ return si->reader.peek(db::no_timeout).then([si] (mutation_fragment* mfp) {
+  if (!mfp) {
```
denesb: This looks like only half-indent.

asias: It is on purpose to avoid massive indentation changes.

asias: @denesb Please take a look at the peek() usage. Looks good to you?

denesb: > It is on purpose to avoid massive indentation changes.

Oh, I didn't see the closing parens below. Makes sense.

denesb: > @denesb Please take a look at the peek() usage. Looks good to you?

Yes.

tgrabiec:
I wonder how useful this optimization is. For large-enough tables, ranges may never be empty, right? So we need another solution for this problem anyway.

tgrabiec:
Closed by mistake.

asias:
> I wonder how useful this optimization is. For large-enough tables, ranges may never be empty, right? So we need another solution for this problem anyway.

It is really depends on the workload. If there are thousands of sparse tables, this could be useful. 

The optimization is easy and straightforward. I do not see a reason why we should not apply such optimization. It saves some work on both the sender and receiver.

asias:
> So we need another solution for this problem anyway.

Even if we have solution in the rpc framework, e.g., reuse the tcp connection. This is still useful because it avoid unnecessary rpc streaming connection and handler creation.

asias:
Ping.  I know we recently added the support to reuse the port at rpc level. This patch still has its value.

gleb-cloudius:
On Mon, Oct 07, 2019 at 08:08:38PM -0700, Asias He wrote:
> Ping.  I know we recently added the support to reuse the port at rpc level. This patch still has its value. 
> 
I thought it was applied. It is a good chance the rpc patch would not be
needed it it was.

--
			Gleb.

PR #4997, opened 2019-09-10
CDC: Implement minimal version that logs only primary key of each change

### LINKED ISSUE #4986: Implement minimal CDC version that logs only primary key for each change
The first version of Change Data Capture should be minimal but provide some end-to-end functionality. It should make it possible to create a table with CDC enabled as well as alter an existing table to enable/disable CDC for it. It also should log the primary key of each change applied to the data.

This is a subtask of #4985.

haaawk:
This PR introduces CDC in it's minimal version.

It is possible now to create a table with CDC enabled or to enable/disable CDC on existing table.
There is a management of CDC log and description related to enabling/disabling CDC for a table.

For now only primary key of the changed data is logged.

To be able to co-locate cdc streams with related base table partitions it was needed to propagate the information about the number of shards per node. This was node through gossip.

There is an assumption that all the nodes use the same value for sharding_ignore_msb_bits. If it does not hold we would have to gossip sharding_ignore_msb_bits around together with the number of shards.

Fixes #4986.

Tests: unit(dev, release, debug)


haaawk:
Here's recommended order of reading: https://github.com/haaawk/scylla/commits/cdc1-logging-keys

cdc/cdc.cc
```diff
@@ -0,0 +1,197 @@
...
+    for_each_endpoint([&] (const gms::inet_address& endpoint) {
+        const auto decomposed_ip = inet_addr_type->decompose(endpoint.addr());
+
+        for_each_shard(endpoint, [&] (unsigned int shard_id, int shard_count) {
+            const auto pk = partition_key::from_exploded(
+                    *desc_schema, { decomposed_ip, int32_type->decompose(static_cast<int>(shard_id)) });
+            mutations.emplace_back(desc_schema, pk);
+
+            auto stream_id = utils::make_random_uuid();
+            while (!belongs_to(endpoint, shard_id, shard_count, stream_id)) {
```
haaawk: After brainstorming with @kbr- we came to the conclusion that this can be done better by generating stream_ids for all the endpoints and shards at the same time instead of generating it one by one. Will improve and upload soon.

slivne:
@kbr- / @elcallio please review prior to Avi ...

cql3/statements/modification_statement.cc
```diff
@@ -416,7 +421,16 @@ modification_statement::execute_without_condition(service::storage_proxy& proxy,
...
 
-        return proxy.mutate_with_triggers(std::move(mutations), cl, timeout, false, qs.get_trace_state(), qs.get_permit(), this->is_raw_counter_shard_write());
+        auto mutate = [&, cl, timeout] (std::vector<mutation> mutations) {
+            return proxy.mutate_with_triggers(std::move(mutations), cl, timeout, false, qs.get_trace_state(), qs.get_permit(), this->is_raw_counter_shard_write());
+        };
+
+        if (qs.get_client_state().is_internal() || qs.get_client_state().is_thrift() || !s->cdc_enabled()) {
+            return mutate(std::move(mutations));
+        }
+        return cdc::apply(proxy, s, timeout, qs.get_permit(), std::move(mutations)).then([cl, timeout, &proxy, &qs, mutate = std::move(mutate)] (std::vector<mutation> mutations) {
```
elcallio: Given that schema is reachable from each mutation, we could just as well move the cdc call into storage proxy (and generalize it to extract the schema set from mutation set -> batch support (?)). 
Not sure how much it adds in simplicity for alternator etc.

cdc/cdc.cc
```diff
@@ -194,4 +195,194 @@ future<> setup(service::storage_proxy& proxy, schema_ptr s) {
...
+};
+
+static future<transformer> get_transformer(
+        service::storage_proxy& proxy,
+        schema_ptr s,
+        service::storage_proxy::clock_type::time_point timeout,
+        service_permit permit) {
+    auto desc_schema =
+        proxy.get_db().local().find_schema(s->ks_name(), desc_name(s->cf_name()));
+    query::read_command cmd(
```
haaawk: It might be a good idea to cache this somehow.

cql3/statements/create_table_statement.cc
```diff
@@ -97,24 +97,26 @@ std::vector<column_definition> create_table_statement::get_columns()
 }
 
 future<shared_ptr<cql_transport::event::schema_change>> create_table_statement::announce_migration(service::storage_proxy& proxy, bool is_local_only) {
-    return make_ready_future<>().then([this, is_local_only, &proxy] {
```
kbr-: Why did you unwrap this? It changes the semantics.
Previously, if announce_new_column_family or get_cf_meta_data threw an exception, it would result in an exceptional future, which would be inspected later in the lambda passed to then_wrapped.
Now it will always be propagated out of announce_migration (and not as an exceptional future; it will be thrown).

cdc/cdc.cc
```diff
@@ -0,0 +1,197 @@
...
+#include "service/migration_manager.hh"
+#include "service/storage_proxy.hh"
+#include "service/storage_service.hh"
+
+using seastar::sstring;
+
+namespace cdc {
+
+static sstring log_name(const sstring& table_name) {
+    static constexpr const auto cdc_log_suffix = "_scylla_cdc_log";
```
kbr-: `const` not needed :)

cdc/cdc.cc
```diff
@@ -0,0 +1,197 @@
...
+    return table_name + cdc_log_suffix;
+}
+
+static sstring desc_name(const sstring& table_name) {
+    static constexpr const auto cdc_desc_suffix = "_scylla_cdc_desc";
+    return table_name + cdc_desc_suffix;
+}
+
+static future<> remove_log(const sstring& ks_name, const sstring& table_name) {
+    auto& mm = service::get_local_migration_manager();
```
kbr-: why do you introduce these intermediate variables (`mm`)? They make my brain use more power.

kbr-: By the way, since we don't like globals, wouldn't it be possible to pass the migration manager and storage_service down from the caller of CDC functions?
From what I know we're in the middle of a war with globals, but this patchset introduced a lot of global variable references.

cdc/cdc.cc
```diff
@@ -0,0 +1,197 @@
...
+    b.with_column("shard_id", int32_type, column_kind::partition_key);
+    b.with_column("created_at", timestamp_type, column_kind::clustering_key);
+    b.with_column("stream_id", uuid_type);
+    auto& mm = service::get_local_migration_manager();
+    return mm.announce_new_column_family(b.build(), false);
+}
+
+template <typename Func>
+static void for_each_endpoint(Func&& f) {
+    auto& topology = service::get_local_storage_service().get_token_metadata().get_topology();
```
kbr-: globals

cdc/cdc.cc
haaawk: Intermediate variables are just to make the formatting better. It helps in making long lines shorter.

Of course, we can pass them down and use globals on the call points :). I will check how much more messy the code will be if I pass down all the globals instead of using them directly.

kbr-: > Intermediate variables are just to make the formatting better. It helps in making long lines shorter.

Most of the lines wouldn't be so long anyway (yeah I know, you want to have them <= 80 characters, which is insane :P). You can always do
```
    return service::get_local_migration_manager()
            .announce_column_family_drop(ks_name, desc_name(table_name), false);
```

which is much better for readability IMO.

cdc/cdc.cc
haaawk: Not sure. I believe `constexpr auto cdc_log_suffix` will be equivalent to `constexpr char* const cdc_log_suffix` here while `constexpr const auto cdc_log_suffix` is `constexpr const char* const cdc_log_suffix`. So `constexpr` makes a pointer const but `constexpr const` makes both the pointer and the element const. I may be wrong though.

cdc/cdc.cc
kbr-: > Of course, we can pass them down and use globals on the call points :)

Preferably don't use globals at all. For example `storage_service` is passed down all the way from main in a lot of places, similarly to `storage_proxy`. I don't know about `migration_manager` or `snitch_ptr` though.

cdc/cdc.cc
kbr-: `const auto x = y;` is equivalent to `auto const x = y;`. If `auto` deduces to `char*` say, then it will become `char* const x = y;`. It will make the pointer const.

What you describe would be achieved with `constexpr const auto* x = y;`. Then when `y` is of type `char*`, auto would deduce to `char`.

cdc/cdc.cc
```diff
@@ -0,0 +1,197 @@
...
+static void for_each_shard(const gms::inet_address& endpoint, Func&& f) {
+    auto& snitch_ptr = locator::i_endpoint_snitch::get_local_snitch_ptr();
+    const unsigned int shard_count = snitch_ptr->get_shard_count(endpoint);
+    for (unsigned int shard_id = 0; shard_id < shard_count; ++shard_id) {
+        f(shard_id, shard_count);
+    }
+}
+
+static future<>
+populate_desc(service::storage_proxy& proxy, const schema& s) {
```
kbr-: This function makes some important assumptions about the state of DB when it's called, for example that `setup_desc` with the same schema `s` was previously called (and no `setup_desc`s with different schemas after that). It would make it easier to understand the function if these assumptions were written in a comment.

Understanding these assumptions will become much more important when we'll implement schema/topology changes.

cdc/cdc.cc
haaawk: Fair enough. It seems with auto the const is not necessary, indeed. Without auto it would make a difference:
```
    constexpr const auto a = "_scylla_cdc_log";
    constexpr auto b = "_scylla_cdc_log";
    constexpr const char* c = "_scylla_cdc_log";
    constexpr char* d = "_scylla_cdc_log";
    std::cerr << demangle(typeid(a).name()) << std::endl;
    std::cerr << demangle(typeid(b).name()) << std::endl;
    std::cerr << demangle(typeid(c).name()) << std::endl;
    std::cerr << demangle(typeid(d).name()) << std::endl;
```

```
char const*
char const*
char const*
char*
```

cdc/cdc.cc
haaawk: > Preferably don't use globals at all. For example `storage_service` is passed down all the way from main in a lot of places, similarly to `storage_proxy`. I don't know about `migration_manager` or `snitch_ptr` though.

Yes. That's why I said I will see what can be done. I agree with the principles.

> which is much better for readability IMO.

Since it's a matter of taste - let's agree to disagree here :).

cdc/cdc.cc
kbr-: "With `auto`" it would also make a difference, but only if you used `auto*`. I.e.
`constexpr auto* e = "_scylla_cdc_log";` would make `e` have type `char*`.

cdc/cdc.cc
kbr-: > Since it's a matter of taste - let's agree to disagree here :).

But I've just shown you how to achieve short lines AND not have intermediate variables. Isn't that a win-win situation? :P

cdc/cdc.hh
```diff
@@ -0,0 +1,48 @@
...
+
+namespace service {
+
+class storage_proxy;
+
+} // namespace service
+
+namespace cdc {
+
+seastar::future<> setup(service::storage_proxy& proxy, schema_ptr schema);
```
kbr-: Please write a doc string describing what is being "setup".

cdc/cdc.hh
```diff
@@ -0,0 +1,48 @@
...
+
+seastar::future<>
+remove(const seastar::sstring& ks_name, const seastar::sstring& table_name);
```
kbr-: Please write a doc string describing what is being removed.

cdc/cdc.cc
haaawk: > But I've just shown you how to achieve short lines AND not have intermediate variables. Isn't that a win-win situation? :P

I still find the intermediate variable more readable. Both solutions need two lines. Solution with no Intermediate variable requires a break that is IMO way worse than a variable just above the usage. To be honest I totally don't understand the problem with the variable so I don't think this discussion brings much value.

cdc/cdc.cc
```diff
@@ -0,0 +1,197 @@
...
+    return proxy.mutate(std::move(mutations),
+                        db::consistency_level::QUORUM,
+                        db::no_timeout,
+                        nullptr,
+                        empty_service_permit());
+}
+
+future<> setup(service::storage_proxy& proxy, schema_ptr s) {
+    return setup_log(*s).then([&proxy, s = std::move(s)] () mutable {
+        return setup_desc(*s).then_wrapped([&proxy, s = std::move(s)] (future<> f) mutable {
```
kbr-: `announce_new_column_family` might throw.

If that happens inside `setup_desc`, it will fly out of `setup_desc` and make `setup` return an exceptional future. But whatever happened in `setup_log` won't be reversed.
You might want to wrap `setup_desc` in a `make_ready_future<>().then([] {... setup_desc ...})` construct.

If that happens inside `setup_log`, `setup` will propagate the exception too. Perhaps we want to return an exceptional future instead.

cdc/cdc.cc
kbr-: In that particular place (two line function) it probably doesn't make a difference. But when the function is longer, any variable introduced early in the function makes me wonder "is that variable used anywhere else?" which consumes CPU cycles in the human brain which would be better spent elsewhere.
It can sometimes be a source of bugs too, when a variable is used by accident where it shouldn't have been (that's why I really don't like the common pattern where a variable is declared as a result of a function call and then immediately moved from; it would be better to call the function right where it's needed)

cql3/statements/create_table_statement.cc
```diff
@@ -96,10 +97,31 @@ std::vector<column_definition> create_table_statement::get_columns()
...
+    });
+}
+
 future<shared_ptr<cql_transport::event::schema_change>> create_table_statement::announce_migration(service::storage_proxy& proxy, bool is_local_only) {
-    auto create_table = [this, is_local_only, &proxy] {
+    auto schema = get_cf_meta_data(proxy.get_db().local());
+    auto create_table = [this, is_local_only, schema] () mutable {
         auto& mm = service::get_local_migration_manager();
-        return mm.announce_new_column_family(get_cf_meta_data(proxy.get_db().local()), is_local_only).then_wrapped([this] (auto&& f) {
+        auto res = mm.announce_new_column_family(std::move(schema), is_local_only);
```
kbr-: :(

cql3/statements/create_table_statement.cc
```diff
@@ -96,10 +97,31 @@ std::vector<column_definition> create_table_statement::get_columns()
...
+                return make_exception_future<ret_type>(std::move(ep));
+            });
+        });
+    });
+}
+
 future<shared_ptr<cql_transport::event::schema_change>> create_table_statement::announce_migration(service::storage_proxy& proxy, bool is_local_only) {
-    auto create_table = [this, is_local_only, &proxy] {
+    auto schema = get_cf_meta_data(proxy.get_db().local());
+    auto create_table = [this, is_local_only, schema] () mutable {
```
kbr-: Why not make it a private method or a static function in this unit? What do we gain with all the lambda begin passed around? `create_table_with_cdc` can just call the method/function.

cdc/cdc.cc
haaawk: Per your other request all those will become variables passed to the functions so I really start to worry about your brain cpu :)

cql3/statements/create_table_statement.cc
```diff
@@ -96,10 +97,31 @@ std::vector<column_definition> create_table_statement::get_columns()
...
+future<shared_ptr<cql_transport::event::schema_change>>
+create_table_statement::create_table_with_cdc(service::storage_proxy& proxy,
+                                              schema_ptr schema,
+                                              CreateTable&& create_table) {
+    if (_if_not_exists) {
+        throw exceptions::invalid_request_exception(
+                "Can't create table with CDC support using IF NOT EXISTS");
+    }
+    return cdc::setup(proxy, schema).then([create_table = std::move(create_table), schema = std::move(schema)] () mutable{
+        return create_table().handle_exception([schema = std::move(schema)](std::exception_ptr ep) mutable {
```
kbr-: This confirms my previous suspicion. We really don't want `create_table` to throw and instead return an exceptional future, but it might throw as it currently is (from the `announce_new_column_family` call).

cdc/cdc.cc
kbr-: Yay!

cql3/statements/create_table_statement.cc
haaawk: Private method and static function work just as well. The benefit of having lambda here is encapsulation. With lambda it's blunt clear that it's just a helper method for announce_migration and nothing else. Having said that, I'm fine with private method. I can change it if you want?

elcallio: Please don't. Always keep all used entities as local as possible. If the lambda function was needed in a different function also, make it a private function. Otherwise, keep it as close to the usage as possible.

kbr-: Is there any other place in the code where we do something like that: there's a generic function (`create_table_with_cdc`) that takes a function parameter (`CreateTable&& create_table`) and is called in **exactly one** place (`announce_migration`) with **exactly one** function (the `create_table` lambda)?

It just seems like a very non-standard thing.

Also, since `create_table_with_cdc` takes a function parameter, it suggests that it's called in multiple places with different functions passed in for this parameter. And that it doesn't really care what the function parameter is.
However for `create_table_with_cdc` to be correct there is only one function that can be passed in here, which is now the `create_table` lambda. So the parametrization is artificial.

Locality can be easily achieved my using a static function. It will then be local to the compilation unit. You can easily find all usages of the function in the unit by using text search.

elcallio: Less local than inside the function. And requires search/parse of rest of compilation unit/class.

cql3/statements/create_table_statement.cc
haaawk: You're right. Will add make_ready_future<> back. Good catch!

cql3/statements/drop_table_statement.cc
```diff
@@ -76,18 +80,24 @@ future<shared_ptr<cql_transport::event::schema_change>> drop_table_statement::an
...
         try {
             f.get();
             using namespace cql_transport;
-            return make_shared<event::schema_change>(
-                    event::schema_change::change_type::DROPPED,
-                    event::schema_change::target_type::TABLE,
-                    this->keyspace(),
-                    this->column_family());
+            auto fut = make_ready_future<>();
+            if (proxy.get_db().local().has_schema(keyspace(), cdc::desc_name(column_family()))) {
```
kbr-: Could it happen that `proxy.get_db().local().has_schema(keyspace(), cdc::desc_name(column_family()))` returns false on our shard/node, but returns true on a different shard/node, because the information about the cdc table didn't yet arrive to our shard?

Wouldn't unconditional cdc::remove be better? Announce the remove to each node/shard, and simply make it noop if the table doesn't exist locally.

tests/cql_test_env.cc
```diff
@@ -216,6 +216,12 @@ class single_node_cql_env : public cql_test_env {
         return make_ready_future<>();
     }
 
+    virtual future<> require_table_dont_exists(const sstring& ks_name, const sstring& table_name) override {
```
kbr-: doesnt_exist

cdc/cdc.cc
```diff
@@ -194,4 +195,194 @@ future<> setup(service::storage_proxy& proxy, schema_ptr s) {
...
+public:
+    struct map_cmp {
+        bool operator()(const std::pair<net::inet_address, unsigned int>& a,
+                        const std::pair<net::inet_address, unsigned int>& b) const {
+            if (a.first.is_ipv4()) {
+                if (b.first.is_ipv4()) {
+                    if (a.first.as_ipv4_address().ip > b.first.as_ipv4_address().ip) {
+                        return false;
+                    }
+                    return a.second < b.second;
```
kbr-: Wasn't that intended to be a lexicographical compare?
In that case, if `a.first.as_ipv4_address().ip < b.first.as_ipv4_address().ip` and `a.second == b.second`, shouldn't it return true?

cdc/cdc.cc
```diff
@@ -194,4 +195,194 @@ future<> setup(service::storage_proxy& proxy, schema_ptr s) {
...
+        }
+    };
+    using streams_type = std::map<std::pair<net::inet_address, unsigned int>, utils::UUID, map_cmp>;
+private:
+    schema_ptr _schema;
+    schema_ptr _log_schema;
+    utils::UUID _time;
+    bytes _decomposed_time;
+    std::reference_wrapper<locator::token_metadata> _token_metadata;
+    std::reference_wrapper<const locator::abstract_replication_strategy> _replication_strategy;
```
kbr-: why the need for `reference_wrapper`s?

cdc/cdc.cc
```diff
@@ -194,4 +195,194 @@ future<> setup(service::storage_proxy& proxy, schema_ptr s) {
...
+    // The assumption here is that all nodes in the cluster are using the
+    // same murmur3_partitioner_ignore_msb_bits. If it's not true, we will
+    // have to gossip it around together with shard_count.
+    unsigned _ignore_msb_bits;
+    streams_type _streams;
+
+    clustering_key set_pk_columns(const partition_key& pk, int batch_no, mutation& m) const {
```
kbr-: why return the `clustering_key`? The return value is ignored below.

cdc/cdc.cc
```diff
@@ -194,4 +195,194 @@ future<> setup(service::storage_proxy& proxy, schema_ptr s) {
...
+            auto cdef = m.schema()->get_column_definition(to_bytes("_" + column.name()));
+            auto value = atomic_cell::make_live(*column.type,
+                                                _time.timestamp(),
+                                                bytes_view(pk_value[pos]));
+            m.set_cell(log_ck, *cdef, std::move(value));
+            ++pos;
+        }
+        return log_ck;
+    }
+    partition_key stream_id(const schema& s, const net::inet_address& ip, unsigned int shard_id) const {
```
kbr-: suggestion: don't take the `schema` argument, since it's always equal to `*_log_schema`.

cdc/cdc.cc
```diff
@@ -194,4 +195,194 @@ future<> setup(service::storage_proxy& proxy, schema_ptr s) {
...
+                    }
+                    return a.second < b.second;
+                }
+            }
+        }
+    };
+    using streams_type = std::map<std::pair<net::inet_address, unsigned int>, utils::UUID, map_cmp>;
+private:
+    schema_ptr _schema;
+    schema_ptr _log_schema;
```
kbr-: suggestion, but probably not worth the effort: `transformer` doesn't need to live across different tasks. Well yeah, currently `get_transformer` returns a `future<transformer>`, but the `transformer` is only constructed in the last return statement, so its futurization is artificial. The meat of `get_transformer` is building the streams. So:
`get_transformer` could be changed to `get_streams`, and then `transformer` could store only `const schema&` instead of `schema_ptr` (it doesn't need to manage the `schema`'s lifetime).

cdc/cdc.cc
```diff
@@ -194,4 +195,194 @@ future<> setup(service::storage_proxy& proxy, schema_ptr s) {
...
+        auto shard_id = dht::murmur3_partitioner::shard_of(
+                t, locator::i_endpoint_snitch::get_local_snitch_ptr()->get_shard_count(eps[0]), _ignore_msb_bits);
+        mutation res(_log_schema, stream_id(*_log_schema, eps[0].addr(), shard_id));
+        set_pk_columns(m.key(), 0, res);
+        return res;
+    }
+};
+
+class streams_builder {
+    schema_ptr _schema;
```
kbr-: suggestion: this can be a `const schema&`, since `streams_builder` doesn't have to manage the `schema`'s lifetime: it's only a temporary object that doesn't live across different tasks.

The same applies to `transformer`, but changing that would require a bit more effort.

cdc/cdc.cc
```diff
@@ -194,4 +195,194 @@ future<> setup(service::storage_proxy& proxy, schema_ptr s) {
...
+    { }
+
+    mutation transform(const mutation& m) const {
+        auto& t = m.token();
+        auto&& eps = _replication_strategy.get().calculate_natural_endpoints(t, _token_metadata.get());
+        if (eps.empty()) {
+            throw std::runtime_error(format("No primary replica found for key {}", m.decorated_key()));
+        }
+        auto shard_id = dht::murmur3_partitioner::shard_of(
+                t, locator::i_endpoint_snitch::get_local_snitch_ptr()->get_shard_count(eps[0]), _ignore_msb_bits);
```
kbr-: globals

cdc/cdc.cc
```diff
@@ -194,4 +195,194 @@ future<> setup(service::storage_proxy& proxy, schema_ptr s) {
...
+future<std::vector<mutation>> apply(
+        service::storage_proxy& proxy,
+        schema_ptr s,
+        service::storage_proxy::clock_type::time_point timeout,
+        service_permit permit,
+        std::vector<mutation> mutations) {
+
+    return get_transformer(proxy, std::move(s), timeout, permit).then([mutations = std::move(mutations)] (transformer trans) mutable {
+        mutations.reserve(2 * mutations.size());
+        for(int i = 0, size = mutations.size(); i < size; ++i) {
```
kbr-: suggestion: `std::transform` + `std::back_inserter`

cdc/cdc.cc
```diff
@@ -194,4 +195,194 @@ future<> setup(service::storage_proxy& proxy, schema_ptr s) {
...
+    void accept_partition_end(const query::result_row_view& static_row) {
+        _streams.emplace(std::make_pair(_node_ip, _shard_id), _latest_row_stream_id);
+    }
+
+    transformer::streams_type build() {
+        return std::move(_streams);
+    }
+};
+
+static future<transformer> get_transformer(
```
kbr-: suggestion: the meat of this function is building the streams. The futurization of `transformer` is artificial (and forces it to take the schema by `schema_ptr` instead of `const schema&`, by the way). Maybe `get_streams` would be better?

cdc/cdc.hh
```diff
@@ -45,4 +49,11 @@ seastar::sstring log_name(const seastar::sstring& table_name);
 
 seastar::sstring desc_name(const seastar::sstring& table_name);
 
+seastar::future<std::vector<mutation>>apply(
```
kbr-: `apply` is a very generic, non-descriptive name. What about `append_log_mutations`?
In any case, please write a doc string that explain what the function does.

cql3/statements/modification_statement.cc
kbr-: I think what shows up in the log should depend on the frontend used. For example, each frontend has its own notion of "operation type".

cdc/cdc.cc
```diff
@@ -279,7 +279,60 @@ class transformer final {
         auto shard_id = dht::murmur3_partitioner::shard_of(
                 t, locator::i_endpoint_snitch::get_local_snitch_ptr()->get_shard_count(eps[0]), _ignore_msb_bits);
         mutation res(_log_schema, stream_id(*_log_schema, eps[0].addr(), shard_id));
-        set_pk_columns(m.key(), 0, res);
+        auto& p = m.partition();
+        if(p.partition_tombstone()) {
+            // Partition deletion
+            set_pk_columns(m.key(), 0, res);
+        } else if (!p.row_tombstones().empty()) {
```
kbr-: We need to remember that when batch operations are supported, this case and the next will not necessarily be disjoint.

cdc/cdc.cc
kbr-: never mind, I see that it's used in further commit.

cdc/cdc.cc
haaawk: We don't - removed.

cdc/cdc.cc
haaawk: Good catch - will fix. Thanks.

tests/cql_test_env.cc
haaawk: done.

cdc/cdc.cc
haaawk: Inserting invalidates iterators.

cql3/statements/create_table_statement.cc
haaawk: Fixed. Thanks.

cql3/statements/create_table_statement.cc
haaawk: Fixed.

cdc/cdc.cc
haaawk: I don't understand why you say that futurization is artificial? There's a query happening inside.

cdc/cdc.hh
haaawk: done.

cql3/statements/drop_table_statement.cc
haaawk: unconditional remove is probably better. Will change.

haaawk:
I uploaded v2 with the following changes:
1. Added comments
2. Dropped CDC tables unconditionally
3. Few style improvements
4. Fixed exception handling
5. s/apply/append_log_mutations
6. Made CDC code global-free
7. Fixed streams comparator

@kbr- @elcallio @avikivity please have a look.

cdc/cdc.hh
```diff
@@ -0,0 +1,160 @@
...
+///
+/// This function cleans up all CDC related tables created for a given table.
+/// At the moment, CDC Log and CDC Description are the only affected tables.
+/// It's ok if some/all of them don't exist.
+///
+/// \param[in] ctx object with references to database components
+/// \param[in] ks_name keyspace name of a table for which CDC tables are removed
+/// \param[in] table_name name of a table for which CDC tables are removed
+///
+/// \pre This function works correctly no matter CDC Log and/or CDC Describtion
```
kbr-: no matter if

cdc/cdc.cc
```diff
@@ -0,0 +1,471 @@
...
+    static constexpr auto cdc_desc_suffix = "_scylla_cdc_desc";
+    return table_name + cdc_desc_suffix;
+}
+
+static future<>
+remove_log(db_context ctx, const sstring& ks_name, const sstring& table_name) {
+    try {
+        return ctx._migration_manager.announce_column_family_drop(
+                ks_name, log_name(table_name), false);
+    } catch (exceptions::configuration_exception& e) {
```
kbr-: This remove is still conditional, you've just moved the check to a different place (inside `announce_column_family_drop` the node checks if it knows about the schema).

kbr-: Maybe that's not a problem, because we don't support concurrent administrative operations.

kbr-:
LGTM

cdc/cdc.cc
```diff
@@ -0,0 +1,471 @@
...
+                          const int shard_count,
+                          const utils::UUID& stream_id) {
+        const auto log_pk = partition_key::from_singular(*log_schema,
+                                                         data_value(stream_id));
+        const auto token = ctx._partitioner.decorate_key(*log_schema, log_pk).token();
+        auto&& eps =
+            replication_strategy.calculate_natural_endpoints(token, ctx._token_metadata);
+        if (eps.empty() || eps[0] != endpoint) {
+            return false;
+        }
```
kbr-: Unfortunately I found out that this approach is not correct.

If we use NetworkTopologyStrategy and the node we're trying to generate `stream_id` for belongs to a datacenter with RF = 0, the algorithm will never end, because the node will not be a natural endpoint for any token (in other words, no token range is primary for this node; this node is not a replica of the vnodes assigned to it).

What we should be looking for instead is a `stream_id` that belongs to a token range that is _owned_ by this node:
```
ctx._token_metadata.get_endpoint(ctx._token_metadata.first_token(token)) == endpoint
```
Then mutations with this `stream_id` will go exactly to the nodes that are replicas for the token ranges (vnodes) owned by this node. And I think that this is our intention.

haaawk: Will fix. Good catch.

cdc/cdc.cc
```diff
@@ -0,0 +1,471 @@
...
+        , _time(utils::UUID_gen::get_time_UUID())
+        , _decomposed_time(timeuuid_type->decompose(_time))
+        , _replication_strategy(ctx._proxy.get_db().local().find_keyspace(_schema.ks_name()).get_replication_strategy())
+        , _ignore_msb_bits(ctx._proxy.get_db().local().get_config().murmur3_partitioner_ignore_msb_bits())
+        , _streams(streams)
+    { }
+
+    mutation transform(const mutation& m) const {
+        auto& t = m.token();
+        auto&& eps = _replication_strategy.calculate_natural_endpoints(t, _ctx._token_metadata);
```
kbr-: this will have to be fixed too

cdc/cdc.cc
kbr-: Or we could change the definition of the desc table and say that only nodes that are primary replicas (for some token ranges) have corresponding entries in the table, but then:
1. that would probably be harder to implement and introduce more complications when topology changes are handled
2. that would require an additional assumption: the set of replicas for a token range is determined by the primary replica. I think for our current implementations of replication strategies that is true, but I don't know if it's by design or by accident.
And the situation where the owner of a range is not its primary replica is rare, so we shoud optimize for the common case I guess.

haaawk:
I pushed next version with fixed stream_id generation.

nyh:
Hi @haaawk I want to review this patch series, and it would be nice to be able to read a longer description of the implementation you picked and why before diving into the actual patches. Such a description could come as the commit message, but even better would be a design document or an md document in docs/. If you have a document like this, can you please paste a link to it here - or just send it to me directly? Thanks.

haaawk:
Hi @nyh there's a design doc here: https://docs.google.com/document/d/1WI5QaDjTfL-aDKqHywi-ET0ueZhIb5sZxFDdrMafLqU/edit?usp=sharing. It's not perfect but has at least some explanation. I've sent it to the mailing list before.

haaawk:
@nyh, @avikivity do you have any comments? If you want we can schedule a short call to answer all your questions if you have any.

nyh:
@haawk sorry about that. @slivne assigned me to review these patches with 1. no prior knowledge (this is why I went to review your design documents first), and 2. no time (I spent most of last week preparing a webinar, and this week it's a holiday in Israel). I'll be back at work tomorrow and review your patches.

haaawk:
Thanks @nyh. Please let me know if there's anything I can do to make it easier for you. And Happy New Year!

cdc/cdc.hh
nyh: Also "Description".
I think you need to write what you mean by "works correctly" - Does it just do nothing? Throws an exception? Or what?

cql3/statements/create_table_statement.cc
```diff
@@ -96,10 +97,30 @@ std::vector<column_definition> create_table_statement::get_columns()
     return column_defs;
 }
 
+template <typename CreateTable>
+future<shared_ptr<cql_transport::event::schema_change>>
+create_table_statement::create_table_with_cdc(service::storage_proxy& proxy,
```
nyh: I'm curious why this is a separate function instead of just being an option inside the normal create table operation?

cql3/statements/create_table_statement.cc
```diff
@@ -96,10 +97,30 @@ std::vector<column_definition> create_table_statement::get_columns()
...
+        });
+    });
+}
+
 future<shared_ptr<cql_transport::event::schema_change>> create_table_statement::announce_migration(service::storage_proxy& proxy, bool is_local_only) {
-    auto create_table = [this, is_local_only, &proxy] {
-        return make_ready_future<>().then([this, is_local_only, &proxy] {
-            return service::get_local_migration_manager().announce_new_column_family(get_cf_meta_data(proxy.get_db().local()), is_local_only);
+    auto schema = get_cf_meta_data(proxy.get_db().local());
+    auto create_table = [this, is_local_only, schema] () mutable {
```
nyh: I'm having a sense of deja-vu - haven't I seen this lambda in a previous patch?

tests/cdc_test.cc
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <seastar/testing/thread_test_case.hh>
+
+#include "tests/cql_test_env.hh"
+
+SEASTAR_THREAD_TEST_CASE(test_with_cdc_parameter) {
```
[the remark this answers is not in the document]
nyh: Oh, I see that later commits add more checks to this one test... So at the end it tests more than it tests here... Maybe you can say that in the commit message.

tests/cdc_test.cc
```diff
@@ -39,6 +41,19 @@ SEASTAR_THREAD_TEST_CASE(test_with_cdc_parameter) {
         auto test = [&] (const sstring& create_stmt, bool expected_cdc_value) {
             e.execute_cql(create_stmt).get();
             assert_cdc(expected_cdc_value);
+            if (expected_cdc_value) {
```
nyh: All of these checks are deep implementation tests, and may change in later implementation, while the API does't. So I would have made them a separate test - so that if we later change the implementation we can see the API checks continue to succeed, while this implementation test is the only one which will break.

cdc/cdc.cc
```diff
@@ -218,4 +219,195 @@ db_context db_context::builder::build() {
...
+        bool cmp(const std::pair<net::inet_address, unsigned int>& a,
+                 const std::pair<net::inet_address, unsigned int>& b,
+                 Func&& func) const {
+            auto&& a_ip = func(a.first).ip;
+            auto&& b_ip = func(b.first).ip;
+            return a_ip < b_ip || (a_ip == b_ip && a.second < b.second);
+        }
+        bool operator()(const std::pair<net::inet_address, unsigned int>& a,
+                        const std::pair<net::inet_address, unsigned int>& b) const {
+            if (a.first.is_ipv4()) {
```
nyh: This is too clever for me :-) what is this operator() supposed to return - equality? But then, if a is ipv4, and b is NOT ipv4, why do we return true?

cdc/cdc.cc
```diff
@@ -218,4 +219,195 @@ db_context db_context::builder::build() {
...
+        _latest_row_stream_id = utils::UUID();
+    }
+
+    void accept_new_partition(uint32_t row_count) {
+        assert(false);
+    }
+
+    void accept_new_row(
+            const clustering_key& key,
+            const query::result_row_view& static_row,
```
nyh: You don't seem to use the static_row parameter - is this ok?

cql3/statements/modification_statement.cc
```diff
@@ -414,7 +419,22 @@ modification_statement::execute_without_condition(service::storage_proxy& proxy,
...
 
-        return proxy.mutate_with_triggers(std::move(mutations), cl, timeout, false, qs.get_trace_state(), qs.get_permit(), this->is_raw_counter_shard_write());
+        auto mutate = [&, cl, timeout] (std::vector<mutation> mutations) {
+            return proxy.mutate_with_triggers(std::move(mutations), cl, timeout, false, qs.get_trace_state(), qs.get_permit(), this->is_raw_counter_shard_write());
+        };
+
+        if (qs.get_client_state().is_internal() || qs.get_client_state().is_thrift() || !s->cdc_enabled()) {
+            return mutate(std::move(mutations));
+        }
+        return cdc::append_log_mutations(
```
nyh: I don't remember, why does this append_log_mutations (which just adds more mutations to the vector) return a future instead of working immediately?

service/storage_service.cc
```diff
@@ -1679,7 +1679,8 @@ future<> storage_service::gossip_snitch_info() {
     auto rack = snitch->get_rack(addr);
     return _gossiper.add_local_application_state({
         { gms::application_state::DC, value_factory.datacenter(dc) },
-        { gms::application_state::RACK, value_factory.rack(rack) }
+        { gms::application_state::RACK, value_factory.rack(rack) },
+        { gms::application_state::SHARD_COUNT, value_factory.shard_count(smp::count) }
```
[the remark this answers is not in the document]
kbr-: Assume that there's no resharding and that the cluster topology doesn't change at all. Currently a lot of things can break if nodes change. I'm working on a version that supports topology changes.

cql3/statements/modification_statement.cc
kbr-: It reads the CDC description table to create the mutations.
I'm implementing a topology-change-proof version which won't read the description table, but we will still have to read the base table to create the pre-image, so it will eventually have to return a future.

schema.cc
```diff
@@ -584,6 +585,7 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
     os << ",bloomFilterFpChance=" << s._raw._bloom_filter_fp_chance;
     os << ",memtableFlushPeriod=" << s._raw._memtable_flush_period;
     os << ",caching=" << s._raw._caching_options.to_sstring();
+    os << ",cdc=" << std::boolalpha << s._raw._cdc_enabled;
```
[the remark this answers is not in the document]
haaawk: Fixed.

cql3/statements/cf_prop_defs.cc
```diff
@@ -172,11 +174,19 @@ std::optional<utils::UUID> cf_prop_defs::get_id() const {
     return std::nullopt;
 }
 
+bool cf_prop_defs::is_cdc_enabled() const {
```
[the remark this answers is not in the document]
haaawk: In general we would probably need a way to decide what should be included into CDC (change itself, pre-image, post-image). The question is what's the best way to do it. Should we do a complex configuration like in compaction and topology strategy?

service/storage_service.cc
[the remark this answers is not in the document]
haaawk: @kbr- is right. This is a first minimal version that ignores many complexities. @nyh please assume no topology or schema changes can occur. Handling those is the goal of next steps that will build on this patchset. I will add a comment in commit message to make it more clear.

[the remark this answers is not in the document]
haaawk: @nyh I tried to express the fact that number of shards on each node is needed for creation of streams of changes in commit message:
"We would like to share with other nodes
    the number of shards available at the node.
    
    This is needed by CDC to be able to create
    streams of changes for each shard on each node."

I will expand this a bit more.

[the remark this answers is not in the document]
haaawk: I improved the commit message.

cdc/cdc.cc
haaawk: This is not equality. This is used in std::map so it has to return true if a < b.
If a is ipv4 and b is ipv6, the comparator returns true because I ordered all ipv4 addresses before all ipv6 addresses. Does it make sense?

cql3/statements/cf_prop_defs.cc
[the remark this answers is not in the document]
nyh: I haven't thought this through, but I assumed that because this configuration comes from a CQL command, which is a string, it would most likely be a string. I was worried that if we use a bool now, and want to upgrade to a string later, we will need to deal with upgrade issues, and maybe it's better to use a string right now. On the other hand, maybe we should just call CDC an experimental feature, and not promise any update path yet.

cdc/cdc.cc
nyh: Yes, it makes sense, but please add a comment saying this - that this returns true if a < b, as std::map requires, and that you order ipv4 addresses before ipv6 addresses. It's hard to guess and there are many other possibilities on what this "cmp" function does.

tests/cdc_test.cc
haaawk: Which checks do you have in mind when you say that they check only the implementation? Existance of Log and Description is part of API. Content of Description is also part of API.

tests/cdc_test.cc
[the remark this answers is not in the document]
haaawk: I added commit message. Regarding testing persistence of the parameter - I haven't found a way to do it with cql_test_env.

cql3/statements/create_table_statement.cc
```diff
@@ -97,24 +97,27 @@ std::vector<column_definition> create_table_statement::get_columns()
...
-            using namespace cql_transport;
-            return make_shared<event::schema_change>(
-                    event::schema_change::change_type::CREATED,
-                    event::schema_change::target_type::TABLE,
-                    this->keyspace(),
-                    this->column_family());
-        } catch (const exceptions::already_exists_exception& e) {
-            if (_if_not_exists) {
-                return ::shared_ptr<cql_transport::event::schema_change>();
+    auto create_table = [this, is_local_only, &proxy] {
```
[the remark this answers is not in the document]
haaawk: I separated them on purpose. This one is a harmless preparation but touches multiple lines. The next commit is the exact change that carries some risk. I wanted to make it clear what's an easy refactoring and what's a new code that needs careful review. That's what I was always asked to do by maintainers.

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+using locator::topology;
+using seastar::sstring;
+using service::migration_manager;
+using service::storage_proxy;
+
+namespace cdc {
+
+sstring log_name(const sstring& table_name) {
+    static constexpr auto cdc_log_suffix = "_scylla_cdc_log";
+    return table_name + cdc_log_suffix;
```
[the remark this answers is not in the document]
haaawk: Those tables will be queried by uses using CQL so characters that are not allowed probably won't work.

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+    b.with_column("ttl", long_type);
+    auto add_columns = [&] (const auto& columns) {
+        for (const auto& column : columns) {
+            b.with_column("_" + column.name(), column.type);
+        }
+    };
+    add_columns(s.partition_key_columns());
+    add_columns(s.clustering_key_columns());
+    add_columns(s.static_columns());
+    add_columns(s.regular_columns());
```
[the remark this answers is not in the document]
haaawk: The most important reason is that in the future patches (Calle's patchset) static and regular columns are handled differently.

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+
+// This function assumes setup_desc was called on |s| before the call to this
+// function.
+static future<> populate_desc(db_context ctx, const schema& s) {
+    auto& db = ctx._proxy.get_db().local();
+    auto desc_schema =
+        db.find_schema(s.ks_name(), desc_name(s.cf_name()));
+    auto log_schema =
+        db.find_schema(s.ks_name(), log_name(s.cf_name()));
+    // The assumption here is that all nodes in the cluster are using the
```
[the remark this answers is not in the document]
haaawk: I believe this never changes and will break the cluster if someone changes it on a running cluster. I added the comment to get some input from people who may know more on the subject than me. @avikivity What do you think about it?

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+        // It's fine if the table does not exist.
+        return make_ready_future<>();
+    }
+}
+
+future<>
+remove(db_context ctx, const sstring& ks_name, const sstring& table_name) {
+    return when_all(remove_log(ctx, ks_name, table_name),
+                    remove_desc(ctx, ks_name, table_name)).discard_result();
+}
```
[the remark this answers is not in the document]
haaawk: Fixed.

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+                             empty_service_permit());
+}
+
+future<> setup(db_context ctx, schema_ptr s) {
+    return setup_log(ctx, *s).then([ctx, s = std::move(s)] () mutable {
+        return make_ready_future<>().then([ctx, &s = *s] {
+            return setup_desc(ctx, s);
+        }).then_wrapped([ctx, s = std::move(s)] (future<> f) mutable {
+            if (f.failed()) {
+                return remove_log(ctx, s->ks_name(), s->cf_name());
```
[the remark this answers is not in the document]
haaawk: Good point. Fixed.

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+            }
+            return make_ready_future<>().then([ctx, &s = *s] {
+                return populate_desc(ctx, s);
+            }).handle_exception([ctx, s = std::move(s)] (std::exception_ptr ep) mutable {
+                return remove_desc(ctx, s->ks_name(), s->cf_name()).then([ctx, s = std::move(s)] {
```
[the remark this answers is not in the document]
haaawk: Calling remove() sounds like a reasonable thing to do. Done.

cdc/cdc.hh
haaawk: Fixed.

cql3/statements/create_table_statement.cc
haaawk: Exactly. It was prepared in a previous patch. Here I only change &proxy to schema and add mutable to the lambda.

cql3/statements/create_table_statement.cc
haaawk: I wanted to separate this CDC related logic to make it possible for a reader that's not interested in CDC to easily ignore this part.

locator/snitch_base.cc
```diff
@@ -37,8 +37,23 @@
  */
 
 #include "locator/snitch_base.hh"
+#include "gms/gossiper.hh"
+#include "gms/application_state.hh"
 
 namespace locator {
+std::optional<sstring>
+snitch_base::get_endpoint_info(inet_address endpoint,
```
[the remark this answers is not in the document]
haaawk: I guess get_endpoint_info could be just a standalone function. I extracted it from production_snitch_base::get_endpoint_info because I needed exactly the same logic in get_shard_count. I thought that it would be clean to extract a common logic to a base class and then use it both in a base class and subclass. I can make it static if you like. Or turn it into a fully standalone function if you insist. Which one do you prefer?

dht/murmur3_partitioner.hh
```diff
@@ -53,13 +53,14 @@ public:
...
     virtual unsigned sharding_ignore_msb() const override;
+    static unsigned shard_of(const token& t, unsigned shard_count, unsigned sharding_ignore_msb_bits);
 private:
     using uint128_t = unsigned __int128;
     static int64_t normalize(int64_t in);
     token get_token(bytes_view key);
     token get_token(uint64_t value) const;
-    token bias(uint64_t value) const;      // translate from a zero-baed range
-    uint64_t unbias(const token& t) const; // translate to a zero-baed range
+    token bias(uint64_t value) const;      // translate from a zero-based range
```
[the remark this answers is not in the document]
haaawk: I can't. bias uses get_token which is virtual and not static.

dht/murmur3_partitioner.cc
```diff
@@ -28,6 +28,11 @@
 
 namespace dht {
 
+unsigned
```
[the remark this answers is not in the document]
haaawk: If you're worried about confusing shard_of and zero_based_shard_of the I don't think it's a problem. One is public, the other is private so users of partitioner can't mix them. Someone who's changing the internals can easily see that shard_of calls zero_based_shard_of and figure out the relation between them.

cdc/cdc.cc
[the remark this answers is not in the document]
kbr-: Until now nodes didn't care about other nodes' shards. Sharding was a "private matter" to each node. So I don't think the value of ignore_msb_bits has to be in sync.

In my version this is taken care of (I don't require anything to be in sync, but if it is, I use it make things faster).
My suggestion would be to ignore this issue for now, maybe leave a FIXME. I will change most of this code anyway.

db/schema_tables.cc
```diff
@@ -269,7 +269,7 @@ schema_ptr tables() {
          {"dclocal_read_repair_chance", double_type},
          {"default_time_to_live", int32_type},
          {"extensions", map_type_impl::get_instance(utf8_type, bytes_type, false)},
-         {"flags", set_type_impl::get_instance(utf8_type, false)}, // SUPER, COUNTER, DENSE, COMPOUND
+         {"flags", set_type_impl::get_instance(utf8_type, false)}, // SUPER, COUNTER, DENSE, COMPOUND, CDC
          {"gc_grace_seconds", int32_type},
          {"id", uuid_type},
```
avikivity: Better to add a boolean field to scylla_tables as this is not compatible with Cassandra.

cql3/statements/cf_prop_defs.cc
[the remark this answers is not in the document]
avikivity: The Cassandra syntax is `WITH CDC = true`; we can support an alterantive syntax WITH CDC = property_map`.

The property map would indicate that it's enabled, and whether we log just the primary key, changes, pre/post, and retention options.

cql3/statements/create_table_statement.cc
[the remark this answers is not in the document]
avikivity: Yes, and it makes it easier to review, at least for me.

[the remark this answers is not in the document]
avikivity: Even better than a lambda is to put it into a separate method. It makes the containing method smaller.

cdc/cdc.cc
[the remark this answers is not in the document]
avikivity: Bikeshedding, but maybe we should make the suffix `_change_log` to make it clearer.

cdc/cdc.cc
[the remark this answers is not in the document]
avikivity: The assumption is not correct. But if we don't have to make it, it doesn't matter.

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+    auto desc_schema =
+        db.find_schema(s.ks_name(), desc_name(s.cf_name()));
+    auto log_schema =
+        db.find_schema(s.ks_name(), log_name(s.cf_name()));
+    // The assumption here is that all nodes in the cluster are using the
+    // same murmur3_partitioner_ignore_msb_bits. If it's not true, we will
+    // have to gossip it around together with shard_count.
+    auto ignore_msb_bits =
+        db.get_config().murmur3_partitioner_ignore_msb_bits();
+
```
avikivity: should use global_partitioner().sharding_ignore_msb() since that's what the system uses (it derives it from config, but better get it from the source of truth)

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+    };
+    add_columns(s.partition_key_columns());
+    add_columns(s.clustering_key_columns());
+    add_columns(s.static_columns());
+    add_columns(s.regular_columns());
+    return ctx._migration_manager.announce_new_column_family(b.build(), false);
+}
+
+static future<> setup_desc(db_context ctx, const schema& s) {
+    schema_builder b(s.ks_name(), desc_name(s.cf_name()));
```
avikivity: What's a desc?

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+    schema_builder b(s.ks_name(), log_name(s.cf_name()));
+    b.set_default_time_to_live(gc_clock::duration{86400}); // 24h
+    b.set_comment(sprint("CDC log for %s.%s", s.ks_name(), s.cf_name()));
+    b.with_column("stream_id", uuid_type, column_kind::partition_key);
+    b.with_column("time", timeuuid_type, column_kind::clustering_key);
+    b.with_column("batch_seq_no", int32_type, column_kind::clustering_key);
+    b.with_column("operation", int32_type);
+    b.with_column("ttl", long_type);
+    auto add_columns = [&] (const auto& columns) {
+        for (const auto& column : columns) {
```
avikivity: Please avoid gratuituous generic lambdas, because IDEs won't be able to understand the type of `columns` and assist the developer.

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+    b.with_column("node_ip", inet_addr_type, column_kind::partition_key);
+    b.with_column("shard_id", int32_type, column_kind::partition_key);
+    b.with_column("created_at", timestamp_type, column_kind::clustering_key);
+    b.with_column("stream_id", uuid_type);
+    return ctx._migration_manager.announce_new_column_family(b.build(), false);
+}
+
+template <typename Func>
+static void for_each_endpoint(db_context ctx, Func&& f) {
+    for (const auto& dc : ctx._token_metadata.get_topology().get_datacenter_endpoints()) {
```
avikivity: Please use std::function instead of a template. This both helps getting better error messages, and reduces code bloat.

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+        for (const auto& endpoint : dc.second) {
+            f(endpoint);
+        }
+    }
+}
+
+template <typename Func>
+static void
+for_each_shard(db_context ctx, const gms::inet_address& endpoint, Func&& f) {
+    const unsigned int shard_count = ctx._snitch->get_shard_count(endpoint);
```
avikivity: Here too (and in any non-performance-critical path).

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+                stream_id = utils::make_random_uuid();
+            }
+            auto value = atomic_cell::make_live(*uuid_type,
+                                                ts,
+                                                uuid_type->decompose(stream_id));
+            mutations.back().set_cell(ck, *cdef, std::move(value));
+        });
+    });
+    return ctx._proxy.mutate(std::move(mutations),
+                             db::consistency_level::QUORUM,
```
avikivity: It can be nicer to use cql prepared statements instead of generating mutations by hand. It's certainly a more stable API, the mutations stuff can change.

cdc/cdc.cc
```diff
@@ -0,0 +1,221 @@
...
+                             db::no_timeout,
+                             nullptr,
+                             empty_service_permit());
+}
+
+future<> setup(db_context ctx, schema_ptr s) {
+    return setup_log(ctx, *s).then([ctx, s = std::move(s)] () mutable {
```
avikivity: For administrative stuff like this it's better to use a seastar::thread. The code is simpler to understand and there's no performance impact.

dht/murmur3_partitioner.cc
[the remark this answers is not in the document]
avikivity: Is it really needed? You can constuct a temporary murmur3_partitioner and call its regular shard_of() function.

cql3/statements/alter_table_statement.cc
```diff
@@ -388,7 +393,17 @@ future<shared_ptr<cql_transport::event::schema_change>> alter_table_statement::a
...
+        f = f.then([&proxy, schema = std::move(schema)] {
+            return cdc::setup(cdc::db_context::builder(proxy).build(), schema);
+        });
+    } else if (delete_cdc) {
+        f = f.then([&proxy, schema = std::move(schema)] {
+            return cdc::remove(cdc::db_context::builder(proxy).build(), schema->ks_name(), schema->cf_name());
+        });
+    }
+    return f.then([this] {
         using namespace cql_transport;
```
avikivity: This is totally broken if we crash between creating the table and creating the cdc log.

However, there's no easy way to fix it until we get transactional schema, so let's live with it for now.

cdc/cdc.cc
avikivity: The traditional name for such a comparator is less(). But why can't you use inet_type's comparator?

Also (this is a question for where this is used) it's better to use an unordered_map when possible.

cdc/cdc.cc
```diff
@@ -218,4 +219,195 @@ db_context db_context::builder::build() {
...
+private:
+    db_context _ctx;
+    const schema& _schema;
+    schema_ptr _log_schema;
+    utils::UUID _time;
+    bytes _decomposed_time;
+    // The assumption here is that all nodes in the cluster are using the
+    // same murmur3_partitioner_ignore_msb_bits. If it's not true, we will
+    // have to gossip it around together with shard_count.
+    unsigned _ignore_msb_bits;
```
avikivity: We'll need a map here as the assumption is incorrect.

cdc/cdc.cc
```diff
@@ -218,4 +219,195 @@ db_context db_context::builder::build() {
...
+    const streams_type& _streams;
+
+    clustering_key set_pk_columns(const partition_key& pk, int batch_no, mutation& m) const {
+        const auto log_ck = clustering_key::from_exploded(
+                *m.schema(), { _decomposed_time, int32_type->decompose(batch_no) });
+        auto pk_value = pk.explode(_schema);
+        size_t pos = 0;
+        for (const auto& column : _schema.partition_key_columns()) {
+            assert (pos < pk_value.size());
+            auto cdef = m.schema()->get_column_definition(to_bytes("_" + column.name()));
```
avikivity: This is okay for now, but we should precalculate this stuff: build a vector of source table column_id -> log table column_id. The vector changes each time the schema changes, so it's tricky.

cdc/cdc.cc
```diff
@@ -218,4 +219,195 @@ db_context db_context::builder::build() {
...
+            if (a.first.is_ipv4()) {
+                return !b.first.is_ipv4() ||
+                    cmp(a, b, [] (const net::inet_address& addr) { return addr.as_ipv4_address(); });
+            } else {
+                return !b.first.is_ipv4() &&
+                    cmp(a, b, [] (const net::inet_address& addr) { return addr.as_ipv6_address(); });
+            }
+        }
+    };
+    using streams_type = std::map<std::pair<net::inet_address, unsigned int>, utils::UUID, map_cmp>;
```
avikivity: unordered_map

cdc/cdc.cc
```diff
@@ -218,4 +219,195 @@ db_context db_context::builder::build() {
...
+private:
+    db_context _ctx;
+    const schema& _schema;
+    schema_ptr _log_schema;
+    utils::UUID _time;
```
avikivity: How can the same time be valid for all transformations? Shouldn't it come from the mutation itself?

cdc/cdc.cc
```diff
@@ -218,4 +219,195 @@ db_context db_context::builder::build() {
...
+                                                bytes_view(pk_value[pos]));
+            m.set_cell(log_ck, *cdef, std::move(value));
+            ++pos;
+        }
+        return log_ck;
+    }
+    partition_key stream_id(const net::inet_address& ip, unsigned int shard_id) const {
+        auto it = _streams.find(std::make_pair(ip, shard_id));
+        if (it == std::end(_streams)) {
+                throw std::runtime_error(format("No stream found for node {} and shard {}", ip, shard_id));
```
avikivity: This is tricky, it will probably fail during topology changes. We should really move logging to the replica.

kbr-: The whole thing assumes there are no topology changes. And it will most certainly fail.
Please read my "CDC with topology change support" document, which is a modification to the original design. I'm working on the code right now.

cdc/cdc.cc
```diff
@@ -218,4 +219,195 @@ db_context db_context::builder::build() {
...
+            const clustering_key& key,
+            const query::result_row_view& static_row,
+            const query::result_row_view& row) {
+        auto row_iterator = row.iterator();
+        api::timestamp_type timestamp = value_cast<db_clock::time_point>(
+                timestamp_type->deserialize(key.explode(_schema)[0])).time_since_epoch().count();
+        if (timestamp <= _latest_row_timestamp) {
+            return;
+        }
+        _latest_row_timestamp = timestamp;
```
avikivity: For just logging the partition key, this is fine. But for real data logging we'll need to split the mutation according to timestamps. Since it's super-rare for a single mutation to have multiple timestamps (I think it can be constructed with a BATCH statement) we can defer it, but we need to implement it at some point. @denesb can recommend a timestamp-based mutation splitter for you.

avikivity: Looks like I misunderstood this. Please add documentation to streams_builder to explain its role.

cdc/cdc.cc
```diff
@@ -218,4 +219,195 @@ db_context db_context::builder::build() {
...
+        });
+    });
+}
+
+future<std::vector<mutation>> append_log_mutations(
+        db_context ctx,
+        schema_ptr s,
+        service::storage_proxy::clock_type::time_point timeout,
+        service::query_state& qs,
+        std::vector<mutation> muts) {
```
avikivity: Why copy here?

cdc/cdc.cc
```diff
@@ -218,4 +219,195 @@ db_context db_context::builder::build() {
...
+        size_t pos = 0;
+        for (const auto& column : _schema.partition_key_columns()) {
+            assert (pos < pk_value.size());
+            auto cdef = m.schema()->get_column_definition(to_bytes("_" + column.name()));
+            auto value = atomic_cell::make_live(*column.type,
+                                                _time.timestamp(),
+                                                bytes_view(pk_value[pos]));
+            m.set_cell(log_ck, *cdef, std::move(value));
+            ++pos;
+        }
```
avikivity: Note that here working with the mutation API is the right thing to do, the CQL stuff should be reserved for administrative operations.

cdc/cdc.cc
```diff
@@ -298,7 +298,60 @@ class transformer final {
         auto shard_id = dht::murmur3_partitioner::shard_of(
                 t, _ctx._snitch->get_shard_count(*ep), _ignore_msb_bits);
         mutation res(_log_schema, stream_id(ep->addr(), shard_id));
-        set_pk_columns(m.key(), 0, res);
+        auto& p = m.partition();
+        if(p.partition_tombstone()) {
```
avikivity: Space after if.

cdc/cdc.cc
```diff
@@ -298,7 +298,60 @@ class transformer final {
...
+            // Partition deletion
+            set_pk_columns(m.key(), 0, res);
+        } else if (!p.row_tombstones().empty()) {
+            // range deletion
+            int batch_no = 0;
+            for (auto& rt : p.row_tombstones()) {
+                auto set_bound = [&] (const clustering_key& log_ck, const clustering_key_prefix& ckp) {
```
avikivity: Shouldn't we have an indication here of the operation (range delete start/end)? Also whether the bound is inclusive or exclusive.

cdc/cdc.cc
avikivity: You should have a thread::maybe_yield() here (and run this in a thread), or @glommer will be sad

avikivity: Perhaps each shard should be responsible for populating its own entry in the table. This works better with replica-side logging: when a replica is notified it becomes the owner of a token range, it initializes the stream id for it.

tests/cdc_test.cc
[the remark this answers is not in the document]
nyh: We could add to cql_test_env a feature where we can shut down the test environment without removing the data files, and then bring up the test environment again - with the same data files. But this is beyond the scope of this patch. And in any case I think what we really need isn't a more advanced cql_test_env but rather dropping it completely, and replacing it by functional tests in Python, similar to what we did in Alternator (and similar in some respects to what @kostja has also been suggesting). But since it's beyond the scope of this patch set, I'll mark the discussion resolved.

cdc/cdc.cc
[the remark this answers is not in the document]
nyh: @haaawk I see, I thought the CDC table is an implementation detail and the user will have new instructions to read from it - I completely missed the fact that users will need to read from it directly and understand its inner format to do so. Indeed, if users need to read from it, they'll need to know the table's name. But in that case maybe you should add checking to the CDC creation code that the CDC tables do not previously exist - because they may.

locator/snitch_base.cc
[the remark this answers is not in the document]
nyh: I don't know what I prefer, I think I still don't understand what this function is trying to do. If get_endpoint_info() isn't virtual, and there are two implementations, which one will be called from get_shard_count()? Does making get_shard_count() virtual, but with just a single implementation, help?

tests/cdc_test.cc
nyh: Ok, I completely missed that all the details of the CDC table are the actual API of this feature, not just implementation details....

cdc/cdc.cc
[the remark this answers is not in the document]
haaawk: @nyh makes sense. I was planning to solve this by making sure that it's impossible to create a table with _scylla_cdc_log and _scylla_cdc_desc suffix manually. I also planned to do it as a separate task that would be done by one of the new employees. Would that be acceptable?

haaawk:
I was hoping to get this merged in. I'm addressing Avi's comments right now.

avikivity:
Please make it depend on the --experimental flag, so we don't have to commit to support anything and can change it later.

haaawk:
Will do @avikivity

cql3/statements/cf_prop_defs.cc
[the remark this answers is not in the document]
haaawk: Will do the property map.

cdc/cdc.cc
haaawk: description

cdc/cdc.cc
haaawk: Kamil will change this in exactly that way in his patchset. Each shard will be responsible for its streamid.

locator/snitch_base.cc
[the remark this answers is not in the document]
haaawk: snitch_base::get_endpoint_info is not virtual (and I made it static) on purpose. It's a helper method that's used in snitch_base::get_shard_count and in production_snitch_base::get_endpoint_info. production_snitch_base::get_endpoint_info and snitch_base::get_endpoint_info are not related. Then only relation is that production_snitch_base::get_endpoint_info uses snitch_base::get_endpoint_info.

cql3/statements/alter_table_statement.cc
haaawk: I guess I could reverse the order so that I first setup CDC and only after that announce the schema change. It would be a bit safer.

cdc/cdc.cc
haaawk: The operation type is added by Calle in the following patchset. There start and end are easily distinguished.
For inclusive/exclusive, that's something we have to do. The question is how. There are some options. One is to make the operation type RT_start_inclusive, RT_start_exclusive, RT_end_inclusive and RT_end_exclusive. It's probably the simplest. Another would be an addition of an extra column to mark inclusiveness but I like the first option more. What do you think @avikivity? Also FYI @elcallio

cdc/cdc.cc
haaawk: It's not copy. Argument is moved here. See https://github.com/scylladb/scylla/pull/4997/commits/abbe97e07e7e7d7971b2a44d134ffb8bf72558b8#diff-b080f995ee3e937cb0f38d2617833c64R434

cdc/cdc.cc
haaawk: It's ok because CDC description table has no static rows.

cdc/cdc.cc
haaawk: I added a comment.

cdc/cdc.cc
haaawk: It's the time the change reached the cluster.

cdc/cdc.cc
elcallio: inclusive/exclusive markers is symmetric with clustering bounds so makes most sense. Ii.e. agree on first option)

cdc/cdc.cc
haaawk: I changed the map type to unordered_map so this comparator goes away.

cdc/cdc.cc
haaawk: done

dht/murmur3_partitioner.cc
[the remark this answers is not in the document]
haaawk: I guess it should be possible to just create another instance of murmur3_partitioner. Will give it a try.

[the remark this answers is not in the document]
haaawk: It works. I'm removing this patch. Thanks.

cdc/cdc.cc
haaawk: Done.

cdc/cdc.cc
haaawk: I removed those helper at all.

cdc/cdc.cc
haaawk: done.

cdc/cdc.cc
haaawk: done.

haaawk:
I pushed a version with most of your comments addressed. I still have to figure out how to use scylla_tables and use cql prepared statements instead of mutations for description.

haaawk:
I moved the cdc settings to scylla_tables. @nyh I think I addressed all the comments. Please have a look.

locator/snitch_base.cc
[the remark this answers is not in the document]
nyh: Ok. I find this very confusing but as long as you checked that this actually does what you want, I'll believe you :-)

cdc/cdc.cc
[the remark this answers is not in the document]
nyh: @haaawk we can certainly do this later, but please remember this issue somewhere (FIXME, bug tracker, or wherever you save your list of tasks...).
@avikivity are you ok with that long name, "_scylla_cdc_log" (that users will need to be aware of) or do you request that they change it? We can change it later, of course, while this is still an experimental feature.

cdc/cdc.cc
nyh: @avikivity also Alternator uses direct calls to mutate(). If mutate() changes in the future, we'll need to change all callers - including CDC and Alternator, or, eventually, we'll need to create some internal APIs that all of these use - but these internal APIs should not use CQL concepts like "prepared statements".

cdc/cdc.cc
nyh: @haaawk maybe need to add a FIXME here? Or you're just assuming that everything around this area will be rewritten in the near future anyway?

nyh:
@avikivity any objections that I commit this patch set? I see there are still a few small pending conversations with you in the review.
I think much of the details in these patches will need to change later - there are already other patch sets waiting for some (e.g., topology changes) but I am worried that even basic stuff like how "streams" work and table names will change, but I think it may be a good idea to commit this start to make it easier for the guys working on CDC to make incremental progress.

cdc/cdc.cc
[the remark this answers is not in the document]
haaawk: @nyh there's already an issue https://github.com/scylladb/scylla/issues/4991

cdc/cdc.cc
haaawk: Another thing is that with prepared statements I can't insert all the rows with one request like I'm doing now. I would have to call the prepared statement for each row. That sounds wasteful.

cdc/cdc.cc
haaawk: I fixed the issue by gossiping the ignore_msb_bits for each node and using the right value when needed.

haaawk:
Thanks @penberg

slivne:
@avikivity ping ... can we merge this ?

avikivity:
> @avikivity ping ... can we merge this ?

I have no idea how to tell. @haaawk can you summarize what issues were resolved, and what issues (if any) are still pending?

avikivity:
One thing we need for sure (if not done already) is to qualify the feature on --experimental since we can't freeze any detail yet.

haaawk:
It's already done @avikivity

haaawk:
@avikivity anything else I have to do to get this merged?

cdc/cdc.cc
avikivity: I guessed it's a description, but it's still a vague name. Maybe setup_stream_description_table()?

avikivity:
I had a small comment about a function name, but it can be addressed later.

nyh:
Merged to "next".

Issue #4796, opened 2019-08-04
add counters for schema version 


slivne:
Export the schema by breaking it up to multiple counters (4) - so it will be easy to detect cases in which the schema is not in synch acorss the cluster 

Exporting https://github.com/scylladb/scylla/issues/4785#issuecomment-517163716 to an issue

slivne:
patch was dropped as we did not find a tehcincal solution - leaving the issue open as we may go back to this.

PR #5025, opened 2019-09-15
database.cc: Export current schema version as metrics

amnonh:
It is useful to be able to monitor the schema version per-node.
From the monitoring perspective, what is important is that all node
would have the same schema version.

This patch report the UUID as gauges. The UUID id reported as two gauges
one for the LSB and one for the MSB

The reported metrics loook like:

scylla_database_schema_version{shard="0",type="gauge",uuid="LSB"} -7498393077721644032.000000
scylla_database_schema_version{shard="0",type="gauge",uuid="MSB"} 6462017092514037760.000000

Fixes: #4796

### DIFF
--- a/database.cc
+++ b/database.cc
@@ -298,6 +298,7 @@ dirty_memory_manager::setup_collectd(sstring namestr) {
 }
 
 static const metrics::label class_label("class");
+static const metrics::label uuid_label("uuid");
 
 void
 database::setup_metrics() {
@@ -333,6 +334,7 @@ database::setup_metrics() {
                        sm::description("Holds the number of failed memtable flushes. "
                                        "High value in this metric may indicate a permanent failure to flush a memtable.")),
     });
+
 
     _metrics.add_group("database", {
         sm::make_gauge("requests_blocked_memory_current", [this] { return _dirty_memory_manager.region_group().blocked_requests(); },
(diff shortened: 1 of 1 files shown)


avikivity:
Maybe a better way is schema{version="$version"} = 1.

amnonh:
you cannot use labels, it will keep on changing all the time.

the idea is to identify a situation that one of the nodes has a different version than the others.

amnonh:
@avikivity see my last comment, I prefer not to abuse the labels by continuously creating and deleting metrics.

If we want to have the full information
I can use 4 gauges instead of 2, so each unit64 will be represented by 2 doubles, will that be acceptable by you?

avikivity:
No, it's terrible, let's look for antoher solution.

avikivity:
I don't see a better solution than using labels. It's also recommended here: https://www.robustperception.io/exposing-the-software-version-to-prometheus.

Yes it create lots of series, but we already expose tens of thousands. Exposing one per version isn't so terrible, we don't expect more than a few versions per day.

amnonh:
Exposing the software version as a metric is a legit solution and we are using it, but that change less than the number of restarts.

Anyone that would start creating lots of changes would add junk to the Prometheus server.

There's already a counter that changes each time a schema changed, so we get some notifications on it.

I'm against keep changing the metrics name (by changing the labels). we can close the PR and wait until will have a logger integration.

avikivity:
Closing as suggested.

avikivity:
I did see people export random info as labels.

PR #5031, opened 2019-09-15
Simplify and explain from_varint_to_integer

espindola:
This is the second version of the patch series. The previous one was just the second patch, this one adds more tests an another patch to make it easier to test that the new code has the same behavior as  the old one.

### DIFF
--- a/tests/castas_fcts_test.cc
+++ b/tests/castas_fcts_test.cc
@@ -167,6 +167,31 @@ SEASTAR_TEST_CASE(test_varint_to_bigint) {
     });
 }
 
+SEASTAR_TEST_CASE(test_varint_to_int) {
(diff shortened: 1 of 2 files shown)


espindola:
ping.  @kostja Could you review this one?

espindola:
ping 2. @tgrabiec maybe?

espindola:
merge ping

kostja:
@espindola pinging me on github is OK only as long as you're OK with a very late response.
Generally I prefer reviewing patches over email, just in case you would wish to solicit my review another time and make it easy for me to provide

avikivity:
Queed as 9e990725d90502feac24075779ccbc426667fa5f

PR #5051, opened 2019-09-18
recreate dependent user types

espindola:
When a user type changes we were not recreating other uses types that use it. This patch series fixes that and makes it clear which code is responsible for it.

### DIFF
--- a/db/schema_tables.cc
+++ b/db/schema_tables.cc
@@ -1131,10 +1131,31 @@ static std::vector<V> get_list(const query::result_set_row& row, const sstring& 
 // Create types for a given keyspace. This takes care of topologically sorting user defined types.
 template <typename T> static std::vector<user_type> create_types(keyspace_metadata& ks, T&& range) {
     cql_type_parser::raw_builder builder(ks);
+    std::unordered_set<bytes> names;
     for (const query::result_set_row& row : range) {
-        builder.add(row.get_nonnull<sstring>("type_name"),
-                        get_list<sstring>(row, "field_names"),
-                        get_list<sstring>(row, "field_types"));
+        auto name = row.get_nonnull<sstring>("type_name");
+        names.insert(to_bytes(name));
+        builder.add(std::move(name), get_list<sstring>(row, "field_names"), get_list<sstring>(row, "field_types"));
+    }
+    // Add user types that use any of the above types. From the
+    // database point of view they haven't changed since the content
+    // of system.types is the same for them. The runtime objects in
+    // the other hand now point to out of date types, so we need to
+    // recreate them.
+    for (const auto& p : ks.user_types()->get_all_types()) {
+        const user_type& t = p.second;
+        if (names.count(t->_name) != 0) {
+            continue;
+        }
+        for (const auto& name : names) {
+            if (t->references_user_type(t->_keyspace, name)) {
+                std::vector<sstring> field_types;
+                for (const data_type& f : t->field_types()) {
+                    field_types.push_back(f->as_cql3_type().to_string());
+                }
+                builder.add(t->get_name_as_string(), t->string_field_names(), std::move(field_types));
+            }
+        }
     }
     return builder.build();
 }
--- a/cql3/cql3_type.cc
+++ b/cql3/cql3_type.cc
@@ -37,7 +37,7 @@ namespace cql3 {
 cql3_type cql3_type::raw::prepare(database& db, const sstring& keyspace) {
     try {
         auto&& ks = db.find_keyspace(keyspace);
-        return prepare_internal(keyspace, ks.metadata()->user_types());
+        return prepare_internal(keyspace, *ks.metadata()->user_types());
     } catch (no_such_keyspace& nsk) {
         throw exceptions::invalid_request_exception("Unknown keyspace " + keyspace);
     }
@@ -62,7 +62,7 @@ public:
     virtual cql3_type prepare(database& db, const sstring& keyspace) {
         return _type;
     }
-    cql3_type prepare_internal(const sstring&, lw_shared_ptr<user_types_metadata>) override {
+    cql3_type prepare_internal(const sstring&, user_types_metadata&) override {
         return _type;
     }
 
@@ -110,7 +110,7 @@ public:
         return true;
     }
 
-    virtual cql3_type prepare_internal(const sstring& keyspace, lw_shared_ptr<user_types_metadata> user_types) override {
+    virtual cql3_type prepare_internal(const sstring& keyspace, user_types_metadata& user_types) override {
         assert(_values); // "Got null values type for a collection";
 
         if (!_frozen && _values->supports_freezing() && !_values->_frozen) {
@@ -180,7 +180,7 @@ public:
         _frozen = true;
     }
 
-    virtual cql3_type prepare_internal(const sstring& keyspace, lw_shared_ptr<user_types_metadata> user_types) override {
+    virtual cql3_type prepare_internal(const sstring& keyspace, user_types_metadata& user_types) override {
         if (_name.has_keyspace()) {
             // The provided keyspace is the one of the current statement this is part of. If it's different from the keyspace of
             // the UTName, we reject since we want to limit user types to their own keyspace (see #6643)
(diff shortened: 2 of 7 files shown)


bhalevy:
@tgrabiec please review

db/schema_tables.cc
```diff
@@ -1131,10 +1131,28 @@ static std::vector<V> get_list(const query::result_set_row& row, const sstring&
...
+        names.push_back(to_bytes(name));
+        builder.add(std::move(name), get_list<sstring>(row, "field_names"), get_list<sstring>(row, "field_types"));
+    }
+    // Add user types that use any of the above types. From the
+    // database point of view they haven't changed since the content
+    // of system.types is the same for them. The runtime objects in
+    // the other hand now point to out of date types, so we need to
+    // recreate them.
+    for (const auto& p : ks.user_types()->get_all_types()) {
+        const auto&t = p.second;
```
tgrabiec: missing space after &

tgrabiec: Please replace `auto` in a declaration of `t` with concrete type, it's hard to keep track of what we're working on here.

db/schema_tables.cc
```diff
@@ -1131,10 +1131,28 @@ static std::vector<V> get_list(const query::result_set_row& row, const sstring&
...
+        for (const auto& name : names) {
+            if (t->_name != name && t->references_user_type(t->_keyspace, name)) {
+                std::vector<sstring> field_types;
+                for (const data_type& f : t->field_types()) {
+                    field_types.push_back(f->as_cql3_type().to_string());
+                }
+                builder.add(t->get_name_as_string(), t->string_field_names(), std::move(field_types));
```
tgrabiec: What if `t` is also in `names` (meaning it's being modified here), wouldn't we add its entry twice, with a stale field names?

Also, since we changed `t->_name`, shouldn't we also update types which reference `t->_name` ?

database.cc
```diff
@@ -588,9 +580,10 @@ future<> database::parse_system_tables(distributed<service::storage_proxy>& prox
     }).then([&proxy, this] {
         return do_parse_schema_tables(proxy, db::schema_tables::TYPES, [this, &proxy] (schema_result_value_type &v) {
             auto& ks = this->find_keyspace(v.first);
+            auto& types = ks.metadata()->user_types();
             auto&& user_types = create_types_from_schema_partition(*ks.metadata(), v.second);
             for (auto&& type : user_types) {
-                ks.add_user_type(type);
+                types->add_type(type);
```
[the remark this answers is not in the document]
espindola: True. The code is even more redundant than I thought. I will remove it. Thanks.

db/schema_tables.cc
espindola: We do check if "t->_name != name", so I don't think we add it twice, is that the question?

For the second part, references_user_type checks the transitive property, so if we modified A, B references A and C references B, C->references_user_type(A) returns true.

espindola:
I pushed a v2 which:

* Removes redundant calls to add_type after create_types_from_schema_partition.
* Replaces "auto&t" with "user_type& t"

bhalevy:
@tgrabiec please re-review.

db/schema_tables.cc
tgrabiec: If `names` has more than one element, `t->_name != name` doesn't guarantee that `t` is not in `names`, it could be that `t->_name == name` for some later value of `name`. 

Suppose the type A references B before, and both A and B were changed. Suppose names = [A, B]. You first add A because it was changed, with the new field definitions. Then when you iterate over all types, you add it again because it references B.

It may also happen that when you add A with the new field definitions, you reference the old definition of B because it was not added yet.

espindola:
Tomasz Grabiec <notifications@github.com> writes:

> tgrabiec requested changes on this pull request.
>
>
>
>> +        builder.add(std::move(name), get_list<sstring>(row, "field_names"), get_list<sstring>(row, "field_types"));
> +    }
> +    // Add user types that use any of the above types. From the
> +    // database point of view they haven't changed since the content
> +    // of system.types is the same for them. The runtime objects in
> +    // the other hand now point to out of date types, so we need to
> +    // recreate them.
> +    for (const auto& p : ks.user_types()->get_all_types()) {
> +        const auto&t = p.second;
> +        for (const auto& name : names) {
> +            if (t->_name != name && t->references_user_type(t->_keyspace, name)) {
> +                std::vector<sstring> field_types;
> +                for (const data_type& f : t->field_types()) {
> +                    field_types.push_back(f->as_cql3_type().to_string());
> +                }
> +                builder.add(t->get_name_as_string(), t->string_field_names(), std::move(field_types));
>
> If `names` has more than one element, `t->_name != name` doesn't guarantee that `t` is not in `names`, it could be that `t->_name == name` for some later value of `name`. 
>
> Suppose the type A references B before, and both A and B were changed. Suppose names = [A, B]. You first add A because it was changed, with the new field definitions. Then when you iterate over all types, you add it again because it references B.
>
> It may also happen that when you add A with the new field definitions, you reference the old definition of B because it was not added yet.

Duh, no idea how I wrote that.

It is surprisingly hard to test. ALTER TYPE always creates a mutation of
size 1. I added a test that manually creates a larger mutation, and the
existing code indeed creates the types multiple times, but we get
"lucky" with the order we get types out of

        std::unordered_multimap<entry *, entry *> adjacency;

In any case, I will upload a new version soon that includes a test were
we were at least creating the type twice.

Cheers,
Rafael

espindola:
@tgrabiec updated.

database.cc
```diff
@@ -588,10 +580,7 @@ future<> database::parse_system_tables(distributed<service::storage_proxy>& prox
     }).then([&proxy, this] {
         return do_parse_schema_tables(proxy, db::schema_tables::TYPES, [this, &proxy] (schema_result_value_type &v) {
             auto& ks = this->find_keyspace(v.first);
-            auto&& user_types = create_types_from_schema_partition(*ks.metadata(), v.second);
-            for (auto&& type : user_types) {
-                ks.add_user_type(type);
-            }
+            create_types_from_schema_partition(*ks.metadata(), v.second);
```
tgrabiec: This comment is not directly targeted at this patch, but rather at 4f1260f3e36210e178869739c7e8557d038bb1da which started this. I am not sure that coupling type parsing with insertion into the keyspace is a good direction. It means that the operation which involves parsing cannot be safely rolled-back if it fails mid-way. There is no way to obtain a type instance without adding it to the keyspace.

db/schema_tables.cc
```diff
@@ -1131,10 +1131,31 @@ static std::vector<V> get_list(const query::result_set_row& row, const sstring&
...
+        if (names.count(t->_name) != 0) {
+            continue;
+        }
+        for (const auto& name : names) {
+            if (t->references_user_type(t->_keyspace, name)) {
+                std::vector<sstring> field_types;
+                for (const data_type& f : t->field_types()) {
+                    field_types.push_back(f->as_cql3_type().to_string());
+                }
+                builder.add(t->get_name_as_string(), t->string_field_names(), std::move(field_types));
```
tgrabiec: I think this version still doesn't deal with full transitivity. We changed `t` here, so we now have to also change all types depending on `t`.

database.cc
espindola: I think it could be changed to copy user_types_metadata, in which case this patch could not depend on build() adding the types. Do you want me to do that and rebase?

db/schema_tables.cc
espindola: references_user_type should handle transitivity. We only changed t because it references some type X. A type that references t also references X.

espindola:
replied. No idea if github sent you a notification or not.

db/schema_tables.cc
tgrabiec: Can it happen that a type that references t was updated before we updated t, and hence what we put into the builder will reference a stale definition of `t`?

database.cc
tgrabiec: Yes please

db/schema_tables.cc
espindola: The order types are added to the builder should not be important. The type references in the builder are by name and the builder does a topological sort.

database.cc
espindola: OK. I will send you another pull request with just that today.

db/schema_tables.cc
tgrabiec: I see. It's fine then.

espindola:
Turns out it is a lot easier to just add the patches to not modify the current types in place to this pull request.

I have updated the branch. Note that github shows the patchs in the wrong order :-(

Issue #5045, opened 2019-09-16
missing key after cleanup during decommission


bentsi:
*Installation details*
Scylla version (or git commit hash): `3.1.0.rc7-0.20190914.024d1563a`
Cluster size: 4
OS (RHEL/CentOS/Ubuntu/AWS AMI): ami-07103eb2a63bc2155 [eu-west-1]

Scenario:

1. 2019-09-16 03:23:34 Decommission node=Node longevity-200gb-48h-sst-corrupt-rep-db-node-1191a793-3 [34.253.143.14 | 10.0.123.58] (seed: False)\
```
Sep 16 03:23:43 ip-10-0-123-58.eu-west-1.compute.internal scylla[2360]:  [shard 9] compaction - Compacted 6 sstables to [/var/lib/scylla/data/keyspace1/standard1-c3a74c
Sep 16 03:23:43 ip-10-0-123-58.eu-west-1.compute.internal scylla[2360]:  [shard 0] storage_service - DECOMMISSIONING: starts
Sep 16 03:23:43 ip-10-0-123-58.eu-west-1.compute.internal scylla[2360]:  [shard 9] compaction - Compacting [/var/lib/scylla/data/keyspace1/standard1-c3a74c90d80811e9808
Sep 16 03:23:44 ip-10-0-123-58.eu-west-1.compute.internal scylla[2360]:  [shard 0] storage_service - LEAVING: sleeping 30000 ms for batch processing and pending range s
```

2. 2019-09-16 03:28:35 Kill nodetool decommission command (command exited, node continued to decommission
```
Sep 16 03:24:14 ip-10-0-123-58.eu-west-1.compute.internal scylla[2360]:  [shard 2] LeveledManifest - Adding high-level (L3) /var/lib/scylla/data/keyspace1/standard1-c3a
Sep 16 03:24:14 ip-10-0-123-58.eu-west-1.compute.internal scylla[2360]:  [shard 0] storage_service - DECOMMISSIONING: unbootstrap starts
Sep 16 03:24:14 ip-10-0-123-58.eu-west-1.compute.internal scylla[2360]:  [shard 0] storage_service - LEAVING: replaying batch log and streaming data to other nodes
Sep 16 03:24:14 ip-10-0-123-58.eu-west-1.compute.internal scylla[2360]:  [shard 0] range_streamer - Unbootstrap starts, nr_ranges_remaining=3076
[...]

PR #5072, opened 2019-09-20
storage_service: Reject nodetool cleanup when there is pending ranges

asias:
From Shlomi:

4 node cluster Node A, B, C, D (Node A: seed)
cassandra-stress write n=10000000 -pop seq=1..10000000 -node <seed-node>
cassandra-stress read duration=10h -pop seq=1..10000000 -node <seed-node>
while read is progressing
Node D: nodetool decommission
Node A: nodetool status node - wait for UL
Node A: nodetool cleanup (while decommission progresses)

I get the error on c-s once decommission ends
  java.io.IOException: Operation x0 on key(s) [383633374d31504b5030]: Data returned was not validated

The problem is when a node gets new ranges, e.g, the bootstrapping node, the
existing nodes after a node is removed or decommissioned, nodetool cleanup will
remove data within the new ranges which the node just gets from other nodes.

To fix, we should reject the nodetool cleanup when there is pending ranges on that node.

Note, rejecting nodetool cleanup is not a full protection because new ranges
can be assigned to the node while cleanup is still in progress. However, it is
a good start to reject until we have full protection solution.

Fixes: #5045

### DIFF
--- a/api/storage_service.cc
+++ b/api/storage_service.cc
@@ -304,17 +304,24 @@ void set_storage_service(http_context& ctx, routes& r) {
         if (column_families.empty()) {
             column_families = map_keys(ctx.db.local().find_keyspace(keyspace).metadata().get()->cf_meta_data());
         }
-        return ctx.db.invoke_on_all([keyspace, column_families] (database& db) {
-            std::vector<column_family*> column_families_vec;
-            auto& cm = db.get_compaction_manager();
-            for (auto cf : column_families) {
-                column_families_vec.push_back(&db.find_column_family(keyspace, cf));
+        return service::get_local_storage_service().is_cleanup_allowed(keyspace).then([&ctx, keyspace,
+                column_families = std::move(column_families)] (bool is_cleanup_allowed) mutable {
+            if (!is_cleanup_allowed) {
+                return make_exception_future<json::json_return_type>(
+                        std::runtime_error("Can not perform cleanup operation when topology changes"));
             }
-            return parallel_for_each(column_families_vec, [&cm] (column_family* cf) {
-                return cm.perform_cleanup(cf);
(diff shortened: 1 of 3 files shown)


asias:
Ping.

service/storage_service.cc
```diff
@@ -3554,5 +3554,16 @@ db::schema_features storage_service::cluster_schema_features() const {
     return f;
 }
 
+future<bool> storage_service::allow_cleanup(sstring keyspace) {
```
tgrabiec: Predicates should not read as imperatives, so I suggest e.g. "is_cleanup_allowed"

api/storage_service.cc
```diff
@@ -304,17 +304,24 @@ void set_storage_service(http_context& ctx, routes& r) {
...
-        return ctx.db.invoke_on_all([keyspace, column_families] (database& db) {
-            std::vector<column_family*> column_families_vec;
-            auto& cm = db.get_compaction_manager();
-            for (auto cf : column_families) {
-                column_families_vec.push_back(&db.find_column_family(keyspace, cf));
+        return service::get_local_storage_service().allow_cleanup(keyspace).then([&ctx, keyspace,
+                column_families = std::move(column_families)] (bool allow_cleanup) mutable {
+            if (!allow_cleanup) {
+                return make_exception_future<json::json_return_type>(
+                        std::runtime_error("Can not perform cleanup operation when topology changes"));
```
tgrabiec: It's not a complete solution to #5045, so you cannot close it with this change.

service/storage_service.cc
asias: Done.

api/storage_service.cc
asias: I changed to Refs.

asias:
Updated:
- Rename to is_cleanup_allowed
- Refs the issue not fix

PR #5073, opened 2019-09-20
repair: Fix tracker::start and tracker::done in case of error

### LINKED ISSUE #5074: repair: Handle tracker::start and tracker::done in case of error
The operation after gate.enter() in tracker::start() can fail and throw,
we should call gate.leave() in such case to avoid unbalanced enter and
leave calls. tracker::done() has similar issue too.

asias:
repair: Fix tracker::start and tracker::done in case of error

The operation after gate.enter() in tracker::start() can fail and throw,
we should call gate.leave() in such case to avoid unbalanced enter and
leave calls. tracker::done() has similar issue too.

Fix it by removing the gate enter and leave logic in tracker start and
done. A helper tracker::run() is introduced to take care of the gate and
repair status.

In addition, the errro log is improved. It now logs exceptions on all
shards in the summary. e.g.,

[shard 0] repair - repair id 1 failed: std::runtime_error
({shard 0: std::runtime_error (error0), shard 1: std::runtime_error (error1)})

Fixes #5074

### DIFF
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -235,7 +235,6 @@ tracker::~tracker() {
 }
 
 void tracker::start(int id) {
-    _gate.enter();
     _status[id] = repair_status::RUNNING;
 }
 
@@ -245,7 +244,6 @@ void tracker::done(int id, bool succeeded) {
     } else {
         _status[id] = repair_status::FAILED;
     }
-    _gate.leave();
 }
 repair_status tracker::get(int id) {
     if (id >= _next_repair_command) {
@@ -323,6 +321,19 @@ void tracker::abort_all_repairs() {
 
 semaphore& tracker::range_parallelism_semaphore() {
     return _range_parallelism_semaphores[engine().cpu_id()];
+}
+
+future<> tracker::run(int id, std::function<future<> ()> func) {
+    return seastar::with_gate(_gate, [this, id, func =std::move(func)] {
+        start(id);
+        return func().then([this, id] {
+            rlogger.info("repair id {} completed successfully", id);
+            done(id, true);
+        }).handle_exception([this, id] (std::exception_ptr ep) {
+            rlogger.info("repair id {} failed: {}", id, ep);
+            done(id, false);
+        });
+    });
 }
 
 void check_in_shutdown() {
@@ -1300,7 +1311,7 @@ static future<> repair_ranges(lw_shared_ptr<repair_info> ri) {
         repair_tracker().remove_repair_info(ri->id);
         return make_ready_future<>();
     }).handle_exception([ri] (std::exception_ptr eptr) {
-        rlogger.info("repair id {} failed - {}", ri->id, eptr);
+        rlogger.info("repair id {} on shard {} failed: {}", ri->id, engine().cpu_id(), eptr);
         repair_tracker().remove_repair_info(ri->id);
         return make_exception_future<>(std::move(eptr));
     });
@@ -1323,8 +1334,6 @@ static int do_repair_start(seastar::sharded<database>& db, sstring keyspace,
     // yet. Real ids returned by next_repair_command() will be >= 1.
     int id = repair_tracker().next_repair_command();
     rlogger.info("starting user-requested repair for keyspace {}, repair id {}, options {}", keyspace, id, options_map);
-    repair_tracker().start(id);
-    auto fail = defer([id] { repair_tracker().done(id, false); });
 
     if (!gms::get_local_gossiper().is_normal(utils::fb_utilities::get_broadcast_address())) {
         throw std::runtime_error("Node is not in NORMAL status yet!");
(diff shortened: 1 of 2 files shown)


slivne:
@asias is this relevant for 3.1 / 3.0 if so please open an issue so we cab do backports

asias:
> @asias is this relevant for 3.1 / 3.0 if so please open an issue so we cab do backports

Yes, backport candidate. I created https://github.com/scylladb/scylla/issues/5074.

repair/repair.cc
```diff
@@ -236,16 +236,26 @@ tracker::~tracker() {
...
-        _status.erase(id);
-    } else {
-        _status[id] = repair_status::FAILED;
+    try {
+        if (succeeded) {
+            _status.erase(id);
+        } else {
+            _status[id] = repair_status::FAILED;
+        }
+        _gate.leave();
```
vladzcloudius: It's definitely better but still not perfect.
What if `tracker::done(id)` is called without the preceding successful `tracker::start(id)` call?
In this case there still going to be unbalanced `gate::enter()` + `gate::leave()` calls amount.

The same is if `start(id)` is called twice for the same `id`...

You need to check that `_status[id]` exists before going forward both in `start(id)` and in `done(id)`.

asias: I think it is the job of the caller to guarantee it will have balanced call of start() and done(). It is like the gate won't protect you from calling unbalanced enter and leave().

Checking if _status[id] won't help for example, in done() it goes to the branch that does not erase the id. If we call done() multiple times, we will call gate::leave() more than gate::enter().

vladzcloudius: Gate has `with_gate()` to ensure that. 
gate::enter() and leave() are very low level methods and one indeed should be very careful when calling them and this low level API is indeed very unsafe - hence there is `with_gate()`

Ensuring safety with `tracker` *is* possible. Since it's much more primitive abstraction than the `gate` you can even make it safe without needing to implement `with_xxx()`. All what's needed is to separate the "active" state from the "status".

You can go with the `with_xxx` semantics of course - which will be MUCH safer than the split semantics you have now.

asias: Currently it use `defter` to make sure gate.leave() is called when the defter is destroyed. But, yes, I think with_gate is better here. I will send a new version shortly.

asias:
Vlad, do you have any idea why there is a duplicated comment. Did you use email to respond?

asias:
Pushed a new branch.
Changes:
- Move the gate enter and leave logic out of start() and done()
- Use with_gate

vladzcloudius:
> Vlad, do you have any idea why there is a duplicated comment. Did you use email to respond?

Nope. Wonders of github I guess... ;)

repair/repair.cc
vladzcloudius: > Currently it use `defter` to make sure gate.leave() is called when the defter is destroyed. But, yes, I think with_gate is better here. I will send a new version shortly.

I meant to implement something like `with_tracker(id, callable)`. This will make sure the above complications are not needed.

You implemented a split API in the tracker and it always leaves a place of a misuse (as you and I mentioned before). You can still make it quite safe even with the current API as I described above however if you implement a `with_tracker(...)` single-call notation this will be a much safer API.

Another small benefit is going to be that you would be able to easily incorporate a `with_gate(...)` into a `with_tracker(...)` API implementation.

repair/repair.hh
```diff
@@ -251,6 +251,7 @@ public:
     void abort_all_repairs();
     semaphore& range_parallelism_semaphore();
     static size_t max_repair_memory_per_range() { return _max_repair_memory_per_range; }
+    seastar::gate& gate() { return _gate; };
```
vladzcloudius: Following my comment above: this seems to be unsafe.

repair/repair.cc
asias: See my latest patch. It ready has a with_gate. with_tracker only wraps with_gate, I do not think it is very useful.

repair/repair.cc
```diff
@@ -1323,8 +1321,6 @@ static int do_repair_start(seastar::sharded<database>& db, sstring keyspace,
     // yet. Real ids returned by next_repair_command() will be >= 1.
     int id = repair_tracker().next_repair_command();
     rlogger.info("starting user-requested repair for keyspace {}, repair id {}, options {}", keyspace, id, options_map);
-    repair_tracker().start(id);
```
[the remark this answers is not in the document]
asias: See the comments of repair_tracker. 
Most of the stuff is accessed on shard zero only, the field that is accessed on all shards is stored inside a vector, e.g.
```
 // Map repair id into repair_info. The vector has smp::count elements, each
 // element will be accessed by only one shard.
 std::vector<std::unordered_map<int, lw_shared_ptr<repair_info>>> _repairs;
```

repair/repair.hh
asias: It is safe, because _gate is accessed from shard zero only.

asias:
> > Vlad, do you have any idea why there is a duplicated comment. Did you use email to respond?
> 
> Nope. Wonders of github I guess... ;)

I only saw this problem when discussing issues with you. Do you have special settings with you web browser?

vladzcloudius:
> > > Vlad, do you have any idea why there is a duplicated comment. Did you use email to respond?
> > 
> > 
> > Nope. Wonders of github I guess... ;)
> 
> I only saw this problem when discussing issues with you. Do you have special settings with you web browser?

Not that I know of.

repair/repair.cc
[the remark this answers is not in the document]
vladzcloudius: I see. I see that you have two such vectors. It looks like you are mirroring the `sharded` functionality here. I'd consider reworking this too.

repair/repair.hh
vladzcloudius: I see. You could consider separating the "sharded" part of the tracker and the "shard0 only" part into separate data structures. This would make maintaining a lot easier. But this is for future patches.

asias:
> Looks ok.
> It would be great if you'd implement tracker::do_with() that would have encorporated the `with_gate` call and the rest of `_status` logic. But the way it is now it's at least safe.
> It's still hard to with but it's definitely better than before.

Yes, wrapping with_gate alone is not very useful. But it makes sense to wrap the status as well. I had a new version which does that.

asias:
Updated a new version (d12cd2a): Introduced a new helper tracker::run which takes care of the gate and repair status.

asias:
Updated a new version (2d55765): 
- Introduced a new helper tracker::run which takes care of the gate and repair status.
- Fixed a typo in commit message.

slivne:
@nyh - can you please review / merge - vlad blessed this

nyh:
Committed into next (commit a9b95f5f01827968d3b2c39ec56a456e08ef6bfb)

Issue #4978, opened 2019-09-08
Failed to allocate during destroy_data_then_repair in large partition longevity


ShlomiBalalis:
This is Scylla's bug tracker, to be used for reporting bugs only.
If you have a question about Scylla, and not a bug, please ask it in
our mailing-list at scylladb-dev@googlegroups.com or in our slack channel.

- [x] I have read the disclaimer above, and I am reporting a suspected malfunction in Scylla.

*Installation details*
Scylla version (or git commit hash):3.1.0.rc5-0.20190902.623ea5e3d
Cluster size:4
OS (RHEL/CentOS/Ubuntu/AWS AMI):ami-027fe33bfba73c292

In large partition longevity, after a destroy_data_then_repair nemesis started, two of the four nodes, one of which was the target node, experienced an error and created a coredump with the following backtrace:

```
0x0000000004180762
0x000000000406d7e5
0x000000000406dae5
0x000000000406db93
0x00007f308813d02f
/libreloc/libc.so.6+0x000000000003853e
/libreloc/libc.so.6+0x0000000000022894
0x000000000402185a
0x00000000040228be
0x000000000131bf23
0x000000000137d9ff
0x00000000013174cf
0x000000000136cb21
0x0000000001455825
0x0000000001456a7e
0x000000000406acd1
0x000000000406aece
0x000000000414cf1d
0x000000000417755b
0x0000000004038f8d
/libreloc/libpthread.so.0+0x000000000000858d
/libreloc/libc.so.6+0x00000000000fd6a2
```

```
void seastar::backtrace<seastar::backtrace_buffer::append_backtrace()::{lambda(seastar::frame)#1}>(seastar::backtrace_buffer::append_backtrace()::{lambda(seastar::frame)#1}&&) at /usr/include/boost/program_options/variables_map.hpp:146
seastar::backtrace_buffer::append_backtrace() at /usr/include/boost/program_options/variables_map.hpp:146
 (inlined by) print_with_backtrace at /jenkins/workspace/scylla-3.1/relocatable-pkg/scylla/seastar/build/release/../../src/core/reactor.cc:1104
seastar::print_with_backtrace(char const*) at /usr/include/boost/program_options/variables_map.hpp:146
sigabrt_action at /usr/include/boost/program_options/variables_map.hpp:146
 (inlined by) operator() at /jenkins/workspace/scylla-3.1/relocatable-pkg/scylla/seastar/build/release/../../src/core/reactor.cc:5012
[...]

PR #5090, opened 2019-09-23
lsa: Assert no cross-shard region locking

tgrabiec:
We observed an abort on bad_alloc which was not caused by real OOM,
but could be explained by cache region being locked from a different
shard, which is not allowed, concurrently with memory reclamation.
    
It's impossible now to prove this, or, if that was indeed the case, to
determine which code path was attempting such lock. This patch adds an
assert which would catch such incorrect locking at the attempt.
    
Refs #4978

Tests:
  - unit (dev, release, debug)

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -843,7 +843,6 @@ pure_boost_tests = set([
     'tests/range_test',
     'tests/crc_test',
     'tests/checksum_utils_test',
-    'tests/managed_vector_test',
     'tests/dynamic_bitset_test',
     'tests/idl_test',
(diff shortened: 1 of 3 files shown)


bhalevy:
@avikivity please consider backporting to 3.1 to help debug #4978

avikivity:
Can we try to reproduce it on master?

glommer:
I personally would like to see assert patches like this backported even to
stable versions.

When the problems happen in the field, they can be undebuggable

PR #5098, opened 2019-09-25
messaging_service: use rpc::tuple instead of variadic futures for rpc

avikivity:
Since variadic future<> is deprecated, switch to rpc::tuple for multiple
return values in rpc calls. This is more or less mechanical translation.

### DIFF
--- a/service/storage_proxy.hh
+++ b/service/storage_proxy.hh
@@ -63,6 +63,14 @@
 #include "mutation_query.hh"
 #include "service_permit.hh"
 
+
+namespace seastar::rpc {
+
+template <typename... T>
+class tuple;
+
+}
+
 namespace locator {
 
 class token_metadata;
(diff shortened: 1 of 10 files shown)


service/storage_proxy.hh
```diff
@@ -63,6 +63,14 @@
...
 #include "service_permit.hh"
 
+
+namespace seastar::rpc {
+
+template <typename... T>
+class tuple;
+
+}
+
```
gleb-cloudius: why not  #include <seastar/rpc/rpc_types.hh>?

avikivity:
On 25/09/2019 10.21, Gleb Natapov wrote:
>
> *@gleb-cloudius* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In service/storage_proxy.hh 
> <https://github.com/scylladb/scylla/pull/5098#discussion_r327966759>:
>
> > @@ -63,6 +63,14 @@
>   #include "mutation_query.hh"
>   #include "service_permit.hh"
>   
> +
> +namespace seastar::rpc {
> +
> +template <typename... T>
> +class tuple;
> +
> +}
> +
>
> why not #include <seastar/rpc/rpc_types.hh>?
>
>

Reduce dependency proliferation.

espindola:

> On 25/09/2019 10.21, Gleb Natapov wrote:
>>
>> *@gleb-cloudius* commented on this pull request.
>>
>> ------------------------------------------------------------------------
>>
>> In service/storage_proxy.hh 
>> <https://github.com/scylladb/scylla/pull/5098#discussion_r327966759>:
>>
>> > @@ -63,6 +63,14 @@
>>   #include "mutation_query.hh"
>>   #include "service_permit.hh"
>>   
>> +
>> +namespace seastar::rpc {
>> +
>> +template <typename... T>
>> +class tuple;
>> +
>> +}
>> +
>>
>> why not #include <seastar/rpc/rpc_types.hh>?
>>
>>
>
> Reduce dependency proliferation.

Given that std::tuple has its own header, maybe it makes sense for us to
have seastar/rpc/tuple.hh. It would be a lot lighter than rpc_types.hh.

That can also be done as a followup since cleaning up the current
warnings is a big win.

LGTM

Cheers,
Rafael

PR #5099, opened 2019-09-25
mvcc: Fix incorrect schema version being used to copy the mutation when applying

### LINKED ISSUE #5095: Writes inserted into memtable may be interpreted using incorrect schema version on schema alters of a counter table
Scylla version (or git commit hash): 2.3+

Introduced in 27014a23d76ecbb169f75b73c2cf4f3b885e7178.

Relevant code:
```
void partition_entry::apply(const schema& s, const mutation_partition& mp, const schema& mp_schema)
{
    apply(s, mutation_partition(s, mp), mp_schema);
}
```

`mutation_partition(s, mp)` is incorrect, because it uses `s` to interpret `mp`, while it should use `mp_schema`.

We may hit this if the current node has a newer schema than the incoming mutation. This can happen during alter when we receive the mutation from a node which hasn't processed the schema change yet.

This is undefined behavior in general. If the alter was adding or removing columns, this may result in corruption of the write where values of one column are inserted into a different column.

tgrabiec:
Currently affects only counter tables.

Introduced in 27014a2.

mutation_partition(s, mp) is incorrect because it uses s to interpret
mp, while it should use mp_schema.

We may hit this if the current node has a newer schema than the
incoming mutation. This can happen during table schema altering when we receive the
mutation from a node which hasn't processed the schema change yet.

This is undefined behavior in general. If the alter was adding or
removing columns, this may result in corruption of the write where
values of one column are inserted into a different column.

Fixes #5095.

### DIFF
--- a/mutation_partition.cc
+++ b/mutation_partition.cc
@@ -146,7 +146,14 @@ mutation_partition::mutation_partition(const schema& s, const mutation_partition
         , _static_row(s, column_kind::static_column, x._static_row)
         , _static_row_continuous(x._static_row_continuous)
         , _rows()
-        , _row_tombstones(x._row_tombstones) {
+        , _row_tombstones(x._row_tombstones)
+#ifdef SEASTAR_DEBUG
+        , _schema_version(s.version())
+#endif
+{
+#ifdef SEASTAR_DEBUG
+    assert(x._schema_version == _schema_version);
+#endif
     auto cloner = [&s] (const auto& x) {
         return current_allocator().construct<rows_entry>(s, x);
     };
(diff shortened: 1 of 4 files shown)


mutation_partition.cc
```diff
@@ -146,7 +146,14 @@ mutation_partition::mutation_partition(const schema& s, const mutation_partition
...
         , _rows()
-        , _row_tombstones(x._row_tombstones) {
+        , _row_tombstones(x._row_tombstones)
+#ifdef SEASTAR_DEBUG
+        , _schema_version(s.version())
+#endif
+{
+#ifdef SEASTAR_DEBUG
+    assert(x._schema_version == _schema_version);
+#endif
```
bhalevy: how about defining SEASTAR_DEBUG_ASSERT to improve readability?

avikivity: It is not really related to Seastar. Scylla should define its own fine-grained debug macros.

For now it is okay to hitch a ride in SEASTAR_DEBUG.

PR #5107, opened 2019-09-26
Added separate counters for accesses to system tables

jul-stas:
Welcome to my first PR to Scylla!

The task was intended as a warm-up ("noob") exercise; its description is here: https://github.com/scylladb/scylla/issues/4182
Sorry, I also couldn't help it and did some scouting: edited descriptions of some metrics and shortened few annoyingly long LoC.

### DIFF
--- a/cql3/query_processor.cc
+++ b/cql3/query_processor.cc
@@ -96,14 +96,22 @@ query_processor::query_processor(service::storage_proxy& proxy, database& db, qu
                                      std::chrono::milliseconds(_db.get_config().permissions_update_interval_in_ms()),
                                      mcfg.authorized_prepared_cache_size, authorized_prepared_statements_cache_log) {
     namespace sm = seastar::metrics;
+    namespace stm = statements;
     using clevel = db::consistency_level;
     sm::label cl_label("consistency_level");
+
+    sm::label who_label("who");  // Who queried system tables
+    const auto user_who_label_instance = who_label("user");
+    const auto internal_who_label_instance = who_label("internal");
+
+    sm::label ks_label("ks");
+    const auto system_ks_label_instance = ks_label("system");
 
     std::vector<sm::metric_definition> qp_group;
     qp_group.push_back(sm::make_derive(
         "statements_prepared",
         _stats.prepare_invocations,
-        sm::description("Counts a total number of parsed CQL requests.")));
+        sm::description("Counts the total number of parsed CQL requests.")));
     for (auto cl = size_t(clevel::MIN_VALUE); cl <= size_t(clevel::MAX_VALUE); ++cl) {
         qp_group.push_back(
             sm::make_derive(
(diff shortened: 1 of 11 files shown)


cql3/query_processor.cc
```diff
@@ -103,7 +103,7 @@ query_processor::query_processor(service::storage_proxy& proxy, database& db, qu
     qp_group.push_back(sm::make_derive(
         "statements_prepared",
         _stats.prepare_invocations,
-        sm::description("Counts a total number of parsed CQL requests.")));
+        sm::description("Counts the total number of parsed CQL requests.")));
```
kbr-: Put the change that edits descriptions to a separate commit, and describe what's the nature of the change in the commit message (improving grammar or something).

cql3/statements/modification_statement.cc
```diff
@@ -45,6 +45,7 @@
 #include "cql3/restrictions/single_column_restriction.hh"
 #include "validation.hh"
 #include "db/consistency_level_validations.hh"
+#include "db/system_keyspace.hh"  // Decl. of `is_system_keyspace()'
```
kbr-: If anyone uses something else from this header in the future, I don't think they'll remember to update this comment.

cql3/statements/modification_statement.cc
```diff
@@ -71,14 +72,20 @@ modification_statement_timeout(const schema& s) {
     }
 }
 
-modification_statement::modification_statement(statement_type type_, uint32_t bound_terms, schema_ptr schema_, std::unique_ptr<attributes> attrs_, uint64_t* cql_stats_counter_ptr)
+modification_statement::modification_statement(statement_type type_,
+                                               uint32_t bound_terms,
+                                               schema_ptr schema_,
+                                               std::unique_ptr<attributes> attrs_,
+                                               uint64_t* cql_stats_counter_ptr,
+                                               uint64_t* cql_system_stats_counter_ptr)
```
kbr-: it's not instantly obvious what is "new" and what is just a cosmetic change (shortening LoCs). Please put the cosmetic change into a separate commit.

cql3/query_processor.cc
jul-stas: Sure, I'll do it, but...
While I understand the reason, I'm afraid that strict enforcing of such a rule might discourage future Boy Scouts. I'm not saying it's wrong, I'm just drawing the attention to something that might need more thought.

kbr-: True that, I myself have many times resigned from doing cosmetic changes like this due to the need to create a separate commit.
However when I feel a change is very important, I still do it (properly, with a commit).
But if I get a "meh, don't care so much" feeling, that tells me I shouldn't have wasted any thought on it in the first place (just my interpretation, I know a lot of people would disagree).

cql3/query_processor.cc
```diff
@@ -103,7 +103,7 @@ query_processor::query_processor(service::storage_proxy& proxy, database& db, qu
     qp_group.push_back(sm::make_derive(
```
denesb: Can't comment on the commit description, so will comment here: Please split the commit description into two parts. A single line summary (or subject line) that is preferably 72 chars or shorter (not always possible), followed by the the more elaborate description of what the commit does, separated from the subject line with an empty line.

Also please move the cosmetic changes into a separate commit, like @kbr- requested.

cql3/statements/modification_statement.cc
denesb: Broken lines should be indented just with a double indent (2x4 spaces).

cql3/statements/select_statement.cc
```diff
@@ -56,6 +56,7 @@
 #include "partition_slice_builder.hh"
 #include "cql3/untyped_result_set.hh"
 #include "db/timeout_clock.hh"
+#include "db/system_keyspace.hh"  // Decl. of `is_system_keyspace()'
```
denesb: Please remove this comment.

psarna: Better yet, forward declare this function only. Ref: 4f6b892aa1fb994c7ad6f19b8e5c3b660ff8f5c6

cql3/statements/modification_statement.cc
```diff
@@ -45,6 +45,7 @@
 #include "cql3/restrictions/single_column_restriction.hh"
 #include "validation.hh"
 #include "db/consistency_level_validations.hh"
+#include "db/system_keyspace.hh"
```
psarna: Please forward-declare instead, as in 4f6b892aa1

cql3/statements/select_statement.cc
```diff
@@ -56,6 +56,7 @@
 #include "partition_slice_builder.hh"
 #include "cql3/untyped_result_set.hh"
 #include "db/timeout_clock.hh"
+#include "db/system_keyspace.hh"
```
psarna: ditto

jul-stas:
Thanks everyone! I addressed your concerns, indentation included.

So, what now? Any formal DoD checklist to fill? Should I mark reviewers' points as resolved? wait for them to hit "resolve"/give me a written "OK"? Cover new code with tests? Give it to some testers? (Sorry, I still know little about the company and our SW pipelines)

denesb:
In general it is enough for all reviewers to approve the changes and then a maintainer to find the time to merge the PR. The latter is usually what takes more time. I usually ping PR:s that are stale for more than a week to remind reviewers/maintainers to finish the review/merge.

We try to cover all non-trivial code with tests. Cosmetic changes don't need tests and in the past we haven't written tests for counters either. So I think this PR is good to merge, unless somebody else has any objections.

cql3/query_processor.cc
```diff
@@ -137,6 +137,26 @@ query_processor::query_processor(service::storage_proxy& proxy, database& db, qu
...
+                            "system_updates",
+                            _cql_stats.system_updates,
+                            sm::description("Counts the number of CQL UPDATE requests executed on SYSTEM keyspace.")),
+
+                    sm::make_derive(
+                            "system_deletes",
+                            _cql_stats.system_deletes,
+                            sm::description("Counts the number of CQL DELETE requests executed on SYSTEM keyspace.")),
+
                     sm::make_derive(
```
avikivity: Maybe these should be done as a separate label (so we have reads{who="user"} and reads{who="system"}).

Pro:
 - can see all reads by summing across the who label

Con:
 - lots more stats flooding prometheus

cql3/statements/modification_statement.cc
```diff
@@ -394,6 +403,9 @@ modification_statement::do_execute(service::storage_proxy& proxy, service::query
     }
 
     inc_cql_stats();
+    if (::is_system_keyspace(keyspace())) {
+        inc_cql_system_stats();
+    }
```
avikivity: It's bad form to slow down stuff just for stats. A single increment instruction now has to go through multiple string compares.

You can compute a pointer during prepare() and increment via the pointer.

Another question is whether a system read is a read of the system keyspace, or a read that the system executes for its internal purposes. It's even not clear what "system" is.

 - should writes to the traces tables be considered system writes? I think so
 - should driver queries of the system tables to get the schema be considered system reads? I think not

/cc @glommer who eats metrics for one meal a day

cql3/statements/modification_statement.hh
```diff
@@ -110,12 +110,21 @@ private:
         };
 
     uint64_t* _cql_modification_counter_ptr = nullptr;
+    /// Like `_cql_modification_counter_ptr' but incremented upon updates of "system" keyspace.
+    uint64_t* _cql_system_modification_counter_ptr = nullptr;
+
```
avikivity: Should have just one of these pointers, computed at prepare time.

avikivity: (or a different cql_stats structure passed by the caller, if we decide to make it depend on who initiated the query rather than the keyspace the query ended up affecting)

amnonh:
On Sun, Sep 29, 2019 at 10:35 AM Avi Kivity <notifications@github.com>
wrote:

>
> In cql3/query_processor.cc
> <https://github.com/scylladb/scylla/pull/5107#discussion_r329341975>:
>
> > +
> +                    sm::make_derive(
> +                            "system_inserts",
> +                            _cql_stats.system_inserts,
> +                            sm::description("Counts the number of CQL INSERT requests executed on SYSTEM keyspace.")),
> +
> +                    sm::make_derive(
> +                            "system_updates",
> +                            _cql_stats.system_updates,
> +                            sm::description("Counts the number of CQL UPDATE requests executed on SYSTEM keyspace.")),
> +
> +                    sm::make_derive(
> +                            "system_deletes",
> +                            _cql_stats.system_deletes,
> +                            sm::description("Counts the number of CQL DELETE requests executed on SYSTEM keyspace.")),
> +
>                      sm::make_derive(
>
> Maybe these should be done as a separate label (so we have
> reads{who="user"} and reads{who="system"}).
>
> Pro:
>
>    - can see all reads by summing across the who label
>
> Con:
>
>    - lots more stats flooding prometheus
>
> The system should be a key-space label not part of the name.
An example of that you can find at table.cc for ks/table labels, it's
likely that we'll add more ks/tables specific metrics in the future.

also, I prefer that the CQL command will be a label and not part of the
name, the cons is that you cannot create a specific description
the plus, it makes it easier to use with function and dashboards.
An example of that you can find at alternator/stats.cc

cql3/query_processor.cc
haaawk: It doesn't have to be a separate PR. Just a separate commit. Shouldn't be too hard and discourage anyone :).

cql3/statements/modification_statement.cc
avikivity: Another way to address this is to add a pointer to the keyspace at prepare time, increment a keyspace metric, and have the metric fetch lambda sum over all system keyspaces. That's if we want by keyspace rather than by caller.

slivne: @avikivity some of the questions you raised are inluded in the issue description https://github.com/scylladb/scylla/issues/4182#issuecomment-535011816

- we do need to count the driver queries to system tables as they are misleading our users in the cql_optimization dahsboard - full info in the issue

- we do need to split put the internal system read/writes as well

jul-stas:
[UPDATED]
I chose a simple approach: to split the counters and increment one of them depending on the value of `client_state.is_internal()`. Also, moved string comparison to statements' ctors and added label.

jul-stas:
@avikivity @amnonh Sorry for bothering, would you mind having a look at my current version? I kept your comments as "unresolved" so you have a reminder of what was wrong.

cql3/query_processor.cc
```diff
@@ -119,62 +124,134 @@ query_processor::query_processor(service::storage_proxy& proxy, database& db, qu
...
+                            [this] {
+                                return _cql_stats.query_cnt(source_selector::USER, ks_selector::SYSTEM, stm::statement_type::DELETE)
+                                        + _cql_stats.query_cnt(source_selector::USER, ks_selector::NONSYSTEM, stm::statement_type::DELETE)
+                                        + _cql_stats.query_cnt(source_selector::INTERNAL, ks_selector::SYSTEM, stm::statement_type::DELETE)
+                                        + _cql_stats.query_cnt(source_selector::INTERNAL, ks_selector::NONSYSTEM, stm::statement_type::DELETE);
+                            },
+                            sm::description("Counts the total number of CQL DELETE requests.")),
+
+                    sm::make_derive(
+                            "system_reads",
```
amnonh: Please don't use system as a metric name, it should be a keyspace label (see my comment and references on the previous version)

The same goes for the rest of the operations.

make sure that when summing a metric over all labels you don't count the same metric twice

cql3/query_processor.cc
```diff
@@ -250,12 +327,12 @@ query_processor::query_processor(service::storage_proxy& proxy, database& db, qu
...
                             "reverse_queries",
                             _cql_stats.reverse_queries,
-                            sm::description("Counts number of CQL SELECT requests with ORDER BY DESC.")),
+                            sm::description("Counts the number of CQL SELECT requests with ORDER BY DESC.")),
 
                     sm::make_counter(
                             "unpaged_select_queries",
                             _cql_stats.unpaged_select_queries,
-                            sm::description("Counts number of unpaged CQL SELECT requests.")),
+                            sm::description("Counts the number of unpaged CQL SELECT requests.")),
```
amnonh: it will be good to add a specific internal counter for that as well.

cql3/query_processor.cc
```diff
@@ -250,12 +327,12 @@ query_processor::query_processor(service::storage_proxy& proxy, database& db, qu
                     sm::make_counter(
                             "reverse_queries",
                             _cql_stats.reverse_queries,
-                            sm::description("Counts number of CQL SELECT requests with ORDER BY DESC.")),
+                            sm::description("Counts the number of CQL SELECT requests with ORDER BY DESC.")),
```
amnonh: The description is not correct, but I understand you only fixing the grammar.

those are requests with reverse ORDER BY.

cql3/query_processor.cc
jul-stas: Thanks, do you mean to split it like this?
scylla_cql_unpaged_select_queries{shard="x",**keyspace="system"**,type="counter"} 123
scylla_cql_unpaged_select_queries{shard="x",**keyspace="nonsystem"**,type="counter"} 456
Or maybe even this (?) :
scylla_cql_unpaged_queries{shard="x",**keyspace="system",query="select"**,type="counter"} 123

amnonh: first keyspace label as defined in table.cc (see my comment on the previous version) is called ks, use the same name.

don't use keyspace nonsystem.

for the unpaged_select_queries, you can leave the name as it is.
Paging is only relevant to select, no point in adding the extra labels

jul-stas: My apologies: the example of `scylla_cql_unpaged_select_queries` was unfortunate. But the general questions still remain:
1. would you like to have label `query` as above?
2. Example from table.cc reads:
`auto ks = keyspace_label(_schema->ks_name());`
Are you suggesting to make separate counters per every KS? The motivation was to isolate the metrics for `system` KS, not **every** KS.
2.a. To avoid counting metrics twice when summing over labels, labels should be "mutually exclusive". What else can it be if not `nonsystem`? The name is wrong?

amnonh: 1. a query label can replace the cql command that is part of the name, I think it's better but it's not a must.

2. the problem with nonsystem, is that it's not a real keyspace name and it will block us in the future from adding keyspace names.
I see two options:
a. don't add a keyspace label for non-system, but then don't sum over it.
b. use a different name, that includes only metrics with keyspace, for now, it will only have system, but in the future, we can add more keyspaces to it.

cql3/query_processor.cc
```diff
@@ -119,62 +127,134 @@ query_processor::query_processor(service::storage_proxy& proxy, database& db, qu
...
+                            {user_who_label_instance, system_ks_label_instance}),
+
+                    sm::make_derive(
+                            "per_ks_reads",
+                            _cql_stats.query_cnt(source_selector::INTERNAL, ks_selector::SYSTEM, stm::statement_type::SELECT),
+                            sm::description(),
+                            {internal_who_label_instance, system_ks_label_instance}),
+
+                    sm::make_derive(
+                            "per_ks_inserts",
```
amnonh: if you are adding the per_ks please add it as a suffix, it would make searching easier

jul-stas: Done.
Apparently Github didn't reload this PR after I force pushed, but it should be able to be cleanly merged (even by ff-only). Unless you have more comments, of course.

cql3/query_processor.cc
```diff
@@ -119,62 +124,134 @@ query_processor::query_processor(service::storage_proxy& proxy, database& db, qu
...
                             "reads",
-                            _cql_stats.reads,
-                            sm::description("Counts a total number of CQL read requests.")),
+                            [this] {
+                                return _cql_stats.query_cnt(source_selector::USER, ks_selector::SYSTEM, stm::statement_type::SELECT)
+                                        + _cql_stats.query_cnt(source_selector::USER, ks_selector::NONSYSTEM, stm::statement_type::SELECT)
+                                        + _cql_stats.query_cnt(source_selector::INTERNAL, ks_selector::SYSTEM, stm::statement_type::SELECT)
+                                        + _cql_stats.query_cnt(source_selector::INTERNAL, ks_selector::NONSYSTEM, stm::statement_type::SELECT);
+                            },
+                            sm::description("Counts the total number of CQL read requests.")),
```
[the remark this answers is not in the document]
haaawk: What lambda function do you mean, @amnonh? I can't find any.

jul-stas:
@amnonh Hi Amnon, do you think we can finally merge it?

nyh:
@amnonh @avikivity both of you had requests on this issue earlier, and I have no idea if the new version addresses your concern, or not enough. Can you please officially "review" this issue with a decision (request changes or accept)? Thanks.

cql3/query_processor.cc
```diff
@@ -124,96 +124,96 @@ query_processor::query_processor(service::storage_proxy& proxy, database& db, qu
                     sm::make_derive(
                             "reads",
                             _cql_stats.statements[size_t(statement_type::SELECT)],
-                            sm::description("Counts a total number of CQL read requests.")),
```
haaawk: Why are you removing ')'? Does this compile?

cql3/query_processor.cc
```diff
@@ -123,97 +131,209 @@ query_processor::query_processor(service::storage_proxy& proxy, database& db, qu
...
+                    sm::make_derive(
+                            "inserts_per_ks",
+                            _cql_stats.query_cnt(source_selector::USER, ks_selector::SYSTEM, cond_selector::NO_CONDITIONS, stm::statement_type::INSERT),
+                            sm::description("Counts the number of CQL INSERT requests executed on particular keyspaces. "
+                                            "Label `who' indicates where the reqs come from (clients or DB internals)"),
+                            {user_who_label_instance, system_ks_label_instance, non_cas_label_instance}),
+                    sm::make_derive(
+                            "inserts_per_ks",
+                            _cql_stats.query_cnt(source_selector::INTERNAL, ks_selector::SYSTEM, cond_selector::NO_CONDITIONS, stm::statement_type::INSERT),
+                            sm::description(),
```
amnonh: why the empty description?

amnonh:
The descriptions are missing, besides that it's ok

cql3/query_processor.cc
jul-stas: thanks, I added descriptions

haaawk:
@nyh @avikivity any chances of getting this merged?

haaawk:
@nyh @avikivity ping

PR #5108, opened 2019-09-26
alternator: Actually use BEGINS_WITH in its tests

dekimir:
For some reason, BEGINS_WITH tests used EQ as comparison operator.

Tests: pytest test_expected.py


nyh:
Oops :-) As you can guess, I copied these tests over and over for the different types of comparison operators, so although I tried to be careful, it's not entirely surprising I forgot to update some of the copied lines. I'll commit.

PR #5132, opened 2019-10-02
Redis API in Scylla

fastio:
The detailed design and implementation of Redis API in Scylla is provided in the doc/redis/redis.md.
The redis commands (such as PING, GET, SET, DEL) are already implemented.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -456,6 +456,8 @@ arg_parser.add_argument('--enable-alloc-failure-injector', dest='alloc_failure_i
                         help='enable allocation failure injection')
 arg_parser.add_argument('--with-antlr3', dest='antlr3_exec', action='store', default=None,
                         help='path to antlr3 executable')
+arg_parser.add_argument('--with-ragel', dest='ragel_exec', action='store', default='ragel',
+        help='path to ragel executable')
 args = arg_parser.parse_args()
 
 defines = ['XXH_PRIVATE_API',
@@ -802,6 +804,26 @@ alternator = [
        'alternator/auth.cc',
 ]
 
+redis = [
+        'redis/service.cc',
+        'redis/server.cc',
+        'redis/query_processor.cc',
+        'redis/protocol_parser.rl',
+        'redis/keyspace_utils.cc',
+        'redis/options.cc',
+        'redis/stats.cc',
+        'redis/mutation_utils.cc',
+        'redis/query_utils.cc',
+        'redis/abstract_command.cc',
+        'redis/command_factory.cc',
+        'redis/commands/unknown.cc',
+        'redis/commands/ping.cc',
+        'redis/commands/set.cc',
+        'redis/commands/get.cc',
+        'redis/commands/del.cc',
+        'redis/commands/select.cc',
+        ]
+
 idls = ['idl/gossip_digest.idl.hh',
         'idl/uuid.idl.hh',
         'idl/range.idl.hh',
@@ -847,7 +869,7 @@ scylla_tests_dependencies = scylla_core + idls + scylla_tests_generic_dependenci
 ]
 
 deps = {
-    'scylla': idls + ['main.cc', 'release.cc'] + scylla_core + api + alternator,
+    'scylla': idls + ['main.cc', 'release.cc'] + scylla_core + api + alternator + redis,
 }
 
 pure_boost_tests = set([
@@ -1237,6 +1259,11 @@ if args.antlr3_exec:
 else:
     antlr3_exec = "antlr3"
 
+if args.ragel_exec:
+    ragel_exec = args.ragel_exec
+else:
+    ragel_exec = "ragel"
+
 for mode in build_modes:
     configure_zstd(outdir, mode)
 
@@ -1271,6 +1298,11 @@ with open(buildfile_tmp, 'w') as f:
             command = {ninja} -C $subdir $target
             restat = 1
             description = NINJA $out
+        rule ragel
+            # sed away a bug in ragel 7 that emits some extraneous _nfa* variables
+            # (the $$ is collapsed to a single one by ninja)
+            command = {ragel_exec} -G2 -o $out $in && sed -i -e '1h;2,$$H;$$!d;g' -re 's/static const char _nfa[^;]*;//g' $out
+            description = RAGEL $out
         rule run
             command = $in > $out
             description = GEN $out
@@ -1335,6 +1367,7 @@ with open(buildfile_tmp, 'w') as f:
         swaggers = {}
         serializers = {}
         thrifts = set()
+        ragels = {}
         antlr3_grammars = set()
         seastar_dep = 'build/{}/seastar/libseastar.a'.format(mode)
         for binary in build_artifacts:
@@ -1392,6 +1425,9 @@ with open(buildfile_tmp, 'w') as f:
                 elif src.endswith('.json'):
                     hh = '$builddir/' + mode + '/gen/' + src + '.hh'
                     swaggers[hh] = src
+                elif src.endswith('.rl'):
+                    hh = '$builddir/' + mode + '/gen/' + src.replace('.rl', '.hh')
+                    ragels[hh] = src
                 elif src.endswith('.thrift'):
                     thrifts.add(src)
                 elif src.endswith('.g'):
@@ -1420,6 +1456,7 @@ with open(buildfile_tmp, 'w') as f:
             gen_headers += g.headers('$builddir/{}/gen'.format(mode))
         gen_headers += list(swaggers.keys())
         gen_headers += list(serializers.keys())
+        gen_headers += list(ragels.keys())
         gen_headers_dep = ' '.join(gen_headers)
 
         for obj in compiles:
@@ -1433,6 +1470,9 @@ with open(buildfile_tmp, 'w') as f:
         for hh in serializers:
             src = serializers[hh]
             f.write('build {}: serializer {} | idl-compiler.py\n'.format(hh, src))
+        for hh in ragels:
+            src = ragels[hh]
+            f.write('build {}: ragel {}\n'.format(hh, src))
         for thrift in thrifts:
             outs = ' '.join(thrift.generated('$builddir/{}/gen'.format(mode)))
             f.write('build {}: thrift.{} {}\n'.format(outs, mode, thrift.source))
--- a/db/config.cc
+++ b/db/config.cc
@@ -713,6 +713,21 @@ db::config::config(std::shared_ptr<db::extensions> exts)
     , alternator_address(this, "alternator_address", value_status::Used, "0.0.0.0", "Alternator API listening address")
     , alternator_enforce_authorization(this, "alternator_enforce_authorization", value_status::Used, false, "Enforce checking the authorization header for every request in Alternator")
     , abort_on_ebadf(this, "abort_on_ebadf", value_status::Used, true, "Abort the server on incorrect file descriptor access. Throws exception when disabled.")
+    , redis_transport_port(this, "redis_transport_port", value_status::Used, 6379, "Port on which the REDIS transport listens for clients.")
+    // FIXME: 9142 has been used by native_transport_port_ssl
+    , redis_transport_port_ssl(this, "redis_transport_port_ssl", value_status::Used, 9142, "Port on which the REDIS TLS native transport listens for clients.")
+    , enable_redis_protocol(this, "enable_redis_protocol", value_status::Used, false, "Enable redis protocol; Scylla will process redis protocol as a redis cluster if enable.")
+    , redis_read_consistency_level(this, "redis_read_consistency_level", value_status::Used, "ONE", "Consistency level for read operations for redis.")
+    , redis_write_consistency_level(this, "redis_write_consistency_level", value_status::Used, "ANY", "Consistency level for write operations for redis.")
+    , redis_default_database_count(this, "redis_default_database_count", value_status::Used, 16, "Default database count for the redis cluster.")
+    , redis_keyspace_options(this, "redis_keyspace_options", value_status::Used, {}, 
+            "Enable redis protocol, list of properties of replication of redis keyspace. The available options are:\n"
+            "\n"
+            "\tclass : (Default: SimpleStrategy ).\n"
+            "\treplication_factor: (Default: 1) If the class is SimpleStrategy, set the replication factor for the strategy.\n"
+            "\tdatacenter_i: (Default: 1) If the class is NetworkTopologyStrategy, set the replication factor per datacenter.\n"
+            "\n"
+            "The properties of replication for redis keyspace.")
 
     , default_log_level(this, "default_log_level", value_status::Used)
     , logger_log_level(this, "logger_log_level", value_status::Used)
--- a/docs/redis/redis.md
+++ b/docs/redis/redis.md
@@ -1,0 +1,362 @@
+Redis API in Scylla
+
+## 1. Overview
+
+Redis is a very  famous NoSQL in-memory data structure store that stores a
+mapping of keys to different types of values, and familiar to many developers.
+It supports data structures such as STRINGs, HASHes, LISTs, SETs, ZSETs(sorted sets),
+and other data structures. Redis has built-in replication, Lua
+scripting, LRU eviction, transactions and different levels of on-disk
+persistence. Now Redis as memory store service is provided by many cloud
+platforms.
+
+In this document, the detailed design and implementation of Redis that build on
+top of Scylla is provided. At the beginning, the main feature, which is as a
+data structures store, is design and  implmented. In the future, the reset
+features of Redis will be supported.
+
+## 2. Motivation
+
+In contrast,  [Scylla](https://www.scylladb.com/) has advantage and amazing
+features:
+
+* Low and consistent latency
+* Data persistence
+* High availability
+* High throughtput
+* Highly scalable
+* Auto tuning
+
+If Redis build on the top of Scylla, it has the above features automatically.
+It's achived great progress in cluster master managment, data persistence,
+failover and replication.
+
+The benefits to the users are easy to use and develop in their production
+environment, and taking avantages of Scylla.
+
+## 3. The Protocol Server
+
+Redis clients communicate with the Redis server using a protocol called
+**RESP** (REdis Serialization Protocol). The protocol is binary-safe and
+easy to be implement at client side, which was designed specifically for
+Redis.
+
+Redis server accepts commands composed of different arguments. Once a
+command is received, it's processed and a reply is sent back to the client.
+Two kind of Redis clients are widely used, the smart Redis client and
+simple Redis client. When using smart Redis client to connect the Redis
+cluster, it will send the commands to the right Redis server. Because the
+client will get the hash slots and server nodes mapping data at the bootrap
+time. When using simple Redis client,  the server address is needed, and
+all the requests are sent to this address, and the requests only send to
+the single Redis server.
+
+In Scylla terminology, the node receiving the request acts as the the
+coordinator, and often passes the request on to one or more other nodes,
+which hold copies of the requested data.  Any node of Scylla cluster can
+process the request correctly. In other words, Client can send request to
+any node of Redis cluster built on the top of Scylla. Obviously, the simple
+client can accesses the Redis cluster built on the top of Scylla, when
+provided address is not an IP but rather a single domain name, which a
+client resloves a random Scylla node IP address. The smart Redis client
+queries the mapping information between hash slots and Server nodes. We
+just fill the mapping with random Scylla node IP address.
+
+Before you can start using Scylla with Redis API, you must set the
+`enable_redis_protocol` configuration option (visa the command line or YAML),
+to `true`, which default is `false`. By default, the Scylla will listen on the
+port (default is 6379, you can set the `redis_transport_prot` to other value).
+
+With Redis enabled, every Scylla node listens for Redis requests on this port.
+These requests, in [RESP](https://redis.io/topics/protocol) format over TCP,
+are parsed and result in calls to internal Scylla C++ functions.
+
+## 4. Data Model
+
+Out of box, every Redis cluster supports 16 databases. The default database
+is 0 but it allows us to change that to any number from 0-15 (and allow us
+to configure Redis to support more databases). Each database provides
+a distinct keyspace, independent from the others. Use `SELECT n`
+to change databases. Redis allows us to store keys that map to any one of
+five different  basic data structure types:  STRINGs,  LISTs,  SETs,
+HASHes, and  ZSETs. (In fact, Now Redis has support other structure types,
+but the basic structure types are used widely).
+
+In this proposal, We use the column family of Scylla to simulate the
+database within Redis. At the bootrap phase, the column familes with fixed
+name should be created (if not exists) or loaded.
+
+We known that, Redis allows us to store keys that map to any one of the
+basic data structures. Scylla supports variant type, which allows us to
+define one table to store all kind of these Redis data. For example ,we
+create the table with the schame as follow:
+
+```
+CREATE TABLE redis （
+    key text,
+    value variant\<text, list\<text\>, map\<text\>\>,
+    PRIMARY KEY(key)
+）WITH ...
+```
+
+However,  it's need to prefetch the whole elements of this type structure
+into memory in Scylla, even modifying only one element of the structures.
+The main limitation is that the size of the data structure is not allowed
+to be exceeded the limit of the server's memory size. And the operations
+will have more performance cost with bigger data structure.
+
+> In Redis, maximum length of a list is 23^2  - 1 elements (4294967295,
+> more than 4 billion of elements per list). And maximum length of
+> element (as a string) is 512MB. It's very common that the size of data
+> structure is exceeded the server's memory size.
+
+So single table to store all the data strucutures is not good idea, instead
+of five independent tables are created to hold the the different Redis
+structure types within each column family related Redis database. In other
+words, all of the STRINGs of the Redis will be stored in a table within the
+column family, and all of the LISTs' elements will be stored in anther
+table within the column family, and so on.
+
+> When building Redis on top of Scylla, for every Redis  database (16
+> databases as default), 1 column family with 5 tables in Scylla
+> will be created.
+
+The disadvantages of this proposal is that, we can not the TYPE command of
+Redis. The different type strutures are stored in the different Scylla
+tables, and each Scylla table provide the independent keyspace. In Redis,
+keys are unique in the database. However,  the keys with column family of
+Scylla are not unique. Beacause the keys are splited into independent
+Scylla tables, it's different to original Redis.
+
+**IMPORTANT NOTE**: The keys with database of Redis are not unique  in this
+proposal, is the main difference with original Redis.
+
+> This limitation is acceptable. The top layer of bussness system knows the
+> context of the keys.
+
+In the rest of this section, the details of schema of tables within column
+family are provided.
+
+### 4.1  Table Schema of STRINGs
+
+In Redis, STRINGs are similar to strings that we see in other languages or
+other key-value stores. Every data item only has two parts, KEY and VALUE.
+The partition in the Scylla table owns the TTL property. We just need two
+columns in the table to store Redis STRING data.
+
+The maximum allowed key and value size is 512 MB. Very long key size is not
+good idea.
+
+Within the Scylla column family, we create the table named `STRINGs to
+store the string structure of Redis. The table STRINGs is created by
+following CQL:
+
+```
+CREATE TABLE STRINGs (
+    pkey text,
+    data text,
+    PRIMARY KEY (key)
+) WITH ... ;
+```
+
+Every Key-Value pair is stored as one partition within STRINGs table.
+All commands of STRINGs and some shared commands will operate this table.
+
+### 4.2 Table Schema of LISTs
+
+In Redis, LISTs store an ordered sequence of strings which sorted by
+insertion order. It allows us to push items to the front and the back of
+the LIST with LPUSH/RPUSH, and to pop items from the front and back of the
+list with LPOP/RPOP, and to fetch an item at a given position with LINDEX,
+and to fetch a range of items with LRANGE.
+
+Maximum length of a list is 23^2  - 1 elements (4294967295, more than 4
+billion of elements per list). And maximum length of element (as a string) is 512MB.
+
+As we known, Scylla supports 3 kinds of collections, lists, maps,
+and sets. We do not use the lists collection to store Redis LISTs
+structure. Because, the LISTs structure in Redis may contains many
+elements ( Maximum is more than 4 billion of elements per list).
+When updating the lists collection in Scylla, the whole elements
+will be loaded into memory.
+
+The better way is that the LISTs of Redis are stored as the
+partition of the Scylla table. All commons of Redis LISTs will
+operate on this table, which is created by following CQL:
+
+```
+CREATE TABLE LISTs (
+    pkey text,
+    ckey text,
+    data text,
+    PRIMARY KEY(pkey, ckey)
+) WITH ... ;
+```
+
+The pkey is mapped to Redis LISTs key, and ckey is the UUID of the
+insertion timestamp, which will be keep the right order as the insertion
+order. The element's value is stored in the data column within LISTs table.
+
+### 4.3  Table Schema of HASHes
+
+In Redis, HASHes are maps between the string fields and the string values.
+The fields are unique within a HASH structure in Redis. The values that can
+be stored in HASHes are the same as what can be stored as normal STRINGs.
+
+HASHes provide constant time basic operations like HGET, HSET, HEXISTS etc.
+
+As mentioned above, Scylla provides the 3 kinds of collection data type,
+including map.
+We do not use this collection type of Scylla to strore HASHes data with the
+same reason.
+
+The better way is that the HASHes of Redis are stored as the partition of
+the Scylla table. All commons of Redis HASHes will operate on this table,
+which is created by following CQL:
+
+```
+CREATE TABLE HASHes (
+    pkey text,
+    ckey text,
+    data text,
+    PRIMARY KEY(pkey, ckey)
+) WITH ... ;
+```
+
+The HASHes structure is stored as the partition of Scylla table. The pkey
+is the partition key. The HASHes structure's field is storted as ckey
+(cluster key in HASHes table). The HASHes structure's value is stored as
+data column.
+
+### 4.4 Table Schema of SETs
+
+In Redis, SETs allows us to store sequence of strings, and uses a hash
+table to keep all strings unique.
+
+For the same reason as LISTs/HASHes, we do not use the collection type of
+Scylla to store Redis SETs data. The table is created by following CQL:
+
+```
+CREATE TABLE SETs (
+    pkey text,
+    ckey text,
+    PRIMARY KEY(pkey, ckey)
+) WITH ... ;
+```
+
+Unlike HASHes, There is not data column in SETs table. Every SETs structure
+is stored as a partition in SETs table. We use the cluster key to store the
+item of Redis SETs.
+
+### 4.5 Table Schema of ZSETs(Sorted SETs)
+
+In fact, ZSETs in Redis is similar to HASHes, which maps the STRINGs key to
+value. Like HASHes, the keys (called  MEMBERS) are unique. But the values (called
+SCORES) are limited to floating-point numbers.  ZSETs have the
+unique property in Redis of being able to be accessed by member (like aHASH),
+but items can also be accessed by the sorted order and values of the scores.
+
+Essentially speaking, the ZSETs looks like special SETs. The SETs structure
+maps a key (as STRING) to a collection of STRINGs. But the ZSETs structure
+maps a key (as STRING) to a collection of ITEMs which associated with a
+score. It allows us to fetch data by score.
+
+To store ZSETs data,  the scylla table is created by following CQL:
+
+```
+CREATE TABLE ZSETs (
+    pkey text,
+    ckey double,
+    data text,
+    PRIMARY KEY(pkey, ckey)
+) WITH ... ;
+```
+
+Like other stutures mentioned above, a ZSETs strucutre is stored as a
+partition within the ZSETs table.
+
+## 5. Implementation of Commands
+
+In Scylla, high write performance is achieved by ensuring that writes do
+not require reads from disk. However, in Redis, the object level atomic
+updates is [supported](https://redis.io/topics/transactions),  and the
+transaction is supported (e.g. CAS).  In this proposal, the commands, which
+require a read before the write (a.k.a read-modify-write), are implemented
+in an unsafe way, that simply performs a read before the write.
+
+Fortunately, the LWT is soon comming to Scylla.  By then the
+read-modify-write commands will be implemented based on LWT.
+
+**IMPORTANT NOTE**:
+
+The commands, that require a read before the write, are not atomic. It's
+anther difference with original Redis.
+
+### 5.1 RMW Command
+
+The RMW Commond is that the sequence of operations (Read-Modify-Write)
+require by this command needs to be performed atomically.
+
+For instance, APPEND command of STRINGs, is typical RMW command. First,
+read the key from database, modify it's value (based on the exists value),
+and write the new value back. These sequence of operations should be
+performed atomically.
+
+As mentioned above, Scylla currently does not support LWT. The RMW commands
+are implemented in an unsafe way in this proposal. It simply performs a
+read before the write.
+
+Asumming Scylla has supported the LWT, there is no difference with the
+original Reids about the RMW commands.
+
+### 5.2 Non-RMW Command
+
+The Non-RMW command only performs a single read or write operation. For
+instance, The SET, GET, EXISTS of STRINGs strucutures. The behaviour of
+these commands are not difference with the original Redis.
+
+### 5.3 Transaction
+
+For the same reason as RMW command, the transaction is not supported
+currently.
+
+> In fact, some commands (e.g. MSET of STRINGs) are also not supported,
+> which also need the transcation mechanism.
+
+### 5.4 Time to Live (TTL)
+
+In Redis, the TTL is only specified to entrie key-value pair. Scylla has
+compaction mechanism to remove expired data.  We can set the TTL the
+partition for the given partition key. We also can rewrite the  partition
+with new TTL value.
+
+The partition ( stores the key-value pair of Redis strucutres) will be
+deleted eventually, when the associated TTL is older than current
+timestamp.
+
+### 5.5 Consistency
+
+Redis Cluster is not able to [guarantee strong
+onsistency](https://redis.io/topics/cluster-tutorial).
+
+In Scylla, data is always replicated automatically.  Read  or  write
+operations can occur to data stored on any of the replicated nodes. The
+[Consistency Level
+(CL)](https://docs.scylladb.com/glossary/#term-consistency-level-cl)
+determines how many replicas in a cluster that must acknowledge read or
+[write
+operations](https://docs.scylladb.com/glossary/#term-write-operation)
+before it is considered successful.
+
+We can use the fault tolerance mechanism of Scylla, to make the
+consistency of Redis operations configurable.
+
+### 5.6 Implementation of Shared Commands
+
+In Redis, data strucutures have the shared commands (e.g. DEL, EXISTS).
+And other commands usually operate on special one data structure of Redis.
+As mentioned above,
+we split the key set into 5 different Scylla table. The shared commands
+should operate on these tables in parallel. For instance, DEL command,
+which allows user to remove any key in Redis database, should delete the
+given key from 5 different Scylla tables.
--- a/main.cc
+++ b/main.cc
@@ -70,6 +70,7 @@
 #include "cql3/cql_config.hh"
 
 #include "alternator/server.hh"
+#include "redis/service.hh"
 
 namespace fs = std::filesystem;
 
@@ -1103,6 +1104,12 @@ int main(int ac, char** av) {
                 alternator_server.init(addr, alternator_port, alternator_https_port, creds, cfg->alternator_enforce_authorization()).get();
             }
 
+            static redis_service redis;
+            if (cfg->enable_redis_protocol()) {
+                redis.init(std::ref(proxy), std::ref(db), std::ref(auth_service), *cfg).get();
+                startlog.info("Redis server listening on port {}", cfg->redis_transport_port());
+            }
+
             if (cfg->defragment_memory_on_idle()) {
                 smp::invoke_on_all([] () {
                     engine().set_idle_cpu_handler([] (reactor::work_waiting_on_reactor check_for_work) {
@@ -1122,6 +1129,7 @@ int main(int ac, char** av) {
             api::set_server_done(ctx).get();
             supervisor::notify("serving");
             // Register at_exit last, so that storage_service::drain_on_shutdown will be called first
+
             auto stop_repair = defer_with_log_on_error([] {
                 startlog.info("stopping repair");
                 repair_shutdown(service::get_local_storage_service().db()).get();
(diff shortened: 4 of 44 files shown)


redis/server.hh
```diff
@@ -0,0 +1,148 @@
...
+};
+
+class redis_server {
+    std::vector<server_socket> _listeners;
+    distributed<service::storage_proxy>& _proxy;
+    distributed<redis::query_processor>& _query_processor;
+    redis_server_config _config;
+    size_t _max_request_size;
+    semaphore _memory_available;
+    seastar::metrics::metric_groups _metrics;
```
amnonh: I see that you define a metric_gruup object, which is good, but I don't see any metrics defined.

Adding metrics can be in a follow-up patch, what is important is that you'll have a counter (i.e. ever-growing unsigned int) for each of your commands, so it can be monitored separately.

Also, I suggest monitoring the latencies of your read and write, take and store a timestamp at the beginning of each operation and use an estimated_histogram for latencies calculations.

fastio: Hi @amnonh , thanks for your suggestion about metrics.
Now, only some basic metrics are added.

asias:
- The commit message has extra space. Run `git log` and compare the commit message with other commits in the tree you will notice.

-  Commit 7e549dc1bff866949f820e7cc0899f2a7c3686fb and 1ef0f1b15c7bb65d8849902543352153a4be5634 can be merged together.

- Commit 4d67faeee85ca71e21d6fa7441d3201c890c1d84 is the only commit that adds a individual command. To make the review easier, you can organize the commits as 1) first introduce the infrastructure 2) then implement the individual redis commands 3) enable the redis option and wire it up.

```
commit 4d67faeee85ca71e21d6fa7441d3201c890c1d84 (fastio/master)
Author: Peng Jian <pengjian.uestc@gmail.com>
Date:   Fri Oct 4 19:45:48 2019 +0800

        ADD select command.
    
        process the exception when received wrong arguments.
    
        Signed-off-by: Peng Jian, <pengjian.uestc@gmail.com>

commit 7e549dc1bff866949f820e7cc0899f2a7c3686fb
Author: Peng Jian <pengjian.uestc@gmail.com>
Date:   Fri Oct 4 15:11:17 2019 +0800

        Redis API in Scylla
    
        Remove unnecessary logs.
        Add metrics for for Redis API.
    
        Signed-off-by: Peng Jian, <pengjian.uestc@gmail.com>

commit 1ef0f1b15c7bb65d8849902543352153a4be5634
Author: Peng Jian <pengjian.uestc@gmail.com>
Date:   Wed Oct 2 17:29:42 2019 +0800

        Redis API in Scylla
    
        Scylla has advantage and amazing features. If Redis build on the top of Scylla,
        it has the above features automatically. It's achived great progress
        in cluster master managment, data persistence, failover and replication.
    
        The benefits to the users are easy to use and develop in their production
        environment, and taking avantages of Scylla.
    
        Using the Ragel to parse the Redis request, server abtains the command name
        and the parameters from the request, invokes the Scylla's internal API to
        read and write the data, then replies to client.
    
        Signed-off-by: Peng Jian, <pengjian.uestc@gmail.com>
```

redis/server.hh
```diff
@@ -0,0 +1,156 @@
...
+private:
+    uint64_t _connects = 0;
+    uint64_t _connections = 0;
```
amoskong: Is it repeated with _total_connections & _current_connections in line 145?

```
    uint64_t _total_connections = 0;
    uint64_t _current_connections = 0;
```

amoskong: BTW, it's difficult to know the difference between those two variables from the name.

main.cc
```diff
@@ -1089,6 +1092,17 @@ int main(int ac, char** av) {
                 startlog.info("Alternator server listening on {} port {}", addr, cfg->alternator_port());
             }
 
+            if (cfg->enable_redis_protocol()) {
+                supervisor::notify("starting create redis databases if not exists.");
+                redis::create_keyspace_if_not_exists(cfg).get();
```
amoskong: alternator has a similar method `executor::maybe_create_keyspace()` in alternator/executor.cc

And that's included in start() stage.

redis/server.hh
fastio: total_connections, means that this is an incremental counter, while current_connections means that current active client connections counter which increased when connection is connected, reduced when the connection is closed.

main.cc
fastio: Thanks. I will read the code of executor::maybe_create_keyspace. Maybe it 's useful for me.

fastio: Thanks for your suggestion.

redis/server.cc
```diff
@@ -0,0 +1,259 @@
...
+    size_t nr_total = _listeners.size();
+    logging.debug("redis_server: abort accept nr_total={}", nr_total);
+    for (auto&& l : _listeners) {
+        l.abort_accept();
+        logging.debug("redis_server: abort accept {} out of {} done", ++nr, nr_total);
+    }
+    auto nr_conn = make_lw_shared<size_t>(0);
+    auto nr_conn_total = _connections_list.size();
+    logging.debug("redis_server: shutdown connection nr_total={}", nr_conn_total);
+    return parallel_for_each(_connections_list.begin(), _connections_list.end(), [nr_conn, nr_conn_total] (auto&& c) {
```
nyh: We can fix this later, but it's probably not a great idea to do something in parallel for a a hue number of connections

redis/server.cc
```diff
@@ -0,0 +1,259 @@
...
+            return stop_iteration::no;
+        }).handle_exception([] (auto ep) {
+            logging.debug("accept failed: {}", ep);
+            return stop_iteration::no;
+        });
+    });
+}
+
+future<redis_server::result> redis_server::connection::process_request_one(redis::request&& request, redis::redis_options& opts, service_permit permit) {
+    return futurize_apply([this, request = std::move(request), &opts, permit] () mutable {
```
nyh: Why do you need futurize_apply and a lambda here? I'm not sure I understand what you're trying to do...

redis/server.cc
```diff
@@ -0,0 +1,259 @@
...
+    }, [this] {
+        return with_gate(_pending_requests_gate, [this] {
+            return process_request();
+        });
+    }).then_wrapped([this] (future<> f) {
+        try {
+            f.get();
+        }
+        catch (redis_exception& e) {
+        logging.info(" redis exception = {} ", e.what());
```
nyh: please indent this line in

redis/server.cc
```diff
@@ -0,0 +1,259 @@
...
+            write_reply(e);
+        }
+        catch (std::exception& e) {
+        logging.info(" std::exception = {} ", e.what());
```
nyh: again indent

redis/server.cc
```diff
@@ -0,0 +1,259 @@
...
+            write_reply(redis_exception { e.what() });
+        }
+        catch (...) {
+        logging.info(" ... exception ");
```
nyh: also indent here

redis/server.cc
```diff
@@ -0,0 +1,259 @@
...
+
+future<redis_server::result> redis_server::connection::process_request_internal() {
+    return _process_request_stage(this, std::move(_parser.get_request()), seastar::ref(_options), empty_service_permit());
+}
+
+void redis_server::connection::write_reply(const redis_exception& e)
+{
+    _ready_to_respond = _ready_to_respond.then([this, exception_message = e.what_message()] () mutable {
+        return redis_message::make_exception(exception_message).then([this] (auto&& result) {
+            auto message = result.message();
```
nyh: Why do you need to create a ready future only to extract the message from it? Or what are you trying to do here?

redis/server.cc
```diff
@@ -0,0 +1,259 @@
...
+    _parser.init();
+    return _read_buf.consume(_parser).then([this] {
+        ++_server._requests_served;
+        ++_server._requests_serving;
+        _pending_requests_gate.enter();
+        auto leave = defer([this] { _pending_requests_gate.leave(); });
+        return process_request_internal().then([this, leave = std::move(leave)] (auto&& result) mutable {
+            --_server._requests_serving;
+            auto message = result.make_message();
+            return _write_buf.write(std::move(*message)).then([this] {
```
nyh: I'm not sure I understood the purpose of the pending_request_gate - don't you need to also put "leave" into this continuation so the gate keeps getting taken, so we don't kill the request before we finished writing its response? Or is that fine?

redis/server.hh
nyh: By the way, you can see examples of metrics similar to what you may want to add in Alternator - alternator/stats.{cc,hh}

service/storage_service.cc
```diff
@@ -3554,5 +3555,108 @@ db::schema_features storage_service::cluster_schema_features() const {
     return f;
 }
 
-} // namespace service
+future<> storage_service::start_redis_transport(distributed<redis::query_processor>& qp) {
```
nyh: Can this code please moved to one of the source files in redis/? Maybe redis/server.cc or something?
I'd really like to see most of the redis-related code in redis/, most of the DynamoDB-specific code in alternator/, etc. storage_service.cc is already too huge.

redis/server.cc
```diff
@@ -50,6 +50,23 @@ redis_server::redis_server(distributed<service::storage_proxy>& proxy, distribut
     , _memory_available(_max_request_size)
     , _auth_service(auth_service)
 {
```
nyh: Please use a different title for each patch. If this patch is about statistics, title it something like "Redis API: add statistics".

redis/server.cc
```diff
@@ -50,6 +50,23 @@ redis_server::redis_server(distributed<service::storage_proxy>& proxy, distribut
...
+    namespace sm = seastar::metrics;
+
+    _metrics.add_group("redis_transport", {
```
nyh: If this patch is about adding statistics, maybe you can move log-related stuff from the previous huge patch, to this one.
Not critical, though.

redis/server.cc
```diff
@@ -163,15 +180,12 @@ future<> redis_server::connection::process()
             f.get();
         }
         catch (redis_exception& e) {
-        logging.info(" redis exception = {} ", e.what());
```
nyh: You can drop these lines from the original patch - you don't need to add them in patch 2 and drop them in patch 3.

main.cc
fastio: It's different from the case of Alternator. It's better to create the keyspaces/tables in the start stage before the server serving the requests.

docs/redis/redis.md
```diff
@@ -0,0 +1,360 @@
...
+
+Redis is a very  famous NoSQL in-memory data structure store that stores a
+mapping of keys to different types of values, and familiar to many developers.
+It supports data structures such as STRINGs, HASHes, LISTs, SETs, ZSETs(sorted sets),
+and other data structures. Redis has built-in replication, Lua
+scripting, LRU eviction, transactions and different levels of on-disk
+persistence. Now Redis as memory store service is provided by many cloud
+platforms.
+
+**IMPORTANT NOTE**: As we known, Redis owns many useful features.  Storing the
```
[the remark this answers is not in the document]
fastio: Sorry, this `IMPORTANT NOT` is not necessary here. And it should be removed.
What I want to say is that suporting the Redis API for storing data is enough to many users.
Of course, we can support all of the Redis API.

docs/redis/redis.md
```diff
@@ -0,0 +1,360 @@
...
+top of Scylla is provided. At the beginning, the main feature, which is as a
+data structures store, is design and  implmented. In the future, the reset
+features of Redis will be supported.
+
+## 2. Motivation
+
+In contrast,  [Scylla](https://www.scylladb.com/) has advantage and amazing
+features:
+
+* Low and consistent latency
```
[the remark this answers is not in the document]
fastio: Yes, persistence is the very important.
I will update it.

docs/redis/redis.md
```diff
@@ -0,0 +1,360 @@
...
+Redis clients communicate with the Redis server using a protocol called
+**RESP** (REdis Serialization Protocol). The protocol is binary-safe and
+easy to be implement at client side, which was designed specifically for
+Redis.
+
+Redis server accepts commands composed of different arguments. Once a
+command is received, it's processed and a reply is sent back to the client.
+Two kind of Redis clients are widely used, the smart Redis client and
+simple Redis client. When using smart Redis client to connect the Redis
+cluster, it will send the commaonds to the right Redis server. Because the
```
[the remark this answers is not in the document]
fastio: Got it.

docs/redis/redis.md
```diff
@@ -0,0 +1,360 @@
...
+
+In Scylla terminology, the node receiving the request acts as the the
+coordinator, and often passes the request on to one or more other nodes,
+which hold copies of the requested data.  Any node of Scylla cluster can
+process the request correctly. In other words, Client can send request to
+any node of Redis cluster built on the top of Scylla. Obviously, the simple
+client can accesses the Redis cluster built on the top of Scylla, when
+provided address is not an IP but rather a single domain name, which a
+client resloves a random Scylla node IP address. The smart Redis client
+will query the mapping information between hash slots and Server nodes. We
```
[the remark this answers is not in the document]
fastio: This is a syntax error.  Actually, I have implemented the command named 'CLUSTER SLOTS' in the peids project.  The redis client uses this command to query the mapping of hash slots and server nodes.
I'll fix this.

docs/redis/redis.md
```diff
@@ -0,0 +1,360 @@
...
+just fill the mapping with random Scylla node IP address.
+
+With Redis enabled on port 6379 (which is configurable), every Scylla node
```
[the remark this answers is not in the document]
fastio: OK. I agree it.

service/storage_service.cc
fastio: Yes, of course.

redis/server.cc
fastio: OK. I got it.

redis/server.cc
fastio: OK.

redis/server.cc
fastio: Got it.

nyh:
@fastio  I have another request.

It's fairly obvious that while you were developing this code, you needed to test its functionality - test that the protocol is actually parsed correctly, that responses and errors are generated in a way that Redis clients understand them, that writing data and reading it back actually works - and so on. You must have done such tests yourself, either manually or programatically in some way. I would really like to see automated test that check these things.

These test will be super-important in the future when you, or somebody else, modifies the Redis-support code or even general Scylla code and want to verify that we didn't break anything in Redis support. 
Though you can argue that we can add such tests later, I believe that we need to add at least a seed for them - even if just a small amount of test cases - now - for this seed to grow later. If we don't start adding such tests now, it is unlikely we'll ever have them.

In the current Scylla codebase, you can find two example of how we did such functional tests and you can use one of them - or propose another one if you had something else in mind. The older approach you can see in `tests/cql_query_test.cc`. This is C++ code linked with Scylla (so test executables are pretty huge, unfortunately), a test quickly starts one-shard Scylla server - and then does all sorts of tests on it using various undocumented C++-code hooks into Scylla. The second, newer, approach is used in testing Alternator (the DynamoDB API) which you can find in the `alternator-test/` directory: This approach has the user starting Scylla manually, and then test code written in any language (in `alternator-test` we used Python)  use a standard client library to connect to Scylla via the API being tested (in `alternator-test`, we used Amazon's "boto3" library which connects to DynamoDB). 

In my opinion, the second approach has two important benefits over the first which is why I chose it for Alternator. First it makes writing tests very easy - almost fun - which encourages better test coverage. Tests are easy to write Python code, don't need lengthy compilation, and the pytest framework we used for Python tests make it very easy to manage the tests, understand where they failed, etc. Second, we can run the same tests against **both** Scylla and the server we're trying to emulate (DynamoDB or Redis). This is important, if you want to make sure that Scylla's behavior not only does not regress, it was actually right in the first place. 

We actually have a third testing approach in Scylla, called "dtest", which also has tests in Python, but these tests start their own Scylla clusters. However, these tests are so incredibly slow (since each small test starts a separate Scylla cluster) that I would recommend against this third approach.

fastio:
@avikivity Yes, you are right.  All the commands in redis received by one redis-server  are exectued sequentially.

Perhaps we should serialize accesses to the same key, but allow accesses to different keys to run in parallel.
-------------
I agree with you. We can submit the requests with the same key to the same shard by the hash of key.

fastio:
@nyh Thank you for your advice. The codes should be tested. And it's very important for production system. Actually, there are several test cases in the [pedis project](https://github.com/fastio/pedis/blob/master/tests/redis/kv_test.cc) . I agree with you to use the second approach to write test cases for Redis API.

main.cc
```diff
@@ -1121,12 +1113,6 @@ int main(int ac, char** av) {
             api::set_server_done(ctx).get();
             supervisor::notify("serving");
             // Register at_exit last, so that storage_service::drain_on_shutdown will be called first
-            auto stop_redis_query_processor = defer([&redis_query_processor, cfg] {
```
nyh: How is this destruction code not needed any more?

fastio:
Change logs:

*  Add redis/service.{hh, cc}:  move the Redis transport related codes from service/storage_service.{hh,cc} to redis/service.{hh, cc}.
*  Add redis/stat.{hh,cc}: implement the basic metrics for Redis API.
*  Add the test cases:  the test codes written in Python using the  redis-py library.
*  Fix the reply of the DEL command.
*  Fix some errors and incorrect description in the doc/redis/redis.md.

@nyh . Thanks for review.

main.cc
fastio: The shutdown hook for redis is added now.

fastio:
Change logs:
* Add the shutdown hook for redis to stop redis service.

redis/mutation_utils.cc
```diff
@@ -0,0 +1,93 @@
...
+    return proxy.mutate(std::vector<mutation> {std::move(m)}, write_consistency_level, timeout, nullptr, permit);
+}
+
+
+mutation make_tombstone(service::storage_proxy& proxy, const redis_options& options, const sstring& cf_name, const bytes& key) {
+    auto schema = get_schema(proxy, options.get_keyspace_name(), cf_name);
+    auto pkey = partition_key::from_single_value(*schema, key);
+    auto m = mutation(schema, std::move(pkey));
+    m.partition().apply(tombstone { api::new_timestamp(), gc_clock::now() }); 
+    return std::move(m);
```
amoskong:
```
FAILED: build/release/redis/mutation_utils.o 
g++ -MD -MT build/release/redis/mutation_utils.o -MF build/release/redis/mutation_utils.o.d -I/home/amos/scylladb.com/scylla-build-example/scylla-redis/seastar/include -I/home/amos/scylladb.com/scylla-build-example/scylla-redis/build/release/seastar/gen/include -std=gnu++17 -U_FORTIFY_SOURCE -Wno-maybe-uninitialized -march=westmere -Werror=unused-result -fconcepts -DSEASTAR_API_LEVEL=2 -DSEASTAR_HAVE_GCC6_CONCEPTS -DSEASTAR_USE_STD_OPTIONAL_VARIANT_STRINGVIEW -DSEASTAR_HAVE_DPDK -DFMT_SHARED -I/usr/include/p11-kit-1 -ffile-prefix-map=/home/amos/scylladb.com/scylla-build-example/scylla-redis=.  -march=westmere -DBOOST_TEST_DYN_LINK -DTHRIFT_USES_BOOST  -fvisibility=hidden -Wall -Werror -Wno-maybe-uninitialized -Wno-tautological-compare -Wno-missing-braces -Wno-misleading-indentation -Wno-overflow -Wno-noexcept-type -Wno-nonnull-compare -Wno-error=cpp -Wno-ignored-attributes -Wno-overloaded-virtual -Wno-stringop-overflow -Wno-error=deprecated-declarations -DXXH_PRIVATE_API -DSEASTAR_TESTING_MAIN -DHAVE_LIBSYSTEMD=1 -DHAVE_LZ4_COMPRESS_DEFAULT -O3 --param inline-unit-growth=300 -gz  -g -I. -I build/release/gen  -c -o build/release/redis/mutation_utils.o redis/mutation_utils.cc
redis/mutation_utils.cc: In function 'mutation redis::make_mutation(service::storage_proxy&, const redis::redis_options&, bytes&&, bytes&&, long int)':
redis/mutation_utils.cc:61:21: error: moving a local object in a return statement prevents copy elision [-Werror=pessimizing-move]
   61 |     return std::move(m);
      |            ~~~~~~~~~^~~
redis/mutation_utils.cc:61:21: note: remove 'std::move' call
redis/mutation_utils.cc: In function 'mutation redis::make_tombstone(service::storage_proxy&, const redis::redis_options&, const sstring&, const bytes&)':
redis/mutation_utils.cc:77:21: error: moving a local object in a return statement prevents copy elision [-Werror=pessimizing-move]
   77 |     return std::move(m);
      |            ~~~~~~~~~^~~
redis/mutation_utils.cc:77:21: note: remove 'std::move' call
cc1plus: all warnings being treated as errors
```

redis/protocol_parser.rl
```diff
@@ -0,0 +1,141 @@
...
+}
+
+postpop {
+    postpop();
+}
+
+}%%
+
+class redis_protocol_parser : public ragel_parser_base<redis_protocol_parser> {
+    %% write data nofinal noprefix;
```
amoskong:
```
FAILED: build/release/redis/service.o 
g++ -MD -MT build/release/redis/service.o -MF build/release/redis/service.o.d -I/home/amos/scylladb.com/scylla-build-example/scylla-redis/seastar/include -I/home/amos/scylladb.com/scylla-build-example/scylla-redis/build/release/seastar/gen/include -std=gnu++17 -U_FORTIFY_SOURCE -Wno-maybe-uninitialized -march=westmere -Werror=unused-result -fconcepts -DSEASTAR_API_LEVEL=2 -DSEASTAR_HAVE_GCC6_CONCEPTS -DSEASTAR_USE_STD_OPTIONAL_VARIANT_STRINGVIEW -DSEASTAR_HAVE_DPDK -DFMT_SHARED -I/usr/include/p11-kit-1 -ffile-prefix-map=/home/amos/scylladb.com/scylla-build-example/scylla-redis=.  -march=westmere -DBOOST_TEST_DYN_LINK -DTHRIFT_USES_BOOST  -fvisibility=hidden -Wall -Werror -Wno-maybe-uninitialized -Wno-tautological-compare -Wno-missing-braces -Wno-misleading-indentation -Wno-overflow -Wno-noexcept-type -Wno-nonnull-compare -Wno-error=cpp -Wno-ignored-attributes -Wno-overloaded-virtual -Wno-stringop-overflow -Wno-error=deprecated-declarations -DXXH_PRIVATE_API -DSEASTAR_TESTING_MAIN -DHAVE_LIBSYSTEMD=1 -DHAVE_LZ4_COMPRESS_DEFAULT -O3 --param inline-unit-growth=300 -gz  -g -I. -I build/release/gen  -c -o build/release/redis/service.o redis/service.cc
In file included from ./redis/server.hh:39,
                 from redis/service.cc:24:
build/release/gen/redis/protocol_parser.hh:43:20: error: in-class initialization of static data member 'const char redis_protocol_parser::_nfa_targs []' of incomplete type
   43 |  static const char _nfa_targs[] = {
      |                    ^~~~~~~~~~
build/release/gen/redis/protocol_parser.hh:47:20: error: in-class initialization of static data member 'const char redis_protocol_parser::_nfa_offsets []' of incomplete type
   47 |  static const char _nfa_offsets[] = {
      |                    ^~~~~~~~~~~~
build/release/gen/redis/protocol_parser.hh:53:20: error: in-class initialization of static data member 'const char redis_protocol_parser::_nfa_push_actions []' of incomplete type
   53 |  static const char _nfa_push_actions[] = {
      |                    ^~~~~~~~~~~~~~~~~
build/release/gen/redis/protocol_parser.hh:57:20: error: in-class initialization of static data member 'const char redis_protocol_parser::_nfa_pop_trans []' of incomplete type
   57 |  static const char _nfa_pop_trans[] = {
      |                    ^~~~~~~~~~~~~~
```

fastio: Hi @amoskong , what's the version of ragel which you used?

I use the ragel with version 6.8, and there are no errors as you mentioned.

> Ragel State Machine Compiler version 6.8 Feb 2013
> Copyright (c) 2001-2009 by Adrian Thurston

redis/mutation_utils.cc
fastio: Thanks. The std::move is unnecessary. I'll fix it.

fastio:
Change logs:

* Fix the warnings being treated as errors: remove unnecessary std::move.

redis/protocol_parser.rl
amoskong: I built by `./tools/toolchain/dbuild ./reloc/build_reloc.sh`.

[scylla-redis]$ ./tools/toolchain/dbuild -it -- rpm -q ragel
ragel-7.0.0.11-3.fc30.x86_64

fastio: I found that the same problem was reported in this issue: https://github.com/scylladb/seastar/issues/296

fastio: I change the ragel to latest version (Ragel State Machine Compiler version 7.0.0.12 May 2019
). The errors are still exists.
But the ragel with version 6.8 works fine.

Can we change the ragel version to 6.8 ?

redis/reply.hh
```diff
@@ -0,0 +1,105 @@
...
+        write_bytes(m, result);
+        return make_ready_future<redis_message>(m);
+    }
+    static future<redis_message> unknown(const bytes& name) {
+        return from_exception(make_message("-ERR unknown command '%s'\r\n", to_sstring(name)));
+    }
+    static future<redis_message> exception(const redis_exception& e) {
+        return from_exception(make_message("-ERR %s\r\n", e.what_message()));
+    }
+    static future<redis_message> make_exception(const sstring& em) {
```
amoskong: It's not making exception, but make a redis message from exception message.

How about this? then we don't need from_exception(), we can consider exception() is one kind of redis_message.

```
    static future<redis_message> exception(const sstring& em) {
        auto m = make_lw_shared<scattered_message<char>> ();
        m->append(make_message("-ERR %s\r\n", em));
        return make_ready_future<redis_message>(m);
    }
    static future<redis_message> exception(const redis_exception& e) {
        return exception(e.what_message());
    }
```

redis/reply.hh
```diff
@@ -0,0 +1,105 @@
...
+        m->append_static(":0\r\n");
+        return make_ready_future<redis_message>(m);
+    }
+    static future<redis_message> one() {
+        auto m = make_lw_shared<scattered_message<char>> ();
+        m->append_static(":1\r\n");
+        return make_ready_future<redis_message>(m);
+    }
+    static future<redis_message> zero() {
+        return err();
```
amoskong: Calling zero() in err() is better. BTW err message is a general message, there are many kind of error, not just return zero.

```
    static future<redis_message> zero() {
        auto m = make_lw_shared<scattered_message<char>> ();
        m->append_static(":0\r\n");
        return make_ready_future<redis_message>(m);
    }
    static future<redis_message> err() {
        return zero();
    }
```

redis/commands/get.cc
```diff
@@ -0,0 +1,56 @@
...
+    return seastar::make_shared<get> (std::move(req._command), std::move(req._args[0]));
+}
+
+future<redis_message> get::execute(service::storage_proxy& proxy, redis::redis_options& options, service_permit permit)
+{
+    return redis::read_strings(proxy, options, _key, permit).then([] (auto result) {
+        if (result->has_result()) {
+            return redis_message::make_strings_result(std::move(result->result()));
+        }
+        return redis_message::err();
```
amoskong: Currently the err() will return zero  `":0\r\n"`.

> Quote from https://redis.io/commands/get
```
Return value
Bulk string reply: the value of key, or nil when key does not exist.
```

The correct response message should be `"$-1\r\n"`

> Referenced https://redis.io/topics/protocol:
```
RESP Bulk Strings can also be used in order to signal non-existence of a value using a special format that is used to represent a Null value. In this special format the length is -1, and there is no data, so a Null is represented as:

"$-1\r\n"
This is called a Null Bulk String.

The client library API should not return an empty string, but a nil object, when the server replies with a Null Bulk String. For example a Ruby library should return 'nil' while a C library should return NULL (or set a special flag in the reply object), and so forth.
```

redis/commands/del.cc
```diff
@@ -0,0 +1,53 @@
...
+    if (req.arguments_size() != 1) {
+        throw wrong_arguments_exception(1, req.arguments_size(), req._command);
+    }
+    return seastar::make_shared<del> (std::move(req._command), std::move(req._args[0]));
+}
+
+future<redis_message> del::execute(service::storage_proxy& proxy, redis::redis_options& options, service_permit permit)
+{
+    return redis::delete_object(proxy, options, std::move(_key), permit).then([] {
+       return redis_message::one();
```
amoskong: Currently your implement only supports to delete one key once.
Redis supports to delete multiple keys once: https://redis.io/commands/del

It returns the number of deleted keys, not always to be one: `":1\r\n"`

redis/commands/del.hh
```diff
@@ -0,0 +1,44 @@
...
+
+#include "redis/request.hh"
+#include "redis/abstract_command.hh"
+
+namespace redis {
+
+namespace commands {
+
+class del : public abstract_command {
+    bytes _key;
```
amoskong: std::vector<bytes> _keys;

DEL command supports to delete multiple keys once.

redis/keyspace_utils.cc
```diff
@@ -0,0 +1,210 @@
...
+    auto table_gen = [] (sstring ks_name, sstring cf_name, schema_ptr schema) {
+        auto& proxy = service::get_local_storage_proxy();
+        if (proxy.get_db().local().has_schema(ks_name, cf_name)) {
+            return make_ready_future<>();
+        }
+        logger.info("Create keyspace: {}, table: {} for redis.", ks_name, cf_name);
+        return service::get_local_migration_manager().announce_new_column_family(schema, false);
+    };
+    // create 16 default database for redis.
+    return parallel_for_each(boost::irange<unsigned>(0, 16), [keyspace_gen = std::move(keyspace_gen), table_gen = std::move(table_gen)] (auto c) {
```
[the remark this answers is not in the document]
amoskong: SELECT command is supported, it would switch between keyspaces. The limitation is got from configuration.

So if we config redis_default_database_count to be larger than 16, SELECT allows to switch to unexisting keyspace, and it will fail.

docs/redis/redis.md
```diff
@@ -0,0 +1,360 @@
...
+Out of box, every Redis cluster supports 16 databases. The default database
+is 0 but it allows us to change that to any number from 0-15 (and allow us
+to configure Redis to support more databases). Each database provides
+a distinct keyspace, independent from the others. Use `SELECT n`
+to change databases. Redis allows us to store keys that map to any one of
+five different  basic data structure types:  STRINGs,  LISTs,  SETs,
+HASHes, and  ZSETs. (In fact, Now Redis has support other structure types,
+but the basic structure types are used widely). 
+
+In this proposal, We use the column family of Scylla to simulate the
```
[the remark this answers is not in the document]
amoskong: What is really used is the keyspace in Scylla.

----

Quote from https://docs.scylladb.com/glossary/:

```
Keyspace
A collection of tables with attributes which define how data is replicated on nodes. 

Column Family
See table.

Table
A collection of columns fetched by row. Columns are ordered by Clustering Key.
```

docs/redis/redis.md
```diff
@@ -0,0 +1,360 @@
...
+database within Redis. At the bootrap phase, the column familes with fixed
+name should be created (if not exists) or loaded. 
+
+We known that, Redis allows us to store keys that map to any one of the
+basic data structures. Scylla supports variant type, which allows us to
```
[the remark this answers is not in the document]
amoskong: @fastio the user-defined type can implement this:

```
create type redis_type (v0 text, v1 list<text>, v2 map<text, text>);
create table redis (a int, b frozen<redis_type>, primary key ((a), b));
insert into redis (a, b) values (1, ('scylla', ['l1', 'l2'], {'k1': 'v1', 'k2': 'v2'}));
cqlsh:ks1> SELECT * from redis ;

 a | b
---+----------------------------------------------------------------
 1 | {v0: 'scylla', v1: ['l1', 'l2'], v2: {'k1': 'v1', 'k2': 'v2'}}

(1 rows)
cqlsh:ks1>
```

docs/redis/redis.md
```diff
@@ -0,0 +1,360 @@
...
+define one table to store all kind of these Redis data. For example ,we
+create the table with the schame as follow:
+
+```
+CREATE TABLE redis （
+    key text,
+    value variant\<text, list\<text\>, map\<text\>\>,
```
amoskong: The map only has one item?

redis/reply.hh
fastio: I agree with you. I'll change it.

redis/reply.hh
fastio: Yes. It's better to call zeor() in err().

redis/commands/get.cc
fastio: Yes, you are right. Here we should returns nil string if key does not exist.

redis/commands/del.cc
amoskong: zero `":0\r\n"` should be returned if the key doesn't exist.

redis/protocol_parser.rl
amoskong: > I change the ragel to latest version (Ragel State Machine Compiler version 7.0.0.12 May 2019). The errors are still exists. But the ragel with version 6.8 works fine.
> 
> Can we change the ragel version to 6.8 ?

/CC @avikivity ?

redis/commands/del.cc
fastio: It's easy to delete multi keys. 
In fact, we should determine that how many keys should be actually deleted. This is a read-before-write operation. As we known that Scylla does not support LWT. So should we just read all keys, then delete the exist keys?

docs/redis/redis.md
fastio: No. This is one of the optional sulotion. As mentioned in the redis.md, this sulotion is not the not better one.

redis/protocol_parser.rl
amoskong: The building completed without error after downgraded ragel to 6.8-5

#### The downgrade command:
```
sudo dnf downgrade -y https://rpmfind.net/linux/mageia/distrib/5/x86_64/media/core/release/ragel-6.8-5.mga5.x86_64.rpm
```

#### New docker image with ragel-6.8-5
```
diff --git a/tools/toolchain/image b/tools/toolchain/image
index 3d89f8f..2e9992e 100644
--- a/tools/toolchain/image
+++ b/tools/toolchain/image
@@ -1 +1 @@
-docker.io/scylladb/scylla-toolchain:fedora-30-20190911
+docker.io/amoskong/scylla-toolchain:fedora-30-ragel6.8
```

nyh:
If we continue to improve this patch set until it absolutely perfect, I think it will be a very long process, and very frustrating for you @fastio to keep rebasing it and maintaining it.

So I want to do something different. I want to commit this work **this week**; Get this basic support in, and then later you can send incremental patches improving the Redis support. For this, I would like you to please conflate the newer patches in this patch set into the older ones - i.e., I don't want to see a patch like "Redis API: Move the redis transport related codes to redis/service" - I just want the original patch to put it right in the right place immediately. I then want to do one final review over this code which will focus on verifying just one thing: That most of the Redis code is in the redis/ directory, and the amount of changes to other source files is minimal. The reason why I wanted to reach this state (and now to verify we've indeed reached it) is two-fold:

1. First it reduces the **risks** of including the new Redis code now - even before it is perfect. If it's very clear that Redis cannot possibly do any harm to users not using it - because the patches don't touch any of the existing Scylla code (or very little of it) -  then there is no risk of including the new Redis support now, even before it's 100% ready.

2. Second, having separate source files will make it easier - and more fun - for you to continue working on improving this Redis support incrementally. This is because your patches will only (or mostly) go to new redis/ files, so the risk that you'll clash with other people's changes will be lower.

Amos has made some good comments in this thread recently, so it would be nice if you could also take them into account for the next (and hopefully final) iteration, but you don't have to fix all of them - you can also create "FIXME"s in the code - and/or bug tracker issues - to remember that these issues exist without fixing them yet.

By the way, the API issues that Amos discovered by running the tests against the "real" redis is exactly why I suggested that we write tests that use the Redis API (and not some internal C++ code) so that they can be run against both our code and against Redis.

redis/protocol_parser.rl
nyh: I don't follow the details of this discussion (I'm not at all familiar with Ragel or what changed in its versions), but please don't make your code rely on an antique version of Ragel. My build machine has ragel 7.0.0.11 and I don't have a problem building Seastar of Scylla. The new code needs to work also on the modern Ragel 7.0 - it can't just work on Ragel 6.8. That's not a viable workaround, in my opinion.

asias:
> If we continue to improve this patch set until it absolutely perfect, I think it will be a very long process, and very frustrating for you @fastio to keep rebasing it and maintaining it.
> 
> So I want to do something different. I want to commit this work **this week**; Get this basic support in, and then later you can send incremental patches improving the Redis support. 

Nadav, I really liked this approach. 

> For this, I would like you to please conflate the newer patches in this patch set into the older ones - i.e., I don't want to see a patch like "Redis API: Move the redis transport related codes to redis/service" - I just want the original patch to put it right in the right place immediately. I then want to do one final review over this code which will focus on verifying just one thing: That most of the Redis code is in the redis/ directory, and the amount of changes to other source files is minimal. The reason why I wanted to reach this state (and now to verify we've indeed reached it) is two-fold:
> 
>     1. First it reduces the **risks** of including the new Redis code now - even before it is perfect. If it's very clear that Redis cannot possibly do any harm to users not using it - because the patches don't touch any of the existing Scylla code (or very little of it) -  then there is no risk of including the new Redis support now, even before it's 100% ready.
> 
>     2. Second, having separate source files will make it easier - and more fun - for you to continue working on improving this Redis support incrementally. This is because your patches will only (or mostly) go to new redis/ files, so the risk that you'll clash with other people's changes will be lower.
> 
> 
> Amos has made some good comments in this thread recently, so it would be nice if you could also take them into account for the next (and hopefully final) iteration, but you don't have to fix all of them - you can also create "FIXME"s in the code - and/or bug tracker issues - to remember that these issues exist without fixing them yet.
> 
> By the way, the API issues that Amos discovered by running the tests against the "real" redis is exactly why I suggested that we write tests that use the Redis API (and not some internal C++ code) so that they can be run against both our code and against Redis.

redis/commands/del.cc
amoskong: > It's easy to delete multi keys.
> In fact, we should determine that how many keys should be actually deleted. This is a read-before-write operation. As we known that Scylla does not support LWT. So should we just read all keys, then delete the exist keys?

@psarna , @nyh can you help to answer @fastio's question?

@fastio , Scylla had supported lwt: https://github.com/scylladb/scylla/issues/1359

nyh: This is an interesting question. Indeed Scylla's classic MO is to do writes **without read**, so you can delete a key without knowing whether it actually existed before or it did not. Redis's DEL documentation also says that it "Removes the specified keys. A key is ignored if it does not exist.", but it is indeed sad that the API also returns the number of keys actually removed. I don't know if users actually use this capability often, but sadly it is officially documented and we'll eventually need to support it. 

But we don't need to support it **now**. What I suggest we do now is to remember that we still have this problem - I suggest doing two things: First have an "xfailing" test for this case (a test deleting a non-existent key expecting the returned integer to be 0) which will succeed on real redis, but fail on Scylla. Second is to create a bug-tracker issue to remember this.

For multiple keys, again, if you don't want to fix this issue immediately, please add an xfailing test and an issue to remember this missing feature. Or if it's trivial, just fix that now.

Eventually, we'll need to decide how to do all sorts of RMW (read-modify-write aka read-before-write) operations that Redis has. Maybe we'll use LWT, maybe some sort of leader model or Raft, or I don't know what; Scylla's DynamoDB support ("Alternator") has the same problems. But we can start without this for now. We just need to remember what we don't support yet, and in my opinion xfailing tests (which succeed on the real redis) is a good way to remember that.

amoskong:
The latest PR only changed configure.py (for building) and db/config.* (for adding new options), and very small & clean change for main.cc for starting/stopping redis service.

The patches are organized a little bit messy ;-) @fastio you can reference my example:
https://github.com/amoskong/scylla/commits/redis-patchset

- [PATCH 1/5] Document: add docs/redis/redis.md
- [PATCH 2/5] redis: Redis API in Scylla
- [PATCH 3/5] Redis API: graft redis module to Scylla
- [PATCH 4/5] redis-test: add test cases for Redis API
- [PATCH 5/5] toolchain/Docker: downgrade ragel to 6.8-5

BTW: there are many trailing whitespace in your code, it should be cleaned.

nyh:
Yes, @amoskong it's hard to see which files changed when one patch changes one file and a later patch moves it to another file. This can probably be easily cleaned up (like you did). I would really appreciate not having that "toolchain/Docker: downgrade ragel to 6.8-5" patch, though. I want to be able to build with the ragel I have (7.0) - the toolchain should not be mandatory for developers using modern distributions. I'm sure this can be fixed without downgrading Ragel (but I have no idea how - I know nothing about Ragel).

amoskong:
toolchain/dbuild is used for building master packages by our daily jenkins job, the master building will be broken if we don't downgrade ragel to 6.8.

We can revert the patch when if ragel issue is fixed.

nyh:
As much as I care about the daily Jenkins job, I also want it to build on **my** development machine - which uses Ragel 7.0.0.11 (on Fedora 30) and I don't plan to downgrade....

 Is this really something we cannot work around in Ragel 7.0 without downgrading? And why doesn't it cause problems in Seastar as well (yes, I see there's an open issue, but I didn't see any problems on my own machine which uses Ragel 7.0, so what am I missing?)

amoskong:
> As much as I care about the daily Jenkins job, I also want it to build on **my** development machine - which uses Ragel 7.0.0.11 (on Fedora 30) and I don't plan to downgrade....
> 
> Is this really something we cannot work around in Ragel 7.0 without downgrading? And why doesn't it cause problems in Seastar as well (yes, I see there's an open issue, but I didn't see any problems on my own machine which uses Ragel 7.0, so what am I missing?)

Did you successfully build @fastio 's PR with Ragel 7.0?   or just successfully build latest scylla master with this PR?

nyh:
No, I did not try to compile this pull request. What I meant was that I didn't see any problem with Seastar or its HTTP parser (which I think is where we used ragel?) with the Ragel 7.0 I have in the machine. What is the problem with Ragel 7.0, and how difficult is it to work around it?

amoskong:
> No, I did not try to compile this pull request. What I meant was that I didn't see any problem with Seastar or its HTTP parser (which I think is where we used ragel?) with the Ragel 7.0 I have in the machine. What is the problem with Ragel 7.0, and how difficult is it to work around it?

seastar uses fixed version of ragel  ()
```
$ git grep ragel
CMakeLists.txt:find_package (ragel 6.10 REQUIRED)
CMakeLists.txt:function (seastar_generate_ragel)
CMakeLists.txt:    COMMAND ${ragel_RAGEL_EXECUTABLE} -G2 -o ${args_OUT_FILE} ${args_IN_FILE}
....
cooking_recipe.cmake:    URL http://www.colm.net/files/ragel/ragel-6.10.tar.gz
···

@fastio added an option `--with-ragel`, we can assign different version of ragel.

The only concern is daily jenkins will be broken, this is why I downgrade ragel of dbuild image.
Note: only ragel version of docker image will be downgraded, @nyh it won't downgrade ragel of your machine.

**Comment by fastio** [comment]
@avikivity The problem is that  the code generated with ragel 7 contains the in-class initialization of static const integral type members, and these members are never used by the rest codes.

The compiler errors:
```
build/release/gen/redis/protocol_parser.hh:43:20: error: in-class initialization of static data member 'const char redis_protocol_parser::_nfa_targs []' of incomplete type
   43 |  static const char _nfa_targs[] = {
      |                    ^~~~~~~~~~
build/release/gen/redis/protocol_parser.hh:47:20: error: in-class initialization of static data member 'const char redis_protocol_parser::_nfa_offsets []' of incomplete type
   47 |  static const char _nfa_offsets[] = {
      |                    ^~~~~~~~~~~~
build/release/gen/redis/protocol_parser.hh:53:20: error: in-class initialization of static data member 'const char redis_protocol_parser::_nfa_push_actions []' of incomplete type
   53 |  static const char _nfa_push_actions[] = {
      |                    ^~~~~~~~~~~~~~~~~
build/release/gen/redis/protocol_parser.hh:57:20: error: in-class initialization of static data member 'const char redis_protocol_parser::_nfa_pop_trans []' of incomplete type
   57 |  static const char _nfa_pop_trans[] = {
```
The code generated by ragel 7:

```
 static const char _nfa_targs[] = { 
        0, 0
    };  
    
    static const char _nfa_offsets[] = { 
        0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0
    };  
    
    static const char _nfa_push_actions[] = { 
        0, 0
    };  
    
    static const char _nfa_pop_trans[] = { 
        0, 0
    };  
```

BTW, I found that you mentioned this issue in the ragel-users mail list in 2017, http://www.colm.net/pipermail/ragel-users/2017-June/003481.html

**Comment by avikivity** [comment]
I suppose you can apply the same workaround as in scylladb/seastar@1cb8b0eec2141bbc7bdd297362c0c306dd566f48

File: db/config.cc [hunk 32]
```diff
@@ -702,6 +702,20 @@ db::config::config(std::shared_ptr<db::extensions> exts)
             "especially when multiple clustering key columns have IN restrictions. Increasing this value can result in server instability.")
     , alternator_port(this, "alternator_port", value_status::Used, 0, "Alternator API port")
     , alternator_address(this, "alternator_address", value_status::Used, "0.0.0.0", "Alternator API listening address")
+    , redis_transport_port(this, "redis_transport_port", value_status::Used, 6379, "Port on which the REDIS transport listens for clients.")
+    , redis_transport_port_ssl(this, "redis_transport_port_ssl", value_status::Used, 9142, "Port on which the REDIS TLS native transport listens for clients.")
+    , enable_redis_protocol(this, "enable_redis_protocol", value_status::Used, false, "Enable redis protocol; Scylla will process redis protocol as a redis cluster if enable.")
```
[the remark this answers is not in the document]
nyh: I still don't understand what enable_redis_protocol is supposed to do (enable http? https? both? or what) but you can fix this later.

db/config.cc
```diff
@@ -702,6 +702,20 @@ db::config::config(std::shared_ptr<db::extensions> exts)
...
+    , redis_read_consistency_level(this, "redis_read_consistency_level", value_status::Used, "ONE", "Consistency level for read operations for redis.")
+    , redis_write_consistency_level(this, "redis_write_consistency_level", value_status::Used, "ANY", "Consistency level for write operations for redis.")
```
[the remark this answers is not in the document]
nyh: Please remember these review comments, I think you'll still need to address them, but it doesn't need to block committing this version.

docs/redis/redis.md
```diff
@@ -0,0 +1,360 @@
...
+maps a key (as STRING) to a collection of ITEMs which associated with a
+score. It allows us to fetch data by score. 
+
+To store ZSETs data,  the scylla table is created by following CQL:
+
+```
+CREATE TABLE ZSETs (
+    pkey text,
+    ckey double,
+    data text,
```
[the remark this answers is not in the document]
nyh: Again, I think this is a pretty important point which you didn't address. Please, after this version is committed, do go over all my comments and write down as issues, fixmes, or followup patches, whatever needs to be done.

nyh:
P.S. Don't try to understand the review comments from this "discussion window". Got to the "files changed: 44" link at the top, and there you can see the review comments on the actual lines of code they apply to.

configure.py
```diff
@@ -1299,7 +1299,9 @@ def configure_zstd(build_dir, mode):
             restat = 1
             description = NINJA $out
         rule ragel
-            command = {ragel_exec} -G2 -o $out $in
+            # sed away a bug in ragel 7 that emits some extraneous _nfa* variables
+            # (the $$ is collapsed to a single one by ninja)
+            command = {ragel_exec} -G2 -o $out $in && sed -i -e '1h;2,$$H;$$!d;g' -re 's/static const char _nfa[^;]*;//g' $out
             description = RAGEL $out
```
avikivity: Please fold this into the commit that introduces the problem, so we don't break bisectability.

fastio:
> @fastio please note that there a still a significant number of comments that I left during my review which you have not addressed. I don't think any of these issues needs to delay the commit - you can address them later - but please do take a look at all of them. I might have got some of my comments wrong, but I think at least some of them are right and worth fixing or at least remembering to fix (using issues in the bug tracker and/or FIXMEs in the code).

Thanks. I got it. I'll take a look at these issues and fix them soon.

configure.py
nyh: @fastio please merge these two patches (unless I'm missing something, this appears to be Avi's last request), and I will commit the resulting version. Thanks.

fastio: Hi @nyh, What do you mean is that the commit 5082a54 should be merged to the previous one (e9738a4) ?

PR #5136, opened 2019-10-03
build: fix double linkage of Seastar libraries in tests

avikivity:
Tests that use the seastar test framework link with its libraries (as described
pkg-config --libs --static seastar-testing.pc). The output of that command however
also includes seastar's libraries, because seastar-testing.pc explicitly depends
on seastar. As a result, the linker command line specifies each seastar library
(and its dependencies) twice.

Up to now this has not been a problem; the linker ignores the doubly-specified
libraries. However with a recent Seastar fix, this starts failing with --enable-dpdk,
because dpdk libraries are linked with --whole-archive. This forces the linker to
include the library in the executable, and if this is done twice, the linker
complains about duplicate symbols.

Fix by removing the --static flag from the pkg-config command line. Since we
explicitly include seastar dependencies, it is not needed. With this we can build
Scylla tests with recent Seastar and with --enable-dpdk.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1348,7 +1348,7 @@ with open(buildfile_tmp, 'w') as f:
                         local_libs += ' ' + maybe_static(args.staticboost, '-lboost_unit_test_framework')
                     if binary not in tests_not_using_seastar_test_framework:
                         pc_path = pc[mode].replace('seastar.pc', 'seastar-testing.pc')
-                        local_libs += ' ' + pkg_config(pc_path, '--libs', '--static')
+                        local_libs += ' ' + pkg_config(pc_path, '--libs')
                     if has_thrift:
                         local_libs += ' ' + thrift_libs + ' ' + maybe_static(args.staticboost, '-lboost_system')
                     # Our code's debugging information is huge, and multiplied


nyh:
@avikivity , we had in the past a huge email thread about this "--static" - I said it was wrong to use this option, and I lost the argument. Isn't it strange to make this change now incidentally, as a workaround for some pretty specific issue?
Moreover, I didn't understand why we can't fix this issue more directly - i.e., not depend on Seastar twice. I'm afraid I'm not familiar enough with the new cmake stuff to understand your explanation on why it appeared twice in the first place.

espindola:
--whole-archive is a horrible hack that breaks the abstraction of what a library is. With it present there is not right answer to "how do I link with this library", since the answer depends on whether there is another dependency path using the library.

So I think the correct solution is to not use it. My suggestion for how to do that is:

* Use "ld -r" to create a seartar-dpdk.o:
 ld -r -o seastar-dpdk.o --whole-archive dpdk-libs --no-whole-archive
* include seastar-dpdk.o in seastar.o
* have an undefined reference (maybe with -u command line) to bring in that file.

With that seastar.a can be included as many times as needed in a link line. I am happy to write that patch if you want.

avikivity:
Agree your solution is better, but I lack the cmake skills to implement it. Can you cook a patch?

espindola:
I am writing a patch.

avikivity:
Withdrawing in favor of @espindola's patch.

Issue #3912, opened 2018-11-13
add memtable partition/row hit/miss counters


vladzcloudius:
*Installation details*
Scylla version: master at 1c5934c9348de560c851a61ddc0da7a75d5a19c1

*Description*
Sometimes we need to reason about the amount of overwrites out of all writes.
Deciding about compaction strategy is one of the cases. 

Right now we have counters that allow us to reason about this in some particular cases. 

For instance we can deduct it in a single rows partitions non-BATCHED writes case using `cache/rows_processed_from_memtable` and `database/total_writes`.

However for more complicated use cases these counters may not be enough.

It would be easier to have explicit partition/row hit/miss counters similar to those we cache for a `cache` in a read flow today.

PR #5138, opened 2019-10-03
Partition/row hit/miss counters for memtable write operations

piodul:
Fixes issue: #3912

Adds four new per-table metrics for counting writes performed on memtables:
- `memtable_partition_writes` - number of write operations performed on partitions in memtables,
- `memtable_partition_hits` - number of write operations performed on partitions that previously existed in a memtable,
- `memtable_row_writes` - number of row write operations performed in memtables,
- `memtable_row_hits` - number of row write operations that overwrote rows previously present in a memtable.

### DIFF
--- a/api/column_family.cc
+++ b/api/column_family.cc
@@ -75,28 +75,28 @@ future<> foreach_column_family(http_context& ctx, const sstring& name, function<
 }
 
 future<json::json_return_type>  get_cf_stats(http_context& ctx, const sstring& name,
-        int64_t column_family::stats::*f) {
+        int64_t column_family_stats::*f) {
     return map_reduce_cf(ctx, name, int64_t(0), [f](const column_family& cf) {
         return cf.get_stats().*f;
     }, std::plus<int64_t>());
 }
 
 future<json::json_return_type>  get_cf_stats(http_context& ctx,
-        int64_t column_family::stats::*f) {
+        int64_t column_family_stats::*f) {
     return map_reduce_cf(ctx, int64_t(0), [f](const column_family& cf) {
         return cf.get_stats().*f;
     }, std::plus<int64_t>());
 }
 
 static future<json::json_return_type>  get_cf_stats_count(http_context& ctx, const sstring& name,
-        utils::timed_rate_moving_average_and_histogram column_family::stats::*f) {
+        utils::timed_rate_moving_average_and_histogram column_family_stats::*f) {
     return map_reduce_cf(ctx, name, int64_t(0), [f](const column_family& cf) {
         return (cf.get_stats().*f).hist.count;
     }, std::plus<int64_t>());
 }
 
 static future<json::json_return_type>  get_cf_stats_sum(http_context& ctx, const sstring& name,
-        utils::timed_rate_moving_average_and_histogram column_family::stats::*f) {
+        utils::timed_rate_moving_average_and_histogram column_family_stats::*f) {
     auto uuid = get_uuid(name, ctx.db.local());
     return ctx.db.map_reduce0([uuid, f](database& db) {
         // Histograms information is sample of the actual load
@@ -112,14 +112,14 @@ static future<json::json_return_type>  get_cf_stats_sum(http_context& ctx, const
 
 
 static future<json::json_return_type>  get_cf_stats_count(http_context& ctx,
-        utils::timed_rate_moving_average_and_histogram column_family::stats::*f) {
+        utils::timed_rate_moving_average_and_histogram column_family_stats::*f) {
     return map_reduce_cf(ctx, int64_t(0), [f](const column_family& cf) {
         return (cf.get_stats().*f).hist.count;
     }, std::plus<int64_t>());
 }
 
 static future<json::json_return_type>  get_cf_histogram(http_context& ctx, const sstring& name,
-        utils::timed_rate_moving_average_and_histogram column_family::stats::*f) {
+        utils::timed_rate_moving_average_and_histogram column_family_stats::*f) {
     utils::UUID uuid = get_uuid(name, ctx.db.local());
     return ctx.db.map_reduce0([f, uuid](const database& p) {
         return (p.find_column_family(uuid).get_stats().*f).hist;},
@@ -130,7 +130,7 @@ static future<json::json_return_type>  get_cf_histogram(http_context& ctx, const
     });
 }
 
-static future<json::json_return_type> get_cf_histogram(http_context& ctx, utils::timed_rate_moving_average_and_histogram column_family::stats::*f) {
+static future<json::json_return_type> get_cf_histogram(http_context& ctx, utils::timed_rate_moving_average_and_histogram column_family_stats::*f) {
     std::function<utils::ihistogram(const database&)> fun = [f] (const database& db)  {
         utils::ihistogram res;
         for (auto i : db.get_column_families()) {
@@ -146,7 +146,7 @@ static future<json::json_return_type> get_cf_histogram(http_context& ctx, utils:
 }
 
 static future<json::json_return_type>  get_cf_rate_and_histogram(http_context& ctx, const sstring& name,
-        utils::timed_rate_moving_average_and_histogram column_family::stats::*f) {
+        utils::timed_rate_moving_average_and_histogram column_family_stats::*f) {
     utils::UUID uuid = get_uuid(name, ctx.db.local());
     return ctx.db.map_reduce0([f, uuid](const database& p) {
         return (p.find_column_family(uuid).get_stats().*f).rate();},
@@ -157,7 +157,7 @@ static future<json::json_return_type>  get_cf_rate_and_histogram(http_context& c
     });
 }
 
-static future<json::json_return_type> get_cf_rate_and_histogram(http_context& ctx, utils::timed_rate_moving_average_and_histogram column_family::stats::*f) {
+static future<json::json_return_type> get_cf_rate_and_histogram(http_context& ctx, utils::timed_rate_moving_average_and_histogram column_family_stats::*f) {
     std::function<utils::rate_moving_average_and_histogram(const database&)> fun = [f] (const database& db)  {
         utils::rate_moving_average_and_histogram res;
         for (auto i : db.get_column_families()) {
@@ -407,11 +407,11 @@ void set_column_family(http_context& ctx, routes& r) {
     });
 
     cf::get_memtable_switch_count.set(r, [&ctx] (std::unique_ptr<request> req) {
-        return get_cf_stats(ctx,req->param["name"] ,&column_family::stats::memtable_switch_count);
+        return get_cf_stats(ctx,req->param["name"] ,&column_family_stats::memtable_switch_count);
     });
 
     cf::get_all_memtable_switch_count.set(r, [&ctx] (std::unique_ptr<request> req) {
-        return get_cf_stats(ctx, &column_family::stats::memtable_switch_count);
+        return get_cf_stats(ctx, &column_family_stats::memtable_switch_count);
     });
 
     // FIXME: this refers to partitions, not rows.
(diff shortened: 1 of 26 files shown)


database.cc
```diff
@@ -1331,7 +1331,7 @@ future<> memtable_list::request_flush() {
 }
```
avikivity: Is this only for memtables? If so, the pull request description, patch subjects, and variable names should mention it to avoid confusion.

memtable.cc
```diff
@@ -211,8 +212,14 @@ memtable::find_or_create_partition(const dht::decorated_key& key) {
         memtable_entry* entry = current_allocator().construct<memtable_entry>(
             _schema, dht::decorated_key(key), mutation_partition(_schema));
         partitions.insert_before(i, *entry);
+        if (_table_stats) {
+            ++_table_stats->partition_creations;
+        }
         return entry->partition();
```
avikivity: Better to force the caller to supply a stats object than to add conditionals everywhere. Statistics should be cheap.

mutation_partition.cc
```diff
@@ -249,16 +249,16 @@ void mutation_partition::ensure_last_dummy(const schema& s) {
     }
 }
 
-void mutation_partition::apply(const schema& s, const mutation_partition& p, const schema& p_schema) {
-    apply_weak(s, p, p_schema);
+mutation_application_stats mutation_partition::apply(const schema& s, const mutation_partition& p, const schema& p_schema) {
+    return apply_weak(s, p, p_schema);
```
avikivity: Maybe this should be an output parameter. I don't think it's important enough to be the return value. @tgrabiec ?

It's good that you made mutation_application_stats a new type.

avikivity: (btw I usually dislike output parameters)

mutation_partition.cc
```diff
@@ -304,17 +304,19 @@ mutation_partition::apply(const schema& s, const mutation_fragment& mf) {
     mf.visit(applier);
 }
 
-stop_iteration mutation_partition::apply_monotonically(const schema& s, mutation_partition&& p, cache_tracker* tracker, is_preemptible preemptible) {
+std::tuple<stop_iteration, mutation_application_stats> mutation_partition::apply_monotonically(const schema& s, mutation_partition&& p, cache_tracker* tracker, is_preemptible preemptible) {
 #ifdef SEASTAR_DEBUG
```
avikivity: Again would be better as an output parameter. The stop_iteration is the main point of the function, the stats are a side effect.

mutation_partition.hh
```diff
@@ -878,6 +878,11 @@ public:
     friend std::ostream& operator<<(std::ostream& os, const printer& p);
 };
 
+struct mutation_application_stats {
+    uint64_t row_creations;
+    uint64_t row_overwrites;
+};
```
avikivity: You can use = 0; to initialize them here and not worry about forgetting to initialize them later.

table.cc
```diff
@@ -1089,7 +1089,11 @@ void table::set_metrics() {
...
                 ms::make_gauge("live_sstable", ms::description("Live sstable count"), _stats.live_sstable_count)(cf)(ks),
-                ms::make_gauge("pending_compaction", ms::description("Estimated number of compactions pending for this column family"), _stats.pending_compactions)(cf)(ks)
+                ms::make_gauge("pending_compaction", ms::description("Estimated number of compactions pending for this column family"), _stats.pending_compactions)(cf)(ks),
+                ms::make_gauge("partition_creations", ms::description("Number of times a write operation created a new partition"), _stats.partition_creations)(cf)(ks),
+                ms::make_gauge("partition_reuses", ms::description("Number of times a write operation was issued in an existing partition"), _stats.partition_reuses)(cf)(ks),
+                ms::make_gauge("row_creations", ms::description("Number of times a new row was created"), _stats.row_creations)(cf)(ks),
+                ms::make_gauge("row_overwrites", ms::description("Number of times a row was updated"), _stats.row_overwrites)(cf)(ks)
         });
```
avikivity: Maybe better is rows (with the sum) and row_overwrites (similar for partitions). This gives you a single counter for partition and row writes which is the most interesting thing, and a secondary counter for overwrites.

api/column_family.cc
```diff
@@ -75,28 +75,28 @@ future<> foreach_column_family(http_context& ctx, const sstring& name, function<
 }
```
avikivity: Missing motivation in commit message.

database.cc
haaawk: It is so probably it's better to update the names of the counters and the other things mentioned by Avi.

mutation_partition.cc
tgrabiec: I also think that stats are better collected by passing a stats collecting object as a parameter.

memtable.cc
```diff
@@ -211,8 +212,14 @@ memtable::find_or_create_partition(const dht::decorated_key& key) {
...
     } else {
+        if (_table_stats) {
+            ++_table_stats->partition_reuses;
+        }
```
tgrabiec: We already use the terms "partition_insertions" and "partition_hits" in the cache. Perhaps it's better to be consistent here in naming.

database.cc
piodul: Updated commit and PR descriptions. Added "memtable_" prefix to variable and counter names.

memtable.cc
piodul: Done. On memtable construction, if no `table_stats` are provided, a dummy `thread_local` object will be supplied, similarly to the `dirty_memory_manager` object.

mutation_partition.cc
piodul: Done, now `mutation_application_stats` is now passed as an output parameter. To make this parameter optional, I added overloads of `apply` and `apply_*` methods of `mutation_partition` that do not take this argument. Those variants use a temporary, stack-allocated `mutations_application_stats` structure.

mutation_partition.cc
piodul: Done.

mutation_partition.hh
piodul: Done.

memtable.cc
piodul: Changed names, now using names like "memtable_partition_writes" and "memtable_partition_hits".

table.cc
piodul: OK. Changed those counters to report total writes and overwrites (`memtable_partition_writes` + `memtable_partition_hits`, `memtable_row_writes` + `memtable_row_hits`).

api/column_family.cc
piodul: Added.

database.hh
```diff
@@ -312,9 +312,35 @@ struct cf_stats {
...
+    utils::timed_rate_moving_average_and_histogram reads{256};
+    utils::timed_rate_moving_average_and_histogram writes{256};
+    utils::estimated_histogram estimated_read;
+    utils::estimated_histogram estimated_write;
+    utils::estimated_histogram estimated_sstable_per_read{35};
+    utils::timed_rate_moving_average_and_histogram tombstone_scanned;
+    utils::timed_rate_moving_average_and_histogram live_scanned;
+    utils::estimated_histogram estimated_coordinator_read;
+};
+
```
vladzcloudius: Why not to have cheap stats always enabled while expensive (histograms) conditionally enabled?

This would also resolve that Avi's comment about the conditional stats update and would allow you drop that workaround that you made.

mutation_partition.cc
```diff
@@ -250,15 +250,33 @@ void mutation_partition::ensure_last_dummy(const schema& s) {
 }
 
 void mutation_partition::apply(const schema& s, const mutation_partition& p, const schema& p_schema) {
-    apply_weak(s, p, p_schema);
+    mutation_application_stats app_stats;
+    return apply(s, p, p_schema, app_stats);
```
vladzcloudius: I'm not sure I'm following: where those app_stats eventually go after apply() updates them?

vladzcloudius: And C++ wise the above looks weird at least (it's unclear how this even compiles): you have a return statement in the function returning `void`.

api/column_family.cc
```diff
@@ -75,28 +75,28 @@ future<> foreach_column_family(http_context& ctx, const sstring& name, function<
...
 
 future<json::json_return_type>  get_cf_stats(http_context& ctx, const sstring& name,
-        int64_t column_family::stats::*f) {
+        int64_t column_family_stats::*f) {
     return map_reduce_cf(ctx, name, int64_t(0), [f](const column_family& cf) {
```
avikivity: Note, we often alias the old name to the new name to avoid this busywork (no need to undo that work).

database.hh
```diff
@@ -312,9 +312,35 @@ struct cf_stats {
...
+    /** Number of times flush has resulted in the memtable being switched out. */
+    int64_t memtable_switch_count = 0;
+    /** Estimated number of tasks pending for this column family */
+    int64_t pending_flushes = 0;
+    int64_t live_disk_space_used = 0;
+    int64_t total_disk_space_used = 0;
+    int64_t live_sstable_count = 0;
+    /** Estimated number of compactions pending for this column family */
+    int64_t pending_compactions = 0;
+    int64_t partition_creations = 0;
```
avikivity: Looks like you snuck in some extra changes thinking no one would notice.

database.hh
```diff
@@ -327,10 +333,10 @@ struct table_stats {
...
     /** Estimated number of compactions pending for this column family */
     int64_t pending_compactions = 0;
-    int64_t partition_creations = 0;
-    int64_t partition_reuses = 0;
-    int64_t row_creations = 0;
-    int64_t row_overwrites = 0;
+    int64_t memtable_partition_insertions = 0;
+    int64_t memtable_partition_hits = 0;
+    int64_t memtable_row_insertions = 0;
+    int64_t memtable_row_hits = 0;
```
avikivity: These should have been introduced in this patch, not in the struct rename patch.

memtable.cc
avikivity: Yes, that's fine.

mutation_partition.cc
avikivity: Why did you change apply_weak() to apply()?

avikivity: @vladzcloudius it's legal to return void (very useful for templates). But should be when not needed, it's surprising.

mutation_partition.cc
```diff
@@ -250,15 +250,33 @@ void mutation_partition::ensure_last_dummy(const schema& s) {
 }
 
 void mutation_partition::apply(const schema& s, const mutation_partition& p, const schema& p_schema) {
-    apply_weak(s, p, p_schema);
+    mutation_application_stats app_stats;
```
avikivity: You should get this from the caller, or you can lose some row creations. These should always be propagated from the top-most scope that either has a place to store them or knows it can throw them away (like tests).

mutation_partition.cc
avikivity: Ah, I understand the applyapply_weak thing. But leaving a version that doesn't accept a stats object is an invitation to forget to propagate them. It's very annoying to always have to propagate these objects, but also very safe.

partition_version.cc
```diff
@@ -200,7 +201,10 @@ stop_iteration partition_snapshot::merge_partition_versions() {
         }
         while (auto prev = current->prev()) {
             region().allocator().invalidate_references();
-            if (current->partition().apply_monotonically(*schema(), std::move(prev->partition()), _tracker, is_preemptible::yes) == stop_iteration::no) {
+            mutation_application_stats app_stats;
+            const auto do_stop = current->partition().apply_monotonically(*schema(), std::move(prev->partition()),
+                    _tracker, app_stats, is_preemptible::yes);
+            if (do_stop == stop_iteration::no) {
                 return stop_iteration::no;
```
avikivity: Why the big code change?

We could (later) also collect these merging/cleaning statistics.

piodul: Removed temporary `do_stop` variable.

mutation_partition.cc
piodul: Now, the `mutable_application_stats` argument is required in `apply`/`apply_*` methods. I converted all invocations which previously did not supply the argument to provide a throw-away stats argument. If you would like me to, I can try propagating the stats argument higher up the call hierarchy.

database.hh
piodul: Sorry, I messed up a little during the rebase. New counters are properly introduced in the second commit, and the issue with returning void is resolved, too.

database.hh
piodul: @vladzcloudius  If I understand correctly that you mean splitting the struct and adding a required argument to `memtable`'s constructor, then it would generate a big change:
```
[piodul@localhost scylla]$ git grep "make_lw_shared<memtable>" | wc -l
180
```
Almost all of those would need to be modified to include the new dummy parameter.
Furthermore, only one of those invocations is in non-test code and it always is passed a real argument, the rest is the test code.

piodul:
Ping @avikivity

partition_version.cc
```diff
@@ -364,6 +369,8 @@ void partition_entry::apply(const schema& s, mutation_partition&& mp, const sche
     }
     new_version->insert_before(*_version);
     set_version(new_version);
+    app_stats.row_insertions = mp.live_row_count(s);
+    app_stats.row_hits = 0;
```
avikivity: I don't understand these two lines. Shouldn't you only be adding to the stats, never clearing them?

table.cc
```diff
@@ -1085,6 +1085,10 @@ void table::set_metrics() {
     if (_config.enable_metrics_reporting) {
         _metrics.add_group("column_family", {
                 ms::make_derive("memtable_switch", ms::description("Number of times flush has resulted in the memtable being switched out"), _stats.memtable_switch_count)(cf)(ks),
+                ms::make_gauge("memtable_partition_writes", ms::description("Number of write operations performed on partitions in memtables"), [this] () { return _stats.memtable_partition_insertions + _stats.memtable_partition_hits; })(cf)(ks),
+                ms::make_gauge("memtable_partition_hits", ms::description("Number of times a write operation was issued on an existing partition in memtables"), _stats.memtable_partition_hits)(cf)(ks),
+                ms::make_gauge("memtable_row_writes", ms::description("Number of row writes performed in memtables"), [this] () { return _stats.memtable_row_insertions + _stats.memtable_row_hits; })(cf)(ks),
+                ms::make_gauge("memtable_row_hits", ms::description("Number of rows overwritten by write operations in memtables"), _stats.memtable_row_hits)(cf)(ks),
```
avikivity: These should be make_counter(). Gauges are for things that memory in use that can go up and down. Counters are for counted events that are intended to be viewed using the first derivative wrt. time.

partition_version.cc
piodul: Indeed, these values should be added, not assigned. Previously `app_stats` was a struct that was returned from this function. It is now corrected.

memtable.cc
```diff
@@ -710,7 +714,10 @@ memtable::apply(const mutation& m, db::rp_handle&& h) {
           with_linearized_managed_bytes([&] {
             auto& p = find_or_create_partition(m.decorated_key());
             _stats_collector.update(*m.schema(), m.partition());
-            p.apply(*_schema, m.partition(), *m.schema());
+            mutation_application_stats app_stats;
+            p.apply(*_schema, m.partition(), *m.schema(), app_stats);
+            _table_stats.memtable_row_insertions += app_stats.row_insertions;
+            _table_stats.memtable_row_hits += app_stats.row_hits;
```
tgrabiec: Not a must, but you could avoid some code duplication if you stored `_table_stats.memtable` as `mutation_application_stats`, and defined operator+= on `mutation_application_stats`. Then you could just do `_table_stats.memtable += app_stats` here and below and don't care what stats are exactly tracked. So slightly more future proof.

partition_version.cc
```diff
@@ -364,6 +369,7 @@ void partition_entry::apply(const schema& s, mutation_partition&& mp, const sche
     }
     new_version->insert_before(*_version);
     set_version(new_version);
+    app_stats.row_insertions += mp.live_row_count(s);
```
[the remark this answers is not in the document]
piodul: I think that we still have to count rows from `mp` in case it is necessary to create a new version. Otherwise, if an INSERT operation creates a new version, writes will not be counted.

As for undetected overwrites, I am now accounting for them in `merge_partition_versions`. I managed to produce a situation in which a new version is created and writes are applied to it. The new solution correctly updates overwrite counter when versions are squashed, while the old did not.

memtable.cc
piodul: Good idea - converted those two fields into `mutation_application_structs memtable;` and added `+=` operator. I named it `memtable_app_stats` in order not to clash with other `table_stats` members that have names with `memtable_*` prefix.

piodul:
Minor change: renamed `app_stats_for_tests` -> `dummy_app_stats` in `row_cache.cc:53`

partition_version.cc
piodul: Re-added the variable, but it actually is needed now because of the addition-assignment.

piodul:
Ping @tgrabiec, @avikivity, @vladzcloudius

tgrabiec:
@piodul The change has conflicts with master, please resolve

haaawk:
@tgrabiec / @avikivity the change seems to be approved by Tomek. Would you mind merging it please?

PR #5156, opened 2019-10-09
distributed_loader: populate non-system keyspaces in parallel

### LINKED ISSUE #2007: populate non system keyspaces in parallel
When populating non-system keyspaces, we scan data dir and for each entry (keyspace dir) which name isn't 'system', we populate the keyspace. That's done serially.

I think it makes sense to load non system keyspaces in parallel so as to increase parallelism and reduce boot time because population is I/O intensive and barely uses CPU.

The list of non system keyspaces is available, and scanning data dir for them is uneeded. So we only need to get the list and iterate through it in parallel calling distributed_loader::populate_keyspace()

piodul:
Before this change, when populating non-system keyspaces, each data
directory was scanned and for each entry (keyspace directory),
a keyspace was populated. This was done in a serial fashion - populating
of one keyspace was not started until the previous one was done.

Loading keyspaces in such fashion can introduce unnecessary waiting
in case of a large number of keyspaces in one data directory. Population
process is I/O intensive and barely uses CPU.

This change enables parallel loading of keyspaces per data directory.
Populating the next keyspace does not wait for the previous one.

A benchmark was performed measuring startup time, with the following
setup:
  - 1 data directory,
  - 200 keyspaces,
  - 2 tables in each keyspace, with the following schema:
      CREATE TABLE tbl (a int, b int, c int, PRIMARY KEY(a, b))
        WITH CLUSTERING ORDER BY (b DESC),
  - 1024 rows in each table, with values (i, 2*i, 3*i) for i in 0..1023,
  - ran on 6-core virtual machine running on i7-8750H CPU,
  - compiled in dev mode,
  - parameters: --smp 6 --max-io-requests 4 --developer-mode=yes
      --datadir $DIR --commitlog-directory $DIR
      --hints-directory $DIR --view-hints-directory $DIR

The benchmark tested:
  - boot time, by comparing timestamp of the first message in log,
    and timestamp of the following message:
      "init - Scylla version ... initialization completed."
  - keyspace population time, by comparing timestamps of messages:
      "init - loading non-system sstables"
    and
      "init - starting view update generator"

The benchmark was run 5 times for sequential and parallel version,
with the following results:
  - sequential: boot 31.620s, keyspace population 6.051s
  - parallel:   boot 29.966s, keyspace population 4.360s

Keyspace population time decreased by ~27.95%, and overall boot time
by about ~5.23%.

Tests: unit(release)

Fixes #2007

### DIFF
--- a/distributed_loader.cc
+++ b/distributed_loader.cc
@@ -41,6 +41,7 @@
 extern logging::logger dblog;
 
 sstables::sstable::version_types get_highest_supported_format();
+static future<> execute_futures(std::vector<future<>>& futures);
 
 static const std::unordered_set<sstring> system_keyspaces = {
                 db::system_keyspace::NAME, db::schema_tables::NAME
@@ -366,12 +367,18 @@ static std::vector<sstables::shared_sstable> sstables_for_shard(const std::vecto
 }
 
 static future<> populate(distributed<database>& db, sstring datadir) {
-    return lister::scan_dir(datadir, { directory_entry_type::directory }, [&db] (fs::path datadir, directory_entry de) {
-        auto& ks_name = de.name;
-        if (is_system_keyspace(ks_name)) {
+    return do_with(std::vector<future<>>(), [&db, datadir = std::move(datadir)] (std::vector<future<>>& futures) {
+        // push futures that populate keyspaces into array of futures,
+        // so that the supplied callback will not block scan_dir() from
+        // reading the next entry in the directory.
+        return lister::scan_dir(datadir, { directory_entry_type::directory }, [&db, &futures] (fs::path datadir, directory_entry de) {
+            if (!is_system_keyspace(de.name)) {
+                futures.emplace_back(distributed_loader::populate_keyspace(db, datadir.native(), de.name));
+            }
             return make_ready_future<>();
-        }
-        return distributed_loader::populate_keyspace(db, datadir.native(), ks_name);
+        }).then([&futures] {
+            return execute_futures(futures);
+        });
     });
 }
 


raphaelsc:
Great! Reviewed-by: Raphael S. Carvalho <raphaelsc@scylladb.com>

haaawk:
Please add Fixes #<issue number> to commit message.

distributed_loader.cc
```diff
@@ -366,12 +367,20 @@ static std::vector<sstables::shared_sstable> sstables_for_shard(const std::vecto
...
-    return lister::scan_dir(datadir, { directory_entry_type::directory }, [&db] (fs::path datadir, directory_entry de) {
-        auto& ks_name = de.name;
-        if (is_system_keyspace(ks_name)) {
+    return do_with(std::vector<future<>>(), [&db, datadir = std::move(datadir)] (std::vector<future<>>& futures) {
+        // push futures that populate keyspaces into array of futures,
+        // so that the supplied callback will not block scan_dir() from
+        // reading the next entry in the directory.
+        return lister::scan_dir(datadir, { directory_entry_type::directory }, [&db, &futures] (fs::path datadir, directory_entry de) {
+            auto& ks_name = de.name;
+            if (is_system_keyspace(ks_name)) {
```
haaawk: Super nit: I think the code will be shorter if you do:
\[ ... \]( ... ){
if (!is_system_keyspace(de.name)) {
    futures.emplace_back(distributed_loader::populate_keyspace(db, datadir.native(), de.name));
}
};

No need to return at all.

piodul:
@haaawk added information on fixed issue number to the commit message and PR description.

haaawk:
There's one more thing you could do @piodul -> Run all tests in release (test.py --mode=release) and add "tests: unit(release)" to the commit message to make it clear that you have tested the code.

distributed_loader.cc
piodul: I inverted the condition, but the `lister::scan_dir` function requires a `std::function<future<> (something)>`. Unless I return `make_ready_future<>()` at the end, the compiler complains about incompatible types.

piodul:
@haaawk done, added information on unit tests.

avikivity:
Queued as c04e8c37aa7b1bfcf301b06cfea6be978958e4ec

PR #5166, opened 2019-10-11
view: fix bug in virtual columns.

kbr-:
When creating a virtual column of non-frozen map type,
the wrong type was used for the map's keys.

Fixes https://github.com/scylladb/scylla/issues/5165.


nyh:
Wow, excellent catch, and sad the original test didn't catch this (because, as you noticed, it used int for both key and value). I'll commit the patch.

bhalevy:
Merged in ef9d5750c82592afbae5b4174cb7045c3ba5e2df

PR #5167, opened 2019-10-11
db: Added virtual table `system.clients'

jul-stas:
Task is here: https://github.com/scylladb/scylla/issues/4820
I used the following as a reference:
https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/virtual/ClientsTable.java
At this moment the table lacks: SSL info, driver info, hostname and per-client request count. I'll appreciate any advice on where to look for this info.
Also, control connections are reported too.

I'll add support for `nodetool clientlist` in a separate PR once this is stabilized.


db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+                    [this, scdv=std::move(sorted_client_data_vec)] (auto&& mutations_vec) mutable {
+                mutations_vec.reserve(scdv.size());
+
+                std::for_each(scdv.begin(), scdv.end(), [this, &mutations_vec] (auto&& client) {
+                    mutations_vec.emplace_back(db::system_keyspace::make_client_mutation(std::move(client)));
+                });
+                if (!mutations_vec.empty()) {
+                    _mutation_fragmentator_opt = flat_mutation_reader_from_mutations(std::move(mutations_vec));
+                }
+                return make_ready_future<>();
```
haaawk: Is this return needed? It seems that the code could work without it.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+    for (auto&& client : clients) {
+        const auto pk = partition_key::from_single_value(*_schema, inet_addr_type->decompose(client.ip));
+        const dht::token calculated_token = dht::global_partitioner().get_token(*_schema, pk);
+        if (dht::shard_of(calculated_token) != engine().cpu_id()) {
+            continue;
+        }
+        for (auto& range : _slice.row_ranges(*_schema, pk)) {
+            const auto ck_bytes = data_type_for<decltype(system_keyspace::client_data::port
+                    )>()->decompose(client.port);
+            const auto ck_prefix = clustering_key_prefix::from_single_value(*_schema, ck_bytes);
```
denesb: This should be move outside the inner loop.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+        if (_slice.options.contains(query::partition_slice::option::reversed)) {
+            std::sort(
+                    cluster.begin(), cluster.end(),
+                    [] (const auto& client1, const auto& client2) { return client1.port > client2.port; });
+        } else {
+            std::sort(
+                    cluster.begin(), cluster.end(),
+                    [] (const auto& client1, const auto& client2) { return client1.port < client2.port; });
+        }
+        sorted_clients.insert(
```
denesb: Note that clients should be in token order.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+        return make_ready_future<>();
+    }
+
+    std::vector<system_keyspace::client_data> get() && {
+        return std::move(_val);
+    }
+};
+
+class ck_comparator {
+    schema_ptr _schema;
```
haaawk: Nit: reader already ensures that schema_ptr is alive so it should be enough to take const schema& here.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+public:
+    using port_type = decltype(system_keyspace::client_data::port);
+
+    ck_comparator(schema_ptr schema) : _schema(schema) {}
```
haaawk: Nit: Either take const schema& or do _schema(std::move(schema)).

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+};
+}  // anon. namespace
+
+static future<std::vector<system_keyspace::client_data>> get_all_clients_data(
+        distributed<cql_transport::cql_server>& distributed_cql_server) {
+    return distributed_cql_server.map_reduce(clients_reducer{},
+            [] (const cql_transport::cql_server& srv) {
+        const auto& conn_list = srv.get_connections_list();
+        return do_with(std::vector<system_keyspace::client_data>{},
+                [&conn_list] (auto& per_shard_result) {
```
haaawk: generic lambda

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+            return do_for_each(conn_list.begin(), conn_list.end(),
+                    [&per_shard_result] (const auto& conn) {
```
haaawk: generic lambda

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+                system_keyspace::client_data cd;
+                cd.ip = conn.get_client_state().get_client_address().addr();
+                cd.port = conn.get_client_state().get_client_port();
+                cd.connection_stage = conn.get_client_state().get_auth_state();
+                cd.protocol_version = conn.get_protocol_version();
+                cd.username = conn.get_client_state().user()->name;
+
+                per_shard_result.emplace_back(std::move(cd));
```
haaawk: maybe reserve conn_list.size() for per_shard_resurt

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+            }).then([psr = std::move(per_shard_result)] {
+                return make_ready_future<std::vector<system_keyspace::client_data>>(std::move(psr));
```
haaawk: Do you have to call make_ready_future here? Would it be enough to just return psr?

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+future<std::vector<system_keyspace::client_data>>
+clientlist_mutation_reader::filter_and_sort_by_port(std::vector<system_keyspace::client_data>&& clients) const {
+    std::vector<system_keyspace::client_data> sorted_clients;
+    sorted_clients.reserve(clients.size());
+    
+    std::unordered_map<
+            decltype(system_keyspace::client_data::ip),
+            std::vector<system_keyspace::client_data>> pk_clusters;
+    
+    for (auto&& client : clients) {
```
haaawk: Maybe do_for_each to avoid stalls?

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+        _val.insert(_val.end(), std::make_move_iterator(rhs.begin()), std::make_move_iterator(rhs.end()));
+        return make_ready_future<>();
+    }
+
+    std::vector<system_keyspace::client_data> get() && {
+        return std::move(_val);
+    }
+};
+
+class ck_comparator {
```
kbr-: Suggestion: define this at the point of usage, i.e. inside the filter_and_sort_by_port function. It will be easier to understand why this thing's needed without jumping around the code as much.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+#include "transport/server.hh"
+
+#include <algorithm>
+#include <unordered_map>
+#include <vector>
+
+namespace db::clientlist {
+
+namespace {
+class clients_reducer {
```
kbr-: Suggestion: define this at the point of usage, i.e. inside the get_all_clients_data function. It will be easier to understand why this thing's needed without jumping around the code as much.

db/clientlist_virtual_reader.hh
```diff
@@ -0,0 +1,72 @@
...
+class cql_server;
+}  // namespace cql_transport
+
+namespace db::system_keyspace {
+    struct client_data;
+}  // namespace system_keyspace
+
+namespace db::clientlist {
+
+class clientlist_mutation_reader final : public flat_mutation_reader::impl {
```
kbr-: Please write a comment explaining what data this reader generates.
Here or above the virtual_reader struct, wherever you think would be more suitable.

db/clientlist_virtual_reader.hh
```diff
@@ -0,0 +1,72 @@
...
+            const query::partition_slice& slice);
+            
+    future<> fill_buffer(db::timeout_clock::time_point) override;
+    void next_partition() override;
+    future<> fast_forward_to(const dht::partition_range&, db::timeout_clock::time_point) override;
+    future<> fast_forward_to(position_range, db::timeout_clock::time_point) override;
+    size_t buffer_size() const override;
+};
+
+struct virtual_reader {
```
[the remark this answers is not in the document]
kbr-: This is the actual interface to the reader. The above class is just the implementation.
It's used in system_keyspace.cc, in `maybe_add_virtual_reader`:

```
    if (s.get() == clients().get()) {
        db.find_column_family(s).set_virtual_reader(mutation_source(db::clientlist::virtual_reader()));
    }
```

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+}  // anon. namespace
+
+static future<std::vector<system_keyspace::client_data>> get_all_clients_data(
+        distributed<cql_transport::cql_server>& distributed_cql_server) {
+    return distributed_cql_server.map_reduce(clients_reducer{},
+            [] (const cql_transport::cql_server& srv) {
+        const auto& conn_list = srv.get_connections_list();
+        return do_with(std::vector<system_keyspace::client_data>{},
+                [&conn_list] (auto& per_shard_result) {
+            return do_for_each(conn_list.begin(), conn_list.end(),
```
kbr-: Isn't it possible for the _connections_list inside cql_server to change in-between tasks?
In that case these iterators could be invalidated while `do_for_each` is running.

db/clientlist_virtual_reader.cc
kbr-: In fact, no need to define `psr`. Just:
`return std::move(per_shard_result)`
(I think you have to do the `move` because it's a reference)

db/clientlist_virtual_reader.cc
kbr-: It would be nice to have a dtest which creates a lot of connections and concurrently queries this table.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+
+                per_shard_result.emplace_back(std::move(cd));
+            }).then([psr = std::move(per_shard_result)] {
+                return make_ready_future<std::vector<system_keyspace::client_data>>(std::move(psr));
+            });
+        });
+    });
+}
+
+// Clusters clients by IP and sorts them by port number within clusters.
```
kbr-: Clusters in what sense?

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+future<std::vector<system_keyspace::client_data>>
+clientlist_mutation_reader::filter_and_sort_by_port(std::vector<system_keyspace::client_data>&& clients) const {
```
kbr-: Why does it have "filter" in the name?
Do you mean that it clusters clients by IP and then takes (filters) only the "clusters" (partitions) that belong to this shard?
Please elaborate in the comment above.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+    std::vector<system_keyspace::client_data> sorted_clients;
+    sorted_clients.reserve(clients.size());
+    
+    std::unordered_map<
```
kbr-: So you're defining an `unordered_map`, then copying data from it into a `vector` and sorting the `vector`.
Hmmmm...
https://en.cppreference.com/w/cpp/container/map

db/clientlist_virtual_reader.cc
denesb: You can just change this to make a copy of the conn list to make it safe.

db/clientlist_virtual_reader.cc
denesb: I think he meant "groups".

kbr-: Yes, but this comment doesn't explain how this clustering/grouping manifests:
are all these clusters returned in the vector consecutively?
or is only one cluster returned?
or some subset of them? (that's the case: we return the partitions that belong to this shard)

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+}
+
+future<> clientlist_mutation_reader::maybe_setup_mutation_fragmentator() {
+    if (_mutation_fragmentator_opt) {
+        return make_ready_future<>();
+    }
+
+    return smp::submit_to(0, [] {
+        // It must be run on shard 0. Shards with id>=1 see: get_dist_cql_server()==nullptr
+        return std::ref(*service::get_local_storage_service().get_dist_cql_server());
```
kbr-: no need to construct a `reference_wrapper` here I think. Instead you could declare that the lambda returns a reference:
`submit_to(0, [] () -> distributed<cql_transport::cql_server>& { ... })`

But whatever you prefer. Maybe this one's better because we don't repeat the type so much.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+    }).then([] (distributed<cql_transport::cql_server>& distributed_cql_server) {
```
kbr-: wouldn't `then(get_all_clients_data)` work? Do we have to construct the lambda?

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+
+    for (auto& [ip, cluster] : pk_clusters) {
+        if (_slice.options.contains(query::partition_slice::option::reversed)) {
+            std::sort(
+                    cluster.begin(), cluster.end(),
+                    [] (const auto& client1, const auto& client2) { return client1.port > client2.port; });
+        } else {
+            std::sort(
+                    cluster.begin(), cluster.end(),
+                    [] (const auto& client1, const auto& client2) { return client1.port < client2.port; });
```
kbr-: You could reuse the above comparator and sort `cluster.rbegin(), cluster.rend()` for the reversed case.
Would make the code a bit shorter.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+        }
+        sorted_clients.insert(
+                sorted_clients.end(),
+                make_move_iterator(cluster.begin()),
+                make_move_iterator(cluster.end()));
+    }
+    return make_ready_future<std::vector<system_keyspace::client_data>>(std::move(sorted_clients));
```
kbr-: `make_ready_future` probably not needed here.

db/clientlist_virtual_reader.cc
kbr-: Suggestion: to achieve the token order, you could change `pk_clusters` to an (ordered) `map` with a token comparator.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+        return make_ready_future<>();
+    }
+
+    return smp::submit_to(0, [] {
+        // It must be run on shard 0. Shards with id>=1 see: get_dist_cql_server()==nullptr
+        return std::ref(*service::get_local_storage_service().get_dist_cql_server());
+    }).then([] (distributed<cql_transport::cql_server>& distributed_cql_server) {
+        return get_all_clients_data(distributed_cql_server);
+    }).then([this] (auto&& client_data_vec) {
+        return filter_and_sort_by_port(std::move(client_data_vec)).then([this] (auto&& sorted_client_data_vec) {
```
kbr-: couldn't this `then` be attached one level higher so we don't have this nesting?
also, this lambda calling `filter_and_sort_by_port` is probably not needed, similarly to `get_all_clients_data` above (I'm 100% not sure about that though).
If I'm right, you could do
`then(get_all_clients_data).then(filter_and_sort_by_port).then([this], (... sorted_client_data_vec) { ... })`

kbr-: Also don't take `sorted_client_data_vec` by forwarding reference (after fixing Piotr's comment, by rvalue reference).
Just take it by value.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+            return do_with(
+                    std::vector<mutation>{},
+                    [this, scdv=std::move(sorted_client_data_vec)] (auto&& mutations_vec) mutable {
```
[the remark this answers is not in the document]
kbr-: Also use normal lvalue reference for simplicity (like you did in the `do_with` in `get_all_clients_data`). It's not like we're ever trying to bind to rvalues here.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+                mutations_vec.reserve(scdv.size());
+
+                std::for_each(scdv.begin(), scdv.end(), [this, &mutations_vec] (auto&& client) {
```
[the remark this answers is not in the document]
kbr-: Why is even `do_with` used right now? There are no yields here.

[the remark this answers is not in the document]
kbr-: In case you stick to no yields, a _range for_ would probably be better (shorter, simpler) than `for_each`.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+            });
+        });
+    });
+}
+
+clientlist_mutation_reader::clientlist_mutation_reader(
+        schema_ptr schema,
+        const dht::partition_range& prange,
+        const query::partition_slice& slice)
+    : impl(schema)
```
kbr-: move

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+                    _mutation_fragmentator_opt = std::nullopt;
+                }
+            });
+        });
+    });
+}
+
+void clientlist_mutation_reader::next_partition() {
+    // This method should never be needed.
+    api::unimplemented();
```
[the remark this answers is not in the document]
kbr-: If the comment is left like this, then I agree. Simply throwing `std::bad_function_call()` and deleting this comment would be equally bad to throwing and leaving this comment.

What would be infinitely times better is to explain **why** this method should never be needed.
For some reason `bad_function_call` doesn't take a string argument, so the explanation should be written in the comment.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+}
+
+future<> clientlist_mutation_reader::fast_forward_to(const dht::partition_range& pr, db::timeout_clock::time_point timeout) {
+    // This method should never be needed.
+    api::unimplemented();
```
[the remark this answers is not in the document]
kbr-: Same.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+    return make_ready_future<>();
+}
+
+future<> clientlist_mutation_reader::fast_forward_to(position_range pr, db::timeout_clock::time_point timeout) {
+    // This method should never be needed.
+    api::unimplemented();
```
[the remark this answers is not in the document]
kbr-: Same.

db/clientlist_virtual_reader.cc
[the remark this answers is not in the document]
denesb: For me no explanation is needed (maybe because I saw many readers). Readers, that don't support fast-forwarding (neither of them) don't allow these methods to be called. IMO this is implicit, we have many such readers in the codebase. This is also part of the `flat_mutation_reader` interface contract, although it might not be documented.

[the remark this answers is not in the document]
kbr-: Why doesn't this reader support fast-forwarding?

db/system_keyspace.cc
```diff
@@ -1997,6 +2036,60 @@ future<int> increment_and_get_generation() {
...
+        break;
+    case service::client_state::auth_state::AUTHENTICATION:
+        connection_stage_str = "authenticating";
+        break;
+    case service::client_state::auth_state::READY:
+        connection_stage_str = "ready";
+        break;
+    default:
+        connection_stage_str = to_sstring((uint8_t)cd.connection_stage);
+    }
```
kbr-: Wouldn't a `static const std::map<auth_state, sstring>` or sth like that be more elegant?
But I guess this is just a matter of preference.

kbr-: In any case, I'd suggest to move this code to a `make_connection_stage_string` or something like that.

denesb: A case-switch is more efficient.

db/system_keyspace.hh
```diff
@@ -598,6 +616,10 @@ future<> set_bootstrap_state(bootstrap_state state);
 
     api::timestamp_type schema_creation_timestamp();
 
+/**
+ * Builds a mutation for system.clients containing provided connections.
+ */
+mutation make_client_mutation(client_data&& cd);
```
kbr-: This should probably be a local (static) function in the clientlist_virtual_reader module. No need to pollute `system_keyspace`'s header.

db/clientlist_virtual_reader.cc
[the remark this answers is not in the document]
denesb: Fast-forwarding is only used by the cache, so only readers that are used by cache as an underlying reader need to support it. I think it doesn't make sense to cache virtual tables, but I'm not sure if we actually have an optimization for that. @tgrabiec?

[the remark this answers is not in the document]
kbr-: And that deserves a comment.

db/system_keyspace.cc
```diff
@@ -1997,6 +2036,60 @@ future<int> increment_and_get_generation() {
...
+        connection_stage_str = "established";  // This is C* convention
+        break;
+    case service::client_state::auth_state::AUTHENTICATION:
+        connection_stage_str = "authenticating";
+        break;
+    case service::client_state::auth_state::READY:
+        connection_stage_str = "ready";
+        break;
+    default:
+        connection_stage_str = to_sstring((uint8_t)cd.connection_stage);
```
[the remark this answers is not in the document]
jul-stas: Currently all cases are handled, but I wanted to guarantee at least some form of future-proofness. We don't do that here?

db/system_keyspace.cc
```diff
@@ -427,6 +428,41 @@ schema_ptr built_indexes() {
...
+                {"connection_stage", utf8_type},
+                {"driver_name", utf8_type},
+                {"driver_version", utf8_type},
+                {"hostname", utf8_type},
+                {"protocol_version", int32_type},
+                {"request_count", long_type},
+                {"ssl_cipher_suite", utf8_type},
+                {"ssl_enabled", boolean_type},
+                {"ssl_protocol", utf8_type},
+                {"username", utf8_type}
```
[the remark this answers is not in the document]
jul-stas: I had such a version but ultimately I didn't do that for the sake of C*-compatibility. But if it's really valuable then I'll add it.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// Clusters clients by IP and sorts them by port number within clusters.
+future<std::vector<system_keyspace::client_data>>
+clientlist_mutation_reader::filter_and_sort_by_port(std::vector<system_keyspace::client_data>&& clients) const {
+    std::vector<system_keyspace::client_data> sorted_clients;
+    sorted_clients.reserve(clients.size());
+    
+    std::unordered_map<
+            decltype(system_keyspace::client_data::ip),
```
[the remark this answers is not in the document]
jul-stas: Just to avoid repeating exotic types. I won't fight over it :)

db/system_keyspace.cc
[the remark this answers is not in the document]
denesb: The compiler will warn when new members are added to the enum.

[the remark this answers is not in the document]
denesb: If there is no default that is, so not having `default` is actually safer.

db/system_keyspace.cc
jul-stas: I'm sticking to `switch`, but putting this into another function is a good idea.

db/clientlist_virtual_reader.cc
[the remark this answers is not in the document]
tgrabiec: Reads for tables with virtual readers don't go through cache currently.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+    , _slice(slice) {}
+
+future<> clientlist_mutation_reader::fill_buffer(db::timeout_clock::time_point timeout) {
+    // PKs returned from different shards must be distinct, otherwise combining reader
+    // hangs up. Now: one client can maintain concurrent connections with many different
+    // shards but his IP belongs to partition range of a particular (possibly yet another)
+    // shard. That's why every shard must traverse every connection and return only the
+    // clients with IPs that fall within *this shard's* partition range.
+
+    return with_timeout(timeout, maybe_setup_mutation_fragmentator()).then([this, timeout] {
```
tgrabiec: `with_timeout()` seems wrong, what prevents `clientlist_mutation_reader` from being destroyed after timeout while `maybe_setup_mutation_fragmentator()` is still executing (which results in use-after-free)?

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+        if (!_mutation_fragmentator_opt) {
+            // Setup of `_mutation_fragmentator_opt' failed => there are no connections
+            _end_of_stream = true;
+            return make_ready_future<>();
+        }
+        return do_until([this, timeout] { return is_end_of_stream() || is_buffer_full(); }, [this, timeout] {
```
tgrabiec: Could we just call `fill_buffer_from(*_mutation_fragmentator_opt)`?

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+
+            if (range.contains(ck_prefix, ck_comparator(_schema))) {
+                pk_clusters[client.ip].emplace_back(std::move(client));
+                break;
+            }
+        }
+    }
+
+    for (auto& [ip, cluster] : pk_clusters) {
+        if (_slice.options.contains(query::partition_slice::option::reversed)) {
```
tgrabiec: Implementing a reader is not easy, since the interface puts lots of requirements on the implementation. Why not rely on existing implementation? You could put all the mutations into a temporary `memtable` and return `memtable::make_flat_reader()`.

denesb:
On Tue, 2019-10-15 at 03:47 -0700, Tomasz Grabiec wrote:
> @tgrabiec commented on this pull request.
> 
> 
> 
> 
> In db/clientlist_virtual_reader.cc:
> 
> > +            continue;+        }+        for (auto& range :
> _slice.row_ranges(*_schema, pk)) {+            const auto ck_bytes =
> data_type_for<decltype(system_keyspace::client_data::port+           
>          )>()->decompose(client.port);+            const auto
> ck_prefix = clustering_key_prefix::from_single_value(*_schema,
> ck_bytes);++            if (range.contains(ck_prefix,
> ck_comparator(_schema)))
> {+                pk_clusters[client.ip].emplace_back(std::move(clien
> t));+                break;+            }+        }+    }++    for
> (auto& [ip, cluster] : pk_clusters) {+        if
> (_slice.options.contains(query::partition_slice::option::reversed)) {
> Implementing a reader is not easy, since the interface puts lots of
> requirements on the implementation. Why not rely on existing
> implementation? You could put all the mutations into a temporary
> memtable and return memtable::make_flat_reader().

Another alternative is to just return `reader_from_mutations()` as the
virtual reader. The current solution already uses that as the
underlying reader, after creating the mutations.

db/clientlist_virtual_reader.cc
avikivity: But isn't "all the mutations" equivalent to "all the connections"? This could be larger than a page, so we'd be doing O(nr connections) work instead of O(page's worth of connections) work.

db/clientlist_virtual_reader.cc
```diff
@@ -0,0 +1,237 @@
...
+    }
+    int operator()(const clustering_key_prefix& lhs, const port_type& rhs) {
+        const auto rhs_bytes = data_type_for<port_type>()->decompose(rhs);
+        return operator()(lhs, clustering_key_prefix::from_single_value(*_schema, rhs_bytes));
+    }
+};
+}  // anon. namespace
+
+static future<std::vector<system_keyspace::client_data>> get_all_clients_data(
+        distributed<cql_transport::cql_server>& distributed_cql_server) {
```
avikivity: It's easy for the connection table to be in the thousands per shard. Multiplying by 100 shards, that's hundreds of thousands of objects. That's at the verge of starting to cause problems (stalls/bad_allocs).

We should consider an incremental interface, with this function receiving a last key from which to start. Unfortunately it makes a lot harder and uglier, since cql_server will now need to understand the partition key of this virtual table and maintain connections in sorted order. @tgrabiec any ideas?

avikivity: Maybe we should store this in a non-virtual local table. It adds an extra write on connect/disconnect, but that's nothing, and it's a lot simpler. Also need to truncate on startup.

denesb: I don't think we need a local table. For starts each shard should only emit connection info whose partition key it owns. We can filter out non-belonging connections early.

It would all be much simpler if we used a custom partitioner that sharded connections such that each shards own connections belonging to it. Then connections could be members of an intrusive set/map which takes care of ordering. Maybe at this point a local table is simpler though. Also I'm not sure how well our code is prepared for any partitioner other than the global one. Can the partition even be per table?

avikivity: We're not prepared per-table partitioners at all. Even though we only need to specialize shard_of().

Sure each shard will emit only owned connections, but how to do the merge sort? It's not impossible to solve, but also not easy.

A regular disk-based table solves this and doesn't occupy memory when not in use. A virtual table would be nicer, but needs a large amount of work to get right. Some of that work can be reused with other virtual tables if we manage to make it generic enough, but that's not easy either.

jul-stas:
Thanks everyone for your input. 
I'll turn it into a regular table and get back to you. Until then let's consider this PR being 'on hold'.

PR #5182, opened 2019-10-17
misc_services: Don't gossip hit rates when they are not used

### LINKED ISSUE #1100: Ubuntu AMI rootfs is not XFS
Since we are using Ubuntu's official AMI it is using ext4, we need extra ephemeral disks to construct RAID for /var/lib/scylla.

### LINKED ISSUE #5183: Don't gossip cache hit rates when they are not used
When cache_hit_rate_read_balancing is set to false, heat-weighted load balancing is turned off and we gossip cache hit rates for no good reason. The size of gossiped data is linear on number of tables in the database so it's good to avoid such traffic when it is not used.

Ref https://github.com/scylladb/scylla-enterprise/issues/1092

haaawk:
When cache_hit_rate_read_balancing is set to false,
heat-weighted load balancing is turned off and we
gossip cache hit rates for no good reason.
The size of gossiped data is linear on number of tables
in the database so it's good to avoid such traffic
when it is not used.

Fixes #5183

### DIFF
--- a/database.hh
+++ b/database.hh
@@ -876,7 +876,7 @@ public:
     }
 
     void set_hit_rate(gms::inet_address addr, cache_temperature rate);
-    cache_hit_rate get_hit_rate(gms::inet_address addr);
+    std::optional<cache_hit_rate> get_hit_rate(gms::inet_address addr);
     void drop_hit_rate(gms::inet_address addr);
 
     future<> run_with_compaction_disabled(std::function<future<> ()> func);
--- a/db/consistency_level.cc
+++ b/db/consistency_level.cc
@@ -211,10 +211,12 @@ filter_for_query(consistency_level cl,
     const auto remaining_bf = bf - selected_endpoints.size();
 
     if (cf) {
-        auto get_hit_rate = [cf] (gms::inet_address ep) -> float {
+        bool has_missing_hit_rates = false;
+        auto get_hit_rate = [cf, &has_missing_hit_rates] (gms::inet_address ep) -> float {
             constexpr float max_hit_rate = 0.999;
             auto ht = cf->get_hit_rate(ep);
             if (!ht) {
+                has_missing_hit_rates = true;
                 return 0.0;
             }
             if (float(ht->rate) < 0) {
@@ -241,7 +243,7 @@ filter_for_query(consistency_level cl,
             return std::make_pair(ep, ht);
         }));
 
-        if (!old_node && ht_max - ht_min > 0.01) { // if there is old node or hit rates are close skip calculations
+        if (!has_missing_hit_rates && !old_node && ht_max - ht_min > 0.01) { // if there is old node or hit rates are close skip calculations
             // local node is always first if present (see storage_proxy::get_live_sorted_endpoints)
             unsigned local_idx = epi[0].first == utils::fb_utilities::get_broadcast_address() ? 0 : epi.size() + 1;
             live_endpoints = miss_equalizing_combination(epi, local_idx, remaining_bf, bool(extra));
--- a/db/consistency_level.cc
+++ b/db/consistency_level.cc
@@ -214,15 +214,18 @@ filter_for_query(consistency_level cl,
         auto get_hit_rate = [cf] (gms::inet_address ep) -> float {
             constexpr float max_hit_rate = 0.999;
             auto ht = cf->get_hit_rate(ep);
-            if (float(ht.rate) < 0) {
-                return float(ht.rate);
-            } else if (lowres_clock::now() - ht.last_updated > std::chrono::milliseconds(1000)) {
+            if (!ht) {
+                return 0.0;
+            }
+            if (float(ht->rate) < 0) {
+                return float(ht->rate);
+            } else if (lowres_clock::now() - ht->last_updated > std::chrono::milliseconds(1000)) {
                 // if a cache entry is not updates for a while try to send traffic there
                 // to get more up to date data, mark it updated to not send to much traffic there
-                cf->set_hit_rate(ep, ht.rate);
+                cf->set_hit_rate(ep, ht->rate);
                 return max_hit_rate;
             } else {
-                return std::min(float(ht.rate), max_hit_rate); // calculation below cannot work with hit rate 1
+                return std::min(float(ht->rate), max_hit_rate); // calculation below cannot work with hit rate 1
             }
         };
 
--- a/main.cc
+++ b/main.cc
@@ -1027,12 +1027,18 @@ int main(int ac, char** av) {
                 lb->stop_broadcasting().get();
             });
             supervisor::notify("starting cf cache hit rate calculator");
-            cf_cache_hitrate_calculator.start(std::ref(db), std::ref(cf_cache_hitrate_calculator)).get();
-            auto stop_cache_hitrate_calculator = defer_with_log_on_error([&cf_cache_hitrate_calculator] {
+
+            auto stop_cache_hitrate_calculator = defer_with_log_on_error([&cf_cache_hitrate_calculator, &db] {
+                if (!db.local().get_config().cache_hit_rate_read_balancing()) {
+                    return;
+                }
                 startlog.info("stopping cf cache hit rate calculator");
-                return cf_cache_hitrate_calculator.stop().get();
+                cf_cache_hitrate_calculator.stop().get();
             });
-            cf_cache_hitrate_calculator.local().run_on(engine().cpu_id());
+            if (db.local().get_config().cache_hit_rate_read_balancing()) {
+                cf_cache_hitrate_calculator.start(std::ref(db), std::ref(cf_cache_hitrate_calculator)).get();
+                cf_cache_hitrate_calculator.local().run_on(engine().cpu_id());
+            }
 
             supervisor::notify("starting view update backlog broker");
             static sharded<service::view_update_backlog_broker> view_backlog_broker;
--- a/service/misc_services.cc
+++ b/service/misc_services.cc
@@ -40,6 +40,7 @@
 
 #include "load_broadcaster.hh"
 #include "cache_hitrate_calculator.hh"
+#include "db/config.hh"
 #include "db/system_keyspace.hh"
 #include "gms/application_state.hh"
 #include "service/storage_service.hh"
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -3058,7 +3058,8 @@ storage_proxy::query_partition_key_range_concurrent(storage_proxy::clock_type::t
                     struct {
                         column_family* cf = nullptr;
                         float operator()(const gms::inet_address& ep) const {
-                            return float(cf->get_hit_rate(ep).rate);
+                            auto ht = cf->get_hit_rate(ep);
+                            return ht ? float(ht->rate) : 0.0;
                         }
                     } ep_to_hr{pcf};
                     return *boost::range::min_element(range | boost::adaptors::transformed(ep_to_hr));
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -536,7 +536,8 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
     app_states.emplace(gms::application_state::RPC_ADDRESS, value_factory.rpcaddress(broadcast_rpc_address));
     app_states.emplace(gms::application_state::RELEASE_VERSION, value_factory.release_version());
     app_states.emplace(gms::application_state::SUPPORTED_FEATURES, value_factory.supported_features(features));
-    app_states.emplace(gms::application_state::CACHE_HITRATES, value_factory.cache_hitrates(""));
+    app_states.emplace(gms::application_state::CACHE_HITRATES,
+            value_factory.cache_hitrates(db().local().get_config().cache_hit_rate_read_balancing() ? "" : "DISABLED"));
     app_states.emplace(gms::application_state::SCHEMA_TABLES_VERSION, versioned_value(db::schema_tables::version));
     app_states.emplace(gms::application_state::RPC_READY, value_factory.cql_ready(false));
     app_states.emplace(gms::application_state::VIEW_BACKLOG, versioned_value(""));
--- a/table.cc
+++ b/table.cc
@@ -2220,7 +2220,7 @@ void table::set_hit_rate(gms::inet_address addr, cache_temperature rate) {
     e.last_updated = lowres_clock::now();
 }
 
-table::cache_hit_rate table::get_hit_rate(gms::inet_address addr) {
+std::optional<table::cache_hit_rate> table::get_hit_rate(gms::inet_address addr) {
     auto it = _cluster_cache_hit_rates.find(addr);
     if (utils::fb_utilities::get_broadcast_address() == addr) {
         return cache_hit_rate { _global_cache_hit_rate, lowres_clock::now()};
@@ -2233,6 +2233,9 @@ table::cache_hit_rate table::get_hit_rate(gms::inet_address addr) {
             auto* state = eps->get_application_state_ptr(gms::application_state::CACHE_HITRATES);
             float f = -1.0f; // missing state means old node
             if (state) {
+                if (state->value == "DISABLED") {
+                    return {};
+                }
                 sstring me = format("{}.{}", _schema->ks_name(), _schema->cf_name());
                 auto i = state->value.find(me);
                 if (i != sstring::npos) {
@@ -2244,7 +2247,7 @@ table::cache_hit_rate table::get_hit_rate(gms::inet_address addr) {
                 return cache_hit_rate{cache_temperature(f), lowres_clock::now()};
             }
         }
-        return cache_hit_rate {cache_temperature(0.0f), lowres_clock::now()};
+        return {};
     } else {
         return it->second;
     }


haaawk:
I'm still testing but please have a look in the mean-time.

slivne:
@tgrabiec can you please review this I want to push this to try and solve an existing user problem

tgrabiec:
Another way would be to disable HWLB if any of the replicas doesn't advertise the hitrate

haaawk:
I don't understand the first solution but I can certainly do the second. Will do it tomorrow.

haaawk:
Ok. I think I understood the first solution too :)

slivne:
@tgrabiec / @haaawk we will need to backport this to versions that I don't think the SIGHUP is supported. 

So I prefer the alternative approach. 

in my view Its fine to require a complete upgrade to the newer version - before doing the disable

haaawk:
Makes sense @slivne. @tgrabiec how about a simpler solution that just utilizes features? I would add a heat-load balancing feature and enable it only when it's enabled in the config. That would have exactly the same effect and would be very simple. 

Disabling HWLB if any of the replicas doesn't advertise the hit rates is a bit more complicated because I'm not exactly sure how to distinguish a freshly rebooted node from a node that is not advertising the hit rates. Both send an empty string.

tgrabiec:
On Fri, Oct 18, 2019 at 10:22 AM Piotr Jastrzębski <notifications@github.com>
wrote:

> Makes sense @slivne <https://github.com/slivne>. @tgrabiec
> <https://github.com/tgrabiec> how about a simpler solution that just
> utilizes features? I would add a heat-load balancing feature and enable it
> only when it's enabled in the config. That would have exactly the same
> effect and would be very simple.
>
gms features are monotonic, so you cannot use them for on/off settings.

> Disabling HWLB if any of the replicas doesn't advertise the hit rates is a
> bit more complicated because I'm not exactly sure how to distinguish a
> freshly rebooted node from a node that is not advertising the hit rates.
> Both send an empty string.
>

That's probably because of this line in storage_service.cc:

app_states.emplace(gms::application_state::CACHE_HITRATES,
value_factory.cache_hitrates(""));

Let's make it config-dependent too.

>

haaawk:
While testing this, I got few of those:
`ERROR 2019-10-18 14:00:59,222 [shard 0] storage_proxy - Exception when communicating with 127.0.0.3: std::runtime_error (sstable read queue overloaded)`
Is it a big problem?
The performance of the cluster seems to be alright.

haaawk:
> What's your concurrency?

270 clients

haaawk:
I'm getting the error even when I start the load on a cluster that has heat-load balancing disabled on all nodes so this is probably not an effect of mixing nodes with enabled and disabled balancing.

haaawk:
I tested this PR with the following scenario:
1. I created a cluster of 3 nodes
2. Wrote some data there with cassandra-stress and RF=2
3. Started read load and one by one turned off heat-load balancing on the nodes by adding cache_hit_rate_read_balancing: false to scylla.yaml and restarting the node
4. then I re-enabled heat-load balancing on all the nodes one by one by setting cache_hit_rate_read_balancing: true and restarting the node

Nothing blew up. There were no errors in the logs but few `Exception when communicating with 127.0.0.3: std::runtime_error (sstable read queue overloaded)` just after the restart of the node. The load dropped but then got back to the previous amount.

haaawk:
@tgrabiec I added the improvement you requested so that when any of the nodes participating in the query does not support the heat-weighted load balancing, the load balancing is not used.

I tested it the same way as the previous iteration.

asias:
Before we gossip CACHE_HITRATES  unconditionally. After this patch if user turns on the config option to disable, we do not gossip CACHE_HITRATES at all.

To simplify things, we can still gossip CACHE_HITRATES unconditionally, but we never update the value. If a node does not updates its gossip application state, we will not have the gossip traffic on wire for the node which does not update. For example, we set CACHE_HITRATES to -1f in storage_service.cc during bootstrap. All the other nodes will think this node is a old_node in db/consistency_level.cc::filter_for_query, effectively disable HWLB.

haaawk:
@asias Could you explain more why gossiping unconditionally will make other nodes think the node is an old node? My understanding was that old nodes are distinguished by the fact that they don't have application_state::CACHE_HITRATES set. If we set application_state::CACHE_HITRATES to "" unconditionally then the node will be treated as new node, no?

I guess "If a node does not updates its gossip application state, we will not have the gossip traffic on wire for the node which does not update" should gave me some clue but I'm not sure what it means. Won't gossiping CACHE_HITRATES unconditionally update the state and then that state will be gossipped around so that all other nodes will think it's a new node?

haaawk:
What I don't understand @asias is how I could distinguish freshly restarted node with heat-weighted load balancing enabled from a node that has heat-weighted load balancing disabled, if I gossip unconditionally.

asias:
> @asias Could you explain more why gossiping unconditionally will make other nodes think the node is an old node? My understanding was that old nodes are distinguished by the fact that they don't have application_state::CACHE_HITRATES set. If we set application_state::CACHE_HITRATES to "" unconditionally then the node will be treated as new node, no?

I suggested to set application_state::CACHE_HITRATES to -1 during boot up unconditionally. 

Other node will think this node is a old node because the hitrate < 0.
```
233         auto epi = boost::copy_range<std::vector<std::pair<gms::inet_address, float>>>(live_endpoints | boost::adaptors::transformed([&] (gms::inet_address ep) {
234             auto ht = get_hit_rate(ep);
235             old_node = old_node || ht < 0;
236             ht_max = std::max(ht_max, ht);
237             ht_min = std::min(ht_min, ht);
238             return std::make_pair(ep, ht);
239         }));
```


> 
> I guess "If a node does not updates its gossip application state, we will not have the gossip traffic on wire for the node which does not update" should gave me some clue but I'm not sure what it means. Won't gossiping CACHE_HITRATES unconditionally update the state and then that state will be gossipped around so that all other nodes will think it's a new node?

The goal is to reduce the gossip traffic of CACHE_HITRATES if user disables HWLB, right? We can use the fact gossip will only exchange the application state that changes. If the node only set CACHE_HITRATES once, it will be exchanged once and only once.

asias:
> What I don't understand @asias is how I could distinguish freshly restarted node with heat-weighted load balancing enabled from a node that has heat-weighted load balancing disabled, if I gossip unconditionally.
```
if (user wants to gossip hit rate) {
    app_states.emplace(gms::application_state::CACHE_HITRATES, value_factory.cache_hitrates("0.0"));
} else {
    app_states.emplace(gms::application_state::CACHE_HITRATES, value_factory.cache_hitrates("-1"));
}
```

When we get 0, we know it is restarted and it will update the hit rate.
When we get -1, we know it is restarted but it will not update hit rate.

haaawk:
> I suggested to set application_state::CACHE_HITRATES to -1 during boot up unconditionally.
Other node will think this node is a old node because the hitrate < 0.

I may be wrong but if we set application_state::CACHE_HITRATES to -1 during boot up unconditionally then a freshly restarted node that supports heat-weighted load balancing will cause heat-weighted load balancing to be disabled and that will defeat the purpose of the load balancing, no?

haaawk:
```
if (user wants to gossip hit rate) {
    app_states.emplace(gms::application_state::CACHE_HITRATES, value_factory.cache_hitrates("0.0"));
} else {
    app_states.emplace(gms::application_state::CACHE_HITRATES, value_factory.cache_hitrates("-1"));
}
```

If that's what you called 'unconditionally' then I agree :). The only thing is that setting "0.0" or "-1.0" is not correct I think. It seems that cache hit rates are encoded per table so we would have to add <tablename> 0.0 or <tablename> -1.0 for each table.

asias:
> If that's what you called 'unconditionally' then I agree :). 

It is unconditionally in the sense we always inject the application state and gossip.

> The only thing is that setting "0.0" or "-1.0" is not correct I think. It seems that cache hit rates are encoded per table so we would have to add 0.0 or -1.0 for each table.

Sure.  For 0 we can still use the empty string, for -1 we need to specify the tables. 

Btw, I suggested to compress the value of the CACHE_HITRATES application state in other thread with Vlad to further reduce the message size.

Looking at table::cache_hit_rate table::get_hit_rate

```
215 table::cache_hit_rate table::get_hit_rate(gms::inet_address addr) {
2216     auto it = _cluster_cache_hit_rates.find(addr);
2217     if (utils::fb_utilities::get_broadcast_address() == addr) {
2218         return cache_hit_rate { _global_cache_hit_rate, lowres_clock::now()};
2219     }
2220     if (it == _cluster_cache_hit_rates.end()) {
2221         // no data yet, get it from the gossiper
2222         auto& gossiper = gms::get_local_gossiper();
2223         auto* eps = gossiper.get_endpoint_state_for_endpoint_ptr(addr);
2224         if (eps) {
2225             auto* state = eps->get_application_state_ptr(gms::application_state::CACHE_HITRATES);
2226             float f = -1.0f; // missing state means old node
2227             if (state) {
2228                 sstring me = format("{}.{}", _schema->ks_name(), _schema->cf_name());
2229                 auto i = state->value.find(me);
2230                 if (i != sstring::npos) {
2231                     f = strtof(&state->value[i + me.size() + 1], nullptr);
2232                 } else {
2233                     f = 0.0f; // empty state means that node has rebooted
2234                 }
2235                 set_hit_rate(addr, cache_temperature(f)); 
2236                 return cache_hit_rate{cache_temperature(f), lowres_clock::now()};
2237             }
2238         }
2239         return cache_hit_rate {cache_temperature(0.0f), lowres_clock::now()};
2240     } else {
2241         return it->second;
2242     }
2243 }
```

I think we can compress and send the compressed value within a mixed cluster. For the old node, it will think the old was restart because in line 2230, it won't be able to find the ks.table string within the compressed value. After the upgrade is done, every nodes enjoys the compressed hit rate value.

haaawk:
Even wore problem arrives for tables that are created after the node is started. For such table the node with disabled heat-weighed load balancing won't be distinguishable from a freshly restarted node.

haaawk:
> send the compressed value within a mixed cluster. For the old node, it will think the old was restart because in line 2230, it won't be able to find the ks.table string within the

I think that's out of the scope of this change.

asias:
Btw, line 2226 `float f = -1.0f; // missing state means old node`, the default value is never used!

haaawk:
> Btw, line 2226 `float f = -1.0f; // missing state means old node`, the default value is never used!

Yeah - I know.

haaawk:
> > I think that's out of the scope of this change.
> 
> Sure, but I hope this change to be "compatible" with the compression idea.

Sure but the approach that sends unconditionally and later never updates won't work because tables can be created after the node starts.

asias:
> Even wore problem arrives for tables that are created after the node is started. For such table the node with disabled heat-weighed load balancing won't be distinguishable from a freshly restarted node.

Without updating the scylla node to newer version, existing node (with old code) will think such new table has a hit rate of 0 anyway.

Perhaps, we can sent 

gms::application_state::CACHE_HITRATES = "DISABLED"

asias:
> > Btw, line 2226 `float f = -1.0f; // missing state means old node`, the default value is never used!
> 
> Yeah - I know.

This means existing nodes (without any patches from hit rate issue fixes) will think a node was restarted if we gossip gms::application_state::CACHE_HITRATES = "" or do not gossip gms::application_state at all. So, not to gossip gms::application_state if user wants to disable HWLB in this PR will not work with existing nodes.

This gives another reason to redesign the CACHE_HITRATES protocol.

Perhaps:
```
1) during boot up

gms::application_state::CACHE_HITRATES = "REBOOTED"

or

gms::application_state::CACHE_HITRATES = "DISABLED"

2) during normal update
gms::application_state::CACHE_HITRATES = "UPDATE: compress(t1=0.1, t2=0.2, ...)"
```

haaawk:
I think it's better to do:
```
1) during boot up
gms::application_state::CACHE_HITRATES = "" for rebooted
or
not gossiping gms::application_state::CACHE_HITRATES for disabled
2) during normal update
whatever the protocol is for gms::application_state::CACHE_HITRATES
```
It does not seem to bring any value to add "REBOOTED", "DISABLED" and "UPDATE:" because this is already easy to distinguish.

> So, not to gossip gms::application_state if user wants to disable HWLB in this PR will not work with existing nodes

changing CACHE_HITRATES protocol won't work with existing nodes as well.

haaawk:
> > Even wore problem arrives for tables that are created after the node is started. For such table the node with disabled heat-weighed load balancing won't be distinguishable from a freshly restarted node.
> 
> Without updating the scylla node to newer version, existing node (with old code) will think such new table has a hit rate of 0 anyway.
> 
> Perhaps, we can send:
> gms::application_state::CACHE_HITRATES = "DISABLED"

Sending "DISABLED" won't help because old nodes won't understand it either and will still treat a node as restarted, no?

asias:
> I think it's better to do:
> 
> ```
> 1) during boot up
> gms::application_state::CACHE_HITRATES = "" for rebooted
> or
> not gossiping gms::application_state::CACHE_HITRATES for disabled
> 2) during normal update
> whatever the protocol is for gms::application_state::CACHE_HITRATES
> ```
> 
> It does not seem to bring any value to add "REBOOTED", "DISABLED" and "UPDATE:" because this is already easy to distinguish.

I think use "not to gossip CACHE_HITRATES' to express that this node has disabled the HWLB is not a good idea. It looks like a hack to me.

> 
> > So, not to gossip gms::application_state if user wants to disable HWLB in this PR will not work with existing nodes
> 
> changing CACHE_HITRATES protocol won't work with existing nodes as well.

The point is existing node won't work without using newer scylla binary anyway, so we can work on a solution that does not use special cases to communicate the information like this node has disabled CACHE_HITRATES.

Think a user runs nodetool gossipinfo and sees CACHE_HITRATES is missing for a node. Is that node a old node that does not support CACHE_HITRATES or the node is explicitly disabled CACHE_HITRATES. If we express it explicitly we know for sure.

haaawk:
> Think a user runs nodetool gossipinfo and sees CACHE_HITR

Makes sense but since the new code will have to handle both old version and new version I don't think it's a good idea to introduce the new protocol now when we still don't know what it's gonna be. I would much more prefer go with a simple change here and let the protocol be changed once it's agreed on.

At the moment, if a node does not gossip gms::application_state::CACHE_HITRATES it means it does not support heat-weighted load balancing. Before this change it was only due to old version of the code and after this change it's also because of the config flag. So I'm not touching the protocol here. I'm just making use of it.

If you want to change the protocol, it's fine but I think it's not in scope of this change.

asias:
> Makes sense but since the new code will have to handle both old version and new version I don't think it's a good idea to introduce the new protocol now when we still don't know what it's gonna be. I would much more prefer go with a simple change here and let the protocol be changed once it's agreed on.
> 
> At the moment, if a node does not gossip gms::application_state::CACHE_HITRATES it means it does not support heat-weighted load balancing. Before this change it was only due to old version of the code and after this change it's also because of the config flag. So I'm not touching the protocol here. I'm just making use of it.

This patch introduces a new case as you described, the missing of CACHE_HITRATES can be now due to user disabling the config flag. 

Later, we need to be compatible with this newly introduced case as well. Instead of handling more cases in later, we can avoid such cases.

Just gossip CACHE_HITRATES = "DISABLED" in case it is disabled, which is similar to  CACHE_HITRATES = "" in case it is rebooted. They are more consistent. 

This change is minimal as well. 

> 
> If you want to change the protocol, it's fine but I think it's not in scope of this change.

haaawk:
> > Makes sense but since the new code will have to handle both old version and new version I don't think it's a good idea to introduce the new protocol now when we still don't know what it's gonna be. I would much more prefer go with a simple change here and let the protocol be changed once it's agreed on.
> > At the moment, if a node does not gossip gms::application_state::CACHE_HITRATES it means it does not support heat-weighted load balancing. Before this change it was only due to old version of the code and after this change it's also because of the config flag. So I'm not touching the protocol here. I'm just making use of it.
> 
> This patch introduces a new case as you described, the missing of CACHE_HITRATES can be now due to user disabling the config flag.
> 
> Later, we need to be compatible with this newly introduced case as well. Instead of handling more cases in later, we can avoid such cases.

It seems that both cases are handled the same way so there's no need to distinguish them. In both cases we want heat-weighted load balancing to be turned off.
The case is the same -> node does not support heat-weighted load balancing. Whether it's because of the old code or because of the config flag, makes no difference.
Unless there's a reason we need to distinguish the two that I don't know.
Do you know any @asias?

> 
> Just gossip CACHE_HITRATES = "DISABLED" in case it is disabled, which is similar to CACHE_HITRATES = "" in case it is rebooted. They are more consistent.
> 
> This change is minimal as well.
> 
> > If you want to change the protocol, it's fine but I think it's not in scope of this change.

asias:
> It seems that both cases are handled the same way so there's no need to distinguish them. In both cases we want heat-weighted load balancing to be turned off.
> The case is the same -> node does not support heat-weighted load balancing. Whether it's because of the old code or because of the config flag, makes no difference.
> Unless there's a reason we need to distinguish the two that I don't know.
> Do you know any @asias?

I agree that the end result is the same - both turning off HWLB. The difference is user can know for sure what is the real reason CACHE_HITRATES is not seen. We have CACHE_HITRATES support since 2.0 release, we send CACHE_HITRATES unconditionally since 2.0. Practically, we can assume CACHE_HITRATES is always present. I hope we do not change this behavior and have to handle the corner case that  CACHE_HITRATES misses.

if we gossip CACHE_HITRATES = "DISABLED"  when user disables it, it is both easier to reason about the code and the cluster (e.g., when you run nodetool gossipinfo, people know user disables it , not due to a bug in gossip that CACHE_HITRATES is not propagated). 

How do you think?

asias:
Btw, what happens if a node does not receive the CACHE_HITRATES = "" but receives the later CACHE_HITRATES = "table with real hit rate" directly which could happen. Will this node know the peer node has rebooted?

haaawk:
> > It seems that both cases are handled the same way so there's no need to distinguish them. In both cases we want heat-weighted load balancing to be turned off.
> > The case is the same -> node does not support heat-weighted load balancing. Whether it's because of the old code or because of the config flag, makes no difference.
> > Unless there's a reason we need to distinguish the two that I don't know.
> > Do you know any @asias?
> 
> I agree that the end result is the same - both turning off HWLB. The difference is user can know for sure what is the real reason CACHE_HITRATES is not seen. We have CACHE_HITRATES support since 2.0 release, we send CACHE_HITRATES unconditionally since 2.0. Practically, we can assume CACHE_HITRATES is always present. I hope we do not change this behavior and have to handle the corner case that CACHE_HITRATES misses.
> 
> if we gossip CACHE_HITRATES = "DISABLED" when user disables it, it is both easier to reason about the code and the cluster (e.g., when you run nodetool gossipinfo, people know user disables it , not due to a bug in gossip that CACHE_HITRATES is not propagated).
> 
> How do you think?

User-friendliness sounds like a good reason to do it so I will add the DISABLED as you requested.

haaawk:
> Btw, what happens if a node does not receive the CACHE_HITRATES = "" but receives the later CACHE_HITRATES = "table with real hit rate" directly which could happen. Will this node know the peer node has rebooted?

I think the node will treat the stats it got as the source of truth and won't think it's a rebooted node. It is fine though because if the peer was actually rebooted, its hit rates will reflect the cold caches.

nyh:
A sacrilegious question:
It seems the reason why we want the ability to turn off HWLB is because it gossips a separate metric per table. So **could we fix that issue instead**? If there are too many tables (or perhaps even always?) we could gossip an average number for all tables instead of one per table. If we calculate this average correctly (i.e., weighted by the actual number of requests to each table - or in other words simply don't calculate the average separately for each table but just once for all table), I believe the overall HWLB will work as intended, won't it? @gleb-cloudius? If this idea has merit, I can open a new issue for it.

The reason I'm asking this is the spirit of "zero configuration database". It is against our spirit to ask the user to notice, when the number of tables are high, that gossip becomes overwhelming and then manually turn off an option. This option may be useful as a bandaid to solve a specific problem of one client quickly - but I think we should not encourage it in general.

haaawk:
It's a valid question @nyh. This fix does not fix the problem entirely. My understanding is that there's a customer struggling with this issue and we want to help them ASAP. So this is a quick fix, not the final solution. BTW config option to disable HWLB was there before and just wasn't working so this patchset just fixes it and does not add anything new to the config.

asias:
> A sacrilegious question:
> It seems the reason why we want the ability to turn off HWLB is because it gossips a separate metric per table. So **could we fix that issue instead**? If there are too many tables (or perhaps even always?) we could gossip an average number for all tables instead of one per table. If we calculate this average correctly (i.e., weighted by the actual number of requests to each table - or in other words simply don't calculate the average separately for each table but just once for all table), I believe the overall HWLB will work as intended, won't it? @gleb-cloudius? If this idea has merit, I can open a new issue for it.
> 
> The reason I'm asking this is the spirit of "zero configuration database". It is against our spirit to ask the user to notice, when the number of tables are high, that gossip becomes overwhelming and then manually turn off an option. This option may be useful as a bandaid to solve a specific problem of one client quickly - but I think we should not encourage it in general.

So, I suggested to compress the hit rate string with strong compression level to reduce the size. I thought about a more efficient representation. I liked the idea to use the single average to represent all tables which reduces the size to constant ;-)

haaawk:
I added the change requested by @asias to send "DISABLED" instead of an empty string when HWLB is disabled.

asias:
> It's a valid question @nyh. This fix does not fix the problem entirely. My understanding is that there's a customer struggling with this issue and we want to help them ASAP. So this is a quick fix, not the final solution. BTW config option to disable HWLB was there before and just wasn't working so this patchset just fixes it and does not add anything new to the config.

Yes, I agree. Improving the HWLB protocol is beyond the scope this patch. This patch fixes a valid issue.

haaawk:
@tgrabiec would you mind merging?

nyh:
I opened issue #5200 about the idea of gossiping just one per-node statistic instead of per-table statistic.

asias:
On Tue, Oct 22, 2019 at 5:44 PM nyh <notifications@github.com> wrote:

> I opened issue #5200 <https://github.com/scylladb/scylla/issues/5200>
> about the idea of gossiping just one per-node statistic instead of
> per-table statistic.
>

Good. Thanks!

-- 
Asias

haaawk:
Please don't merge this. I think I found an issue with the code.

haaawk:
I fixed the place where gossiping hit rates is disabled. Now it prevents cf_cache_hitrate_calculator from starting in main.cc and load is being gossiped as before. I will retest now.

haaawk:
I'm closing this and opening a new PR against next here: https://github.com/scylladb/scylla/pull/5204

PR #5191, opened 2019-10-20
dist/ami/build_ami.sh: support incremental build of rpms

bhalevy:
If build_rpms was run manually and rpms were built
but not copied to dist/ami/files then copy them
before checking if any rpm is missing.

### DIFF
--- a/dist/ami/build_ami.sh
+++ b/dist/ami/build_ami.sh
@@ -73,14 +73,25 @@ if [ $LOCALRPM -eq 1 ]; then
         pkg_install git
     fi
 
-    if [ ! -f dist/ami/files/$PRODUCT.x86_64.rpm ] || [ ! -f dist/ami/files/$PRODUCT-kernel-conf.x86_64.rpm ] || [ ! -f dist/ami/files/$PRODUCT-conf.x86_64.rpm ] || [ ! -f dist/ami/files/$PRODUCT-server.x86_64.rpm ] || [ ! -f dist/ami/files/$PRODUCT-debuginfo.x86_64.rpm ]; then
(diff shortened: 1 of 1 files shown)


dist/ami/build_ami.sh
```diff
@@ -73,14 +73,25 @@ if [ $LOCALRPM -eq 1 ]; then
         pkg_install git
     fi
```
avikivity: I don't understand the goal. What is build_rpms?

dist/ami/build_ami.sh
```diff
@@ -73,14 +73,25 @@ if [ $LOCALRPM -eq 1 ]; then
...
 
-    if [ ! -f dist/ami/files/$PRODUCT.x86_64.rpm ] || [ ! -f dist/ami/files/$PRODUCT-kernel-conf.x86_64.rpm ] || [ ! -f dist/ami/files/$PRODUCT-conf.x86_64.rpm ] || [ ! -f dist/ami/files/$PRODUCT-server.x86_64.rpm ] || [ ! -f dist/ami/files/$PRODUCT-debuginfo.x86_64.rpm ]; then
+    rpm_names="$PRODUCT $PRODUCT-kernel-conf $PRODUCT-conf $PRODUCT-server $PRODUCT-debuginfo"
```
avikivity: Please use arrays for lists of stuff, they also work if something has a space in it.

dist/ami/build_ami.sh
bhalevy: build_rpm is the script that generates the rpms build_ami uses for the ami when given the `--localrpm` option

amoskong: When `--localrpm` option is assigned for build_ami.sh, it will use build_rpms.sh to build scylla rpms if some scylla packages is missing in ami files directory (dist/ami/files/).

glommer:
@penberg please merge.

PR #5197, opened 2019-10-21
CDC (atomic) delta + (non-optional) pre-image data columns

elcallio:
Adds delta and pre-image data column writes for the atomic columns in a cdc-enabled table. 

Note that in this patch set it is still unconditional. Adding option support comes in next set. 

Uses code more or less derived from alternator to select pre-image, using raw query interface. So should be fairly low overhead to query generation. Pre-image and delta mutations are mixed in with the actual modification mutations to generate the full cdc log (sans post-image). 

Rebased on lastest master as of 2019-10-21.

### DIFF
--- a/cdc/cdc.cc
+++ b/cdc/cdc.cc
@@ -39,6 +39,7 @@
 #include "cql3/statements/select_statement.hh"
 #include "cql3/multi_column_relation.hh"
 #include "cql3/tuples.hh"
+#include "log.hh"
 
 using locator::snitch_ptr;
 using locator::token_metadata;
@@ -58,6 +59,8 @@ template<> struct hash<std::pair<net::inet_address, unsigned int>> {
 }
 
 using namespace std::chrono_literals;
+
+static logging::logger cdc_log("cdc");
 
 namespace cdc {
 
@@ -335,6 +338,34 @@ public:
                     res.set_cell(log_ck, *cdef, std::move(value));
                     ++pos;
                 }
+
+                std::vector<bytes_opt> values(3);
+
+                r.row().cells().for_each_cell([&](column_id id, const atomic_cell_or_collection& cell) {
+                    auto& cdef = _schema->column_at(column_kind::regular_column, id);
+                    auto* dst = _log_schema->get_column_definition(to_bytes("_" + cdef.name()));
+                    // todo: collections.
+                    if (cdef.is_atomic()) {
+                        column_op op;
+
+                        values[1] = values[2] = std::nullopt;
+                        auto view = cell.as_atomic_cell(cdef);
+                        if (view.is_live()) {
+                            op = column_op::set;
+                            values[1] = view.value().linearize();
+                            if (view.is_live_and_has_ttl()) {
+                                values[2] = long_type->decompose(data_value(view.ttl().count()));
+                            }
+                        } else {
+                            op = column_op::del;
+                        }
+
+                        values[0] = data_type_for<column_op_native_type>()->decompose(data_value(static_cast<column_op_native_type>(op)));
+                        res.set_cell(log_ck, *dst, atomic_cell::make_live(*dst->type, _time.timestamp(), tuple_type_impl::build_value(values)));
+                    } else {
+                        cdc_log.warn("Non-atomic cell ignored {}.{}:{}", _schema->ks_name(), _schema->cf_name(), cdef.name_as_text());
+                    }
+                });
 
                 set_operation(log_ck, operation::update, res);
                 ++batch_no;
--- a/cdc/cdc.cc
+++ b/cdc/cdc.cc
@@ -33,6 +33,7 @@
 #include "schema_builder.hh"
 #include "service/migration_manager.hh"
 #include "service/storage_service.hh"
+#include "types/tuple.hh"
 
 using locator::snitch_ptr;
 using locator::token_metadata;
@@ -53,6 +54,9 @@ template<> struct hash<std::pair<net::inet_address, unsigned int>> {
 
 
 namespace cdc {
+
+using operation_native_type = std::underlying_type_t<operation>;
+using column_op_native_type = std::underlying_type_t<column_op>;
 
 sstring log_name(const sstring& table_name) {
     static constexpr auto cdc_log_suffix = "_scylla_cdc_log";
@@ -103,17 +107,21 @@ static future<> setup_log(db_context ctx, const schema& s) {
     b.with_column("stream_id", uuid_type, column_kind::partition_key);
     b.with_column("time", timeuuid_type, column_kind::clustering_key);
     b.with_column("batch_seq_no", int32_type, column_kind::clustering_key);
-    b.with_column("operation", int32_type);
+    b.with_column("operation", data_type_for<operation_native_type>());
     b.with_column("ttl", long_type);
-    auto add_columns = [&] (const schema::const_iterator_range_type& columns) {
+    auto add_columns = [&] (const schema::const_iterator_range_type& columns, bool is_data_col = false) {
         for (const auto& column : columns) {
-            b.with_column("_" + column.name(), column.type);
+            auto type = column.type;
+            if (is_data_col) {
+                type = tuple_type_impl::get_instance({ /* op */ data_type_for<column_op_native_type>(), /* value */ type, /* ttl */long_type});
+            }
+            b.with_column("_" + column.name(), type);
         }
     };
     add_columns(s.partition_key_columns());
     add_columns(s.clustering_key_columns());
-    add_columns(s.static_columns());
-    add_columns(s.regular_columns());
+    add_columns(s.static_columns(), true);
+    add_columns(s.regular_columns(), true);
     return ctx._migration_manager.announce_new_column_family(b.build(), false);
 }
 
--- a/cdc/cdc.cc
+++ b/cdc/cdc.cc
@@ -20,6 +20,8 @@
  */
 
 #include <utility>
+#include <algorithm>
+
 #include <seastar/util/defer.hh>
 #include <seastar/core/thread.hh>
 
@@ -34,6 +36,9 @@
 #include "service/migration_manager.hh"
 #include "service/storage_service.hh"
 #include "types/tuple.hh"
+#include "cql3/statements/select_statement.hh"
+#include "cql3/multi_column_relation.hh"
+#include "cql3/tuples.hh"
 
 using locator::snitch_ptr;
 using locator::token_metadata;
@@ -52,6 +57,7 @@ template<> struct hash<std::pair<net::inet_address, unsigned int>> {
 
 }
 
+using namespace std::chrono_literals;
 
 namespace cdc {
 
@@ -322,6 +328,71 @@ public:
             }
         }
         return res;
+    }
+
+    static db::timeout_clock::time_point default_timeout() {
+        return db::timeout_clock::now() + 10s;
+    }
+
+    future<lw_shared_ptr<cql3::untyped_result_set>> pre_image_select(
+            service::storage_proxy& proxy,
+            service::client_state& client_state,
+            db::consistency_level cl,
+            const mutation& m)
+    {
+        auto& p = m.partition();
+        if (p.partition_tombstone() || !p.row_tombstones().empty() || p.clustered_rows().empty()) {
+            return make_ready_future<lw_shared_ptr<cql3::untyped_result_set>>();
+        }
+
+        dht::partition_range_vector partition_ranges{dht::partition_range(m.decorated_key())};
+
+        auto&& pc = _schema->partition_key_columns();
+        auto&& cc = _schema->clustering_key_columns();
+
+        std::vector<query::clustering_range> bounds;
+        if (cc.empty()) {
+            bounds.push_back(query::clustering_range::make_open_ended_both_sides());
+        } else {
+            for (const rows_entry& r : p.clustered_rows()) {
+                auto& ck = r.key();
+                bounds.push_back(query::clustering_range::make_singular(ck));
+            }
+        }
+
+        std::vector<const column_definition*> columns;
+        columns.reserve(_schema->all_columns().size());
+
+        std::transform(pc.begin(), pc.end(), std::back_inserter(columns), [](auto& c) { return &c; });
+        std::transform(cc.begin(), cc.end(), std::back_inserter(columns), [](auto& c) { return &c; });
+
+        query::column_id_vector static_columns, regular_columns;
+
+        auto sk = column_kind::static_column;
+        auto rk = column_kind::regular_column;
+        // TODO: this assumes all mutations touch the same set of columns. This might not be true, and we may need to do more horrible set operation here. 
+        for (auto& [r, cids, kind] : { std::tie(p.static_row(), static_columns, sk), std::tie(p.clustered_rows().begin()->row().cells(), regular_columns, rk) }) {
+            r.for_each_cell([&](column_id id, const atomic_cell_or_collection&) {
+                auto& cdef =_schema->column_at(kind, id);
+                cids.emplace_back(id);
+                columns.emplace_back(&cdef);
+            });
+        }
+
+        auto selection = cql3::selection::selection::for_columns(_schema, std::move(columns));
+        auto partition_slice = query::partition_slice(std::move(bounds), std::move(static_columns), std::move(regular_columns), selection->get_query_options());
+        auto command = ::make_lw_shared<query::read_command>(_schema->id(), _schema->version(), partition_slice, query::max_partitions);
+
+        return proxy.query(_schema, std::move(command), std::move(partition_ranges), cl, service::storage_proxy::coordinator_query_options(default_timeout(), empty_service_permit(), client_state)).then(
+                [this, partition_slice = std::move(partition_slice), selection = std::move(selection)] (service::storage_proxy::coordinator_query_result qr) -> lw_shared_ptr<cql3::untyped_result_set> {
+                    cql3::selection::result_set_builder builder(*selection, gc_clock::now(), cql_serialization_format::latest());
+                    query::result_view::consume(*qr.query_result, partition_slice, cql3::selection::result_set_builder::visitor(builder, *_schema, *selection));
+                    auto result_set = builder.build();
+                    if (!result_set || result_set->empty()) {
+                        return {};
+                    }
+                    return make_lw_shared<cql3::untyped_result_set>(*result_set);
+        });
     }
 };
 
--- a/cdc/cdc.cc
+++ b/cdc/cdc.cc
@@ -232,7 +232,7 @@ private:
     schema_ptr _log_schema;
     utils::UUID _time;
     bytes _decomposed_time;
-    const streams_type& _streams;
+    ::shared_ptr<const transformer::streams_type> _streams;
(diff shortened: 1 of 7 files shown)


cdc/cdc.cc
```diff
@@ -534,15 +573,20 @@ future<std::vector<mutation>> append_log_mutations(
...
-        }
-        return std::move(muts);
+    auto mp = ::make_lw_shared<std::vector<mutation>>(std::move(muts));
+
+    return get_streams(ctx, s->ks_name(), s->cf_name(), timeout, qs).then([ctx, s = std::move(s), mp, &qs](::shared_ptr<transformer::streams_type> streams) mutable {
+        mp->reserve(2 * mp->size());
+        auto trans = make_lw_shared<transformer>(ctx, s, std::move(streams));
+        auto i = mp->begin();
+        auto e = mp->end();
+        return parallel_for_each(i, e, [ctx, &qs, trans, mp](mutation& m) {
```
nyh: Something to think about: Could we start here too many things parallel if "mp" has many mutations? Is this a likely occurance, or usually mp has only a small number of mutations?

tests/cql_assertions.cc
```diff
@@ -180,7 +180,7 @@ rows_assertions rows_assertions::with_serialized_columns_count(size_t columns_co
 }
 
 shared_ptr<cql_transport::messages::result_message> cquery_nofail(
-        cql_test_env& env, const char* query, std::unique_ptr<cql3::query_options>&& qo, const std::experimental::source_location& loc) {
+        cql_test_env& env, const seastar::sstring& query, std::unique_ptr<cql3::query_options>&& qo, const std::experimental::source_location& loc) {
```
nyh: This change fairly harmless to me (a little bit more work for the compiler...), but @dekimir this is code you wrote, can you please check if you're fine with it too?

nyh:
@haaawk can you please review this too? Looks mostly good to me (I only left a few requests), but you're the real expert in this so please take a look too, and when @elcallio responds (perhaps with a new forced push) and you give the ok too, I can commit this series.

cdc/cdc.hh
```diff
@@ -166,6 +166,21 @@ struct db_context final {
...
+// cdc log table operation
+enum class operation : int8_t {
+    // note: these values will eventually be read by a third party, probably not privvy to this
+    // enum decl, so don't change the constant values. 
+    pre_image = 0, update = 1, row_delete = 2, range_delete_start = 3, range_delete_end = 4, partition_delete = 5
+};
+
+// cdc log data column operation
+enum class column_op : int8_t {
+    // same as above
```
haaawk: better to reference operation enum by name to make it safe in case when someone inserts another code in between.

cdc/cdc.hh
```diff
@@ -166,6 +166,21 @@ struct db_context final {
...
+    set = 0, del = 1, add = 2,
+};
+
+seastar::future<> setup(service::storage_proxy& proxy, schema_ptr schema);
```
haaawk: Are you sure this is correctly rebased? on master this method is already present in cdc.hh -> https://github.com/scylladb/scylla/blob/master/cdc/cdc.hh#L167

haaawk: It uses db_context on master though.

cdc/cdc.cc
```diff
@@ -103,17 +107,21 @@ static future<> setup_log(db_context ctx, const schema& s) {
     b.with_column("stream_id", uuid_type, column_kind::partition_key);
     b.with_column("time", timeuuid_type, column_kind::clustering_key);
     b.with_column("batch_seq_no", int32_type, column_kind::clustering_key);
-    b.with_column("operation", int32_type);
+    b.with_column("operation", data_type_for<operation_native_type>());
```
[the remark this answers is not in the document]
elcallio: What user needs to know the type? It will be an int-like type, describe schema will tell him. Yes, changing the underlying type of "operation" enum type would change this, but it is better that the definition is at _one_ place, i.e. in "enum class operation : int32_t" than spread out in several. 
Dependent types is a good thing.

cql3/untyped_result_set.cc
```diff
@@ -69,25 +69,32 @@ bool cql3::untyped_result_set_row::has(const sstring& name) const {
 
 using cql_transport::messages::result_message;
 
+cql3::untyped_result_set::untyped_result_set(const cql3::result_set& rs) {
+    auto& cn = rs.get_metadata().get_names();
+    for (auto& r : rs.rows()) {
+        _rows.emplace_back(cn, std::move(r));
```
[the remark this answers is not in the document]
elcallio: cn == column_names. As in the result of "get_names()". As in a map name->column index. And yes, right now this is duplicated for each row. Because untyped res-set was initially meant for more or less single-row, cold code, usage. 
Making this type more efficient is on the todo list.

cdc/cdc.cc
```diff
@@ -323,6 +329,75 @@ class transformer final {
...
+    }
+
+    future<lw_shared_ptr<cql3::untyped_result_set>> pre_image_select(
+            service::storage_proxy& proxy,
+            service::client_state& client_state,
+            db::consistency_level cl,
+            const mutation& m)
+    {
+        auto& p = m.partition();
+        if (p.partition_tombstone() || !p.row_tombstones().empty() || p.clustered_rows().empty()) {
```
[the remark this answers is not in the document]
haaawk: How would you create a mutation with an old tombstone and newer data with a CQL statement @nyh. This mutation here is restricted only to mutations that can be created with a single user operation.

cql3/untyped_result_set.cc
```diff
@@ -69,25 +69,32 @@ bool cql3::untyped_result_set_row::has(const sstring& name) const {
...
+    }
+}
+
 cql3::untyped_result_set::untyped_result_set(::shared_ptr<result_message> msg)
     : _rows([msg]{
     class visitor : public result_message::visitor_base {
     public:
-        rows_type rows;
+        std::optional<untyped_result_set> res;
```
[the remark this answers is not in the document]
elcallio: To provide two different ways to construct untyped result set. This changes this signature constructor to use the other. I.e. reduce code duplication.

cdc/cdc.cc
```diff
@@ -323,6 +329,75 @@ class transformer final {
...
+            return make_ready_future<lw_shared_ptr<cql3::untyped_result_set>>();
+        }
+
+        auto& pk = m.key();
+
+        dht::partition_range_vector partition_ranges{dht::partition_range(dht::global_partitioner().decorate_key(*_schema, pk))};
```
haaawk: how about using m.decorated_key() here instead of decorating the key again?

cdc/cdc.cc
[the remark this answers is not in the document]
elcallio: This is a mutation being written. The tombstone should not be expired. Even if it is, the operation is a delete (that then does not really do so) . In any case, we don't need to pre-image. 
And a mutation being written as a result of any query should not contain both PK delete/tombstone and data should it?

cdc/cdc.cc
```diff
@@ -323,6 +329,75 @@ class transformer final {
...
+        auto* cids = &static_columns;
+        auto kind = column_kind::static_column;
+
+        for (const row& r : { std::ref(p.static_row()), std::ref(p.clustered_rows().begin()->row().cells()) }) {
+            r.for_each_cell([&](column_id id, const atomic_cell_or_collection&) {
+                auto& cdef =_schema->column_at(kind, id);
+                cids->emplace_back(id);
+                columns.emplace_back(&cdef);
+            });
+            cids = &regular_columns;
```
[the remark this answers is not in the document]
elcallio: Duplicated code is the devil. Updating a var in a loop is not strange. Would be nicer to make the secondary var a part of the loop data, but since it is just two, this is simpler.

cdc/cdc.cc
elcallio: We probably could. And this is something that the future will tell. But not something to prematurely solve. 
Most of the time, mutations will be order of 1 in length.

cdc/cdc.cc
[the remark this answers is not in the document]
haaawk: or just use transform with back inserter.

tests/cql_assertions.cc
elcallio: For changing the signature of a test method (verified to compile)? Sure. @dekimir, please review.

cdc/cdc.hh
elcallio: What do you mean "by name"? And what do you mean "inserts another code inbetween"?

cdc/cdc.hh
elcallio: Of course not. Rebasing force-pushed stuff is a nightmare. I'll re-check.

cdc/cdc.cc
elcallio: Indeed.

cdc/cdc.cc
```diff
@@ -299,11 +307,13 @@ class transformer final {
                 {
                     auto log_ck = set_pk_columns(m.key(), batch_no, res);
                     set_bound(log_ck, rt.start);
+                    set_operation(log_ck, operation::range_delete_start, res);
```
haaawk: Should we make it range_delete_start_inclusive/exclusive to cover different bound types?

cdc/cdc.cc
```diff
@@ -299,11 +307,13 @@ class transformer final {
...
                     ++batch_no;
                 }
                 {
                     auto log_ck = set_pk_columns(m.key(), batch_no, res);
                     set_bound(log_ck, rt.end);
+                    set_operation(log_ck, operation::range_delete_end, res);
```
haaawk: same question here?

cdc/cdc.cc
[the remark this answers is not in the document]
nyh: You're right that "describe schema" can tell a cqlsh user which type "operation" is, but I suspect that actual application code will not call "describe schema" but rather assume it has a specific type. An important part of the design document should have been (@haaawk do we have this?) the user's API to use this feature, and I think the type of these fields are part of the API, and we won't want to ever change it.
Anyway, I'm not **strongly** against this, just mildly, so if you prefer it this way, then so be it.

cdc/cdc.hh
haaawk: I mean something like this "This class is restricted the same way as enum class operation".
Inserting another code inbetween:
enum class operation {
// important comment
...
}

class foo {
// not important comment
...
}

enm class column_op {
// same as above
}

cdc/cdc.cc
elcallio: Maybe. But let's leave this as a follow-up, no?

haaawk: Sure. Please at least add FIXME so that we don't forget.

cdc/cdc.cc
```diff
@@ -334,6 +334,32 @@ class transformer final {
                     ++pos;
                 }
 
+                std::vector<bytes_opt> values(3);
+
+                r.row().cells().for_each_cell([&](column_id id, const atomic_cell_or_collection& cell) {
+                    auto& cdef = _schema->column_at(column_kind::regular_column, id);
+                    auto* dst = _log_schema->get_column_definition(to_bytes("_" + cdef.name()));
+                    // todo: collections.
```
haaawk: Please print a warning for collections.

cdc/cdc.cc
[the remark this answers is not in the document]
elcallio: Yes. But the docs will (or should) be generated by schema describe... > schema.txt; insert content of schema.txt into docs. 
Point being that, no, op type should not change (unless needed to), and as such is should be defined at one point, and that is the enum type. underlying_type_t<> is not something that changes willy-nilly, it is restricted by enum <name> : <type>. Again, I _really_ dislike spreading out loosly coupled definitions, so when I can make stuff dependent on a single definition, I do.

cdc/cdc.cc
```diff
@@ -321,16 +321,45 @@ class transformer final {
...
             int batch_no = 0;
             for (const rows_entry& r : p.clustered_rows()) {
-                auto log_ck = set_pk_columns(m.key(), batch_no, res);
                 auto ck_value = r.key().explode(*_schema);
+
+                std::optional<clustering_key> pikey;
+                const cql3::untyped_result_set_row * pirow = nullptr;
+
+                if (rs) {
+                    for (auto& utr : *rs) {
```
haaawk: I'm very sad about the quadratic complexity of this code. Would it be possible to depend on the fact that both p.clustered_rows() and rs are sorted and make it linear?

haaawk: Not for this PR though. I will do it as a follow up.

cdc/cdc.cc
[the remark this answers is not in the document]
nyh: @haaawk @elcallio  I imagine more general mutations, coming from e.g. repair. But now that I think of it, I don't think you handle repair at all, right? It's just for new modifications as they happen. So ok.

cdc/cdc.cc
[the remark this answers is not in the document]
nyh: @elcallio the funny thing that the "unduplicated" code is actually longer than the duplicated code, and harder to understand (it took me a minute to figure out why you're overwriting these variables at the end of each iteration - and that it's only meant to happen after the first iteration). Its only merit is that it doesn't have the same line of code twice. Well, sort of (you do have cids= assigned in two places). I guess it's a matter of taste.

cdc/cdc.cc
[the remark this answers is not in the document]
haaawk: Right. Only mutations from user actions.

cdc/cdc.cc
```diff
@@ -357,13 +386,23 @@ class transformer final {
 
                         values[0] = data_type_for<column_op_native_type>()->decompose(data_value(static_cast<column_op_native_type>(op)));
                         res.set_cell(log_ck, *dst, atomic_cell::make_live(*dst->type, _time.timestamp(), tuple_type_impl::build_value(values)));
+
+                        if (pirow && pirow->has(cdef.name_as_text())) {
+                            values[0] = data_type_for<column_op_native_type>()->decompose(data_value(static_cast<column_op_native_type>(column_op::set)));
```
haaawk: Please add FIXME so that we remember to add support for column tombstones here as well.

tests/cql_assertions.cc
dekimir: LGTM (nit: `seastar::` is superfluous here)

cdc/cdc.cc
```diff
@@ -357,13 +386,23 @@ class transformer final {
...
+                            values[1] = pirow->get_blob(cdef.name_as_text());
+                            values[2] = std::nullopt;
```
haaawk: And a FIXME for TTL.

cdc/cdc.cc
elcallio: Uhm, if there is a tombstone here, we will not get a value back, simple as that.

cdc/cdc.cc
elcallio: Again, we don't get a ttl back. This is a snapshot of how this looks right now, by value. We have data bytes, that is is. I think that is fine.

cdc/cdc.cc
haaawk: Right. Sorry I got confused.

nyh:
I wanted to commit this, but it doesn't compile:
```
istcc[6539] ERROR: compile cdc/cdc.cc on compile1/8,lzo failed
cdc/cdc.cc: In member function ‘mutation cdc::transformer::transform(const mutation&, const cql3::untyped_result_set*) const’:
cdc/cdc.cc:414:73: error: no match for call to ‘(cdc::transformer::transform(const mutation&, const cql3::untyped_result_set*) const::<lambda(const row&, column_kind)>) (const lazy_row&, column_kind)’
cdc/cdc.cc:375:38: note: candidate: ‘cdc::transformer::transform(const mutation&, const cql3::untyped_result_set*) const::<lambda(const row&, column_kind)>’
cdc/cdc.cc:375:38: note:   no known conversion for argument 1 from ‘const lazy_row’ to ‘const row&’
cdc/cdc.cc: In member function ‘seastar::future<seastar::lw_shared_ptr<cql3::untyped_result_set> > cdc::transformer::pre_image_select(service::storage_proxy&, service::client_state&, db::consistency_level, const mutation&)’:
cdc/cdc.cc:465:159: error: unable to deduce ‘std::initializer_list<auto>&&’ from ‘{std::tie(_Elements& ...) [with _Elements = {const lazy_row, utils::small_vector<unsigned int, 8>, column_kind}](static_columns, sk), std::tie(_Elements& ...) [with _Elements = {const row, utils::small_vector<unsigned int, 8>, column_kind}](regular_columns, rk)}’
cdc/cdc.cc:465:159: note:   deduced conflicting types for parameter ‘auto’ (‘std::tuple<const lazy_row&, utils::small_vector<unsigned int, 8>&, column_kind&>’ and ‘std::tuple<const row&, utils::small_vector<unsigned int, 8>&, column_kind&>’)
```
Am I doing anything wrong? Is it possible you forgot to include something in your branch?

haaawk:
> I wanted to commit this, but it doesn't compile:
> 
> ```
> istcc[6539] ERROR: compile cdc/cdc.cc on compile1/8,lzo failed
> cdc/cdc.cc: In member function ‘mutation cdc::transformer::transform(const mutation&, const cql3::untyped_result_set*) const’:
> cdc/cdc.cc:414:73: error: no match for call to ‘(cdc::transformer::transform(const mutation&, const cql3::untyped_result_set*) const::<lambda(const row&, column_kind)>) (const lazy_row&, column_kind)’
> cdc/cdc.cc:375:38: note: candidate: ‘cdc::transformer::transform(const mutation&, const cql3::untyped_result_set*) const::<lambda(const row&, column_kind)>’
> cdc/cdc.cc:375:38: note:   no known conversion for argument 1 from ‘const lazy_row’ to ‘const row&’
> cdc/cdc.cc: In member function ‘seastar::future<seastar::lw_shared_ptr<cql3::untyped_result_set> > cdc::transformer::pre_image_select(service::storage_proxy&, service::client_state&, db::consistency_level, const mutation&)’:
> cdc/cdc.cc:465:159: error: unable to deduce ‘std::initializer_list<auto>&&’ from ‘{std::tie(_Elements& ...) [with _Elements = {const lazy_row, utils::small_vector<unsigned int, 8>, column_kind}](static_columns, sk), std::tie(_Elements& ...) [with _Elements = {const row, utils::small_vector<unsigned int, 8>, column_kind}](regular_columns, rk)}’
> cdc/cdc.cc:465:159: note:   deduced conflicting types for parameter ‘auto’ (‘std::tuple<const lazy_row&, utils::small_vector<unsigned int, 8>&, column_kind&>’ and ‘std::tuple<const row&, utils::small_vector<unsigned int, 8>&, column_kind&>’)
> ```
> 
> Am I doing anything wrong? Is it possible you forgot to include something in your branch?

I think it's an effect of Avi's "mutation_partition: make static_row optional to reduce memory footprint".
Since Calle is not working today I will fix it and upload the new version of this PR.

haaawk:
@nyh I rebased this PR on top of current master, fixed the compilation error and removed some trailing whitespaces. You can find a rebased version here -> https://github.com/scylladb/scylla/pull/5217
It should be ready for merge :)

nyh:
Thanks. Committed @haaawk 's newer rebased version.

PR #5198, opened 2019-10-21
combined reader gallop mode

### LINKED ISSUE #3593: combined_reader gallop mode
Inspired by [Timsort](https://en.wikipedia.org/wiki/Timsort).

If a particular input reader contributes a stream of fragments and keeps winning over other readers, we can enter a gallop mode where we assume it will keep winning, and instead of entering the next fragment from that reader into the heap, we can compare it against the min element of the heap and move it to the output directly, avoiding the O(log N) compares needed to heapify. If it fails the comparison, exit gallop mode and push that fragment to the heap.

This is likely to occur with time series workloads, where sstables will have disjoint runs of clustering keys. It's also likely to happen with LCS partition keys, because the lower level sstables have fewer keys per token than higher level sstables, but that's unlikely to be much help because we're only merging two sstables.

piodul:
In case when a single reader contributes a stream of fragments and keeps winning over other readers, mutation_reader_merger will enter gallop mode, in which it is assumed that the reader will keep winning over other readers. Currently, a reader needs to contribute 3 fragments to enter that mode.

In gallop mode, fragments returned by the galloping reader will be compared with the best fragment from _reader_heap or _fragment_heap. If it wins, the fragment is directly returned. Otherwise, gallop mode ends and merging performed as in general case, which involves heap operations.

A microbenchmark was added in order to test performance of the galloping reader optimization. A combining reader that merges results from four other readers is created. Each sub-reader provides a range of 16 clustering rows that is disjoint from others. All sub-readers return rows from the same partition. An improvement can be observed after introducing the galloping reader optimization.

As for other benchmarks from the "combined" group, results are pretty close to the old ones. Usually new results are slightly slower, but it depends on the run - in the comparision below, most results for the optimized version are actually better. Most of the time, the difference is not greater than 6%.
    
Median times from a single run of perf_mutation_readers.combined:
(1s run duration, 5 runs per benchmark, dev mode)

```    
test name                      before    -> after     change
combined.one_row:              103.416ns -> 105.080ns (1.6% worse)
combined.single_active:        101.730us -> 98.141us  (3.5% better)
combined.many_overlapping:     817.900us -> 808.799us (1.1% better)
combined.disjoint_interleaved: 96.008us  -> 92.657us  (3.4% better)
combined.disjoint_ranges:      96.100us  -> 91.972us  (4.3% better)
combined.overlapping_partitions_disjoint_rows:
                               107.784us -> 71.915us  (33.3% better)
```

Tested on AMD Ryzen Threadripper 2950X @ 3.5GHz.

Tests: unit(dev)
Fixes #3593.

### DIFF
--- a/mutation_reader.cc
+++ b/mutation_reader.cc
@@ -166,6 +166,10 @@ public:
     };
 
     using mutation_fragment_batch = boost::iterator_range<merger_vector<mutation_fragment>::iterator>;
+
+    // Determines how many times a fragment should be taken from the same
+    // reader in order to enter gallop mode. Must be greater than one.
+    static constexpr int gallop_mode_entering_threshold = 3;
 private:
     struct reader_heap_compare;
     struct fragment_heap_compare;
@@ -194,12 +198,25 @@ private:
     // end, a call to next_partition() or a call to
     // fast_forward_to(dht::partition_range).
     reader_and_last_fragment_kind _single_reader;
+    // Holds a reference to the reader that previously contributed a fragment.
+    // When hits reader contributes a certain number of fragments in a row,
+    // gallop mode becomes enabled. In this mode, the _galloping_reader
+    // is assumed to be the contributor of the next fragment.
+    reader_and_last_fragment_kind _galloping_reader;
+    // Counts how many times the _galloping_reader contributed a fragment
+    // before entering the gallop mode. It can also be equal to 0, meaning
+    // that the gallop mode was stopped (galloping reader lost to some other reader).
+    int _gallop_mode_hits = 0;
     const schema_ptr _schema;
     streamed_mutation::forwarding _fwd_sm;
     mutation_reader::forwarding _fwd_mr;
 private:
+    void maybe_add_readers_at_partition_boundary();
     void maybe_add_readers(const std::optional<dht::ring_position_view>& pos);
     void add_readers(std::vector<flat_mutation_reader> new_readers);
+    void retire_reader(reader_and_last_fragment_kind rk);
+    bool in_gallop_mode() const;
+    future<stop_iteration> advance_galloping_reader(db::timeout_clock::time_point timeout);
     future<> prepare_next(db::timeout_clock::time_point timeout);
     // Collect all forwardable readers into _next, and remove them from
     // their previous containers (_halted_readers and _fragment_heap).
@@ -303,6 +320,72 @@ struct mutation_reader_merger::fragment_heap_compare {
     }
 };
 
+void mutation_reader_merger::maybe_add_readers_at_partition_boundary() {
+    // We are either crossing partition boundary or ran out of
+    // readers. If there are halted readers then we are just
+    // waiting for a fast-forward so there is nothing to do.
+    if (_fragment_heap.empty() && _halted_readers.empty()) {
+        if (_reader_heap.empty()) {
+            maybe_add_readers(std::nullopt);
+        } else {
+            maybe_add_readers(_reader_heap.front().fragment.as_partition_start().key());
+        }
+    }
+}
+
+void mutation_reader_merger::retire_reader(reader_and_last_fragment_kind rk) {
+    if (_fwd_sm == streamed_mutation::forwarding::yes && rk.last_kind != mutation_fragment::kind::partition_end) {
+        // When in streamed_mutation::forwarding mode we need
+        // to keep track of readers that returned
+        // end-of-stream to know what readers to ff. We can't
+        // just ff all readers as we might drop fragments from
+        // partitions we haven't even read yet.
+        // Readers whoose last emitted fragment was a partition
+        // end are out of data for good for the current range.
+        _halted_readers.push_back(std::move(rk));
+    } else if (_fwd_mr == mutation_reader::forwarding::no) {
+        _to_remove.splice(_to_remove.end(), _all_readers, rk.reader);
+        if (_to_remove.size() >= 4) {
+            _to_remove.clear();
+        }
+    }
+}
+
+bool mutation_reader_merger::in_gallop_mode() const {
+    return _gallop_mode_hits >= gallop_mode_entering_threshold;
+}
+
+future<stop_iteration> mutation_reader_merger::advance_galloping_reader(db::timeout_clock::time_point timeout) {
+    // When _galloping_reader is present, we assume that it will keep producing fragments
+    // that win over other readers. If that assumption turns out to be no longer true,
+    // this function exits gallop mode and we should proceed with standard merging logic.
+    return (*_galloping_reader.reader)(timeout).then([this] (mutation_fragment_opt mfo) {
+        if (mfo) {
+            if (mfo->is_partition_start()) {
+                // Don't assume we will keep winning in the next partition.
+                _reader_heap.emplace_back(_galloping_reader.reader, std::move(*mfo));
+                boost::push_heap(_reader_heap, reader_heap_compare(*_schema));
+                maybe_add_readers_at_partition_boundary();
+            } else {
+                if (_fragment_heap.empty() || position_in_partition::less_compare(*_schema)(mfo->position(), _fragment_heap.front().fragment.position())) {
+                    _current.clear();
+                    _current.push_back(std::move(*mfo));
+                    _galloping_reader.last_kind = _current.back().mutation_fragment_kind();
+                    return stop_iteration::yes;
+                }
+
+                _fragment_heap.emplace_back(_galloping_reader.reader, std::move(*mfo));
+                boost::push_heap(_fragment_heap, fragment_heap_compare(*_schema));
+            }
+        } else {
+            retire_reader(_galloping_reader);
+            maybe_add_readers_at_partition_boundary();
+        }
+        _gallop_mode_hits = 0;
+        return stop_iteration::no;
+    });
+}
+
 future<> mutation_reader_merger::prepare_next(db::timeout_clock::time_point timeout) {
     return parallel_for_each(_next, [this, timeout] (reader_and_last_fragment_kind rk) {
         return (*rk.reader)(timeout).then([this, rk] (mutation_fragment_opt mfo) {
@@ -314,35 +397,13 @@ future<> mutation_reader_merger::prepare_next(db::timeout_clock::time_point time
                     _fragment_heap.emplace_back(rk.reader, std::move(*mfo));
                     boost::range::push_heap(_fragment_heap, fragment_heap_compare(*_schema));
                 }
-            } else if (_fwd_sm == streamed_mutation::forwarding::yes && rk.last_kind != mutation_fragment::kind::partition_end) {
-                // When in streamed_mutation::forwarding mode we need
-                // to keep track of readers that returned
-                // end-of-stream to know what readers to ff. We can't
-                // just ff all readers as we might drop fragments from
-                // partitions we haven't even read yet.
-                // Readers whoose last emitted fragment was a partition
-                // end are out of data for good for the current range.
-                _halted_readers.push_back(rk);
-            } else if (_fwd_mr == mutation_reader::forwarding::no) {
-                _to_remove.splice(_to_remove.end(), _all_readers, rk.reader);
-                if (_to_remove.size() >= 4) {
-                    _to_remove.clear();
-                }
+            } else {
+                retire_reader(rk);
             }
         });
     }).then([this] {
         _next.clear();
-
-        // We are either crossing partition boundary or ran out of
-        // readers. If there are halted readers then we are just
-        // waiting for a fast-forward so there is nothing to do.
-        if (_fragment_heap.empty() && _halted_readers.empty()) {
-            if (_reader_heap.empty()) {
-                maybe_add_readers(std::nullopt);
-            } else {
-                maybe_add_readers(_reader_heap.front().fragment.as_partition_start().key());
-            }
-        }
+        maybe_add_readers_at_partition_boundary();
     });
 }
 
@@ -352,6 +413,10 @@ void mutation_reader_merger::prepare_forwardable_readers() {
     std::move(_halted_readers.begin(), _halted_readers.end(), std::back_inserter(_next));
     if (_single_reader.reader != reader_iterator{}) {
         _next.emplace_back(std::exchange(_single_reader.reader, {}), _single_reader.last_kind);
+    }
+    if (in_gallop_mode()) {
+        _next.emplace_back(_galloping_reader);
+        _gallop_mode_hits = 0;
     }
     for (auto& df : _fragment_heap) {
         _next.emplace_back(df.reader, df.fragment.mutation_fragment_kind());
@@ -392,6 +457,17 @@ future<mutation_reader_merger::mutation_fragment_batch> mutation_reader_merger::
         return make_ready_future<mutation_fragment_batch>(_current);
     }
 
+    if (in_gallop_mode()) {
+        return advance_galloping_reader(timeout).then([this, timeout] (stop_iteration stop) {
+            if (stop == stop_iteration::yes) {
+                return make_ready_future<mutation_fragment_batch>(_current);
+            }
+            // Galloping reader may have lost to some other reader. In that case, we should proceed
+            // with standard merging logic.
+            return (*this)(timeout);
+        });
+    }
+
     if (!_next.empty()) {
         return prepare_next(timeout).then([this, timeout] { return (*this)(timeout); });
     }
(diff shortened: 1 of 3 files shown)


avikivity:
@denesb please review, it's your baby he's mutilating

avikivity:
Please provide measurements in release mode, and normalize to time per fragment.

mutation_reader.cc
```diff
@@ -303,6 +320,72 @@ struct mutation_reader_merger::fragment_heap_compare {
...
+    // this function exits gallop mode and we should proceed with standard merging logic.
+    return (*_galloping_reader.reader)(timeout).then([this] (mutation_fragment_opt mfo) {
+        if (mfo) {
+            if (mfo->is_partition_start()) {
+                // Don't assume we will keep winning in the next partition.
+                _reader_heap.emplace_back(_galloping_reader.reader, std::move(*mfo));
+                boost::push_heap(_reader_heap, reader_heap_compare(*_schema));
+                maybe_add_readers_at_partition_boundary();
+            } else {
+                if (_fragment_heap.empty() || position_in_partition::less_compare(*_schema)(mfo->position(), _fragment_heap.front().fragment.position())) {
```
denesb: It appears to me that this condition is the meat of this optimization, other than that there is not much difference compared to the standard logic.
I think we could merge all three modes (multi-reader, single-reader and gallop) by just specializing `prepare_next()` to single and multiple readers by just looking at how many readers are in `_next`. When there is just one reader we try to see if we can bypass the merging logic altogether by using this very same condition. When there are multiple readers we fall back to the standard logic. We can get away without any separate `_single_reader` and `_gallopping_reader` and friends.

denesb: `prepare_next()` would return `future<needs_merge>`. When `needs_merge` == false, `operator()()` bypasses the merging logic.

avikivity:
@piodul does this work for different rows in the same partition, different partitions, or both?

piodul:
@avikivity Currently, this optimization works only for different rows in the same partition. Galloping is turned off when next partition starts: https://github.com/scylladb/scylla/pull/5198/files#diff-2a2d8752bb687a0e338dfe715fcc3bd3R365

mutation_reader.cc
piodul: Reading the old discussion, I can see that Paweł kept `_single_reader` logic outside the `prepare_next` function in order to skip wrapping in a future in a general case (e.g. when single reader's internal buffer still has some fragments). I will try merging `advance_galloping_reader` with `prepare_next` and will see how the performance changes.

mutation_reader.cc
```diff
@@ -194,12 +198,25 @@ class mutation_reader_merger {
     // end, a call to next_partition() or a call to
     // fast_forward_to(dht::partition_range).
     reader_and_last_fragment_kind _single_reader;
+    // Holds a reference to the reader that previously contributed a fragment.
+    // When hits reader contributes a certain number of fragments in a row,
+    // gallop mode becomes enabled. In this mode, the _galloping_reader
+    // is assumed to be the contributor of the next fragment.
+    reader_and_last_fragment_kind _galloping_reader;
```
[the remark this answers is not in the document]
piodul: I found the old discussion ([here](https://groups.google.com/forum/#!searchin/scylladb-dev/%22combined_reader$3A$20optimise$20for%22|sort:date/scylladb-dev/YtDC7dm1l0c/osVDTeevAQAJ)). The argument against removing the `_single_reader` variable is that re-using `_next` makes it less clear that the single reader mode is being turned on.

I think I agree with Paweł on that. Furthermore, we still need an additional field for tracking which reader previously contributed a fragment in order to know when to increase/clear `_gallop_mode_hits`.

mutation_reader.cc
denesb: I experimented once optimizing for the case where the readers' buffers are not empty, avoding futures. I saw no improvements whatsoever in the our macrobenchmarks.

mutation_reader.cc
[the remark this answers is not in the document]
denesb: If we do what we suggested, we don't need `_gallop_mode_hits`. Gallop mode is implicit and is triggered any time a single reader is in `_next`.

mutation_reader.cc
denesb: (which doesn't mean you shouldn't check to see if the performance changes)

mutation_reader.cc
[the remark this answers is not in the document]
piodul: I'm not sure if I understand correctly. This optimization is supposed to kick in when one reader contributes 3 fragments consecutively, so we need to count it somehow.

`_next` holds all readers that contributed a fragment in previous `operator()()` call. `_next.size()` would be 1 also in the case when two readers contributing alternating fragments are merged - in that case galloping would be detrimental to performance, because we do an extra key comparison before falling back to heap operations.

I think it is possible to remove `_galloping_reader` and use the previous value from `_next` to compare if currently contributing reader is the same as before. We would have to pass this information to `operator()()` or structure the function differently.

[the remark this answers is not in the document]
denesb: Ok, so the reason we want to trigger this optimization after a certain threshold is to avoid the additional key comparison. I think we can do this with my proposed solution as well, albeit with some changes.

Before I go into that I'd like to lay out my reasons why I'd like to do away with the separate states (single reader, multi reader and gallop). Yes, adding another member is not that terrible (like Pawel put it), yet this creates another separate logical stream within the reader. We already have two and are about to add a third one, each with its own way of fetching fragments from the reader, then propagating them downstream. It's almost like having 3 readers in one, with a bunch of duplicate code. This makes one of our most complex reader even more complex. So I think we should also weight code complexity, not just runtime performance.

I think we can make this even simpler then I originally proposed: modify `prepare_next()` to not push all new fragments into the `_fragment_heap`, instead these new fragments should go into another (unordered) container, `_new_fragments`. Then, in `operator()()` we can see how many new fragments were produced and if there is just one, make a "bet" whether we'll need to merge or not. This can be based on a history of recent such "bets". This history is `_galloping_reader_hits` in your proposal.
We start out by assuming we need to merge. We push `_new_fragments.front()` to `_fragment_heap`. If after making the heap it is still the last element (it is the smallest in the heap) we bump the counter as we initiated a merge when we didn't need one. After `N` (3) such events we flip the assumption and enter gallop mode, assuming we won't need a merge and check `_new_fragments.front()` against `_fragment_heap` before pushing it.

mutation_reader.cc
denesb: > Reading the old discussion, I can see that Paweł kept `_single_reader` logic outside the `prepare_next` function in order to skip wrapping in a future in a general case (e.g. when single reader's internal buffer still has some fragments).

Wrapping something in a future is of course more expensive then not having to do it at all, but that is not the case here. We have to handle the case where the buffer is empty anyway, so the compiler has to generate the future handling code anyway, we are not saving anything. Furthermore calling `then()` on a ready future is actually pretty cheap, the continuation lambda is called right away. What is expensive is a future that defers as that causes task objects to be allocated for each continuation. So, as long as we avoid unnecessary defers we should be fine performance wise.

mutation_reader.cc
```diff
@@ -314,35 +397,13 @@ future<> mutation_reader_merger::prepare_next(db::timeout_clock::time_point time
...
-                // just ff all readers as we might drop fragments from
-                // partitions we haven't even read yet.
-                // Readers whoose last emitted fragment was a partition
-                // end are out of data for good for the current range.
-                _halted_readers.push_back(rk);
-            } else if (_fwd_mr == mutation_reader::forwarding::no) {
-                _to_remove.splice(_to_remove.end(), _all_readers, rk.reader);
-                if (_to_remove.size() >= 4) {
-                    _to_remove.clear();
-                }
```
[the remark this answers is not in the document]
piodul: Done, moved into separate commit.

piodul:
@denesb

I tried reorganizing the logic in the way you described (removing single and galloping readers, passing through `_new_fragments`), but observed a performance regression on the perf_mutation_readers microbenchmark.

Measurements from the first implementation of this change (these were run in release mode, on a slightly faster machine):

```
test                                      iterations      median         mad         min         max
combined.one_row                             2510366    51.178ns     0.179ns    50.985ns    51.607ns
combined.single_active                         10745    57.073us    77.093ns    56.830us    57.399us
combined.many_overlapping                       1601   477.804us     1.320us   476.484us   482.929us
combined.disjoint_interleaved                  10408    55.716us   230.158ns    55.284us    55.950us
combined.disjoint_ranges                       10352    54.414us    34.435ns    54.380us    54.739us
combined.overlapping_partitions_disjoint_rows       15147    38.201us    33.760ns    38.137us    38.544us
```

After changing the logic [link](https://github.com/piodul/scylla/compare/765dc86de43730b70cca993b2e8dcf78be424e28..9d723ebc77570ef475cbcfe2551217bf746819ea):

```
test                                      iterations      median         mad         min         max
combined.one_row                             2307771    56.470ns     0.126ns    56.134ns    56.596ns
combined.single_active                          8549    71.943us   582.406ns    70.296us    75.426us
combined.many_overlapping                       1591   476.175us   266.926ns   475.885us   485.330us
combined.disjoint_interleaved                   8662    74.187us   140.655ns    74.046us    74.480us
combined.disjoint_ranges                        8711    70.226us   214.404ns    70.012us    71.927us
combined.overlapping_partitions_disjoint_rows       12018    56.130us   382.621ns    55.734us    56.913us
```

[Reintroducing `_single_reader` special handling](https://github.com/piodul/scylla/commit/aa07f39bf7bddd1c8cd3d875b79af740b4eea590) helped bring back old performance for the half of the tests, but the new benchmark that tests the `galloping reader` optimization target is still the same:

```
test                                      iterations      median         mad         min         max
combined.one_row                             2435163    51.774ns     0.396ns    50.907ns    52.169ns
combined.single_active                         10734    56.813us     8.027ns    56.789us    56.910us
combined.many_overlapping                       1599   475.385us     1.840us   472.539us   479.347us
combined.disjoint_interleaved                  10250    55.051us   205.976ns    54.712us    57.079us
combined.disjoint_ranges                        9693    57.982us   345.795ns    57.221us    63.638us
combined.overlapping_partitions_disjoint_rows       11485    58.646us   263.370ns    58.367us    72.713us
```

I agree that clarity is important, but the goal of this issue was to optimize a special case. I'm not sure if introducing performance regression would count as a valid solution...

I have reorganized the code a little differently while keeping the performance (only `many_overlapping` benchmark time increased from 475us to 500us). It reduced code duplication, but ultimately I'm not sure it is now more clear or not. This version is available here: [link](https://github.com/piodul/scylla/commit/1f47d4f231c4e892c0e2f045109225b187ffe970). Which version would you prefer me to keep?

By the way, could you clarify what you meant by macrobenchmarks? I'm not sure what you meant, so I sticked only to running microbenchmarks.

piodul:
Increased the `overlapping_partitions_disjoint_rows` to have 32 clustering rows per partition, in order to make the total amount of clustering rows used in the test similar to other ones (all benchmarks apart from `one_row` and `many_overlapping` have 512 clustering rows).

@denesb
As expected, changing compile-time parameter to dynamic one did not change the performance of the solution.

I tried running benchmarks from `perf_fast_forward`, but observed no difference in performance.

I also tried to see if a difference can be observed in real queries. I prepared a table with only one partition and 4 sstables. Each sstable had a range of 10*1024 rows, and each range was disjoint from other sstables' ranges. I did not observe any noticeable improvement when performing a `select * from test_table` query.

@avikivity
Benchmarks redone in release mode. New results are included in the message for commit with benchmark, and in the PR description.

As for preserving the gallop mode across partitions, should we implement it in this version, or in a follow-up?

avikivity:
@piodul your select benchmark was perhaps dominated by paging latency. You need to saturate the server to overserve the throughput improvement. Also, did you prevent compaction from merging your tables?

The gain is about 90ns/row. We've seen nodes doing 200k rows/sec/core, which is about 5000 ns/row. So the improvement should be around 2%. But I expect the benefit to come for compactions when merging sstables that have disjoint data (which can be quite common for time series workloads as the data is ordered by timestamp).

piodul:
@avikivity The table used TWCS strategy, I created each sstables in different time window, so they would not be compacted. The sstable files did not change before, during or after the test.

avikivity:
I see. They may have been cached after the first read.

Maybe you can eliminate paging latency by using aggregation: select count(*) from tab where pk = ?.

mutation_reader.cc
```diff
@@ -336,30 +358,59 @@ void mutation_reader_merger::retire_reader(reader_and_last_fragment_kind rk) {
...
+future<mutation_reader_merger::needs_merge> mutation_reader_merger::advance_galloping_reader(db::timeout_clock::time_point timeout) {
+    return prepare_one(timeout, _galloping_reader, handle_galloping::yes).then([this] (needs_merge needs_merge) {
+        maybe_add_readers_at_partition_boundary();
+        return needs_merge;
+    });
+}
+
+future<> mutation_reader_merger::prepare_next(db::timeout_clock::time_point timeout) {
+    return parallel_for_each(_next, [this, timeout] (reader_and_last_fragment_kind rk) {
+        return prepare_one(timeout, rk, handle_galloping::no).discard_result();
```
denesb: `discard_result()` attaches a continuation. Better just adjust the signature of the lambda below and ignore the parameter.

mutation_reader.cc
```diff
@@ -336,30 +358,59 @@ void mutation_reader_merger::retire_reader(reader_and_last_fragment_kind rk) {
...
-                }
-            } else if (_fwd_sm == streamed_mutation::forwarding::yes && rk.last_kind != mutation_fragment::kind::partition_end) {
-                // When in streamed_mutation::forwarding mode we need
-                // to keep track of readers that returned
-                // end-of-stream to know what readers to ff. We can't
-                // just ff all readers as we might drop fragments from
-                // partitions we haven't even read yet.
-                // Readers whoose last emitted fragment was a partition
-                // end are out of data for good for the current range.
-                _halted_readers.push_back(rk);
```
denesb: This code movement should also have been done in a separate commit. It is really hard to see what has changed in this code snippet and what haven't.

mutation_reader.cc
```diff
@@ -303,6 +305,37 @@ struct mutation_reader_merger::fragment_heap_compare {
...
+    if (_fragment_heap.empty() && _halted_readers.empty()) {
+        if (_reader_heap.empty()) {
+            maybe_add_readers(std::nullopt);
+        } else {
+            maybe_add_readers(_reader_heap.front().fragment.as_partition_start().key());
+        }
+    }
+}
+
+void mutation_reader_merger::retire_reader(reader_and_last_fragment_kind rk) {
```
denesb: This extraction doesn't seem necessary in the new version, but if you think it makes the code easier to read we can still do it, but in that case it should be in a commit saying it is extracted for readability.

mutation_reader.cc
```diff
@@ -336,30 +358,59 @@ void mutation_reader_merger::retire_reader(reader_and_last_fragment_kind rk) {
...
-                // partitions we haven't even read yet.
-                // Readers whoose last emitted fragment was a partition
-                // end are out of data for good for the current range.
-                _halted_readers.push_back(rk);
+bool mutation_reader_merger::in_gallop_mode() const {
+    return _gallop_mode_hits >= gallop_mode_entering_threshold;
+}
+
+future<mutation_reader_merger::needs_merge> mutation_reader_merger::prepare_one(db::timeout_clock::time_point timeout,
+        reader_and_last_fragment_kind rk, handle_galloping handle_galloping) {
```
denesb: reader_galloping might be a better name.

mutation_reader.cc
piodul: `parallel_for_each` [accepts only futures which return nothing](https://github.com/scylladb/seastar/blob/master/include/seastar/core/future-util.hh#L194), so we are forced to call `.discard_result()` at the end.

mutation_reader.cc
piodul: Done, changed

mutation_reader.cc
piodul: OK, I inlined `retire_reader` back. I left `maybe_add_readers_at_partition_boundary` intact, and included extraction of `prepare_one` in the first commit.

mutation_reader.cc
piodul: My bad - this branch was accidentally re-introduced during rebase. It belongs to `retire_reader` (which was re-inlined).

mutation_reader.cc
denesb: Right, missed that.

piodul:
@avikivity I re-did the query benchmark with following changes:
- disabled cache by `--enable-cache=no`
- used a larger table (16 sstables, 100*1024 clustering rows each, everything is still in the same partition)
- ran query `select (*) from test_table where pk = 0`

I actually managed to observe an improvement:

```
test           median     min        max        mad
before change  2.171087s  2.166947s  2.222554s  0.002911s
after change   1.870574s  1.817358s  1.917264s  0.007278s
```

median difference: 0.300513s (13.84%)

PR #5203, opened 2019-10-22
cdc: Return db_context::builder by reference

haaawk:
from its with_* functions.


nyh:
Thanks. Committed.

PR #5204, opened 2019-10-22
misc_services: Don't gossip hit rates when they are not used

### LINKED ISSUE #5183: Don't gossip cache hit rates when they are not used

haaawk:
When cache_hit_rate_read_balancing is set to false,
heat-weighted load balancing is turned off and we
gossip cache hit rates for no good reason.
The size of gossiped data is linear on number of tables
in the database so it's good to avoid such traffic
when it is not used.

Fixes #5183

Signed-off-by: Piotr Jastrzebski piotr@scylladb.com

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -1027,12 +1027,20 @@ int main(int ac, char** av) {
                 lb->stop_broadcasting().get();
             });
             supervisor::notify("starting cf cache hit rate calculator");
-            cf_cache_hitrate_calculator.start(std::ref(db), std::ref(cf_cache_hitrate_calculator)).get();
-            auto stop_cache_hitrate_calculator = defer_with_log_on_error([&cf_cache_hitrate_calculator] {
+
+            if (db.local().get_config().cache_hit_rate_read_balancing()) {
+                cf_cache_hitrate_calculator.start(std::ref(db), std::ref(cf_cache_hitrate_calculator)).get();
+            }
+            auto stop_cache_hitrate_calculator = defer_with_log_on_error([&cf_cache_hitrate_calculator, &db] {
+                if (!db.local().get_config().cache_hit_rate_read_balancing()) {
+                    return;
+                }
                 startlog.info("stopping cf cache hit rate calculator");
-                return cf_cache_hitrate_calculator.stop().get();
+                cf_cache_hitrate_calculator.stop().get();
             });
-            cf_cache_hitrate_calculator.local().run_on(engine().cpu_id());
+            if (db.local().get_config().cache_hit_rate_read_balancing()) {
+                cf_cache_hitrate_calculator.local().run_on(engine().cpu_id());
+            }
 
             supervisor::notify("starting view update backlog broker");
             static sharded<service::view_update_backlog_broker> view_backlog_broker;
--- a/database.hh
+++ b/database.hh
@@ -876,7 +876,7 @@ public:
     }
 
     void set_hit_rate(gms::inet_address addr, cache_temperature rate);
-    cache_hit_rate get_hit_rate(gms::inet_address addr);
+    std::optional<cache_hit_rate> get_hit_rate(gms::inet_address addr);
     void drop_hit_rate(gms::inet_address addr);
 
     future<> run_with_compaction_disabled(std::function<future<> ()> func);
--- a/db/consistency_level.cc
+++ b/db/consistency_level.cc
@@ -211,10 +211,12 @@ filter_for_query(consistency_level cl,
     const auto remaining_bf = bf - selected_endpoints.size();
 
     if (cf) {
-        auto get_hit_rate = [cf] (gms::inet_address ep) -> float {
+        bool has_missing_hit_rates = false;
+        auto get_hit_rate = [cf, &has_missing_hit_rates] (gms::inet_address ep) -> float {
             constexpr float max_hit_rate = 0.999;
             auto ht = cf->get_hit_rate(ep);
             if (!ht) {
+                has_missing_hit_rates = true;
                 return 0.0;
             }
             if (float(ht->rate) < 0) {
@@ -241,7 +243,7 @@ filter_for_query(consistency_level cl,
             return std::make_pair(ep, ht);
         }));
 
-        if (!old_node && ht_max - ht_min > 0.01) { // if there is old node or hit rates are close skip calculations
+        if (!has_missing_hit_rates && !old_node && ht_max - ht_min > 0.01) { // if there is old node or hit rates are close skip calculations
             // local node is always first if present (see storage_proxy::get_live_sorted_endpoints)
             unsigned local_idx = epi[0].first == utils::fb_utilities::get_broadcast_address() ? 0 : epi.size() + 1;
             live_endpoints = miss_equalizing_combination(epi, local_idx, remaining_bf, bool(extra));
--- a/db/consistency_level.cc
+++ b/db/consistency_level.cc
@@ -214,15 +214,18 @@ filter_for_query(consistency_level cl,
         auto get_hit_rate = [cf] (gms::inet_address ep) -> float {
             constexpr float max_hit_rate = 0.999;
             auto ht = cf->get_hit_rate(ep);
-            if (float(ht.rate) < 0) {
-                return float(ht.rate);
-            } else if (lowres_clock::now() - ht.last_updated > std::chrono::milliseconds(1000)) {
+            if (!ht) {
+                return 0.0;
+            }
+            if (float(ht->rate) < 0) {
+                return float(ht->rate);
+            } else if (lowres_clock::now() - ht->last_updated > std::chrono::milliseconds(1000)) {
                 // if a cache entry is not updates for a while try to send traffic there
                 // to get more up to date data, mark it updated to not send to much traffic there
-                cf->set_hit_rate(ep, ht.rate);
+                cf->set_hit_rate(ep, ht->rate);
                 return max_hit_rate;
             } else {
-                return std::min(float(ht.rate), max_hit_rate); // calculation below cannot work with hit rate 1
+                return std::min(float(ht->rate), max_hit_rate); // calculation below cannot work with hit rate 1
             }
         };
 
--- a/service/misc_services.cc
+++ b/service/misc_services.cc
@@ -40,6 +40,7 @@
 
 #include "load_broadcaster.hh"
 #include "cache_hitrate_calculator.hh"
+#include "db/config.hh"
 #include "db/system_keyspace.hh"
 #include "gms/application_state.hh"
 #include "service/storage_service.hh"
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -3058,7 +3058,8 @@ storage_proxy::query_partition_key_range_concurrent(storage_proxy::clock_type::t
                     struct {
                         column_family* cf = nullptr;
                         float operator()(const gms::inet_address& ep) const {
-                            return float(cf->get_hit_rate(ep).rate);
+                            auto ht = cf->get_hit_rate(ep);
+                            return ht ? float(ht->rate) : 0.0;
                         }
                     } ep_to_hr{pcf};
                     return *boost::range::min_element(range | boost::adaptors::transformed(ep_to_hr));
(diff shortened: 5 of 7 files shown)


haaawk:
I've tested it the same way as before with 3 nodes cluster and a rolling disable and then enable of HWLB. Nothing bad happened.

main.cc
```diff
@@ -1027,12 +1027,18 @@ int main(int ac, char** av) {
                 lb->stop_broadcasting().get();
             });
             supervisor::notify("starting cf cache hit rate calculator");
-            cf_cache_hitrate_calculator.start(std::ref(db), std::ref(cf_cache_hitrate_calculator)).get();
-            auto stop_cache_hitrate_calculator = defer_with_log_on_error([&cf_cache_hitrate_calculator] {
+
+            auto stop_cache_hitrate_calculator = defer_with_log_on_error([&cf_cache_hitrate_calculator, &db] {
```
tgrabiec: This should be constructed after `cf_cache_hitrate_calculator.start()`, becasue if start() fails, we must not attempt to stop.

haaawk: Ok. Sending fix now.

haaawk:
@tgrabiec I moved the cleaner after the .start() is called.

gleb-cloudius:
I am missing where do we check that entire cluster has the feature disabled?

tgrabiec:
@gleb-cloudius It's disabled on per-request basis, if any of the replicas has the feature disabled. Changed in "filter_for_query: Don't use heat-weighted load balancing when".

gleb-cloudius:
Ah, we skipping the calculation if one of the nodes returned DISABLE, right. Not sure this is enough though. With a cluster of size 4 if one node has it disable but three others are not some requests will be balanced and some will not which may result in disproportional load on some nodes.

tgrabiec:
@gleb-cloudius So are you suggesting to disable the load balancing if any of the known and alive nodes have it disabled?

gleb-cloudius:
On Wed, Oct 23, 2019 at 10:12:07AM -0700, Tomasz Grabiec wrote:
> @gleb-cloudius So are you suggesting to disable the load balancing if any of the known and alive nodes have it disabled?
> 
Or rather if all of them have it disabled.

--
			Gleb.

tgrabiec:
> Or rather if all of them have it disabled.

But then in the mixed cluster the nodes which have HWLB disabled will appear as having 0 hitrate, and that will also cause imbalance of load.

gleb-cloudius:
On Wed, Oct 23, 2019 at 11:31:14AM -0700, Tomasz Grabiec wrote:
> > Or rather if all of them have it disabled.
> 
> But then in the mixed cluster the nodes which have HWLB disabled will appear as having 0 hitrate, and that will also cause imbalance of load.
> 
> 
Why? Without this patch here a node always calculates hitrate, the flag
only controls if the info is actually used or not.

--
			Gleb.

tgrabiec:
śr., 23 paź 2019, 21:00 użytkownik Gleb Natapov <notifications@github.com>
napisał:

> On Wed, Oct 23, 2019 at 11:31:14AM -0700, Tomasz Grabiec wrote:
> > > Or rather if all of them have it disabled.
> >
> > But then in the mixed cluster the nodes which have HWLB disabled will
> appear as having 0 hitrate, and that will also cause imbalance of load.
> >
> >
> Why? Without this patch here a node always calculates hitrate, the flag
> only controls if the info is actually used or not.
>

The flag also disables gossiping of CACHE_HITRATES, that's one of the
reasons we want it. So coordinators will see DISABLED. How do we treat that
when calculating hitrates? The old behavior was to interpret missing
CACHE_HITRATES as 0% hitrate.

>

gleb-cloudius:
On Thu, Oct 24, 2019 at 10:10:09AM -0700, Tomasz Grabiec wrote:
> śr., 23 paź 2019, 21:00 użytkownik Gleb Natapov <notifications@github.com>
> napisał:
> 
> > On Wed, Oct 23, 2019 at 11:31:14AM -0700, Tomasz Grabiec wrote:
> > > > Or rather if all of them have it disabled.
> > >
> > > But then in the mixed cluster the nodes which have HWLB disabled will
> > appear as having 0 hitrate, and that will also cause imbalance of load.
> > >
> > >
> > Why? Without this patch here a node always calculates hitrate, the flag
> > only controls if the info is actually used or not.
> >
> 
> The flag also disables gossiping of CACHE_HITRATES, that's one of the
> reasons we want it. So coordinators will see DISABLED. How do we treat that
> when calculating hitrates? The old behavior was to interpret missing
> CACHE_HITRATES as 0% hitrate.
> 
What I am saying that what the patch does it incorrect. It should not
stop gossiping unless _all_ nodes have it disabled.

--
			Gleb.

haaawk:
> On Thu, Oct 24, 2019 at 10:10:09AM -0700, Tomasz Grabiec wrote: śr., 23 paź 2019, 21:00 użytkownik Gleb Natapov ***@***.***> napisał: > On Wed, Oct 23, 2019 at 11:31:14AM -0700, Tomasz Grabiec wrote: > > > Or rather if all of them have it disabled. > > > > But then in the mixed cluster the nodes which have HWLB disabled will > appear as having 0 hitrate, and that will also cause imbalance of load. > > > > > Why? Without this patch here a node always calculates hitrate, the flag > only controls if the info is actually used or not. > The flag also disables gossiping of CACHE_HITRATES, that's one of the reasons we want it. So coordinators will see DISABLED. How do we treat that when calculating hitrates? The old behavior was to interpret missing CACHE_HITRATES as 0% hitrate.
> What I am saying that what the patch does it incorrect. It should not stop gossiping unless _all_ nodes have it disabled.
> […](#)
> -- Gleb.

How would the nodes know that everyone has HWLB disabled then @gleb-cloudius ? So far we were indicating it by stopping the gossip of hit rates.

tgrabiec:
> What I am saying that what the patch does it incorrect. It should not stop gossiping unless _all_ nodes have it disabled.

Ok.

tgrabiec:
> > On Thu, Oct 24, 2019 at 10:10:09AM -0700, Tomasz Grabiec wrote: śr., 23 paź 2019, 21:00 użytkownik Gleb Natapov _**@**_.***> napisał: > On Wed, Oct 23, 2019 at 11:31:14AM -0700, Tomasz Grabiec wrote: > > > Or rather if all of them have it disabled. > > > > But then in the mixed cluster the nodes which have HWLB disabled will > appear as having 0 hitrate, and that will also cause imbalance of load. > > > > > Why? Without this patch here a node always calculates hitrate, the flag > only controls if the info is actually used or not. > The flag also disables gossiping of CACHE_HITRATES, that's one of the reasons we want it. So coordinators will see DISABLED. How do we treat that when calculating hitrates? The old behavior was to interpret missing CACHE_HITRATES as 0% hitrate.
> > What I am saying that what the patch does it incorrect. It should not stop gossiping unless _all_ nodes have it disabled.
> > […](#)
> > -- Gleb.
> 
> How would the nodes know that everyone has HWLB disabled then @gleb-cloudius ? So far we were indicating it by stopping the gossip of hit rates.

You would need to introduce a new gossip state, HWLB_ENABLED, and check if all live endpoints have it enabled or not when making a decision at the coordinator side or whether to gossip CACHE_HITRATES. Perhaps we can have something like `gms::feature` which listens on gossip changes so that you don't have to actually compute this on each request.

haaawk:
> Ah, we skipping the calculation if one of the nodes returned DISABLE, right. Not sure this is enough though. With a cluster of size 4 if one node has it disable but three others are not some requests will be balanced and some will not which may result in disproportional load on some nodes.

I have to admit that I don't understand your concern @gleb-cloudius. I don't think disabling HWLB on one node will cause other nodes to receive any more load. When HWLB is disabled on all nodes, the load will be balanced more or less equally among all of them. If HWLB is enabled then the load is distributed based on how hot are the caches. Disabling HWLB on a single node will cause to be more balanced as some requests will be load balanced equally to all nodes while others will still be distributed according to cache hotness. 

Since the goal is to disable HWLB on all nodes, we will get to a state when load is balanced more or less equally anyway. Disabling a single node moves the distribution of the load few steps closer to that final distribution. Why would that be a problem? If in equal distribution each node should receive X requests/s then turning HWLB on one node will increase load on nodes that currently receive less than X req/s and will decrease the load on nodes that currently receive more than X req/s.

@gleb-cloudius could you please explain a bit more what you meant by "disproportional load"? Disproportional to what? total load / number of nodes or hotness of the caches?

gleb-cloudius:
On Fri, Oct 25, 2019 at 04:37:26AM -0700, Piotr Jastrzębski wrote:
> > Ah, we skipping the calculation if one of the nodes returned DISABLE, right. Not sure this is enough though. With a cluster of size 4 if one node has it disable but three others are not some requests will be balanced and some will not which may result in disproportional load on some nodes.
> 
> I have to admit that I don't understand your concern @gleb-cloudius. I don't think disabling HWLB on one node will cause other nodes to receive any more load. When HWLB is disabled on all nodes, the load will be balanced more or less equally among all of them. If HWLB is enabled then the load is distributed based on how hot are the caches. Disabling HWLB on a single node will cause to be more balanced as some requests will be load balanced equally to all nodes while others will still be distributed according to cache hotness. 
> 
> Since the goal is to disable HWLB on all nodes, we will get to a state when load is balanced more or less equally anyway. Disabling a single node moves the distribution of the load few steps closer to that final distribution. Why would that be a problem? If in equal distribution each node should receive X requests/s then turning HWLB on one node will increase load on nodes that currently receive less than X req/s and will decrease the load on nodes that currently receive more than X req/s.
> 
The problem is not that a node with disabled HWLB will stop distributing
a load according to cache hotness, this is indeed fine. The problem is
that it starts to advertise cache hit of zero, so _other_ nodes that
still do not have HWLB disabled will start to distribute load using
incorrect data. In affect they will stop sending load to a nodes with
disabled HWLB.

That said the data distributed over gossiper is used very briefly only
after a node reboots and not yet communicated with other nodes, but we
deemed closing this short window to be important, so I do not see the
reason to break it. 

--
			Gleb.

haaawk:
> The problem is not that a node with disabled HWLB will stop distributing a load according to cache hotness, this is indeed fine. The problem is that it starts to advertise cache hit of zero, so _other_ nodes that still do not have HWLB disabled will start to distribute load using incorrect data. In affect they will stop sending load to a nodes with disabled HWLB. That said the data distributed over gossiper is used very briefly only after a node reboots and not yet communicated with other nodes, but we deemed closing this short window to be important, so I do not see the reason to break it.
> […](#)
> -- Gleb.

This patchset does not work the way you described.

If a node has disabled HWLB, it will not advertise cache hit of zero. It will advertise DISABLED.

_other_ nodes that still do not have HWLB disabled will see that DISABLED and whenever a node that advertises DISABLED is involved in a query, they won't use HWLB at all and balance load equally between all the replicas.

haaawk:
I guess you're thinking about an older version of the code that does not understand DISABLED yet, right? In that case, you're right.

haaawk:
> I guess you're thinking about nodes with an older version of the code that does not understand DISABLED yet, right? In that case, you're right.

The solution for this would be to first do a rolling upgrade with a new version that understands DISABLED but don't disable HWLB and only after that do rolling disable of HWLB but I'm not sure this is a practical solution.

gleb-cloudius:
On Sun, Oct 27, 2019 at 01:26:02AM -0700, Piotr Jastrzębski wrote:
> > I guess you're thinking about nodes with an older version of the code that does not understand DISABLED yet, right? In that case, you're right.
> 
> The solution for this would be to first do a rolling upgrade with a new version that understands DISABLED but don't disable HWLB and only after that do rolling disable of HWLB but I'm not sure this is a practical solution.
> 
Looking at the patch again things looks even worse. table::get_hit_rate()
considers gossiper only if there were no previous communication with the
node. Otherwise it uses the data that was piggy backed to previous reply
(it is expensive to check gossiper on each request and data there is
less up-to-date). And since on a destination node HWLB calculation was
disabled the data will be incorrect (initialized to zero as far as I see).

--
			Gleb.

haaawk:
It seems that the issue in the enterprise was fixed already so closing this PR.

PR #5214, opened 2019-10-23
Fix TTL serialization breakage

### LINKED ISSUE #4855: Updates with TTL fail to replicate in a mixed cluster of 3.0 and 3.1
Reported by @ultrabug via ticket #793

*Installation details*
3.0.9 

Platform (physical/VM/cloud instance type/docker): Gentoo

@ultrabug built current master of scylla and tried upgrading a node of his current 3.0.9 staging cluster with it.

It failed with :

Aug 14 14:44:07 scy-st-1-p1 scylla: [shard 17] storage_proxy - Failed to apply mutation from 10.6.18.12#1: std::out_of_range (deserialization buffer underflow)on the master version node 
Aug 14 14:43:58 scy-st-1-p2 scylla: [shard 1] storage_proxy - Failed to apply mutation from 10.6.18.11#24: std::runtime_error (Truncated frame) 
on the other nodes

Logs available in ticket #793

avikivity:
Commit 93270dd8e04e3caf0e7586311badbaf121c14d3f changed gc_clock to be 64-bit, to fix the Y2038 problem. While 64-bit tombstone::deletion_time is serialized in a compatible way, TTLs (gc_clock::duration) were not.

This patchset reverts TTL serialization to the 32-bit serialization format, and also allows opting-in to the 64-bit format in case a cluster was installed with the broken code. Only Scylla 3.1.0 is vulnerable.

Fixes #4855 

Tests: unit (dev)

### DIFF
--- a/serializer.hh
+++ b/serializer.hh
@@ -303,6 +303,38 @@ struct normalize<bytes_ostream> {
 template <typename T, typename U>
 struct is_equivalent : std::is_same<typename normalize<std::remove_const_t<std::remove_reference_t<T>>>::type, typename normalize<std::remove_const_t <std::remove_reference_t<U>>>::type> {
 };
+
+// gc_clock duration values were serialized as 32-bit prior to 3.1, and
+// are serialized as 64-bit in 3.1.0.
+//
+// TTL values are capped to 20 years, which fits into 32 bits, so
+// truncation is not a concern.
+
+inline bool gc_clock_using_3_1_0_serialization = false;
+
+template <typename Output>
+void
+serialize_gc_clock_duration_value(Output& out, int64_t v) {
+    if (!gc_clock_using_3_1_0_serialization) {
+        // This should have been caught by the CQL layer, so this is just
+        // for extra safety.
+        assert(int32_t(v) == v);
+        serializer<int32_t>::write(out, v);
+    } else {
+        serializer<int64_t>::write(out, v);
+    }
+}
+
+template <typename Input>
+int64_t
+deserialize_gc_clock_duration_value(Input& in) {
+    if (!gc_clock_using_3_1_0_serialization) {
+        return serializer<int32_t>::read(in);
+    } else {
+        return serializer<int64_t>::read(in);
+    }
+}
+
 }
 
 /*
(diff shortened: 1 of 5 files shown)


avikivity:
Note: completely untested, so for early review only,

serializer.hh
```diff
@@ -303,6 +303,61 @@ struct normalize<bytes_ostream> {
...
+void
+serialize_gc_clock_duration_value(Output& out, int64_t v) {
+    if (gc_clock_using_3_1_0_serialization) {
+        serializer<int64_t>::write(out, v);
+    } else {
+        if (int32_t(v) == v) {
+            serializer<int32_t>::write(out, int32_t(v));
+        } else {
+            serializer<int32_t>::write(out, gc_clock_magic_value);
+            serializer<int64_t>::write(out, v);
```
denesb: This assumes that there will be no mixed clusters by the time a legit TTL can overflow the 32 bit value.

serializer.hh
```diff
@@ -303,6 +303,61 @@ struct normalize<bytes_ostream> {
...
+
+constexpr int32_t gc_clock_magic_value = int32_t(0x81828384);
+
+template <typename Output>
+void
+serialize_gc_clock_duration_value(Output& out, int64_t v) {
+    if (gc_clock_using_3_1_0_serialization) {
+        serializer<int64_t>::write(out, v);
+    } else {
+        if (int32_t(v) == v) {
```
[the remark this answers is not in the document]
avikivity: if a 32-bit value is compared with a 64-bit value, the 32-bit value is extended.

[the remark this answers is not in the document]
avikivity: https://godbolt.org/z/Fe3bw4

serializer.hh
avikivity: As @bhalevy TTLs can't really overflow. 31-bit == 68 years. Only time_points can overflow.

avikivity: So we can just serialize as 32-bit, unless 3.1.0. No need for 12-byte serialization.

serializer.hh
```diff
@@ -303,6 +303,40 @@ struct normalize<bytes_ostream> {
...
+
+// gc_clock duration values were serialized as 32-bit prior to 3.1, and
+// are serialized as 64-bit in 3.1.0.
+//
+// TTL values are capped to 20 years, which fits into 32 bits, so
+// truncation is not a concern.
+
+inline bool gc_clock_using_3_1_0_serialization = false;
+
+constexpr int32_t gc_clock_magic_value = int32_t(0x81828384);
```
tgrabiec: Unused.

denesb: Not used anymore.

avikivity:
Update 1: since 32-bit TTLs cannot overflow (they are durations, not time points, so the passage of time doesn't affect them) we now always serialize TTLs as 32-bit (except when talking to 3.1.0 heritage clusters)

Update 2: dropped unused gc_clock_magic_value

avikivity:
@tgrabiec anything else needed?

tgrabiec:
No, I was waiting for Botond to re-review.

PR #5217, opened 2019-10-24
CDC (atomic) delta + (non-optional) pre-image data columns

haaawk:
This PR is a clone of @elcallio's https://github.com/scylladb/scylla/pull/5197. Just rebased on top of more recent master and with few trailing whitespaces removed.

Adds delta and pre-image data column writes for the atomic columns in a cdc-enabled table.

Note that in this patch set it is still unconditional. Adding option support comes in next set.

Uses code more or less derived from alternator to select pre-image, using raw query interface. So should be fairly low overhead to query generation. Pre-image and delta mutations are mixed in with the actual modification mutations to generate the full cdc log (sans post-image).

Rebased on lastest master as of 2019-10-24.

### DIFF
--- a/cdc/cdc.cc
+++ b/cdc/cdc.cc
@@ -20,6 +20,8 @@
  */
 
 #include <utility>
+#include <algorithm>
+
 #include <seastar/util/defer.hh>
 #include <seastar/core/thread.hh>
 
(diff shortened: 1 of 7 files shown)


avikivity:
Dequeued,  debug mode cdc_test fails

PR #5218, opened 2019-10-24
Return preimage from CDC only when it's enabled

haaawk:
Users should be able to decide whether they need preimage or not. There is already an option for that but it's not respected by the implementation. This PR adds support for this functionality.

Tests: unit(dev).

### DIFF
--- a/cdc/cdc.cc
+++ b/cdc/cdc.cc
@@ -583,19 +583,19 @@ future<std::vector<mutation>> append_log_mutations(
         service::storage_proxy::clock_type::time_point timeout,
         service::query_state& qs,
         std::vector<mutation> muts) {
-    auto mp = ::make_lw_shared<std::vector<mutation>>(std::move(muts));
-
-    return get_streams(ctx, s->ks_name(), s->cf_name(), timeout, qs).then([ctx, s = std::move(s), mp, &qs](::shared_ptr<transformer::streams_type> streams) mutable {
-        mp->reserve(2 * mp->size());
-        auto trans = make_lw_shared<transformer>(ctx, s, std::move(streams));
-        auto i = mp->begin();
-        auto e = mp->end();
-        return parallel_for_each(i, e, [ctx, &qs, trans, mp](mutation& m) {
-            return trans->pre_image_select(qs.get_client_state(), db::consistency_level::LOCAL_QUORUM, m).then([trans, mp, &m](lw_shared_ptr<cql3::untyped_result_set> rs) {
-                mp->push_back(trans->transform(m, rs.get()));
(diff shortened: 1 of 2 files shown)


haaawk:
@nyh ping

haaawk:
@nyh ping2

cdc/cdc.cc
```diff
@@ -578,23 +578,40 @@ static future<::shared_ptr<transformer::streams_type>> get_streams(
...
-        return do_with(std::move(muts), transformer(ctx, s, std::move(streams)), [&qs] (std::vector<mutation>& muts, transformer& trans) {
+        return do_with(std::move(muts), transformer(ctx, s, std::move(streams)), [&qs, s] (std::vector<mutation>& muts, transformer& trans) {
             muts.reserve(2 * muts.size());
-            return parallel_for_each(boost::irange(static_cast<decltype(muts.size())>(0), muts.size()), [&qs, &trans, &muts] (int idx) mutable {
-                return trans.pre_image_select(qs.get_client_state(), db::consistency_level::LOCAL_QUORUM, muts[idx]).then([&trans, &muts, idx] (lw_shared_ptr<cql3::untyped_result_set> rs) mutable {
-                    muts.push_back(trans.transform(muts[idx], rs.get()));
+            if (!s->cdc_options().preimage()) {
+                constexpr int batch_size = 100;
+                int muts_len = muts.size();
+                return transform_mutations(muts, batch_size, [&muts, &trans, batch_size, muts_len] (int idx) {
```
nyh: I don't understand (but I'll still commit....) the whole issue of the "batch size", and why without preimage you use a batch size of 100 (why 100? if you have 100,000,000 mutations this still leaves 1 million concurrent operations), but with preimage you use batch size of 1.

haaawk: This is done here to avoid stalls. pre_image_select was written by Calle and I didn't inspect it for stalls.

nyh: With a parallel_for_each, there are no stalls. I assume you used "batch size 100" not to avoid stalls but rather, to reduce parallelism, and only then, you switched back to batch size 1 to avoid stalls. All of these measures look very arbitrary to me... A "batch size 100" indeed reduces parallelism, but 100-fold. Why 100-fold? If the original parallelism was 1 million, is 100-fold enough?
There are other ways to do a loop with reasonable parallelism, such as a do_for_each (not parallel) with a semaphore. I'm not saying that the method you used here is completely broken and needs to be fixed immediately, but I am saying it looks pretty weird and arbitrary and it's not clear what it is trying to guarantee and why.

haaawk: I did batches not to reduce the parallelism but to reduce the overhead of future loops.
My reasoning was as follows:
1. I need to break a compute loop to avoid stalls
2. I assume that there's a constant overhead of a iteration of do_for_each or parallel_for_each because of futures and checks so I decided to batch calls to 'transforms' to reduce the impact constant cost of the future loop has on each iteration.

Your remark about the parallelism limitation is very interesting though and worth keeping in mind.

PR #5223, opened 2019-10-26
reloc: do not install dependencies when building the relocatable package

### LINKED ISSUE #5222: relocatable builds (via reloc/build_reloc.sh) are not reproducible
reloc/build_reloc.sh calls ./install-dependencies.sh. Usually, it ought to do nothing. But if it does something (because a packages was missing in the frozen toolchain), then every build can pick up a different package version, as it is updated.

avikivity:
The dependencies are provided by the frozen toolchain. If a dependency
is missing, we must update the toolchain rather than rely on build-time
installation, which is not reproducible (as different package versions
are available at different times).

Luckily "dnf install" does not update an already-installed package. Had
that been a case, none of our builds would have been reproducible, since
packages would be updated to the latest version as of the build time rather
than the version selected by the frozen toolchain.

So, to prevent missing packages in the frozen toolchain translating to
an unreproducible build, remove the support for installing dependencies
from reloc/build_reloc.sh. We still parse the --nodeps option in case some
script uses it.

Fixes #5222.

Tests: reloc/build_reloc.sh.

### DIFF
--- a/reloc/build_reloc.sh
+++ b/reloc/build_reloc.sh
@@ -14,7 +14,6 @@ print_usage() {
     echo "  --clean                 clean build directory"
     echo "  --compiler PATH         C++ compiler path"
     echo "  --c-compiler PATH       C compiler path"
-    echo "  --nodeps                skip installing dependencies"
     exit 1
 }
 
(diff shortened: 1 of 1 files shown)


avikivity:
ping @scylladb/scylla-maint

nyh:
Committed to next.

PR #5229, opened 2019-10-27
test_udt_mutations fixes

bhalevy:
mutation_test/test_udt_mutations kept failing on my machine and I tracked it down to the 3rd patch in this series (use int64_t constants for long_type).  While at it, this series also fixes a comment and the end iterator in `BOOST_REQUIRE(std::all_of(...))`

mutation_test: test_udt_mutations: fixup udt comment
mutation_test: test_udt_mutations: fix end iterator in call to std::all_of
mutation_test: test_udt_mutations: use int64_t constants for long_type

Test: mutation_test(dev, debug)


kbr-:
LGTM, sorry for the mistake and thanks for fixing it.

bhalevy:
> LGTM, sorry for the mistake and thanks for fixing it.

No problem.  I used it to educate myself a bit about this area of the code... :)

PR #5230, opened 2019-10-27
commitlog: change variadic stream in read_log_file to future<struct>

avikivity:
Since seastar::streams are based on future/promise, variadic streams
suffer the same fate as variadic futures - deprecation and eventual
removal.

This patch therefore replaces a variadic stream in commitlog::read_log_file()
with a non-variadic stream, via a helper struct.

Tests: unit (dev)

### DIFF
--- a/db/commitlog/commitlog.cc
+++ b/db/commitlog/commitlog.cc
@@ -1814,7 +1814,7 @@ const db::commitlog::config& db::commitlog::active_config() const {
 
 // No commit_io_check needed in the log reader since the database will fail
 // on error at startup if required
-future<std::unique_ptr<subscription<fragmented_temporary_buffer, db::replay_position>>>
+future<std::unique_ptr<subscription<db::commitlog::buffer_and_replay_position>>>
 db::commitlog::read_log_file(const sstring& filename, const sstring& pfx, seastar::io_priority_class read_io_prio_class, commit_load_reader_func next, position_type off, const db::extensions* exts) {
     struct work {
     private:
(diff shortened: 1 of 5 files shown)


espindola:
-            std::bind(&impl::process, this, s.get(), std::placeholders::_1,
-                    std::placeholders::_2), p, &exts).then([](auto s) {
+            std::bind(&impl::process, this, s.get(), std::placeholders::_1
+                    ), p, &exts).then([](auto s) {

The indentation looks really odd now.


+            auto&& [buf, rp] = buf_rp;

What is the meaning of && in this context? My guess is that since we
know buf_rp is an rvalue this is the same as a single &. If that is the
case, using a single & is best, no?

LGTM with that.

Cheers,
Rafael

avikivity:
I use auto&& as "inherit rvalue-ness from caller". Usually in normal binding:

   auto&& foo = bar();

if bar returns a regular reference, then foo is a regular reference, and if bar() returns an rvalue, then so is foo (and it's lifetime is then extended). So I used the same scheme here.

avikivity:
Updated: fixed indentation in commitlog_replayer where @espindola pointed out a problem.

espindola:

> I use auto&& as "inherit rvalue-ness from caller". Usually in normal binding:
>
>    auto&& foo = bar();
>
> if bar returns a regular reference, then foo is a regular reference, and if bar() returns an rvalue, then so is foo (and it's lifetime is then extended). So I used the same scheme here.

Yes, which is why it is a bit odd to see it used in

 +            auto&& [buf, rp] = buf_rp;

since it looks like it is trying to do something that auto& would not.

Patch LGTM since that is just a small question of style.

Cheers,
Rafael

avikivity:
I'll leave it to the committing maintainer to make the call, since it's just style.

Issue #3748, opened 2018-09-04
Improve error message


glommer:
If invalid prepared statements are sent down to Scylla, we will return the following exception:

```
        throw exceptions::invalid_request_exception("Invalid amount of bind variables");
```

It can be the case that users may not know where did they made the mistake in their code. Scylla could help them, by making this exception a lot more verbose, and saying exactly what was expected and what was found.

withings-sas:
I agree especially since I have the case where the same query will fail in Scylla but succeed in Cassandra... How can I debug this ?

jul-stas:
Hi @glommer, I can easily tell the number of parameters (expectation vs. reality), but do you think it's enough? Perhaps you have something more on your mind?

avikivity:
Cosmetic change, so not backporting. Removing backport candidate label.

PR #5236, opened 2019-10-28
cql3: error msg w/ arg counts for prepared stmts with wrong arg cnt

jul-stas:
Issue: https://github.com/scylladb/scylla/issues/3748
Very small change: added argument count (expectation vs. reality) to error msg within `invalid_request_exception` when number of bound params does not match the statement.


haaawk:
Add "Fixes #3748" to commit message so that the issue is closed automatically when the commit is merged.

jul-stas:
@nyh Nadav, perhaps you could merge it? (if you like it ofc)

nyh:
Thanks. Committed to "next".

PR #5256, opened 2019-10-31
alternator-test: Skip tests that prevent restart

dekimir:
Although currently marked `xfail`, these tests still run and leave
sstables in some bogus state.  Upon restart, Scylla exits with error
message:

database - Exception while populating keyspace 'alternator' with column family 'alternator_test_1572540235701:hello' from file '.../data/alternator/alternator_test_1572540235701:hello-a10640b2fbfd11e9b09e000000000001': sstables::malformed_sstable_exception (invalid version for file mc-33-big-Filter.db with path .../data/alternator/alternator_test_1572540235701:hello-a10640b2fbfd11e9b09e000000000001. Path doesn't match known pattern.)

### DIFF
--- a/alternator-test/test_gsi.py
+++ b/alternator-test/test_gsi.py
@@ -481,7 +481,7 @@ def test_gsi_5(test_table_gsi_5):
 # "ProjectionType:: KEYS_ONLY" works. We note that it projects both
 # the index's key, *and* the base table's key. So items which had different
 # base-table keys cannot suddenly become the same item in the index.
-@pytest.mark.xfail(reason="GSI not supported")
+@pytest.mark.skip(reason="Leaves sstable in bogus format, preventing Scylla restart")
 def test_gsi_projection_keys_only(dynamodb):
     table = create_test_table(dynamodb,
         KeySchema=[ { 'AttributeName': 'p', 'KeyType': 'HASH' } ],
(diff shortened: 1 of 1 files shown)


nyh:
I don't understand why these specific tests cause this problem, which seems to have to do with files being left behind on a deleted table. Also I can't reproduce this issue on my machine, I don't know why. I think this is a real bug we need to fix, not work around. I'll try to understand the problem and create an issue in the bug tracker. I don't think we should apply this patch as-is, as these three specific tests shouldn't really have any reason to create corrupt sstables.

nyh:
Found and described the real cause for this problem in https://github.com/scylladb/scylla/issues/5280. I don't know why you saw this specifically in these tests - maybe you happened to stop those specific tests in the middle, or something?), but this is actually a much more serious problem, where you can't restart alternator if you have any GSI or LSI! So I'm closing this pull request, and plan to fix that issue instead.

nyh:
@dekimir Avi just committed my patch ("sstables: allow non-traditional characters in table name") to next. It should, I belive, solve this problem without needing to disable any test. If you can easily reproduce this problem, can you please test it again with my patch and confirm it solves the problem you saw? Thanks.

dekimir:
> If you can easily reproduce this problem, can you please test it again with my patch and confirm it solves the problem you saw?

I confirm that the problem exists right before 2fb2eb27a and goes away after it.

PR #5258, opened 2019-10-31
Implement Expected operators LT and GT

dekimir:
Tests: pytest in alternator-test.

### DIFF
--- a/alternator/conditions.cc
+++ b/alternator/conditions.cc
@@ -255,6 +255,12 @@ struct cmp_lt {
     const char* diagnostic() const { return "LT operator"; }
 };
 
+struct cmp_gt {
+    // bytes only has <
+    template <typename T> bool operator()(const T& lhs, const T& rhs) const { return rhs < lhs; }
+    const char* diagnostic() const { return "GT operator"; }
+};
+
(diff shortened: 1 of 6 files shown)


alternator/conditions.cc
```diff
@@ -234,19 +237,24 @@ static bool check_LT(const rjson::value* v1, const rjson::value& v2) {
...
+        return cmp(base64_decode(kv1.value), base64_decode(kv2.value));
     }
-    clogger.error("check_LT panic: LHS type equals RHS type, but one is in {N,S,B} while the other isn't");
+    clogger.error("check_compare panic: LHS type equals RHS type, but one is in {N,S,B} while the other isn't");
     return false;
 }
 
+struct cmp_lt {
+    template <typename T> bool operator()(const T& lhs, const T& rhs) const { return lhs < rhs; }
+    const char* diagnostic() const { return "LT operator"; }
```
nyh: Nitpic - this could have been a string constant instead of a function to make it clear it's just a constant. But this is also ok.

nyh:
Thanks - looks very good. I committed  to next.

PR #5264, opened 2019-11-06
dist/docker: Add SCYLLA_REPO_URL argument to Dockerfile

yaronkaikov:
This change adds a `SCYLLA_REPO_URL` argument to Dockerfile, which defines the RPM repository used to install Scylla from.

When building a new Docker image, users can specify the argument by passing the `--build-arg SCYLLA_REPO_URL=<url>` option to the `docker build` command. If the argument is not specified, the same RPM repository is used as before, retaining the old default behavior.

We intend to use this in release engineering infrastructure to specify RPM repositories for nightly  builds of release branches (for example, 3.1.x), which are currently only using the stable RPMs.

### DIFF
--- a/dist/docker/redhat/Dockerfile
+++ b/dist/docker/redhat/Dockerfile
@@ -4,6 +4,9 @@ MAINTAINER Avi Kivity <avi@cloudius-systems.com>
 
 ENV container docker
 
+# The SCYLLA_REPO_URL argument specifies the URL to the RPM repository this Docker image uses to install Scylla. The default value is the Scylla's unstable RPM repository, which contains the daily build.
+ARG SCYLLA_REPO_URL=http://downloads.scylladb.com/rpm/unstable/centos/master/latest/scylla.repo
+
 ADD scylla_bashrc /scylla_bashrc
 
 # Scylla configuration:
@@ -28,7 +31,7 @@ ADD commandlineparser.py /commandlineparser.py
 ADD docker-entrypoint.py /docker-entrypoint.py
 ADD node_exporter_install /node_exporter_install
 # Install Scylla:
-RUN curl http://downloads.scylladb.com/rpm/unstable/centos/master/latest/scylla.repo -o /etc/yum.repos.d/scylla.repo && \
+RUN curl $SCYLLA_REPO_URL -o /etc/yum.repos.d/scylla.repo && \
     yum -y install epel-release && \
     yum -y clean expire-cache && \
     yum -y update && \


penberg:
@yaronkaikov Please do this against `master` (and make the pull request against `next`). We can then backport the change to 3.1.

penberg:
And please describe the change accurately. This is about introducing a `SCYLLA_REPO_URL` argument to `Dockerfile`, which can be used to define the Scylla repository URL to build a Docker image from.

dist/docker/redhat/Dockerfile
```diff
@@ -28,7 +28,8 @@ ADD commandlineparser.py /commandlineparser.py
 ADD docker-entrypoint.py /docker-entrypoint.py
 ADD node_exporter_install /node_exporter_install
 # Install Scylla:
-RUN curl http://downloads.scylladb.com/rpm/centos/scylla-3.1.repo -o /etc/yum.repos.d/scylla.repo && \
+ARG SCYLLA_REPO_URL=http://downloads.scylladb.com/rpm/centos/scylla-3.1.repo
```
penberg: Let's move this `ARG` definition to the beginning of the `Dockerfile` where it's easier to discover. Let's also add a comment on top of it explaining what it does.

dist/docker/redhat/Dockerfile
```diff
@@ -4,6 +4,9 @@ MAINTAINER Avi Kivity <avi@cloudius-systems.com>
 
 ENV container docker
 
+# default argument for scylla repo url, will be used incase no external argument provided
```
penberg: Can we do something like:

```
```

dist/docker/redhat/Dockerfile
```diff
@@ -4,6 +4,9 @@ MAINTAINER Avi Kivity <avi@cloudius-systems.com>
...
+ARG SCYLLA_REPO_URL=http://downloads.scylladb.com/rpm/centos/scylla-3.1.repo
```
penberg: The value is still incorrect. It needs to be `http://downloads.scylladb.com/rpm/unstable/centos/master/latest/scylla.repo` like the one below which you are replacing.

dist/docker/redhat/Dockerfile
yaronkaikov: Fixed

dist/docker/redhat/Dockerfile
yaronkaikov: Sorry,, fixed

PR #5278, opened 2019-11-14
scylla-gdb: active sstables: support k_l/mc sstable readers

### LINKED ISSUE #5277: scylla-gdb: scylla active sstables does not work with k_l/mc sstable_mutation_reader
`find_single_sstable_readers` looks up active sstable readers using their template arguments.
It currently only looks for `sstable_range_wrapping_reader`, but this applies only to legacy versions (< 2.1 according to comment).

Since version 3.0 we need to support `mp_row_consumer_{k_l,m}`

bhalevy:
Fixes #5277

Tests: (gdb) scylla active sstables

In v3:
- Defined _lookup_obj helper to simplify cache lookup

In v2:
- Simplified construction of `types` list of dicts
- Added break after yielding result
- Added a cache for resolved vtable_addr

### DIFF
--- a/scylla-gdb.py
+++ b/scylla-gdb.py
@@ -840,8 +840,8 @@ def find_single_sstable_readers():
 
     def _lookup_obj(obj_addr, vtable_addr):
         vtable_pfx = 'vtable for '
-        name = resolve(vtable_addr)
-        if not name or not name.startswith(vtable_pfx):
+        name = resolve(vtable_addr, startswith=vtable_pfx)
+        if not name:
             return None
         name = name[len(vtable_pfx):]
         for t in types:
@@ -994,8 +994,8 @@ def find_instances(type_name):
     ptr_type = gdb.lookup_type(type_name).pointer()
     vtable_name = 'vtable for %s ' % type_name
     for obj_addr, vtable_addr in find_vptrs():
-        name = resolve(vtable_addr)
-        if name and name.startswith(vtable_name):
+        name = resolve(vtable_addr, startswith=vtable_name)
+        if name:
             yield gdb.Value(obj_addr).cast(ptr_type)
 
 
@@ -1769,7 +1769,7 @@ class scylla_lsa(gdb.Command):
 names = {}  # addr (int) -> name (str)
 
 
-def resolve(addr, cache=True):
+def resolve(addr, cache=True, startswith=None):
     if addr in names:
         return names[addr]
 
@@ -1778,6 +1778,8 @@ def resolve(addr, cache=True):
         name = None
     else:
         name = infosym[:infosym.find('in section')]
+    if startswith and not name.startswith(startswith):
+        return None
     if cache:
         names[addr] = name
     return name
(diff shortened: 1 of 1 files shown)


bhalevy:
@tgrabiec can you please merge this PR?

scylla-gdb.py
```diff
@@ -838,18 +838,29 @@ def _lookup_type(type_name):
...
-    def _lookup_obj(obj_addr, vtable_addr):
+    def _lookup_obj(obj_addr, vtable_addr, vtable_addr_to_ptr_type):
+        try:
+            # lookup cache
+            return vtable_addr_to_ptr_type[vtable_addr]
+        except KeyError:
+            pass
+        # insert negative entry in cache in case
+        # vtable_addr is not found or not matching any reader type
+        vtable_addr_to_ptr_type[vtable_addr] = None
```
tgrabiec: I think this can make `vtable_addr_to_ptr_type` grow to O(core dump size), which may overflow host's memory.

Rather than optimizing this, we should do a code change similar to 33bef82f6bb30e6623b84a18a2ac245f3292042a

denesb: I suggested this optimization because it greatly reduced the runtime of another (similar) command that I wrote. However I just found out today, that if there are millions of virtual objects in memory (happens often when scylla dumps core), the command still takes hours to run becoming effectively useless :(.
So I agree with Tomek, commands whose runtime scales with the number of virtual objects will be of limited usefullness at best.

I think we can go one step further and link all readers into an intrusive list. Filtering through that list will be much faster than filtering all virtual objects.

bhalevy: I was under the (false) assumption that vtable_addr is O(#types) rather than O(#objects)...

I'll resend just the first, unoptimized, patch then and later we can better track the readers in scylla as Botond suggested.

I see that we have the same scalability issue in `resolve` with addr -> name mapping and it has a `cache` param to enable the cache. We can add a flag to `scylla active sstables` or even a global flag (e.g. `scylla set caching [on|off]` command) that will control default metadata caching in scylla-gdb so we could turn it off if we run out of memory due to having too many objects.

bhalevy:
In v4:
- Removed `vtable_addr_to_ptr_type` patch

In v5:
- Added `scylla-gdb: resolve: add startswith parameter`
  to deal with the potential host memory explosion when calling `resolve` on all values returned by `find_vptrs`

scylla-gdb.py
denesb: `vtable_addr` *is* `O(#types)`. It is the scan of the entire memory that is `O(#objects)`. Caching speeds that up to some extent but if there are millions of objects the scan still takes a lot of time (hours even).

bhalevy: Yes. this pr is not attempting to fix that.
I suggest we merge this one since it does improve functionality until we do the intrusive list of sstable readers, and do the latter as a follow up.

denesb: Agree.

bhalevy:
@tgrabiec please re-review

Issue #4146, opened 2019-01-25
Default secondary index name contains redundant suffix


duarten:
An unnamed index is given the "idx_index" suffix, which is weird and redundant:

Given

```
create table t (p int, c int, v int, primary key (p, c));
create index on t(c);
```

The underlying MV will be named "t_c_idx_index". It should be either "t_c_idx" or "t_c_index".

nyh:
@duarten this redundancy happens because:

1. If the index name passed in "create index" is called xyz, we create a MV called xyz_index.
2. If no index name is given, we generate a name automatically looking like "t_c_idx".

We can't change 2 because it needs to be compatible with Cassandra because that's the only way to drop an index (see issue #3403). Something we could do is to change 1, so that if if the index name is xyz, the MV name will also be xyz. I think it makes sense, but maybe @penberg can comment on why he didn't do this.

Sadly, if we want to support data written by release 3.0 and we decide to change the MV's name, we'll need to have code which will look for both old and new names...

duarten:
On Sun, Jan 27, 2019 at 13:49 nyh <notifications@github.com> wrote:

> @duarten <https://github.com/duarten> this redundancy happens because:
>
>    1. If the index name passed in "create index" is called xyz, we create
>    a MV called xyz_index.
>    2. If no index name is given, we generate a name automatically looking
>    like "t_c_idx".
>
> We can't change 2 because it needs to be compatible with Cassandra because
> that's the only way to drop an index (see issue #3403
> <https://github.com/scylladb/scylla/issues/3403>). Something we could do
> is to change 1, so that if if the index name is xyz, the MV name will also
> be xyz. I think it makes sense, but maybe @penberg
> <https://github.com/penberg> can comment on why he didn't do this.
>
> Sadly, if we want to support data written by release 3.0 and we decide to
> change the MV's name, we'll need to have code which will look for both old
> and new names...
>

Why would this be needed?

nyh:
@duarten because if somebody creates a secondary index with Scylla 3.0, it will create a materialized view called (for example) `t_c_idx_index`. If we change the code in ` indexed_table_select_statement::prepare()` to not add this "_index" (we also need to change `index_table_name()` in secondary_index_manager.cc), using the old index won't work. What I meant was that the code will in prepare() will need to look for both options, to support upgrades... No?

duarten:
On Sun, Jan 27, 2019 at 14:48 nyh <notifications@github.com> wrote:

[...]

PR #5294, opened 2019-11-18
Adding a schema file when creating a snapshot

### LINKED ISSUE #4192: Scylla does not include schema.cql file in snapshot folder

*Installation details*
Scylla version (or git commit hash): 3.0.2-0.20190131.5174b1cd1
Cluster size: 3
OS (RHEL/CentOS/Ubuntu/AWS AMI): Ubuntu 16.04

Cassandra provides schema.cql file in snapshot folder (for each table, next to SSTables and others), that describes the schema at the time of the snapshot.
Scylla does not behave this way.

This functionality makes backup/restore on Cassandra easier and should probably be part of Scylla as well.

amnonh:
To use a snapshot we need a schema file that is similar to the result of running cql DESCRIBE command.

The DESCRIBE is implemented in the cql driver so the functionality needs to be re-implemented inside scylla.

This series adds a describe method to the schema file and use it when doing a snapshot.

There are different approach of how to handle materialize views and secondary indexes. 

This implementation creates each schema.cql file in its own relevant directory, so the schema for materializing view, for example, will be placed in the snapshot directory of the table of that view.

Fixes #4192

### DIFF
--- a/database.hh
+++ b/database.hh
@@ -827,6 +827,24 @@ public:
     future<> snapshot(sstring name);
     future<std::unordered_map<sstring, snapshot_details>> get_snapshot_details();
 
+    /*!
+     * \brief write the schema to a 'schema.cql' file at the given directory.
+     *
+     * When doing a snapshot, the snapshot directory contains a 'schema.cql' file
+     * with a CQL command that can be used to generate the schema.
+     * The content is is similar to the result of the CQL DESCRIBE command of the table.
+     *
+     * When a schema has indexes, local indexes or views, those indexes and views
+     * are represented by their own schemas.
+     * In those cases, the method would write the relevant information for each of the schemas:
+     *
+     * The schema of the base table would output a file with the CREATE TABLE command
+     * and the schema of the view that is used for the index would output a file with the
+     * CREATE INDEX command.
+     * The same is true for local index and MATERIALIZED VIEW.
+     */
+    future<> write_schema_as_cql(sstring dir) const;
+
     const bool incremental_backups_enabled() const {
         return _config.enable_incremental_backups;
     }
--- a/index/secondary_index_manager.cc
+++ b/index/secondary_index_manager.cc
@@ -50,6 +50,7 @@
 
 #include <boost/range/adaptor/map.hpp>
 #include <boost/algorithm/cxx11/any_of.hpp>
+#include <boost/algorithm/string/predicate.hpp>
 
 namespace secondary_index {
 
@@ -98,6 +99,13 @@ void secondary_index_manager::add_index(const index_metadata& im) {
 
 sstring index_table_name(const sstring& index_name) {
     return format("{}_index", index_name);
+}
+
+sstring index_name_from_table_name(const sstring& table_name) {
+    if (table_name.size() < 7 || !boost::algorithm::ends_with(table_name, "_index")) {
+        throw std::runtime_error(format("Table {} does not have _index suffix", table_name));
+    }
+    return table_name.substr(0, table_name.size() - 6); // remove the _index suffix from an index name;
 }
 
 static bytes get_available_token_column_name(const schema& schema) {
--- a/schema.cc
+++ b/schema.cc
@@ -34,6 +34,8 @@
 #include <boost/algorithm/cxx11/any_of.hpp>
 #include "view_info.hh"
 #include "partition_slice_builder.hh"
+#include "database.hh"
+#include "service/storage_service.hh"
 
 constexpr int32_t schema::NAME_LENGTH;
 
@@ -635,6 +637,159 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
         os << ", viewInfo=" << *s.view_info();
     }
     os << "]";
+    return os;
+}
+
+std::ostream& map_as_cql_param(std::ostream& os, const std::map<sstring, sstring>& map, bool first = true) {
+    for (auto i: map) {
+        if (first) {
+            first = false;
+        } else {
+            os << ",";
+        }
+        os << "'" << i.first << "': '" << i.second << "'";
+    }
+
+    return os;
+}
+
+std::ostream& column_definition_as_cql_key(std::ostream& os, const column_definition & cd) {
+    os << cd.name_as_cql_string();
+    os << " " << cd.type->cql3_type_name();
+
+    if (cd.kind == column_kind::static_column) {
+        os << " STATIC";
+    }
+    return os;
+}
+
+static bool is_global_index(const utils::UUID& id, const schema& s) {
+    return  service::get_local_storage_service().db().local().find_column_family(id).get_index_manager().is_global_index(s);
+}
+
+static bool is_index(const utils::UUID& id, const schema& s) {
+    return  service::get_local_storage_service().db().local().find_column_family(id).get_index_manager().is_index(s);
+}
+
+
+std::ostream& schema::describe(std::ostream& os) const {
+    os << "CREATE ";
+    int n = 0;
+
+    if (is_view()) {
+        if (is_index(view_info()->base_id(), *this)) {
+            auto is_local = !is_global_index(view_info()->base_id(), *this);
+
+            os << "INDEX " << cql3::util::maybe_quote(secondary_index::index_name_from_table_name(cf_name())) << " ON "
+                    << cql3::util::maybe_quote(ks_name()) << "." << cql3::util::maybe_quote(view_info()->base_name()) << "(";
+            if (is_local) {
+                os << "(";
+            }
+            for (auto& pk : partition_key_columns()) {
+                if (n++ != 0) {
+                    os << ", ";
+                }
+                os << pk.name_as_cql_string();
+            }
+            if (is_local) {
+                os << ")";
+                if (!clustering_key_columns().empty()) {
+                    os << ", " << clustering_key_columns().front().name_as_cql_string();
+                }
+            }
+            os <<");\n";
+            return os;
+        } else {
+            os << "MATERIALIZED VIEW " << cql3::util::maybe_quote(ks_name()) << "." << cql3::util::maybe_quote(cf_name()) << " AS\n";
+            os << "    SELECT ";
+            for (auto& cdef : all_columns()) {
+                if (cdef.is_hidden_from_cql()) {
+                    continue;
+                }
+                if (n++ != 0) {
+                    os << ", ";
+                }
+                os << cdef.name_as_cql_string();
+            }
+            os << "\n    FROM " << cql3::util::maybe_quote(ks_name()) << "." <<  cql3::util::maybe_quote(view_info()->base_name());
+            os << "\n    WHERE " << view_info()->where_clause();
+        }
+    } else {
+        os << " TABLE " << cql3::util::maybe_quote(ks_name()) << "." << cql3::util::maybe_quote(cf_name()) << " (";
+        for (auto& cdef : all_columns()) {
+            os << "\n    ";
+            column_definition_as_cql_key(os, cdef);
+            os << ",";
+        }
+    }
+
+    os << "\n    PRIMARY KEY (";
+    if (partition_key_columns().size() > 1) {
+        os << "(";
+    }
+    n = 0;
+    for (auto& pk : partition_key_columns()) {
+        if (n++ != 0) {
+            os << ", ";
+        }
+        os << pk.name_as_cql_string();
+    }
+    if (partition_key_columns().size() > 1) {
+        os << ")";
+    }
+    for (auto& pk : clustering_key_columns()) {
+        os << ", ";
+        os << pk.name_as_cql_string();
+    }
+    os << ")";
+    if (is_view()) {
+        os << "\n    ";
+    } else {
+        os << "\n) ";
+    }
+    os << "WITH ";
+    if (!clustering_key_columns().empty()) {
+        // Adding clustering key order can be optional, but there's no harm in doing so.
+        os << "CLUSTERING ORDER BY (";
+        n = 0;
+        for (auto& pk : clustering_key_columns()) {
+            if (n++ != 0) {
+                os << ", ";
+            }
+            os << pk.name_as_cql_string();
+            if (pk.type->is_reversed()) {
+                os << " DESC";
+            } else {
+                os << " ASC";
+            }
+        }
+        os << ")\n    AND ";
+    }
+    if (is_compact_table()) {
+        os << "COMPACT STORAGE\n    AND ";
+    }
+    os << "bloom_filter_fp_chance = " << bloom_filter_fp_chance();
+    os << "\n    AND caching = {";
+    map_as_cql_param(os, caching_options().to_map());
+    os << "}";
+    os << "\n    AND comment = '" << comment()<< "'";
+    os << "\n    AND compaction = {'class': '" <<  sstables::compaction_strategy::name(compaction_strategy()) << "'";
+    map_as_cql_param(os, compaction_strategy_options()) << "}";
+    os << "\n    AND compression = {";
+    map_as_cql_param(os,  get_compressor_params().get_options());
+    os << "}";
+
+    os << "\n    AND crc_check_chance = " << crc_check_chance();
+    os << "\n    AND dclocal_read_repair_chance = " << dc_local_read_repair_chance();
+    os << "\n    AND default_time_to_live = " << default_time_to_live().count();
+    os << "\n    AND gc_grace_seconds = " << gc_grace_seconds().count();
+    os << "\n    AND max_index_interval = " << max_index_interval();
+    os << "\n    AND memtable_flush_period_in_ms = " << memtable_flush_period();
+    os << "\n    AND min_index_interval = " << min_index_interval();
+    os << "\n    AND read_repair_chance = " << read_repair_chance();
+    os << "\n    AND speculative_retry = '" << speculative_retry().to_sstring() << "';";
+    os << "\n";
+
     return os;
 }
 
--- a/schema.hh
+++ b/schema.hh
@@ -909,6 +909,24 @@ public:
     // Search for an existing index with same kind and options.
     std::optional<index_metadata> find_index_noname(const index_metadata& target) const;
     friend std::ostream& operator<<(std::ostream& os, const schema& s);
+    /*!
+     * \brief stream the CQL DESCRIBE output.
+     *
+     * CQL DESCRIBE is implemented at the driver level. This method mimic that functionality
+     * inside Scylla.
+     *
+     * The output of DESCRIBE is the CQL command to create the described table with its indexes and views.
+     *
+     * For tables with Indexes or Materialized Views, the CQL DESCRIBE is split between the base and view tables.
+     * Calling the describe method on the base table schema would result with the CQL "CREATE TABLE"
+     * command for creating that table only.
+     *
+     * Calling the describe method on a view schema would result with the appropriate "CREATE MATERIALIZED VIEW"
+     * or "CREATE INDEX" depends on the type of index that schema describes (ie. Materialized View, Global
+     * Index or Local Index).
+     *
+     */
+    std::ostream& describe(std::ostream& os) const;
     friend bool operator==(const schema&, const schema&);
     const column_mapping& get_column_mapping() const;
     friend class schema_registry_entry;
--- a/table.cc
+++ b/table.cc
@@ -1526,7 +1526,7 @@ const std::vector<sstables::shared_sstable>& table::compacted_undeleted_sstables
 
 inline bool table::manifest_json_filter(const fs::path&, const directory_entry& entry) {
     // Filter out directories. If type of the entry is unknown - check its name.
-    if (entry.type.value_or(directory_entry_type::regular) != directory_entry_type::directory && entry.name == "manifest.json") {
+    if (entry.type.value_or(directory_entry_type::regular) != directory_entry_type::directory && (entry.name == "manifest.json" || entry.name == "schema.cql")) {
         return false;
     }
 
@@ -1713,6 +1713,27 @@ seal_snapshot(sstring jsondir) {
     });
 }
 
+future<> table::write_schema_as_cql(sstring dir) const {
+    std::ostringstream ss;
+    try {
+        this->schema()->describe(ss);
+    } catch (...) {
+        return make_exception_future<>(std::current_exception());
+    }
+    auto schema_description = ss.str();
+    auto schema_file_name = dir + "/schema.cql";
+    return open_checked_file_dma(general_disk_error_handler, schema_file_name, open_flags::wo | open_flags::create | open_flags::truncate).then([schema_description = std::move(schema_description)](file f) {
+        return do_with(make_file_output_stream(std::move(f)), [schema_description  = std::move(schema_description)] (output_stream<char>& out) {
+            return out.write(schema_description.c_str(), schema_description.size()).then([&out] {
+               return out.flush();
+            }).then([&out] {
+               return out.close();
+            });
+        });
+    });
+
+}
+
 future<> table::snapshot(sstring name) {
     return flush().then([this, name = std::move(name)]() {
        return with_semaphore(_sstable_deletion_sem, 1, [this, name = std::move(name)]() {
@@ -1747,7 +1768,7 @@ future<> table::snapshot(sstring name) {
                     auto rf = f.substr(sst->get_dir().size() + 1);
                     table_names.insert(std::move(rf));
                 }
-                return smp::submit_to(shard, [requester = engine().cpu_id(), jsondir = std::move(jsondir),
+                return smp::submit_to(shard, [requester = engine().cpu_id(), jsondir = std::move(jsondir), this,
                                               tables = std::move(table_names), datadir = _config.datadir] {
 
                     if (pending_snapshots.count(jsondir) == 0) {
@@ -1762,10 +1783,15 @@ future<> table::snapshot(sstring name) {
                     auto my_work = make_ready_future<>();
                     if (requester == engine().cpu_id()) {
                         my_work = snapshot->requests.wait(smp::count).then([jsondir = std::move(jsondir),
-                                                                            snapshot] () mutable {
-                            return seal_snapshot(jsondir).then([snapshot] {
-                                snapshot->manifest_write.signal(smp::count);
+                                                                            snapshot, this] {
+                            return write_schema_as_cql(jsondir).handle_exception([jsondir](std::exception_ptr ptr) {
+                                tlogger.error("Failed writing schema file in snapshot in {} with exception {}", jsondir, ptr);
                                 return make_ready_future<>();
+                            }).finally([jsondir = std::move(jsondir), snapshot] () mutable {
+                                return seal_snapshot(jsondir).then([snapshot] {
+                                    snapshot->manifest_write.signal(smp::count);
+                                    return make_ready_future<>();
+                                });
                             });
                         });
                     }
@@ -1814,7 +1840,7 @@ future<std::unordered_map<sstring, table::snapshot_details>> table::get_snapshot
                         //
                         // All the others should just generate an exception: there is something wrong, so don't blindly
                         // add it to the size.
-                        if (name != "manifest.json") {
+                        if (name != "manifest.json" && name != "schema.cql") {
                             sstables::entry_descriptor::make_descriptor(snapshot_dir.native(), name);
                             all_snapshots.at(snapshot_name).total += size;
                         } else {
--- a/test/boost/cql_query_test.cc
+++ b/test/boost/cql_query_test.cc
@@ -49,6 +49,7 @@
 #include "json.hh"
 #include "schema_builder.hh"
 #include "service/migration_manager.hh"
+#include <regex>
 
 using namespace std::literals::chrono_literals;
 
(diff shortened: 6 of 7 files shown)


view_info.hh
```diff
@@ -58,6 +58,10 @@ public:
         return _raw.where_clause();
     }
 
+    /*!
+     * \brief returns true if this view is backing a secondary index
+     */
+    bool is_secondary_index() const;
```
denesb: This is a layering violation, a materialized view shouldn't know about different things it is used for as an infrastructure. In fact we used to have this method and @psarna removed it in an effort to fix this layering violation.

schema.cc
```diff
@@ -656,6 +656,135 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
...
+    return os;
+}
+
+std::ostream& schema::describe(std::ostream& os) const {
+    os << "CREATE ";
+    int n = 0;
+
+    if (is_view()) {
+        if (view_info()->is_secondary_index()) {
+            os << "INDEX " << cf_name().substr(0, cf_name().size() - 6) << " ON " << ks_name() << "." << view_info()->base_name() << "(";
```
denesb: What is this `- 6`?

denesb:
Please write unit tests that verify that parsing back the output of `schema::describe()` results in a schema equivalent to the original one. You can use `tests/random_schema.hh` to generate random schemas but you should also generate random (but valid) schema properties as well (random_schema doesn't do that). You can then generate a materialized view and secondary index for the random schema and check that they describe correctly as well.

view_info.hh
amnonh: so you remove the method, but still do that in the code,
see the code in: cql3::statements::select_statement& view_info::select_statement()

I'll place the logic somewhere else

amnonh:
ok, naturally I checked it with the different options, but I'll add tests for it as well.

schema.cc
amnonh: remove the _index prefix I'll make it clearer

view_info.hh
psarna: The code in `view_info::select_statement` is a necessary evil due to backward compatibility, it's marked with this comment:
```cpp
// FIXME(sarna): legacy code, should be removed after "computed_columns" feature is guaranteed
// to be available on every node. Then, we won't need to check if this view is backing a secondary index.
```
It would be best to avoid checking if if the view backs an index in the view code itself, but instead on an upper layer (e.g. a code that iterates over all views of a table could exclude ones backed by an index, because a table knows its indexes and their accompanying views)

amnonh: @psarna you're right but that is not the case I'm facing, I'm writing a describe method in a schema (similar to a cql DESCRIBE command) and if the schema in hand is a view I need to know how to treat it.

Anyway, as I replied to @denesb I'll move the same hack that is used here to somewhere else for my need.

avikivity:
What does "A schema file when doing a snapshot" mean? I can't parse it.

amnonh:
I'm adding some tests and I'll update the description when I'm pushing the next version

amnonh:
@denesb I've forced push V2.

Main changes, I remove the is_index from the view layer and mimic that functionality where it is needed.

I've added some basic testing for a table, an index and materialized view.

schema.cc
```diff
@@ -656,6 +658,145 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
...
+        os << " STATIC";
+    }
+    return os;
+}
+
+static sstring remove_index_suffix(const sstring& name) {
+    return name.substr(0, name.size() - 6); // remove the _index suffix from an index name;
+}
+
+static bool is_secondery_index(const utils::UUID& id, const schema& s) {
```
psarna: there's a typo here :) is_second**a**ry_index

Also, the commit message  has `schmea` instead of `schema`

amnonh: @psarna thanks, I've forced push V3 with a fix

slivne:
@psarna please comment / approve the last version

tests/cql_query_test.cc
```diff
@@ -4261,3 +4261,135 @@ SEASTAR_TEST_CASE(test_rf_expand) {
...
+                "    AND compression = {'sstable_compression': 'org.apache.cassandra.io.compress.LZ4Compressor'}\n"
+                "    AND crc_check_chance = 1\n"
+                "    AND dclocal_read_repair_chance = 0.1\n"
+                "    AND default_time_to_live = 0\n"
+                "    AND gc_grace_seconds = 864000\n"
+                "    AND max_index_interval = 2048\n"
+                "    AND memtable_flush_period_in_ms = 0\n"
+                "    AND min_index_interval = 128\n"
+                "    AND read_repair_chance = 0\n"
+                "    AND speculative_retry = '99.0PERCENTILE';\n");
```
psarna: What about comparing the schemas with `cmp_schema` for this case?

Also, I think this big description strings should be moved to a string variable and used both during table creation and here. Reasons for that:
1. You start by creating the table without specifying all these configuration variables like `bloom_filter_fp_chance`, and then compare the results *with* these values - which means that every single change in the default value will also unnecessarily invalidate this test case.
2. It will simply result in less code duplication.

tests/cql_query_test.cc
```diff
@@ -4261,3 +4261,135 @@ SEASTAR_TEST_CASE(test_rf_expand) {
...
+        auto new_schema = e.local_db().find_schema("ks", "cf_view");
+
+        cmp_schema(schema, new_schema);
+
+        e.execute_cql("CREATE INDEX cf_index ON ks.cf (col2);").get();
+        schema = e.local_db().find_schema("ks", "cf_index_index");
+
+        std::ostringstream ss1;
+        schema->describe(ss1);
+        BOOST_CHECK_EQUAL(ss1.str(), "CREATE INDEX cf_index ON ks.cf(col2);\n");
```
psarna: again, shouldn't schemas be compared again with `cmp_schema` after the second change?

tests/cql_query_test.cc
```diff
@@ -4261,3 +4261,135 @@ SEASTAR_TEST_CASE(test_rf_expand) {
...
+
+
+
+/*
+ * The following tests are for schema describe command
+ */
+
+static void cmp_schema(schema_ptr s1, schema_ptr s2) {
+    BOOST_CHECK_EQUAL(s1->ks_name(), s2->ks_name());
+    BOOST_CHECK_EQUAL(s1->cf_name(), s2->cf_name());
```
psarna: One more really easy check to add would be to check if the schemas have the same column names (and perhaps types). There are multiple ways of listing column names from the schema, one of them is `columns_by_name()`, which gives you a map where keys are ordered column names - easy to compare.

I'm also fine with leaving it as a todo for the future.

slivne:
@amnonh ping

slivne:
@amnonh ping

tests/cql_query_test.cc
amnonh: @psarna I've forced push a new version, switching to a single string to create a table and to compare the generated schema to remove some of the tests.

Because  it's the same string, we know it works (or there wouldn't be a schema)

table.cc
```diff
@@ -1747,7 +1764,7 @@ future<> table::snapshot(sstring name) {
                     auto rf = f.substr(sst->get_dir().size() + 1);
                     table_names.insert(std::move(rf));
                 }
-                return smp::submit_to(shard, [requester = engine().cpu_id(), jsondir = std::move(jsondir),
+                return smp::submit_to(shard, [requester = engine().cpu_id(), jsondir = std::move(jsondir),this,
```
psarna: I only found one super-tiny style mistake - please put a space between the comma and 'this' - I can push the series after this.

amnonh: I have forced push an updated version with the missing space

psarna:
queued to /next

psarna:
Dequeued: a unit test failed (database_test).

Please fix and include an information that unit tests were run, e.g.:
```
Tests: unit(dev)
```
Also, note that I'm starting vacation in an hour, so this pull request likely needs attention of another maintainer :)

amnonh:
@psarna Enjoy your vacation.

Found the issue, the problem is with a different test that needs to be changed

amnonh:
I have forced push a new version, get_snapshot_details in table.cc should ignore the schema.cql file or the relevant test will fail.

slivne:
@nyh can you please help ... sarna is on vacation

@amnonh please rebase

nyh:
On Sun, Dec 22, 2019 at 12:36 PM Shlomi Livne <notifications@github.com>
wrote:

> @nyh <https://github.com/nyh> can you please help ... sarna is on vacation
>
> @amnonh <https://github.com/amnonh> please rebase
>

I'll review it now, but to your long list of problems in our long review
process you can add another one:
When you switch reviewer in the middle of a long review, the new reviewer
may have different opinions and care about new things that the original
reviewer never cared about, resulting in one of two options, I don't know
which of them is worse:
1. The new reviewer may demand new things and slow the patch's commit even
more.
2. The new reviewer may commit the patch contrary to his personal opinions,
because the previous reviewer as almost fine with this patch.

table.cc
```diff
@@ -1713,6 +1713,23 @@ seal_snapshot(sstring jsondir) {
     });
 }
 
+future<> table::write_schema(sstring dir) const {
+    std::ostringstream ss;
+    this->schema()->describe(ss);
+    auto json = ss.str();
+    auto jsonfile = dir + "/schema.cql";
+    return open_checked_file_dma(general_disk_error_handler, jsonfile, open_flags::wo | open_flags::create | open_flags::truncate).then([json](file f) {
+        return do_with(make_file_output_stream(std::move(f)), [json] (output_stream<char>& out) {
```
nyh: I think you can move json (which isn't json...) instead of copying it?

tests/cql_query_test.cc
```diff
@@ -4261,3 +4261,106 @@ SEASTAR_TEST_CASE(test_rf_expand) {
...
+        e.execute_cql(create_normal_table).get();
+        auto schema = e.local_db().find_schema("ks", "cf");
+        std::ostringstream ss;
+
+        schema->describe(ss);
+        BOOST_CHECK_EQUAL(ss.str(), create_normal_table);
+        e.execute_cql("DROP TABLE ks.cf").get();
+        e.execute_cql(ss.str()).get();
+
+        std::string create_table_with_compund_key = "CREATE  TABLE ks.cf1 (\n"
```
nyh: There are other variants you should try, e.g., multiple partition key columns (which is a special case you have special code with parentheses for), and importantly - uppercase table names, quoted column names, etc.
To reduce the amount of duplicate code, you can have an array of strings with just the different part (appending at the end the common part with all those silly parameters), and do the test with a loop instead of duplicate code.

tests/cql_query_test.cc
```diff
@@ -4261,3 +4261,106 @@ SEASTAR_TEST_CASE(test_rf_expand) {
...
+        schema->describe(ss);
+        BOOST_CHECK_EQUAL(ss.str(), create_materialized_view);
+        std::string create_index = "CREATE INDEX cf_index ON ks.cf(col2);\n";
+        e.execute_cql(create_index).get();
+        schema = e.local_db().find_schema("ks", "cf_index_index");
+
+        std::ostringstream ss1;
+        schema->describe(ss1);
+        BOOST_CHECK_EQUAL(ss1.str(), create_index);
+    });
```
nyh: Please test at the end of this test that describe() of "cf", (not "cf_view" or "cf_index_index"). I don't think it's obvious whether describe of a table *with* materialized views should list them or not - but since in the code we decided NOT to list them, the test should verify this: describe() of cf, despite it having a materialized view and an index, should contain CQL to just create cf, not its view.

schema.cc
```diff
@@ -656,6 +658,145 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
...
+}
+
+static sstring remove_index_suffix(const sstring& name) {
+    return name.substr(0, name.size() - 6); // remove the _index suffix from an index name;
+}
+
+static bool is_secondary_index(const utils::UUID& id, const schema& s) {
+    return  service::get_local_storage_service().db().local().find_column_family(id).get_index_manager().is_global_index(s);
+}
+std::ostream& schema::describe(std::ostream& os) const {
```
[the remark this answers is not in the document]
amnonh: Sure, I'll add some explenation

schema.cc
```diff
@@ -656,6 +658,145 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
...
+    os << "CREATE ";
+    int n = 0;
+
+    if (is_view()) {
+        if (is_secondary_index(view_info()->base_id(), *this)) {
+            os << "INDEX " << remove_index_suffix(cf_name()) << " ON " << ks_name() << "." << view_info()->base_name() << "(";
```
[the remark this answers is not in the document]
amnonh: it's based on a comment from the previous committer it removes the _index suffix from the index name, if you didn't find it clear, I'll think of a different name for it.

schema.cc
```diff
@@ -656,6 +658,145 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
...
+            for (auto& pk : partition_key_columns()) {
```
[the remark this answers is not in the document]
amnonh: I'll add one

schema.cc
```diff
@@ -656,6 +658,145 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
...
+                if (n++ != 0) {
+                    os << ", ";
+                }
+                os << pk.name_as_cql_string();
+            }
+            os <<");\n";
+            return os;
+        } else {
+            os << "MATERIALIZED VIEW " << ks_name() << "." << cf_name() << " AS\n";
```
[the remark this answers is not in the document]
amnonh: it should work, but I'll verify.

schema.cc
```diff
@@ -656,6 +658,145 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
...
+                if (n++ != 0) {
+                    os << ", ";
+                }
+                os << cdef.name_as_cql_string();
+            }
+            os << "\n    FROM " << view_info()->base_name();
+            os << "\n    WHERE " << view_info()->where_clause();
+        }
+    } else {
+        os << " TABLE " << ks_name() << "." << cf_name() << " (";
```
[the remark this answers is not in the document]
amnonh: I'll add test for it

schema.cc
```diff
@@ -656,6 +658,145 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
...
+    }
+    if (partition_key_columns().size() > 1) {
+        os << ")";
+    }
+    for (auto& pk : clustering_key_columns()) {
+        os << ", ";
+        os << pk.name_as_cql_string();
+    }
+    os << ")";
+    if (is_view()) {
```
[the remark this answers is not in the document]
amnonh: I'll check

table.cc
```diff
@@ -1713,6 +1713,23 @@ seal_snapshot(sstring jsondir) {
     });
 }
 
+future<> table::write_schema(sstring dir) const {
+    std::ostringstream ss;
+    this->schema()->describe(ss);
+    auto json = ss.str();
```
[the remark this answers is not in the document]
amnonh: yes, you're right, I'll change it

tests/cql_query_test.cc
amnonh: sounds good, I'll squeeze it into a single test with a loop

tests/cql_query_test.cc
amnonh: good point, I'll add a check for that as well

schema.cc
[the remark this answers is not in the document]
nyh: The problem is that by default, there are two "index suffixes" see issue #4146 - and by this name it's impossible to guess which of these two "suffixes" the function removes.
Maybe the name is ok if it includes a clear comment on what it does, but I don't see the definition of this function - again - where is it? a separate patch? This function should be in `index/secondary_index_manager.cc`, right next to `index_table_name()`. You could call your new function `from_index_table_name()` to make it clearer what it does, and that it is the reverse of `index_table_name()`.

schema.cc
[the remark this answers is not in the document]
nyh: I don't see how this can work without being fixed. If cf_name is FooBar (capitalized like this), the only correct way to create is with the CQL command:
```
CREATE TABLE "FooBar"
```
Like this - with double quotes. Without the double quotes, it will just create a table called foobar - and this is not the table you wanted to create. Fixing this is trivial - you can use the existing function, `cql3::util::maybe_quote(sstring(cf_name))` to do this CQL quoting if necessary.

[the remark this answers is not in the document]
amnonh: I already tested it again and validate that there is an issue with the table name. The next version will take care of that and I have tests that validate that it's correct.

While the parameters are quoted correctly, that table/ks wheren't (but it will be ok)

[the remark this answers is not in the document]
nyh: Right. The column names you did quote correctly - just not the table names (the index name, view name, and table name, can all have this issue).

amnonh:
@nyh Thanks for the review. I'm just forced pushed a new version. Capitalization of table and views are now handled correctly and is tested. 
Additional changes, I moved the function that returns the index name to secondary_index_manager and rename it.

I also cleaned the variables that used json in their names.

slivne:
@nyh ping

index/secondary_index_manager.cc
```diff
@@ -100,6 +100,10 @@ sstring index_table_name(const sstring& index_name) {
     return format("{}_index", index_name);
 }
 
+sstring index_name_from_table_name(const sstring& table_name) {
+    return table_name.substr(0, table_name.size() - 6); // remove the _index suffix from an index name;
```
nyh: It would have been better to check and throw an exception if the table_name doesn't end with "_index". It's a very cheap test (this is a very rarely called function) and can help serious bugs in using this function.

schema.hh
```diff
@@ -909,6 +909,15 @@ public:
     // Search for an existing index with same kind and options.
     std::optional<index_metadata> find_index_noname(const index_metadata& target) const;
     friend std::ostream& operator<<(std::ostream& os, const schema& s);
+    /*!
+     * \brief print the CQL DESCRIBE output.
+     *
+     * CQL DESCRIBE is implemented at the driver level. This method mimic that functionality
+     * inside Scylla.
+     *
+     * It should be possible to use the method result to re-create the schema.
```
nyh: Please be more explicit about what this function does schemas which **have** a view or an index, and for schemas which **are** a view or index itself. For example something like:
This function returns a "CREATE TABLE", "CREATE MATERIALIZED VIEW" or "CREATE INDEX" command depending on whether the given schema is a base table, a materialized view or a materialized view created for a secondary index. If the given schema is a base table, only the command to create the table itself is returned - its materialized views or indexes, if any, need to be created separately.

schema.cc
```diff
@@ -638,6 +640,141 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
...
+    os << " " << cd.type->cql3_type_name();
+
+    if (cd.kind == column_kind::static_column) {
+        os << " STATIC";
+    }
+    return os;
+}
+
+static bool is_secondary_index(const utils::UUID& id, const schema& s) {
+    return  service::get_local_storage_service().db().local().find_column_family(id).get_index_manager().is_global_index(s);
```
nyh: You left out local indexes. Your patch is incomplete without recognizing those, and also recreating them with the proper CQL command. And of course, you need to add a test case for them, which will show you you missed them. I already pointed out this in my previous round of review - see my other comment below which remains.

schema.cc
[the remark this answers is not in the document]
nyh: Did you? You have to. Otherwise, your code will seem to work until someone happens to use it with local secondary indexes, and things just silently break :-(
We *can* commit code without the local-secondary-index part and fix this bug later, hoping nobody uses local secondary indexes (?) but that's not a technical decision but rather a decision on how quickly we need a almost-working fix in. I have no input on that. Maybe @slivne does?

slivne:
we need secondary indexes to work if anyone is in doubt

database.hh
```diff
@@ -827,6 +827,10 @@ public:
     future<> snapshot(sstring name);
     future<std::unordered_map<sstring, snapshot_details>> get_snapshot_details();
 
+    /*!
+     * \brief write the table schema to a schema file at the given directory
```
nyh: Would have been nice to give more context in this comment... That this function has to do with snapshots (this "schema file" isn't something we use normally in normal database operation), and the "schema file" is really a CQL command to recreate the schema.
I think it would also be better to rename this function **write_snapshot_schema** or similar, to make clearer this function is just used by snapshots (if "snapshot" is the right word... Or is it "backup"?).

schema.cc
```diff
@@ -638,6 +640,141 @@ std::ostream& operator<<(std::ostream& os, const schema& s) {
...
+        }
+        os << "'" << i.first << "': '" << i.second << "'";
+    }
+
+    return os;
+}
+
+std::ostream& column_definition_as_cql_key(std::ostream& os, const column_definition & cd) {
+    os << cd.name_as_cql_string();
+    os << " " << cd.type->cql3_type_name();
```
nyh: Please consider: What happens when UDT are used? Do we need to save the UDT definition in the CQL file as well? Does this cql3_type_name() method (I'm not familiar with it) work correctly also with UDTs and also with UDTs with quoted names (I believe this is possible, but didn't check...)
If UDTs don't work, we can leave it as a TODO, but we should throw an exception from describe() rather than silently doing the wrong thing.

test/boost/cql_query_test.cc
```diff
@@ -4313,3 +4315,198 @@ SEASTAR_TEST_CASE(test_rf_expand) {
...
+ */
+std::string normalize_white_space(const std::string& str) {
+  return std::regex_replace(" " + str + " ", std::regex("\\s+"), " ");
+}
+
+SEASTAR_TEST_CASE(test_describe_simple_schema) {
+    return do_with_cql_env_thread([] (cql_test_env& e) {
+        std::unordered_map<std::string, std::string> cql_create_tables {
+            {"cf", "CREATE TABLE ks.cf (\n"
+            "    pk blob,\n"
```
nyh: nitpick: the indentation of this "cf" is different from the following ones.

test/boost/cql_query_test.cc
```diff
@@ -4313,3 +4315,198 @@ SEASTAR_TEST_CASE(test_rf_expand) {
...
+                    "    AND default_time_to_live = 0\n"
+                    "    AND gc_grace_seconds = 864000\n"
+                    "    AND max_index_interval = 2048\n"
+                    "    AND memtable_flush_period_in_ms = 0\n"
+                    "    AND min_index_interval = 128\n"
+                    "    AND read_repair_chance = 0\n"
+                    "    AND speculative_retry = '99.0PERCENTILE';\n"
+            },
+            {"CF2", "CREATE TABLE ks.\"CF2\" (\n"
+                    "    pk blob,\n"
```
nyh: It seems in none of these tests you tried a quoted partition-key column (only clustering key), and neither did you try quoted regular column names.

test/boost/cql_query_test.cc
```diff
@@ -4313,3 +4315,198 @@ SEASTAR_TEST_CASE(test_rf_expand) {
...
+                    "    AND dclocal_read_repair_chance = 0.2\n"
+                    "    AND default_time_to_live = 0\n"
+                    "    AND gc_grace_seconds = 954000\n"
+                    "    AND max_index_interval = 3048\n"
+                    "    AND memtable_flush_period_in_ms = 10\n"
+                    "    AND min_index_interval = 128\n"
+                    "    AND read_repair_chance = 0\n"
+                    "    AND speculative_retry = '99.0PERCENTILE';\n"
+            },
+            {"cf4", "CREATE TABLE ks.cf4 (\n"
```
nyh: what does cf4 test which cf2 didn't?

nyh:
> we need secondary indexes to work if anyone is in doubt

I don't think that regular secondary indexes work with this patch, I was worried about the newer "local secondary index" feature and whether this code knows how to describe its schema. Maybe it does and I just don't understand how? But we should at least have a test case for this case.

schema.hh
amnonh: ok

database.hh
amnonh: ok, I'll take your view, though it's a valid schema, if anyone would need the schema description (i.e pretty print instead of the redirection operator) it would be useful.

schema.cc
amnonh: UDT and UDF are defined at the keyspace level. We assume that the keyspace is defined correctly before creating the table.

you are right that I should test it as well

test/boost/cql_query_test.cc
amnonh: ok

test/boost/cql_query_test.cc
amnonh: no clustering key

nyh: Oh, I see you have a column called "CK" but it's no longer a clustering key :-) Confusing, but right. Ok.

database.hh
nyh: So maybe write_schema_as_cql() or something like that?
To emphasize that it writes CQL, it's not the normal function we use to persist a schema to disk or something.

amnonh: I like that better

amnonh:
@nyh I have forced push a new version, I hope everything is covered.
Besides the changes, there is also a test for a local index, a combination of compound key and a test with User define type.

nyh:
Thanks. I'll review - and hopefully push - now.

test/boost/cql_query_test.cc
```diff
@@ -4313,3 +4315,207 @@ SEASTAR_TEST_CASE(test_rf_expand) {
...
+              "    AND crc_check_chance = 1\n"
+              "    AND dclocal_read_repair_chance = 0.1\n"
+              "    AND default_time_to_live = 0\n"
+              "    AND gc_grace_seconds = 864000\n"
+              "    AND max_index_interval = 2048\n"
+              "    AND memtable_flush_period_in_ms = 0\n"
+              "    AND min_index_interval = 128\n"
+              "    AND read_repair_chance = 0\n"
+              "    AND speculative_retry = '99.0PERCENTILE';\n"},
+          {"cf_index_index", "CREATE INDEX cf_index ON ks.cf(col2);"},
```
nyh: Would also be good to check three more indexes: ks.cf(pk), ks.cf(pk1) and ks.cf(ck) - because it's possible to index also a key column (I think your code is correct in this case - I don't suspect a bug - but it wouldn't hurt to check, and it should be very easy with your code that can easily add more tests in this looo).

test/boost/cql_query_test.cc
```diff
@@ -4313,3 +4315,207 @@ SEASTAR_TEST_CASE(test_rf_expand) {
...
+                "    AND gc_grace_seconds = 954000\n"
+                "    AND max_index_interval = 3048\n"
+                "    AND memtable_flush_period_in_ms = 10\n"
+                "    AND min_index_interval = 128\n"
+                "    AND read_repair_chance = 0\n"
+                "    AND speculative_retry = '99.0PERCENTILE';\n"
+            }
+
+        };
+        e.execute_cql("CREATE TYPE phone ("
```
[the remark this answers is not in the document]
amnonh: From what I've tested, if the type is quoted-capitalized the type will be created, but the create table will fail because it will still look for the lower case name.

amnonh:
Thanks @nyh that was my last forced-push for 2019, let 2020 brings joy to all!

schema.cc
nyh: I keep asking and don't understand the response - is it fine that there is no maybe_quote on the cql3_type_name?

test/boost/cql_query_test.cc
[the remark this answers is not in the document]
nyh: So, the code is wrong, right (I left you a comment exactly where I think the code is wrong)? So why not fix it? And why not add the test here to confirm the code has been fixed?

amnonh:
On Tuesday, December 31, 2019, nyh <notifications@github.com> wrote:

> *@nyh* requested changes on this pull request.
>
> In schema.cc
> <https://github.com/scylladb/scylla/pull/5294#discussion_r362255690>:
>
> > +std::ostream& map_as_cql_param(std::ostream& os, const std::map<sstring, sstring>& map, bool first = true) {
> +    for (auto i: map) {
> +        if (first) {
> +            first = false;
> +        } else {
> +            os << ",";
> +        }
> +        os << "'" << i.first << "': '" << i.second << "'";
> +    }
> +
> +    return os;
> +}
> +
> +std::ostream& column_definition_as_cql_key(std::ostream& os, const column_definition & cd) {
> +    os << cd.name_as_cql_string();
> +    os << " " << cd.type->cql3_type_name();
>
> I keep asking and don't understand the response - is it fine that there is
> no maybe_quote on the cql3_type_name?
>
I missed that you were asking about it, it's a wrapper that call maybe
quote internally.

------------------------------
>
> In test/boost/cql_query_test.cc
> <https://github.com/scylladb/scylla/pull/5294#discussion_r362255954>:
>
> > +                "    AND comment = ''\n"
> +                "    AND compaction = {'class': 'SizeTieredCompactionStrategy'}\n"
> +                "    AND compression = {'sstable_compression': 'org.apache.cassandra.io.compress.LZ4Compressor'}\n"
> +                "    AND crc_check_chance = 1\n"
> +                "    AND dclocal_read_repair_chance = 0.2\n"
> +                "    AND default_time_to_live = 0\n"
> +                "    AND gc_grace_seconds = 954000\n"
> +                "    AND max_index_interval = 3048\n"
> +                "    AND memtable_flush_period_in_ms = 10\n"
> +                "    AND min_index_interval = 128\n"
> +                "    AND read_repair_chance = 0\n"
> +                "    AND speculative_retry = '99.0PERCENTILE';\n"
> +            }
> +
> +        };
> +        e.execute_cql("CREATE TYPE phone ("
>
> So, the code is wrong, right (I left you a comment exactly where I think
> the code is wrong)? So why not fix it? And why not add the test here to
> confirm the code has been fixed?
>
Are you talking about the user define type?
If so, I responded on the previous patch, I've tested with quoted name and
it does not supported in a kind of weird way.

If it's something else, then Im not sure what you are referring to.

nyh:
On Wed, Jan 1, 2020 at 12:47 AM Amnon Heiman <notifications@github.com>
wrote:

> > + os << " " << cd.type->cql3_type_name();
> >
> > I keep asking and don't understand the response - is it fine that there
> is
> > no maybe_quote on the cql3_type_name?
> >
> I missed that you were asking about it, it's a wrapper that call maybe
> quote internally.
>

Excellent, I didn't find where this is implemented so I was worried it is
missing a maybe_quote.
Especially since you said that types with quoted names didn't work
correctly in your test.


> > + e.execute_cql("CREATE TYPE phone ("
> >
> > So, the code is wrong, right (I left you a comment exactly where I think
> > the code is wrong)? So why not fix it? And why not add the test here to
> > confirm the code has been fixed?
> >
> Are you talking about the user define type?
>

Yes, exactly. You can have exactly the same test with the type name "pHoNe"
and
it should work too (unless I'm misunderstanding something in how type names
work
in Cassandra or Scylla).

If so, I responded on the previous patch, I've tested with quoted name and
> it does not supported in a kind of weird way.
>

"not supported in a weird way", means it doesn't work? Is it *supposed* to
work (works in Cassandra)? If so, why not fix it?

If it's a completely unrelated bug in Scylla with support for quoted type
names, please open a separate issue about this.
But is it completely unrelated and just a bug in describe()? I don't
understand how. In any case, if it's easy to fix, better just fix it - and
add a test for this case.

amnonh:
@nyh My first forced-push to 2020 added the additional index tests

nyh:
Committed to next. Mazal tov ;-)

avikivity:
Dequeueing, fails cql_query_test:

```
ecbabe2f42ae269ed8b8f5dc860cf01081f8b7de is the first bad commit
commit ecbabe2f42ae269ed8b8f5dc860cf01081f8b7de
Author: Amnon Heiman <amnon@scylladb.com>
Date:   Mon Nov 18 09:03:22 2019 +0200

    database: add schema.cql file when creating a snapshot
    
    When creating a snapshot we need to add a schema.cql file in the
    snapshot directory that describes the table in that snapshot.
    
    This patch adds the file using the schema describe method.
    
    get_snapshot_details was modified to ignore the schema.cql file.
    
    Fixes #4192
    
 database.hh | 18 ++++++++++++++++++
 table.cc    | 30 +++++++++++++++++++++++++-----
 2 files changed, 43 insertions(+), 5 deletions(-)
```

amnonh:
@nyh 
I have forced push a version that prints an error when writing a schema but does not stop the snapshot.
cql_query_test now passes

nyh:
@amnonh it would have been nice to put here (for lack of a better place... I'm not a fan of these pull requests...) a description of what you fixed in this version (although I know :-)). I'll review and commit.

nyh:
Hmm, I now realized that "prints an error when writing a schema but does not stop the snapshot" was supposed to be a description of what you think... But I'm not sure how anyone except the two of us will know what this means and which problem in the test it fixed.

nyh:
@amnonh unfortunately your branch does not apply cleanly to "next". Please rebase.

slivne:
@amnonh ping

amnonh:
I have forced push a newer version this morning, but I see that there is already another conflict

nyh:
It seems that today, everybody is sending patches to cql_query_test.cc, and all of these patches want to add a new test at the end, so all of them conflict :-( Whoever rebases their patches more quickly, has a better chance to be committed without yet another rebase - so - the race is on! :-)

amnonh:
2AM is as good a time to force push pull-rebase

nyh:
Committed to next. Hopefully 10th time is the charm :-)

amnonh:
Praying for the old gods and the new

avikivity:
Dequeued due to dtest failures.

amnonh:
I've forced push a new version, but there is also a change in dtest that need to be merged first:
https://github.com/scylladb/scylla-dtest/pull/1247

avikivity:
What are the changes in the new version?

nyh:
@amnonh I see your dtest patch was already merged into scylla-dtest next. @fruch what is the process of how it reaches scylla-dtest master?

But you said you forced-pushed a new version of the Scylla patch as well. Why? What changed? (@avikivity asked this as well).

fruch:
The dtest would be in master once next is passing

amnonh:
I've updated the message following @avikivity comment to state that table::manifest_json_filter now filter the schema.cql file.

I see that there is now a conflict in the cql_query_test as well, I'll rebase

nyh:
Yes, I saw your changes to dtest, but you also said that you forced-push a version of Scylla, here, so I was wondering why. Was it just a rebase? Any other change?

amnonh:
@nyh there is an additional change:
table::manifest_json_filter now filter the schema.cql file.

nyh:
Ok, I don't know what to do now but commit the new version and pray you fixed the tests.
I'll do that - minus the praying ;-)

nyh:
Well, I can't commit it yet, actually, because your dtest patch **still** hasn't reached master (https://github.com/scylladb/scylla-dtest/commits/master), so if I commit this patch, it will fail the dtest again. This is so frustrating :-(  If I forget to commit your series tomorrow, please remind me again.

amnonh:
@nyh I forced-push a rebase of the series

nyh:
Pushed to next. Please consider praying to a more powerful god this time :-)

Issue #4263, opened 2019-02-24
Slight different between Scylla and Apache Cassandra TTL


tzach:
*Installation details*
Scylla version (or git commit hash): 3.0.3
Apache Cassandra 3.11

Just before a row is TTLed, Scylla return a null value, while Cassandra does not

### Setup
```
cqlsh> CREATE KEYSPACE mykeyspace WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;
cqlsh> USE mykeyspace;
cqlsh:mykeyspace> CREATE TABLE heartrate_ttl (
              ...  pet_chip_id  uuid,
              ...  time timestamp,
              ...  heart_rate int,
              ... PRIMARY KEY (pet_chip_id, time));
cqlsh:mykeyspace>  INSERT INTO heartrate_ttl(pet_chip_id, time, heart_rate) VALUES (123e4567-e89b-12d3-a456-426655440b23, '2019-03-04 07:04:00', 131) USING TTL 21;
cqlsh:mykeyspace> select pet_chip_id,time, heart_rate, TTL(heart_rate) from heartrate_ttl;

pet_chip_id                          | time                            | heart_rate | ttl(heart_rate)
--------------------------------------+---------------------------------+------------+-----------------
 123e4567-e89b-12d3-a456-426655440b23 | 2019-03-04 07:04:00.000000+0000 |        131 |              10
```

Just before TTL is over Scylla will return
```
cqlsh:mykeyspace> select pet_chip_id,time, heart_rate, TTL(heart_rate) from heartrate_ttl;

 pet_chip_id                          | time                            | heart_rate | ttl(heart_rate)
--------------------------------------+---------------------------------+------------+-----------------
 123e4567-e89b-12d3-a456-426655440b23 | 2019-03-04 07:04:00.000000+0000 |       null |            null
```

and after a second
```
pet_chip_id | time | heart_rate | ttl(heart_rate)
-------------+------+------------+-----------------
```

I cound not repreduce this null value with Cassandra. Not sure its a bug

avikivity:
Maybe the row marker's liveness and the column's liveness are evaluted using different base timestamps (at two different points in the query lifecycle).

haaawk:
@eliransin Can we reassign to @piodul since his working on #5290 which seems related?

eliransin:
@haaawk NP. Let me know if it turns out to be unrelated :)

avikivity:
[...]

Issue #5290, opened 2019-11-16
Data expired via TTL is returned with null values for a short period of time


eyusim:
*Installation details*
Scylla version (or git commit hash): 2.1.6 / 3.1.1
OS (RHEL/CentOS/Ubuntu/AWS AMI): CentOS

We have noticed after upgrading from 2.0 -> 3.1.1 (starting with 2.1.6) that scylla will return rows that have been recently expired due to TTL with null values in non-key columns. 

This issue happens in both 2.1.6 and 3.1.1. 

Below is python code to reproduce the issue (just change the port in the Cluster constructor). 

The code inserts 2 rows with same partition key, one row has a 5 second ttl and the other 1 hour. Continuous querying of the data will show that about 5 times the query returns 2 rows with the expired rows data having been wiped out but it's still being returned. 

Running the code against a docker container for the latest cassandra or scylla 2.0.2 behaves as you would expect, with the result set eventually being 1 row and no missing data in-between. 

```python
from cassandra.cluster import Cluster
import time

cluster = Cluster(['localhost'], port=9043)

session = cluster.connect()
session.execute('drop keyspace if exists ttltest ')

session.execute("""create keyspace if not exists ttltest WITH replication = { 'class': 'SimpleStrategy', 'replication_factor': '1' }""")
session.set_keyspace('ttltest')

session.execute("""
create table if not exists data_with_ttl(
            thekey text,
            num bigint,
            data text,
            PRIMARY KEY (thekey, num)
          ) 
""")

session.execute("insert into data_with_ttl (thekey, num, data) values('key1', 123, 'long live data') using TTL 3600")
session.execute("insert into data_with_ttl (thekey, num, data) values('key1', 456, 'quick expire data') using TTL 5")

tries = 0
while tries < 100:
    tries = tries + 1
    rows = session.execute("select * from data_with_ttl where thekey = 'key1'")
    resultSize = len(rows.current_rows)
    print "try number:", tries, "result size:", resultSize
    for row in rows:
        print '\t',  row.num, row.data
    if resultSize == 1:
        break
    time.sleep(0.2)

session.execute('drop keyspace ttltest')

```

tzach:
Look like https://github.com/scylladb/scylla/issues/4263

slivne:
@haaawk please assign to have someone look at this

avikivity:
Backported to 3.0, 3.1, 3.2.

avikivity:
Already backported.

PR #5301, opened 2019-11-18
row_marker: correct row expiry condition

piodul:
This change corrects condition on which a row was considered expired by its TTL.

The logic that decides when a row becomes expired was inconsistent with the logic that decides if a single cell is expired. A single cell becomes expired when `expiry_timestamp <= now`, while a row became expired when `expiry_timestamp < now` (notice the strict inequality). For rows inserted with TTL, this caused non-key cells to expire (change their values to null) one second before the row disappeared. Now, row expiry logic uses non-strict inequality.

Fixes: #4263, #5290.

Tests:
- unit(dev)
- python test described in issue #5290

### DIFF
--- a/mutation_partition.cc
+++ b/mutation_partition.cc
@@ -1876,7 +1876,7 @@ bool row_marker::compact_and_expire(tombstone tomb, gc_clock::time_point now,
         _timestamp = api::missing_timestamp;
         return false;
     }
-    if (_ttl > no_ttl && _expiry < now) {
+    if (_ttl > no_ttl && _expiry <= now) {
         _expiry -= _ttl;
         _ttl = dead;
     }
--- a/tests/mutation_test.cc
+++ b/tests/mutation_test.cc
@@ -1320,6 +1320,104 @@ SEASTAR_THREAD_TEST_CASE(test_mutation_upgrade_type_change) {
     assert_that(m).is_equal_to(m2);
 }
 
+// This test checks the behavior of row_marker::{is_live, is_dead, compact_and_expire}. Those functions have some
+// duplicated logic that decides if a row is expired, and this test verifies that they behave the same with respect
+// to TTL.
+SEASTAR_THREAD_TEST_CASE(test_row_marker_expiry) {
+    can_gc_fn never_gc = [] (tombstone) { return false; };
+
+    auto must_be_alive = [&] (row_marker mark, gc_clock::time_point t) {
+        BOOST_TEST_MESSAGE(format("must_be_alive({}, {})", mark, t));
+        BOOST_REQUIRE(mark.is_live(tombstone(), t));
+        BOOST_REQUIRE(mark.is_missing() || !mark.is_dead(t));
+        BOOST_REQUIRE(mark.compact_and_expire(tombstone(), t, never_gc, gc_clock::time_point()));
+    };
+
+    auto must_be_dead = [&] (row_marker mark, gc_clock::time_point t) {
+        BOOST_TEST_MESSAGE(format("must_be_dead({}, {})", mark, t));
+        BOOST_REQUIRE(!mark.is_live(tombstone(), t));
+        BOOST_REQUIRE(mark.is_missing() || mark.is_dead(t));
+        BOOST_REQUIRE(!mark.compact_and_expire(tombstone(), t, never_gc, gc_clock::time_point()));
+    };
+
+    const auto timestamp = api::timestamp_type(1);
+    const auto t0 = gc_clock::now();
+    const auto t1 = t0 + 1s;
+    const auto t2 = t0 + 2s;
+    const auto t3 = t0 + 3s;
+
+    // Without timestamp the marker is missing (doesn't exist)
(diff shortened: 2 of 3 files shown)


tests/mutation_test.cc
```diff
@@ -1320,6 +1320,99 @@ SEASTAR_THREAD_TEST_CASE(test_mutation_upgrade_type_change) {
...
+
+    // Without timestamp it is never alive
+    row_marker m1;
+    must_be_dead(m1, t0);
+    must_be_dead(m1, t1);
+    must_be_dead(m1, t2);
+    must_be_dead(m1, t3);
+
+    // Without ttl it is always alive
+    row_marker m2(api::new_timestamp());
```
nyh: is api::new_timestamp() supposed to be the same as t1?
The reason I'm asking is that not a good property of the test that it slightly depends on how much time passed between t1 and right now. Could we just use t1 here?

tests/mutation_test.cc
```diff
@@ -1320,6 +1320,99 @@ SEASTAR_THREAD_TEST_CASE(test_mutation_upgrade_type_change) {
     assert_that(m).is_equal_to(m2);
 }
 
+SEASTAR_THREAD_TEST_CASE(test_row_marker) {
+    auto ttl = std::chrono::seconds(1);
```
nyh: nitpick: in C++, there is no need to define variables at the top of the function. It's clearer if you define them right before actually using them. By the time you used "ttl" below, I forgot what it was :-)
By the way, you can use "const" to emphasize you only set it once, and don't plan to change it below.

tests/mutation_test.cc
```diff
@@ -1320,6 +1320,99 @@ SEASTAR_THREAD_TEST_CASE(test_mutation_upgrade_type_change) {
     assert_that(m).is_equal_to(m2);
 }
 
+SEASTAR_THREAD_TEST_CASE(test_row_marker) {
```
nyh: I would really appreciate a comment above the test explaining what it's testing, and why. If it served you to reproduce an issue with a number, you can say it reproduces issue #1234, for example. Or if this test is just a general thing, and just one specific line (e.g., the test for t2 - the equality case) reproduces issue #1234 you can say it there.

mutation_partition.cc
```diff
@@ -1876,7 +1876,7 @@ bool row_marker::compact_and_expire(tombstone tomb, gc_clock::time_point now,
         _timestamp = api::missing_timestamp;
```
denesb: (commenting here because I can't comment on the description)

A good commit description also explains why the logic was invalid and how it is fixed.

tests/mutation_test.cc
denesb: After `using std::chrono_literals` you can just write `1s`.

tests/mutation_test.cc
```diff
@@ -1320,6 +1320,99 @@ SEASTAR_THREAD_TEST_CASE(test_mutation_upgrade_type_change) {
...
+    };
+
+    auto must_be_dead = [&] (row_marker mark, gc_clock::time_point t) {
+        BOOST_TEST_MESSAGE(format("must_be_dead({}, {})", mark, t));
+        BOOST_REQUIRE(!mark.is_live(tombstone(), t));
+        BOOST_REQUIRE(mark.is_missing() || mark.is_dead(t));
+        BOOST_REQUIRE(!mark.compact_and_expire(tombstone(), t, never_gc, gc_clock::time_point()));
+    };
+
+    // Without timestamp it is never alive
```
denesb: Without timestamp the marker is missing (doesn't exist).

tests/mutation_test.cc
piodul: `t1` cannot be used here, `gc_clock::time_point` is a different type from `api::timestamp_type`.

Actually, the logic that this test verifies does not take the timestamp into consideration. I changed values used in this test into one constant. I initially used `api::new_timestamp` because it was also used in other tests.

tests/mutation_test.cc
piodul: Added `const` qualifiers and increased locality of the code.

tests/mutation_test.cc
piodul: OK, I added some description above the test.

mutation_partition.cc
piodul: I reworded the description, hope that it is more complete and clear now.

tests/mutation_test.cc
piodul: Done.

tests/mutation_test.cc
piodul: Changed.

mutation_partition.cc
denesb: Much better. Previously I had to guess and go look at the other `is_expired()` implementations to see why this one was invalid.

tests/mutation_test.cc
avikivity: You can also write `auto ttl = 1s;` if you want.

PR #5310, opened 2019-11-19
toolchain: update to Fedora 31

avikivity:
This is a minor update as gcc and boost versions did not change. A noteable update is patchelf 0.10, which adds support to large binaries.

A few minor issues exposed by the update are fixed in preparatory patches.


espindola:
Tested on fedora 30. All tests pass on dev mode, so LGTM.
Not sure if I should approve since I am not familiar with rpm building.
@glommer ?

Issue #5139, opened 2019-10-03
max and min on a inet column returns unexpected result


fgelcer:
This is Scylla's bug tracker, to be used for reporting bugs only.
If you have a question about Scylla, and not a bug, please ask it in
our mailing-list at scylladb-dev@googlegroups.com or in our slack channel.

- [] I have read the disclaimer above, and I am reporting a suspected malfunction in Scylla.

*Installation details*
Scylla version (or git commit hash): ReleaseVersion: 3.0.8
Cluster size: 
OS (RHEL/CentOS/Ubuntu/AWS AMI):

*Hardware details (for performance issues)*          Delete if unneeded
Platform (physical/VM/cloud instance type/docker):
Hardware: sockets= cores= hyperthreading= memory=
Disks: (SSD/HDD, count)

c* returns the value, as it is represented as a string:
```
cqlsh> CREATE KEYSPACE group_functions WITH replication={'class':'SimpleStrategy', 'replication_factor':1} ;
cqlsh> USE group_functions ;
cqlsh:group_functions> create TABLE all_types (id int PRIMARY KEY, my_inet inet);
cqlsh:group_functions> SELECT * from all_types ;

 id | my_inet
----+---------

(0 rows)
cqlsh:group_functions> INSERT INTO all_types (id, my_inet) VALUES (0, '1.1.1.1');
cqlsh:group_functions> INSERT INTO all_types (id, my_inet) VALUES (1, '1.1.1.2');
cqlsh:group_functions> INSERT INTO all_types (id, my_inet) VALUES (2, '1.1.1.3');
cqlsh:group_functions> INSERT INTO all_types (id, my_inet) VALUES (3, '1.1.1.4');
cqlsh:group_functions> INSERT INTO all_types (id, my_inet) VALUES (4, '1.1.1.5');
cqlsh:group_functions> INSERT INTO all_types (id, my_inet) VALUES (6, '1.1.1.6');
cqlsh:group_functions> SELECT cou
                       <blobLiteral>          <colname>              <float>                <identifier>           <pgStringLiteral>      <quotedStringLiteral>  <uuid>                 <wholenumber>          COUNT                  
cqlsh:group_functions> SELECT count(my_inet) from all_types ;

 system.count(my_inet)
-----------------------
                     6

(1 rows)

Warnings :
Aggregation query used without partition key

cqlsh:group_functions> SELECT min(my_inet) from all_types ;

 system.min(my_inet)
---------------------
             1.1.1.1

[...]

PR #5311, opened 2019-11-20
5139 minmax bad printing

jul-stas:
This is a partial solution to #5139 (only for two types) because of the above and because collections are much harder to do. They are coming in a separate PR.
What this PR does is simple: adds `system.min()` and `system.max()` fns for `time_type` and `inet_address`, so 'overload resolution' does not use `blob->blob` overload, which caused results to be printed as blobs.

### DIFF
--- a/cql3/functions/aggregate_fcts.hh
+++ b/cql3/functions/aggregate_fcts.hh
@@ -247,6 +247,25 @@ struct aggregate_type_for<timeuuid_native_type> {
     using type = timeuuid_native_type::primary_type;
 };
 
+template<>
+struct aggregate_type_for<time_native_type> {
+    using type = time_native_type::primary_type;
+};
+
+template <typename Type>
+const Type& max_wrapper(const Type& t1, const Type& t2) {
+    using std::max;
+    return max(t1, t2);
+}
+
+inline const net::inet_address& max_wrapper(const net::inet_address& t1, const net::inet_address& t2) {
+    using family = seastar::net::inet_address::family;
+    const size_t len =
+            (t1.in_family() == family::INET || t2.in_family() == family::INET)
+            ? sizeof(::in_addr) : sizeof(::in6_addr);
+    return std::memcmp(t1.data(), t2.data(), len) >= 0 ? t1 : t2;
+}
+
 template <typename Type>
 class impl_max_function_for final : public aggregate_function::aggregate {
    std::optional<typename aggregate_type_for<Type>::type> _max{};
@@ -268,7 +287,7 @@ public:
         if (!_max) {
             _max = val;
         } else {
-            _max = std::max(*_max, val);
+            _max = max_wrapper(*_max, val);
         }
     }
 };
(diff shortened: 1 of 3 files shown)


denesb:
On Thu, 2019-11-21 at 07:36 -0800, nyh wrote:
> @nyh commented on this pull request.
> 
> 
> Looks good to me, but:
> 
> 
> It needs the Seastar patch first, and it isn't in yet.
> If the Seastar patch changes the sort order, this one needs to change
> as well.
> 
> Note that 2 is slightly worrying: I means that if Scylla needs a
> specific ordering (perhaps to be 100% compatible with what Cassandra
> does in this case?) we need Seastar's ordering to be the same. Could
> we somehow define Scylla's require ordering as a seperate comparator
> object, instead of operator<?

How would that guarantee compatible ordering? If it relies on the
seastar operator< it could change without notice just the same. We rely
on other libs (most prominently libstdc++) for other orderings as well.
The only guarantee is open coding our own comparators, but even better
is writing tests that verify the expected ordering.

avikivity:
Most types have natual orders and we don't expect a library to just change it.

Where there is no natural order, or if Cassandra's ordering is not the same as the natural order, then we must open-code our comparator to insulate ourselves from the problem.

jul-stas:
I was considering to define this comparison in scylla, but:
- having `operator<` for `inet_address`es solves this problem right away, with minimal code, without messing with implementation of aggregates. (simplicity, safety)
- Due to ADL, the operator should be in `namespace seastar` and I don't know any trick to work around this (anyone?).

Anyway, supplying comparator object to aggregare fns wouldn't be a big deal. The question is if it's better..?

denesb:
I think that for those types that don't have a natural ordering and yet we want to guarantee a certain one (compatible with C*), we can just add specific unit tests. The tests will make sure the ordering will not change unnoticed.

jul-stas:
@denesb  @avikivity @nyh Let's make a conclusion or this will hang forever.
Custom comparator in scylla and no `inet_address::operator<`?

nyh:
@jul-stas I think it is easiest and safest to include this operator< or comparator in Scylla, not in Seastar. The problem isn't just making sure it doesn't change in Seastar; As you said, this would be easy to ensure by using unit tests. The bigger problem is getting the Seastar committers to agree to commit a suspicious-looking operator< that has no real justification besides "this is what Cassandra does". As you can see, your patch to do this was not committed yet into Seastar by any of its committers.

By the way, following up on your question: using "namespace seastar" in Scylla would be "cheating" but an acceptable form of cheating, I think. We already have a bunch of these in Scylla, and even "namespace std"! An example of the latter is how we define a has function in bytes.hh.

jul-stas:
@denesb @nyh Updated - please have a look. Logic of ordering (or lack thereof) is still the same, but now defined in scylla.

cql3/functions/aggregate_fcts.hh
```diff
@@ -247,6 +247,25 @@ struct aggregate_type_for<timeuuid_native_type> {
...
 
+template<>
+struct aggregate_type_for<time_native_type> {
+    using type = time_native_type::primary_type;
+};
+
+template <typename Type>
+const Type& max_wrapper(const Type& t1, const Type& t2) {
+    using std::max;
+    return max(t1, t2);
```
nyh: nitpick: why do with two lines what you can do with one, i.e., call "std::max()" directly?

jul-stas: Thanks! I use this trick sometimes to allow custom implementations, e.g. `max(const WeirdType& wt1, const WeirdType& wt2)` in the namespace where `WeirdType` belongs. Then ADL will prefer the custom over `std::max()`.
But you're right, at this moment it does the same thing, it's only for future-proofness.

PR #5316, opened 2019-11-20
api: export system uptime via REST

glommer:
This will be useful for tools like nodetool that want to query the uptime
of the system.


avikivity:
Merged as 73aff1fc95b44ec26acb25a6a417ca1dcabb2bc5, don't know why github didn't pick it up.

PR #5329, opened 2019-11-21
Sstables: limit snapshot dirs

### LINKED ISSUE #5324: snapshot only sstables in base directory
We've hit a case where snapshot failed since an sstable in staging got removed while taking its snapshot. We need to snapshot only sstables located in the base table directory, not in "staging" or "uploads".

bhalevy:
Fixes #5324

We've hit a case where snapshot failed since an sstable in staging got removed while taking its snapshot. We need to snapshot only sstables located in the base table directory, not in "staging" or "uploads".

The patchset introduces an enum class sstable::location_types that indicates where the sstable is in the data hierarchy: base, staging, upload, or snapshots.

The sstable location is determined when the sstable object is constructed using the sstable directory using a regular expression shared with `make_descriptor`.  `sstable::get_location()` retrieves the location.

The sstable location is used by `table::snapshot` to filter `_sstables->all()` and ignore any sstable that is not in the base directory.

A unit test for sstable location has been added to `sstable_test`.

### DIFF
--- a/sstables/entry_descriptor.hh
+++ b/sstables/entry_descriptor.hh
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2019 ScyllaDB
+ */
+
+/*
+ * This file is part of Scylla.
+ *
+ * Scylla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Scylla is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <seastar/core/sstring.hh>
+
+#include "sstables/component_type.hh"
+#include "sstables/version.hh"
+
+using namespace seastar;
+
+namespace sstables {
+
+struct entry_descriptor {
+    sstring sstdir;
+    sstring ks;
+    sstring cf;
+    int64_t generation;
+    sstable_version_types version;
+    sstable_format_types format;
+    component_type component;
+
+    static entry_descriptor make_descriptor(sstring sstdir, sstring fname);
+
+    entry_descriptor(sstring sstdir, sstring ks, sstring cf, int64_t generation,
+                     sstable_version_types version, sstable_format_types format,
+                     component_type component)
+        : sstdir(sstdir), ks(ks), cf(cf), generation(generation), version(version), format(format), component(component) {}
+};
+
+}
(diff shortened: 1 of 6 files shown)


sstables/entry_descriptor.hh
```diff
@@ -21,6 +21,8 @@
 
 #pragma once
```
avikivity: What's the motivation? You're declaring static file/dir types, but I can't guess why.

table.cc
```diff
@@ -1720,7 +1720,10 @@ seal_snapshot(sstring jsondir) {
 future<> table::snapshot(sstring name) {
```
avikivity: Fixes ????
Explanation?

avikivity:
looks okay, needs documentary improvement

nyh:
@bhalevy I'm trying to understand your statement that "we need to snapshot only sstables located in the base table directory, not in "staging" or "uploads"."
Please correct me if I'm wrong (it's been a while since I looked at that stuff), but I think that sstables in staging/ are already readable by a client, aren't they? In this case, why should they not be snapshotted as well?

sstables/entry_descriptor.hh
bhalevy: We're practically using `pragma once` as an include guard.
Isn't that by convention?

table.cc
bhalevy: The motivation is described in the pull request message.
I can add a comment here as well.

bhalevy:
> @bhalevy I'm trying to understand your statement that "we need to snapshot only sstables located in the base table directory, not in "staging" or "uploads"."
> Please correct me if I'm wrong (it's been a while since I looked at that stuff), but I think that sstables in staging/ are already readable by a client, aren't they? In this case, why should they not be snapshotted as well?

I was under the impression from @slivne's comment that we should not take snapshots in staging directory, but after discussion and re-reading the code I think you're right and these are readable and need to be part of the snapshot.  That said, [apparently](https://github.com/scylladb/scylla-enterprise/issues/1146) we need to do that carefully.

bhalevy:
Retracting this PR in favor of proper synchronization of snapshot and view_update_generator.

sstables/sstables.cc
```diff
@@ -2697,7 +2720,7 @@ void sstable::move_to_new_dir_in_thread(sstring new_dir, int64_t new_generation)
 
 std::regex entry_descriptor::la_mc_filename_pattern("(la|mc)-(\\d+)-(\\w+)-(.*)");
 std::regex entry_descriptor::ka_filename_pattern("(\\w+)-(\\w+)-ka-(\\d+)-(.*)");
-std::regex entry_descriptor::sst_dir_pattern(".*/([^/]*)/([^/]+)-[\\da-fA-F]+(?:/staging|/upload|/snapshots/[^/]+)?/?");
+std::regex entry_descriptor::sst_dir_pattern(".*/([^/]*)/([^/]+)-[\\da-fA-F]+(/(staging|upload|snapshots)(/[^/]+)?)?/*");
```
denesb: I don't see the relaxing, the `staging|upload|snapshot` part is now mandatory. Furthermore there is now an unmatched `(`. Also, this pattern now permits any amount of trailing `/`.

sstables/types.hh
```diff
@@ -49,6 +49,14 @@ static inline bytes_view to_bytes_view(const temporary_buffer<char>& b) {
 
 namespace sstables {
 
+enum class sstable_location_types {
```
denesb: Why the plural?

PR #5330, opened 2019-11-21
support new aws instance types: c5d, m5d, m5ad, r5d, z1d

amoskong:
Currently scylla_io_setup will skip in scylla_setup, because we didn't support
those new instance types.

I manually executed scylla_io_setup, and the scylla-server started and worked
well.

Let's apply this patch first, then check if there is some new problem in
ami-test.

Fixes: https://github.com/scylladb/scylla/issues/5312

@syuu1228 @avikivity

### DIFF
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -167,7 +167,7 @@ class aws_instance:
         return self._type.split(".")[0]
 
     def is_supported_instance_class(self):
-        if self.instance_class() in ['i2', 'i3', 'i3en']:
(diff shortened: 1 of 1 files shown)


avikivity:
@amoskong there are 9 commits here

amoskong:
> @amoskong there are 9 commits here

Rebased to latest next.

avikivity:
@amoskong in the future, base your patches on master to avoid such problems.

amoskong:
> @amoskong in the future, base your patches on master to avoid such problems.

Ok

PR #5346, opened 2019-11-25
cdc: allow cluster rolling upgrade 

### LINKED ISSUE #5288: upgrade from 3.1 to pre-3.2: Found db err in verify test after upgraded one db node
*Installation details*
Scylla version (or git commit hash): Upgrade from `3.1.1-0.20191024.f32ec885c` to `666.development-0.20191114.552c56633e0`
Cluster size: 4
OS (RHEL/CentOS/Ubuntu/AWS AMI): GCE centos7
### Living GCE cluster:
- https://console.cloud.google.com/compute/instances?project=skilled-adapter-452&instancessize=50&instancesquery=%255B%257B_22k_22_3A_22_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22050cdd46_5C_22_22%257D%255D
- Node rolling-upgrade-upgrade--centos-db-node-050cdd46-0-1 [104.196.53.201 | 10.142.0.39]
- Node rolling-upgrade-upgrade--centos-db-node-050cdd46-0-2 [34.73.170.11 | 10.142.0.52]
- Node rolling-upgrade-upgrade--centos-db-node-050cdd46-0-3 [34.74.3.45 | 10.142.0.122]
- Node rolling-upgrade-upgrade--centos-db-node-050cdd46-0-4 [34.74.46.139 | 10.142.0.123]
- Node rolling-upgrade-upgrade--centos-loader-node-050cdd46-0-1 [35.237.158.147 | 10.142.0.124]
- Node rolling-upgrade-upgrade--centos-monitor-node-050cdd46-0-1 [34.73.132.20 | 10.142.0.125]

### Logs:
- db logs: https://cloudius-jenkins-test.s3.amazonaws.com/050cdd46-736c-4a04-b5f7-ce74fc8d403e/rolling-upgrade-upgrade--centos-db-cluster-050cdd46.zip 
- sct event log: https://cloudius-jenkins-test.s3.amazonaws.com/050cdd46-736c-4a04-b5f7-ce74fc8d403e/events_log.zip 

### Errors after upgraded one db node:

- migration_manager - Failed to update definitions 
from 10.142.0.39:2: exceptions::configuration_exception (Missing enabled CDC option)
- storage_proxy - Exception when communicating with
 10.142.0.122: std::runtime_error (Failed to load schema version b9d0630f-b374-386d-8f4d-bc409a5be4ab)
    - Note: currently we create all the tables together, and sleep 30 seconds after that, then execute truncate.
- auth_service - Unexpected exception while revoking all permissions on dropped table: exceptions::read_failure_exception (Operation failed for system_auth.role_permissions - received 0 responses and 1 failures from 1 CL=LOCAL_ONE.)
-  migration_manager - Failed to update definitions from 10.142.0.39:2: query::deserialization_error (failed on column system_schema.scylla_tables#cdc (version: d15e4b38-4294-3cf3-b7e2-5485bd5c9367, id: 0, index: 0, type: org.apache.cassandra.db.marshal.FrozenType(org.apache.cassandra.db.marshal.MapType(org.apache.cassandra.db.marshal.UTF8Type,org.apache.cassandra.db.marshal.UTF8Type))): seastar::internal::backtraced<marshal_exception> (marshaling error: read_simple_bytes - not enough bytes (requested 131207658, got 8) Backtrace:   0x2d67f2d#012  0xa7f75e#012  0x44b353#012  0x4e7094#012  0x13d38f2#012  0x1b16f6e#012  0x1b18e70#012  0x1b1b6b0#012  0x1b192a7#012  0x1b19844#012  0x1b1a662#012  0xb9469c#012  0x177c6e0#012  0x17805e5#012  0x1783cdb#012  0x17853d8#012  0x7e2143#012))
      - Occurred in re-fill the test data
- seastar - Exceptional future ignored: std::_Nested_exception<no_such_column_family> (Can't find a column family with UUID 083fffc0-07d2-11ea-8234-000000000002): std::out_of_range (_Map_base::at), backtrace: 
  0x2d67f2d#012  0x2ab82c4#012  0x961d51#012  0x2ab5a91#012  0x2ab5c9f#012  0x2b1c2fd#012  0x2a62af1#012  0x2a6370e#012  0x795c11#012  /opt/scylladb/libreloc/libc.so.6+0x23f32#012  0x7e456d
    - Occurred in re-fill the test data

### Test Scenario:
- execute verify test of https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/fill_db_data.py
- upgrade one db node (db-node-2)
- re-execute verify test of https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/fill_db_data.py

```
Upgrade from 3.1.1-0.20191024.f32ec885c to 666.development-0.20191114.552c56633e0

11-15 17:01:35,208 > Populate DB with many types of tables and data
11-15 17:05:00,354 > Run some Queries to verify data BEFORE UPGRADE
11-15 17:06:11,836 > Re-Populate DB with many types of tables and data

### Upgrade onde node (db-node-2)
< t:2019-11-15 17:21:58,511 f:upgrade_test.py l:449  c:sdcm.tester          p:INFO  > Upgrade Node rolling-upgrade-upgrade--centos-db-node-050cdd46-0-2 begin
< t:2019-11-15 17:21:58,512 f:upgrade_test.py l:106  c:sdcm.tester          p:INFO  > Upgrading a Node
< t:2019-11-15 17:28:49,435 f:upgrade_test.py l:451  c:sdcm.tester          p:INFO  > Upgrade Node rolling-upgrade-upgrade--centos-db-node-050cdd46-0-2 ended

### Run verify test of fill_db_data.py
11-15 17:40:14,480 > Run some Queries to verify data after upgraded one node

11-15T17:40:18+00:00  rolling-upgrade-upgrade--centos-db-node-050cdd46-0-2 !ERR    
 | scylla: [shard 2] migration_manager - Failed to update definitions 
10.142.0.39:2: exceptions::configuration_exception (Missing enabled CDC option)

11-15T17:47:36+00:00  rolling-upgrade-upgrade--centos-db-node-050cdd46-0-2 !ERR     
| scylla: [shard 0] storage_proxy - Exception when communicating with
42.0.122:std::runtime_error (Failed to load schema version b9d0630f-b374-386d-8f4d-bc409a5be4ab)

11-15T17:47:36+00:00  rolling-upgrade-upgrade--centos-db-node-050cdd46-0-2 !ERR     | scylla: 
[shard 0] auth_service - Unexpected exception while revoking all permissions on dropped table: 
exceptions::read_failure_exception (Operation failed for system_auth.role_permissions - received 0 responses and 1 failures from 1 CL=LOCAL_ONE.)

### Re-fill the test data
11-15 18:01:22,224 > Re-Populate DB with many types of tables and data

11-15 18:02:10,162 > 2019-11-15T18:02:05+00:00  rolling-upgrade-upgrade--centos-db-node-050cdd46-0-2 !ERR     | scylla: [shard 2] migration_manager - Failed to update definitions from 10.142.0.39:2: query::deserialization_error (failed on column system_schema.scylla_tables#cdc (version: d15e4b38-4294-3cf3-b7e2-5485bd5c9367, id: 0, index: 0, type: org.apache.cassandra.db.marshal.FrozenType(org.apache.cassandra.db.marshal.MapType(org.apache.cassandra.db.marshal.UTF8Type,org.apache.cassandra.db.marshal.UTF8Type))): seastar::internal::backtraced<marshal_exception> (marshaling error: read_simple_bytes - not enough bytes (requested 131207658, got 8) Backtrace:   0x2d67f2d#012  0xa7f75e#012  0x44b353#012  0x4e7094#012  0x13d38f2#012  0x1b16f6e#012  0x1b18e70#012  0x1b1b6b0#012  0x1b192a7#012  0x1b19844#012  0x1b1a662#012  0xb9469c#012  0x177c6e0#012  0x17805e5#012  0x1783cdb#012  0x17853d8#012  0x7e2143#012))

11-15T18:02:28+00:00  rolling-upgrade-upgrade--centos-db-node-050cdd46-0-2 !WARNING | scylla: [shard 0] seastar - Exceptional future ignored: std::_Nested_exception<no_such_column_family> (Can't find a column family with UUID 083fffc0-07d2-11ea-8234-000000000002): std::out_of_range (_Map_base::at), backtrace:   0x2d67f2d#012  0x2ab82c4#012  0x961d51#012  0x2ab5a91#012  0x2ab5c9f#012  0x2b1c2fd#012  0x2a62af1#012  0x2a6370e#012  0x795c11#012  /opt/scylladb/libreloc/libc.so.6+0x23f32#012  0x7e456d
```

@roydahan @slivne

haaawk:
Addition of cdc column in scylla_tables changes how schema
digests are calculated, and affect the ABI of schema update
messages (adding a column changes other columns' indexes
in frozen_mutation).

To fix this, extend the schema_tables mechanism with support
for the cdc column, and adjust schemas and mutations to remove
that column when sending schemas during upgrade.

Fixes #5288 
Tests: unit(dev)


haaawk:
I still have to test this on a cluster but please have a look in the meantime.

db/schema_features.hh
```diff
@@ -33,12 +33,14 @@ enum class schema_feature {
...
     COMPUTED_COLUMNS,
+    CDC_OPTIONS,
 };
 
 using schema_features = enum_set<super_enum<schema_feature,
     schema_feature::VIEW_VIRTUAL_COLUMNS,
     schema_feature::DIGEST_INSENSITIVE_TO_EXPIRY,
-    schema_feature::COMPUTED_COLUMNS
+    schema_feature::COMPUTED_COLUMNS,
+    db::schema_feature::CDC_OPTIONS
```
dekimir: `db::` is superfluous

haaawk: Fixed. Thanks.

haaawk:
I only removed unnecessary "db:" prefix in db/schema_features.hh

elcallio:
Question: If we made cdc an extension attribute instead we would not have to deal with schema conflict issues (it would just be another map in the extensions map-map). Then we could simply refuse to process CDC until whole cluster reports it supports the feature.

haaawk:
What's an extension attribute, @elcallio? Sounds good in general but I have to learn more about the mechanism for extension attributes because right now I don't know anything about it.

elcallio:
Extension attributes are what encodes schema_extension:s. Making CDC an extension might be a little cumbersome though, so one could instead take the cdc options struct and serialize it into the extensions map at schema_tables.cc:1660 and the special case a "cdc" value at line 2125. 
This could be a stop-gap way of putting this in schemas until next major version when it can fully graduate to its own column.

haaawk:
Sounds like a good idea to put CDC stuff in extensions until we stabilize the API. The question is whether we have a migration process established for switching from extension to own column? or will we have to write something new from scratch just for CDC?

haaawk:
It seems that we have to do this change anyway because we need CDC feature to guard other CDC tables in system tables as well.

db/config.cc
```diff
@@ -691,6 +691,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
     , cpu_scheduler(this, "cpu_scheduler", value_status::Used, true, "Enable cpu scheduling")
     , view_building(this, "view_building", value_status::Used, true, "Enable view building; should only be set to false when the node is experience issues due to view building")
     , enable_sstables_mc_format(this, "enable_sstables_mc_format", value_status::Used, true, "Enable SSTables 'mc' format to be used as the default file format")
+    , enable_cdc(this, "enable_cdc", value_status::Used, false, "Enable Change Data Capture")
```
avikivity: Why do we need this flag? There is already a per-table flag.

haaawk: We need this to be able to do rolling upgrade of the new Scylla binary with CDC schema changes not present and then enable CDC on such cluster. It would work like this:
1. Do rolling upgrade with enable-cdc = false
2. Switch enable-cdc to true on all nodes
Otherwise we will get the problem described in #5288

avikivity:
I'd like to see new tests in schema_change_test.cc.

I'd also like to understand why the existing tests in schema_change_test.cc did not catch the breakage. It's supposed to catch exactly this! Maybe the test doesn't enable --experimental?

haaawk:
Yes. It is weird that the test didn't break. Will investigate.

haaawk:
I explicitly set:
db_cfg.experimental({true}, db::config::config_source::CommandLine);
db_cfg.enable_cdc({true}, db::config::config_source::CommandLine);
and it's still not failing.

haaawk:
I checked and the test uses schema with my new column so it turns out that adding a column does not break digest? I will dig some more.

haaawk:
I think I might have broken this test with one of my previous commits:
```
diff --git a/tests/schema_change_test.cc b/tests/schema_change_test.cc
index 2494dc5c9..607ae6425 100644
--- a/tests/schema_change_test.cc
+++ b/tests/schema_change_test.cc
@@ -371,8 +371,10 @@ SEASTAR_TEST_CASE(test_merging_does_not_alter_tables_which_didnt_change) {
             muts2.push_back(db::schema_tables::make_scylla_tables_mutation(s0, api::new_timestamp()));
             mm.announce(muts2).get();
 
-            BOOST_REQUIRE(s1 == find_table().schema());
-            BOOST_REQUIRE_EQUAL(legacy_version, find_table().schema()->version());
+            // SCYLLA_TABLES have additional columns so announcing its mutation
+            // changes the tables
+            BOOST_REQUIRE(s1 != find_table().schema());
+            BOOST_REQUIRE(legacy_version != find_table().schema()->version());
         });
     });
 }

I will fix that.
```

elcallio:
It is true that once/if we want to move cdc to an explicit column, the problem arises again. We could of course just put cdc (and similar) in a new system table(s). Full schema digest will change, but the actual schema for individual (existing) tables will remain the same.

haaawk:
Ok here's what I've learned:
1. test_merging_does_not_alter_tables_which_didnt_change was a test for a migration from 1.7 to 2.0 and it wasn't relevant to the addition of cdc column to scylla_tables
2. test_schema_digest_does_not_change_with_disabled_features probably did not fail because of cdc column for tested tables is empty. I will add a test similar to UDF test to see if I can get it fail.

haaawk:
I added a test for cdc_options in schema_change_test.

avikivity:
Do we know why it didn't fail before? Does the new test fail if you run it without the fixes in this patchset?

haaawk:
I don't know yet why the tests didn't fail before, @avikivity - still looking into it.
The new test does not fail without the fix because it tests the case when CDC is enabled. Older tests should be failing with CDC enabled but they don't :(.

haaawk:
Ok @avikivity - the answer why new column in scylla_tables does not cause test to fail is that we don't take system tables into account when calculating digiest:
https://github.com/scylladb/scylla/blob/master/db/schema_tables.cc#L629

haaawk:
@avikivity Do you want me to do anything else here or can we merge?

service/migration_manager.cc
```diff
@@ -903,8 +904,9 @@ future<> migration_manager::announce(std::vector<mutation> mutations, bool annou
 future<> migration_manager::push_schema_mutation(const gms::inet_address& endpoint, const std::vector<mutation>& schema)
 {
     netw::messaging_service::msg_addr id{endpoint, 0};
-    auto fm = std::vector<frozen_mutation>(schema.begin(), schema.end());
-    auto cm = std::vector<canonical_mutation>(schema.begin(), schema.end());
+    auto adjusted_schema = db::schema_tables::adjust_schema_for_schema_features(schema, get_local_storage_service().cluster_schema_features());
```
tgrabiec: Why do we need this change?

The "cdc" column should not be set until you can push it to all nodes.

tgrabiec: We still need to adjust the schema of the mutations we send, so ok.

haaawk:
I rebased the code on top of master. No other changes.

db/config.cc
avikivity: No need for a flag. Condition everything on cluster_supports_cdc() (which is only true if all nodes report they support cdc, and each node only enables the feature if it is in experimental mode).

avikivity:
> Ok @avikivity - the answer why new column in scylla_tables does not cause test to fail is that we don't take system tables into account when calculating digiest:
> https://github.com/scylladb/scylla/blob/master/db/schema_tables.cc#L629

But what about the non-system tables in that schema? They have a new cdc column, and even if the column is null, it still changes the indexes of all columns that follow it.

avikivity:
> @avikivity Do you want me to do anything else here or can we merge?

Yes, I'm not satisfied with the explanation about why the tests didn't fail previously, and I think enable_cdc should go away.

haaawk:
> > Ok @avikivity - the answer why new column in scylla_tables does not cause test to fail is that we don't take system tables into account when calculating digiest:
> > https://github.com/scylladb/scylla/blob/master/db/schema_tables.cc#L629
> 
> But what about the non-system tables in that schema? They have a new cdc column, and even if the column is null, it still changes the indexes of all columns that follow it.

I'm not sure I follow. The new column was added only to scylla_tables table. I clearly am missing something - why would all tables get a new column?

avikivity:
A new column was added to scylla_tables. Every table gets a row in scylla_tables, with a value (usually null) for the new column.

When calculating digests, we go over every row in scylla_tables, but we skip over rows that correspond to system tables. We use the rows that correspond to user tables to calculate a digest.

Nodes that don't have the new column will calculate a different digest. The digest calculation depends not only on the value of the column, but on its very existence in the schema, since that alters the indexes of following columns, which alter the digest.

This last paragraph doesn't apply to digests calculated using canonical mutations, but the test that should have failed predates that.

haaawk:
I see. I will run some tests and try to figure that out.

haaawk:
So here's what I found @avikivity and @tgrabiec:
schema_change_test does not test scylla_tables at all. SSTables for scylla_tables used by the test contain only cell tombstones so digest calculation ignores them and effectively scylla_tables is empty from the test perspective.

It seems that the sstables used by the test are wrong - @tgrabiec could you confirm?
I tried to regenerated them using the test but the test does not generate scylla_tables sstables at all.

tgrabiec:
On Wed, Dec 4, 2019 at 9:30 AM Piotr Jastrzębski <notifications@github.com>
wrote:

> So here's what I found @avikivity <https://github.com/avikivity> and
> @tgrabiec <https://github.com/tgrabiec>:
> schema_change_test does not test scylla_tables at all. SSTables for
> scylla_tables used by the test contain only cell tombstones so digest
> calculation ignores them and effectively scylla_tables is empty from the
> test perspective.
>

scylla_tables is empty in the currently pre-generated sstables and that's
expected because scylla_tables didn't contain any columns except for
"version" (which is deleted by design) at the time the sstables were
generated. The currently generated sstables reflect the pre-upgrade state
of a node during upgrade.

The test uses calculate_schema_digest() to calculate the digest, which does
not ignore scylla_tables. It uses schema_tables::all_tables() to get the
list of tables, and scylla_tables is there unconditionally.


> It seems that the sstables used by the test are wrong - @tgrabiec
> <https://github.com/tgrabiec> could you confirm?
>

Why do you think they're wrong?


> I tried to regenerated them using the test but the test does not generate
> scylla_tables sstables at all.
>

The test uses the standard "create table" CQL statement, so you should get
the same as you would if you created the table using cqlsh.


>

haaawk:
@tgrabiec ok so if sstables for scylla_tables are empty for test_schema_digest_does_not_change and test_schema_digest_does_not_change_after_computed_columns then any modification to SCYLLA_TABLES won't be caught by the test. scylla_tables are included in digest calculation but because there are no rows in this table they have no impact on digest.

test_schema_digest_does_not_change_with_functions has no sstables for scylla_tables at all.

The question is what we want to do about this? It seems that before my addition of a column to scylla_tables it was always empty and had no impact on digest. This means it's ok that the tests didn't break after addition of a new column. @avikivity are you satisfied with the explanation now?

tgrabiec:
On Wed, Dec 4, 2019 at 1:48 PM Piotr Jastrzębski <notifications@github.com>
wrote:

> @tgrabiec <https://github.com/tgrabiec> ok so if sstables for
> scylla_tables are empty for test_schema_digest_does_not_change and
> test_schema_digest_does_not_change_after_computed_columns then any
> modification to SCYLLA_TABLES won't be caught by the test. scylla_tables
> are included in digest calculation but because there are no rows in this
> table they have no impact on digest.
>
> test_schema_digest_does_not_change_with_functions has no sstables for
> scylla_tables at all.
>
> The question is what we want to do about this? It seems that before my
> addition of a column to scylla_tables it was always empty and had no impact
> on digest. This means it's ok that the tests didn't break after addition of
> a new column. @avikivity <https://github.com/avikivity> are you satisfied
> with the explanation now?
>

In other words, the test verifies that a schema read from disk before and
after the upgrade has the same digest.

What would you like the test to verify instead?

>

tgrabiec:
On Wed, Dec 4, 2019 at 3:27 PM Tomasz Grabiec <tgrabiec@gmail.com> wrote:

>
>
> On Wed, Dec 4, 2019 at 1:48 PM Piotr Jastrzębski <notifications@github.com>
> wrote:
>
>> @tgrabiec <https://github.com/tgrabiec> ok so if sstables for
>> scylla_tables are empty for test_schema_digest_does_not_change and
>> test_schema_digest_does_not_change_after_computed_columns then any
>> modification to SCYLLA_TABLES won't be caught by the test. scylla_tables
>> are included in digest calculation but because there are no rows in this
>> table they have no impact on digest.
>>
>> test_schema_digest_does_not_change_with_functions has no sstables for
>> scylla_tables at all.
>>
>> The question is what we want to do about this? It seems that before my
>> addition of a column to scylla_tables it was always empty and had no impact
>> on digest. This means it's ok that the tests didn't break after addition of
>> a new column. @avikivity <https://github.com/avikivity> are you
>> satisfied with the explanation now?
>>
>
> In other words, the test verifies that a schema read from disk before and
> after the upgrade has the same digest.
>
> What would you like the test to verify instead?
>

We may want to add a test case with CDC-enabled tables to this test, to
verify that CDC-enabled tables have the same digest after future upgrades.

avikivity:
Ok, so we got "lucky" - adding a column corrupts columns that come after it, but in that test, all the columns were empty so there was no column after cdc.

Still, it must be fixed, because in other scenarios scylla_tables will not be empty.

slivne:
@haaawk - its in your court

haaawk:
So here's what I found @avikivity and @tgrabiec:
schema_change_test does not test scylla_tables at all. SSTables for scylla_tables used by the test contain only cell tombstones so digest calculation ignores them and effectively scylla_tables is empty from the test perspective.

It seems that the sstables used by the test are wrong - @tgrabiec could you confirm?
I tried to regenerated them using the test but the test does not generate scylla_tables sstables at all.

> On Wed, Dec 4, 2019 at 3:27 PM Tomasz Grabiec ***@***.***> wrote: On Wed, Dec 4, 2019 at 1:48 PM Piotr Jastrzębski ***@***.***> wrote: > @tgrabiec <https://github.com/tgrabiec> ok so if sstables for > scylla_tables are empty for test_schema_digest_does_not_change and > test_schema_digest_does_not_change_after_computed_columns then any > modification to SCYLLA_TABLES won't be caught by the test. scylla_tables > are included in digest calculation but because there are no rows in this > table they have no impact on digest. > > test_schema_digest_does_not_change_with_functions has no sstables for > scylla_tables at all. > > The question is what we want to do about this? It seems that before my > addition of a column to scylla_tables it was always empty and had no impact > on digest. This means it's ok that the tests didn't break after addition of > a new column. @avikivity <https://github.com/avikivity> are you > satisfied with the explanation now? > In other words, the test verifies that a schema read from disk before and after the upgrade has the same digest. What would you like the test to verify instead?
> We may want to add a test case with CDC-enabled tables to this test, to verify that CDC-enabled tables have the same digest after future upgrades.

I've done that already.

haaawk:
> Ok, so we got "lucky" - adding a column corrupts columns that come after it, but in that test, all the columns were empty so there was no column after cdc.
> 
> Still, it must be fixed, because in other scenarios scylla_tables will not be empty.

I already added a test when cdc column is not empty so this should be covered now.

haaawk:
I got rid of the enable-cdc flag

haaawk:
@avikivity I rebased the change on top of master and added a commit that explicitly disables CDC on old tests. Please have a look.

tests/schema_change_test.cc
```diff
@@ -715,3 +715,24 @@ SEASTAR_TEST_CASE(test_schema_digest_does_not_change_with_functions) {
             e.execute_cql("create aggregate my_agg(int) sfunc my_add stype int finalfunc twice;").get();
         });
```
avikivity: Please document in the change log what cases you are testing. This will allow a reviewer to consider whether there are missing tests, and whether the tests do in fact test what the commit claims they test.

tests/schema_change_test.cc
```diff
@@ -715,3 +715,24 @@ SEASTAR_TEST_CASE(test_schema_digest_does_not_change_with_functions) {
...
+        utils::UUID("018fccba-8050-3bb9-a0a5-2b3c5f0371fe"),
+        utils::UUID("018fccba-8050-3bb9-a0a5-2b3c5f0371fe"),
+        utils::UUID("58f4254e-cc3b-3d56-8a45-167f9a3ea423"),
+        utils::UUID("48fda4f8-d7b5-3e59-a47a-7397989a9bf8"),
+        utils::UUID("8049bcfe-eb01-3a59-af33-16cef8a34b45"),
+        utils::UUID("2195a821-b2b8-3cb8-a179-2f5042e90841")
+    };
+    return test_schema_digest_does_not_change_with_disabled_features(
+        "./tests/sstables/schema_digest_test_cdc_options",
+        std::set<sstring>{},
```
avikivity: How were those sstables created? Should be documented here.

haaawk: They were generated the same ways as sstables for other test_schema_digest_does_not_change_with_disabled_features tests. There's a regenerate flag that you set to true to generate them.

tests/schema_change_test.cc
haaawk: I added commit message.

haaawk:
I added commit message in a patch that adds new test case @avikivity

haaawk:
I fixed dtests by making sure new node is using old schema for all schema updates it receives before CDC feature is enabled.

BYO is running here: https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/789/

tests/schema_change_test.cc
nyh: @haaawk and where was *that* documented?
There is some general "NOTICE" about "Regenerating data for this test may be necessary when a system ta
ble is added." but it is completely mysterious what it means, and what happens when you regenerate - I would be worried that if I regenerate the tables I would just make the existing bug permanent by committing wrong "expected" values. There definitely needs to be some more precise documentation on how these tests work, and how to properly "generate" them.

tests/schema_change_test.cc
```diff
@@ -708,7 +708,7 @@ SEASTAR_TEST_CASE(test_schema_digest_does_not_change_with_functions) {
     };
     return test_schema_digest_does_not_change_with_disabled_features(
         "./tests/sstables/schema_digest_with_functions_test",
-        std::set<sstring>{},
+        std::set<sstring>{"CDC"},
```
nyh: I don't understand why a test called "test_schema_digest_does_not_change_with_functions", which apparently tests something with "functions" (UDF?), needs to list specifically CDC out of the gazillion other features we have... Is it because we have pre-CDC checksums saved, and now with CDC these saved checksums no longer match so we need to disable CDC? Is this the best we can do? Should a comment here (and in other places) explain this issue to the uninitiated like myself who would be reading this code?

avikivity:
> I fixed dtests by making sure new node is using old schema for all schema updates it receives before CDC feature is enabled.
 

Can you elaborate? Did you change dtest? Or did you change scylla in order to fix the dtest failures?

What is the change?

haaawk:
> > I fixed dtests by making sure new node is using old schema for all schema updates it receives before CDC feature is enabled.
> 
> Can you elaborate? Did you change dtest? Or did you change scylla in order to fix the dtest failures?
> 
> What is the change?

The dtest was good. I didn't change the test. The problem was in the code. I forgot to update the receiving side of the schema communication. New node was sending a schema without new column as long as the cluster wasn't supporting CDC but at the same time new nodes were expecting schema with the new column. I made them expect schema without new column until CDC feature is enabled. Then they expect schema with the new column.

avikivity:
Sounds reasonable, can you point me to the change?

avikivity:
Also, can you detail the local upgrade tests you made? If they look sufficient I'll merge in advance of @amoskong tests.

haaawk:
> Sounds reasonable, can you point me to the change?

https://github.com/scylladb/scylla/pull/5346/commits/b0439039239ca3968d6b5b21fc04fb1770f65d46#diff-5f8642de298aefafdc25e42d09aaee45R319

haaawk:
> Also, can you detail the local upgrade tests you made? If they look sufficient I'll merge in advance of @amoskong tests.

I'm testing it now. I added two more fixes to schema_mutations::cdc_options() to handle the case when we upgrade a node and old data does not contain the new column. Fixes are in:
https://github.com/scylladb/scylla/pull/5346/commits/6c6a0d9a4909c9c95363462bb60bf83010494257#diff-ae7c3ae27fc99a69ed208d7ab253aa61R112
and:
https://github.com/scylladb/scylla/pull/5346/commits/cac4db428891ff86d7d28b3dbb220081c79abe8f#diff-ae7c3ae27fc99a69ed208d7ab253aa61R109

haaawk:
I'm also running dtests locally @avikivity because byo failed in some random manner which in no way implicate CDC

tests/schema_change_test.cc
haaawk: That's a question for the creator of the test @nyh. I can't speed for their intentions. What I did was to look at how it was done before, talked to Sarna how he did that last time and followed the same route. Of course you can persist a bug by committing wrong expected values if you're not careful.

tests/schema_change_test.cc
haaawk: Yes. We have to either disable CDC feature on that test or update the checksums. Both ways are more or less equivalent.

haaawk:
I pushed a version with a fix included in canonical_mutation::to_mutation. Do you think it's better @avikivity ?

canonical_mutation.cc
```diff
@@ -77,7 +78,9 @@ mutation canonical_mutation::to_mutation(schema_ptr s) const {
 
     mutation m(std::move(s), std::move(pk));
 
-    if (version == m.schema()->version()) {
+    // We have to check CDC feature here because until cluster starts supporting
+    // CDC, schema used to exchange info between nodes does not contain CDC column in scylla_tables
+    if (version == m.schema()->version() && service::get_local_storage_service().cluster_supports_cdc()) {
         auto partition_view = mutation_partition_view::from_view(mv.partition());
```
avikivity: canonical_mutation is a primitive, it shouldn't know anything about cluster_supports_cdc(). We should build complicated stuff from primitive stuff, not push special cases into primitives.

I don't understand how that equality test succeeded. If we're sending a schema from a node that has cdc to a node that doesn't have cdc, the schema versions should be different. So I think that check is broken somehow, and you compenated for it by adding the extra check.

haaawk: My suspicious is that a new node (with CDC) sends a canonical mutation to another new node (with CDC). They both have the same schema but before sending CDC column gets removed from the mutation. That's probably why the version of the schema matches. Will check this by testing.

avikivity: Ok, the schema matches, but then what's the problem? It should just work.

We won't pass the cdc column, but nobody should care about it.

avikivity: (we should have debug mode add the schema_ptr to mutation objects and check them on any access, that would remove a whole class of such bugs)

avikivity: (not mutation, but the various mutation_fragments that don't have schemas attached)

haaawk: > Ok, the schema matches, but then what's the problem? It should just work.
> 
> We won't pass the cdc column, but nobody should care about it.

If we try to deserialize the mutation that does not have cdc column with schema that has cdc column, it takes the value of version column as value of cdc column and puts garbage in version column.

avikivity: In which call?

It can't be canonical_mutation::to_mutation(), because that explicitly checks that the schemas match.

db/schema_tables.cc
```diff
@@ -674,6 +693,14 @@ future<std::vector<canonical_mutation>> convert_schema_to_mutations(distributed<
     return map_reduce(all_table_names(features), map, std::vector<canonical_mutation>{}, reduce);
 }
 
+std::vector<mutation>
+adjust_schema_for_schema_features(std::vector<mutation> schema, schema_features features) {
+    for (auto& m : schema) {
+        m = redact_columns_for_missing_features(m, features);
```
avikivity: std::move(m) in the call

canonical_mutation.cc
avikivity: > My suspicious is that a new node (with CDC) sends a canonical mutation to another new node (with CDC). They both have the same schema but before sending CDC column gets removed from the mutation. That's probably why the version of the schema matches. Will check this by testing.

How confident are you in this?

avikivity: Note that when a column gets removed, the schema version changes:

```c++
static
mutation
redact_columns_for_missing_features(mutation m, schema_features features) {
    if (features.contains(schema_feature::CDC_OPTIONS)) {
        return std::move(m);
    }
    if (m.schema()->cf_name() != SCYLLA_TABLES) {
        return std::move(m);
    }
    slogger.debug("adjusting schema_tables mutation due to possible in-progress cluster upgrade");
    m.upgrade(scylla_tables(features));
    return std::move(m);
}
```

mutation::upgrade() updates the schema of the mutation.

denesb: > > Ok, the schema matches, but then what's the problem? It should just work.
> > We won't pass the cdc column, but nobody should care about it.
> 
> If we try to deserialize the mutation that does not have cdc column with schema that has cdc column, it takes the value of version column as value of cdc column and puts garbage in version column.

I don't see how this can happen. We use a visitor pattern to deserialize mutations, and thus it is the content of the serialized mutation that drives the deserialization process. How can we visit cells that are not there?

db/schema_tables.cc
```diff
@@ -613,14 +616,40 @@ schema_ptr aggregates() {
     }
 #endif
 
+static
+mutation
+redact_columns_for_missing_features(mutation m, schema_features features) {
+    if (features.contains(schema_feature::CDC_OPTIONS)) {
+        return std::move(m);
```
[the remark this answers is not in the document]
denesb: You need it when your return a function parameter. Otherwise this will copy.

db/schema_tables.cc
```diff
@@ -299,19 +299,22 @@ schema_ptr tables() {
 }
 
 // Holds Scylla-specific table metadata.
-schema_ptr scylla_tables() {
-    static thread_local auto schema = [] {
+schema_ptr scylla_tables(schema_features features) {
```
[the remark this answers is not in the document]
denesb: I think that instead of passing `features` here and making the addition of the `cdc` column conditional on it, we should just remove this column when we need to redact it, during upgrade.
When writing code for mixed cluster support, we want to disturb the common path of a homogeneous cluster the least possible, and do all the support related extra work in the mixed cluster case. Here you add extra code and logic for the vastly more common case of a non-mixed cluster.

db/schema_tables.cc
```diff
@@ -613,14 +616,28 @@ schema_ptr aggregates() {
...
+mutation
+redact_columns_for_missing_features(mutation m, schema_features features) {
+    if (features.contains(schema_feature::CDC_OPTIONS)) {
+        return std::move(m);
+    }
+    if (m.schema()->cf_name() != SCYLLA_TABLES) {
+        return std::move(m);
+    }
+    slogger.debug("adjusting schema_tables mutation due to possible in-progress cluster upgrade");
+    m.upgrade(scylla_tables(features));
```
denesb: We should just remove the `cdc` column here, instead of passing `features` to `scylla_tables()`.

canonical_mutation.cc
tgrabiec: > (we should have debug mode add the schema_ptr to mutation objects and check them on any access, that would remove a whole class of such bugs)

There was a patch series posted to do just that: "Catch invalid schema version access on rows", I need to resurrect it.

db/schema_tables.cc
avikivity: This is removing the cdc column, no? And adjusting the mutation's schema.

denesb: Yes, it does. What I don't like is that it pushes this removal down to `schema_tables()`. If we keep using this pattern, `schema_tables()` will be full of branches that conditionally add columns to the schema, when this code is used only in a brief transitional period. I think we should strive to contain such redacting code to an as small as possible area, instead of spreading it in the codebase. In theory all such redacting will be removed one day and it will be much easier if this will just entail removing one function, as opposed to going through the code and tracing all the different places this logic have spread to.

haaawk: schema_features is a way to handle rolling upgrade in other places so I think we should be consistent and have a single approach.

haaawk: Notice that there's no if in scylla_tables. It selects schema from a table based on condition.

denesb: I see this:
```
        if (has_cdc_options) {
            sb.with_column("cdc", map_type_impl::get_instance(utf8_type, utf8_type, false));
        }
```

haaawk: it's executed only once:

```
schema_ptr scylla_tables(schema_features features) {
    static auto make = [] (bool has_cdc_options) -> schema_ptr {
        auto id = generate_legacy_id(NAME, SCYLLA_TABLES);
        auto sb = schema_builder(NAME, SCYLLA_TABLES, std::make_optional(id))
            .with_column("keyspace_name", utf8_type, column_kind::partition_key)
            .with_column("table_name", utf8_type, column_kind::clustering_key)
            .with_column("version", uuid_type)
            .set_gc_grace_seconds(schema_gc_grace)
            .with_version(generate_schema_version(id));
        if (has_cdc_options) {
            sb.with_column("cdc", map_type_impl::get_instance(utf8_type, utf8_type, false));
        }
        return sb.build();
    };
    +---- notice the static
    v
    STATIC thread_local schema_ptr schemas[2] = { make(false), make(true) };
    return schemas[features.contains(schema_feature::CDC_OPTIONS)];
}
```

db/schema_tables.cc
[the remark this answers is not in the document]
avikivity: The schema_features parameter makes it impossible to use the wrong schema accidentally.

If you're worried about performance, this isn't in any hot path, and in any case translates to an array lookup.

db/schema_tables.cc
denesb: Ok, maybe I'm overreacting this. I would have done it something like:
```
auto adjusted_s = schema_builder(schema_tables()).remove_column("cdc_options").build();
m.upgrade(adjusted_s);
```
Just to keep all this adjusting/redacting thing localized to one function. I won't insist though.

avikivity: This breaks down when you start having multiple orthogonal features. Each feature is tied to a column, so if you have many features, you end up with many possible variations.

canonical_mutation.cc
haaawk: I added some prints to canonical_mutation::to_mutation:
```
mutation canonical_mutation::to_mutation(schema_ptr s) const {
    auto in = ser::as_input_stream(_data);
    auto mv = ser::deserialize(in, boost::type<ser::canonical_mutation_view>());

    auto cf_id = mv.table_id();
    if (s->id() != cf_id) {
        throw std::runtime_error(format("Attempted to deserialize canonical_mutation of table {} with schema of table {} ({}.{})",
                                        cf_id, s->id(), s->ks_name(), s->cf_name()));
    }

    auto version = mv.schema_version();
    auto pk = mv.key();

    mutation m1(s, pk);
    mutation m2(s, pk);

    {
    auto partition_view = mutation_partition_view::from_view(mv.partition());
    mutation_application_stats app_stats;
    m1.partition().apply(*m1.schema(), partition_view, *m1.schema(), app_stats);
    }
    {
    column_mapping cm = mv.mapping();
    converting_mutation_partition_applier v(cm, *m2.schema(), m2.partition());
    auto partition_view = mutation_partition_view::from_view(mv.partition());
    partition_view.accept(cm, v);
    }
    std::cerr << "haaawk " << version << " <-> " << s->version() << std::endl << std::endl << m1 << std::endl << std::endl << m2 << std::endl;
    return version == s->version() ? m1 : m2;
}
```
and I got:
```
haaawk d15e4b38-4294-3cf3-b7e2-5485bd5c9367 <-> d15e4b38-4294-3cf3-b7e2-5485bd5c9367

{system_schema.scylla_tables {key: pk{00026b73}, token:3274692326281147944} {mutation_partition:
 static: cont=1 {row: },
 clustered: {
    {rows_entry: cont=true dummy=false {position: clustered,ckp{00026366},0} {deletable_row: {row:
      {0, { atomic_cell{7aa08101217911eab225000000000000;ts=1576661622032685;expiry=-1,ttl=0} }}}}}}}}

{system_schema.scylla_tables {key: pk{00026b73}, token:3274692326281147944} {mutation_partition:
 static: cont=1 {row: },
 clustered: {
    {rows_entry: cont=true dummy=false {position: clustered,ckp{00026366},0} {deletable_row: {row:
      {1, { atomic_cell{7aa08101217911eab225000000000000;ts=1576661622032685;expiry=-1,ttl=0} }}}}}}}}
```

The cannonical_mutation that the method is called on is a mutation without CDC column received from another node.
I performed both normal deserialization and schema conversion to compare the results.
As you can see, schema version of canonical mutation (that does not contain CDC column) and schema version on the node (that contains CDC column) are the same so the method "thinks" that it should perform deserialization without schema conversion.
Is |s|, column 0 is CDC column and 1 is version column.
cannonical_mutation contains only data for version column.
You can see that deserialization without schema conversion places data from cannonical mutation into CDC column.
Deserialization with schema conversion places data from cannonical mutation into version column, correctly.

So the problem is that when we exchange schema with other nodes and remove a column, the version of the schema stays the same.

Any thoughts, @avikivity ?

haaawk: That's why comparing not only schema version but also CDC feature solved the problem.
I understand @avikivity why you didn't like the approach so would you rather accept the previous solution that was removing CDC column from |s| before passing it to cannonical_mutation::to_mutation in migration_manager?

avikivity: That's the bug for sure, now to understand how it happened.

avikivity: It's this line:

```
            .with_version(generate_schema_version(id));
```

avikivity: So I propose to make the version input "1" without cdc, and "2" after cdc (and make generate_schema_version accept version_sequence_number as a parameter instead of reading it from the global version_sequence_number)

avikivity: @tgrabiec?

avikivity: > That's why comparing not only schema version but also CDC feature solved the problem.
> I understand @avikivity why you didn't like the approach so would you rather accept the previous solution that was removing CDC column from |s| before passing it to cannonical_mutation::to_mutation in migration_manager?

Can you refer me to that? I have no idea what |s| is.

haaawk: > > That's why comparing not only schema version but also CDC feature solved the problem.
> > I understand @avikivity why you didn't like the approach so would you rather accept the previous solution that was removing CDC column from |s| before passing it to cannonical_mutation::to_mutation in migration_manager?
> 
> Can you refer me to that? I have no idea what |s| is.

|s| is schema_ptr taken by canonical_mutation::to_mutation as parameter. I think fixing generate_schema_version is a better solution though.

haaawk: There's this comment on generate_schema_version:
```
// Increase whenever changing schema of any system table.
// FIXME: Make automatic by calculating from schema structure.
static const uint16_t version_sequence_number = 1;

table_schema_version generate_schema_version(utils::UUID table_id) {
    md5_hasher h;
    feed_hash(h, table_id);
    feed_hash(h, version_sequence_number);
    return utils::UUID_gen::get_name_UUID(h.finalize());
}
```
Maybe I just need to increase version_sequence_number?

haaawk: It seems that increasing version_sequence_number helped but I need to run more tests to be sure. Tests that don't fail don't prove anything ;).

haaawk: When I think about it, I can't explain why would it help though. Both nodes will be using increased version_sequence_number so the schema version should be the same again and lead to the same problem.

avikivity: Yes, it needs to be dependent on schema_features.

avikivity: also, I wouldn't want to change the sequence number for all tables needlessly.

avikivity: Maybe it worked because the schema registry happened to return the right schema for the version this time around (we register two schemas with the same version, so it is implementation-defined which schema we get when we look up by version)

haaawk: Ok. So maybe we could take an additional parameter to generate_schema_version that is an offset above version_sequence_number?

avikivity: Makes sense. version_sequence_number must be frozen for this, but it effectively is.

haaawk:
@avikivity I pushed a new version that uses the latest approach we agreed on -> passing offset to generate_schema_version. The change is in:
https://github.com/scylladb/scylla/pull/5346/commits/e9d03c0ddbdeab2deb18a87649091a48205893d4#diff-94c960641daa44618de006a784f3a143R312
https://github.com/scylladb/scylla/pull/5346/commits/e9d03c0ddbdeab2deb18a87649091a48205893d4#diff-f97687fcb8f7030421c64fd164d00b78R107
https://github.com/scylladb/scylla/pull/5346/commits/e9d03c0ddbdeab2deb18a87649091a48205893d4#diff-4a118af3eac66a6f1c68d5a44ef19469R155

db/schema_tables.cc
avikivity: ^

avikivity:
Looks reasonable to me. What kind of testing did it pass? And you undid the previous hacks, yes?

haaawk:
> Looks reasonable to me. What kind of testing did it pass? And you undid the previous hacks, yes?

Yes - I removed previous hacks. I run some dtests that have previously failed. I guess it would be best to ask @amoskong to rerun the upgrade test. What do you think?

avikivity:
If you can run some manual tests it would be better. Waiting for @amoskong means another 18h latency.

haaawk:
Sure. I will do some manual testing and let you know.

haaawk:
One test I run was to perform rolling upgrade of the cluster from the one without cdc enabled to the one with cdc enabled. It worked ok. I will try running two scylla versions as well.

haaawk:
I have troubles compiling 3.1 so I asked @amoskong to rerun the test @avikivity

avikivity:
@haaawk what troubles are these?

avikivity:
@tgrabiec if you have time, please go over this too.

haaawk:
@avikivity it does not compile neither normally or using dbuild.
Example failure (but not the only one):
```
In file included from ../../include/seastar/core/ragel.hh:24,
                 from gen/include/seastar/http/request_parser.hh:24,
                 from ../../include/seastar/http/httpd.hh:24,
                 from ../../include/seastar/core/prometheus.hh:24,
                 from ../../src/core/prometheus.cc:22:
In constructor 'seastar::basic_sstring<char_type, Size, max_size, NulTerminate>::basic_sstring(size_t, char_type) [with char_type = char; Size = unsigned int; Size max_size = 15; bool NulTerminate = true]',
    inlined from 'void seastar::basic_sstring<char_type, Size, max_size, NulTerminate>::resize(size_t, char_type) [with char_type = char; Size = unsigned int; Size max_size = 15; bool NulTerminate = true]' at ../../include/seastar/core/sstring.hh:367:19,
    inlined from 'bool seastar::prometheus::metrics_handler::trim_asterisk(seastar::sstring&)' at ../../src/core/prometheus.cc:655:24,
    inlined from 'virtual seastar::future<std::unique_ptr<seastar::httpd::reply, std::default_delete<seastar::httpd::reply> > > seastar::prometheus::metrics_handler::handle(const sstring&, std::unique_ptr<seastar::httpd::request>, std::unique_ptr<seastar::httpd::reply>)' at ../../src/core/prometheus.cc:667:36:
../../include/seastar/core/sstring.hh:244:15: warning: 'void* memset(void*, int, size_t)' specified bound 18446744073709551613 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
```

another one:
```
[2/76] Building CXX object CMakeFiles/seastar.dir/gen/src/proto/metrics2.pb.cc.o
FAILED: CMakeFiles/seastar.dir/gen/src/proto/metrics2.pb.cc.o 
/usr/bin/g++  -DFMT_SHARED -DSEASTAR_API_LEVEL=2 -DSEASTAR_HAS_MEMBARRIER -DSEASTAR_HAVE_ASAN_FIBER_SUPPORT -DSEASTAR_HAVE_GCC6_CONCEPTS -DSEASTAR_HAVE_HWLOC -DSEASTAR_HAVE_LZ4_COMPRESS_DEFAULT -DSEASTAR_HAVE_NUMA -DSEASTAR_USE_STD_OPTIONAL_VARIANT_STRINGVIEW -I../../include -Igen/include -I../../src -Igen/src -O2 -g -DNDEBUG   -std=gnu++17 -U_FORTIFY_SOURCE -Wno-maybe-uninitialized -Wno-error=unused-result -fvisibility=hidden -UNDEBUG -Wall -Werror -Wno-error=deprecated-declarations -gz -Wno-error -march=westmere -O3 --param inline-unit-growth=300 -fconcepts -MD -MT CMakeFiles/seastar.dir/gen/src/proto/metrics2.pb.cc.o -MF CMakeFiles/seastar.dir/gen/src/proto/metrics2.pb.cc.o.d -o CMakeFiles/seastar.dir/gen/src/proto/metrics2.pb.cc.o -c gen/src/proto/metrics2.pb.cc
In file included from gen/src/proto/metrics2.pb.cc:4:
gen/src/proto/metrics2.pb.h:12:2: error: #error This file was generated by a newer version of protoc which is
 #error This file was generated by a newer version of protoc which is
  ^~~~~
gen/src/proto/metrics2.pb.h:13:2: error: #error incompatible with your Protocol Buffer headers. Please update
 #error incompatible with your Protocol Buffer headers.  Please update
  ^~~~~
gen/src/proto/metrics2.pb.h:14:2: error: #error your headers.
 #error your headers.
  ^~~~~
In file included from gen/src/proto/metrics2.pb.cc:4:
gen/src/proto/metrics2.pb.h:27:10: fatal error: google/protobuf/inlined_string_field.h: No such file or directory
 #include <google/protobuf/inlined_string_field.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
[3/76] Building CXX object CMakeFiles/seastar.dir/src/core/prometheus.cc.o
FAILED: CMakeFiles/seastar.dir/src/core/prometheus.cc.o 
/usr/bin/g++  -DFMT_SHARED -DSEASTAR_API_LEVEL=2 -DSEASTAR_HAS_MEMBARRIER -DSEASTAR_HAVE_ASAN_FIBER_SUPPORT -DSEASTAR_HAVE_GCC6_CONCEPTS -DSEASTAR_HAVE_HWLOC -DSEASTAR_HAVE_LZ4_COMPRESS_DEFAULT -DSEASTAR_HAVE_NUMA -DSEASTAR_USE_STD_OPTIONAL_VARIANT_STRINGVIEW -I../../include -Igen/include -I../../src -Igen/src -O2 -g -DNDEBUG   -std=gnu++17 -U_FORTIFY_SOURCE -Wno-maybe-uninitialized -Wno-error=unused-result -fvisibility=hidden -UNDEBUG -Wall -Werror -Wno-error=deprecated-declarations -gz -Wno-error -march=westmere -O3 --param inline-unit-growth=300 -fconcepts -MD -MT CMakeFiles/seastar.dir/src/core/prometheus.cc.o -MF CMakeFiles/seastar.dir/src/core/prometheus.cc.o.d -o CMakeFiles/seastar.dir/src/core/prometheus.cc.o -c ../../src/core/prometheus.cc
In file included from ../../src/core/prometheus.cc:25:
gen/src/proto/metrics2.pb.h:12:2: error: #error This file was generated by a newer version of protoc which is
 #error This file was generated by a newer version of protoc which is
  ^~~~~
gen/src/proto/metrics2.pb.h:13:2: error: #error incompatible with your Protocol Buffer headers. Please update
 #error incompatible with your Protocol Buffer headers.  Please update
  ^~~~~
gen/src/proto/metrics2.pb.h:14:2: error: #error your headers.
 #error your headers.
  ^~~~~
In file included from ../../src/core/prometheus.cc:25:
gen/src/proto/metrics2.pb.h:27:10: fatal error: google/protobuf/inlined_string_field.h: No such file or directory
 #include <google/protobuf/inlined_string_field.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
[4/76] Building CXX object CMakeFiles/seastar.dir/src/core/posix.cc.o
[5/76] Building CXX object CMakeFiles/seastar.dir/src/core/fsqual.cc.o
[6/76] Building CXX object CMakeFiles/seastar.dir/src/core/fair_queue.cc.o
[7/76] Building CXX object CMakeFiles/seastar.dir/src/core/linux-aio.cc.o
[8/76] Building CXX object CMakeFiles/seastar.dir/src/core/thread_pool.cc.o
[9/76] Building CXX object CMakeFiles/seastar.dir/src/core/exception_hacks.cc.o
[10/76] Building CXX object CMakeFiles/seastar.dir/src/core/alien.cc.o
[11/76] Building CXX object CMakeFiles/seastar.dir/src/core/reactor_backend.cc.o
[12/76] Building CXX object CMakeFiles/seastar.dir/src/core/execution_stage.cc.o
[13/76] Building CXX object CMakeFiles/seastar.dir/src/core/memory.cc.o
[14/76] Building CXX object CMakeFiles/seastar.dir/src/core/app-template.cc.o
[15/76] Building CXX object CMakeFiles/seastar.dir/src/core/future-util.cc.o
[16/76] Building CXX object CMakeFiles/seastar.dir/src/core/metrics.cc.o
[17/76] Building CXX object CMakeFiles/seastar.dir/src/core/fstream.cc.o
[18/76] Building CXX object CMakeFiles/seastar.dir/src/core/file.cc.o
[19/76] Building CXX object CMakeFiles/seastar.dir/src/core/reactor.cc.o
ninja: build stopped: subcommand failed.
[50/340] CXX build/release/gen/cql3/CqlParser.o
ninja: build stopped: subcommand failed.
```

avikivity:
Zap gen/src/proto/metrics2.pb.h (a generated file).

avikivity:
Any news on testing?

avikivity:
I was able to upgrade from 3.1.2 to this branch, both with --experimental-features clear and with --experimental-features cdc.

avikivity:
schema_registry_test fails because of:

```c++
|   109         std::optional<std::map<sstring, sstring>> schema_mutations::cdc_options() const {                                                                                                                                                             │
│  >110             if (_scylla_tables && service::get_local_storage_service().cluster_supports_cdc()) {                                                                                                                                                      │
│   111                 auto rs = query::result_set(*_scylla_tables);                                                                                                                                                                                         │
│   112                 if (!rs.empty()) {                                                                                                                                                                                                                    │
│   113                     auto map = db::schema_tables::get_map<sstring, sstring>(rs.row(0), "cdc");                                                                                                                                                        │
│   114                     if (map && !map->empty()) {                                                                                                                                                                                                       │
│   115                         return map;                                                                                                                                                                                                                   │
│   116                     }                                                                                                                                                                                                                                 │
│   117                 }                                                                                                                                                                                                                                     │
│   118             }                                                                                                                                                                                                                                         │
│   119             return { };                                                                                                                                                                                                                               │
│   120         }        
```

The global isn't initialized.

avikivity:
I'll just drop the second part of the conditional. If the cluster doesn't support cdc yet, then map will be std::nullopt and we'll return an empty value.

avikivity:
No, that's not true.

```
        system |                large_partitions | {'enabled': 'false', 'postimage': 'false', 'preimage': 'false', 'ttl': '86400'} | 04fa9920-9369-3a96-be39-6dd9fdc816b6
```

We need to also change the builder to create tables with no cdc as tables with an empty map.

avikivity:
I think we can do that by adjusting `cdc::options::to_map()`.

avikivity:
I did that and re-verified that everything works.

avikivity:
Queued as ce2beeab302994c79d63885df3b7fa79df220208

avikivity:
Dequeued. There is some bug that causes schema versions calculated with one set of schema_features to be compared against schema versions calculated with against another set. The mismatch happens in querier_cache, and as a result we no longer reuse queries, failing paging_additional_test.TestPagingSavedQueryStateSingularRanges.test_two_partitions.

avikivity:
Rebased and lightly modified branch: https://github.com/avikivity/scylla/tree/cdc-rolling-upgrade

haaawk:
This test looks wrong:
```
Node metrics doesn't match any of the expected metrics:
node_metrics: {'querier_cache_misses': 0, 'querier_cache_memory_based_evictions': 0, 'querier_cache_drops': 2, 'querier_cache_lookups': 2, 'querier_cache_time_based_evictions': 0, 'querier_cache_resource_based_evictions': 0}
expected_metrics: ({'lookups': 2, 'drops': 1}, {})
```

I guess `lookups` should probably be replaced with `querier_cache_lookups` and `drops` with `querier_cache_drops`.

Which does not change the fact that we dropped 2 queries while only one was expected.

haaawk:
When I compare the schemas, they are the same except for the version:
```
schema:
org.apache.cassandra.config.CFMetaData@0x600001140700[cfId=e8b7daa0-2b2c-11ea-8d62-000
000000000,ksName=paging_additional_test_querier_reuse,cfName=test_singular,cfType=Standard,comparator=org.apache.cassandra.db.marshal.CompositeType(org.apache
.cassandra.db.marshal.Int32Type,org.apache.cassandra.db.marshal.UTF8Type),comment=,readRepairChance=0,dcLocalReadRepairChance=0.1,gcGraceSeconds=864000,keyVal
idator=org.apache.cassandra.db.marshal.Int32Type,minCompactionThreshold=4,maxCompactionThreshold=32,columnMetadata=[ColumnDefinition{name=pk, type=org.apache.
cassandra.db.marshal.Int32Type, kind=PARTITION_KEY, componentIndex=0, droppedAt=-9223372036854775808}, ColumnDefinition{name=ck, type=org.apache.cassandra.db.
marshal.Int32Type, kind=CLUSTERING_COLUMN, componentIndex=0, droppedAt=-9223372036854775808}, ColumnDefinition{name=val, type=org.apache.cassandra.db.marshal.
UTF8Type, kind=REGULAR, componentIndex=null, droppedAt=-9223372036854775808}],compactionStrategyClass=class org.apache.cassandra.db.compaction.SizeTieredCompa
ctionStrategy,compactionStrategyOptions={enabled=true},compressionParameters={sstable_compression=org.apache.cassandra.io.compress.LZ4Compressor},bloomFilterF
pChance=0.01,memtableFlushPeriod=0,caching={"keys":"ALL","rows_per_partition":"ALL"},cdc={},defaultTimeToLive=0,minIndexInterval=128,maxIndexInterval=2048,spe
culativeRetry=99.0PERCENTILE,triggers=[],isDense=false,version=5c7329b6-42d5-3696-b86b-eef862d5cba2,droppedColumns={},collections={},indices={}]

vs querier schema:
org.apache.cassandra.config.CFMetaData@0x60000113f880[cfId=e8b7daa0-2b2c-11ea-8d62-000000000000,ksName=paging_additional_test_querier_reuse,cfName=test
_singular,cfType=Standard,comparator=org.apache.cassandra.db.marshal.CompositeType(org.apache.cassandra.db.marshal.Int32Type,org.apache.cassandra.db.marshal.U
TF8Type),comment=,readRepairChance=0,dcLocalReadRepairChance=0.1,gcGraceSeconds=864000,keyValidator=org.apache.cassandra.db.marshal.Int32Type,minCompactionThr
eshold=4,maxCompactionThreshold=32,columnMetadata=[ColumnDefinition{name=pk, type=org.apache.cassandra.db.marshal.Int32Type, kind=PARTITION_KEY, componentInde
x=0, droppedAt=-9223372036854775808}, ColumnDefinition{name=ck, type=org.apache.cassandra.db.marshal.Int32Type, kind=CLUSTERING_COLUMN, componentIndex=0, drop
pedAt=-9223372036854775808}, ColumnDefinition{name=val, type=org.apache.cassandra.db.marshal.UTF8Type, kind=REGULAR, componentIndex=null, droppedAt=-922337203
6854775808}],compactionStrategyClass=class org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy,compactionStrategyOptions={enabled=true},compressio
nParameters={sstable_compression=org.apache.cassandra.io.compress.LZ4Compressor},bloomFilterFpChance=0.01,memtableFlushPeriod=0,caching={"keys":"ALL","rows_pe
r_partition":"ALL"},cdc={},defaultTimeToLive=0,minIndexInterval=128,maxIndexInterval=2048,speculativeRetry=99.0PERCENTILE,triggers=[],isDense=false,version=eb
861fa4-5bc1-3969-bbec-d85106eac871,droppedColumns={},collections={},indices={}]
```

they both have `cdc = {}`.

avikivity:
Yes,  I'm trying to understand why master preserves version while the series changes it.

avikivity:
> This test looks wrong:
> 
> ```
> Node metrics doesn't match any of the expected metrics:
> node_metrics: {'querier_cache_misses': 0, 'querier_cache_memory_based_evictions': 0, 'querier_cache_drops': 2, 'querier_cache_lookups': 2, 'querier_cache_time_based_evictions': 0, 'querier_cache_resource_based_evictions': 0}
> expected_metrics: ({'lookups': 2, 'drops': 1}, {})
> ```
> 
> I guess `lookups` should probably be replaced with `querier_cache_lookups` and `drops` with `querier_cache_drops`.
> 


The test code adds the extra prefix.

haaawk:
What's interesting, the version changes from `eb861fa4-5bc1-3969-bbec-d85106eac871` to `5c7329b6-42d5-3696-b86b-eef862d5cba2` and then back from `5c7329b6-42d5-3696-b86b-eef862d5cba2` to `eb861fa4-5bc1-3969-bbec-d85106eac871`. During a single select.

Issue #5335, opened 2019-11-23
The configration items related to Redis API should be described clearly


fastio:
The configuration items related to redis are as follows:

    redis_transport_port;
    redis_transport_port_ssl;
    enable_redis_protocol;
    redis_read_consistency_level;
    redis_write_consistency_level;
    redis_default_database_count;
    redis_keyspace_options;

These should be described clearly.

#5132

nyh:
Some questions I had while reading the code, and should probably be answered as code comments or in redis.md - or resolved by modifying the code (all of these are from my previous comments on your original pull request):

1. "enable_redis_protocol" - What does this "enable"? The transport_port? The transport_port_ssl? Both?
Because of this ambiguity, in Alternator we decided to have have "--alternator_port" and "--alternator_https_port", which need to be set (no default value) to enable the specific listener - and we don't have a separate "--enable-alternator" option because it's not needed.

2. "redis_transport_port_ssl" - is this actually in use? Isn't its default (9142) a mistake because Scylla already does something else on this port?

3. "redis_read/write_consistency_level" - mention its default, and why. I wonder if the current default (ANY/ONE) is a great idea, or a safer default should be LOCAL_QUORUM.

4. "redis_default_database_count" - maybe this is the "max" database count, not "default" database count? Or maybe I'm misunderstand what you mean here by "default" (can it be modified later to be different from this "default"?)

5.  "redis_keyspace_option" - the description string for this option isn't clear - it is currently "Enable redis protocol, list of properties of replication of redis keyspace." - but does it actually "enable redis protocol"? we have a different option with that name. Also, we need to spell out when does this option to change keyspace options takes effect - the very first node to boot, probably, and the others' options will have to be ignored, because the keyspace already exists? If so, I think this should be said explicitly somewhere.

fastio:
@nyh Thanks for your reply.

> 1. "enable_redis_protocol" - What does this "enable"? The transport_port? The transport_port_ssl? Both?
> Because of this ambiguity, in Alternator we decided to have have "--alternator_port" and "--alternator_https_port", which need to be set (no default value) to enable the specific listener - and we don't have a separate "--enable-alternator" option because it's not needed.

Indeed, the `enable_redis_protocol is unnecessary. I think it's better to replace them with`--redis_port and `--redis_ssl_port.

> 2. "redis_transport_port_ssl" - is this actually in use? Isn't its default (9142) a mistake because Scylla already does something else on this port?

[...]

Issue #5334, opened 2019-11-23
Implement the atomic DEL command in Scylla to support the multiple keys.


fastio:
The current implementation of Redis DEL command in Scylla does not support to delete multiple keys.
It should support delete multiple keys atomically, return the number of actually deleted keys.

As @nyh mentioned in the [Redis API in Scylla](https://github.com/scylladb/scylla/pull/5132) , there features should be supported eventually.

nyh:
@fastio, what did you mean "delete multiple keys atomically" - if one of them is missing, none of the deletes are done? Is this a Redis feature?
I'm aware of two things that the current code is missing, please correct me if this is the wrong list:
1. We need to support the *syntax* of deleting multiple keys in one DEL command.
2. We need to support DEL's feature of returning the number of really deleted keys. I don't know how we can do this efficiently, as this requires a read-before-write (LWT, if we want to be really consistent), which will significantly slow us down. I wonder how often users actually use the return value of DEL, and if they don't often do - maybe it makes sense to have an option to not support it.

fastio:
>  @fastio, what did you mean "delete multiple keys atomically" - if one of them is missing, none of the deletes are done? Is this a Redis feature?

Sorry, It's a mistake here. I want say that we can delete multiple keys at once using the DEL command of Redis.  In the Redis implementation, the DEL command does not remove multiple keys
atomically. The command returns the number of removed keys in the operation.

> I'm aware of two things that the current code is missing, please correct me if this is the wrong list:
> 
> 1. We need to support the syntax of deleting multiple keys in one DEL command.
> 2. We need to support DEL's feature of returning the number of really deleted keys. 
> I don't know  how we can do this efficiently, as this requires a read-before-write (LWT, if we want to be really consistent), which will significantly slow us down. I wonder how often users actually use the return value of DEL, and if they don't often do - maybe it makes sense to have an option to not support it.

You are right. I agree with you. I think users are concerned about whether the keys are deleted correctly in most use scenarios.

PR #5353, opened 2019-11-26
	Redis API: Rename options related to Redis API, describe them clearl…

fastio:
Rename redis_port, which the redis transport listens on for clients.

Rename redis_ssl_port, which the redis TLS transport listens on for clients.

Rename redis_database_count. Set the redis database count.

Rename redis_keyspace_replication_strategy_options. Set the replication strategy for redis keyspace.

Remove unecessary option named .

#5335 

Signed-off-by: Peng Jian <pengjian.uestc@gmail.com>

### DIFF
--- a/cql3/query_processor.hh
+++ b/cql3/query_processor.hh
@@ -158,15 +158,15 @@ public:
         return _cql_stats;
     }
 
-    statements::prepared_statement::checked_weak_ptr get_prepared(const auth::authenticated_user* user_ptr, const prepared_cache_key_type& key) {
-        if (user_ptr) {
-            auto it = _authorized_prepared_cache.find(*user_ptr, key);
+    statements::prepared_statement::checked_weak_ptr get_prepared(const std::optional<auth::authenticated_user>& user, const prepared_cache_key_type& key) {
+        if (user) {
+            auto it = _authorized_prepared_cache.find(*user, key);
             if (it != _authorized_prepared_cache.end()) {
                 try {
                     return it->get()->checked_weak_from_this();
                 } catch (seastar::checked_ptr_is_null_exception&) {
                     // If the prepared statement got invalidated - remove the corresponding authorized_prepared_statements_cache entry as well.
-                    _authorized_prepared_cache.remove(*user_ptr, key);
+                    _authorized_prepared_cache.remove(*user, key);
                 }
             }
         }
--- a/cql3/statements/list_users_statement.cc
+++ b/cql3/statements/list_users_statement.cc
@@ -100,17 +100,16 @@ cql3::statements::list_users_statement::execute(service::storage_proxy& proxy, s
 
     const auto& cs = state.get_client_state();
     const auto& as = *cs.get_auth_service();
-    const auto user = cs.user();
 
-    return auth::has_superuser(as, *user).then([&cs, &as, user](bool has_superuser) {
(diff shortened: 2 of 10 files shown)


fastio:
Rename `redis_transport_port` to `redis_port`, which the redis transport listens on for clients.

Rename `redis_transport_port_ssl` to `redis_ssl_port`, which the redis TLS transport listens on for clients.

Rename `redis_default_database_count` to `redis_database_count`. Set the redis database count.

Rename `redis_keyspace_options` to `redis_keyspace_replication_strategy_options`. Set the replication strategy for redis keyspace.
        
Remove unecessary option named `enable_redis_protocol`.

db/config.cc
```diff
@@ -715,21 +715,20 @@ db::config::config(std::shared_ptr<db::extensions> exts)
...
-    , redis_keyspace_options(this, "redis_keyspace_options", value_status::Used, {}, 
-            "Enable redis protocol, list of properties of replication of redis keyspace. The available options are:\n"
-            "\n"
-            "\tclass : (Default: SimpleStrategy ).\n"
-            "\treplication_factor: (Default: 1) If the class is SimpleStrategy, set the replication factor for the strategy.\n"
-            "\tdatacenter_i: (Default: 1) If the class is NetworkTopologyStrategy, set the replication factor per datacenter.\n"
-            "\n"
-            "The properties of replication for redis keyspace.")
+    , redis_port(this, "redis_port", value_status::Used, 6379, "Port on which the REDIS transport listens for clients.")
+    , redis_ssl_port(this, "redis_ssl_port", value_status::Used, 9142, "Port on which the REDIS TLS native transport listens for clients.")
```
amoskong: The FIXME comment is removed, but the problem is still there.

Do you want to share same port with native_transport_port_ssl? does it has conflict?

nyh:
Two procedural requests:

1. The second commit has nothing to do with this pull request. I can work around this (I will commit each of them separately), but in general it's confusing.

2. Does the first commit fix issue #5335 completely and we should close it? If so, you should use the text "Fixes #5335", not just "#5335" because this causes github to automatically close this issue when the patch is committed. Moreover, you should use the text "Fixes #5335" in the individual patch's description - not in the pull request text (especially in this case when the pull request has multiple unrelated patch).

3. The second commit you added now, do it also fix an existing issue?

fastio:
Change logs:
* DEL： Support to delete multiple keys.
* GET:    Fix the return value when the required key is not exists.
* redis-test: Add the xfailed for DEL command.

db/config.cc
fastio: @amoskong , Sorry. I want change the redis_ssl_port to 6389.

fastio:
> Two procedural requests:
> 1. The second commit has nothing to do with this pull request. I can work around this (I will commit each of them separately), but in general it's confusing.

@nyh  Thanks.

> 2. Does the first commit fix issue #5335 completely and we should close it? If so, you should use the text "Fixes #5335", not just "#5335" because this causes github to automatically close this issue when the patch is committed. Moreover, you should use the text "Fixes #5335" in the individual patch's description - not in the pull request text (especially in this case when the pull request has multiple unrelated patch).

I got it. I will add the text "Fixes #5335" to the patch's description.

> 3. The second commit you added now, do it also fix an existing issue?

Yes, Fixes #5334

PR #5358, opened 2019-11-27
scylla_util: get_scylla_dirs: return default data/commitlog directories if they aren't set

amoskong:
The default values of data_file_directories and commitlog_directory were
commented by commit e0f40ed16aa. It causes scylla_util.py:get_scylla_dirs() to
fail in checking the values.

This patch changed get_scylla_dirs() to return default data/commitlog
directories if they aren't set.

/CC @xemul @vladzcloudius

### DIFF
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -471,13 +471,15 @@ def get_scylla_dirs():
     y = yaml.safe_load(open(scylla_yaml_name))
 
     # Check that mandatory fields are set
+    if 'workdir' not in y or not y['workdir']:
+        y['workdir'] = '/var/lib/scylla'
     if 'data_file_directories' not in y or \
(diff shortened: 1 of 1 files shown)


bhalevy:
@xemul your [comment](https://github.com/scylladb/scylla/commit/217055aca5d392cd6d229e968a47f31f9e281a0a#commitcomment-36154754) on 217055aca5d392cd6d229e968a47f31f9e281a0a is going to be lost in the noise...
I suggest you review this PR and give it your blessing formally.

amoskong:
### Test:
I tested this PR manually by applying this change to an problem AMI, then the scylla-ami-setup can complete successfully.

penberg:
Merged, thanks for analyzing and fixing the bug, @amoskong!

mykaul:
I assume we are not going to backport it to anything right now.

PR #5365, opened 2019-11-27
gossip: make sleeps react to shutdown signal

### LINKED ISSUE #3749: gossip: Make the sleep code react to stop
There are multiple usage of sleep in the gossip code base. The sleep lasts for 30s by default which will hang the stop operation.

piodul:
This change makes most sleeps in gossip.cc abortable. It is now possible to quickly shut down a node during startup, most notably during the phase while it waits for gossip to settle.

Fixes #3749.

Tests: unit(dev)

### DIFF
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -125,8 +125,9 @@ public:
     void on_restart(inet_address, endpoint_state) override {}
 };
 
-gossiper::gossiper(feature_service& features, db::config& cfg)
-        : _feature_service(features)
+gossiper::gossiper(abort_source& as, feature_service& features, db::config& cfg)
+        : _abort_source(as)
+        , _feature_service(features)
         , _cfg(cfg)
         , _fd(cfg.phi_convict_threshold()) {
     // Gossiper's stuff below runs only on CPU0
@@ -1072,7 +1073,7 @@ future<> gossiper::advertise_removing(inet_address endpoint, utils::UUID host_id
         logger.info("Removing host: {}", host_id);
         auto ring_delay = std::chrono::milliseconds(_cfg.ring_delay_ms());
         logger.info("Sleeping for {}ms to ensure {} does not change", ring_delay.count(), endpoint);
-        sleep(ring_delay).get();
+        sleep_abortable(ring_delay, _abort_source).get();
         // make sure it did not change
         auto& eps = get_endpoint_state(endpoint);
         if (eps.get_heart_beat_state().get_generation() != generation) {
@@ -1102,7 +1103,7 @@ future<> gossiper::advertise_token_removed(inet_address endpoint, utils::UUID ho
         endpoint_state_map[endpoint] = eps;
         replicate(endpoint, eps).get();
         // ensure at least one gossip round occurs before returning
-        sleep(INTERVAL * 2).get();
+        sleep_abortable(INTERVAL * 2, _abort_source).get();
     });
 }
 
@@ -1136,7 +1137,7 @@ future<> gossiper::assassinate_endpoint(sstring address) {
                 auto ring_delay = std::chrono::milliseconds(gossiper._cfg.ring_delay_ms());
                 logger.info("Sleeping for {} ms to ensure {} does not change", ring_delay.count(), endpoint);
                 // make sure it did not change
-                sleep(ring_delay).get();
+                sleep_abortable(ring_delay, gossiper._abort_source).get();
 
                 es = gossiper.get_endpoint_state_for_endpoint_ptr(endpoint);
                 if (!es) {
@@ -1158,7 +1159,7 @@ future<> gossiper::assassinate_endpoint(sstring address) {
             auto expire_time = gossiper.compute_expire_time();
             ep_state.add_application_state(application_state::STATUS, gossiper._value_factory.left(tokens_set, expire_time.time_since_epoch().count()));
             gossiper.handle_major_state_change(endpoint, ep_state);
-            sleep(INTERVAL * 4).get();
+            sleep_abortable(INTERVAL * 4, gossiper._abort_source).get();
             logger.warn("Finished assassinating {}", endpoint);
         });
     });
@@ -1773,7 +1774,7 @@ future<> gossiper::do_shadow_round() {
                     logger.trace("Fail to send GossipDigestSyn (ShadowRound) to {}: {}", id, ep);
                 });
             }
-            sleep(std::chrono::seconds(1)).get();
+            sleep_abortable(std::chrono::seconds(1), _abort_source).get();
             if (this->_in_shadow_round) {
                 if (clk::now() > t + std::chrono::milliseconds(_cfg.shadow_round_ms())) {
                     throw std::runtime_error(format("Unable to gossip with any seeds (ShadowRound)"));
@@ -2112,9 +2113,9 @@ future<> gossiper::wait_for_gossip(std::chrono::milliseconds initial_delay, std:
 
         auto delay = initial_delay;
 
-        sleep(GOSSIP_SETTLE_MIN_WAIT_MS).get();
+        sleep_abortable(GOSSIP_SETTLE_MIN_WAIT_MS, _abort_source).get();
         while (num_okay < GOSSIP_SETTLE_POLL_SUCCESSES_REQUIRED) {
-            sleep(delay).get();
+            sleep_abortable(delay, _abort_source).get();
             delay = GOSSIP_SETTLE_POLL_INTERVAL_MS;
 
             int32_t current_size = endpoint_state_map.size();
--- a/gms/gossiper.hh
+++ b/gms/gossiper.hh
@@ -62,6 +62,7 @@
 #include <set>
 #include <seastar/core/condition-variable.hh>
 #include <seastar/core/metrics_registration.hh>
+#include <seastar/core/abort_source.hh>
 
 namespace db {
 class config;
@@ -259,7 +260,7 @@ private:
     // The value must be kept alive until completes and not change.
     future<> replicate(inet_address, application_state key, const versioned_value& value);
 public:
-    explicit gossiper(feature_service& features, db::config& cfg);
+    explicit gossiper(abort_source& as, feature_service& features, db::config& cfg);
 
     void set_last_processed_message_at();
     void set_last_processed_message_at(clk::time_point tp);
@@ -592,6 +593,7 @@ private:
 
     class msg_proc_guard;
 private:
+    abort_source& _abort_source;
     condition_variable _features_condvar;
     feature_service& _feature_service;
     db::config& _cfg;
--- a/main.cc
+++ b/main.cc
@@ -668,7 +668,7 @@ int main(int ac, char** av) {
             (void)cql_config_updater.start(std::ref(cql_config), std::ref(*cfg));
             auto stop_cql_config_updater = defer([&] { cql_config_updater.stop().get(); });
             auto& gossiper = gms::get_gossiper();
-            gossiper.start(std::ref(feature_service), std::ref(*cfg)).get();
+            gossiper.start(std::ref(stop_signal.as_sharded_abort_source()), std::ref(feature_service), std::ref(*cfg)).get();
             // #293 - do not stop anything
             //engine().at_exit([]{ return gms::get_gossiper().stop(); });
             supervisor::notify("initializing storage service");
--- a/test/boost/gossip_test.cc
+++ b/test/boost/gossip_test.cc
@@ -67,7 +67,7 @@ SEASTAR_TEST_CASE(test_boot_shutdown){
         netw::get_messaging_service().start(gms::inet_address("127.0.0.1"), 7000, false /* don't bind */).get();
         auto stop_messaging_service = defer([&] { netw::get_messaging_service().stop().get(); });
 
-        gms::get_gossiper().start(std::ref(feature_service), std::ref(cfg)).get();
+        gms::get_gossiper().start(std::ref(abort_sources), std::ref(feature_service), std::ref(cfg)).get();
         auto stop_gossiper = defer([&] { gms::get_gossiper().stop().get(); });
 
         service::storage_service_config sscfg;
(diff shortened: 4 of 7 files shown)


gms/gossiper.hh
```diff
@@ -548,7 +548,6 @@ public:
     future<> add_local_application_state(std::initializer_list<std::pair<application_state, utils::in<versioned_value>>>);
 
     // Needed by seastar::sharded
-    future<> stop();
```
tgrabiec: You should also remove `// Needed by seastar::sharded`, which applied top `stop()` only.

gms/gossiper.cc
```diff
@@ -125,10 +125,11 @@ class feature_enabler : public i_endpoint_state_change_subscriber {
     void on_restart(inet_address, endpoint_state) override {}
 };
 
-gossiper::gossiper(feature_service& features, db::config& cfg)
-        : _feature_service(features)
+gossiper::gossiper(feature_service& features, db::config& cfg, abort_source& as)
```
tgrabiec: I'm not sure interrupting all services at once is the right approach in general.

Shouldn't we rather have the stop operation (`gossiper::stop_gossiping()`) trigger the interruption of internal operations, on a private abort_source? So that we don't interrupt anything until it's really time to stop the component. Just theorizing now, there could be another component which is relying on the gossiper to complete its (the component's) operations, so we should first let that component to stop itself (which may want to wait for those operations to complete), and only then stop the gossiper.

piodul: We already use a similar approach in `storage_proxy.cc` - take global `abort_source`, and use `sleep_abortable` with it everywhere.

I can see that all paths that pass through `sleep_abortable`s from gossiper.cc are called either from HTTP API (directly or indirectly), or from initialization. When initialization is aborted, it is desirable to stop waiting. As for HTTP API, it will just fail the request if user is unlucky enough to perform it during shutdown.

gms/gossiper.hh
piodul: Removed the comment as well

gms/gossiper.cc
tgrabiec: Do you mean `storage_service`? I don't see abort_source usage in storage_proxy.cc. It's true that storage_service uses this approach, I still think it's a bad approach in general.

It's also true that this approach is not currently harmful given what operations it affects.

tgrabiec: Another argument for the alternative approach is that we want to stop the gossiper operations not only on ^C, but also when exception interrupts the boot process. Making `stop()` trigger abort handles both cases.

piodul: Sorry, I meant `storage_service.cc`, my mistake.

I agree with you argument, I remade aborting so as to abort on a private `abort_source`. I left aborting on Ctrl+C `abort_source` in initialization code.

gms/gossiper.cc
```diff
@@ -1739,15 +1739,16 @@ future<> gossiper::start_gossiping(int generation_nbr, std::map<application_stat
         }).then([this, generation] {
             logger.trace("gossip started with generation {}", generation);
             _enabled = true;
+            _abort_source = abort_source();
```
tgrabiec: Why is this line needed? _abort_source should be already constructed in the constructor by then.

gms/gossiper.cc
```diff
@@ -1133,7 +1133,7 @@ future<> gossiper::assassinate_endpoint(sstring address) {
                 auto ring_delay = std::chrono::milliseconds(gossiper._cfg.ring_delay_ms());
                 logger.info("Sleeping for {} ms to ensure {} does not change", ring_delay.count(), endpoint);
                 // make sure it did not change
-                sleep(ring_delay).get();
+                sleep_abortable(ring_delay, gossiper._abort_source).get();
```
tgrabiec: This sleep is part of the `assassinate_endpoint()` operation (the operation joins with this, it doesn't run in the background). The sleeps in `wait_for_gossip_to_settle()` are also like that. In the latter case we chose the caller to interrupt the operation (by providing an abort_source as an argument). In the former case, we chose stop_gossiping() to interrupt it. We could have let the caller interrupt `assasinate_endpoint()` as well. I wonder why did you chose otherwise.

The idea was that whoever uses the gossiper (having an active foreground operation counts as a use), should be stopped before the gossiper is stopped. And when the gossiper is stopped, it interrupts only operations started by itself, which don't count as external use.

piodul: I converted other methods to take an abort source as well. The only internal task I see is `gossiper::run` and it does not sleep, so we may drop the `abort_source` altogether.

I observed that we do not wait for HTTP requests to finish before stopping the gossiper, so it is possible for somebody to start, for example, an `assasinate` operation while we are shutting down. I'm not sure what to do about it - guard such operations with a gate?

tgrabiec: Before the gossiper is stopped (sharded::stop), the API server should be stopped. Or at least the part of the API which uses the gossiper. Stopping is interrupt (abort_source) + await (gate for tracking).

gms/gossiper.cc
```diff
@@ -1099,18 +1099,18 @@ future<> gossiper::advertise_token_removed(inet_address endpoint, utils::UUID ho
...
-    return assassinate_endpoint(address);
+    return assassinate_endpoint(address, as);
 }
 
-future<> gossiper::assassinate_endpoint(sstring address) {
-    return get_gossiper().invoke_on(0, [address] (auto&& gossiper) {
-        return seastar::async([&gossiper, g = gossiper.shared_from_this(), address] {
+future<> gossiper::assassinate_endpoint(sstring address, abort_source& as) {
+    return get_gossiper().invoke_on(0, [address, &as] (auto&& gossiper) {
+        return seastar::async([&gossiper, g = gossiper.shared_from_this(), address, &as] {
```
tgrabiec: `abort_source` is not safe to use across shards (across threads). It's a shard-local object.

invoke_on() may execute the lambda on a different shard if the current shard is not 0.

api/gossiper.cc
```diff
@@ -57,12 +57,13 @@ void set_gossiper(http_context& ctx, routes& r) {
     });
 
     httpd::gossiper_json::assassinate_endpoint.set(r, [](std::unique_ptr<request> req) {
+        abort_source as;
```
nyh: This variable is on the stack, below you pass it by reference to asynchronous functions which keep it after this function returns a future. I think you need to use some existing abort source, or do_with()?

gms/gossiper.cc
piodul: Proper shutdown of services seems to be a large topic (https://github.com/scylladb/scylla/issues/293). I think this is a topic too big to be handled here. I tried calling `sharded::stop` on the gossip service, but some other services still tried to use it during shutdown, which caused the nodes on my local cluster to crash.

I don't see a way to unregister a subset of HTTP routes, so I added a gate for gossiper methods that face the HTTP API. The gate is closed during shutdown operation (via `defer_verbose_shutdown`). Those routes will just return an error in case they are called during shutdown.

gms/gossiper.cc
piodul: I removed the `abort_source&` argument from the function. It wouldn't be used anyway (as we don't handle cancellation of HTTP requests), and I can't think of a solution for passing the abort source that would look nice.

piodul:
Additional change: I removed the commit that gets rid of `gossiper::stop` function. Although this function is not necessary, it looks like it is a convention for services, so I would like to keep it.

gms/gossiper.cc
```diff
@@ -1064,45 +1064,49 @@ future<> gossiper::replicate(inet_address ep, application_state key, const versi
...
-        auto ring_delay = std::chrono::milliseconds(_cfg.ring_delay_ms());
-        logger.info("Sleeping for {}ms to ensure {} does not change", ring_delay.count(), endpoint);
-        sleep(ring_delay).get();
-        // make sure it did not change
-        auto& eps = get_endpoint_state(endpoint);
-        if (eps.get_heart_beat_state().get_generation() != generation) {
-            throw std::runtime_error(format("Endpoint {} generation changed while trying to remove it", endpoint));
-        }
+future<> gossiper::advertise_removing(inet_address endpoint, utils::UUID host_id, utils::UUID local_host_id, abort_source& as) {
+    return seastar::with_gate(_gate, [this, g = this->shared_from_this(), endpoint, host_id, local_host_id, &as] {
```
asias: You can use a half indentation to reduce the changes and restore the indentation later.

gms/gossiper.cc
```diff
@@ -1064,45 +1064,49 @@ future<> gossiper::replicate(inet_address ep, application_state key, const versi
...
-        auto expire_time = compute_expire_time();
-        eps.add_application_state(application_state::STATUS, _value_factory.removed_nonlocal(host_id, expire_time.time_since_epoch().count()));
-        logger.info("Completing removal of {}", endpoint);
-        add_expire_time_for_endpoint(endpoint, expire_time);
-        endpoint_state_map[endpoint] = eps;
-        replicate(endpoint, eps).get();
-        // ensure at least one gossip round occurs before returning
-        sleep(INTERVAL * 2).get();
+future<> gossiper::advertise_token_removed(inet_address endpoint, utils::UUID host_id, abort_source& as) {
+    return seastar::with_gate(_gate, [this, g = this->shared_from_this(), endpoint, host_id, &as] {
```
asias: Same here.

gms/gossiper.cc
```diff
@@ -1112,54 +1116,56 @@ future<> gossiper::unsafe_assassinate_endpoint(sstring address) {
...
+                        throw std::runtime_error(format("Unable to calculate tokens for {}", endpoint));
+                    }
 
-                int generation = ep_state.get_heart_beat_state().get_generation();
-                int heartbeat = ep_state.get_heart_beat_state().get_heart_beat_version();
-                auto ring_delay = std::chrono::milliseconds(gossiper._cfg.ring_delay_ms());
-                logger.info("Sleeping for {} ms to ensure {} does not change", ring_delay.count(), endpoint);
-                // make sure it did not change
-                sleep(ring_delay).get();
+                    int generation = ep_state.get_heart_beat_state().get_generation();
```
asias: Extra white space?

piodul:
@asias I used half-indentation as you suggested, and fixed whitespace in a separate commit.

main.cc
```diff
@@ -962,7 +962,7 @@ int main(int ac, char** av) {
             //FIXME: discarded future
             (void)api::set_server_cache(ctx);
             startlog.info("Waiting for gossip to settle before accepting client requests...");
-            gms::get_local_gossiper().wait_for_gossip_to_settle().get();
+            gms::get_local_gossiper().wait_for_gossip_to_settle(stop_signal.as_local_abort_source()).get();
```
asias: In  storage_service class, we have 
 
   abort_source& _abort_source

member. Why not do this to gossiper class as well so we can avoid passing abort_source to all the calls.

gms/gossiper.cc
```diff
@@ -1065,7 +1065,8 @@ future<> gossiper::replicate(inet_address ep, application_state key, const versi
 }
 
 future<> gossiper::advertise_removing(inet_address endpoint, utils::UUID host_id, utils::UUID local_host_id, abort_source& as) {
-    return seastar::async([this, g = this->shared_from_this(), endpoint, host_id, local_host_id, &as] {
+  return seastar::with_gate(_gate, [this, g = this->shared_from_this(), endpoint, host_id, local_host_id, &as] {
```
asias: class gossiper inherits from class seastar::async_sharded_service. This guarantees sharded::stop() will wait until all the references disappear before returning. This looks redundant to me. In general I like using gate to wait for the existing and prevent further operations, but it is not enough just for those functions you changed here. 

So, I would recommend to contain only the abort source changes in this PR so we can get it merged faster.

asias:
> @asias I used half-indentation as you suggested, and fixed whitespace in a separate commit.

Thanks. It is much easier to review now!

main.cc
piodul: The idea is to let the caller of the operation supply an `abort_source`, so that they control when the operation is cancelled. Initially, I did it in the way you described, but changed after comments from @tgrabiec.

gms/gossiper.cc
piodul: OK, I dropped the last two commits (related to the gate).

main.cc
asias: There is only one caller that is storage_service.cc. So every abort source passed to gossiper is from storage_service which in turn from the object from main.cc. What's the point to make storage_service to pass abort source?

slivne:
@asias ping - please approve if its ok now

asias:
> @asias ping - please approve if its ok now

We have pending discussion on passing the abort source to gossip object instead of passing to gossip member functions everywhere. 

@piodul I do not see practical benefits to passing abort source to gossip member functions everywhere.

piodul:
@asias I changed the PR so that the gossiper holds a reference to the abort source in a field. All but one sleeps are now abortable. I left one non-abortable sleep in `do_stop_gossiping` because it is on a path that may run at shutdown and it seemed intentional.

asias:
Thanks @piodul . It looks good to me.

slivne:
@tgrabiec / @nyh - comments seems to been have handled - @asias blessed this - please re-review

nyh:
Looks good to me, but there is one open discussion with @tgrabiec about details of the shutdown order that I didn't understand enough to be able to judge whether this discussion was resolved or not. Tomek?

main.cc
tgrabiec: @asias In general, having one abort source per component is inflexible.

First, we may want to interrupt operations not only when the user hits ^C. If an operation on gossiper is part of a background operation started by storage_service, we will want to interrupt that operation when storage_service::stop() is called. When we fail to boot with an exception we will trigger a cascade of stops on each component (according to the depenency chain), without the main `abort_source` being triggered. 

Another problem is that, in general, interrupting all operations from all components at once may not be correct due to dependencies. If component A depends on B, then A needs to be stopped before B is stopped. If you interrupt B before stopping A, then A may not be able to stop gracefully. B::stop() should interrupt background operations managed by B, only after A stopped itself.

Those are theoretical problems though for now, so I committed the patch.

PR #5366, opened 2019-11-27
CDC: handle schema changes

elcallio:
Moves schema creation/alter/drop awareness to use new "before" callbacks from migration manager, and adds/modifies log and streams table as part of the base table modification. 

Makes schema changes semi-atomic per node. While this does not deal with updates coming in before a schema change has propagated cluster, it now falls into the same pit as when this happens without CDC. 

Added side effect is also that now schemas are transparent across all subsystems, not just cql. 

v2:
Pushed new version addressing the mis-squash for cdc.hh
v3:
Rebased on master
v4:
Added two asserts.
v5:
removed "use namespace seastar" and added include "seastarx.h"
v6: 
removed asserts again (gah), and simplified condition in favour of comment

### DIFF
--- a/cdc/cdc.cc
+++ b/cdc/cdc.cc
@@ -64,6 +64,24 @@ using namespace std::chrono_literals;
 
 static logging::logger cdc_log("cdc");
 
+class cdc::cdc_service::impl {
+    db_context _ctxt;
+public:
+    impl(db_context ctxt)
+        : _ctxt(std::move(ctxt))
+    {}
+};
+
+cdc::cdc_service::cdc_service(service::storage_proxy& proxy)
+    : cdc_service(db_context::builder(proxy).build())
+{}
+
+cdc::cdc_service::cdc_service(db_context ctxt)
+    : _impl(std::make_unique<impl>(std::move(ctxt)))
+{}
+
+cdc::cdc_service::~cdc_service() = default;
+
 cdc::options::options(const std::map<sstring, sstring>& map) {
     if (map.find("enabled") == std::end(map)) {
         throw exceptions::configuration_exception("Missing enabled CDC option");
--- a/cdc/cdc.cc
+++ b/cdc/cdc.cc
@@ -35,6 +35,7 @@
 #include "schema.hh"
 #include "schema_builder.hh"
 #include "service/migration_manager.hh"
+#include "service/migration_listener.hh"
 #include "service/storage_service.hh"
 #include "types/tuple.hh"
 #include "cql3/statements/select_statement.hh"
@@ -64,12 +65,119 @@ using namespace std::chrono_literals;
 
 static logging::logger cdc_log("cdc");
 
-class cdc::cdc_service::impl {
+namespace cdc {
+static schema_ptr create_log_schema(const schema&, std::optional<utils::UUID> = {});
+static schema_ptr create_stream_description_table_schema(const schema&, std::optional<utils::UUID> = {});
+static future<> populate_desc(db_context ctx, const schema& s);
+}
+
+class cdc::cdc_service::impl : service::migration_listener::empty_listener {
     db_context _ctxt;
 public:
     impl(db_context ctxt)
         : _ctxt(std::move(ctxt))
-    {}
+    {
+        if (engine().cpu_id() == 0) {
+            _ctxt._migration_manager.register_listener(this);
+        }
+    }
+    ~impl() {
+        if (engine().cpu_id() == 0) {
+            _ctxt._migration_manager.unregister_listener(this);
+        }
+    }
+
+    void on_before_create_column_family(const schema& schema, std::vector<mutation>& mutations, api::timestamp_type timestamp) override {
+        if (schema.cdc_options().enabled()) {
+            auto& db = _ctxt._proxy.get_db().local();
+            auto logname = log_name(schema.cf_name());
+            if (!db.has_schema(schema.ks_name(), logname)) {
+                // in seastar thread
+                auto log_schema = create_log_schema(schema);
+                auto stream_desc_schema = create_stream_description_table_schema(schema);
+                auto& keyspace = db.find_keyspace(schema.ks_name());
+
+                auto log_mut = db::schema_tables::make_create_table_mutations(keyspace.metadata(), log_schema, timestamp);
+                auto stream_mut = db::schema_tables::make_create_table_mutations(keyspace.metadata(), stream_desc_schema, timestamp);
+
+                mutations.insert(mutations.end(), std::make_move_iterator(log_mut.begin()), std::make_move_iterator(log_mut.end()));
+                mutations.insert(mutations.end(), std::make_move_iterator(stream_mut.begin()), std::make_move_iterator(stream_mut.end()));
+            }
+        }
+    }
+
+    void on_before_update_column_family(const schema& new_schema, const schema& old_schema, std::vector<mutation>& mutations, api::timestamp_type timestamp) override {
+        bool is_cdc = new_schema.cdc_options().enabled();
+        bool was_cdc = old_schema.cdc_options().enabled();
+
+        // we need to create or modify the log & stream schemas iff either we changed cdc status (was != is)
+        // or if cdc is on now unconditionally, since then any actual base schema changes will affect the column 
+        // etc.
+        if (was_cdc || is_cdc) {
+            auto logname = log_name(old_schema.cf_name());
+            auto descname = desc_name(old_schema.cf_name());
+            auto& db = _ctxt._proxy.get_db().local();
+            auto& keyspace = db.find_keyspace(old_schema.ks_name());
+            auto log_schema = was_cdc ? db.find_column_family(old_schema.ks_name(), logname).schema() : nullptr;
+            auto stream_desc_schema = was_cdc ? db.find_column_family(old_schema.ks_name(), descname).schema() : nullptr;
+
+            if (!is_cdc) {
+                auto log_mut = db::schema_tables::make_drop_table_mutations(keyspace.metadata(), log_schema, timestamp);
+                auto stream_mut = db::schema_tables::make_drop_table_mutations(keyspace.metadata(), stream_desc_schema, timestamp);
+
+                mutations.insert(mutations.end(), std::make_move_iterator(log_mut.begin()), std::make_move_iterator(log_mut.end()));
+                mutations.insert(mutations.end(), std::make_move_iterator(stream_mut.begin()), std::make_move_iterator(stream_mut.end()));
+                return;
+            }
+
+            auto new_log_schema = create_log_schema(new_schema, log_schema ? std::make_optional(log_schema->id()) : std::nullopt);
+            auto new_stream_desc_schema = create_stream_description_table_schema(new_schema, stream_desc_schema ? std::make_optional(stream_desc_schema->id()) : std::nullopt);
+
+            auto log_mut = log_schema 
+                ? db::schema_tables::make_update_table_mutations(keyspace.metadata(), log_schema, new_log_schema, timestamp, false)
+                : db::schema_tables::make_create_table_mutations(keyspace.metadata(), new_log_schema, timestamp)
+                ;
+            auto stream_mut = stream_desc_schema 
+                ? db::schema_tables::make_update_table_mutations(keyspace.metadata(), stream_desc_schema, new_stream_desc_schema, timestamp, false)
+                : db::schema_tables::make_create_table_mutations(keyspace.metadata(), new_stream_desc_schema, timestamp)
+                ;
+
+            mutations.insert(mutations.end(), std::make_move_iterator(log_mut.begin()), std::make_move_iterator(log_mut.end()));
+            mutations.insert(mutations.end(), std::make_move_iterator(stream_mut.begin()), std::make_move_iterator(stream_mut.end()));
+        }
+    }
+
+    void on_before_drop_column_family(const schema& schema, std::vector<mutation>& mutations, api::timestamp_type timestamp) override {
+        if (schema.cdc_options().enabled()) {
+            auto logname = log_name(schema.cf_name());
+            auto descname = desc_name(schema.cf_name());
+            auto& db = _ctxt._proxy.get_db().local();
+            auto& keyspace = db.find_keyspace(schema.ks_name());
+            auto log_schema = db.find_column_family(schema.ks_name(), logname).schema();
+            auto stream_desc_schema = db.find_column_family(schema.ks_name(), descname).schema();
+
+            auto log_mut = db::schema_tables::make_drop_table_mutations(keyspace.metadata(), log_schema, timestamp);
+            auto stream_mut = db::schema_tables::make_drop_table_mutations(keyspace.metadata(), stream_desc_schema, timestamp);
+
+            mutations.insert(mutations.end(), std::make_move_iterator(log_mut.begin()), std::make_move_iterator(log_mut.end()));
+            mutations.insert(mutations.end(), std::make_move_iterator(stream_mut.begin()), std::make_move_iterator(stream_mut.end()));
+        }
+    }
+
+    void on_create_column_family(const sstring& ks_name, const sstring& cf_name) override {
+        auto& db = _ctxt._proxy.get_db().local();
+        auto& cf = db.find_column_family(ks_name, cf_name);
+        auto schema = cf.schema();
+        if (schema->cdc_options().enabled()) {
+            populate_desc(_ctxt, *schema).get();
+        }
+    }
+
+    void on_update_column_family(const sstring& ks_name, const sstring& cf_name, bool columns_changed) override {
+        on_create_column_family(ks_name, cf_name);
+    }
+
+    void on_drop_column_family(const sstring& ks_name, const sstring& cf_name) override {}
 };
 
 cdc::cdc_service::cdc_service(service::storage_proxy& proxy)
@@ -137,39 +245,7 @@ sstring desc_name(const sstring& table_name) {
     return table_name + cdc_desc_suffix;
 }
 
-static future<>
-remove_log(db_context ctx, const sstring& ks_name, const sstring& table_name) {
-    try {
-        return ctx._migration_manager.announce_column_family_drop(
-                ks_name, log_name(table_name), false);
-    } catch (exceptions::configuration_exception& e) {
-        // It's fine if the table does not exist.
-        return make_ready_future<>();
-    } catch (...) {
-        return make_exception_future<>(std::current_exception());
-    }
-}
-
-static future<>
-remove_desc(db_context ctx, const sstring& ks_name, const sstring& table_name) {
-    try {
-        return ctx._migration_manager.announce_column_family_drop(
-                ks_name, desc_name(table_name), false);
-    } catch (exceptions::configuration_exception& e) {
-        // It's fine if the table does not exist.
-        return make_ready_future<>();
-    } catch (...) {
-        return make_exception_future<>(std::current_exception());
-    }
-}
-
-future<>
-remove(db_context ctx, const sstring& ks_name, const sstring& table_name) {
-    return when_all(remove_log(ctx, ks_name, table_name),
-                    remove_desc(ctx, ks_name, table_name)).discard_result();
-}
-
-static future<> setup_log(db_context ctx, const schema& s) {
+static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID> uuid) {
     schema_builder b(s.ks_name(), log_name(s.cf_name()));
     b.set_default_time_to_live(gc_clock::duration{s.cdc_options().ttl()});
     b.set_comment(sprint("CDC log for %s.%s", s.ks_name(), s.cf_name()));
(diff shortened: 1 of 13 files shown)


tgrabiec:
Regarding:

> While this does not deal with updates coming in before a schema change has propagated cluster, it now falls into the same pit as when this happens without CDC.

We implement causal consistency for writes and their schema, meaning that if a write arriving at the replica was generated using schema version S1, we make sure that node's schema is at least as recent as S1 before applying it (by pulling and syncing from the source if necessary).

That's done by service::get_schema_for_write(), called by RPC write handlers (storage_proxy::init_messaging_service())

haaawk:
LGTM in general. Modulo a little bug in condition for update schema.

cdc/cdc.cc
```diff
@@ -64,12 +65,116 @@ using namespace std::chrono_literals;
...
 static logging::logger cdc_log("cdc");
 
-class cdc::cdc_service::impl {
+namespace cdc {
+static schema_ptr create_log_schema(const schema&, std::optional<utils::UUID> = {});
+static schema_ptr create_stream_description_table_schema(const schema&, std::optional<utils::UUID> = {});
+static future<> populate_desc(db_context ctx, const schema& s);
+}
+
+class cdc::cdc_service::impl : service::migration_listener::empty_listener {
```
kbr-: I would like to see a textual description of what this service does, which replaces the documentation that `cdc::setup` and `cdc::remove` functions previously had.

cdc/cdc_options.hh
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+#pragma once
+
+#include <map>
+#include <seastar/core/sstring.hh>
+
+namespace cdc {
+
+using namespace seastar;
```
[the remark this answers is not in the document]
denesb: Unfortunately all our code is written assuming `using namespace sestar`. This is a relic of our history, from the past where `seastar` wasn't yet a separate framework, just a part of the Scylla codebase.

service/migration_listener.hh
```diff
@@ -74,25 +80,43 @@ public:
     virtual void on_drop_aggregate(const sstring& ks_name, const sstring& aggregate_name) = 0;
     virtual void on_drop_view(const sstring& ks_name, const sstring& view_name) = 0;
 
+    // The callback runs inside seastar thread
+    // called before adding/updating/dropping column family. 
+    // listener can add additional type altering mutations if he knows what he is doing. 
+    virtual void on_before_create_column_family(const schema&, std::vector<mutation>&, api::timestamp_type) {}
```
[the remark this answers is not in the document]
elcallio: That breaks all other extensions of the type and requires changes to unrelated files.

service/migration_manager.cc
```diff
@@ -761,35 +778,46 @@ future<> migration_manager::announce_column_family_drop(const sstring& ks_name,
...
-        }
-        auto mutations = db::schema_tables::make_drop_table_mutations(keyspace, schema, api::new_timestamp());
-        mutations.insert(mutations.end(), std::make_move_iterator(drop_si_mutations.begin()), std::make_move_iterator(drop_si_mutations.end()));
-        for (auto& v : views) {
-            if (!old_cfm.get_index_manager().is_index(v)) {
-                mlogger.info("Drop view '{}.{}' of table '{}'", v->ks_name(), v->cf_name(), schema->cf_name());
-                auto m = db::schema_tables::make_drop_view_mutations(keyspace, v, api::new_timestamp());
-                mutations.insert(mutations.end(), std::make_move_iterator(m.begin()), std::make_move_iterator(m.end()));
+
+        return seastar::async([this, keyspace, schema, &old_cfm, drop_views] {
```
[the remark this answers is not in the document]
elcallio: Because it makes the code more readable and has no ill effects on its function.

cdc/cdc.cc
elcallio: Sure.

cdc/cdc.cc
```diff
@@ -64,6 +64,24 @@ using namespace std::chrono_literals;
...
+class cdc::cdc_service::impl {
+    db_context _ctxt;
+public:
+    impl(db_context ctxt)
+        : _ctxt(std::move(ctxt))
+    {}
+};
+
+cdc::cdc_service::cdc_service(service::storage_proxy& proxy)
+    : cdc_service(db_context::builder(proxy).build())
```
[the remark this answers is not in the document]
elcallio: Again, I would prefer to hang the service as as subobject of something like storage proxy, which would also make things like caching more useful (not global static cache). There you would assume creating the object using all required params. This is just stopgap stuff. Does not make anything worse or better.

elcallio:
> 
> 
> > semi-atomic per node
> 
> Does this mean "semi-atomic, i.e. atomic per node", or "even within one node, semi-atomic"?
> If the latter, what does it mean to be semi-atomic?
> 
> Is the vector of mutations that does the schema change applied atomically or not? If not, is there a window where the base table is changed but the log table is not?

No, the schema changes are "atomic" per shard, as in they are applied together and uninterruptible. However, the "populate" part of schema management is not executed until schema change notification, which can defer (though no changes can happen inbetween), so we could have a write come in before populate is done.

service/migration_manager.cc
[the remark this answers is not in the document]
kbr-: How is it more readable? Now I'm wondering if any of this stuff executes asynchronous operations so this fiber can yield.

kbr-:
> so we could have a write come in before populate is done.

If I understand correctly, by "populate" you mean what happens in `on_create_column_family`, which only inserts stuff into the description table? And it shouldn't be a problem if it happens later. (after we introduce topology change support the streams table will be independent of any particular base table anyway)

service/migration_manager.cc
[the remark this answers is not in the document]
elcallio: Of course it does. Even the listener callbacks are allowed to yield. Not to mention that the map-reduce certainly is. At the end of the day, it is a matter of taste I guess, but doing the map-reduce inside async call did at least to me look nicer in the editor than doing then(map_reduce).then(async(...)).then(...)

Honestly, the whole migrate call could probably be inside an async, but I did not want to change the overall behaviour calling out until 100% verified that it is kosher. So path of least resistance.

elcallio:
If I am not mistaken the current code sort of assumes the description table is populated when wrapping an update. But right now (and not because of this change), someone could do a schema create and a send a write to the schema before getting a create ACK, and we might try to augment this without populate being called. 
My only point was that description management should eventually not have this restriction, because this is the only lag between "schema exists so write can be done" and "desc is fully set up".

service/migration_manager.cc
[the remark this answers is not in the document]
kbr-: Yeah, in the case above where map_reduce is used I agree it looks nice and makes sense.

But here (in `announce_column_family_drop`) all the code from start of async to the for loop over listeners, which is most of the code in this async block, is not yielding.

But I'm not going to fight to change it, as you said, it's a matter of taste.

slivne:
@kbr- / @denesb / @haaawk ping - please re-review this

slivne:
@kbr- / @denesb / @haaawk ping - please re-review this

service/migration_manager.cc
```diff
@@ -638,19 +648,26 @@ future<> migration_manager::announce_column_family_update(schema_ptr cfm, bool f
...
-                auto mutations = db::schema_tables::make_update_view_mutations(keyspace, view_ptr(old_view), std::move(view), ts, false);
-                return make_ready_future<std::vector<mutation>>(std::move(mutations));
-            }, std::move(mutations),
-            [] (auto&& result, auto&& view_mutations) {
-                std::move(view_mutations.begin(), view_mutations.end(), std::back_inserter(result));
-                return std::move(result);
-            })
-        .then([keyspace, announce_locally] (auto&& mutations) {
+
+        return seastar::async([this, cfm, old_schema, ts, keyspace, from_thrift, view_updates] {
```
denesb: Would have been better to add these with a half indent and reindent in a later patch.

cdc/cdc.cc
```diff
@@ -63,6 +64,46 @@ using namespace std::chrono_literals;
 
 static logging::logger cdc_log("cdc");
 
+cdc::options::options(const std::map<sstring, sstring>& map) {
```
haaawk: Why not place this below inside namespace cdc and drop cdc::? There's already code there. It would be more consistent to do it in one way.

cdc/cdc.cc
```diff
@@ -64,6 +64,24 @@ using namespace std::chrono_literals;
 
 static logging::logger cdc_log("cdc");
 
+class cdc::cdc_service::impl {
```
haaawk: Why not place this below inside namespace cdc and drop cdc::? There's already code there. It would be more consistent to do it in one way.

cdc/cdc.cc
elcallio: Honestly because I hate code inside namespace{}. It makes it harder to see where I am, and it also makes it uglier whenever you want to group related, but namespace separated functions (or file local statics).

haaawk:
LGTM in general. I left few minor comments.
I would like to make a statement here -> GitHub code reviews mechanism is rubbish.
It messes up the comments so much that I won't blame Calle for missing some of them.

cdc/cdc.cc
haaawk: Ok. Let's then fix it the other way around and get rid of namespace cdc in .cc at all. Not in this patchset though.

elcallio: Woho!

cdc/cdc.hh
```diff
@@ -132,21 +103,7 @@ enum class column_op : int8_t {
...
-/// It's ok if some/all of them don't exist.
-///
-/// \param[in] ctx object with references to database components
-/// \param[in] ks_name keyspace name of a table for which CDC tables are removed
-/// \param[in] table_name name of a table for which CDC tables are removed
-///
-/// \pre This function works correctly no matter if CDC Log and/or CDC Description
-///      exist.
-seastar::future<>
-remove(db_context ctx, const seastar::sstring& ks_name, const seastar::sstring& table_name);
```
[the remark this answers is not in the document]
elcallio: Funny how github kept this comment together with the change even when moved to the correct patch. 
Anyhow being in this patch the above is no longer true.

cdc/cdc.cc
```diff
@@ -63,6 +65,168 @@ using namespace std::chrono_literals;
...
+        if (is_cdc != was_cdc || is_cdc) {
+            auto logname = log_name(old_schema.cf_name());
+            auto descname = desc_name(old_schema.cf_name());
+            auto& db = _ctxt._proxy.get_db().local();
+            auto& keyspace = db.find_keyspace(old_schema.ks_name());
+            auto log_schema = was_cdc ? db.find_column_family(old_schema.ks_name(), logname).schema() : nullptr;
+            auto stream_desc_schema = was_cdc ? db.find_column_family(old_schema.ks_name(), descname).schema() : nullptr;
+
+            if (!is_cdc) {
+                auto log_mut = db::schema_tables::make_drop_table_mutations(keyspace.metadata(), log_schema, timestamp);
```
kbr-: I know it's an invariant that
`cdc was enabled => log schema and desc schema exist`
but what if there's some subtle race or just bug where we break this invariant (maybe just for a short period of time)?

I'd suggest to make a check here that log_schema != nullptr. (i.e., `find_column_family` succeeded). Maybe not an assert, but if + throw runtime_error. Or simply log it with warn level (in `drop` I'd go for log, in `update` I'd go for an error probably).

haaawk:
@elcallio Could you address the final comments please? We would try to get this merged today.

cdc/cdc.cc
elcallio: log_schema etc are always created in the same batch of mutations as base schema. We cannot race two schema updates at the same time (uses semaphore lock), and we are not notified until all changes are done (still inside lock). 
If log_schema was null here with was_cdc=true we should honestly either crash or assert, because there is no valid way we can reach that state _unless_ we had a crash in a run before, between two "db.add_column_family_and_make_directory" calls. 
But sure, an assert might do.

cdc/cdc_options.hh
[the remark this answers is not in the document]
nyh: Indeed. Many of our header files already have "using namespace seastar;" so if you include any one of those, you don't need to do this yourself, but sometimes in new header files which don't include a lot of header files - you do. Even then we recommend to get the "using namespace seastar" by doing "#include "seastarx.hh", not by directly calling "using namespace". All of this is documented in docs/coding-style.md (in fact, it's more or less the only thing documented there, since the rest of the coding style is inherited from Seasatar).

cdc/cdc.cc
```diff
@@ -64,12 +65,119 @@ using namespace std::chrono_literals;
...
+            _ctxt._migration_manager.unregister_listener(this);
+        }
+    }
+
+    void on_before_create_column_family(const schema& schema, std::vector<mutation>& mutations, api::timestamp_type timestamp) override {
+        if (schema.cdc_options().enabled()) {
+            auto& db = _ctxt._proxy.get_db().local();
+            auto logname = log_name(schema.cf_name());
+            if (!db.has_schema(schema.ks_name(), logname)) {
+                // in seastar thread
```
nyh: While this comment is true, why is it relevant? Unless I'm missing something, all of the code below is immediate, and doesn't use get() or anything?

cdc/cdc.cc
```diff
@@ -64,12 +65,119 @@ using namespace std::chrono_literals;
...
+                mutations.insert(mutations.end(), std::make_move_iterator(stream_mut.begin()), std::make_move_iterator(stream_mut.end()));
+            }
+        }
+    }
+
+    void on_before_update_column_family(const schema& new_schema, const schema& old_schema, std::vector<mutation>& mutations, api::timestamp_type timestamp) override {
+        bool is_cdc = new_schema.cdc_options().enabled();
+        bool was_cdc = old_schema.cdc_options().enabled();
+
+        if (is_cdc != was_cdc || is_cdc) {
```
nyh: Unless I'm missing something this condition is the same as if (is_cdc || was_cdc) - but your version is harder to understand.

cdc/cdc.cc
```diff
@@ -64,12 +65,119 @@ using namespace std::chrono_literals;
...
+            auto logname = log_name(old_schema.cf_name());
+            auto descname = desc_name(old_schema.cf_name());
+            auto& db = _ctxt._proxy.get_db().local();
+            auto& keyspace = db.find_keyspace(old_schema.ks_name());
+            auto log_schema = was_cdc ? db.find_column_family(old_schema.ks_name(), logname).schema() : nullptr;
+            auto stream_desc_schema = was_cdc ? db.find_column_family(old_schema.ks_name(), descname).schema() : nullptr;
+
+            assert(!was_cdc || log_schema);
```
nyh: These asserts are rather pointless, considering you defined log_schema just two lines above, to be non-null exactly when was_cdc?

cdc/cdc.cc
nyh: How can log_schema be null? It was defined above to not null if was_cdc... If the find_column_family() above failed, we would get an exception here, not a null.

cdc/cdc.cc
kbr-: It was my suggestion, but now you're saying:

> If the find_column_family() above failed, we would get an exception here, not a null.

So my suggestion was stupid indeed.

cdc/cdc.cc
kbr-: > If the find_column_family() above failed, we would get an exception here, not a null

For some reason I was convinced it would return `nullptr`. Sorry for introducing the confusion

cdc/cdc.cc
elcallio: Irrelevant, but useful in case anyone ever needs to introduce a continuation.

cdc/cdc.cc
elcallio: It is, but imho this expresses better what we are looking for. Of course, had I really wanted to be fully expressive, I would have put a comment, so shame on me.

cdc/cdc.cc
elcallio: pointless, yes, but otoh makes sure noone messes with the assignment. Otoh, I really wish our asserts were debug-only so one could drizzle them more abundantly.

service/migration_listener.hh
```diff
@@ -74,25 +80,43 @@ public:
...
-    virtual void on_update_user_type(const sstring& ks_name, const sstring& type_name) { }
-    virtual void on_update_function(const sstring& ks_name, const sstring& function_name) { }
-    virtual void on_update_aggregate(const sstring& ks_name, const sstring& aggregate_name) { }
-    virtual void on_drop_keyspace(const sstring& ks_name) { }
-    virtual void on_drop_column_family(const sstring& ks_name, const sstring& cf_name) { }
-    virtual void on_drop_user_type(const sstring& ks_name, const sstring& type_name) { }
-    virtual void on_drop_function(const sstring& ks_name, const sstring& function_name) { }
-    virtual void on_drop_aggregate(const sstring& ks_name, const sstring& aggregate_name) { }
+public:
+    void on_create_keyspace(const sstring& ks_name) override {}
```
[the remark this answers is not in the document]
elcallio: In my interpretation of our code style we use virtual for declarations (i.e. first occurrence in hierarchy), and override for overrides/impls. So yes, only cleanup. Because I really hate double decoration.

service/migration_manager.cc
```diff
@@ -617,8 +617,18 @@ future<> migration_manager::announce_new_column_family(schema_ptr cfm, api::time
         }
 
         mlogger.info("Create new ColumnFamily: {}", cfm);
-        auto mutations = db::schema_tables::make_create_table_mutations(keyspace.metadata(), cfm, timestamp);
-        return include_keyspace_and_announce(*keyspace.metadata(), std::move(mutations), announce_locally);
+
+        auto ksm = keyspace.metadata();
+        return seastar::async([this, cfm, timestamp, ksm] {
```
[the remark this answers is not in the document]
elcallio: Because "on_before_create_column_family" is, like all listener functions, declared to be in a seastar::thread (for uniformity). 
It is not really needed in our use case, _yet_, but I prefer to make the change slightly future-proof (pun intended).

service/migration_manager.cc
```diff
@@ -638,19 +648,26 @@ future<> migration_manager::announce_column_family_update(schema_ptr cfm, bool f
...
+                    mlogger.info("Update view '{}.{}' From {} To {}", view->ks_name(), view->cf_name(), *old_view, *view);
+                    auto mutations = db::schema_tables::make_update_view_mutations(keyspace, view_ptr(old_view), std::move(view), ts, false);
+                    return make_ready_future<std::vector<mutation>>(std::move(mutations));
+                }, db::schema_tables::make_update_table_mutations(keyspace, old_schema, cfm, ts, from_thrift),
+                [] (auto&& result, auto&& view_mutations) {
+                    std::move(view_mutations.begin(), view_mutations.end(), std::back_inserter(result));
+                    return std::move(result);
+                }).get0();
+
+            for (auto&& listener : _listeners) {
```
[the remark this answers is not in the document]
elcallio: Maybe, but would have had to deal with keeping mutations in one place, since they are mutable in the callback. As with anything dealing with then(...), it gets ugly and unreadable quickly. And this is not a hot path, so resigning to a seastar::thread for some debuggability feels like a fair trade.

cdc/cdc.cc
nyh: To be honest, I didn't understand what you were looking for in this condition. If it was just "is_cdc != was_cdc", that's one thing - looking for adding or dropping cdc. But when you added "|| is_cdc" you completely changed this condition. Maybe a comment would have helped understand why you wanted to have this complicated way to represent the simple condition was_cdc || is_cdc :-)

cdc/cdc.cc
nyh: Note that asserts cause crashes of the entire node - and if they get caught on multiple nodes - the entire cluster. If we have a bug that effects just a specific operation and not the overall data integrity, it's better to throw an exception and not assert().
And, since in this case they directly follow the assignment, I don't understand the point (yes, someone can change the assignment, but they can change any other line of code and ruin it - we don't put an assert after every statement ;-)).

cdc/cdc.cc
elcallio: fixed

cdc/cdc.cc
elcallio: removed (again)

elcallio:
All outstanding comments should be resolved, either by changes or snarky remarks.

elcallio:
I cannot really close the thread where @nyh asks @tgrabiec  to comment, so interpret that as you wish...

service/migration_manager.cc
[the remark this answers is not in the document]
tgrabiec: @nyh I don't

elcallio:
Le ping? I think all requests/comments have been addressed? If not, please enlighten me.

elcallio:
Sorry, can't click...

haaawk:
@nyh ping

haaawk:
@nyh @avikivity @tgrabiec This PR blocks further work on CDC. Could you please let @elcallio know whether he needs to fix anything or otherwise merge?

slivne:
@tgrabiec / @nyh please review

nyh:
committed to next

kbr-:
Please don't backport this to 3.2 until that bug is fixed: https://github.com/scylladb/scylla/issues/5582

PR #5368, opened 2019-11-28
fix scylla setup of processing default scylla directories

amoskong:
Fixed two problem in scylla_io_setup
- Problem 1: paths of default directories is invalid, introduced by commit 5ec191536e  /CC @bhalevy 
- Problem 2: wrong path join, introduced by commit 31ddb2145a /CC @syuu1228 

scylla_io_setup, scylla_fstrim and scylla_blocktune.py:
- Fixed default scylla directories when they aren't assigned in scylla.yaml

### DIFF
--- a/dist/common/scripts/scylla_io_setup
+++ b/dist/common/scripts/scylla_io_setup
@@ -101,8 +101,8 @@ if __name__ == "__main__":
                 key = "%s_directory" % t
                 if key in cfg:
                     data_dirs += [ cfg[key] ]
-                elif os.path.isdir(default_path + key):
-                    data_dirs += [ default_path + key ]
+                elif os.path.isdir(os.path.join(default_path, t)):
+                    data_dirs += [ os.path.join(default_path, t) ]
 
             iotune_args = []
             for data_dir in data_dirs:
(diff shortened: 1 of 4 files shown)


amoskong:
Hi @xemul 

scylla-setup in AMI works well, but scylla_setup doesn't work if we executed manually.
We have different code path for AMI or non-AMI.

dist/common/scripts/scylla_io_setup
```diff
@@ -91,12 +91,15 @@ if __name__ == "__main__":
             else:
                 conf_dir = etcdir() + "/scylla"
             cfg = yaml.safe_load(open(os.path.join(conf_dir, "scylla.yaml")))
+
+            if 'workdir' not in cfg or not cfg['workdir']:
+                default_path = datadir()
+            else:
+                default_path = cfg['workdir']
```
xemul: This can be rewritten in one line.

    default_path = cfg.get('workdir', '') or datadir()

amoskong: Thanks, a simple one:

```
default_path = cfg.get('workdir') or datadir()
```

xemul:
Ack

penberg:
Queued to next. @amoskong, are any of these fixes need for 3.2 or is only `master` affected?

amoskong:
It fixed some problems introduced to master recently. That won't affect 3.2

PR #5369, opened 2019-11-28
RFC: Add `experimental_features` option

### LINKED ISSUE #5338: Change experimental flag granularity to per feature
The current experimental flag has a granularity of all features. This means that if one of the features is broken it might force the user to turn off the experimental flag altogether, preventing him from using the features that do work.
An example for this situation is #5288 where this solution is suggested, where CDC breaks rolling upgrade from 3.1 to 3.2 forcing turning off the experimental flag and not using LWT since it is also experimental.
The solution suggested in #5288 needs to be implemented, it is roughly outlined in https://github.com/scylladb/scylla/issues/5288#issuecomment-557110306

dekimir:
Add `--experimental-features` -- a vector of features to unlock.  Make corresponding changes in the YAML parser.

Fixes #5338

### DIFF
--- a/conf/scylla.yaml
+++ b/conf/scylla.yaml
@@ -245,7 +245,10 @@ batch_size_fail_threshold_in_kb: 50
 # broadcast_rpc_address: 1.2.3.4
 
 # Uncomment to enable experimental features
-# experimental: true
+# experimental_features:
+#     - cdc
+#     - lwt
+#     - udf
 
 # The directory where hints files are stored if hinted handoff is enabled.
 # hints_directory: /var/lib/scylla/hints
--- a/db/config.cc
+++ b/db/config.cc
@@ -22,6 +22,7 @@
 
 #include <unordered_map>
 #include <regex>
+#include <sstream>
 
 #include <boost/any.hpp>
 #include <boost/program_options.hpp>
@@ -108,6 +109,10 @@ const config_type config_type_for<int32_t> = config_type("integer", value_to_jso
 template <>
 const config_type config_type_for<db::seed_provider_type> = config_type("seed provider", seed_provider_to_json);
 
+template <>
+const config_type config_type_for<std::vector<enum_option<db::experimental_features_t>>> = config_type(
+        "experimental features", value_to_json<std::vector<sstring>>);
+
 }
 
 namespace YAML {
@@ -148,6 +153,23 @@ struct convert<db::config::seed_provider_type> {
                     }
                 }
             }
+        }
+        return true;
+    }
+};
+
+template <>
+class convert<enum_option<db::experimental_features_t>> {
+public:
+    static bool decode(const Node& node, enum_option<db::experimental_features_t>& rhs) {
+        std::string name;
+        if (!convert<std::string>::decode(node, name)) {
+            return false;
+        }
+        try {
+            std::istringstream(name) >> rhs;
+        } catch (boost::program_options::invalid_option_value&) {
+            return false;
         }
         return true;
     }
@@ -671,7 +693,8 @@ db::config::config(std::shared_ptr<db::extensions> exts)
     , shutdown_announce_in_ms(this, "shutdown_announce_in_ms", value_status::Used, 2 * 1000, "Time a node waits after sending gossip shutdown message in milliseconds. Same as -Dcassandra.shutdown_announce_in_ms in cassandra.")
     , developer_mode(this, "developer_mode", value_status::Used, false, "Relax environment checks. Setting to true can reduce performance and reliability significantly.")
     , skip_wait_for_gossip_to_settle(this, "skip_wait_for_gossip_to_settle", value_status::Used, -1, "An integer to configure the wait for gossip to settle. -1: wait normally, 0: do not wait at all, n: wait for at most n polls. Same as -Dcassandra.skip_wait_for_gossip_to_settle in cassandra.")
-    , experimental(this, "experimental", value_status::Used, false, "Set to true to unlock experimental features.")
+    , experimental(this, "experimental", value_status::Used, false, "Set to true to unlock all experimental features.")
+    , experimental_features(this, "experimental_features", value_status::Used, {}, "Unlock experimental features provided as the option arguments (possible values: 'lwt', 'cdc', 'udf'). Can be repeated.")
     , lsa_reclamation_step(this, "lsa_reclamation_step", value_status::Used, 1, "Minimum number of segments to reclaim in a single step")
     , prometheus_port(this, "prometheus_port", value_status::Used, 9180, "Prometheus port, set to zero to disable")
     , prometheus_address(this, "prometheus_address", value_status::Used, "0.0.0.0", "Prometheus listening address")
@@ -706,7 +729,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
         "Set to true if the cluster was initially installed from 3.1.0. If it was upgraded from an earlier version,"
         " or installed from a later version, leave this set to false. This adjusts the communication protocol to"
         " work around a bug in Scylla 3.1.0")
-    , enable_user_defined_functions(this, "enable_user_defined_functions", value_status::Used, false,  "Enable user defined functions")
+    , enable_user_defined_functions(this, "enable_user_defined_functions", value_status::Used, false,  "Enable user defined functions. You must also set experimental-features=udf")
     , user_defined_function_time_limit_ms(this, "user_defined_function_time_limit_ms", value_status::Used, 10, "The time limit for each UDF invocation")
     , user_defined_function_allocation_limit_bytes(this, "user_defined_function_allocation_limit_bytes", value_status::Used, 1024*1024, "How much memory each UDF invocation can allocate")
     , user_defined_function_contiguous_allocation_limit_bytes(this, "user_defined_function_contiguous_allocation_limit_bytes", value_status::Used, 1024*1024, "How much memory each UDF invocation can allocate in one chunk")
@@ -827,10 +850,12 @@ db::fs::path db::config::get_conf_sub(db::fs::path sub) {
     return get_conf_dir() / sub;
 }
 
-void db::config::check_experimental(const sstring& what) const {
-    if (!experimental()) {
-        throw std::runtime_error(format("{} is currently disabled. Start Scylla with --experimental=on to enable.", what));
+bool db::config::check_experimental(experimental_features_t::feature f) const {
+    if (experimental()) {
+        return true;
     }
+    const auto& optval = experimental_features();
+    return find(begin(optval), end(optval), enum_option<experimental_features_t>{f}) != end(optval);
 }
 
 namespace bpo = boost::program_options;
@@ -873,6 +898,12 @@ logging::settings db::config::logging_settings(const bpo::variables_map& map) co
 
 const db::extensions& db::config::extensions() const {
     return *_extensions;
+}
+
+std::unordered_map<sstring, db::experimental_features_t::feature> db::experimental_features_t::map() {
+    // We decided against using the construct-on-first-use idiom here:
+    // https://github.com/scylladb/scylla/pull/5369#discussion_r353614807
+    return {{"lwt", LWT}, {"udf", UDF}, {"cdc", CDC}};
 }
 
 template struct utils::config_file::named_value<seastar::log_level>;
--- a/tests/enum_option_test.cc
+++ b/tests/enum_option_test.cc
@@ -1,0 +1,175 @@
+/*
+ * Copyright (C) 2019 ScyllaDB
+ */
+
+/*
+ * This file is part of Scylla.
+ *
+ * Scylla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Scylla is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#define BOOST_TEST_MODULE core
+
+#include <boost/test/unit_test.hpp>
+
+#include <array>
+#include <boost/program_options.hpp>
+#include <map>
+#include <set>
+#include <sstream>
+#include <string>
+#include <unordered_map>
+
+#include "utils/enum_option.hh"
+
+namespace po = boost::program_options;
+
+namespace {
+
+struct days {
+    enum enumeration { Mo, Tu, We, Th, Fr, Sa, Su };
+    static std::unordered_map<std::string, enumeration> map() {
+        return {{"Mon", Mo}, {"Tue", Tu}, {"Wed", We}, {"Thu", Th}, {"Fri", Fr}, {"Sat", Sa}, {"Sun", Su}};
+    }
+};
+
+template <typename T>
+enum_option<T> parse(const char* value) {
+    po::options_description desc("Allowed options");
+    desc.add_options()("opt", po::value<enum_option<T>>(), "Option");
+    po::variables_map vm;
+    const char* argv[] = {"$0", "--opt", value};
+    po::store(po::parse_command_line(3, argv, desc), vm);
+    return vm["opt"].as<enum_option<T>>();
+}
+
+template <typename T>
+std::string format(typename T::enumeration d) {
+    std::ostringstream os;
+    os << enum_option<T>(d);
+    return os.str();
+}
+
+} // anonymous namespace
+
+BOOST_AUTO_TEST_CASE(test_parsing) {
+    BOOST_CHECK_EQUAL(parse<days>("Sun"), days::Su);
+    BOOST_CHECK_EQUAL(parse<days>("Mon"), days::Mo);
+    BOOST_CHECK_EQUAL(parse<days>("Tue"), days::Tu);
+    BOOST_CHECK_EQUAL(parse<days>("Wed"), days::We);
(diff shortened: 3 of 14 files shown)


tzach:
@dekimir 
Any way to extract the experimental feature available in each release (3 in this case) ?

dekimir:
> @dekimir
> Any way to extract the experimental feature available in each release (3 in this case) ?

Yes: the `--help` flag and the sample `scylla.yaml` file both list them.  That currently has to be maintained by hand, though it's possible to make it automatic in the future with some effort.

utils/enum_option.hh
```diff
@@ -0,0 +1,106 @@
...
+///   return *inst;
+/// }
+/// int main(int ac, char* av[]) {
+///   namespace po = boost::program_options;
+///   po::options_description desc("Allowed options");
+///   desc.add_options()
+///     ("val", po::value<enum_option<Type>>(), "Single Type")
+///     ("vec", po::value<vector<enum_option<Type>>>()->multitoken(), "Type vector");
+/// }
+template<typename Mapper>
```
avikivity: Please document the Mapper type. The best way to do that is with a GCC6_CONCEPT().

conf/scylla.yaml
```diff
@@ -245,7 +245,10 @@ batch_size_fail_threshold_in_kb: 50
 # broadcast_rpc_address: 1.2.3.4
 
 # Uncomment to enable experimental features
-# experimental: true
+# experimental:
+#     - cdc
+#     - lwt
+#     - udf
```
avikivity: Is there a way to keep supporting `experimental: true` to avoid breaking scripts?

avikivity: (it should just enable the entire set)

db/config.cc
```diff
@@ -153,6 +157,24 @@ struct convert<db::config::seed_provider_type> {
     }
 };
 
+template <>
+class convert<enum_option<db::experimental_features>> {
+  public:
```
avikivity: Our coding style doesn't use half indents for access specifiers.

db/config.cc
```diff
@@ -876,6 +896,13 @@ const db::extensions& db::config::extensions() const {
     return *_extensions;
 }
 
+std::unordered_map<sstring, db::experimental_features::feature>& db::experimental_features::map() {
+    static const auto inst = new std::unordered_map<sstring, db::experimental_features::feature>{
```
avikivity: Why `new`?

dist/docker/redhat/commandlineparser.py
```diff
@@ -4,7 +4,7 @@
 def parse():
     parser = argparse.ArgumentParser()
     parser.add_argument('--developer-mode', default='1', choices=['0', '1'], dest='developerMode')
```
avikivity: Should be folded with previous patch, to avoid bisect breakage (unless we manage to make experimental: true work)

conf/scylla.yaml
dekimir: In YAML that might prove difficult, because our YAML parser expects a sequence for `vector`-typed properties.  At best, we can support `experimental:\n  - true` (which would still break scripts).  Or we can have two properties: a binary `experimental` and a vector `experimental_foo`.

In CLI we have a bit more flexibility in implementation, so we could support things like `--experimental=all` or any of the backward-compatible boolean tokens (`on`, `yes`, `1`, and `true`).

db/config.cc
dekimir: To avoid the destruction-order problem: https://isocpp.org/wiki/faq/ctors#construct-on-first-use-v2

conf/scylla.yaml
nyh: We could have two different options: "experimental: true" enables everything, while "experimental_feature: ..." enables just a list of specific features.
This will also allow us to keep the "--experimental" with no options.

db/config.cc
nyh: What destruction-order? If it's a static variable it will only be destructed when the program shuts down, if at all. How can the order of this cause us trouble? And of course, with "new", it will never be destructed at all...

avikivity: It only applies if you have a user in some static destructor, which isn't the case.

dekimir: I don't currently have any static constructors or destructors using this, so technically I don't need the function at all -- I could get away with just a static member.  But this is an established best practice, to avoid future bugs and problems.  We like those, don't we? ;)

tests/enum_option_test.cc
```diff
@@ -0,0 +1,183 @@
...
+    BOOST_CHECK_EQUAL(parse<cities>("YYZ"), cities::TO);
+    BOOST_CHECK_EQUAL(parse<cities>("TheSix"), cities::TO);
+    BOOST_CHECK_EQUAL(parse<cities>("NewYork"), cities::NY);
+    BOOST_CHECK_EQUAL(parse<cities>("NY"), cities::NY);
+    BOOST_CHECK_EQUAL(parse<cities>("NYC"), cities::NY);
+    BOOST_CHECK_EQUAL(parse<cities>("BigApple"), cities::NY);
+}
+
+BOOST_AUTO_TEST_CASE(test_multiple_format) {
+    BOOST_CHECK((std::set<std::string>{"SanFrancisco", "SF", "SFO", "Frisco"}).count(format<cities>(cities::SF)));
```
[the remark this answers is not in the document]
dekimir: Yes.  This is because, in an unordered map, we cannot guarantee which reverse-match we'll get.

tests/enum_option_test.cc
```diff
@@ -0,0 +1,183 @@
...
+
+namespace {
+
+struct days {
+    enum enumeration { Mo, Tu, We, Th, Fr, Sa, Su };
+    static const std::unordered_map<std::string, enumeration>& map() {
+        static const auto inst = new std::unordered_map<std::string, enumeration>{
+            {"Mon", Mo}, {"Tue", Tu}, {"Wed", We}, {"Thu", Th}, {"Fri", Fr}, {"Sat", Sa}, {"Sun", Su}
+        };
+        return *inst;
```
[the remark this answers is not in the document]
dekimir: It will, until it doesn't.  And then we'll be chasing a shifty bug.  See discussion above.

utils/enum_option.hh
```diff
@@ -0,0 +1,106 @@
...
+    // For program_options parser:
+    friend std::istream& operator>>(std::istream& s, enum_option<Mapper>& opt) {
+        typename map_t::key_type key;
+        s >> key;
+        const auto found = Mapper::map().find(key);
+        if (found == Mapper::map().end()) {
+            std::string text;
+            if (s.rdstate() & s.failbit) {
+                // key wasn't read successfully.
+                s >> text;
```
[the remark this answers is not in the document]
dekimir: Because it makes for a better error message! :laughing: 

This read is into a string, while the failed one was into `map_t::key_type`.  For instance, maybe the key type is `int`, and `s` holds "abc", so the first read will fail.  But I still want the user to see an error saying "'abc' is not a valid option value here.

I hope I understood correctly that the `invalid_option_value` constructor takes the problem text as the argument.  I'll double-check...

conf/scylla.yaml
dekimir: That's what I was saying above, except with "foo" instead of "feature".

utils/enum_option.hh
```diff
@@ -0,0 +1,106 @@
...
+///
+/// Mapper must have a static method `map()` that returns a map from a streamable key type (eg, string) to the
+/// enum in question.
+///
+/// Example:
+///
+/// struct Type {
+///   enum ty { a1, a2, b1 };
+///   const static unordered_map<string, ty>& map();
+/// };
```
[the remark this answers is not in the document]
dekimir: You're right, it's not strictly necessary.  If enough people prefer passing the map directly as the `enum_option` template argument, I'm happy to do that.

utils/enum_option.hh
```diff
@@ -0,0 +1,106 @@
...
+///   static const auto inst =
+///     new unordered_map<string, Type::ty>{{"a1", Type::a1}, {"a2", Type::a2}, {"b1", Type::b1}};
+///   return *inst;
+/// }
+/// int main(int ac, char* av[]) {
+///   namespace po = boost::program_options;
+///   po::options_description desc("Allowed options");
+///   desc.add_options()
+///     ("val", po::value<enum_option<Type>>(), "Single Type")
+///     ("vec", po::value<vector<enum_option<Type>>>()->multitoken(), "Type vector");
```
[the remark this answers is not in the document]
dekimir: >     1. Is it possible to make it a set instead of a vector, to avoid duplicates and to allow the code to easily find out which option was enabled?

No, because `set` here doesn't compile.  It violates some static assert deep in Boost.

>     2. What is multitoken()?

It's what allows multiple option values: https://theboostcpplibraries.com/boost.program_options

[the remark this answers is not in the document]
dekimir: I actually might remove even `vector` here, because it's an unnecessary complication.  I'll explain in the blurb that this holds a single value (just like `po::value<int>` holds just a single `int`, not a list of them).

dekimir:
All feedback addressed, PTAL.

Note that this will still break some dtests, because `enable_user_defined_functions` is deleted.

db/config.cc
avikivity: You can find best practices for anything. Another best practice is not to use naked new, and another is not to leak. Yet another (my favorite) is not to use statics. You can avoid violating all best practices here by returning a temporary rather than a reference to a global. If parsing --experimental starts to dominate execution time, we can consider which best practice to violate in order to make it a static.

btw C++20 will avoid these problems with `constinit`.

dekimir: > You can avoid violating all best practices here by returning a temporary 

OK, done.

dekimir:
Addressed feedback and rebased on top of latest `next`.  PTAL.

dekimir:
Actually, wait.  I botched the rebase, so now it shows ~2,000 files changed. :scream: I'm fixing it now...

avikivity:
The note on the top is outdated, yes? Since --experimental is still supported.

dekimir:
K, fixed the rebase.

dekimir:
> The note on the top is outdated, yes? Since --experimental is still supported.

Yes, fixed now.  Thx.

avikivity:
> Delete `--enable-user-defined-functions`.


Er, this is wrong, since it is present in Cassandra and wasn't added to gate an experimental feature.

(Cassandra needs this because their UDF are/were insecure)

dekimir:
> > Delete `--enable-user-defined-functions`.
> 
> Er, this is wrong, since it is present in Cassandra and wasn't added to gate an experimental feature.
> 
> (Cassandra needs this because their UDF are/were insecure)

Do we currently consider UDF experimental in Scylla?  If so, should UDF be a valid value for `--experimental-features`?

avikivity:
Yes, UDF is experimental. In general any new feature that hasn't been blessed by QA is experimental, as a sort of CYA. It also allows us to change storage formats.

dekimir:
> Yes, UDF is experimental. In general any new feature that hasn't been blessed by QA is experimental, as a sort of CYA. It also allows us to change storage formats.

Again, should I allow `--experimental-feature=UDF`?  And if so, how should that interact with `--enable-user-defined-functions`?

avikivity:
Yes please, and both should be set for it to work.

avikivity:
But don't add it to this patch set, let's leave it for later.

dekimir:
If I don't change this patchset, `--enable-user-defined-functions` will disappear, dtests will fail, and we'll stop accepting a valid Cassandra option.  If I do change this patchset _not_ to remove `--enable-user-defined-functions`, then I must also add some reasonable interaction between that and `--experimental-features=UDF`.

avikivity:
I see. We have to un-disappear it. And the interaction is AND: both must be set for the feature to work.

dekimir:
> I see. We have to un-disappear it. And the interaction is AND: both must be set for the feature to work.

K, done.

avikivity:
Is the "NOTE: merging must be coordinated with changes to https://github.com/scylladb/scylla-dtest" still applicable? Since we kept --experimental.

conf/scylla.yaml
```diff
@@ -245,7 +245,10 @@ batch_size_fail_threshold_in_kb: 50
 # broadcast_rpc_address: 1.2.3.4
```
avikivity: Is "Remove the enable_user_defined_functions option, which is now redundant." still correct? I think it's not.

utils/enum_option.hh
```diff
@@ -0,0 +1,106 @@
...
+/// Example:
+///
+/// struct Type {
+///   enum ty { a1, a2, b1 };
+///   const static unordered_map<string, ty>& map();
+/// };
+/// const unordered_map<string, Type::ty>& Type::map() {
+///   static const auto inst =
+///     new unordered_map<string, Type::ty>{{"a1", Type::a1}, {"a2", Type::a2}, {"b1", Type::b1}};
+///   return *inst;
```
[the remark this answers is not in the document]
nyh: Sorry, I still don't understand why there's a "new" here...

utils/enum_option.hh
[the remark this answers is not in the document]
nyh: I don't want to continue bikeshedding on this relatively unimportant issue, and it appears you did test this. But I'm still curious whether it is guaranteed to work... Does a failed read guarantee that the position of the stream does not move?

conf/scylla.yaml
dekimir: No, it's incorrect now.  Thanks for catching this.  Fixed.

utils/enum_option.hh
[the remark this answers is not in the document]
dekimir: Fixed.

utils/enum_option.hh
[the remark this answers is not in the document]
dekimir: > I don't want to continue bikeshedding on this relatively unimportant issue, and it appears you did test this. But I'm still curious whether it is guaranteed to work... Does a failed read guarantee that the position of the stream does not move?

That's my reading of the [official word](https://en.cppreference.com/w/cpp/locale/num_get/get).  In practice, when `key_type` is `int`, this works exactly as expected (and without it, the offending option value is not included in the error message).

avikivity:
I dequeued this:

 - doesn't show up in --help for some reason:

```
$ ./build/release/scylla --help | grep experimental
  --experimental arg                    Set to true to unlock experimental 
```

 - misleading error message:

```
11:01:29      assert e.code == 0000 and e.message == "Not implemented: LWT", e
11:01:29  AssertionError: <Error from server: code=0000 [Server error] message="Paxos is currently disabled. Start Scylla with --experimental_features=lwt to enable.">
```

should be --experimental-features (the command line options auto-translate _ to -)

 - breaks dtest:

```
11:01:29      assert e.code == 0000 and e.message == "Not implemented: LWT", e
11:01:29  AssertionError: <Error from server: code=0000 [Server error] message="Paxos is currently disabled. Start Scylla with --experimental_features=lwt to enable.">
```

This might be fixed by a coordinated change to dtest.

nyh:
That this patch will necessitate dtest changes was obvious (and IIRC, predicted by the patch author). But how can we "coordinate" this change? Should we change the dtest code to support both pre-patch and post-patch scylla (perhaps by using "--help" or something to figure out what option dtest should use)? Is this complication worth it?

nyh:
By the way, how do we mark this pull request "not really merged after all"? :-(

avikivity:
I usually write "I dequeued this".

nyh:
Writing "I dequeued this" doesn't help when someone is using https://github.com/scylladb/scylla/pulls to list the pull requests that need attention. Never mind, yet another annoying github feature...

avikivity:
> That this patch will necessitate dtest changes was obvious (and IIRC, predicted by the patch author).

After the change that brought back --experimental, I thought changes to dtest wouldn't be needed. What broke dtest was a change to the error message. 

> But how can we "coordinate" this change?

Send pull requests to dtest and scylla simultaneously, and coordinate with maintainers to merge simultaneously.

> Should we change the dtest code to support both pre-patch and post-patch scylla (perhaps by using "--help" or something to figure out what option dtest should use)? Is this complication worth it?

No need to fly through such hoops, just prepare changes to both. These things happen from time to time, it's no big deal.

glommer:
On Tue, Dec 10, 2019 at 5:59 AM nyh <notifications@github.com> wrote:

> Writing "I dequeued this" doesn't help when someone is using
> https://github.com/scylladb/scylla/pulls to list the pull requests that
> need attention. Never mind, yet another annoying github feature...
>

We can use tags for that, and reopen the PR.

avikivity:
There's no "reopen" button for pull requests.

glommer:
On Tue, Dec 10, 2019 at 10:01 AM Avi Kivity <notifications@github.com>
wrote:

> There's no "reopen" button for pull requests.
>

How nice!

In this case we can use tags. I will scan for such tags.
(Please tell me we can add a tag to a closed PR)

avikivity:
@glommer you can add a tag to a closed PR

glommer:
Ok, so if we standardize on a "dequeued" tag (or whatever you want) being added, I can change the scripts I am writing to recognize it.

dekimir:
>  doesn't show up in --help for some reason:
> 
> 
> ```
> $ ./build/release/scylla --help | grep experimental
>   --experimental arg                    Set to true to unlock experimental 
> ```

Weird, it shows up for me:
```bash
localhost>./scylla --help|grep experimental
  --experimental arg                    Set to true to unlock all experimental 
  --experimental-features arg           Unlock experimental features provided 
                                        also set experimental-features=udf
```

avikivity:
User error, I built dev and ran release.

dekimir:
I cannot update this PR anymore, so I put my fixes in #5456.

nyh:
By the way, all the problems we had with this pull-request being prematurely closed, before the patch actually was promoted to master, would have been solved if the branch was marked for **master**, not **next**, so the pull request will only be finally closed when it reaches master.

Note that to do this, we need pull request targets to be "master", not "next", and we need to merge them using manual git commands - not with the github GUI buttons because the latter will need to have the target being marked "next". But that (manual git commands) is what I'm usually doing anyway.

PR #5371, opened 2019-11-28
gossiper: stop inheriting from async_sharded_service

haaawk:
It seems that async_sharded_service::_delete_cb is never
set so there's no reason to inherit from async_sharded_service
at all.

### DIFF
--- a/gms/gossiper.hh
+++ b/gms/gossiper.hh
@@ -103,7 +103,7 @@ struct ack_msg_pending {
  * Upon hearing a GossipShutdownMessage, this module will instantly mark the remote node as down in
  * the Failure Detector.
  */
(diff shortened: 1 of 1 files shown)


haaawk:
I checked that in enterprise _delete_cb is not used as well so the change should be safe there too.

haaawk:
It turns out the field is used through reflection.

asias:
@haaawk async_sharded_service is used by gossiper as a guard to prevent to destroy while gossiper object is still in use. Any reason you want to avoid using async_sharded_service other than you suspected a bug with the implementation of  async_sharded_service?

haaawk:
The PR is already closed @asias - I don't think it's correct any more.

Issue #5354, opened 2019-11-26
hintedhandoff_removenode_test Missing: ['hints_manager - Draining for 127.0.9.3: end']


bhalevy:
Seen in [dtest-debug/314](https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug/314/testReport/hintedhandoff_additional_test/TestHintedHandoff/hintedhandoff_removenode_test)
```
tacktrace
Traceback (most recent call last):
  File "/usr/lib64/python2.7/unittest/case.py", line 367, in run
    testMethod()
  File "/jenkins/workspace/scylla-master/dtest-debug/scylla-dtest/hintedhandoff_additional_test.py", line 127, in hintedhandoff_removenode_test
    node1.watch_log_for(msg, from_mark=from_mark, timeout=timeout)
  File "/jenkins/workspace/scylla-master/dtest-debug/scylla-ccm/ccmlib/node.py", line 435, in watch_log_for
    [e.pattern for e in tofind]) + ":\n" + reads[:50] + ".....\nSee {} for remainder".format(filename))
TimeoutError: 21 Nov 2019 05:04:14 [node1] Missing: ['hints_manager - Draining for 127.0.9.3: end']:
```

Scylla version 9a686ac55193f0df4fc923b267b7b36684565a9d
Scylla-dtest version 78c93908540fcf5f05ed48d1845005a027b4c904
(That is scylladb/scylla-dtest#1134 fixing #5244)

[node1.log](https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug/314/artifact/logs-debug.2/1574312654267_hintedhandoff_additional_test.TestHintedHandoff.hintedhandoff_removenode_test/node1.log) contains only:
```
TRACE 2019-11-21 05:03:41,849 [shard 0] hints_manager - on_leave_cluster: 127.0.9.3 is removed/decommissioned
TRACE 2019-11-21 05:03:41,849 [shard 0] hints_manager - on_leave_cluster: 127.0.9.3 is removed/decommissioned
TRACE 2019-11-21 05:03:41,853 [shard 1] hints_manager - on_leave_cluster: 127.0.9.3 is removed/decommissioned
TRACE 2019-11-21 05:03:41,853 [shard 1] hints_manager - on_leave_cluster: 127.0.9.3 is removed/decommissioned
INFO  2019-11-21 05:03:41,873 [shard 0] gossip - Completing removal of 127.0.9.3
[...]

PR #5372, opened 2019-11-28
hinted handoff: log message after removing hints directory

bhalevy:
To be used by dtest as an indicator that endpoint's hints
were drained and hints directory is removed.

Refs #5354

### DIFF
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -571,6 +571,8 @@ void manager::drain_for(gms::inet_address endpoint) {
                 manager_logger.error("Exception when draining {}: {}", endpoint, eptr);
             });
         });
+    }).finally([endpoint] {
+        manager_logger.trace("drain_for: finished draining {}", endpoint);
     });
 }
 


vladzcloudius:
@bhalevy I'm working on the corresponding fix to both scylla and dtest.
The patch I think should be sufficient is as follows:
```
diff --git a/db/hints/manager.cc b/db/hints/manager.cc
index 9e8e60261..2c42d86a2 100644
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -571,6 +571,8 @@ void manager::drain_for(gms::inet_address endpoint) {
                 manager_logger.error("Exception when draining {}: {}", endpoint, eptr);
             });
         });
+    }).finally([endpoint] {
+        manager_logger.trace("drain_for: finished draining {}", endpoint);
     });
 }

```

slivne:
@bhalevy / @vladzcloudius ping

bhalevy:
> @bhalevy I'm working on the corresponding fix to both scylla and dtest.
> The patch that IMO would be sufficient is as follows:
> 
> ```
> diff --git a/db/hints/manager.cc b/db/hints/manager.cc
> index 9e8e60261..2c42d86a2 100644
> --- a/db/hints/manager.cc
> +++ b/db/hints/manager.cc
> @@ -571,6 +571,8 @@ void manager::drain_for(gms::inet_address endpoint) {
>                  manager_logger.error("Exception when draining {}: {}", endpoint, eptr);
>              });
>          });
> +    }).finally([endpoint] {
> +        manager_logger.trace("drain_for: finished draining {}", endpoint);
>      });
>  }
> ```

@vladzcloudius I update the patch accordingly.

Issue #5281, opened 2019-11-14
name reader concurrency semaphores


denesb:
We should do to reader concurrency semaphores what https://github.com/scylladb/scylla/pull/5269 is doing for all regular semaphores in the codebase, that is name them and add this name to error messages generated from said semaphores.
We have three concurrency semaphores: user, system and streaming and albeit conceptually they do have names, these are not propagated down to the semaphore itself.

denesb:
This would greatly help with debugging problems, as these semaphores are some of the most often timing out ones.

tgrabiec:
I'd also throw the ones we have on the write path: in the commitlog (`segment_manager::_request_controller`) and before the memtable (`region_group::run_when_memory_available()`)

PR #5376, opened 2019-11-28
Include semaphore name in reader_concurrency_semaphore exception

jul-stas:
Semaphore names are now explicitly contained in runtime_error message. Thrown from lambda (the last parameter of `reader_concurrency_semaphore` ctor).

### DIFF
--- a/database.cc
+++ b/database.cc
@@ -202,16 +202,20 @@ database::database(const db::config& cfg, database_config dbcfg)
         max_inactive_queue_length(),
         [this] {
             ++_stats->sstable_read_queue_overloaded;
-            return std::make_exception_ptr(std::runtime_error("sstable read queue overloaded"));
+            return std::make_exception_ptr(std::runtime_error("_read_concurrency_sem overloaded"));
         })
     // No timeouts or queue length limits - a failure here can kill an entire repair.
     // Trust the caller to limit concurrency.
     , _streaming_concurrency_sem(
             max_count_streaming_concurrent_reads,
-            max_memory_streaming_concurrent_reads())
+            max_memory_streaming_concurrent_reads(),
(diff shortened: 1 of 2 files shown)


database.cc
```diff
@@ -206,8 +206,12 @@ database::database(const db::config& cfg, database_config dbcfg)
         })
     // No timeouts or queue length limits - a failure here can kill an entire repair.
```
avikivity: Pull request title should be in English

database.cc
```diff
@@ -202,16 +202,20 @@ database::database(const db::config& cfg, database_config dbcfg)
...
     // Trust the caller to limit concurrency.
     , _streaming_concurrency_sem(
             max_count_streaming_concurrent_reads,
-            max_memory_streaming_concurrent_reads())
+            max_memory_streaming_concurrent_reads(),
+            std::numeric_limits<size_t>::max(),
+            [] { return std::make_exception_ptr(std::runtime_error("_streaming_concurrency_sem overloaded")); })
     , _system_read_concurrency_sem(
             max_count_system_concurrent_reads,
-            max_memory_system_concurrent_reads())
```
tgrabiec: #5281 is also about changing timeout errors, so you need to propagate the name into `reader_concurrency_semaphore` and into `expiring_fifo`.

Also, there is more semaphores which deserve naming, e.g. the ones in the commitlog and region_group::run_when_memory_available()

jul-stas: Part 1: Absolutely my bad, I didn't notice that it's timed. I'll do it.
Part 2: Sure, I can do it, but I'll need the task to indicate which semaphores are expected to be named. I don't want this PR to hang forever with debates on name for every mutex...

PR #5378, opened 2019-11-28
hinted handoff: handle hints on separate connection and scheduling group

### LINKED ISSUE #4974: Hinted handoff: take the number of nodes and shards in the cluster into an account when determining the maximum send parallelism

piodul:
Introduce a new verb dedicated for receiving and sending hints: `HINT_MUTATION`. It is handled on the streaming connection, which is separate from the one used for handling mutations sent by coordinator during a write.

The intent of using a separate connection is to increase fairness while handling hints and user requests - this way, a situation can be avoided in which one type of requests saturate the connection, negatively impacting the other one.

Information about new RPC support is propagated through new gossip feature `HINTED_HANDOFF_SEPARATE_CONNECTION`.

Fixes #4974.

Tests: unit(release)

### DIFF
--- a/message/messaging_service.cc
+++ b/message/messaging_service.cc
@@ -480,6 +480,7 @@ static constexpr unsigned do_get_rpc_client_idx(messaging_verb verb) {
     case messaging_verb::REPAIR_GET_ROW_DIFF_WITH_RPC_STREAM:
     case messaging_verb::REPAIR_PUT_ROW_DIFF_WITH_RPC_STREAM:
     case messaging_verb::REPAIR_GET_FULL_ROW_HASHES_WITH_RPC_STREAM:
+    case messaging_verb::HINT_MUTATION:
         return 2;
     case messaging_verb::MUTATION_DONE:
     case messaging_verb::MUTATION_FAILED:
@@ -1273,6 +1274,19 @@ future<> messaging_service::send_paxos_learn(msg_addr id, clock_type::time_point
     std::vector<inet_address> forward, inet_address reply_to, unsigned shard, response_id_type response_id,
     std::optional<tracing::trace_info> trace_info) {
     return send_message_oneway_timeout(this, timeout, messaging_verb::PAXOS_LEARN, std::move(id), decision, std::move(forward),
+        std::move(reply_to), std::move(shard), std::move(response_id), std::move(trace_info));
+}
+
+void messaging_service::register_hint_mutation(std::function<future<rpc::no_wait_type> (const rpc::client_info&, rpc::opt_time_point, frozen_mutation fm, std::vector<inet_address> forward,
+        inet_address reply_to, unsigned shard, response_id_type response_id, rpc::optional<std::optional<tracing::trace_info>> trace_info)>&& func) {
+    register_handler(this, netw::messaging_verb::HINT_MUTATION, std::move(func));
+}
+void messaging_service::unregister_hint_mutation() {
+    _rpc->unregister_handler(netw::messaging_verb::HINT_MUTATION);
+}
+future<> messaging_service::send_hint_mutation(msg_addr id, clock_type::time_point timeout, const frozen_mutation& fm, std::vector<inet_address> forward,
+        inet_address reply_to, unsigned shard, response_id_type response_id, std::optional<tracing::trace_info> trace_info) {
+    return send_message_oneway_timeout(this, timeout, messaging_verb::HINT_MUTATION, std::move(id), fm, std::move(forward),
         std::move(reply_to), std::move(shard), std::move(response_id), std::move(trace_info));
 }
 
--- a/message/messaging_service.cc
+++ b/message/messaging_service.cc
@@ -995,7 +995,7 @@ void messaging_service::unregister_mutation() {
 future<> messaging_service::send_mutation(msg_addr id, clock_type::time_point timeout, const frozen_mutation& fm, std::vector<inet_address> forward,
     inet_address reply_to, unsigned shard, response_id_type response_id, std::optional<tracing::trace_info> trace_info) {
     return send_message_oneway_timeout(this, timeout, messaging_verb::MUTATION, std::move(id), fm, std::move(forward),
-        std::move(reply_to), std::move(shard), std::move(response_id), std::move(trace_info));
+        std::move(reply_to), shard, std::move(response_id), std::move(trace_info));
 }
 
 void messaging_service::register_counter_mutation(std::function<future<> (const rpc::client_info&, rpc::opt_time_point, std::vector<frozen_mutation> fms, db::consistency_level cl, std::optional<tracing::trace_info> trace_info)>&& func) {
@@ -1015,7 +1015,7 @@ void messaging_service::unregister_mutation_done() {
     _rpc->unregister_handler(netw::messaging_verb::MUTATION_DONE);
 }
 future<> messaging_service::send_mutation_done(msg_addr id, unsigned shard, response_id_type response_id, db::view::update_backlog backlog) {
-    return send_message_oneway(this, messaging_verb::MUTATION_DONE, std::move(id), std::move(shard), std::move(response_id), std::move(backlog));
+    return send_message_oneway(this, messaging_verb::MUTATION_DONE, std::move(id), shard, std::move(response_id), std::move(backlog));
 }
 
 void messaging_service::register_mutation_failed(std::function<future<rpc::no_wait_type> (const rpc::client_info& cinfo, unsigned shard, response_id_type response_id, size_t num_failed, rpc::optional<db::view::update_backlog> backlog)>&& func) {
@@ -1025,7 +1025,7 @@ void messaging_service::unregister_mutation_failed() {
     _rpc->unregister_handler(netw::messaging_verb::MUTATION_FAILED);
 }
 future<> messaging_service::send_mutation_failed(msg_addr id, unsigned shard, response_id_type response_id, size_t num_failed, db::view::update_backlog backlog) {
-    return send_message_oneway(this, messaging_verb::MUTATION_FAILED, std::move(id), std::move(shard), std::move(response_id), num_failed, std::move(backlog));
+    return send_message_oneway(this, messaging_verb::MUTATION_FAILED, std::move(id), shard, std::move(response_id), num_failed, std::move(backlog));
 }
 
 void messaging_service::register_read_data(std::function<future<rpc::tuple<foreign_ptr<lw_shared_ptr<query::result>>, cache_temperature>> (const rpc::client_info&, rpc::opt_time_point t, query::read_command cmd, ::compat::wrapping_partition_range pr, rpc::optional<query::digest_algorithm> oda)>&& func) {
@@ -1274,7 +1274,7 @@ future<> messaging_service::send_paxos_learn(msg_addr id, clock_type::time_point
     std::vector<inet_address> forward, inet_address reply_to, unsigned shard, response_id_type response_id,
     std::optional<tracing::trace_info> trace_info) {
     return send_message_oneway_timeout(this, timeout, messaging_verb::PAXOS_LEARN, std::move(id), decision, std::move(forward),
-        std::move(reply_to), std::move(shard), std::move(response_id), std::move(trace_info));
+        std::move(reply_to), shard, std::move(response_id), std::move(trace_info));
 }
 
 void messaging_service::register_hint_mutation(std::function<future<rpc::no_wait_type> (const rpc::client_info&, rpc::opt_time_point, frozen_mutation fm, std::vector<inet_address> forward,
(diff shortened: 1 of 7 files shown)


piodul:
@vladzcloudius

I tried testing this PR in order to observe if it gives any improvement. I tried the following test:

- Start 3 or 9 node ccm cluster,
- Set up following resources:

```
CREATE KEYSPACE keyspace1 WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 2};
CREATE TABLE keyspace1.standard1 (
    key blob PRIMARY KEY,
    "C0" blob,
    "C1" blob,
    "C2" blob,
    "C3" blob,
    "C4" blob
);
```

- Optionally, created a materialized view like this: `CREATE MATERIALIZED VIEW IF NOT EXISTS keyspace1.mv AS SELECT key, "C0", "C1", "C2", "C3" FROM keyspace1.standard1 WHERE "C0" IS NOT NULL AND key > 0x PRIMARY KEY ("C0", key);`
- Start cassandra-stress with 50-2000 threads,
- Stop one node for 1 minute, then start again,
- Observe metrics in grafana.

I looked at all dashboards in grafana from `scylla-monitoring`, and additionally at hint sending and writing numbers, and did not see any noticeable differences between the versions. I tried different combinations of parameters (with/without the view, 3 or 9 nodes, number of threads and write rate for `cassandra-stress`).

The test was run locally, on a cluster which was setup with ccm.

I suspect that the test setup or implementation is wrong. To be frank, I'm also not sure what kind of differences in behavior to look for. Could you tell me how to set up the test correctly?

message/messaging_service.cc
```diff
@@ -484,6 +484,8 @@ static constexpr unsigned do_get_rpc_client_idx(messaging_verb verb) {
     case messaging_verb::MUTATION_DONE:
     case messaging_verb::MUTATION_FAILED:
         return 3;
+    case messaging_verb::HINT_MUTATION:
+        return 4;
```
gleb-cloudius: Why not use streaming connection?

service/storage_proxy.cc
```diff
@@ -258,6 +258,41 @@ class shared_mutation : public mutation_holder {
     }
 };
 
+// shared mutation, but gets sent as a hint
+class hint_mutation : public mutation_holder {
```
gleb-cloudius: You can inherit from shared_mutation

service/storage_proxy.cc
```diff
@@ -258,6 +258,41 @@ class shared_mutation : public mutation_holder {
...
+public:
+    explicit hint_mutation(frozen_mutation_and_schema&& fm_a_s)
+            : _mutation(make_lw_shared<const frozen_mutation>(std::move(fm_a_s.fm))) {
+        _size = _mutation->representation().size();
+        _schema = std::move(fm_a_s.s);
+    }
+    explicit hint_mutation(const mutation& m) : hint_mutation(frozen_mutation_and_schema{freeze(m), m.schema()}) {
+    }
+    virtual bool store_hint(db::hints::manager& hm, gms::inet_address ep, tracing::trace_state_ptr tr_state) override {
+            return hm.store_hint(ep, _schema, _mutation, tr_state);
```
gleb-cloudius: It does not make send to store a hint for a hint.

service/storage_proxy.cc
```diff
@@ -258,6 +259,31 @@ class shared_mutation : public mutation_holder {
     }
 };
 
+// shared mutation, but gets sent as a hint
+class hint_mutation : public shared_mutation {
+public:
+    explicit hint_mutation(frozen_mutation_and_schema&& fm_a_s) : shared_mutation(std::move(fm_a_s)){
+    }
+    explicit hint_mutation(const mutation& m) : shared_mutation(m) {
+    }
```
gleb-cloudius: I think you can do `using shared_mutation::shared_mutation`

service/storage_proxy.cc
```diff
@@ -258,6 +259,31 @@ class shared_mutation : public mutation_holder {
...
+    virtual bool store_hint(db::hints::manager& hm, gms::inet_address ep, tracing::trace_state_ptr tr_state) override {
+        return false;
+    }
```
gleb-cloudius: Why not exception? But I do not insist.

piodul: I wouldn't like to risk it. I can see a possibility where `store_hint` could be called at shutdown (`drain_on_shutdown` -> `timeout_cb` -> `hint_to_dead_endpoints` -> `store_hint`).

gleb-cloudius: You are calling send_to_endpoint with disabled hints, this should not happen

vladzcloudius: I agree with @gleb-cloudius if this one is hit it would definitely be a bug and we would want to know about this ASAP. You may keep it as 'return false' but then, please, add a very scary error message there.

message/messaging_service.cc
```diff
@@ -1276,4 +1277,17 @@ future<> messaging_service::send_paxos_learn(msg_addr id, clock_type::time_point
...
+void messaging_service::register_hint_mutation(std::function<future<rpc::no_wait_type> (const rpc::client_info&, rpc::opt_time_point, frozen_mutation fm, std::vector<inet_address> forward,
+        inet_address reply_to, unsigned shard, response_id_type response_id, rpc::optional<std::optional<tracing::trace_info>> trace_info)>&& func) {
+    register_handler(this, netw::messaging_verb::HINT_MUTATION, std::move(func));
+}
+void messaging_service::unregister_hint_mutation() {
+    _rpc->unregister_handler(netw::messaging_verb::HINT_MUTATION);
+}
+future<> messaging_service::send_hint_mutation(msg_addr id, clock_type::time_point timeout, const frozen_mutation& fm, std::vector<inet_address> forward,
+        inet_address reply_to, unsigned shard, response_id_type response_id, std::optional<tracing::trace_info> trace_info) {
+    return send_message_oneway_timeout(this, timeout, messaging_verb::MUTATION, std::move(id), fm, std::move(forward),
```
vladzcloudius: Should not the verb be `messaging_verb::HINT_MUTATION`?

message/messaging_service.cc
```diff
@@ -1276,4 +1277,17 @@ future<> messaging_service::send_paxos_learn(msg_addr id, clock_type::time_point
...
+        std::move(reply_to), std::move(shard), std::move(response_id), std::move(trace_info));
```
vladzcloudius: Some nitpicks (I'm surprised @gleb-cloudius missed these! :)):
`std::move(shard)` seems to be pointless - it's type is `unsigned`.
The same about std::move(response_id) which is uint64_t. Although I would keep this in case it changes in the future like `inet_address` did.

service/storage_proxy.cc
```diff
@@ -258,6 +259,28 @@ class shared_mutation : public mutation_holder {
...
+    virtual future<> apply_locally(storage_proxy& sp, storage_proxy::clock_type::time_point timeout,
+            tracing::trace_state_ptr tr_state) override {
+        return make_exception_future<>(std::runtime_error("Executing hint locally doesn't make sense"));
+    }
+    virtual future<> apply_remotely(storage_proxy& sp, gms::inet_address ep, std::vector<gms::inet_address>&& forward,
+            storage_proxy::response_id_type response_id, storage_proxy::clock_type::time_point timeout,
+            tracing::trace_state_ptr tr_state) override {
+        tracing::trace(tr_state, "Sending a hint to /{}", ep);
+        auto& ms = netw::get_local_messaging_service();
+        return ms.send_hint_mutation(netw::messaging_service::msg_addr{ep, 0}, timeout, *_mutation,
```
vladzcloudius: Why `cpu_id` is hard coded to `0`?

vladzcloudius: I don't see where you send this on a separate (or streaming) socket.
Do I miss something?

vladzcloudius:
@piodul About the test: an indication of the problem is an increased latency of regular writes beyond what <main class shares>:<streaming shares> would define.

In order to make this easier I'd recommend to use larger payload than the default c-s one, e.g. 10KB (the default is ~300B).

You may also want to make the node to be down for a little longer than 1 minute in order to allow about 5-10m worth hints to get accumulated (here I refer the amount of time it would take a node to receive all of these hints).

Then, you may want to switch to CL=ALL when hints are being sent so that your query latency get hit by those hints for sure. 

(Note that you can't use CL=ALL during the phase when you generate hints).

During the hints sending phase the rate of queries should correspond to ~30%-70% per-shard CPU load without hints. 
However with hints being sent CPU load should be 100%.

So, the expected observation should be:
 * Without your patches - latency degradation should be beyond 20% compared to the same writes rate without any hints.
 * With your patches a latency degradation should be not more than 20%.

Please, let me know if you need further assistance with a test case. We can jump on a quick call if needed.

message/messaging_service.cc
gleb-cloudius: Ha. That's a good catch!

message/messaging_service.cc
gleb-cloudius: I had to leave something to you!

service/storage_proxy.cc
gleb-cloudius: cpu_id is always ignored. It is fine to hardcode it to whatever.

service/storage_proxy.cc
piodul: Ok. In the end I changed it into an exception throw.

message/messaging_service.cc
piodul: Oops... Fixed!

message/messaging_service.cc
piodul: It seems that some other handlers have the same problem. I got rid of moves of primitive integral types (`unsigned`, `uint32_t` and `uint64_t` - left alone typedefs such as `response_id`).

service/storage_proxy.cc
```diff
@@ -2230,6 +2253,20 @@ future<> storage_proxy::send_to_endpoint(
...
+    if (!service::get_local_storage_service().cluster_supports_hinted_handoff_separate_connection()) {
+        return send_to_endpoint(std::move(fm_a_s), target, { }, db::write_type::SIMPLE, allow_hints::no);
+    }
+
+    return send_to_endpoint(
+            std::make_unique<hint_mutation>(std::move(fm_a_s)),
+            std::move(target),
+            {},
+            db::write_type::SIMPLE,
+            _stats,
```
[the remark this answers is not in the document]
piodul: I changed the first call to use `_stats` instead, so that both use the same overload (the one with the actual logic). This way we don't need to introduce yet another overload.

service/storage_proxy.cc
piodul: @vladzcloudius Relevant changes are situated in `messaging_service.cc`. The client for given request is chosen based on `s_rpc_client_idx_table` array, which is initialized based on `do_get_rpc_client_idx` function.

service/storage_proxy.cc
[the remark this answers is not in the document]
vladzcloudius: Fine with me.

message/messaging_service.cc
vladzcloudius: Hmmm... I still see `MUTATION` verb. Are you sure you pushed your changes?

service/storage_proxy.cc
vladzcloudius: The message in the exception is confusing. It's not that you would try to store a hint for a second time but rather you would store a hint for a hint.

service/storage_proxy.cc
vladzcloudius: > @vladzcloudius Relevant changes are situated in `messaging_service.cc`. The client for given request is chosen based on `s_rpc_client_idx_table` array, which is initialized based on `do_get_rpc_client_idx` function.

Oh, I see it now. Thanks.

service/storage_proxy.cc
piodul: Fixed

message/messaging_service.cc
piodul: Well, I did, but I applied the fix to a wrong commit. It should be all in order now.

piodul:
Fixed the spelling of the "move register_mutation handler to local lambda" commit.
Changed slightly the description of the last commit (getting rid of moves), as it actually gets rid of moves of `unsigned` values (there were no moves of `uint32_t`s and `uint64_t`s).

slivne:
@gleb-cloudius please review (hopefully bless) ... I want to try and  have sarna merge this in.

message/messaging_service.cc
```diff
@@ -994,7 +995,7 @@ void messaging_service::unregister_mutation() {
 future<> messaging_service::send_mutation(msg_addr id, clock_type::time_point timeout, const frozen_mutation& fm, std::vector<inet_address> forward,
     inet_address reply_to, unsigned shard, response_id_type response_id, std::optional<tracing::trace_info> trace_info) {
     return send_message_oneway_timeout(this, timeout, messaging_verb::MUTATION, std::move(id), fm, std::move(forward),
-        std::move(reply_to), std::move(shard), std::move(response_id), std::move(trace_info));
+        std::move(reply_to), shard, std::move(response_id), std::move(trace_info));
```
gleb-cloudius: For the future those cleanups should be in a separate patch.

PR #5379, opened 2019-11-28
Adding functionality to enforce optimal Linux clocksource setting.

### LINKED ISSUE #4474: scylla doesn't enforce the use of TSC clocksource, potentially ends up with bad performance
Because Scylla is timestamp-intensive, the choice of the clocksource can have a very big impact on performance. For our AMI, we enforce the TSC clocksource through a boot option, but for users who are not in the AMI that is not enforced.

I debugged a problem today that was pinned to a clocksource other than tsc being used even though tsc was available.

Scylla_setup should enforce that.

flucchese:
A Linux machine typically has multiple clocksources with distinct
performances. Setting a high-performant clocksource might result in
better performance for ScyllaDB, so this should be considered whenever
starting it up.

This patch introduces the possibility of enforcing optimized Linux
clocksource to Scylla's setup/start-up processes. It does so by adding
an interactive question about enforcing clocksource setting to scylla_setup,
which modifies the parameter "CLOCKSOURCE" in scylla_server configuration
file. This parameter is read by perftune.py which, if set to "yes", proceeds
to (non persistently) setting the clocksource. On x86, TSC clocksource is used.

Fixes #4474

### DIFF
--- a/dist/common/scripts/scylla_prepare
+++ b/dist/common/scripts/scylla_prepare
@@ -75,9 +75,5 @@ if __name__ == '__main__':
             run('hugeadm --create-mounts')
         fi
     else:
-        set_nic_and_disks = get_set_nic_and_disks_config_value(cfg)
-        ifname = cfg.get('IFNAME')
-        if set_nic_and_disks == 'yes':
-            create_perftune_conf(ifname)
+        if create_perftune_conf(cfg):
             run("{} --options-file /etc/scylla.d/perftune.yaml".format(perftune_base_command()))
-
--- a/dist/common/scripts/scylla_setup
+++ b/dist/common/scripts/scylla_setup
@@ -182,6 +182,8 @@ if __name__ == '__main__':
                         help='skip daily version check')
     parser.add_argument('--no-node-exporter', action='store_true', default=False,
                         help='do not install the node exporter')
+    parser.add_argument('--no-clocksource', action='store_true', default=False,
+                        help='do not enforce the use of Linux clock source')
     parser.add_argument('--no-cpuscaling-setup', action='store_true', default=False,
                         help='skip cpu scaling setup')
     parser.add_argument('--no-fstrim-setup', action='store_true', default=False,
@@ -215,6 +217,7 @@ if __name__ == '__main__':
     io_setup = not args.no_io_setup
     version_check = not args.no_version_check
     node_exporter = not args.no_node_exporter
+    set_clocksource = not args.no_clocksource
     cpuscaling_setup = not args.no_cpuscaling_setup
     fstrim_setup = not args.no_fstrim_setup
     selinux_reboot_required = False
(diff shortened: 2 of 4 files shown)


glommer:
@nyh do you mind taking a look at this ?

dist/common/scripts/scylla_util.py
```diff
@@ -167,7 +166,7 @@ def instance_class(self):
         return self._type.split(".")[0]
 
     def is_supported_instance_class(self):
-        if self.instance_class() in ['i2', 'i3', 'i3en', 'c5d', 'm5d', 'm5ad', 'r5d', 'z1d']:
```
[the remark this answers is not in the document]
flucchese: Indeed, that shouldn't have been applied. Probably an editing mistake.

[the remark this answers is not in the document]
flucchese: The additional strings were added after I forked the repo. Nick to port them to my patch.

flucchese:
I forced pushed a correction to the scylla_util line 169 issue.

dist/common/scripts/scylla_prepare
```diff
@@ -20,8 +20,6 @@
 # You should have received a copy of the GNU General Public License
 # along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
 
-import os
-import sys
```
vladzcloudius: Are these two lines related to this patch?
I see a similar complain from @nyh below about a different hunk.
I join Nadav's recommendation. - Send cleanups in a separate patch, please.

dist/common/scripts/scylla_prepare
```diff
@@ -75,9 +73,5 @@ if __name__ == '__main__':
             run('hugeadm --create-mounts')
         fi
     else:
-        set_nic_and_disks = get_set_nic_and_disks_config_value(cfg)
-        ifname = cfg.get('IFNAME')
```
vladzcloudius: I'm not sure I understand why you choose to run `perftune.py` unconditionally.
What if the user answered "No" both to the question if he/she wants to tune a clocksource and about NIC and I/O tuning? I think that in this case we are still expected *not* to run a `perftune.py` like before.

Please, let me know if I missed something.

dist/common/scripts/scylla_setup
```diff
@@ -417,4 +423,4 @@ if __name__ == '__main__':
             print(' SELinux.')
 
         if dist_name() == 'Ubuntu':
-            run('apt-get install -y hugepages')
+            run('apt-get install -y hugepages')
```
vladzcloudius: This is the most obvious example - something is not right with your editor/laptop/OS... ;)

dist/common/scripts/scylla_util.py
```diff
@@ -520,14 +519,36 @@ def get_tune_mode(nic):
...
+        if not nic:
+            nic = 'eth0'
+        params += '--tune net --nic "{nic}"'.format(nic=nic)
+
+    if cfg.get('CLOCKSOURCE') == 'yes':
+        params += ' --tune-clock'
+
+    if len(params) > 0:
+        if os.path.exists('/etc/scylla.d/perftune.yaml'):
+            return True
```
vladzcloudius: Hmmm... I'm not sure this is correct. What if the content of `perftune.yaml` was generated with one set of parameters and now you are given a different one? I think you should re-generate it regardless if it exists.

glommer: We don't want to regenerate, Vlad.

What if the user has changed something manually? We don't regenerate today, as far as I understand, so I am not sure this behavior should change.

We can force regenerating the file by deleting it if anything changes. Regenerating unconditionally is dangerous. We wouldn't need a conf file if we were to always regenerate it.

flucchese: Yes, this is the current behavior. I didn't want to change it since it was not related to this patch.

dist/common/scripts/scylla_prepare
flucchese: I'll remove this and the one Nadav pointed out from the PR.

dist/common/scripts/scylla_prepare
flucchese: We're not running perftune unconditionally. We're doing exactly as you described. However, the logic was moved to scylla-Util, where it looked more natural. The condition on scylla_prepare is "if create_perftune_conf(cfg): run("..."

dist/common/scripts/scylla_setup
flucchese: Probably a tab instead of white spaces. Need to fix that too.

flucchese: This is because I remove the empty line at the end of the files I changed. I'll roll back these changes.

flucchese:
I've just removed the changes that were not directly related to this patch's proposition, such as unused imports and empty lines. I forced pusshed this update to the same commit.

dist/common/scripts/scylla_prepare
vladzcloudius: I thought I have deleted this comment. ;)
Yeah, I saw that further in the patch.

This also means that I have probably deleted some other comment I shouldn't have... :D

dist/common/scripts/scylla_util.py
vladzcloudius: You are both right.

dist/common/scripts/scylla_util.py
```diff
@@ -520,14 +520,36 @@ def get_tune_mode(nic):
...
+        
+        mode = get_tune_mode(nic)
+        params += ' --mode {mode} --dump-options-file'.format(mode=mode)
+        yaml = out('/opt/scylladb/scripts/perftune.py '+params)
```
vladzcloudius: Nitpick: spaces around "+" would be cleaner.

flucchese: Ok, fixed.

glommer:
@nyh Fabiano addressed your requested changes.

I have added two requests that are fairly simple and related to wording and string names.
After those are done I am ok with the PR.

flucchese:
Hi, @glommer. I pushed the changes you requested as well.

glommer:
This can go in. @nyh please merge.

penberg:
@flucchese Thanks! For future pull requests, please set the merge target to `next`.

PR #5380, opened 2019-11-28
api: replace swagger definition to use long instead of int

### LINKED ISSUE #5347: swagger: wrong return type of `column_family/metrics/total_disk_space_used`
*Installation details*
Scylla version (or git commit hash): master

[API](https://github.com/scylladb/scylla/blob/master/api/column_family.cc#L554) endpoint [returns uint64_t](https://github.com/scylladb/scylla/blob/master/api/column_family.cc#L218) but swagger file [returns integer](https://github.com/scylladb/scylla/blob/master/api/api-doc/column_family.json#L1623) which maps to int32. 
Other endpoints around may also have wrong return type.

amnonh:
In swagger 1.2 int is defined as int32.

We originally used int following the jmx definition, in practice
internally we use uint and int64 in many places.

While the API format the type correctly, an external system that uses
swagger-based code generator can face a type issue problem.

This patch replace all use of int in a return type with long that is defined as int64.

Changing the return type, have no impact on the system, but it does help
external systems that use code generator from swagger.

Fixes #5347

### DIFF
--- a/api/api-doc/cache_service.json
+++ b/api/api-doc/cache_service.json
@@ -13,7 +13,7 @@
             {
                "method":"GET",
                "summary":"get row cache save period in seconds",
-               "type":"int",
+               "type": "long",
                "nickname":"get_row_cache_save_period_in_seconds",
                "produces":[
                   "application/json"
@@ -35,7 +35,7 @@
                      "description":"row cache save period in seconds",
                      "required":true,
                      "allowMultiple":false,
-                     "type":"int",
+                     "type": "long",
                      "paramType":"query"
                   }
                ]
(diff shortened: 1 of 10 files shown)


amnonh:
@slivne it's worth backporting

avikivity:
Where are these defined? In https://swagger.io/docs/specification/data-models/data-types/#numbers I see "number" and "integer", not "int" and "long".

amnonh:
Remember that the v1 of the API uses Swagger 1.2:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/1.2.md

slivne:
@penberg ping

penberg:
For the record, it looks like roughly 30% of our API methods in scylla-jmx use 64-bit integers:

```
[penberg@nero scylla-jmx]$ git grep "public int" | wc -l
125
[penberg@nero scylla-jmx]$ git grep "public long" | wc -l
71
```

But, using a 64-bit integer for 32-bit integers should be safe as Amnon says, so I'm merging this pull request.

penberg:
@amnonh Please use `next` as the target for future pull requests.

PR #5381, opened 2019-11-28
Redis：fix the options related to Redis API, fix the DEL and GET command

fastio:
* Rename the options related to Redis API, and describe them clearly.
   * Rename `redis_transport_port` to `redis_port`
   * Rename `redis_transport_port_ssl` to `redis_ssl_port`
   * Rename `redis_default_database_count` to  `redis_database_count`
   * Remove unnecessary option `enable_redis_protocol`
   * Modify the default value of opition `redis_read_consistency_level` and `redis_write_consistency_level` to `LOCAL_QUORUM`

* Fix the DEL command: support to delete mutilple keys in one command.
* Fix the GET command: return the empty string when the required key is not exists.
* Fix the redis-test/test_del_non_existent_key: mark xfail.

### DIFF
--- a/docs/redis/redis.md
+++ b/docs/redis/redis.md
@@ -63,11 +63,13 @@ queries the mapping information between hash slots and Server nodes. We
 just fill the mapping with random Scylla node IP address.
 
 Before you can start using Scylla with Redis API, you must set the
-`enable_redis_protocol` configuration option (visa the command line or YAML),
-to `true`, which default is `false`. By default, the Scylla will listen on the
-port (default is 6379, you can set the `redis_transport_prot` to other value).
+`redis-port` configuration option (visa the command line or YAML),
+to available port. By default, the Redis API is disabled. 
+In Scylla, SSL for Redis API is supported. To enable it,
+you must set the `redis-ssl-port` configuration option to available port.
+This feature is disabled by default.
 
-With Redis enabled, every Scylla node listens for Redis requests on this port.
+With Redis enabled, every Scylla node listens for Redis requests on the port.
 These requests, in [RESP](https://redis.io/topics/protocol) format over TCP,
 are parsed and result in calls to internal Scylla C++ functions.
 
--- a/main.cc
+++ b/main.cc
@@ -1108,9 +1108,8 @@ int main(int ac, char** av) {
             }
 
             static redis_service redis;
-            if (cfg->enable_redis_protocol()) {
+            if (cfg->redis_port() || cfg->redis_ssl_port()) {
                 redis.init(std::ref(proxy), std::ref(db), std::ref(auth_service), *cfg).get();
-                startlog.info("Redis server listening on port {}", cfg->redis_transport_port());
             }
 
             if (cfg->defragment_memory_on_idle()) {
@@ -1158,7 +1157,7 @@ int main(int ac, char** av) {
             });
 
             auto stop_redis_service = defer_verbose_shutdown("redis service", [&cfg] {
-                if (cfg->enable_redis_protocol()) {
+                if (cfg->redis_port() || cfg->redis_ssl_port()) {
                     redis.stop().get();
                 }
             });
--- a/redis/commands/del.cc
+++ b/redis/commands/del.cc
@@ -36,16 +36,18 @@ namespace commands {
 
 shared_ptr<abstract_command> del::prepare(service::storage_proxy& proxy, request&& req)
 {
-    if (req.arguments_size() != 1) {
-        throw wrong_arguments_exception(1, req.arguments_size(), req._command);
+    if (req.arguments_size() == 0) {
+        throw wrong_number_of_arguments_exception(req._command);
     }
-    return seastar::make_shared<del> (std::move(req._command), std::move(req._args[0]));
+    return seastar::make_shared<del> (std::move(req._command), std::move(req._args));
 }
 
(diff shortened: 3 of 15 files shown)


main.cc
```diff
@@ -1108,9 +1108,9 @@ int main(int ac, char** av) {
             }
 
             static redis_service redis;
-            if (cfg->enable_redis_protocol()) {
+            if (cfg->redis_port()) {
```
nyh: But, doesn't this mean that Redis will now be enabled by default, because redis_port has a default?
Because the Redis support is experimental, we don't want to enable it by default.
So we either need redis_port to default to 0, or keep a separate enable-redis option...

fastio: I think it's better to keep a separate enable-redis option.

fastio:
Change logs:

* Change the default of redis-port to 0, which means that Redis will now be enabled by default.

redis/commands/del.cc
```diff
@@ -36,16 +36,18 @@ namespace commands {
...
 }
 
 future<redis_message> del::execute(service::storage_proxy& proxy, redis::redis_options& options, service_permit permit)
 {
-    return redis::delete_object(proxy, options, std::move(_key), permit).then([] {
-       return redis_message::one();
+    //FIXME: We should return the count of the actually deleted keys.
+    auto size = _keys.size();
+    return redis::delete_objects(proxy, options, std::move(_keys), permit).then([size] {
+       return redis_message::number(size);
```
amoskong: The returned size is the number of really deleted keys.

For example:
- try to delete  a, b, c
- only a, and c eixst
- will return 2, not 3.

redis/reply.hh
```diff
@@ -60,8 +60,18 @@ public:
...
-        return err();
+    static future<redis_message> nil() {
+        auto m = make_lw_shared<scattered_message<char>> ();
+        m->append_static("$-1\r\n");
+        return make_ready_future<redis_message>(m);
+    }
+    static future<redis_message> err() {
+        return zero();
+    }
+    static future<redis_message> number(size_t n) {
```
amoskong: How about calling number() in zero(), and one()?

redis/commands/del.cc
nyh: This is why he has a FIXME on this...

redis/reply.hh
nyh: @amoskong this pull request seems to conflict with your one (which changed err/zero). It's confusing....

fastio:
@nyh 

> This is why he has a FIXME on this...

Current implement of DEL only supports to delete multiple keys in one command.
We should return the count of  really keys deleted to client in the future.

redis/commands/del.cc
fastio: Yes.

fastio:
@nyh 

> Looks good, but the patch that changes the options on how to enable redis should also update docs/redis/redis.md.

I have update the docs/redis/redis.md for changes of the options on how to enable redis.

docs/redis/redis.md
```diff
@@ -63,11 +63,13 @@ queries the mapping information between hash slots and Server nodes. We
 just fill the mapping with random Scylla node IP address.
 
 Before you can start using Scylla with Redis API, you must set the
-`enable_redis_protocol` configuration option (visa the command line or YAML),
-to `true`, which default is `false`. By default, the Scylla will listen on the
-port (default is 6379, you can set the `redis_transport_prot` to other value).
+`redis-port` configuration option (visa the command line or YAML),
+to availiable port. By default, the Redis API is disable.
```
nyh: is **disabled**

docs/redis/redis.md
```diff
@@ -63,11 +63,13 @@ queries the mapping information between hash slots and Server nodes. We
...
 
 Before you can start using Scylla with Redis API, you must set the
-`enable_redis_protocol` configuration option (visa the command line or YAML),
-to `true`, which default is `false`. By default, the Scylla will listen on the
-port (default is 6379, you can set the `redis_transport_prot` to other value).
+`redis-port` configuration option (visa the command line or YAML),
+to availiable port. By default, the Redis API is disable. 
+In Scylla, SSL for Redis API is supported. To enable it,
+you must set the `redis-ssl-port` configuration option to availiable port.
+This feature is disable by default too.
```
nyh: is **disabled**

docs/redis/redis.md
```diff
@@ -63,11 +63,13 @@ queries the mapping information between hash slots and Server nodes. We
 just fill the mapping with random Scylla node IP address.
 
 Before you can start using Scylla with Redis API, you must set the
-`enable_redis_protocol` configuration option (visa the command line or YAML),
-to `true`, which default is `false`. By default, the Scylla will listen on the
-port (default is 6379, you can set the `redis_transport_prot` to other value).
+`redis-port` configuration option (visa the command line or YAML),
```
nyh: You put this change in the wrong commit....

nyh:
@fastio I see you pushed a new version but I don't see how it answers my request to move the README.md changes to the appropriate commit

fastio:
Change logs:
* Fix the typo in the docs/redis/redis.md
* Push the changes of docs/redis/redis.md into the correct commit.

fastio:
@nyh Thanks for your review.
I have moved the changes of docs/redis/redis.md to the correct commit.

fastio:
Ping

PR #5387, opened 2019-11-29
redis: implement echo command

amoskong:
This patch implemented echo command, which return the string back to client.

Reference:
- https://redis.io/commands/echo

/CC @fastio @nyh

### DIFF
--- a/redis/commands/echo.cc
+++ b/redis/commands/echo.cc
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2019 ScyllaDB
+ */
+
+/*
+ * This file is part of Scylla.
+ *
+ * Scylla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Scylla is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "redis/commands/echo.hh"
+#include "seastar/core/shared_ptr.hh"
+#include "redis/request.hh"
+#include "redis/reply.hh"
+#include "types.hh"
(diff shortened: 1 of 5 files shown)


redis/commands/echo.hh
```diff
@@ -0,0 +1,50 @@
...
+    bytes _str;
+public:
+    static shared_ptr<abstract_command> prepare(service::storage_proxy& proxy, request&& req);
+    echo(bytes&& name, bytes&& str)
+        : abstract_command(std::move(name))
+        , _str(std::move(str))
+    {
+    }
+
+    ~echo()
```
denesb: No need to declare this.

redis/commands/echo.hh
```diff
@@ -0,0 +1,50 @@
...
+    {
+    }
+
+    future<redis_message> execute(service::storage_proxy&, redis_options&, servcie_permit) override;
```
denesb: We prepend `virtual` to overriden virtual method declarations.

redis/commands/echo.cc
```diff
@@ -0,0 +1,48 @@
...
+#include "types.hh"
+#include "service/storage_proxy.hh"
+#include "redis/options.hh"
+#include "service_permit.hh"
+
+namespace redis{
+namespace commands {
+
+shared_ptr<abstract_command> echo::prepare(service::storage_proxy& proxy, request&& req)
+{
```
denesb: Opening parens on the same line as the function signature.

redis/commands/echo.cc
```diff
@@ -0,0 +1,48 @@
...
+    if (req.arguments_size() != 1) {
+        throw wrong_arguments_exception(1, req.arguments_size(), req._command);
+    }
+    request seastar::make_shared<echo> (std::moved(req._command), std::move(req._args[0]);
+}
+
+future<redis_message> echo::execute(service::storage_proxy&, redis::redis_options&, service_permit)
+{
```
denesb: Here too.

redis/commands/echo.cc
```diff
@@ -0,0 +1,48 @@
...
+#include "redis/commands/echo.hh"
+#include "seastar/core/shared_ptr.hh"
+#include "redis/request.hh"
+#include "redis/reply.hh"
+#include "types.hh"
+#include "service/storage_proxy.hh"
+#include "redis/options.hh"
+#include "service_permit.hh"
+
+namespace redis{
```
denesb: Space between `redis` and `{`.

amoskong:
@denesb Thanks for your review, I just updated the patches.

PR #5393, opened 2019-11-30
redis: skip processing request of EOF

amoskong:
When I test the redis cmd by echo and nc, there is a redundant error in the end.
    I checked by strace, currently if client read nothing from stdin, it will
    shutdown the socket, redis server will read nothing (0 byte) from socket. But
    it tries to process the empty command and returns an error.
    
    $ echo -n -e '*1\r\n$4\r\nping\r\n' |strace nc localhost 6379
    | ...
    |    read(0, "*1\r\n$4\r\nping\r\n", 8192)   = 14
    |    select(5, [4], [4], [], NULL)           = 1 (out [4])
    |>>> sendto(4, "*1\r\n$4\r\nping\r\n", 14, 0, NULL, 0) = 14
    |    select(5, [0 4], [], [], NULL)          = 1 (in [0])
    |    recvfrom(0, 0x7ffe4d5b6c70, 8192, 0, 0x7ffe4d5b6bf0, 0x7ffe4d5b6bec) = -1 ENOTSOCK (Socket operation on non-socket)
    |    read(0, "", 8192)                       = 0
    |>>> shutdown(4, SHUT_WR)                    = 0
    |    select(5, [4], [], [], NULL)            = 1 (in [4])
    |    recvfrom(4, "+PONG\r\n-ERR unknown command ''\r\n", 8192, 0, 0x7ffe4d5b6bf0, [0]) = 32
    |    write(1, "+PONG\r\n-ERR unknown command ''\r\n", 32+PONG
    |    -ERR unknown command ''
    |    ) = 32
    |    select(5, [4], [], [], NULL)            = 1 (in [4])
    |    recvfrom(4, "", 8192, 0, 0x7ffe4d5b6bf0, [0]) = 0
    |    close(1)                                = 0
    |    close(4)                                = 0
    
    Current result:
      $ echo -n -e '' |nc localhost 6379
      -ERR unknown command ''
      $ echo -n -e '*1\r\n$4\r\nping\r\n' |nc localhost 6379
      +PONG
      -ERR unknown command ''
    
    Expected:
      $ echo -n -e '' |nc localhost 6379
      $ echo -n -e '*1\r\n$4\r\nping\r\n' |nc localhost 6379
      +PONG

/CC @fastio

### DIFF
--- a/redis/protocol_parser.rl
+++ b/redis/protocol_parser.rl
@@ -81,7 +81,7 @@ blob := any+ >start_blob $advance_blob;
 command := any+ >start_command $advance_command;
 arg = '$' u32 crlf ${ _arg_size = _u32;};
 
-main := (args_count (arg @{fcall command; } crlf) (arg @{fcall blob; } crlf)+) ${_req._state = request_state::ok;};
+main := (args_count (arg @{fcall command; } crlf) (arg @{fcall blob; } crlf)+) ${_req._state = request_state::ok;} >eof{_req._state = request_state::eof;};
 
 prepush {
     prepush();
--- a/redis-test/test_raw_cmd.py
+++ b/redis-test/test_raw_cmd.py
@@ -1,0 +1,74 @@
+#
+# Copyright (C) 2019 ScyllaDB
+#
+#
+# This file is part of Scylla.
+#
+# Scylla is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Scylla is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+import socket
+import logging
+
+logger = logging.getLogger('redis-test')
+
+
+class RedisSocket:
+
+    def __init__(self, host='localhost', port=6379):
+        self.socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.connect(host, port)
+
+    def connect(self, host, port):
+        self.socket.connect((host, port))
+
+    def send(self, content=''):
+        self.socket.send(content)
+
+    def recv(self, buf_len=1024):
+        return self.socket.recv(buf_len)
+
+    def shutdown(self):
+        self.socket.shutdown(socket.SHUT_WR)
+
+    def close(self):
+        self.socket.close()
+
+def verify_cmd_response(cmd, expect_ret, shutdown=False):
+    rs = RedisSocket()
+    rs.send(cmd.encode())
+    if shutdown:
+        rs.shutdown()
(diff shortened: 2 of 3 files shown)


redis/commands/del.cc
```diff
@@ -44,9 +45,16 @@ shared_ptr<abstract_command> del::prepare(service::storage_proxy& proxy, request
 
 future<redis_message> del::execute(service::storage_proxy& proxy, redis::redis_options& options, service_permit permit)
 {
-    return redis::delete_object(proxy, options, std::move(_key), permit).then([] {
-       return redis_message::one();
+    // FIXME: del command doesn't support to delete multiple keys once
```
nyh: Amos, can you please separate unrelated patches to separate pull requests or mailing-list patches?
In this series, some of your patches I can commit in an instant and some I have more problems with, and I'm not sure if I should break up your pull request by committing some of the patches (technically I can, but I'm not sure I should).

redis/commands/del.cc
```diff
@@ -44,9 +45,16 @@ shared_ptr<abstract_command> del::prepare(service::storage_proxy& proxy, request
...
+    return redis::read_strings(proxy, options, _key, permit).then([&proxy, &options, permit, this] (auto result) {
```
nyh: why read_strings? What if the object being deleted is of a different type?

redis/protocol_parser.rl
```diff
@@ -81,7 +81,7 @@ blob := any+ >start_blob $advance_blob;
 command := any+ >start_command $advance_command;
 arg = '$' u32 crlf ${ _arg_size = _u32;};
 
-main := (args_count (arg @{fcall command; } crlf) (arg @{fcall blob; } crlf)+) ${_req._state = request_state::ok;};
+main := (args_count (arg @{fcall command; } crlf) (arg @{fcall blob; } crlf)+) ${_req._state = request_state::ok;} >eof{_req._state = request_state::eof;};
```
nyh: If it's not too hard, would be nice to have test case for this, not just a commit-log comment.

redis/commands/get.cc
```diff
@@ -48,8 +48,7 @@ future<redis_message> get::execute(service::storage_proxy& proxy, redis::redis_o
         if (result->has_result()) {
             return redis_message::make_strings_result(std::move(result->result()));
         }
-        // FIXME: we should returns nil string if key does not exist
-        return redis_message::err();
+        return redis_message::nil();
```
nyh: Does this fix issue https://github.com/scylladb/scylla/issues/5317 ?
If so, please add the "Fixes" line.
If not, can we add a new test for what you're fixing here?

redis/commands/del.cc
amoskong: OK, I will split it to multiple PRs.

redis/protocol_parser.rl
amoskong: Currently we test by redis module, this kind of problem can only be found by sending raw string & checking original response.

I will add a series of my tests to redis-test/, thanks.

redis/commands/get.cc
amoskong: Yes, I will update the commitlog.

I saw @fastio already posted another PR to fix this issue, I will remove this patch from this PR.

redis/commands/del.cc
amoskong: will address this issue by another pr.

amoskong:
Hi @nyh @fastio 

I updated the PR, only left an independent fix, and added a test for this fix.

PR #5395, opened 2019-12-01
gdb: change 'scylla thread' command to access fs_base register directly

avikivity:
Currently, 'scylla thread' uses arch_prctl() to extract the value of
fsbase, used to reference thread local variables. gdb 8 added support
for directly accessing the value as $fs_base, so use that instead. This
works from core dumps as well as live processes, as you don't need to
execute inferior functions.

The patch is required for debugging threads in core dumps, but not
sufficient, as we still need to set $rip and $rsp, and gdb still[1]
doesn't allow this.

[1] https://sourceware.org/bugzilla/show_bug.cgi?id=9370

### DIFF
--- a/scylla-gdb.py
+++ b/scylla-gdb.py
@@ -2016,15 +2016,7 @@ class seastar_thread_context(object):
             gdb.execute('set $%s = %s' % (reg, value))
 
     def get_fs_base(self):
-        holder_addr = get_seastar_memory_start_and_size()[0]
-        holder = gdb.Value(holder_addr).reinterpret_cast(self.ulong_type.pointer())
-        saved = holder.dereference()
-        gdb.execute('set *(void**)%s = 0' % holder_addr)
-        if gdb.parse_and_eval('arch_prctl(0x1003, %d)' % holder_addr) != 0:
-            raise Exception('arch_prctl() failed')
-        fs_base = holder.dereference()
-        gdb.execute('set *(void**)%s = %s' % (holder_addr, saved))
-        return fs_base
+        return gdb.parse_and_eval('$fs_base')
 
     def regs_from_jmpbuf(self, jmpbuf):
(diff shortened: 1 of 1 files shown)


avikivity:
Note: Fedora 31's gdb fails for me before and after the patch. See https://sourceware.org/bugzilla/show_bug.cgi?id=25239 for details. I believe the failure is orthogonal to the patch.

espindola:
It still fails for me with Fedora 30's gdb:

Switched to thread 24, (seastar::thread_context*) 0x609000087800
  File "scylla-gdb.py", line 2122, in invoke
    active_thread_context.__enter__()
  File "scylla-gdb.py", line 2056, in __enter__
    self.restore_regs(self.new_regs)
  File "scylla-gdb.py", line 2016, in restore_regs
    gdb.execute('set $%s = %s' % (reg, value))
gdb.error: You can't do that without a process to debug.
Error occurred in Python: You can't do that without a process to debug.

In a running process I got the same error as you, and without this patch
I got

Error occurred in Python: 'arch_prctl' has unknown return type; cast the call to its declared return type

In sumary, I think this is a step in the right direction.

LGTM

Thanks,
Rafael

denesb:
@avikivity does this make my patch "[PATCH v1] scylla-gdb.py: document and add safety net to seastar::thread related commands" obsolete, i.e. does it make using this command safe (no risk of crashing the process)?

avikivity:
No, it's still risky.

avikivity:
This patch is too optimistic. Yes, it avoids the need to call a function, but gdb still doesn't allow writing to registers while debugging a core dump.

avikivity:
https://sourceware.org/bugzilla/show_bug.cgi?id=9370

slivne:
@avikivity if this pull request is not going to be merged - lets close it.

avikivity:
The patch still has merit. I'll reword the commit message though.

avikivity:
Updated.

PR #5402, opened 2019-12-02
Named semaphores in concurrency reader, segment_manager and region_group 

### LINKED ISSUE #5281: name reader concurrency semaphores

jul-stas:
Selected semaphores' names are now included in exception messages in case of timeout or when admission queue overflows.
Resolves #5281

### DIFF
--- a/database.cc
+++ b/database.cc
@@ -206,8 +206,12 @@ database::database(const db::config& cfg, database_config dbcfg)
         })
     // No timeouts or queue length limits - a failure here can kill an entire repair.
     // Trust the caller to limit concurrency.
-    , _streaming_concurrency_sem(max_count_streaming_concurrent_reads, max_memory_streaming_concurrent_reads())
-    , _system_read_concurrency_sem(max_count_system_concurrent_reads, max_memory_system_concurrent_reads())
+    , _streaming_concurrency_sem(
+            max_count_streaming_concurrent_reads,
+            max_memory_streaming_concurrent_reads())
+    , _system_read_concurrency_sem(
+            max_count_system_concurrent_reads,
+            max_memory_system_concurrent_reads())
     , _data_query_stage("data_query", &column_family::query)
     , _mutation_query_stage()
     , _apply_stage("db_apply", &database::do_apply)
--- a/database.cc
+++ b/database.cc
@@ -199,19 +199,21 @@ database::database(const db::config& cfg, database_config dbcfg)
     }))
     , _read_concurrency_sem(max_count_concurrent_reads,
         max_memory_concurrent_reads(),
+        "_read_concurrency_sem",
         max_inactive_queue_length(),
         [this] {
             ++_stats->sstable_read_queue_overloaded;
-            return std::make_exception_ptr(std::runtime_error("sstable read queue overloaded"));
         })
     // No timeouts or queue length limits - a failure here can kill an entire repair.
     // Trust the caller to limit concurrency.
     , _streaming_concurrency_sem(
             max_count_streaming_concurrent_reads,
-            max_memory_streaming_concurrent_reads())
+            max_memory_streaming_concurrent_reads(),
+            "_streaming_concurrency_sem")
     , _system_read_concurrency_sem(
             max_count_system_concurrent_reads,
-            max_memory_system_concurrent_reads())
+            max_memory_system_concurrent_reads(),
+            "_system_read_concurrency_sem")
     , _data_query_stage("data_query", &column_family::query)
     , _mutation_query_stage()
     , _apply_stage("db_apply", &database::do_apply)
--- a/db/commitlog/commitlog.cc
+++ b/db/commitlog/commitlog.cc
@@ -205,8 +205,19 @@ public:
     bool _shutdown = false;
     std::optional<shared_promise<>> _shutdown_promise = {};
 
+    struct request_controller_timeout_exception_factory {
+        class request_controller_timed_out_error : public timed_out_error {
+        public:
+            virtual const char* what() const noexcept override {
+                return "commitlog: timed out";
+            }
+        };
+        static auto timeout() {
+            return request_controller_timed_out_error();
+        }
+    };
     // Allocation must throw timed_out_error by contract.
-    using timeout_exception_factory = default_timeout_exception_factory;
+    using timeout_exception_factory = request_controller_timeout_exception_factory;
 
     basic_semaphore<timeout_exception_factory> _flush_semaphore;
 
@@ -1023,7 +1034,7 @@ db::commitlog::segment_manager::segment_manager(config c)
     // an existing in-flight buffer. Since we'll force the cycling() of any buffer that is bigger
     // than default_size at the end of the allocation, that allows for every valid mutation to
     // always be admitted for processing.
-    , _request_controller(max_request_controller_units())
+    , _request_controller(max_request_controller_units(), request_controller_timeout_exception_factory{})
     , _reserve_segments(1)
     , _reserve_replenisher(make_ready_future<>())
 {
--- a/reader_concurrency_semaphore.cc
+++ b/reader_concurrency_semaphore.cc
@@ -20,6 +20,7 @@
  */
 
 #include <seastar/core/reactor.hh>
+#include <seastar/core/print.hh>
 
 #include "reader_concurrency_semaphore.hh"
 
@@ -77,7 +78,12 @@ bool reader_concurrency_semaphore::try_evict_one_inactive_read() {
 future<lw_shared_ptr<reader_concurrency_semaphore::reader_permit>> reader_concurrency_semaphore::wait_admission(size_t memory,
         db::timeout_clock::time_point timeout) {
     if (_wait_list.size() >= _max_queue_length) {
-        return make_exception_future<lw_shared_ptr<reader_permit>>(_make_queue_overloaded_exception());
+        if (_prethrow_action) {
+            _prethrow_action();
+        }
+        return make_exception_future<lw_shared_ptr<reader_permit>>(
+                std::make_exception_ptr(std::runtime_error(
+                        format("{}: restricted mutation reader queue overload", _name))));
     }
     auto r = resources(1, static_cast<ssize_t>(memory));
     auto it = _inactive_reads.begin();
--- a/reader_concurrency_semaphore.hh
+++ b/reader_concurrency_semaphore.hh
@@ -39,16 +39,16 @@
 /// construction. New readers will only be admitted when there is both
 /// enough count and memory units available. Readers are admitted in
 /// FIFO order.
-/// It's possible to specify the maximum allowed number of waiting
+/// Semaphore's `name` must be provided in ctor and its only purpose is
+/// to increase readability of exceptions: both timeout exceptions and
+/// queue overflow exceptions (read below) include this `name` in messages.
+/// It's also possible to specify the maximum allowed number of waiting
 /// readers by the `max_queue_length` constructor parameter. When the
-/// number waiting readers would be equal or greater than this number
-/// (when calling `wait_admission()`) an exception will be thrown.
-/// The type of the exception and optionally some additional code
-/// that should be executed when this happens can be customized by the
-/// `raise_queue_overloaded_exception` constructor parameter. This
-/// function will be called every time the queue limit is surpassed.
-/// It is expected to return an `std::exception_ptr` that will be
-/// injected into the future.
+/// number of waiting readers becomes equal or greater than
+/// `max_queue_length` (upon calling `wait_admission()`) an exception of
+/// type `std::runtime_error` is thrown. Optionally, some additional
+/// code can be executed just before throwing (`prethrow_action` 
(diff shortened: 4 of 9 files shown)


database.cc
```diff
@@ -199,19 +199,22 @@ database::database(const db::config& cfg, database_config dbcfg)
     }))
     , _read_concurrency_sem(max_count_concurrent_reads,
         max_memory_concurrent_reads(),
+        "_read_concurrency_sem",
         max_inactive_queue_length(),
         [this] {
             ++_stats->sstable_read_queue_overloaded;
-            return std::make_exception_ptr(std::runtime_error("sstable read queue overloaded"));
+            return std::make_exception_ptr(std::runtime_error("_read_concurrency_sem: sstable read queue overloaded"));
```
denesb: Just thinking out loud here: now that the semaphore has a name, maybe it could generate this error message on its own, there is no need to pass this factory method any-more.

reader_concurrency_semaphore.hh
```diff
@@ -195,12 +206,20 @@ public:
...
+        , _name(std::move(name))
         , _max_queue_length(max_queue_length)
-        , _make_queue_overloaded_exception(raise_queue_overloaded_exception) {
-    }
+        , _make_queue_overloaded_exception(custom_queue_overloaded_handler
+                ? custom_queue_overloaded_handler
+                : [prefix = _name.empty() ? "" : ("semaphore `" + _name + "`: ")] {
+                    return std::make_exception_ptr(std::runtime_error(
+                            prefix + "restricted mutation reader queue overload"));
+                })
```
denesb: Referring back to my previous comment about this factory method, I think we can just ditch it. Compose the message ourselves and export a counter.

reader_concurrency_semaphore.hh
```diff
@@ -195,12 +206,20 @@ public:
 
     reader_concurrency_semaphore(int count,
             ssize_t memory,
+            sstring name = "",
```
denesb: Let's make this mandatory (no default) and remove all the special logic for a `""` name.

db/commitlog/commitlog.cc
```diff
@@ -205,8 +205,19 @@ class db::commitlog::segment_manager : public ::enable_shared_from_this<segment_
     bool _shutdown = false;
     std::optional<shared_promise<>> _shutdown_promise = {};
 
+    struct request_controller_timeout_exception_factory {
+        class request_controller_timed_out_error : public timed_out_error {
+        public:
+            const char* what() const noexcept override {
```
denesb: We prepend overrides of virtual methods with `virtual` keyword.

utils/logalloc.hh
```diff
@@ -208,7 +208,10 @@ class region_group {
         }
     };
 
-    struct on_request_expiry {
+    class on_request_expiry {
+        sstring _name;
+    public:
+        explicit on_request_expiry(sstring name) : _name(name) {}
```
denesb: `std::move(name)`

db/commitlog/commitlog.cc
jul-stas: OK, but why?

reader_concurrency_semaphore.hh
jul-stas: In general I agree - I kept it because it's more future-proof (so we can attach any action and throw custom exception). I'm not insisting, but it seemed fair enough for me to stay here.

db/commitlog/commitlog.cc
denesb: Because you only need to read one word to see that this is a virtual method vs. having to read all of them to the end.

reader_concurrency_semaphore.hh
denesb: The role of the concurrency semaphore is quite set, not much change is expected here. This factory is a remnant of an age where we used `seastar::semaphore` as the concurrency semaphore. I think it would be better to do it as part of this patch as it would ensure all error messages from the semaphore are consistent. It also would result in a smaller patch ;).

database.cc
jul-stas: Please see my comment below

reader_concurrency_semaphore.hh
jul-stas: I did a hybrid solution: I preserved ctor's last argument (callable), but now it's just a `prethrow_action` - no possibility to alter exception type/message there. I hope you're OK with that - after all, being able to run arbitrary code before throwing is a nice possibility to have (?)

reader_concurrency_semaphore.hh
jul-stas: Done

db/commitlog/commitlog.cc
jul-stas: There are people among us who read in opposite direction :)
Anyway, done

utils/logalloc.hh
jul-stas: I removed `name` in favor of static message (if this semaphore is used in one place only, noone would benefit from the ability to give it a generic `name`...)

reader_concurrency_semaphore.cc
```diff
@@ -74,10 +74,20 @@ bool reader_concurrency_semaphore::try_evict_one_inactive_read() {
...
+}
+
 future<lw_shared_ptr<reader_concurrency_semaphore::reader_permit>> reader_concurrency_semaphore::wait_admission(size_t memory,
         db::timeout_clock::time_point timeout) {
     if (_wait_list.size() >= _max_queue_length) {
-        return make_exception_future<lw_shared_ptr<reader_permit>>(_make_queue_overloaded_exception());
+        if (_prethrow_action) {
+            _prethrow_action();
+        }
+        return make_queue_overflow_exception_future();
```
denesb: nitpick: the content of this function might as well be inlined here.

reader_concurrency_semaphore.hh
```diff
@@ -151,28 +152,33 @@ private:
...
+
 private:
     resources _resources;
 
     expiring_fifo<entry, expiry_handler, db::timeout_clock> _wait_list;
 
+    sstring _name;
     size_t _max_queue_length = std::numeric_limits<size_t>::max();
-    std::function<std::exception_ptr()> _make_queue_overloaded_exception;
+    std::function<void()> _prethrow_action = nullptr;
```
denesb: `std::function<>` has a default constructor, so this initialization is not needed.

reader_concurrency_semaphore.hh
```diff
@@ -195,18 +201,24 @@ public:
...
             size_t max_queue_length = std::numeric_limits<size_t>::max(),
-            std::function<std::exception_ptr()> raise_queue_overloaded_exception = default_make_queue_overloaded_exception)
+            std::function<void()> prethrow_action = nullptr)
         : _resources(count, memory)
+        , _wait_list(expiry_handler(name))
+        , _name(std::move(name))
         , _max_queue_length(max_queue_length)
-        , _make_queue_overloaded_exception(raise_queue_overloaded_exception) {
-    }
+        , _prethrow_action(prethrow_action)
```
denesb: `std::move(prethrow_action)`

reader_concurrency_semaphore.hh
```diff
@@ -195,18 +201,24 @@ public:
...
 
     /// Create a semaphore with practically unlimited count and memory.
     ///
     /// And conversely, no queue limit either.
     explicit reader_concurrency_semaphore(no_limits)
-        : reader_concurrency_semaphore(std::numeric_limits<int>::max(), std::numeric_limits<ssize_t>::max()) {
+        : reader_concurrency_semaphore(
+                std::numeric_limits<int>::max(),
+                std::numeric_limits<ssize_t>::max(),
+                format("unlimited RCS @{}", fmt::v5::ptr(this))) {
```
denesb: I think just `unlimited reader_concurrency_semaphore` will do. We don't expect to see an exception with this ever, since it is only used in tests and it has practically unlimited units so it will never time out. A simpler name will also not require adding the `print.hh` include.

db/commitlog/commitlog.cc
```diff
@@ -205,8 +205,19 @@ class db::commitlog::segment_manager : public ::enable_shared_from_this<segment_
     bool _shutdown = false;
     std::optional<shared_promise<>> _shutdown_promise = {};
 
+    struct request_controller_timeout_exception_factory {
+        class request_controller_timed_out_error : public timed_out_error {
+        public:
+            virtual const char* what() const noexcept override {
+                return "request_controller: timed out";
```
denesb: I think this message needs some more context, `request_controller` is a too generic name.

reader_concurrency_semaphore.cc
jul-stas: Done (inlined)

reader_concurrency_semaphore.hh
jul-stas: Removed (although I did it on purpose)

reader_concurrency_semaphore.hh
jul-stas: Removed address

db/commitlog/commitlog.cc
jul-stas: Having no better idea, I wrote `segment_manager::_request_controller`

tests/mutation_reader_test.cc
```diff
@@ -1079,7 +1079,7 @@ class dummy_file_impl : public file_impl {
 
 SEASTAR_TEST_CASE(reader_restriction_file_tracking) {
     return async([&] {
-        reader_concurrency_semaphore semaphore(100, 4 * 1024);
+        reader_concurrency_semaphore semaphore(100, 4 * 1024, "reader_restriction_file_tracking");
```
denesb: hint: `get_name()` will return the name of the test.

reader_concurrency_semaphore.hh
```diff
@@ -195,19 +197,23 @@ public:
...
     /// Create a semaphore with practically unlimited count and memory.
     ///
     /// And conversely, no queue limit either.
     explicit reader_concurrency_semaphore(no_limits)
-        : reader_concurrency_semaphore(std::numeric_limits<int>::max(), std::numeric_limits<ssize_t>::max()) {
-    }
+        : reader_concurrency_semaphore(
+                std::numeric_limits<int>::max(),
+                std::numeric_limits<ssize_t>::max(),
+                "unlimited RCS") {}
```
denesb: I would have liked it better without the acronym, but I won't insist as I don't expect to ever actually see it.

tests/mutation_reader_test.cc
jul-stas: Good one! Done.

reader_concurrency_semaphore.hh
jul-stas: This is an all inclusive, 5-star pull request - acronym removed.

denesb: Haha, thanks, I feel spoiled.

db/commitlog/commitlog.cc
```diff
@@ -205,8 +205,19 @@ class db::commitlog::segment_manager : public ::enable_shared_from_this<segment_
     bool _shutdown = false;
     std::optional<shared_promise<>> _shutdown_promise = {};
 
+    struct request_controller_timeout_exception_factory {
+        class request_controller_timed_out_error : public timed_out_error {
+        public:
+            virtual const char* what() const noexcept override {
+                return "segment_manager::_request_controller: timed out";
```
tgrabiec: Better: `commitlog: timed out`

Logs are also read by users, so it's better to use names of high-level concepts than code pointers.

utils/logalloc.hh
```diff
@@ -209,6 +209,12 @@ class region_group {
     };
 
     struct on_request_expiry {
+        class blocked_requests_timed_out_error : public timed_out_error {
+        public:
+            virtual const char* what() const noexcept override {
+                return "region_group::_blocked_requests timed out";
```
tgrabiec: For the same reasons as in the commitlog we would want to print `memtable: timed out`. That requires the ability to assign a name to a `region_group`.

Currently all region_groups are created on behalf of memtables (in dirty_memory_manager), but logalloc should not know about this as we want to keep it abstract and move to seastar one day.

db/commitlog/commitlog.cc
jul-stas: Done.

utils/logalloc.hh
jul-stas: Thanks, done.

slivne:
@tgrabiec are you merging this - do we need another maintainer review.

tgrabiec:
I see "1 pending reviewer" (haaawk). Maintainer review is necessary to merge, but not sufficient.

haaawk:
LGTM

haaawk:
@tgrabiec I think we can merge now ;)

PR #5405, opened 2019-12-03
redis: create all redis keyspaces as configured

amoskong:
Currently we always try to create 16 keyspaces for Redis, the config
    redis_default_database_count doesn't effect.
    
    SELECT command is supported, it would switch between keyspaces. The limitation
    is got from configuration. So if we config redis_default_database_count to be
    larger than 16, SELECT allows to switch to unexisting keyspace, and it will
    fail in read/write data.
    
    Example:
    | # scylla --developer-mode 1 --workdir ./ --enable-redis-protocol 1  --redis-default-database-count 20
    | # redis-cli
    | 127.0.0.1:6379> select 18
    | OK
    | 127.0.0.1:6379[18]> set a 1
    | (error) ERR Can't find a column family STRINGs in keyspace REDIS_18
    | 127.0.0.1:6379[18]> get a
    | (error) ERR Can't find a column family STRINGs in keyspace REDIS_18

This patchset fixed the issue and added a test for that.

/CC @fastio

### DIFF
--- a/redis-test/test_strings.py
+++ b/redis-test/test_strings.py
@@ -21,6 +21,7 @@
 import pytest
 import redis
 import logging
+import requests
 from util import random_string, connect
 
 logger = logging.getLogger('redis-test')
@@ -100,9 +101,15 @@ def test_select():
     r.delete(key)
     assert r.get(key) == None
 
-    logger.debug('Try to switch to invalid database 16')
+def test_select_invalid_db():
+    r = connect()
+    logger.debug('Assume that user will not set redis_database_count to be bigger as 100')
+    invalid_db_idx = 100
+
+    logger.debug('Try to switch to invalid database %d' % invalid_db_idx)
     try:
-        r.execute_command('SELECT 16')
-        raise Exception('Expect that `SELECT 16` does not work')
+        query = 'SELECT %d' % invalid_db_idx
+        r.execute_command(query)
+        raise Exception('Expect that `%s` does not work' % query)
     except redis.exceptions.ResponseError as ex:
         assert str(ex) == 'invalid DB index'
--- a/redis-test/test_strings.py
+++ b/redis-test/test_strings.py
@@ -113,3 +113,23 @@ def test_select_invalid_db():
         raise Exception('Expect that `%s` does not work' % query)
     except redis.exceptions.ResponseError as ex:
         assert str(ex) == 'invalid DB index'
+
+def test_redis_database_count():
+    """
+    Test enough keyspaces can be prepared if redis_database_count is set to be larger than default 16.
+    Test will fail if redis_database_count isn't set correctly.
+    """
+    logger.info('Make sure redis_database_count is set to be larger than default 16')
+
+    r = connect()
+
+    try:
+        r.execute_command('SELECT 16')  # Use the select to verify the redis_database_count configuration
+        key = random_string(10)
+        r.get(key)
+    except redis.exceptions.ResponseError as ex:
+        if str(ex) == 'invalid DB index':
+            raise Exception('Make sure redis_database_count is set to be larger than default 16')
+    finally:
+        logger.debug('Switch back to default database 0')
+        assert r.execute_command('SELECT 0') == 'OK'
--- a/redis/keyspace_utils.cc
+++ b/redis/keyspace_utils.cc
@@ -179,8 +179,9 @@ future<> create_keyspace_if_not_exists_impl(db::config& config, int default_repl
         logger.info("Create keyspace: {}, table: {} for redis.", ks_name, cf_name);
         return service::get_local_migration_manager().announce_new_column_family(schema, false);
     };
-    // create 16 default database for redis.
-    return parallel_for_each(boost::irange<unsigned>(0, 16), [keyspace_gen = std::move(keyspace_gen), table_gen = std::move(table_gen)] (auto c) {
+    // Create keyspaces as configured by redis_default_database_count, which are used as redis databases.
+    return parallel_for_each(boost::irange<unsigned>(0, config.redis_default_database_count()),
+                             [keyspace_gen = std::move(keyspace_gen), table_gen = std::move(table_gen)] (auto c) {
         auto ks_name = sprint("REDIS_%d", c);
         return keyspace_gen(ks_name).then([ks_name, table_gen] {
             return when_all_succeed(


redis-test/test_strings.py
```diff
@@ -89,8 +90,12 @@ def test_select():
     r.set(key, val)
     assert r.get(key) == val
 
-    logger.debug('Switch to database 1')
-    assert r.execute_command('SELECT 1') == 'OK'
+    response = requests.get('http://127.0.0.1:10000/v2/config/redis_default_database_count')
```
nyh: Tests are supposed to work correctly against both our code and redis (I hope that you and @fastio are verifying this, because I am not...). But this test will not work against Redis, because it doesn't have Scylla's REST API port...
In general, these Python functional tests aren't a good way to check configuration options, because they assume Redis is already running - and don't run it specially with different configuration options. So I think we can't test it like this.

redis-test/test_strings.py
```diff
@@ -113,3 +113,18 @@ def test_select():
         raise Exception('Expect that `%s` does not work' % query)
     except redis.exceptions.ResponseError as ex:
         assert str(ex) == 'invalid DB index'
+
+def test_redis_default_database_count():
+    response = requests.get('http://127.0.0.1:10000/v2/config/redis_default_database_count')
```
nyh: Again, this won't work against regular Redis :-(

redis-test/test_strings.py
amoskong: I always test with Redis manually, and the test code works with Redis server. So I can't break this by using Scylla REST API.

It's a big change to prepare the scylla-redis or real redis server by test code. We need additional PR to change this.

For testing this bug, we need to set redis_default_database_count to be larger than default 16. I will verify the count by `select 16` command. If it succeeded, it means the configuration is correct, and we can try to write data to trigger the bug, or verify the bug is fixed.

amoskong: Hi @nyh, just updated the tests, removed the Scylla API usage, and keep all tests work with native Redis server.

redis-test/test_strings.py
amoskong: fixed.

redis-test/test_strings.py
```diff
@@ -113,3 +113,23 @@ def test_select_invalid_db():
...
+
+    r = connect()
+
+    try:
+        r.execute_command('SELECT 16')  # Use the select to verify the redis_database_count configuration
+        key = random_string(10)
+        r.get(key)
+    except redis.exceptions.ResponseError as ex:
+        if str(ex) == 'invalid DB index':
+            raise Exception('Make sure redis_database_count is set to be larger than default 16')
```
nyh: So I don't understand, this test will fail if Scylla + redis is just run normally, with the default redis_database_count? It will also fail against regular redis if also run as usual (with a default number of 16)? Why do we want to see this failure all the time?

amoskong: The problem we fixed only occurred when redis_database_count is set to be larger than default 16 for Scylla-redis. The problem is not about verify index in switching db by select command, but not all keyspaces will be created as expected.

So this test will fail if Scylla-redis or regular Redis is just run normally, with the default redis_database_count. It doesn't satisfy our test request.

When we see the error of 'invalid DB index', they will get back to start scylla-redis or regular redis with larger redis_database_count.

Test example:
- start scylla-redis with 20 redis_database_count , or start regular redis with 20 dbs
- select 18  (will success)
- try to write data to database 17 (the idx starts from 0),
    -  expect to succeed
    - scylla-redis will fail before applying this PR.

amoskong: In future, we will start scylla or redis by test code, then we don't need to verify the current redis_database_count.

nyh: I'm not saying we can't have the tests assume Scylla or Redis was run with specific paramters - we also do that in the Alternator tests, where we assume they were run with a specific setup regarding authentication, otherwise the authentication tests will fail. HOWEVER, if you need to run Scylla and Redis with specific parameters for all the tests to pass, you should document these parameters in redis-test/README.md which explains how to run the tests.
And you need to consider whether complicating the test run for testing this small issue is worth it. Currently the tests don't run Scylla or Redis from scratch, so they don't check **any** of the redis command-line options - it isn't just redis_database_count which they don't check. So maybe it simply shouldn't check any of the command line parameters, and those need to be tested manually (or in the future, with a more general test framework)?

amoskong: The manual test was executed.

There are two options to process the error of database count isn't set to be larger than default.

- option 1: output the err msg and break the subtest, but it still PASSes.
- option 2: raise Exception with err msg and failed the job.

option 2 will reminder us to test manually or test with correct parameter(redis_database_count)

Current test is simple and crude, we can't improve it to be prefect without refactor and add a general test framework.

So I suggest to commit it as current, or remove the automation tests, and add test in future.

amoskong: @nyh ping

nyh: @amoskong I'll commit for now because you seem to really want this test, but I urge you to rethink this. I think you should do one of the following two options:
1. Document in redis-test/README.md how the user is expected to run Scylla and/or Redis in order for all the tests to pass (we do the same in Alternator, by the way - the user needs to run Scylla in a certain way if he wants the authentication tests to succeed).
2. Mark the tests as "xfail" - meaning that they will fail but this failure is expected, or "skip" - meaning that the test won't even run normally. In either case the user can ask pytest to manually run this test if he wants to try a special configuration with this test.

avikivity:
Dequeued, does not compile:

```
10:00:03  g++ -MD -MT build/release/redis/keyspace_utils.o -MF build/release/redis/keyspace_utils.o.d -I/jenkins/workspace/scylla-master/next/scylla/seastar/include -I/jenkins/workspace/scylla-master/next/scylla/build/release/seastar/gen/include -std=gnu++17 -U_FORTIFY_SOURCE -Wno-maybe-uninitialized -march=westmere -Werror=unused-result -fconcepts -DSEASTAR_API_LEVEL=2 -DSEASTAR_HAVE_GCC6_CONCEPTS -DSEASTAR_USE_STD_OPTIONAL_VARIANT_STRINGVIEW -DSEASTAR_HAVE_DPDK -DFMT_SHARED -I/usr/include/p11-kit-1 -march=westmere -DBOOST_TEST_DYN_LINK -DTHRIFT_USES_BOOST   -fvisibility=hidden -Wall -Werror -Wno-maybe-uninitialized -Wno-tautological-compare -Wno-missing-braces -Wno-misleading-indentation -Wno-overflow -Wno-noexcept-type -Wno-nonnull-compare -Wno-error=cpp -Wno-ignored-attributes -Wno-overloaded-virtual -Wno-stringop-overflow -Wno-error=deprecated-declarations -DXXH_PRIVATE_API -DSEASTAR_TESTING_MAIN -DHAVE_LIBSYSTEMD=1 -DHAVE_LZ4_COMPRESS_DEFAULT -O3 --param inline-unit-growth=300   -I. -I build/release/gen  -c -o build/release/redis/keyspace_utils.o redis/keyspace_utils.cc
10:00:03  redis/keyspace_utils.cc: In function 'seastar::future<> redis::create_keyspace_if_not_exists_impl(db::config&, int)':
10:00:03  redis/keyspace_utils.cc:183:64: error: 'class db::config' has no member named 'redis_default_database_count'; did you mean 'redis_database_count'?
10:00:03    183 |     return parallel_for_each(boost::irange<unsigned>(0, config.redis_default_database_count()),
10:00:03        |                                                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
10:00:03        |                                              
```

nyh:
Sorry Avi :-(
@amoskong can you please fix and resubmit? Thanks.
NOTE: This pull request is already marked "merged" and Github gives us no way to undo this. So you'll need to remember this is still open - github won't.

amoskong:
> Sorry Avi :-(
> @amoskong can you please fix and resubmit? Thanks.

Sorry the parameter is changed recently. I will send a new PR.

> NOTE: This pull request is already marked "merged" and Github gives us no way to undo this. So you'll need to remember this is still open - github won't.

Got it.

PR #5417, opened 2019-12-04
sysctl: reduce kernel tendency to swap anonymous pages relative to page cache

avikivity:
The vm.swappiness sysctl controls the kernel's prefernce for swapping
anonymous memory vs page cache. Since Scylla uses very large amounts
of anonymous memory, and tiny amounts of page cache, the correct setting
is to prefer swapping page cache. If the kernel swaps anonymous memory
the reactor will stall until the page fault is satisfied. On the other
hand, page cache pages usually belong to other applications, usually
backup processes that read Scylla files.

This setting has been used in production in Scylla Cloud for a while
with good results.

Users can opt out by not installing the scylla-kernel-conf package
(same as with the other kernel tunables).

### DIFF
--- a/dist/common/sysctl.d/99-scylla-vm.conf
+++ b/dist/common/sysctl.d/99-scylla-vm.conf
@@ -1,0 +1,8 @@
+# Prefer swapping page cache to anonymous memory. Scylla doesn't
+# use page cache, so if there is any significant amount, it's the
+# result of some side process like backup.
+#
+# Reduce swappiness to the bare minimum (0 is too low, as it won't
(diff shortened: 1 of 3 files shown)


glommer:
Changes look fine by me, Avi.

Please consider merging it.

glommer:
@nyh or @tgrabiec since Avi is unlikely to merge his own patch, please merge

nyh:
Committed through the github UI, but it wound up on the master branch instead of next! I'm now trying to decide how to resolve this :-(

nyh:
@avikivity reverted my commit to master, and committed this himself to next.

PR #5431, opened 2019-12-07
redis: implement mset command

amoskong:
Ref: https://redis.io/commands/mset
    
    Example:
    | redis> MSET key1 "Hello" key2 "World"
    | "OK"
    | redis> GET key1
    | "Hello"
    | redis> GET key2
    | "World"
    | redis>
    | redis> MSET
    | (error) ERR wrong number of arguments for 'mset' command
    | redis> MSET key1
    | (error) ERR wrong number of arguments for 'mset' command

/cc @nyh @fastio


amoskong:
@fastio posted a PR (Redis: Combine all the source files redis/commands/* into redis/comma… #5432) to combine the command files. I will rebase this PR when his PR is merged.

redis/commands/mset.cc
```diff
@@ -0,0 +1,60 @@
...
+#include "service/storage_proxy.hh"
+#include "service/client_state.hh"
+#include "redis/options.hh"
+#include "redis/mutation_utils.hh"
+
+namespace redis {
+
+namespace commands {
+
+shared_ptr<abstract_command> mset::prepare(service::storage_proxy& proxy, request&& req)
```
nyh: We should document somewhere, at least as a comment here but probably also in redis.md (which should have a compatibility section like we have in docs/alternator/alternator.md), that this "mset" implementation offers no atomicity guarantee. https://redis.io/commands/mset explains that "all given keys are set at once. It is not possible for clients to see that some of the keys were updated while others are unchanged." - but this is not promised by this implementation.

redis/commands/mset.cc
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    for (unsigned i = 0; i < req.arguments_size() / 2; i++) {
+        data_map.insert(std::make_pair(req._args[2 * i], req._args[2 * i + 1]));
+    }
+
+    return seastar::make_shared<mset> (std::move(req._command), std::move(data_map), 0);
+}
+
+future<redis_message> mset::execute(service::storage_proxy& proxy, redis::redis_options& options, service_permit permit)
```
nyh: A question not directly related to this patch but worth thinking about:
Why do we have this separation between the prepare() and execute() steps? Why can't we have a single function which takes a request and both parses it and runs the operation? It can remove a lot of boilerplate code from these operations.... I think it can remove about 2/3rds of the code of each operation...

redis/mutation_utils.cc
```diff
@@ -68,6 +68,11 @@ future<> write_strings(service::storage_proxy& proxy, redis::redis_options& opti
     return proxy.mutate(std::vector<mutation> {std::move(m)}, write_consistency_level, timeout, nullptr, permit);
 }
 
+future<> write_strings_pairs(service::storage_proxy& proxy, redis::redis_options& options, std::map<bytes, bytes>&& data_map, long ttl, service_permit permit) {
```
nyh: Why do you call this "strings_pairs"? Isn't this like write_strings, just a list of them instead of just one?
By the way, why do you use a std::map<> instead of just an std::vector or std::list - you don't really need these the keys sorted or (I think) duplicates avoided - can't you just take the same list of items you got from the user, and write them in the same order?

amoskong: The pair is the key and value, after parsing the arguments, I organizes them to multiple pairs of key - value. It's fine to pass two vectors or lists by two arguments as past.

redis/commands/mset.cc
amoskong: The purpose is splitting the disunity argument parse, each command has special argument, and parse logic.

command_factory->create() is used to parse the cmd itself.
command_factory->prepare() is used to parse the cmd arguments
abstract_command->execute() is used to really execute the cmd

If we merge two functions together, it will convert Object-Oriented to procedure-oriented.

@fastio what's your opinion?

redis/commands/mset.cc
amoskong: We are discussing the atomic issue by email.

Do you prefer to find a solution first?  or we can implement commands without care about the atomicity guarantee (document it somewhere, and fix it in future)?

nyh: I don't know how we can provide this guarantee at all, since in Scylla, the different keys can be in different nodes, and we would need full transaction support to implement this. So let's just start with documenting the problem (as you did), and fix it (much) later.

redis/commands/mset.cc
nyh: It's not really related to this patch, so we don't need to decide now. It's just a matter of style so I'll leave it to you guys to decide which style you like best.

redis/mutation_utils.cc
nyh: Ok, let's ignore these issues now, but for future consideration:
1. write_strings also takes a "pair" in this sense, so it's strange to call just one of the function "pairs"
2. You don't really need an std::map here, a std::list<std::pair>, or similar, would have been enough.

redis-test/test_strings.py
```diff
@@ -83,6 +85,20 @@ def test_echo():
...
     assert r.echo('hello world') == 'hello world'
 
+def test_mset():
+    r = connect()
+    values = {}
+    for i in range(5):
+        key = random_string(10)
+        val = random_string(1000)
+        values[key] = val
+    assert r.mset(values) == True
```
nyh: I'll commit this as-is, but a better test would not just write the keys - it would also read them back to see it actually works as expected and doesn't just ignore its arguments or do only the first one or anything like that.

nyh:
@amoskong can you please rebase this on top of next? I'm getting conflicts in redis/commands.cc and redis/commands.hh. Thanks.

avikivity:
"MSET is atomic, so all given keys are set at once. It is not possible for clients to see that some of the keys were updated while others are unchanged."

nyh:
@avikivity yes, that's what I pointed out in my own review, and I asked Amos to add a FIXME and a note about this in redis.md. We can't implement this in today's Scylla without full multi-partition transaction support.

redis-test/test_strings.py
amoskong: Fixed this in solving the conflicts.

```
    for k,v in values.items():
        assert r.get(k) == v
```

redis/mutation_utils.cc
amoskong: I will fix this by another pr.

amoskong:
> @amoskong can you please rebase this on top of next? I'm getting conflicts in redis/commands.cc and redis/commands.hh. Thanks.

The patch to combine command files hasn't merged to master, so it has conflict.
Just rebased this PR to next.

redis/commands.cc
```diff
@@ -129,6 +129,26 @@ future<redis_message> echo::execute(service::storage_proxy&, redis::redis_option
...
 }
 
+shared_ptr<abstract_command> mset::prepare(service::storage_proxy& proxy, request&& req) {
+    std::map<bytes, bytes> data_map;
+    if (req.arguments_size() == 0 || req.arguments_size() % 2 != 0) {
+        throw wrong_number_of_arguments_exception(req._command);
+    }
+
+    for (unsigned i = 0; i < req.arguments_size() / 2; i++) {
+        data_map.insert(std::make_pair(req._args[2 * i], req._args[2 * i + 1]));
```
denesb: `data_map.emplace(req._args[2 * i], req._args[2 * i + 1]);`

redis/commands.cc
```diff
@@ -129,6 +129,26 @@ future<redis_message> echo::execute(service::storage_proxy&, redis::redis_option
...
+    }
+
+    return seastar::make_shared<mset> (std::move(req._command), std::move(data_map), 0);
```
denesb: Space between '>' and '(' not needed.

redis/mutation_utils.cc
```diff
@@ -68,6 +68,11 @@ future<> write_strings(service::storage_proxy& proxy, redis::redis_options& opti
...
+	return parallel_for_each(data_map.begin(), data_map.end(), [&proxy, &options, ttl, permit] (auto &pair) {
+		return write_strings(proxy, options, to_bytes(pair.first), to_bytes(pair.second), ttl, permit);
```
denesb: Why is this `to_bytes()` needed, aren't both elements of the pair of type `bytes`?

redis/commands.cc
```diff
@@ -129,6 +129,26 @@ future<redis_message> echo::execute(service::storage_proxy&, redis::redis_option
...
+}
+
+future<redis_message> mset::execute(service::storage_proxy& proxy, redis::redis_options& options, service_permit permit) {
+    return redis::write_strings_pairs(proxy, options, std::move(_data_map), _ttl, permit).then([] () {
+        return redis_message::ok();
```
denesb: Contrary to redis, in Scylla such a write *can* fail. I don't know how to reconcile this with the required `ok` return status of the `mset` command.

docs/redis/redis.md
```diff
@@ -362,3 +362,11 @@ we split the key set into 5 different Scylla table. The shared commands
...
 which allows user to remove any key in Redis database, should delete the
 given key from 5 different Scylla tables.
+
+### 5.7 Atomicity Guarantee
+
+* "mset" implementation offers no atomicity guarantee.
+
+https://redis.io/commands/mset explains that "all given keys are set at once.
+It is not possible for clients to see that some of the keys were updated while
+others are unchanged." - but this is not promised by this implementation.
```
denesb: It should also be noted that Scylla cannot guarantee that the write will succeed, like redis' implementation does.

redis/commands.cc
nyh: Also I still think this can be a std::vector<std::pair<bytes,bytes>> (or std::list or std::dequeue similar), there is no reason for this to be a std::map, which does needless work of sorting these parameters (which you don't need, unless I'm misunderstanding something).

redis/commands.cc
nyh: This is an interesting point!
If the write fails, I assume (I hope!) the write_strings_pairs() call will return an exceptional future, so hopefully the entire execution will return an error (which will hopefully - and this needs to be tested, at least manually - cause an error reply instead of the usual "OK". Which seems fine to me, but you're right that this isn't what the Redis documentation promises, and needs to be documented in redis.md - note that this is a bigger issue than just mset and the Redis promise isn't actually correct (see https://stackoverflow.com/questions/12911248/redis-set-command-cant-fail-but-can)

docs/redis/redis.md
nyh: Yes, we should have such comment, but it should be much bigger than "mset": You should say something like the following:

1. Many of Redis's operations (including set, mset, etc.) promise in the documentation that they cannot fail, but in Scylla such writes may fail in rare cases of errors: Although Scylla is highly-available by design, it isn't guaranteed always-available.
2. However, this isn't actually a Scylla-specific problem: Despite what the Redis documentation promises, even in the original Redis every request - including the fail-proof promise in the documentation can actually fail. So this promise is ncorrect! As pointed out, for example, in https://stackoverflow.com/questions/12911248/redis-set-command-cant-fail-but-can: Even in the regular Redis, certain problems - specifically running out of memory - can cause operations to fail, even set or mset. So Scylla's failures in this cases will not introduce a new problem for clients.

redis/commands.cc
denesb: Agree.

amoskong:
The PR is out of data, close it.

PR #5436, opened 2019-12-09
api: Add force_remove_endpoint for gossip

### LINKED ISSUE #2134: api: Add force_remove_endpoint for gossip
It is used to force remove a node from gossip membership if something
goes wrong or to accelerate the decommissioned node's gossip state
from the cluster.

Note: run the force_remove_endpoint api at the same time on all the
nodes in the cluster in order to prevent the removed nodes come back.
Becasue nodes without running the force_remove_endpoint api cmd can
gossip around the removed node information to other nodes in 2 *
ring_delay (2 * 30 seconds by default) time.

For instance, in a 3 nodes cluster, node 3 is decommissioned, to remove
node 3 from gossip membership prior the auto removal (3 days by
default), run the api cmd on both node 1 and node 2 at the same time.

$ curl -X POST --header "Accept: application/json"
"http://127.0.0.1:10000/gossiper/force_remove_endpoint/127.0.0.3"
$ curl -X POST --header "Accept: application/json"
"http://127.0.0.2:10000/gossiper/force_remove_endpoint/127.0.0.3"

Then run 'nodetool gossipinfo' on all the nodes to check the removed nodes
are not present.

asias:
It is used to force remove a node from gossip membership if something
goes wrong.

Note: run the force_remove_endpoint api at the same time on _all_ the
nodes in the cluster in order to prevent the removed nodes come back.
Becasue nodes without running the force_remove_endpoint api cmd can
gossip around the removed node information to other nodes in 2 *
ring_delay (2 * 30 seconds by default) time.

For instance, in a 3 nodes cluster, node 3 is decommissioned, to remove
node 3 from gossip membership prior the auto removal (3 days by
default), run the api cmd on both node 1 and node 2 at the same time.

$ curl -X POST --header "Accept: application/json"
"http://127.0.0.1:10000/gossiper/force_remove_endpoint/127.0.0.3"
$ curl -X POST --header "Accept: application/json"
"http://127.0.0.2:10000/gossiper/force_remove_endpoint/127.0.0.3"

Then run 'nodetool gossipinfo' on all the nodes to check the removed nodes
are not present.

Fixes #2134

### DIFF
--- a/api/api-doc/gossiper.json
+++ b/api/api-doc/gossiper.json
@@ -148,6 +148,30 @@
                ]
             }
          ]
+      },
+      {
+         "path":"/gossiper/force_remove_endpoint/{addr}",
+         "operations":[
+            {
+               "method":"POST",
+               "summary":"Force remove an endpoint from gossip",
+               "type":"void",
+               "nickname":"force_remove_endpoint",
+               "produces":[
+                  "application/json"
+               ],
+               "parameters":[
+                  {
+                     "name":"addr",
+                     "description":"The endpoint address",
+                     "required":true,
+                     "allowMultiple":false,
+                     "type":"string",
+                     "paramType":"path"
+                  }
+               ]
+            }
+         ]
       }
    ]
 }
--- a/api/gossiper.cc
+++ b/api/gossiper.cc
@@ -66,6 +66,13 @@ void set_gossiper(http_context& ctx, routes& r) {
             return make_ready_future<json::json_return_type>(json_void());
         });
     });
+
+    httpd::gossiper_json::force_remove_endpoint.set(r, [](std::unique_ptr<request> req) {
+        gms::inet_address ep(req->param["addr"]);
+        return gms::get_local_gossiper().force_remove_endpoint(ep).then([] {
+            return make_ready_future<json::json_return_type>(json_void());
+        });
+    });
 }
 
 }
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -625,6 +625,21 @@ future<> gossiper::apply_state_locally(std::map<inet_address, endpoint_state> ma
     });
 }
 
+future<> gossiper::force_remove_endpoint(inet_address endpoint) {
+    if (endpoint == get_broadcast_address()) {
+        return make_exception_future<>(std::runtime_error(format("Can not force remove node {} itself", endpoint)));
+    }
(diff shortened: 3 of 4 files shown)


asias:
This is the same patch I sent "[PATCH scylla v1 1/1] api: Add force_remove_endpoint for gossip" 3 years ago. It was rejected by @avikivity  because this api is hard to use. However, it is proven to be useful multiple times over the last few years in the field when the cluster is messed up and there is ghost node leaving in the cluster. 

I hope we can reconsider this patch.

avikivity:
There's no technical barrier, the problem is that it will always be an API only you can use. Nobody else will know if it is safe or not.

asias:
> There's no technical barrier, the problem is that it will always be an API only you can use. Nobody else will know if it is safe or not.

People in the field will know how to use it when they consult me to rescue a messed up cluster. The knowledge and experience is gained over time. So not just me know how to use.  It is almost always safe to run the api, because if the node to be remove is a valid node in the cluster, it will repopulate the gossip status into the cluster again anyway. It is much much safer than the gossip assassinate api. 

Without the patch, people will look for unsafe assassinate. This situation is much worse than you imaged situation "only I can use it and know if it is safe or not".

slivne:
@glommer can you please comment here as well

tehlers320:
A node in this state happens allot in k8s.

avikivity:
@asias I think I'm changing my mind. Can you also contribute a corresponding nodetool command and some documentation?

asias:
Changes in cab59f731344996acf60547b786aa2a51c372b8e:
- Return make_exception_future instead of throw
- Rebased to master

asias:
> @asias I think I'm changing my mind. Can you also contribute a corresponding nodetool command and some documentation?

Good.  

Since we do not want user to use it in normal cases, I think it is better to limit to HTTP API only.  Also, there is no existing nodetool cmd we can extend to host this cmd. We need to introduce a completely new cmd for that. I think it is not worth it.

asias:
@avikivity ping

asias:
@avikivity ping ^2.

slivne:
@avikivity ping AFAIK we agreed to merge this

eliransin:
I know it has been a long time since someone pinged on this. So ping @avikivity , this reviewed and comments were addressed few months ago.

PR #5448, opened 2019-12-10
scylla_io_setup: reconfig by iotune if there is no experience disk properties

### LINKED ISSUE #5438: Could not initialize seastar: YAML::InvalidNode (invalid node; first invalid key: "read_bandwidth")
*Installation details*
- ami-0d2d7cf33dd6822e8
- 666.development-0.20191207.0e319e03599

Cluster size: 1
OS (RHEL/CentOS/Ubuntu/AWS AMI): c5d.xlarge m5d.xlarge m5ad.xlarge r5d.xlarge z1d.xlarge
Instance types: c5d.xlarge m5d.xlarge m5ad.xlarge r5d.xlarge z1d.xlarge

* Description:

We want to support new instance types: c5d.xlarge m5d.xlarge m5ad.xlarge r5d.xlarge z1d.xlarge

But scylla_io_setup won't configure `read_iops/read_bandwidth/write_ipos/write_bandwidth` for those instances. Then it failed to start.

### ERROR:

```
Dec 09 04:13:22 scylla[1742]: Scylla version 666.development-0.20191207.0e319e03599 starting ...
Dec 09 04:13:22 scylla[1742]: command used: "/usr/bin/scylla --log-to-syslog 1 --log-to-stdout 0 --default-log-level info --network-stack posix --io-properties-file=/etc/scylla.d/io_properties.yaml --cpuset 0-3"
Dec 09 04:13:22 scylla[1742]: parsed command line options: [log-to-syslog: 1, log-to-stdout: 0, default-log-level: info, network-stack: posix, io-properties-file: /etc/scylla.d/io_properties.yaml, cpuset: 0-3]
Dec 09 04:13:22 scylla[1742]: Could not initialize seastar: YAML::InvalidNode (invalid node; first invalid key: "read_bandwidth")
Dec 09 04:13:22 systemd[1]: scylla-server.service: main process exited, code=exited, status=1/FAILURE
Dec 09 04:13:22 systemd[1]: Failed to start Scylla Server.
Dec 09 04:13:22 systemd[1]: Dependency failed for Scylla JMX.
Dec 09 04:13:22 systemd[1]: Job scylla-jmx.service/start failed with result 'dependency'.
Dec 09 04:13:22 systemd[1]: Unit scylla-server.service entered failed state.
Dec 09 04:13:22 systemd[1]: scylla-server.service failed.
```

```
[centos@ip-172-10-59-98 ~]$ cat /etc/scylla.d/io.conf 
SEASTAR_IO="--io-properties-file=/etc/scylla.d/io_properties.yaml"
[centos@ip-172-10-59-98 ~]$ cat /etc/scylla.d/io_properties.yaml 
disks:
- mountpoint: /var/lib/scylla
[centos@ip-172-10-59-98 ~]$
```

/CC @glommer 

Related commits:
- https://github.com/scylladb/seastar/commit/1cf9bef03e0ce71e453ec9c7197e3b92ba49b286 [io_scheduler: adjust property names]
- https://github.com/scylladb/scylla/commit/0c29289c226cf1621375842e55a94ff021f463ca [scylla_io_setup: preconfigure i3 and i2 instances with new I/O scheduler properties]

amoskong:
Currently we have some experience disk properties for some kinds of instance
types (eg: i3, i3en, i2), no disk properties will be set if there is no
experience disk properties for the instance type, then scylla-server won't
start successfully because of the following error:

Could not initialize seastar: YAML::InvalidNode (invalid node; first invalid key: "read_bandwidth")

This patch changed scylla_io_setup to run iotune to reconfig io.conf and
io_properties.yaml for the instance types, which doesn't have experience disk
properties.

Fixes #5438

/CC @glommer

### DIFF
--- a/dist/common/scripts/scylla_io_setup
+++ b/dist/common/scripts/scylla_io_setup
@@ -40,6 +40,7 @@ if __name__ == "__main__":
     if not is_developer_mode():
         if args.ami:
             idata = aws_instance()
+            reconfig_by_iotune = False
 
             disk_properties = {}
             disk_properties["mountpoint"] = datadir()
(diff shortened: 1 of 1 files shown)


syuu1228:
When we started to implement AMI, we decided not to run iotune on AMI startup, because iotune takes too long, user might confuse why scylla does not started even the instance able to login by ssh.
So we decided to skip running iotune, provide preset config for supported instance types, and let user to run iotune manually for unsupported instance types (actually it's scylla_io_setup).

Now we changed the idea to support new instance types?

amoskong:
> When we started to implement AMI, we decided not to run iotune on AMI startup, because iotune takes too long, user might confuse why scylla does not started even the instance able to login by ssh.
> So we decided to skip running iotune, provide preset config for supported instance types, and let user to run iotune manually for unsupported instance types (actually it's scylla_io_setup).

> Now we changed the idea to support new instance types?

No.

I know the benefit of avoiding to run iotune for AMI, but for supporting many new instance types (with different size), it's not a nice way to add many set of number one by one.

I didn't find a good solution, so just send this PR to make it works  (can start automatically even it's very slow), not a finally solution for supporting the new instance types.

I saw your PR https://github.com/scylladb/scylla/pull/5561
It prevent to generating invalid config, that's fine.

syuu1228:
Okay, I'm not sure this is the only way to support new instance types (c5d.xlarge m5d.xlarge m5ad.xlarge r5d.xlarge z1d.xlarge) so I need comments from other people, but as a code review I prefer not changing scylla_io_setup but change scylla_ami_setup like this:

```
    if aws.is_supported_instance_class():
        run('/opt/scylladb/scripts/scylla_io_setup --ami')
    else:
        print('some warning message')
        run('/opt/scylladb/scripts/scylla_io_setup')
```

amoskong:
> Okay, I'm not sure this is the only way to support new instance types (c5d.xlarge m5d.xlarge m5ad.xlarge r5d.xlarge z1d.xlarge) so I need comments from other people, but as a code review I prefer not changing scylla_io_setup but change scylla_ami_setup like this:
> 
> ```
>     if aws.is_supported_instance_class():

The new instance types (c5d, m5d, m5ad, r5d, z1d) had been added to support list by commit 817f34d1a94 , the code doesn't work. But the idea is right.

>         run('/opt/scylladb/scripts/scylla_io_setup --ami')
>     else:
>         print('some warning message')
>         run('/opt/scylladb/scripts/scylla_io_setup')
> ```

avikivity:
@glommer please review this

avikivity:
@glommer ping

glommer:
Sorry I had missed it.

glommer:
I just don't like this.
If the instance is unsupported, it means we don't know what kind of hardware it has. Then it is no different than any other physical hardware out there: the user has to run setup manually.

It's more than just io setup: the user has to run the full setup, including I/O setup.

I see I left comments in your original bug report: you complain that it is too tedious to add new instance types and my comment was that we don't have to categorize each instance size in the same family, just the inflection points.

@syuu1228 properly noted that another issue here - this one a real issue - is that we are generating part of the config file so the error message is not clear. We should fix that so the instance is clearly marked as unsupported

mykaul:
Closing, too old and has conflicts.

Issue #5449, opened 2019-12-10
Failed replacing node stays in gossip until full cluster shutdown


asias:
Consider following procedure

1) start n1, n2
2) n2 is down
3) start n3 to replace n2, but kill n3 in the middle of the replace
4) start n4 to replace n2

The problem is: after step 3 or step 4, n3 will stay in gossip forever until a full cluster shutdown.
Note, n3 will only stay in gossip but not in system.peers table.

For example:

```
After step 3,  n3 failed replacing n2
$ nodetool gossipinfo
/127.0.0.3
  generation:1575884618
  heartbeat:53
  SCHEMA:7f3511ce-3da1-3a8e-82b3-3ff40c46b238
  RACK:rack1
  STATUS:hibernate,true
  X3:3
  RPC_ADDRESS:127.0.0.3
  X4:0
  X1:CDC,COMPUTED_COLUMNS,CORRECT_COUNTER_ORDER,CORRECT_NON_COMPOUND_RANGE_TOMBSTONES,CORRECT_STATIC_COMPACT_IN_MC,COUNTERS,DIGEST_INSENSITIVE_TO_EXPIRY,DIGEST_MULTIPARTITION_READ,HINTED_HANDOFF_SEPARATE_CONNECTION_FEATURE,INDEXES,LARGE_PARTITIONS,LA_SSTABLE_FORMAT,MATERIALIZED_VIEWS,MC_SSTABLE_FORMAT,NONFROZEN_UDTS,RANGE_TOMBSTONES,ROLES,ROW_LEVEL_REPAIR,SCHEMA_TABLES_V3,STREAM_WITH_RPC_STREAM,TRUNCATION_TABLE,UNBOUNDED_RANGE_TOMBSTONES,VIEW_VIRTUAL_COLUMNS,WRITE_FAILURE_REPLY,XXHASH
  INTERNAL_IP:127.0.0.3
  HOST_ID:1ab96e22-0f0c-48a1-a7a2-989d5f15d985
  X5:
  X7:12
  DC:dc1
  RELEASE_VERSION:3.0.8
  NET_VERSION:0
  X2:
  X6:2
/127.0.0.2
  generation:1575884543
  heartbeat:2147483647
  SCHEMA:7f3511ce-3da1-3a8e-82b3-3ff40c46b238
  RACK:rack1
  STATUS:shutdown,true
  X3:3
  RPC_ADDRESS:127.0.0.2
  X4:0
[...]

PR #5458, opened 2019-12-11
gossiper: Remove HIBERNATE status from dead state

asias:
In scylla, the replacing node is set as HIBERNATE status. It is the only
place we use HIBERNATE status. The replacing node is supposed to be
alive and updating its heartbeat, so it is not supposed to be in dead
state.

This patch fixes the following problem in replacing.

   1) start n1, n2
   2) n2 is down
   3) start n3 to replace n2, but kill n3 in the middle of the replace
   4) start n4 to replace n2

After step 3 and step 4, the old n3 will stay in gossip forever until a
full cluster shutdown. Note n3 will only stay in gossip but in
system.peers table. User will see the annoying and infinite logs like on
all the nodes

   rpc - client $ip_of_n3:7000: fail to connect: Connection refused

Fixes: #5449
Tests: replace_address_test.py + manual test

### DIFF
--- a/gms/gossiper.hh
+++ b/gms/gossiper.hh
@@ -162,7 +162,6 @@ public:
         versioned_value::REMOVING_TOKEN,
         versioned_value::REMOVED_TOKEN,
         versioned_value::STATUS_LEFT,
-        versioned_value::HIBERNATE
     };
     const std::vector<sstring> SILENT_SHUTDOWN_STATES = {
         versioned_value::REMOVING_TOKEN,


avikivity:
HIBERNATE? What a terrible name.

avikivity:
I see Cassandra has HIBERNATE in DEAD_STATES. Does it suffer from the same problem?

avikivity:
Can you explain why being in dead state means we keep the endpoint in gossip forever, and how removing it from dead state allows removing it from gossip? I don't doubt that it's correct, I just want to understand it.

asias:
> I see Cassandra has HIBERNATE in DEAD_STATES. Does it suffer from the same problem?

Yes.

asias:
> Can you explain why being in dead state means we keep the endpoint in gossip forever, and how removing it from dead state allows removing it from gossip? I don't doubt that it's correct, I just want to understand it.

In gossip run, it checks to remove a node from gossip 1) it is gossip only member gossiper::is_gossip_only_member and 2) it is not updating bearbeat for a while.

In gossiper::is_gossip_only_member, we have 

 ```return !is_dead_state(*es) && !ss.get_token_metadata().is_member(endpoint);```

so, a node in dead state is not treated as gossip only member.

Removing the replacing node HIBERNATE status out of dead state, the failed replacing node will be removed when it is not updating heartbeat for a while.

asias:
@avikivity ping

slivne:
@avikivity ping

slivne:
closed by mistake

avikivity:
Queued as 38d4015619

PR #5481, opened 2019-12-16
dbuild: Fix podman invocation

dekimir:
The is_podman check was depending on `docker -v` printing "podman" in
the output, but that doesn't actually work, since podman prints $0.
Use `docker --help` instead, which will output "podman".

Also return podman's return status, which was previously being
dropped.

### DIFF
--- a/tools/toolchain/dbuild
+++ b/tools/toolchain/dbuild
@@ -107,7 +107,7 @@ MAVEN_LOCAL_REPO="$HOME/.m2"
 
 mkdir -p "$MAVEN_LOCAL_REPO"
 
-is_podman="$(docker --version | grep -o podman)"
+is_podman="$(docker --help | grep -o podman)"
 
 docker_common_args=()
 
@@ -152,7 +152,7 @@ if [[ -n "$interactive" || -n "$is_podman" ]]; then
     docker run --rm "${docker_common_args[@]}"
     ret=$?
     rm -rf "$tmpdir"
-    exit $?
+    exit $ret
 fi
 
 container=$(


espindola:
Dejan Mircevski <notifications@github.com> writes:

> The is_podman check was depending on `docker -v` printing &quot;podman&quot; in
> the output, but that doesn&#39;t actually work, since podman prints $0.
> Use `docker --help` instead, which will output &quot;podman&quot;.

What version do you have? On fedora 31 I get:

% docker --version
podman version 1.6.2

But I see that /usr/bin/docker is a shell script:

...
exec /usr/bin/podman "$@"

Maybe this is a fedora only thing?

Cheers,
Rafael

dekimir:
> What version do you have? On fedora 31 I get: % docker --version podman version 1.6.2 But I see that /usr/bin/docker is a shell script: ... exec /usr/bin/podman "$@" Maybe this is a fedora only thing? 

I have the same podman version, but /usr/bin/docker comes from a different package (podman-docker), which I didn't install.  Instead, I manually made /usr/bin/docker a softlink to podman (a trick commonly recommended on the interwebs).  I simply didn't know any better. :crying_cat_face: 

I don't think this change hurts those who have podman-docker rpm installed, and it helps those like me who live with, erm, alternative solutions. :wink:

espindola:
Dejan Mircevski <notifications@github.com> writes:

>> What version do you have? On fedora 31 I get: % docker --version podman version 1.6.2 But I see that /usr/bin/docker is a shell script: ... exec /usr/bin/podman "$@" Maybe this is a fedora only thing? 
>
> I have the same podman version, but /usr/bin/docker comes from a different package (podman-docker), which I didn't install.  Instead, I manually made /usr/bin/docker a softlink to podman (a trick commonly recommended on the interwebs).  I simply didn't know any better. :crying_cat_face: 
>
> I don't think this change hurts those who have podman-docker rpm installed, and it helps those like me who live with, erm, alternative solutions. :wink: 

I agree. This is a good change. I was just curious why we were getting
different results.

Thanks,
Rafael

Issue #5437, opened 2019-12-09
netstats_test: nodetool: Scylla API server HTTP GET to URL '/stream_manager/' failed: _Map_base::at


bhalevy:
Seen in [dtest-release/339/testReport/nodetool_additional_test/TestNodetool/netstats_test](https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/339/testReport/nodetool_additional_test/TestNodetool/netstats_test/)
Scylla version 7b37a700e11ae298940a2015815e51e1434ede6f
Scylla-dtest version scylladb/scylla-dtest@65a423fc6e6777bbe650ad2218af3c3eb0e8135d
```
Error Message
Nodetool command '/jenkins/workspace/scylla-master/dtest-release/scylla/.ccm/scylla-repository/7b37a700e11ae298940a2015815e51e1434ede6f/scylla-java-tools/bin/nodetool -h 127.0.90.1 -p 7190 netstats' failed; exit status: 1; stdout: Mode: NORMAL
nodetool: Scylla API server HTTP GET to URL '/stream_manager/' failed: _Map_base::at
See 'nodetool help' or 'nodetool help <command>'.
Stacktrace
Traceback (most recent call last):
  File "/usr/lib64/python2.7/unittest/case.py", line 367, in run
    testMethod()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/nodetool_additional_test.py", line 1160, in netstats_test
    stats = self.netstats(node)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/nodetool_additional_test.py", line 1071, in netstats
    out = node.nodetool('netstats', True)[0]
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/node.py", line 724, in nodetool
    raise NodetoolError(" ".join(args), exit_status, stdout, stderr)
NodetoolError: Nodetool command '/jenkins/workspace/scylla-master/dtest-release/scylla/.ccm/scylla-repository/7b37a700e11ae298940a2015815e51e1434ede6f/scylla-java-tools/bin/nodetool -h 127.0.90.1 -p 7190 netstats' failed; exit status: 1; stdout: Mode: NORMAL
nodetool: Scylla API server HTTP GET to URL '/stream_manager/' failed: _Map_base::at
See 'nodetool help' or 'nodetool help <command>'.
```

[...]

PR #5483, opened 2019-12-16
Fix stream manager get progress

asias:
streaming: Fix map access in stream_manager::get_progress
 
 When the progress is queried, e.g., query from nodetool netstats
 the progress info might not be updated yet.
 
 Fix it by checking before access the map to avoid errors like:
 
 std::out_of_range (_Map_base::at)
 
 Fixes: #5437
 Tests: nodetool_additional_test.py:TestNodetool.netstats_test

### DIFF
--- a/streaming/stream_manager.cc
+++ b/streaming/stream_manager.cc
@@ -164,13 +164,25 @@ void stream_manager::remove_progress(UUID plan_id) {
 }
 
 stream_bytes stream_manager::get_progress(UUID plan_id, gms::inet_address peer) const {
-    auto const& sbytes = _stream_bytes.at(plan_id);
-    return sbytes.at(peer);
+    auto it = _stream_bytes.find(plan_id);
+    if (it == _stream_bytes.end()) {
+        return stream_bytes();
+    }
+    auto const& sbytes = it->second;
+    auto i = sbytes.find(peer);
+    if (i == sbytes.end()) {
+        return stream_bytes();
+    }
+    return i->second;
 }
 
 stream_bytes stream_manager::get_progress(UUID plan_id) const {
+    auto it = _stream_bytes.find(plan_id);
+    if (it == _stream_bytes.end()) {
+        return stream_bytes();
+    }
     stream_bytes ret;
-    for (auto const& x : _stream_bytes.at(plan_id)) {
+    for (auto const& x : it->second) {
         ret += x.second;
     }
     return ret;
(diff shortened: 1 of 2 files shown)


streaming/stream_transfer_task.cc
```diff
@@ -176,13 +176,6 @@ future<> send_mutations(lw_shared_ptr<send_info> si) {
 }
```
avikivity: Why is this reverted? the changelog should explain.

streaming/stream_manager.cc
```diff
@@ -164,11 +164,20 @@ void stream_manager::remove_progress(UUID plan_id) {
 }
 
 stream_bytes stream_manager::get_progress(UUID plan_id, gms::inet_address peer) const {
+    if (!_stream_bytes.count(plan_id)) {
```
avikivity: A better pattern is _stream_bytes.find(), it avoids the double-check. It doesn't matter much here but it's better practice.

streaming/stream_transfer_task.cc
asias: Sorry, left over during debug. I will remove it.

streaming/stream_manager.cc
asias: Yes, I agree. I will do it in v2 anyway.

asias:
Changes in b0b5ae21696cc3be06686b6a425ad71e311fb1d9:
- Drop the revert does not belong to this series.
- Use map::find() to instead of map::count

asias:
@avikivity Ping.

slivne:
@avikivity ping

slivne:
@avikivity ping

asias:
@avikivity ping

streaming/stream_result_future.cc
```diff
@@ -131,7 +131,7 @@ void stream_result_future::maybe_complete() {
             }
             *stats = format("tx={:d} KiB, {} KiB/s, rx={:d} KiB, {} KiB/s", sbytes.bytes_sent / 1024, tx_bw, sbytes.bytes_received / 1024, rx_bw);
         }).handle_exception([plan_id] (auto ep) {
-            sslog.warn("[Stream #{}] Fail to get progess on all shards: {}", plan_id, ep);
+            sslog.warn("[Stream #{}] Fail to get progress on all shards: {}", plan_id, ep);
```
avikivity: This was already merged, 6b7344f6e56

avikivity:
Second patch queued as 145fd0313abc1fd4d7eed7d4401e2a152430bf85.

PR #5486, opened 2019-12-17
redis: create all redis keyspaces as configured

amoskong:
Note: this PR is v2 of https://github.com/scylladb/scylla/pull/5405

Currently we always try to create 16 keyspaces for Redis, the config
redis_database_count doesn't effect.

SELECT command is supported, it would switch between keyspaces. The limitation
is got from configuration. So if we config redis_database_count to be larger
than 16, SELECT allows to switch to unexisting keyspace, and it will fail in
read/write data.

Example:
| # scylla --developer-mode 1 --workdir ./ --enable-redis-protocol 1  --redis-database-count 20
| # redis-cli
| 127.0.0.1:6379> select 18
| OK
| 127.0.0.1:6379[18]> set a 1 
| (error) ERR Can't find a column family STRINGs in keyspace REDIS_18
| 127.0.0.1:6379[18]> get a
| (error) ERR Can't find a column family STRINGs in keyspace REDIS_18

v2: rename of redis_database_count

### DIFF
--- a/redis/keyspace_utils.cc
+++ b/redis/keyspace_utils.cc
@@ -179,8 +179,9 @@ future<> create_keyspace_if_not_exists_impl(db::config& config, int default_repl
         logger.info("Create keyspace: {}, table: {} for redis.", ks_name, cf_name);
         return service::get_local_migration_manager().announce_new_column_family(schema, false);
     };
-    // create 16 default database for redis.
-    return parallel_for_each(boost::irange<unsigned>(0, 16), [keyspace_gen = std::move(keyspace_gen), table_gen = std::move(table_gen)] (auto c) {
+    // Create keyspaces as configured by redis_database_count, which are used as redis databases.
+    return parallel_for_each(boost::irange<unsigned>(0, config.redis_database_count()),
+                             [keyspace_gen = std::move(keyspace_gen), table_gen = std::move(table_gen)] (auto c) {
         auto ks_name = sprint("REDIS_%d", c);
         return keyspace_gen(ks_name).then([ks_name, table_gen] {
             return when_all_succeed(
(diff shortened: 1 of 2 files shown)


nyh:
Thanks. Verified this really compiles, and committed to next.
(I assume you marked this pull-request as a "draft" by mistake?)

redis/keyspace_utils.cc
```diff
@@ -179,8 +179,9 @@ future<> create_keyspace_if_not_exists_impl(db::config& config, int default_repl
...
         return service::get_local_migration_manager().announce_new_column_family(schema, false);
     };
-    // create 16 default database for redis.
-    return parallel_for_each(boost::irange<unsigned>(0, 16), [keyspace_gen = std::move(keyspace_gen), table_gen = std::move(table_gen)] (auto c) {
+    // Create keyspaces as configured by redis_database_count, which are used as redis databases.
+    return parallel_for_each(boost::irange<unsigned>(0, config.redis_database_count()),
+                             [keyspace_gen = std::move(keyspace_gen), table_gen = std::move(table_gen)] (auto c) {
         auto ks_name = sprint("REDIS_%d", c);
         return keyspace_gen(ks_name).then([ks_name, table_gen] {
             return when_all_succeed(
```
avikivity: The seastar change (below) is broken. Never ever use `git add -a` and be careful about what you commit.

Dequeued this series.

avikivity:
@nyh I dequeued it due to unrelated change to seastar submodule.

redis/keyspace_utils.cc
amoskong: Thanks for the reminder.
This PR isn't fix something serious, I will put myself in confinement some days, then resend a correct one.

nyh:
Sorry @avikivity . I did **not** see that this happened in v2, and this time I **did** try to compile. How did you notice it? Needless to say, github PR mechanism doesn't make it very easy to notice, and when I applied it manually it also got drowned in the noise...

Issue #5490, opened 2019-12-18
repair: Do not return working_row_buf_nr in get combined row hash verb


asias:
In commit b463d7039cd20445a52191bb94770d87a025e78a (repair: Introduce
get_combined_row_hash_response), working_row_buf_nr is returned in
REPAIR_GET_COMBINED_ROW_HASH in addition to the combined hash. It is
scheduled to be part of 3.1 release. However it is not backported to 3.1
by accident.

In order to be compatible between 3.1 and 3.2 repair. We need to drop the
working_row_buf_nr in 3.2 release.

avikivity:
Already backported to 3.2, only vulnerable version.

PR #5491, opened 2019-12-18
repair: Do not return working_row_buf_nr in get combined row hash verb

asias:
In commit b463d7039cd20445a52191bb94770d87a025e78a (repair: Introduce
get_combined_row_hash_response), working_row_buf_nr is returned in
REPAIR_GET_COMBINED_ROW_HASH in addition to the combined hash. It is
scheduled to be part of 3.1 release. However it is not backported to 3.1
by accident.

In order to be compatible between 3.1 and 3.2 repair. We need to drop
the working_row_buf_nr in 3.2 release.

Fixes: #5490
Backports: 3.1 and 3.2
Tests: Run repair in a mixed 3.1 and 3.2 cluster

### DIFF
--- a/repair/repair.hh
+++ b/repair/repair.hh
@@ -321,11 +321,7 @@ struct get_sync_boundary_response {
 };
 
 // Return value of the REPAIR_GET_COMBINED_ROW_HASH RPC verb
-struct get_combined_row_hash_response {
-    repair_hash working_row_buf_combined_csum;
-    // The number of rows in the working row buf
-    uint64_t working_row_buf_nr;
-};
+using get_combined_row_hash_response = repair_hash;
 
 struct node_repair_meta_id {
     gms::inet_address ip;
--- a/idl/partition_checksum.idl.hh
+++ b/idl/partition_checksum.idl.hh
@@ -69,11 +69,6 @@ struct get_sync_boundary_response {
     uint64_t new_rows_nr;
 };
 
-struct get_combined_row_hash_response {
-    repair_hash working_row_buf_combined_csum;
-    uint64_t working_row_buf_nr;
-};
-
 enum class row_level_diff_detect_algorithm : uint8_t {
     send_full_set,
     send_full_set_rpc_stream,
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -1098,14 +1098,14 @@ private:
         _working_row_buf_combined_hash.clear();
 
         if (_row_buf.empty()) {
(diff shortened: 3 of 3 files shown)


penberg:
@avikivity @slivne Is this a release blocker for 3.2?

asias:
> @avikivity @slivne Is this a release blocker for 3.2?

If we want to allow mixed cluster repair,  we must have this in 3.2.

avikivity:
@asias it looks reasonable, I'll commit and backport. You will reintroduce it later?

How did you catch it?

avikivity:
Wait, I don't understand the "backport to 3.1" comment. If I backport it to 3.1, will a mixed 3.1.2 and 3.1.3 cluster work?

asias:
> @asias it looks reasonable, I'll commit and backport. You will reintroduce it later?

Sadly, yes. I wanted working_row_buf_nr in 3.1, so that it is a little bit future-proof.

> How did you catch it?

When I look at 3.1 and 3.2 difference handling a user issue.

asias:
> Wait, I don't understand the "backport to 3.1" comment. If I backport it to 3.1, will a mixed 3.1.2 and 3.1.3 cluster work?

I meant b463d70 (repair: Introduce get_combined_row_hash_response) is scheduled to be backported to 3.1, but we did not do that.

asias:
Now, we only need to backport "repair: Do not return working_row_buf_nr in get combined row hash verb" to 3.2. We do nothing to 3.1 release.

repair/repair.hh
```diff
@@ -323,8 +323,6 @@ struct get_sync_boundary_response {
 // Return value of the REPAIR_GET_COMBINED_ROW_HASH RPC verb
 struct get_combined_row_hash_response {
     repair_hash working_row_buf_combined_csum;
-    // The number of rows in the working row buf
-    uint64_t working_row_buf_nr;
 };
```
avikivity: I don't think this is enough. 3.1 returns a repair_hash, and that serializes differently from a struct that has one repair_hash member, because we serialize the struct size in front.

asias: Ah, then we need to drop the struct.

asias:
Update in commit: 2df78e41689eea80b88fb43663ff8b5e89e78b3d
- Drop the get_combined_row_hash_response struct

avikivity:
Queued as 7322b749e066abd0990b0ead66a0f39b47f8276a (I adjusted the commit message to reflect that it doesn't need a 3.1 backport).

Issue #5506, opened 2019-12-19
Failed to apply mutation - bad alloc and crash with coredump during bootstrap


bentsi:
*Installation details*

Scylla version (or git commit hash): `3.2.rc3-0.20191215.39b17be5620`
Cluster size: 6
OS (RHEL/CentOS/Ubuntu/AWS AMI): ami-0cab4c6eceb3249ac [eu-west-1]
Test case: [longevity-multi-keyspaces.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/branch-3.2/test-cases/longevity/longevity-multi-keyspaces.yaml)
SCT test id: 71b42f6e-23a5-4359-8ee2-7979311abc5b
DB log: [ip-10-0-233-255.gz](https://github.com/scylladb/scylla/files/3985095/ip-10-0-233-255.gz)

Nemesis: Decommission a node
Failure during happend when bootstrappaing a new node during streaming phase
Job: [ longevity-multi-keyspaces-60h-test](https://jenkins.scylladb.com/view/scylla-3.2/job/scylla-3.2/view/scylla-3.2-test/job/longevity/job/longevity-multi-keyspaces-60h/2/)
failures to write to sstable
```
2019-12-17T19:28:13+00:00  ip-10-0-233-255 !INFO    | scylla: [shard 5] compaction - Compacting [/var/lib/scylla/data/keyspace535/standard1-482e6440206f11ea8987000000000008/mc-1085-big-Data.db:level=0, /var/lib/
scylla/data/keyspace535/standard1-482e6440206f11ea8987000000000008/mc-1535-big-Data.db:level=0, /var/lib/scylla/data/keyspace535/standard1-482e6440206f11ea8987000000000008/mc-275-big-Data.db:level=0, /var/lib/sc
ylla/data/keyspace535/standard1-482e6440206f11ea8987000000000008/mc-815-big-Data.db:level=0, /var/lib/scylla/data/keyspace535/standard1-482e6440206f11ea8987000000000008/mc-1025-big-Data.db:level=0, ]
2019-12-17T19:28:13+00:00  ip-10-0-233-255 !INFO    | scylla: [shard 8] compaction - Compacted 2 sstables to [/var/lib/scylla/data/keyspace713/standard1-048f5860207f11ea848f000000000017/mc-998-big-Data.db:level=
0, ]. 6519665 bytes to 6442078 (~98% of original) in 415ms = 14.80MB/s. ~25728 total partitions merged to 25499.
[...]

PR #5509, opened 2019-12-20
sstables: file_writer: close output_stream on exception during write

bhalevy:
Otherwise we may trip the following
    assert(_closing_state == state::closed);
in ~append_challenged_posix_file_impl
when the output_stream is destructed.

Example stack trace:
    non-virtual thunk to seastar::append_challenged_posix_file_impl::~append_challenged_posix_file_impl() at /jenkins/slave/workspace/scylla-3.2/build/scylla/seastar/include/seastar/core/future.hh:944
    seastar::shared_ptr_count_for<checked_file_impl>::~shared_ptr_count_for() at crtstuff.c:?
    seastar::shared_ptr<seastar::file_impl>::~shared_ptr() at /jenkins/slave/workspace/scylla-3.2/build/scylla/seastar/include/seastar/core/future.hh:944
     (inlined by) seastar::file::~file() at /jenkins/slave/workspace/scylla-3.2/build/scylla/seastar/include/seastar/core/file.hh:155
     (inlined by) seastar::file_data_sink_impl::~file_data_sink_impl() at /jenkins/slave/workspace/scylla-3.2/build/scylla/seastar/src/core/fstream.cc:312
     (inlined by) seastar::file_data_sink_impl::~file_data_sink_impl() at /jenkins/slave/workspace/scylla-3.2/build/scylla/seastar/src/core/fstream.cc:312
    seastar::output_stream<char>::~output_stream() at crtstuff.c:?
    sstables::sstable::write_crc(sstables::checksum const&) [clone .cold] at sstables.cc:?
    sstables::mc::writer::close_data_writer() at crtstuff.c:?
    sstables::mc::writer::consume_end_of_stream() at crtstuff.c:?
    sstables::sstable::write_components(flat_mutation_reader, unsigned long, seastar::lw_shared_ptr<schema const>, sstables::sstable_writer_config const&, encoding_stats, seastar::io_priority_class const&)::{lambda()#1}::operator()() at sstables.cc:?

Refs #5506

Test: unit(dev)

### DIFF
--- a/sstables/sstables.cc
+++ b/sstables/sstables.cc
@@ -928,6 +928,13 @@ void sstable::generate_toc(compressor_ptr c, double filter_fp_chance) {
     _recognized_components.insert(component_type::Scylla);
 }
 
+// Must be called in a seastar thread
+file_writer sstable::make_component_file_writer(component_type c, file_output_stream_options options, open_flags oflags)
+{
+    auto f = new_sstable_component_file(_write_error_handler, c, oflags).get0();
+    return file_writer(std::move(f), std::move(options));
+}
+
 void sstable::write_toc(const io_priority_class& pc) {
     touch_temp_dir().get0();
     auto file_path = filename(component_type::TemporaryTOC);
@@ -938,21 +945,19 @@ void sstable::write_toc(const io_priority_class& pc) {
     // If creation of temporary TOC failed, it implies that that boot failed to
     // delete a sstable with temporary for this column family, or there is a
     // sstable being created in parallel with the same generation.
-    file f = new_sstable_component_file(_write_error_handler, component_type::TemporaryTOC, open_flags::wo | open_flags::create | open_flags::exclusive).get0();
+    file_output_stream_options options;
+    options.buffer_size = 4096;
+    options.io_priority_class = pc;
+    auto w = make_component_file_writer(component_type::TemporaryTOC, std::move(options));
 
     bool toc_exists = file_exists(filename(component_type::TOC)).get0();
     if (toc_exists) {
         // TOC will exist at this point if write_components() was called with
         // the generation of a sstable that exists.
-        f.close().get();
+        w.close();
         remove_file(file_path).get();
         throw std::runtime_error(format("SSTable write failed due to existence of TOC file for generation {:d} of {}.{}", _generation, _schema->ks_name(), _schema->cf_name()));
     }
-
-    file_output_stream_options options;
-    options.buffer_size = 4096;
-    options.io_priority_class = pc;
-    auto w = file_writer(std::move(f), std::move(options));
 
     for (auto&& key : _recognized_components) {
             // new line character is appended to the end of each component name.
@@ -1003,12 +1008,9 @@ void sstable::write_crc(const checksum& c) {
     auto file_path = filename(component_type::CRC);
     sstlog.debug("Writing CRC file {} ", file_path);
 
-    auto oflags = open_flags::wo | open_flags::create | open_flags::exclusive;
-    file f = new_sstable_component_file(_write_error_handler, component_type::CRC, oflags).get0();
-
     file_output_stream_options options;
     options.buffer_size = 4096;
-    auto w = file_writer(std::move(f), std::move(options));
+    auto w = make_component_file_writer(component_type::CRC, std::move(options));
     write(get_version(), w, c);
     w.close();
 }
@@ -1018,12 +1020,9 @@ void sstable::write_digest(uint32_t full_checksum) {
     auto file_path = filename(component_type::Digest);
     sstlog.debug("Writing Digest file {} ", file_path);
 
-    auto oflags = open_flags::wo | open_flags::create | open_flags::exclusive;
-    auto f = new_sstable_component_file(_write_error_handler, component_type::Digest, oflags).get0();
-
     file_output_stream_options options;
     options.buffer_size = 4096;
-    auto w = file_writer(std::move(f), std::move(options));
+    auto w = make_component_file_writer(component_type::Digest, std::move(options));
 
     auto digest = to_sstring<bytes>(full_checksum);
     write(get_version(), w, digest);
@@ -1066,12 +1065,11 @@ void sstable::do_write_simple(component_type type, const io_priority_class& pc,
         noncopyable_function<void (version_types version, file_writer& writer)> write_component) {
     auto file_path = filename(type);
     sstlog.debug(("Writing " + sstable_version_constants::get_component_map(_version).at(type) + " file {} ").c_str(), file_path);
-    file f = new_sstable_component_file(_write_error_handler, type, open_flags::wo | open_flags::create | open_flags::exclusive).get0();
 
     file_output_stream_options options;
     options.buffer_size = sstable_buffer_size;
     options.io_priority_class = pc;
-    auto w = file_writer(std::move(f), std::move(options));
+    auto w = make_component_file_writer(type, std::move(options));
     std::exception_ptr eptr;
     try {
         write_component(_version, w);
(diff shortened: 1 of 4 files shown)


sstables/writer.hh
```diff
@@ -52,8 +52,20 @@ public:
...
+            try {
+                // close the output_stream so it won't
+                // be destructed while open.
+                _out.close().get();
+            } catch (...) {
+                // ignore nested exception
+            }
+            std::rethrow_exception(ep);
+        }
     }
```
avikivity: I don't think this is the right place to close (after every write). We should prepare a closer after the open (with defer()), since we write in a thread we can use RAII.

bhalevy: make sense.  done in v2.

bhalevy:
In v2:
- a couple of infrastructure patches:
  - sstable: add make_component_file_writer
  - sstable: file_writer: add filename member
- sstable: file_writer: auto-close in destructor
  - RAII'ify file_writer and auto-close in destructor
    - log error on close
    - ignore close error in destructor.

sstables/sstables.cc
```diff
@@ -927,6 +927,12 @@ void sstable::generate_toc(compressor_ptr c, double filter_fp_chance) {
     _recognized_components.insert(component_type::Scylla);
 }
 
+file_writer sstable::make_component_file_writer(component_type c, file_output_stream_options options, open_flags oflags)
+{
+    auto f = new_sstable_component_file(_write_error_handler, c, oflags).get0();
+    return file_writer(std::move(f), std::move(options));
```
avikivity: What happens if file_writer's constructor throws? We don't close f.

sstables/writer.hh
```diff
@@ -118,7 +123,7 @@ template <typename T>
 uint64_t
 serialized_size(sstable_version_types v, const T& object) {
     uint64_t size = 0;
-    auto writer = file_writer(make_sizing_output_stream(size));
+    auto writer = file_writer(make_sizing_output_stream(size), "<sizing_output_stream>");
```
avikivity: This is an indication that the file_writer isn't the correct place for the name, if you have to make up a fake name.

bhalevy: The thing is that it's misnamed as it works on output streams that are not files.

sstables/sstables.cc
bhalevy: good catch.  I'll add handling for that.

bhalevy:
In v3:
- sstable: file_writer: add optional filename member
  - engaged only for files
- sstable: handle exception in file_writer constructor
  - close file if file_writer constructor throws

bhalevy:
@avikivity ping

sstables/writer.hh
```diff
@@ -65,16 +78,19 @@ public:
...
     uint64_t offset() const {
         return _offset.offset;
     }
 
     const writer_offset_tracker& offset_tracker() const {
         return _offset;
     }
+
+    sstring get_filename() const {
+        return _filename.value_or("output_stream");
```
espindola: I would change _filename to be a plain std::string and in the constructor that doesn't take a filename initialize _filename to "output_stream".

sstables/writer.hh
```diff
@@ -41,14 +41,27 @@ class metadata_collector;
 class file_writer {
     output_stream<char> _out;
     writer_offset_tracker _offset;
+    std::optional<sstring> _filename;
+    bool _closed = false;
 public:
-    file_writer(file f, file_output_stream_options options)
-        : _out(make_file_output_stream(std::move(f), std::move(options))) {}
+    // It is the responsiblity of the caller to close file f
+    // if the constructor throws
```
espindola: This suggests that the constructor is doing too much. Maybe the constructor should be private and there should be a public static make() function that can handle exceptions.

avikivity: @espindola why? the caller created f, so the caller is responsible for closing f, no?

avikivity:
Looks good to me, but need to clear an issue with @espindola .

sstables/writer.hh
espindola: My preference is for constructors that are super simple (noexcept in particular). That way ownership transfer is simple, once the constructor is called the destructor is responsible for freeing resources.

Any convenience that is lost by the simple constructor (like being able to call non noexcept functinos like  make_file_output_stream) is regained by moving that to a static make function.

That is just a preference and should not block merging this change.

slivne:
@bhalevy ping

sstables/writer.hh
bhalevy: sounds good

sstables/writer.hh
bhalevy: @espindola so let's defer that and do it as a follow up as making the file output stream in the constructor is an existing problem.

I opened https://github.com/scylladb/scylla/issues/5901 for doing that.

bhalevy:
In v4:
- rebased
- changed filename type to plain std::string
  - explicitly initialized to "output_stream" in the respective constructor.

bhalevy:
In v4:
- rebased
- changed filename type to plain std::string
  - explicitly initialized to "output_stream" in the respective constructor.

sstables/writer.hh
```diff
@@ -42,12 +42,22 @@ class metadata_collector;
...
+
+    file_writer(output_stream<char>&& out, std::string filename)
+        : _out(std::move(out))
+        , _filename(std::move(filename))
+    {}
 
     file_writer(output_stream<char>&& out)
-        : _out(std::move(out)) {}
+        : _out(std::move(out))
+        , _filename("output_stream")
```
tgrabiec: IMO passing "output_stream" as the file name is confusing, since it's not the actual file name which is used here. The file name is not known, so _filename should be a disengaged std::optional<> in this case.

sstables/sstables.cc
```diff
@@ -3311,18 +3319,26 @@ delete_atomically(std::vector<shared_sstable> ssts) {
...
+                    w.write(toc.c_str(), toc.size());
+                    w.write("\n", 1);
+                }
 
-            w.flush();
-            w.close();
+                w.flush();
+                w.close();
+            } catch (...) {
+                f.close().handle_exception([tmp_pending_delete_log] (auto ep) {
```
tgrabiec: After you construct the file_writer, it is the owner of the file, so it is responsible for closing it. We should not close "f" under its fit, that may result in undefined behavior (like writing to the wrong file).

sstables/sstables.cc
```diff
@@ -928,6 +928,31 @@ void sstable::generate_toc(compressor_ptr c, double filter_fp_chance) {
...
+}
+
+void file_writer::close() {
+    try {
+        _out.close().get();
+        _closed = true;
+    } catch (...) {
+        auto e = std::current_exception();
+        sstlog.error("Error while closing {}: {}", get_filename(), e);
+        std::rethrow_exception(e);
```
tgrabiec: We should either throw or log, not both. If we want to attach more context (file name), we should nest the exception.

bhalevy: that's doable

sstables/sstables.cc
bhalevy: You're right. We need to do that only if constructing the file_writer failed.

sstables/writer.hh
bhalevy: Using std::optional is similar to a previous implementation that was changed to this one based on a review comment...
When disengaged, what do you think should the printout say?
How about `<output_stream>` to make it more obvious it's not an actual file name?

slivne:
@bhalevy is this still relevant - if so lets send a new version, if not please close

bhalevy:
> @bhalevy is this still relevant - if so lets send a new version, if not please close

@slivne yes, this is still relevant, but I'd like to do the new version based on https://github.com/scylladb/scylla/issues/5901 as it will be much cleaner and simpler this way.

bhalevy:
Need to redo based on 0c34e114e29af59c40470c4d6e00f3cf36e6fe26

bhalevy:
55cf219 closes this PR

PR #5512, opened 2019-12-22
Fix case sensitive variable name

hagitsegev:
Fixed a case sensitive variable.
Adding a log print, to enable understanding what is going on (hard to debug like this).
I need this in order to build debug RPM's https://github.com/scylladb/scylla-pkg/issues/698

### DIFF
--- a/reloc/build_rpm.sh
+++ b/reloc/build_rpm.sh
@@ -34,7 +34,7 @@ while [ $# -gt 0 ]; do
             shift 2
             ;;
         "--builddir")
-            builddir="$2"
+            BUILDDIR="$2"
             shift 2
             ;;
         *)
--- a/reloc/build_rpm.sh
+++ b/reloc/build_rpm.sh
@@ -55,4 +55,5 @@ fi
 mkdir -p $BUILDDIR/scylla-package
 tar -C $BUILDDIR/scylla-package -xpf $RELOC_PKG SCYLLA-RELOCATABLE-FILE SCYLLA-RELEASE-FILE SCYLLA-VERSION-FILE SCYLLA-PRODUCT-FILE dist/redhat
 cd $BUILDDIR/scylla-package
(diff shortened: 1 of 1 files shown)


hagitsegev:
Checked BYO: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/byo_build_tests_dtest/807/

reloc/build_rpm.sh
```diff
@@ -55,4 +55,5 @@ fi
 mkdir -p $BUILDDIR/scylla-package
 tar -C $BUILDDIR/scylla-package -xpf $RELOC_PKG SCYLLA-RELOCATABLE-FILE SCYLLA-RELEASE-FILE SCYLLA-VERSION-FILE SCYLLA-PRODUCT-FILE dist/redhat
 cd $BUILDDIR/scylla-package
+echo "Running ./dist/redhat/build_rpm.sh on |$BUILDDIR/scylla-package| OPTS: |$OPTS|"
```
bhalevy: I don't about this message.  It's not strictly required for this fix, or at all, and I don't like the `|` delimiter convention that you imported from scylla-pkg.

hagitsegev: I can remove the || if you don't like it. I use it, so in case I have a variable that has no value, it is clear that I have || null.
I can also put the echo in a separated commit, if this makes more sense.
Without the echo, it is very hard to debug.
I ran the script on the centos-rpm job in order to create debug RPM. Everything went OK, no error messages, but nothing was created. Now you need to understand. 
The way to do it is long. You can edit this script, create a new reloc (using byo), and run the target job again, or disable some parts of the pipeline, and edit the script on the target machine.
Well, I know some of you don't like echo's. I know that the industry standards are to write "clear" code... but for me... it is hard to debug like this. And debug is not only when you write the script, It can happen a while after that, as happened to me now.

So - tell me if you want me to remove the echo, or to put it in a separated commit (with or without the ||).

syuu1228:
@hagitsegev @bhalevy Fine for me.

PR #5519, opened 2019-12-25
Fix case sensitive variable name

hagitsegev:
Fixed a case sensitive variable.
Adding a log print, to enable understanding what is going to run (hard to debug like this).
I need this in order to build debug RPM's scylladb/scylla-pkg#698

Originally done on #5512 
See review comments by @bhalevy and @syuu1228 there.

### DIFF
--- a/reloc/build_rpm.sh
+++ b/reloc/build_rpm.sh
@@ -34,7 +34,7 @@ while [ $# -gt 0 ]; do
             shift 2
             ;;
         "--builddir")
-            builddir="$2"
+            BUILDDIR="$2"
             shift 2
             ;;
         *)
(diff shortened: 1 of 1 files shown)


reloc/build_rpm.sh
```diff
@@ -55,4 +55,5 @@ fi
 mkdir -p $BUILDDIR/scylla-package
 tar -C $BUILDDIR/scylla-package -xpf $RELOC_PKG SCYLLA-RELOCATABLE-FILE SCYLLA-RELEASE-FILE SCYLLA-VERSION-FILE SCYLLA-PRODUCT-FILE dist/redhat
 cd $BUILDDIR/scylla-package
+echo "Running ./dist/redhat/build_rpm.sh on $BUILDDIR/scylla-package OPTS: $OPTS"
```
amoskong: How about output the raw shell cmdline like this?

echo "Running './dist/redhat/build_rpm.sh $OPTS' under $BUILDDIR/scylla-package directory"

Issue #5525, opened 2019-12-26
scylla crashes with a very cryptic error: scylla: build_id.cc:44: const Elf64_Nhdr* get_nt_build_id(dl_phdr_info*): Assertion `0 && "no NT_GNU_BUILD_ID note"' failed. 


vladzcloudius:
*Installation details*
HEAD: 60a851d3a55f973e65ee71af61516f2750fe4aeb
OS: Ubuntu 19.10
Compiler: g++ (Ubuntu 9.2.1-9ubuntu2) 9.2.1 20191008

*Description*
Scylla build from a local repository crashes right on the start with the following symptom:
```
 scylla: build_id.cc:44: const Elf64_Nhdr* get_nt_build_id(dl_phdr_info*): Assertion `0 && "no NT_GNU_BUILD_ID note"' failed. 
```

This is a regression introduced by the following patch:
```
commit 91c7f5bf4470783dd03d1c3c2c4c972302251e08
Author: Rafael Ávila de Espíndola <espindola@scylladb.com>
Date:   Tue Dec 17 19:15:56 2019 -0800
    Print build-id on startup
    Fixes #5426
    Signed-off-by: Rafael Ávila de Espíndola <espindola@scylladb.com>
    Message-Id: <20191218031556.120089-1-espindola@scylladb.com>
```

*Steps to reproduce*
1) Build scylla:
   * ./configure.py --compiler=g++-9 '--cflags=-Wno-attributes -Wno-unused-result -Wno-missing-attributes -Wno-array-bounds -I/usr/include/lua5.3' --c-compiler=gcc-9 '--ldflags=-fuse-ld=lld -L/usr/local/lib -lfmt'
   * ninja ./build/debug/scylla (the same reproduced with `dev` build as well).
2) Run `scylla --help`

bhalevy:
How did you hit it?
Is it reproducible?

vladzcloudius:
> How did you hit it?
> Is it reproducible?

Yes, it's very easily reproducible.
I updated the description.

espindola:
> To add to the insult when I revert the patch in question scylla is able to print build ID just fine:
> 
> ```
> $ ./build/debug/scylla --version
> 666.development-0.20191226.60a851d3a
> ```

That is not the build-id.

> _Steps to reproduce_
> 
>     1. Build scylla.

How? Please provide the exact commands and environment you used. In particular, are you using dbuild?

vladzcloudius:
[...]

PR #5526, opened 2019-12-26
Build id cleanups

bhalevy:
Refs #5525 

- main: add --build-id option
- build_id: mv sources to utils/
- build_id: throw on errors rather than assert
- build_id: simplify callback pointer type casting


main.cc
```diff
@@ -356,9 +356,10 @@ std::string format_parsed_options(const std::vector<bpo::option>& opts) {
 }
 
 static constexpr char startup_msg[] = "Scylla version {} with build-id {} starting ...\n";
+static std::string build_id;
```
espindola: There is now no guarantee that this is initialized before use. If calling get_build_id is too expensive (is it?) make this a static local variable in get_build_id:
```
string get_build_id() {
   static string ret = compute_build_id();
   return ret;
}

File: main.cc [hunk 2]
```diff
@@ -447,6 +449,18 @@ int main(int ac, char** av) {
         return 0;
     }
 
+    try {
+        build_id = get_build_id();
+    } catch (...) {
+        fmt::print("Could not get build_id: {}\n", std::current_exception());
+        build_id = "<unknown>";
```
espindola: I don't think we should run if we failed to print a build_id. If this fails it means that there is something horribly broken with out build (which there is).

It is obviously OK to print a better error message and exit() instead of failing an assert.

espindola:
vladzcloudius <notifications@github.com> writes:

> vladzcloudius commented on this pull request.
>
> I'm not sure I understand what is the use case this new feature is going to cover.
>
> From what I understood from Benny this is meant to allow connection between a given core file and the corresponding official/QA build. 
>
> However, I don't understand under which circumstances we won't know which scylla version generated it and if do know it then why would we need a build ID in addition to that.
>
> I suspect I'm missing some context here.

With the fix to build_reloc core dumps and binaries become easy to
match. With scylla printing its own build-id it is also easy to match
the log output.

Cheers,
Rafael

vladzcloudius:
> vladzcloudius <notifications@github.com> writes:
> vladzcloudius commented on this pull request. I'm not sure I understand what is the use case this new feature is going to cover. From what I understood from Benny this is meant to allow connection between a given core file and the corresponding official/QA build. However, I don't understand under which circumstances we won't know which scylla version generated it and if do know it then why would we need a build ID in addition to that. I suspect I'm missing some context here.
> With the fix to build_reloc core dumps and binaries become easy to match. With scylla printing its own build-id it is also easy to match the log output. Cheers, Rafael

@espindola 
I don't understand under which circumstances printing scylla version only is not going to be sufficient.
The same goes to core <-> binary match. Which use case would require this extra info?

espindola:
vladzcloudius <notifications@github.com> writes:

>> vladzcloudius <notifications@github.com> writes:
>> vladzcloudius commented on this pull request. I'm not sure I understand what is the use case this new feature is going to cover. From what I understood from Benny this is meant to allow connection between a given core file and the corresponding official/QA build. However, I don't understand under which circumstances we won't know which scylla version generated it and if do know it then why would we need a build ID in addition to that. I suspect I'm missing some context here.
>> With the fix to build_reloc core dumps and binaries become easy to match. With scylla printing its own build-id it is also easy to match the log output. Cheers, Rafael
>
> @espindola 
> I don't understand under which circumstances printing scylla version only is not going to be sufficient.
> The same goes to core <-> binary match. Which use case would require this extra info?

Decoding backtraces from the "reactor stuck" messages for instance. Our
build is not reproducible, so we need to know the exact binary.

Cheers,
Rafael

vladzcloudius:
> vladzcloudius <notifications@github.com> writes:
> > vladzcloudius ***@***.***> writes: > vladzcloudius commented on this pull request. I'm not sure I understand what is the use case this new feature is going to cover. From what I understood from Benny this is meant to allow connection between a given core file and the corresponding official/QA build. However, I don't understand under which circumstances we won't know which scylla version generated it and if do know it then why would we need a build ID in addition to that. I suspect I'm missing some context here. > With the fix to build_reloc core dumps and binaries become easy to match. With scylla printing its own build-id it is also easy to match the log output. Cheers, Rafael @espindola I don't understand under which circumstances printing scylla version only is not going to be sufficient. The same goes to core <-> binary match. Which use case would require this extra info?
> Decoding backtraces from the "reactor stuck" messages for instance. Our build is not reproducible, so we need to know the exact binary. Cheers, Rafael

@espindola 
When you have a reactor stall backtrace you usually know where it came from, namely you know both the exact scylla version (it's printed in the log already) and you know which box it came from (OS flavor). 
This should allow you to get to the exact package that generated it, and we do that on a regular basis.

I don't understand why you need this extra ID for.

espindola:
vladzcloudius <notifications@github.com> writes:

>> vladzcloudius <notifications@github.com> writes:
>> > vladzcloudius ***@***.***> writes: > vladzcloudius commented on this pull request. I'm not sure I understand what is the use case this new feature is going to cover. From what I understood from Benny this is meant to allow connection between a given core file and the corresponding official/QA build. However, I don't understand under which circumstances we won't know which scylla version generated it and if do know it then why would we need a build ID in addition to that. I suspect I'm missing some context here. > With the fix to build_reloc core dumps and binaries become easy to match. With scylla printing its own build-id it is also easy to match the log output. Cheers, Rafael @espindola I don't understand under which circumstances printing scylla version only is not going to be sufficient. The same goes to core <-> binary match. Which use case would require this extra info?
>> Decoding backtraces from the "reactor stuck" messages for instance. Our build is not reproducible, so we need to know the exact binary. Cheers, Rafael
>
> @espindola 
> When you have a reactor stall backtrace you usually know where it came from, namely you know both the exact scylla version (it's printed in the log already) and you know which box it came from (OS flavor). 

That is not sufficient. Two builds of scylla at the same revision can
have different addresses.

Cheers,
Rafael

vladzcloudius:
On Sat, Dec 28, 2019, 1:21 PM Rafael Ávila de Espíndola <
notifications@github.com> wrote:

> vladzcloudius <notifications@github.com> writes:
>
> >> vladzcloudius <notifications@github.com> writes:
> >> > vladzcloudius ***@***.***> writes: > vladzcloudius commented on this
> pull request. I'm not sure I understand what is the use case this new
> feature is going to cover. From what I understood from Benny this is meant
> to allow connection between a given core file and the corresponding
> official/QA build. However, I don't understand under which circumstances we
> won't know which scylla version generated it and if do know it then why
> would we need a build ID in addition to that. I suspect I'm missing some
> context here. > With the fix to build_reloc core dumps and binaries become
> easy to match. With scylla printing its own build-id it is also easy to
> match the log output. Cheers, Rafael @espindola I don't understand under
> which circumstances printing scylla version only is not going to be
> sufficient. The same goes to core <-> binary match. Which use case would
> require this extra info?
> >> Decoding backtraces from the "reactor stuck" messages for instance. Our
> build is not reproducible, so we need to know the exact binary. Cheers,
> Rafael
> >
> > @espindola
> > When you have a reactor stall backtrace you usually know where it came
> from, namely you know both the exact scylla version (it's printed in the
> log already) and you know which box it came from (OS flavor).
>
> That is not sufficient. Two builds of scylla at the same revision can
> have different addresses.
>

I'm not sure we are on the same page here.

Could you, please, give a specific real life example of the above situation
(or any other situation) when you wouldn't know what to do without this new
ID?

Builds do not exist in a vacuum, so do core files. We always know where
they came from and what binary has generated them and under which
circumstances.

At least this is what I knew so far.

For instance, two builds of the same version could have different symbols
but this may only happen in either dev environment or in QA and in both
cases you would know which one you used and hence the problem becomes a
non-issue.

So, again, could you, please, give me an example of the situation when you
would HAVE TO have this new ID because, really, I can't see one by myself.

Thanks,
Vlad

espindola:
vladzcloudius <notifications@github.com> writes:

> On Sat, Dec 28, 2019, 1:21 PM Rafael Ávila de Espíndola <
> notifications@github.com> wrote:
>
>> vladzcloudius <notifications@github.com> writes:
>>
>> >> vladzcloudius <notifications@github.com> writes:
>> >> > vladzcloudius ***@***.***> writes: > vladzcloudius commented on this
>> pull request. I'm not sure I understand what is the use case this new
>> feature is going to cover. From what I understood from Benny this is meant
>> to allow connection between a given core file and the corresponding
>> official/QA build. However, I don't understand under which circumstances we
>> won't know which scylla version generated it and if do know it then why
>> would we need a build ID in addition to that. I suspect I'm missing some
>> context here. > With the fix to build_reloc core dumps and binaries become
>> easy to match. With scylla printing its own build-id it is also easy to
>> match the log output. Cheers, Rafael @espindola I don't understand under
>> which circumstances printing scylla version only is not going to be
>> sufficient. The same goes to core <-> binary match. Which use case would
>> require this extra info?
>> >> Decoding backtraces from the "reactor stuck" messages for instance. Our
>> build is not reproducible, so we need to know the exact binary. Cheers,
>> Rafael
>> >
>> > @espindola
>> > When you have a reactor stall backtrace you usually know where it came
>> from, namely you know both the exact scylla version (it's printed in the
>> log already) and you know which box it came from (OS flavor).
>>
>> That is not sufficient. Two builds of scylla at the same revision can
>> have different addresses.
>>
>
> I'm not sure we are on the same page here.
>
> Could you, please, give a specific real life example of the above situation
> (or any other situation) when you wouldn't know what to do without this new
> ID?

Yes, I have been pointed to a bug with a backtrace and the procedure for
trying to find the corresponding binary involved trying to figure out
which bot run created it.

With the logs including a build-id and we storing build artifact by
build-id all the guesswork is avoided.

Cheers,
Rafael

vladzcloudius:
On Sat, Dec 28, 2019, 8:34 PM Rafael Ávila de Espíndola <
notifications@github.com> wrote:

> vladzcloudius <notifications@github.com> writes:
>
> > On Sat, Dec 28, 2019, 1:21 PM Rafael Ávila de Espíndola <
> > notifications@github.com> wrote:
> >
> >> vladzcloudius <notifications@github.com> writes:
> >>
> >> >> vladzcloudius <notifications@github.com> writes:
> >> >> > vladzcloudius ***@***.***> writes: > vladzcloudius commented on
> this
> >> pull request. I'm not sure I understand what is the use case this new
> >> feature is going to cover. From what I understood from Benny this is
> meant
> >> to allow connection between a given core file and the corresponding
> >> official/QA build. However, I don't understand under which
> circumstances we
> >> won't know which scylla version generated it and if do know it then why
> >> would we need a build ID in addition to that. I suspect I'm missing some
> >> context here. > With the fix to build_reloc core dumps and binaries
> become
> >> easy to match. With scylla printing its own build-id it is also easy to
> >> match the log output. Cheers, Rafael @espindola I don't understand under
> >> which circumstances printing scylla version only is not going to be
> >> sufficient. The same goes to core <-> binary match. Which use case would
> >> require this extra info?
> >> >> Decoding backtraces from the "reactor stuck" messages for instance.
> Our
> >> build is not reproducible, so we need to know the exact binary. Cheers,
> >> Rafael
> >> >
> >> > @espindola
> >> > When you have a reactor stall backtrace you usually know where it came
> >> from, namely you know both the exact scylla version (it's printed in the
> >> log already) and you know which box it came from (OS flavor).
> >>
> >> That is not sufficient. Two builds of scylla at the same revision can
> >> have different addresses.
> >>
> >
> > I'm not sure we are on the same page here.
> >
> > Could you, please, give a specific real life example of the above
> situation
> > (or any other situation) when you wouldn't know what to do without this
> new
> > ID?
>
> Yes, I have been pointed to a bug with a backtrace and the procedure for
> trying to find the corresponding binary involved trying to figure out
> which bot run created it.
>
> With the logs including a build-id and we storing build artifact by
> build-id all the guesswork is avoided.
>

I see.
Please, don't get me wrong, I don't have anything against yet another thing
that is ought to make our life easier.

My only concern here is that we don't mask a real issue which was the fact
that you couldn't recover the configuration of a run from the bug with a
workaround which this new ID printing is.

Adding a build ID printing is definitely a welcome thing however what
bothers me is that it may be seen as a fix for the issue you have described
while we should be improving the automation that generated that bug instead
- in a way that would not require you adding build ID printing at the first
place.

Best regards,
Vlad

main.cc
bhalevy: ok

main.cc
bhalevy: I beg to differ.  Not having the build_id is not a critical error we should crash on.  As seen it could be just a bug on our side, or something that changed in toolchain.

vladzcloudius: Please, use a `startlog` logger instead of printing to `stdout`

utils/build_id.cc
```diff
@@ -41,15 +41,19 @@ static const Elf64_Nhdr* get_nt_build_id(dl_phdr_info* info) {
...
 
 static int callback(dl_phdr_info* info, size_t size, void* data) {
     std::string& ret = *(std::string*)data;
     std::ostringstream os;
 
     // The first DSO is always the main program, which has an empty name.
-    assert(strlen(info->dlpi_name) == 0);
+    if (strlen(info->dlpi_name) != 0) {
+        std::string err = "Expected empty DSO name, but got: ";
+        err += info->dlpi_name;
```
vladzcloudius: This looks rather dangerous: you should limit the maximum amount of characters that may be added to the string as a result of this concatenation.

espindola:
Benny Halevy <notifications@github.com> writes:

> bhalevy commented on this pull request.
>
>
>
>> @@ -447,6 +449,18 @@ int main(int ac, char** av) {
>          return 0;
>      }
>  
> +    try {
> +        build_id = get_build_id();
> +    } catch (...) {
> +        fmt::print("Could not get build_id: {}\n", std::current_exception());
> +        build_id = "<unknown>";
>
> I beg to differ.  Not having the build_id is not a critical error we should crash on.  As seen it could be just a bug on our side, or something that changed in toolchain.

But those are critical things that must be fixes ASAP. To use your
example, if a toolchain change breaks build-id we want it to fail next
promotion.

If we don't we risk ending on the same situation we are now where we
find it hard to pair logs, cores and binaries.

Cheers,
Rafael

vladzcloudius:
> Benny Halevy <notifications@github.com> writes:
> bhalevy commented on this pull request. > @@ -447,6 +449,18 @@ int main(int ac, char** av) { return 0; } + try { + build_id = get_build_id(); + } catch (...) { + fmt::print("Could not get build_id: {}\n", std::current_exception()); + build_id = "<unknown>"; I beg to differ. Not having the build_id is not a critical error we should crash on. As seen it could be just a bug on our side, or something that changed in toolchain.
> But those are critical things that must be fixes ASAP. To use your example, if a toolchain change breaks build-id we want it to fail next promotion. If we don't we risk ending on the same situation we are now where we find it hard to pair logs, cores and binaries. Cheers, Rafael

@espindola Logs, cores and binaries should be possible to pair by other means.
Since it's obvious to me that the use case you have in mind is QA then it should be done be means of proper logging done by QA tools. 

AFAIU from @bhalevy this may even be the case today.

It's not obvious to me why this printing even needed.
Does it have to do with relocatable packaging that breaks `eu-readelf` and `eu-unstrip` ? If yes, can it be fixed?

slivne:
@bhalevy was this merged / abandoned ?

bhalevy:
> @bhalevy was this merged / abandoned ?

@slivne not abandoned but swapped out due to higher priorities.

avikivity:
I see there is lots of commentary, was everything resolved?

bhalevy:
> I see there is lots of commentary, was everything resolved?

I force pushed a simplified version
addressing review comments and rebased.

In v2:
- no `static sstring build_id;` in main.cc
- dequeued `build_id: throw on errors rather than assert`
  - not needed anymore

Tested manually: using
```
ninja build/dev/scylla && ./build/dev/scylla --build-id
```

PR #5533, opened 2019-12-30
Pretty print canonical_mutation objects

avikivity:
canonical_mutation objects are used for schema reconciliation, which is a fragile area and thus deserves some debugging help.

This series makes canonical_mutation objects printable. The end result looks something like

```
INFO  2019-12-30 17:00:40,935 [shard 1] migration_manager - received definitions update from 127.0.0.1:62937: {{{canonical_mutation: table_id afddfb9d-bc1e-3068-8056-eed6c302ba09 schema_version c179c1d7-9503-3f66-a5b3-70e72af3392a partition_key pk{00026b73} partition_tombstone {tombstone: none}{row {position: clustered,ckp{0003746162},0} tombstone {row_tombstone: none} marker {row_marker: }, column bloom_filter_fp_chance atomic_cell{0.01;ts=1577718040933619;expiry=-1,ttl=0}, column caching atomic_cell{({keys : ALL}, {rows_per_partition : ALL});ts=1577718040933619;expiry=-1,ttl=0}, column comment atomic_cell{;ts=1577718040933619;expiry=-1,ttl=0}, column compaction atomic_cell{({class : SizeTieredCompactionStrategy});ts=1577718040933619;expiry=-1,ttl=0}, column compression atomic_cell{({sstable_compression : org.apache.cassandra.io.compress.LZ4Compressor});ts=1577718040933619;expiry=-1,ttl=0}, column crc_check_chance atomic_cell{1;ts=1577718040933619;expiry=-1,ttl=0}, column dclocal_read_repair_chance atomic_cell{0.1;ts=1577718040933619;expiry=-1,ttl=0}, column default_time_to_live atomic_cell{0;ts=1577718040933619;expiry=-1,ttl=0}, column extensions atomic_cell{;ts=1577718040933619;expiry=-1,ttl=0}, column flags atomic_cell{compound;ts=1577718040933619;expiry=-1,ttl=0}, column gc_grace_seconds atomic_cell{864000;ts=1577718040933619;expiry=-1,ttl=0}, column id atomic_cell{255c3950-2b15-11ea-b830-000000000001;ts=1577718040933619;expiry=-1,ttl=0}, column max_index_interval atomic_cell{2048;ts=1577718040933619;expiry=-1,ttl=0}, column memtable_flush_period_in_ms atomic_cell{0;ts=1577718040933619;expiry=-1,ttl=0}, column min_index_interval atomic_cell{128;ts=1577718040933619;expiry=-1,ttl=0}, column read_repair_chance atomic_cell{0;ts=1577718040933619;expiry=-1,ttl=0}, column speculative_retry atomic_cell{99.0PERCENTILE;ts=1577718040933619;expiry=-1,ttl=0}}}, {canonical_mutation: table_id 24101c25-a2ae-3af7-87c1-b40ee1aca33f schema_version a4fe505b-053c-3486-9d6a-bd283d568c10 partition_key pk{00026b73} partition_tombstone {tombstone: none}{row {position: clustered,ckp{000374616200026964},0} tombstone {row_tombstone: none} marker {row_marker: }, column clustering_order atomic_cell{NONE;ts=1577718040933619;expiry=-1,ttl=0}, column column_name_bytes atomic_cell{6964;ts=1577718040933619;expiry=-1,ttl=0}, column kind atomic_cell{partition_key;ts=1577718040933619;expiry=-1,ttl=0}, column position atomic_cell{0;ts=1577718040933619;expiry=-1,ttl=0}, column type atomic_cell{int;ts=1577718040933619;expiry=-1,ttl=0}}}, {canonical_mutation: table_id cc7c7069-3740-33c1-92a4-c3de78dbd2c4 schema_version 723420b4-8501-3ecd-92dd-b01de2e227f6 partition_key pk{00026b73} partition_tombstone {tombstone: none}}, {canonical_mutation: table_id 0feb57ac-311f-382f-ba6d-9024d305702f schema_version a7860a8c-d89c-30b0-98ba-279ae67fae1e partition_key pk{00026b73} partition_tombstone {tombstone: none}}, {canonical_mutation: table_id 5e7583b5-f3f4-3af1-9a39-b7e1d6f5f11f schema_version d4e2a352-3e28-3297-b746-461236df0a1e partition_key pk{00026b73} partition_tombstone {tombstone: none}}, {canonical_mutation: table_id 5d912ff1-f759-3665-b2c8-8042ab5103dd schema_version d15e4b38-4294-3cf3-b7e2-5485bd5c9367 partition_key pk{00026b73} partition_tombstone {tombstone: none}{row {position: clustered,ckp{0003746162},0} tombstone {row_tombstone: none} marker {row_marker: }, column cdc atomic_cell{({enabled : false}, {postimage : false}, {preimage : false}, {ttl : 86400});ts=1577718040933619;expiry=-1,ttl=0}, column version atomic_cell{255c3951-2b15-11ea-b830-000000000001;ts=1577718040933619;expiry=-1,ttl=0}}}, {canonical_mutation: table_id abac5682-dea6-31c5-b535-b3d6cffd0fb6 schema_version 546e75bd-58af-3eb4-8017-5f7c54bad49c partition_key pk{00026b73} partition_tombstone {tombstone: none}{row {position: clustered,ckp{},0} tombstone {row_tombstone: none} marker {row_marker: }, column durable_writes atomic_cell{true;ts=1577718037239063;expiry=-1,ttl=0}, column replication atomic_cell{({class : org.apache.cassandra.locator.SimpleStrategy}, {replication_factor : 3});ts=1577718037239063;expiry=-1,ttl=0}}}}}
```

(this is a vector<canonical_mutation> object, so the word "canonical_mutation" appears multiple times).

### DIFF
--- a/atomic_cell.cc
+++ b/atomic_cell.cc
@@ -233,6 +233,27 @@ operator<<(std::ostream& os, const atomic_cell& ac) {
     return os << atomic_cell_view(ac);
 }
 
+std::ostream&
+operator<<(std::ostream& os, const atomic_cell_view::printer& acvp) {
+    auto& type = acvp._type;
+    auto& acv = acvp._cell;
+    if (acv.is_live()) {
+        return fmt_print(os, "atomic_cell{{{};ts={:d};expiry={:d},ttl={:d}}}",
+            type.to_string(acv.value().linearize()),
+            acv.timestamp(),
+            acv.is_live_and_has_ttl() ? acv.expiry().time_since_epoch().count() : -1,
+            acv.is_live_and_has_ttl() ? acv.ttl().count() : 0);
+    } else {
+        return fmt_print(os, "atomic_cell{{DEAD;ts={:d};deletion_time={:d}}}",
+            acv.timestamp(), acv.deletion_time().time_since_epoch().count());
+    }
+}
+
+std::ostream&
+operator<<(std::ostream& os, const atomic_cell::printer& acp) {
+    return operator<<(os, static_cast<const atomic_cell_view::printer&>(acp));
+}
+
 std::ostream& operator<<(std::ostream& os, const atomic_cell_or_collection::printer& p) {
     if (!p._cell._data.get()) {
         return os << "{ null atomic_cell_or_collection }";
--- a/atomic_cell.cc
+++ b/atomic_cell.cc
@@ -214,6 +214,25 @@ size_t atomic_cell_or_collection::external_memory_usage(const abstract_type& t) 
         + imr_object_type::size_overhead + external_value_size;
 }
 
+std::ostream&
+operator<<(std::ostream& os, const atomic_cell_view& acv) {
+    if (acv.is_live()) {
+        return fmt_print(os, "atomic_cell{{{};ts={:d};expiry={:d},ttl={:d}}}",
+            to_hex(acv.value().linearize()),
+            acv.timestamp(),
+            acv.is_live_and_has_ttl() ? acv.expiry().time_since_epoch().count() : -1,
+            acv.is_live_and_has_ttl() ? acv.ttl().count() : 0);
+    } else {
+        return fmt_print(os, "atomic_cell{{DEAD;ts={:d};deletion_time={:d}}}",
+            acv.timestamp(), acv.deletion_time().time_since_epoch().count());
+    }
+}
+
+std::ostream&
+operator<<(std::ostream& os, const atomic_cell& ac) {
+    return os << atomic_cell_view(ac);
+}
+
 std::ostream& operator<<(std::ostream& os, const atomic_cell_or_collection::printer& p) {
     if (!p._cell._data.get()) {
         return os << "{ null atomic_cell_or_collection }";
--- a/canonical_mutation.cc
+++ b/canonical_mutation.cc
@@ -35,6 +35,7 @@
 #include "idl/uuid.dist.impl.hh"
 #include "idl/keys.dist.impl.hh"
 #include "idl/mutation.dist.impl.hh"
+#include <iostream>
 
 canonical_mutation::canonical_mutation(bytes data)
         : _data(std::move(data))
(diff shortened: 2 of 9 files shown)


avikivity:
@kbr- you were randomly selected for review as you touched the collection_mutation_view stuff recently.

collection_mutation.cc
```diff
@@ -153,6 +153,27 @@ api::timestamp_type collection_mutation_view::last_update(const abstract_type& t
     ));
 }
 
+std::ostream& operator<<(std::ostream& os, const collection_mutation_view::printer& cmvp) {
+    fmt::print(os, "{{collection_mutation_view ");
+    cmvp._cmv.with_deserialized(cmvp._type, [&os, &type = cmvp._type] (const collection_mutation_view_description& cmvd) {
+        auto&& ctype = static_cast<const collection_type_impl&>(type);
+        auto&& key_type = ctype.name_comparator();
+        auto&& value_type = ctype.value_comparator();
```
kbr-: I'm afraid this doesn't work for non-frozen UDTs, where value types depend on the keys. Check out `collection_mutation_view_description::materialize` for an example on how to deal with UDTs:
```
        for (auto&& e : cells) {
            m.cells.emplace_back(to_bytes(e.first), atomic_cell(*utype.type(deserialize_field_index(e.first)), e.second));
        }
```

database.cc
```diff
@@ -1644,6 +1644,27 @@ operator<<(std::ostream& os, const atomic_cell& ac) {
...
+            acv.is_live_and_has_ttl() ? acv.ttl().count() : 0);
+    } else {
+        return fmt_print(os, "atomic_cell{{DEAD;ts={:d};deletion_time={:d}}}",
+            acv.timestamp(), acv.deletion_time().time_since_epoch().count());
+    }
+}
+
+std::ostream&
+operator<<(std::ostream& os, const atomic_cell::printer& acp) {
+    return operator<<(os, static_cast<atomic_cell_view::printer>(acp));
```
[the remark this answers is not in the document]
avikivity: It should. Doesn't matter in practice but it's clearer (and was the intent).

database.cc
```diff
@@ -1644,6 +1644,27 @@ operator<<(std::ostream& os, const atomic_cell& ac) {
     return os << atomic_cell_view(ac);
 }
 
+std::ostream&
```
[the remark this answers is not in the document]
avikivity: It got there because the non-type-aware printer for atomic_cell is there. I'll add a patch that moves the existing printer to atomic_cell.cc, and then this will naturally go there as well.

collection_mutation.cc
avikivity: I knew that conflating UDTs with collections will hit us. I'll try to fix it.

kbr-: I opened an issue:
https://github.com/scylladb/scylla/issues/5534

avikivity:
Update:
 - new patch moving old atomic_cell pretty printers to atomic_cell.cc
 - move new atomic_cell pretty printers to atomic_cell.cc
 - convert unintended slicing copy constructor to cast to base class reference
 - default "first" to true when printing collections
 - handle UDTs when printing collections-that-can-also-be-UDTs
 - reduced some curlies in collection_mutation_view printing

kbr-:
```

        auto&& ctype = static_cast<const collection_type_impl&>(type);
        auto&& key_type = ctype.name_comparator();
        auto&& value_type = ctype.value_comparator(); 
```
This probably works (by accident), but probably wasn't intended and should be moved to the `collection_type_impl` visit branch

avikivity:
> ```
> 
>         auto&& ctype = static_cast<const collection_type_impl&>(type);
>         auto&& key_type = ctype.name_comparator();
>         auto&& value_type = ctype.value_comparator(); 
> ```
> 
> This probably works (by accident), but probably wasn't intended and should be moved to the `collection_type_impl` visit branch

Yup, will fix.

avikivity:
Updated:
 - moved collection-specific casts and accesses to inside the collection-specific branch of the collection/UDT visitor

avikivity:
@scylladb/scylla-maint ping

collection_mutation.cc
```diff
@@ -153,6 +153,44 @@ api::timestamp_type collection_mutation_view::last_update(const abstract_type& t
...
+            for (auto&& [key, value] : cmvd.cells) {
+                if (!first) {
+                    fmt::print(os, ", ");
+                }
+                fmt::print(os, "{}: {}", key_type->to_string(key), atomic_cell_view::printer(*value_type, value));
+                first = false;
+            }
+        },
+        [&] (const user_type_impl& utype) {
+            for (auto&& e : cmvd.cells) {
```
denesb: nit: `for (auto&& [raw_idx, val] : cmvd.cells) {`

mutation_partition_view.cc
```diff
@@ -45,6 +45,10 @@
 
 using namespace db;
 
+GCC6_CONCEPT(static_assert(MutationViewVisitor<mutation_partition_view_virtual_visitor>);)
```
denesb: Not suggesting this is wrong or has to be changed, it is just strange to see a concept "floating" like that. When will this be checked? When the TU is compiled?

canonical_mutation.cc
```diff
@@ -89,3 +90,81 @@ mutation canonical_mutation::to_mutation(schema_ptr s) const {
...
+    return ret;
+}
+
+std::ostream& operator<<(std::ostream& os, const canonical_mutation& cm) {
+    auto in = ser::as_input_stream(cm._data);
+    auto mv = ser::deserialize(in, boost::type<ser::canonical_mutation_view>());
+    column_mapping mapping = mv.mapping();
+    auto partition_view = mutation_partition_view::from_view(mv.partition());
+    fmt::print(os, "{{canonical_mutation: ");
+    fmt::print(os, "table_id {} schema_version {} ", mv.table_id(), mv.schema_version());
```
denesb: Would it be helpful to also print the `keyspace.table_name`?

canonical_mutation.cc
```diff
@@ -89,3 +90,81 @@ mutation canonical_mutation::to_mutation(schema_ptr s) const {
...
+    fmt::print(os, "partition_key {} ", mv.key());
```
denesb: There is a pretty printer for partition keys as well. Wouldn't it be better to use that instead of the raw hex representation?

canonical_mutation.cc
```diff
@@ -89,3 +90,81 @@ mutation canonical_mutation::to_mutation(schema_ptr s) const {
...
+        }
+        virtual void accept_row_tombstone(range_tombstone rt) override {
+            print_separator();
+            fmt::print(_os, "row tombstone {}", rt);
+        }
+        virtual void accept_row(position_in_partition_view pipv, row_tombstone rt, row_marker rm, is_dummy, is_continuous) override {
+            if (_in_row) {
+                fmt::print(_os, "}}, ");
+            }
+            fmt::print(_os, "{{row {} tombstone {} marker {}", pipv, rt, rm);
```
denesb: Pretty print the pip?

collection_mutation.cc
avikivity: Yes.

canonical_mutation.cc
avikivity: I have a follow-on patch (already on the list) that adds support for pretty-printing pipvs.

mutation_partition_view.cc
avikivity: Yes. I don't want to run the check in the header, where it will be checked many times.

canonical_mutation.cc
avikivity: Yes, in my use case I guessed the table from the column names, but it's better to print them.

avikivity: In fact the keyspace and table names are not available in the canonical mutation.

canonical_mutation.cc
avikivity: Again I can't use it because the schema isn't available.

We could add a pretty printer that makes use of the/a schema, but that's against the spirit of canonical_mutation which is supposed to be portable.

avikivity:
Updated:
 - use structured binding when decomposing the key/value of pretty-printed UDTs

Not done:
 - print keyspace and table names (not available in canonical_mutation)
 - pretty print partition keys (requires schema)

The latter may be done if schema information is made available, but that requires examining the call sites.

nyh:
Committed to "next". Thanks.

atomic_cell.cc
```diff
@@ -233,6 +233,27 @@ operator<<(std::ostream& os, const atomic_cell& ac) {
     return os << atomic_cell_view(ac);
 }
 
+std::ostream&
+operator<<(std::ostream& os, const atomic_cell_view::printer& acvp) {
+    auto& type = acvp._type;
+    auto& acv = acvp._cell;
+    if (acv.is_live()) {
+        return fmt_print(os, "atomic_cell{{{};ts={:d};expiry={:d},ttl={:d}}}",
```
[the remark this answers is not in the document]
avikivity: This is already merged, I'll send a follow-up.

I had an idea to pretty print everything as json. This is more verbose but can easily be pretty printed and diffed when debugging something.

e.g. atomic_cell would be pretty printed as

    {'type':'atomic_cell','timestamp':4341232445,'expiry':234242,'ttl':4342}

and it would nest naturally within other json-printed objects.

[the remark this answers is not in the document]
avikivity: btw, this came from the non-type-aware pretty printer and is consistent with it.

I'll update both.

Issue #5527, opened 2019-12-26
Repair in abort_repair and no_corrupt_repair fails, sometimes with a coredump, in 50gb 4 days


ShlomiBalalis:
This is Scylla's bug tracker, to be used for reporting bugs only.
If you have a question about Scylla, and not a bug, please ask it in
our mailing-list at scylladb-dev@googlegroups.com or in our slack channel.

- [x] I have read the disclaimer above, and I am reporting a suspected malfunction in Scylla.

*Installation details*
Scylla version (or git commit hash):3.2.rc3-0.20191215.39b17be5620
Cluster size:6
OS (RHEL/CentOS/Ubuntu/AWS AMI):ami-0cab4c6eceb3249ac (Ireland)
Instance Type: i3.4xlarge

Node list:
```
longevity-tls-50gb-4d-3-2-db-node-f6beb13b-1 [34.245.5.120 | 10.0.212.62]
longevity-tls-50gb-4d-3-2-db-node-f6beb13b-2 [52.50.13.163 | 10.0.241.156]
longevity-tls-50gb-4d-3-2-db-node-f6beb13b-3 [34.244.216.146 | 10.0.217.173]
longevity-tls-50gb-4d-3-2-db-node-f6beb13b-4 [52.50.43.34 | 10.0.122.241]
longevity-tls-50gb-4d-3-2-db-node-f6beb13b-5 [34.245.33.127 | 10.0.172.251]
longevity-tls-50gb-4d-3-2-db-node-f6beb13b-6 [34.244.172.83 | 10.0.239.192]
longevity-tls-50gb-4d-3-2-db-node-f6beb13b-7 [54.154.251.179 | 10.0.239.127]
```

In both the 4 occurrences of abort_repair, and in 3 of the 7 occurrences of no_corrupt_repair, the repair command fails in similar fashion:

abort_repair failures:
(target node#4)
```
2019-12-18 00:33:04,674
Traceback (most recent call last):
  File "/sct/sdcm/nemesis.py", line 1179, in wrapper
    result = method(*args, **kwargs)
  File "/sct/sdcm/nemesis.py", line 1394, in disrupt
    self.call_random_disrupt_method()
  File "/sct/sdcm/nemesis.py", line 622, in call_random_disrupt_method
    disrupt_method()
  File "/sct/sdcm/nemesis.py", line 905, in disrupt_abort_repair
    self.repair_nodetool_repair()
[...]

PR #5535, opened 2019-12-31
repair: Avoid duplicated partition_end write

asias:
Consider this:

1) Write partition_start of p1
2) Write clustering_row of p1
3) Write partition_end of p1
4) Repair aborted before writing partition_start of p2
5) Repair calls repair_row_level_stop() to tear down which calls
   wait_for_writer_done(). A duplicate partition_end is written.

To fix, track the partition_start and partition_end written, avoid
unpaired writes.

Backports: 3.1 and 3.2
Fixes: #5527

### DIFF
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -448,6 +448,10 @@ class repair_writer {
     std::vector<std::optional<seastar::queue<mutation_fragment_opt>>> _mq;
     // Current partition written to disk
     std::vector<lw_shared_ptr<const decorated_key_with_hash>> _current_dk_written_to_sstable;
+    // Is current partition still open. A partition is opened when a
+    // partition_start is written and is closed when a partition_end is
+    // written.
+    std::vector<bool> _partition_opened;
 public:
     repair_writer(
             schema_ptr schema,
@@ -462,10 +466,13 @@ public:
     future<> write_start_and_mf(lw_shared_ptr<const decorated_key_with_hash> dk, mutation_fragment mf, unsigned node_idx)  {
         _current_dk_written_to_sstable[node_idx] = dk;
         if (mf.is_partition_start()) {
-            return _mq[node_idx]->push_eventually(mutation_fragment_opt(std::move(mf)));
+            return _mq[node_idx]->push_eventually(mutation_fragment_opt(std::move(mf))).then([this, node_idx] {
+                _partition_opened[node_idx] = true;
+            });
         } else {
             auto start = mutation_fragment(partition_start(dk->dk, tombstone()));
             return _mq[node_idx]->push_eventually(mutation_fragment_opt(std::move(start))).then([this, node_idx, mf = std::move(mf)] () mutable {
+                _partition_opened[node_idx] = true;
                 return _mq[node_idx]->push_eventually(mutation_fragment_opt(std::move(mf)));
             });
         }
@@ -475,6 +482,7 @@ public:
         _writer_done.resize(_nr_peer_nodes);
         _mq.resize(_nr_peer_nodes);
         _current_dk_written_to_sstable.resize(_nr_peer_nodes);
+        _partition_opened.resize(_nr_peer_nodes, false);
     }
 
     void create_writer(unsigned node_idx) {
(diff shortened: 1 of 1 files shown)


repair/row_level.cc
```diff
@@ -439,6 +439,11 @@ using is_local_reader = bool_class<class is_local_reader_tag>;
 
 };
 
+struct repair_writer_stats {
+    uint64_t nr_parttion_start = 0;
+    uint64_t nr_parttion_end = 0;
+};
+
 class repair_writer {
```
avikivity: I don't understand this. Why do we need to count and compare starts and ends? If we abort something, we should tear down the stream, not try to clean it up. If the other side is dead, it won't respond to additional writes, and the socket buffer may be full.

repair/row_level.cc
```diff
@@ -462,10 +468,13 @@ class repair_writer {
     future<> write_start_and_mf(lw_shared_ptr<const decorated_key_with_hash> dk, mutation_fragment mf, unsigned node_idx)  {
         _current_dk_written_to_sstable[node_idx] = dk;
         if (mf.is_partition_start()) {
-            return _mq[node_idx]->push_eventually(mutation_fragment_opt(std::move(mf)));
+            return _mq[node_idx]->push_eventually(mutation_fragment_opt(std::move(mf))).then([this, node_idx] {
+                _stats[node_idx].nr_parttion_start++;
+            });
```
avikivity: Is this racy? Suppose we start to push and block, then someone stops the repair. We see unincremented nr_parttion_start.

repair/row_level.cc
```diff
@@ -519,12 +529,28 @@ class repair_writer {
         t.stream_in_progress());
     }
 
+    future<> write_partition_end(unsigned node_idx) {
+        if (_stats[node_idx].nr_parttion_start == _stats[node_idx].nr_parttion_end + 1) {
+            return _mq[node_idx]->push_eventually(mutation_fragment(partition_end())).then([this, node_idx] {
+                _stats[node_idx].nr_parttion_end++;
+            });
```
avikivity: Again this looks race prone (since repair aborts are asynchronous to this code), nr_parttion_end increments can be delayed.

repair/row_level.cc
asias: This has nothing to do with streams. partition_end is not synced between nodes on wire. We generate partition_end on the node that needs to write the mutation fragments. partition_end is implied when a new partition_start is received.

repair/row_level.cc
avikivity: Why not use seastar::queue::abort()? It is designed for asynchronous operation, and can be called when the producer is blocking in push_eventually().

repair/row_level.cc
asias: when no more fragments from the peer, i.e., the repair is done, we write a partition_end in repair_row_level_stop.

repair/row_level.cc
asias: Abort is actually synchronous, it checks if the abort flag is set in each repair steps inside the main repair loop.

repair/row_level.cc
asias: We will wait for the push to complete before we check the abort flag.

repair/row_level.cc
avikivity: I still don't understand it.

There is a stream of fragments from one end of the queue to the other, no? Why do you need to count them?

repair/row_level.cc
avikivity: I don't understand either the problem or the solution.

avikivity: If it is syncronous, why is there a problem? Why did we call write_partition_end()? What's the call site from abort's persepective?

Maybe abort shouldn't write the partition end if no partition is open?

repair/row_level.cc
avikivity: Where?

repair/row_level.cc
asias: partition_end is not transferred between nodes

repair/row_level.cc
asias: See ri.check_in_abort() in the code.

asias:
@avikivity to set the flag before push is problematic 
_partition_opened = true
push partition_start returns exceptional future, so the partition_start is not actually written

if (_partition_opened) {
      push partition_end        //  wrong, write parittion_end without parittion_end
}

slivne:
@asias do you need avi to review this - did you address all comments

asias:
@avikivity  Can you please take a look at the current version 61d6e5efefb6459ba0abafc49acdba286c0b00d5, which uses a flag as you suggested. I think it is safe to set the flag after the pushing of the mutation fragment.

avikivity:
Queued as 401854dbafbe1e6be00c9060eff3827bd4786165

PR #5539, opened 2019-12-31
dist/ami: update packer to 1.5.1

syuu1228:
Update Packer to 1.5.1.
Needed to rename clean_ami_name -> clean_resource_name on scylla.json, since
the variable name had been changed.
Also fixed checksum verification code, trimmed unwanted extra strings
from sha256sum output.

### DIFF
--- a/dist/ami/build_ami.sh
+++ b/dist/ami/build_ami.sh
@@ -194,16 +194,16 @@ if [ ! -f variables.json ]; then
 fi
 
 if [ ! -d packer ]; then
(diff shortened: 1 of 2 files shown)


fruch:
@syuu1228 is there a specific reason/feature we are upgrading for ?

fruch:
a mirror commit for this one in scylla-cloud-image:
https://github.com/scylladb/scylla-cloud-image/commit/15a78625eecf6aa88871e5abfed95881f5e766ff

syuu1228:
@fruch No specific reason/feature we are upgrading for, but we are updating the tool periodically otherwise it would be very old / outdated version.
Also, old version may not support new EC2 features, we haven't faced such issue yet though.

Issue #5364, opened 2019-11-27
Support pipelining in Redis API


nyh:
As explained in Redis's documentation in https://redis.io/topics/pipelining, the Redis protocol inherently supports pipelining: The client may send several requests on the same connection before waiting for responses, the server will read all of them, and start all of them, and will then return the responses in one write (in the same order as the requests, which requires buffering of the responses).

Looking at our code, if I understand correctly it seems that  `redis_server::connection::process()` has no problem with the client sending multiple requests - but it just parses one, waits for it to be completed and write the reply - and only then starts to parse and process the next one. We could reduce the latency and the amount of work that we need to do on writing the response (as one write instead of many small writes), if we add full support for pipelining - namely - start multiple requests read from the same socket before waiting for all of them to complete.

This issue is about performance, not correctness. As far as I can tell, the existing behavior is correct, just less efficient than it could have been.

@fastio

fastio:
@nyh I want to try to work around this feature that support pipelining in Redis.

My simple solution is described as follows:

* Allocate one fixed size queue when the connection is created.
* Parse as many requests on the connection as possible util the queue is full.
* Process the requests in the queue as many as possible util the queue is empty.

Please give me some advice. Thanks.

avikivity:
Processing should be done in parallel when possible. This will overlap the latency of contacting other nodes.

One problem is that it's not always possible to process in parallel, requests for the same key should be processed serially, otherwise a SET a x followed by a GET a might execute GET a before the SET.

To serialize, we can use the same locking that LWT uses (@gleb-cloudius where is it?) Note that acquiring the lock, and sending the response, must be in the original order, but everything in between can be parallelized.

gleb-cloudius:
On Sun, Dec 15, 2019 at 05:47:38AM -0800, Avi Kivity wrote:
> Processing should be done in parallel when possible. This will overlap the latency of contacting other nodes.
> 
> One problem is that it's not always possible to process in parallel, requests for the same key should be processed serially, otherwise a SET a x followed by a GET a might execute GET a before the SET.
> 
> To serialize, we can use the same locking that LWT uses (@gleb-cloudius where is it?) Note that acquiring the lock, and sending the response, must be in the original order, but everything in between can be parallelized.
> 
It is in a paxos_state::_locks. Currently taken by
paxos_state::prepare() and paxos_state::accept() but I am moving it up
the call chain. FWIW it is per token, not per key.

--
			Gleb.

[...]

PR #5553, opened 2020-01-03
Redis: support pipelining

fastio:
Fixes: #5364

### DIFF
--- a/redis-test/test_strings.py
+++ b/redis-test/test_strings.py
@@ -21,6 +21,8 @@
 import pytest
 import redis
 import logging
+import socket
+import time 
 from util import random_string, connect
 
 logger = logging.getLogger('redis-test')
@@ -106,3 +108,33 @@ def test_select():
         raise Exception('Expect that `SELECT 16` does not work')
     except redis.exceptions.ResponseError as ex:
         assert str(ex) == 'invalid DB index'
+
+def test_pipeline():
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.connect(("127.0.0.1", 6379))
+    s.send('*1\r\n$4\r\nPING\r\n*1\r\n$4\r\nPING\r\n*1\r\n$4\r\nPING\r\n*1\r\n$4\r\nPING\r\n'.encode())
+    assert s.recv(1024).decode() == '+PONG\r\n+PONG\r\n+PONG\r\n+PONG\r\n'
+    s.close()
+
+def test_pipeline_vs_normal():
+    r = connect()
+    time_pipeline_start = time.time()
+    with r.pipeline(transaction=False) as p:
+        p.set(random_string(10), random_string(100))
+        p.set(random_string(10), random_string(100))
+        p.set(random_string(10), random_string(100))
+        p.set(random_string(10), random_string(100))
+        p.set(random_string(10), random_string(100))
+        p.set(random_string(10), random_string(100))
+        p.execute()
+    time_pipeline = time.time() - time_pipeline_start 
+    time_normal_start = time.time()
+    r.set(random_string(10), random_string(100))
+    r.set(random_string(10), random_string(100))
+    r.set(random_string(10), random_string(100))
+    r.set(random_string(10), random_string(100))
+    r.set(random_string(10), random_string(100))
+    r.set(random_string(10), random_string(100))
+    time_normal = time.time() - time_normal_start
+
+    assert time_pipeline < time_normal
--- a/redis/commands.cc
+++ b/redis/commands.cc
@@ -98,7 +98,7 @@ future<redis_message> select::execute(service::storage_proxy&, redis::redis_opti
     if (_index < 0 || static_cast<size_t>(_index) >= options.get_total_redis_db_count()) {
         throw invalid_db_index_exception();
     }
-    options.set_keyspace_name(sprint("REDIS_%zu", static_cast<size_t>(_index)));
+    options.set_db_index(static_cast<unsigned>(_index));
     return redis_message::ok();
 }
 
--- a/redis/keyspace_utils.cc
+++ b/redis/keyspace_utils.cc
@@ -180,7 +180,7 @@ future<> create_keyspace_if_not_exists_impl(db::config& config, int default_repl
         return service::get_local_migration_manager().announce_new_column_family(schema, false);
     };
     // create 16 default database for redis.
-    return parallel_for_each(boost::irange<unsigned>(0, 16), [keyspace_gen = std::move(keyspace_gen), table_gen = std::move(table_gen)] (auto c) {
+    return parallel_for_each(boost::irange<unsigned>(0, config.redis_database_count()), [keyspace_gen = std::move(keyspace_gen), table_gen = std::move(table_gen)] (auto c) {
         auto ks_name = sprint("REDIS_%d", c);
         return keyspace_gen(ks_name).then([ks_name, table_gen] {
             return when_all_succeed(
--- a/redis/locks_context.cc
+++ b/redis/locks_context.cc
@@ -1,0 +1,26 @@
+/*
+ * Copyright (C) 2020 pengjian.uestc @ gmail.com
+ */
+
+/*
+ * This file is part of Scylla.
+ *
+ * Scylla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Scylla is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "redis/locks_context.hh"
+
+namespace redis {
+
+}
--- a/redis/locks_context.hh
+++ b/redis/locks_context.hh
@@ -1,0 +1,126 @@
+/*
+ * Copyright (C) 2020 pengjian.uestc @ gmail.com
+ */
+
+/*
+ * This file is part of Scylla.
+ *
+ * Scylla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Scylla is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "db/config.hh"
+#include "db/timeout_clock.hh"
+#include "bytes.hh"
+#include "seastar/core/future.hh"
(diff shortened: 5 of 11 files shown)


avikivity:
@nyh please review

redis/locks_context.hh
```diff
@@ -0,0 +1,126 @@
...
+    auto it = locks.find(key);
+    if (it != locks.end() && (*it).second.current() == 1) {
+        locks.erase(it);
+    }   
+}
+
+template<typename Func>
+static
+futurize_t<std::result_of_t<Func()>> with_locked_key(locks_map& locks, const bytes& key, clock_type::time_point timeout, Func func) {
+    return with_semaphore(get_semaphore_for_key(locks, key), 1, timeout - clock_type::now(), std::move(func)).finally([key, &locks] {
```
nyh: @gleb-cloudius I just realized for with_semaphore() we only have a variant getting a duration timeout - whereas in get_units() also supports a time_point timeout. Was this intentional, or just an oversight?
Because this missing variant forced this Scylla code to convert the time_point timeout into a duration, for no good reason.

redis/locks_context.hh
```diff
@@ -0,0 +1,126 @@
...
+
+static thread_local key_locks_manager _key_locks_manager;
+
+static inline locks_context& get_locks_context(unsigned db_index) {
+    return _key_locks_manager.get_locks_context(db_index);
+}
+
+template<typename Func>
+static
+futurize_t<std::result_of_t<Func()>> with_locked_key(locks_map& locks, const bytes& key, clock_type::time_point timeout, Func func) {
```
nyh: Just curious, what's the difference between this with_locked_key and the internal::with_locked_key it calls? Why do we need this extra layer?

redis/query_utils.cc
```diff
@@ -70,22 +71,25 @@ class strings_result_builder {
...
-    dht::partition_range_vector partition_ranges;
-    partition_ranges.emplace_back(std::move(partition_range));
-    auto read_consistency_level = options.get_read_consistency_level();
     db::timeout_clock::time_point timeout = db::timeout_clock::now() + options.get_read_timeout();
-    return proxy.query(schema, make_lw_shared(std::move(cmd)), std::move(partition_ranges), read_consistency_level, {timeout, permit, service::client_state::for_internal_calls()}).then([ps, schema] (auto qr) {
-        return query::result_view::do_with(*qr.query_result, [&] (query::result_view v) {
-            auto pd = make_lw_shared<strings_result>();
-            v.consume(ps, strings_result_builder(pd, schema, ps));
-            return pd;
+    return redis::with_strings_locked_key(redis::get_locks_context(options.db_index()), key, timeout, [timeout, &proxy, &options, &key, permit] () mutable {
```
nyh: I now I understand that the commit message slightly underplays what is being changed here.

As far as I understand now, what you're changing here doesn't **just** effect pipelining in a single request. 
Even if you have multiple requests from multiple connections, this "lock" serializes them so only one happens at a time. Is this intentional?

Although it sounds like this serialization is a good idea in general, but there's an important catch... It only happens on a single shard (the lock manager is thread-local).... So concurrent updates to the same key that happened to have been received on different shards (or even different nodes!) will happen in parallel, without the lock protecting anything. This is good enough for pipelining (since the same request belongs to a single shard) but does it offer anything more?

So given that this global - but per-shard - lock doesn't protect much besides pipelining, why do we make it global at all and not a per-request map, used just for pipelining?
Is it because this way is easier? Or there's something actually better in this approach?

In any case should probably have a comment above the "lock manager" explaining what exactly it tries to serialize, what it doesn't serialize (write on different shards) and why it does what we want related to pipelining.

redis/server.cc
```diff
@@ -152,26 +153,57 @@ redis_server::connection::~connection() {
...
+                return _read_buf.consume(_parser).then([this] {
+                    if (_parser.eof()) {
+                        return make_ready_future<>();
+                    }
+                    ++_server._stats._requests_serving;
+                    utils::latency_counter lc;
+                    lc.start();
+                    promise<redis_server::result> pr;
+                    auto fut = pr.get_future();
+                    (void)process_request_internal(std::move(_parser.get_request())).then_wrapped([this, pr = std::move(pr), lc = std::move(lc)] (auto&& fut) mutable {
```
nyh: "fut" here is different from the "fut" above? It's confusing...

redis/server.cc
```diff
@@ -152,26 +153,57 @@ redis_server::connection::~connection() {
...
+                        fut.forward_to(std::move(pr)); 
+                    });
+                    return _queue.push_eventually(std::move(fut));
+                });
             });
         });
-    }).finally([this] {
+    };
+    auto writer = [this] {
+        return do_until([this] { return _read_buf.eof() && _queue.empty(); }, [this] {
```
nyh: Is this eof() && empty() condition safe?
I'm worried that the reader code may read a request, before pushing onto the queue a future for handling this request, it defers for some reason (?) - so the writer thinks it has nothing to wait for, and fails to write the last reply....
One safer alternative is to have a special "eof" (end-of-file) marker on _queue, and writer will keep reading until it reads this eof marker. If there is no easy way to designate one value for eof, a simple approach is to use std::optional<T> for the queue's items - and a disengaged optional will mark the eof.

redis/server.cc
```diff
@@ -152,26 +153,57 @@ redis_server::connection::~connection() {
...
+                        write_reply(std::move(result));
+                    }
+                    catch (redis_exception& e) {
+                        write_reply(e);
+                    }
+                    catch (std::exception& e) {
+                        write_reply(redis_exception { e.what() });
+                    }
+                    catch (...) {
+                        write_reply(redis_exception { "Unknown exception" });
```
nyh: I know you're just moving pre-existing code around, but you could have done here something like:

```c++
fmt::print("Internal error: {}", std::current_exception());
```
So you'll have more information on the exception, not just the string "Unknown exception".

redis-test/test_strings.py
```diff
@@ -106,3 +108,33 @@ def test_select():
...
+    time_normal_start = time.time()
+    r.set(random_string(10), random_string(100))
+    r.set(random_string(10), random_string(100))
+    r.set(random_string(10), random_string(100))
+    r.set(random_string(10), random_string(100))
+    r.set(random_string(10), random_string(100))
+    r.set(random_string(10), random_string(100))
+    time_normal = time.time() - time_normal_start
+
+    assert time_pipeline < time_normal
```
nyh: This test can be "flaky", meaning it will randomly fail, just because of timing issues. E.g., the pipeline should have slightly lower latency - but if in parallel something else runs on the test machine, suddenly it can run a bit slower and the test will fail.
So I don't think we can have such a test. 

On the other hand, there are things we **can** test, and you don't test. We can and should test for correctness - namely that if we do 10 writes to different keys, all of them took effect and the response includes all of them (and we don't e.g., miss one of them) and in the right order. We should also test that  operations on the same key (e.g., read and write) are not reordered.

redis/options.hh
```diff
@@ -47,19 +47,15 @@ class redis_options {
...
     explicit redis_options(const db::consistency_level rcl,
         const db::consistency_level wcl,
         const timeout_config& tc,
         auth::service& auth,
         const socket_address addr,
         size_t total_redis_db_count)
-		: redis_options(0, rcl, wcl, tc, auth, addr, total_redis_db_count)
-	{
-	}
+    : redis_options(0, rcl, wcl, tc, auth, addr, total_redis_db_count)
```
nyh: This cleanups have nothing to do with this patch (adding a test), but I'll let it slide :-)

redis/mutation_utils.cc
```diff
@@ -80,14 +83,30 @@ mutation make_tombstone(service::storage_proxy& proxy, const redis_options& opti
...
+        return parallel_for_each(keys.begin(), keys.end(), [&proxy, timeout, write_consistency_level, &options, permit, &ctx] (const bytes& key) mutable {
+            return redis::with_locked_key(ctx.locks, key, timeout, [&proxy, timeout, write_consistency_level, &options, key, permit, &ctx] () mutable {
+                auto m = make_tombstone(proxy, options, ctx.table, key);
+                return proxy.mutate(std::vector<mutation> {std::move(m)}, write_consistency_level, timeout, nullptr, permit);
+            });
         });
-    };  
-    return parallel_for_each(tables.begin(), tables.end(), remove);
+    };
+    return do_with(std::move(contexts), [remove = std::move(remove)] (auto& contexts) mutable {
```
nyh: I don't think we really need do_with for "contexts", as it is a more-or-less constant array (for a specific shard and db) we could have kept somewhere central. But I guess we can start this way.

redis/commands.cc
```diff
@@ -98,7 +98,7 @@ future<redis_message> select::execute(service::storage_proxy&, redis::redis_opti
     if (_index < 0 || static_cast<size_t>(_index) >= options.get_total_redis_db_count()) {
         throw invalid_db_index_exception();
     }
-    options.set_keyspace_name(sprint("REDIS_%zu", static_cast<size_t>(_index)));
+    options.set_db_index(static_cast<unsigned>(_index));
```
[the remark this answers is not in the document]
fastio: Thank you @nyh. I'll fix it in a separate patch.

redis/keyspace_utils.cc
```diff
@@ -180,7 +180,7 @@ future<> create_keyspace_if_not_exists_impl(db::config& config, int default_repl
         return service::get_local_migration_manager().announce_new_column_family(schema, false);
     };
     // create 16 default database for redis.
-    return parallel_for_each(boost::irange<unsigned>(0, 16), [keyspace_gen = std::move(keyspace_gen), table_gen = std::move(table_gen)] (auto c) {
+    return parallel_for_each(boost::irange<unsigned>(0, config.redis_database_count()), [keyspace_gen = std::move(keyspace_gen), table_gen = std::move(table_gen)] (auto c) {
```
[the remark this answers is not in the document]
fastio: Go it.

redis/locks_context.cc
```diff
@@ -0,0 +1,26 @@
+/*
```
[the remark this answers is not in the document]
fastio: Okey, This file should be removed.

redis/server.cc
fastio: You are right. The last reply may be lost. As you said, to avoid that, we should not use the eof of connection as the `EOF of writer. I'll fix it. Thank you.

redis/server.cc
fastio: Got it.

redis/query_utils.cc
fastio: > As far as I understand now, what you're changing here doesn't just effect pipelining in a single request.
> Even if you have multiple requests from multiple connections, this "lock" serializes them so only one happens at a time. Is this intentional?

Sorry, I should write more detail about the serialization of processing requests. It's better to implement the serialization in a separate path. Multiple requests form multiple connections with same key should be processed in order. Obviersly the current implemention is not as expected. It's **not** avoid to update same key concurrently on different shareds.

> So given that this global - but per-shard - lock doesn't protect much besides pipelining, why do we make it global at all and not a per-request map, used just for pipelining?
> Is it because this way is easier? Or there's something actually better in this approach?

As mentioned above, the implemention of serialization of processing requests in this patch is totally wrong. Perhaps there is a better approach. Requests of same key processed by only one shard of the node should be guaranteed. And these Requests of same key should be processed in order. In fact, the requests should be forwarded to the target shard based on the hash value of the key associated. And the requests recieved by the target shard are processed
serially with same key(protected by lock), otherwise are precessed parallel.

So it's better than the one with global locks ?

redis/server.cc
fastio: Thanks, I got it.

redis-test/test_strings.py
fastio: Thanks for your suggestion. I'll fix it later.

penberg:
@fastio Are you still working on this actively? I'd like to move this forward so please let me know if you need help on this.

mykaul:
This was not touched in 3+ years, closing.

PR #5554, opened 2020-01-03
dist/ami: use prebuilt rpms on --localrpm

syuu1228:
We made --localrpm option to automatically build rpms from sourcecode,
but we actually use the option to produce AMI using prebuilt rpm on our
CI.
To simplified the script, and to prevent accsidently start rpm build
in the script, drop rpm build part.

### DIFF
--- a/dist/ami/build_ami.sh
+++ b/dist/ami/build_ami.sh
@@ -87,68 +87,32 @@ if [ $LOCALRPM -eq 1 ]; then
         fi
     done
     if [ -n "$rpm_missing" ]; then
-        reloc/build_reloc.sh
-        reloc/build_rpm.sh --dist --target centos7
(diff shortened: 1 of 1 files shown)


dist/ami/build_ami.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-./SCYLLA-VERSION-GEN
+./SCYLLA-VERSION-GEN > /dev/null 2>&1
```
avikivity: Why this change?

syuu1228: @avikivity I thought we don't need to print version number on the beginning of build_ami.sh, but the change is not related with PR so dropped the change.
Pushed new version: https://github.com/scylladb/scylla/pull/5554/files

PR #5561, opened 2020-01-07
dist/common/scripts/scylla_io_setup: raise error when unsupported instance type

syuu1228:
Currently --ami does not check instance types, creates invalid io_properties.yaml on unsupported instance types.
It actually won't occur on AMI startup, since scylla_ami_setup only invoke scylla_io_setup --ami when the instance is supported, so we don't get the issue on startup, but we still get when we run scylla_io_setup manually.

It's better to check instance type on scylla_io_setup, too.

Related #5438

### DIFF
--- a/dist/common/scripts/scylla_io_setup
+++ b/dist/common/scripts/scylla_io_setup
@@ -41,6 +41,9 @@ if __name__ == "__main__":
         if args.ami:
             idata = aws_instance()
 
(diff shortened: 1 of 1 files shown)


syuu1228:
@penberg @avikivity @glommer ping

pveentjer:
I just ran into this problem as well when using `c5d.2xlarge` as instance type.

PR #5564, opened 2020-01-08
install.sh: convert relocate_python_scripts.py to a bash function

syuu1228:
Since we need to run relocate_python_scripts.py on install time,
python script may not able to run on various different environment.
So convert the script to bash script, merge it into install.sh.


install.sh
```diff
@@ -103,6 +103,31 @@ EOF
     chmod +x "$root/$prefix/bin/$bin"
 }
 
+relocate_python3() {
+    local script="$1"
+    local scriptdir="$(dirname $script)"
```
avikivity: for extra safety, "$(dirname "$script")"

install.sh
```diff
@@ -103,6 +103,31 @@ EOF
...
+    local relocated="$scriptdir/libexec/$(basename $script)"
```
avikivity: Here too

install.sh
```diff
@@ -103,6 +103,31 @@ EOF
...
+    local pythonpath="$(dirname $(realpath -ms --relative-to $scriptdir $rpython3))"
```
avikivity: and here

install.sh
```diff
@@ -103,6 +103,31 @@ EOF
...
+
+    mv $script $relocated
```
avikivity: and everywhere

install.sh
```diff
@@ -270,20 +298,13 @@ EOS
     ln -srf "$retc"/systemd/system/node-exporter.service.d/nonroot.conf ~/.config/systemd/user/node-exporter.service.d
 fi
 
-install -m755 scylla-gdb.py -Dt "$rprefix"/scripts/
+install -m644 scylla-gdb.py -Dt "$rprefix"/scripts/
 
-PYSCRIPTS=$(find dist/common/scripts/ -maxdepth 1 -type f -exec grep -Pls '\A#!/usr/bin/env python3' {} +)
+PYSCRIPTS=$(find "$rprefix"/scripts/ -maxdepth 1 -type f -executable -exec grep -Pls '\A#!/usr/bin/env python3' {} +)
```
avikivity: These changes are good, but are unrelated

install.sh
syuu1228: fixed, force pushed.

install.sh
syuu1228: Dropped these changes for now, adjust relocate_python3 function to do same behavior with relocate_python_script.py

syuu1228:
@avikivity ping

PR #5567, opened 2020-01-08
CDC: do mutation augmentation from storage proxy

### LINKED ISSUE #5314: CDC: move mutation augmentation to storage_proxy
We should consider moving the augmentation of mutations into storage proxy. Currently it is hung up inside cql3::modification_statement. However, this both misses batched updates as well as things coming from alternator etc. 

My suggestion would be to introspect and optionally add to the mutation set in storage_proxy::mutate_prepare. Much like batch statement, this will have to deal with mutations potentially belonging to different schemas.

elcallio:
Fixes #5314

Instead of tying CDC handling into cql statement objects, this patch set moves it to storage proxy, i.e. shared code for mutating stuff. This means we automatically handle cdc for code paths outside cql (i.e. alternator). 

It also adds api handling (though initially inefficient) for batch statements. 

CDC is tied into storage proxy by giving the former a ref to the latter (per shard). Initially this is _not_ a constructor parameter, because right now we have chicken and egg issues here. Hopefully, Pavels refactoring of migration manager and notifications will untie these and this relationship can become nicer. 

The actual augmentation can (as stated above) be made much more efficient. Hopefully, the stream management refactoring will deal with expensive stream lookup, and eventually, we can maybe coalesce pre-image selects for batches. However, that is left as an exercise for when deemed needed. 

The augmentation API has an optional return value for a "post-image handler" to be used iff returned after mutation call is finished (and successful). It is not yet actually invoked from storage_proxy, but it is at least in the call chain. 

v2:
- Changed storage proxy batch mutation augment to run outside context object
- Added code in storage proxy to actually handle (currently unimplemented) post image hooks.

### DIFF
--- a/cdc/cdc.cc
+++ b/cdc/cdc.cc
@@ -72,6 +72,7 @@ static future<> populate_desc(db_context ctx, const schema& s);
 }
 
 class cdc::cdc_service::impl : service::migration_listener::empty_listener {
+    friend cdc_service;
     db_context _ctxt;
 public:
     impl(db_context ctxt)
@@ -194,7 +195,9 @@ cdc::cdc_service::cdc_service(service::storage_proxy& proxy)
 
 cdc::cdc_service::cdc_service(db_context ctxt)
     : _impl(std::make_unique<impl>(std::move(ctxt)))
-{}
+{
+    _impl->_ctxt._proxy.set_cdc_service(this);
+}
 
 cdc::cdc_service::~cdc_service() = default;
 
--- a/cdc/cdc.cc
+++ b/cdc/cdc.cc
@@ -178,6 +178,14 @@ public:
     }
 
     void on_drop_column_family(const sstring& ks_name, const sstring& cf_name) override {}
+
+    future<std::tuple<std::vector<mutation>, result_callback>> augment_mutation_call(
+        lowres_clock::time_point timeout,
+        std::vector<mutation>&& mutations
+    );
+
+    template<typename Iter>
+    future<> append_mutations(Iter i, Iter e, schema_ptr s, lowres_clock::time_point, std::vector<mutation>&);
 };
 
 cdc::cdc_service::cdc_service(service::storage_proxy& proxy)
@@ -635,9 +643,9 @@ public:
         auto command = ::make_lw_shared<query::read_command>(_schema->id(), _schema->version(), partition_slice, query::max_partitions);
 
         return _ctx._proxy.query(_schema, std::move(command), std::move(partition_ranges), cl, service::storage_proxy::coordinator_query_options(default_timeout(), empty_service_permit(), client_state)).then(
-                [this, partition_slice = std::move(partition_slice), selection = std::move(selection)] (service::storage_proxy::coordinator_query_result qr) -> lw_shared_ptr<cql3::untyped_result_set> {
+                [s = _schema, partition_slice = std::move(partition_slice), selection = std::move(selection)] (service::storage_proxy::coordinator_query_result qr) -> lw_shared_ptr<cql3::untyped_result_set> {
                     cql3::selection::result_set_builder builder(*selection, gc_clock::now(), cql_serialization_format::latest());
-                    query::result_view::consume(*qr.query_result, partition_slice, cql3::selection::result_set_builder::visitor(builder, *_schema, *selection));
+                    query::result_view::consume(*qr.query_result, partition_slice, cql3::selection::result_set_builder::visitor(builder, *s, *selection));
                     auto result_set = builder.build();
                     if (!result_set || result_set->empty()) {
                         return {};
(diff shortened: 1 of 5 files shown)


service/storage_proxy.cc
```diff
@@ -2153,6 +2161,12 @@ storage_proxy::mutate_atomically(std::vector<mutation> mutations, db::consistenc
         };
 
         future<> run() {
+            if (!std::exchange(_checked_cdc, true) && _p.get_cdc_service()->needs_cdc_augmentation(_mutations)) {
+                return _p.get_cdc_service()->augment_mutation_call(_timeout, std::move(_mutations)).then([this](std::tuple<std::vector<mutation>, cdc::result_callback>&& t) {
+                    _mutations = std::move(std::get<0>(t));
+                    return run();
+                });
+            }
```
gleb-cloudius: Why not call augment code before creating a context?

service/storage_proxy.hh
```diff
@@ -388,6 +393,9 @@ private:
     void maybe_update_view_backlog_of(gms::inet_address, std::optional<db::view::update_backlog>);
 
     db::view::update_backlog get_backlog_of(gms::inet_address) const;
+
+    template<typename Range>
+    future<> mutate_counters(Range&& mutations, db::consistency_level cl, tracing::trace_state_ptr tr_state, service_permit permit, clock_type::time_point timeout);
```
gleb-cloudius: Why is it moved here?

service/storage_proxy.cc
```diff
@@ -2005,6 +2006,11 @@ storage_proxy::get_paxos_participants(const sstring& ks_name, const dht::token &
  * @param tr_state trace state handle
  */
 future<> storage_proxy::mutate(std::vector<mutation> mutations, db::consistency_level cl, clock_type::time_point timeout, tracing::trace_state_ptr tr_state, service_permit permit, bool raw_counters) {
+    if (_cdc && _cdc->needs_cdc_augmentation(mutations)) {
+        return _cdc->augment_mutation_call(timeout, std::move(mutations)).then([this, cl, timeout, tr_state = std::move(tr_state), permit = std::move(permit), raw_counters](std::tuple<std::vector<mutation>, cdc::result_callback>&& t) mutable {
+            return _mutate_stage(this, std::move(std::get<0>(t)), cl, timeout, std::move(tr_state), std::move(permit), raw_counters);
+        });
+    }
```
gleb-cloudius: It looks a lot like what triggers infrastructure is doing in Cassandra, so may we should implement that and register cdc as a trigger.

service/storage_proxy.cc
elcallio: It was just easier to put the future-stuff inside the context. No other reason.

service/storage_proxy.hh
elcallio: Move it to private section, just to discourage any (right-now non-existant) direct calls to this, as it is a sub-path to regular (augmented) paths.

service/storage_proxy.cc
elcallio: Maybe. But I would suggest leaving that as an enhancement, also because some chicken-egg stuff might get resolved with pavels upcoming changes etc.

service/storage_proxy.cc
gleb-cloudius: But them you need to do tricks with a bool flag.

elcallio: This is true. Now I however remember. There is actually a much better reason why it is inside the context object: Post-image. 
The cdc API here gives optionally back a post-image-generation object that iff not empty need to be called after the mutation is done. This is not yet actually used, _but_ when writing so that it is, it will be easier if it exists as a result of the above call inside the context. Alternative is to make it a field in the context iff augmentation returned it.

gleb-cloudius: What do you mean by "after the mutation is done"?

elcallio: I mean after "run()" has successfully finished. When actual handling of post-image is implemented, the idea is that the code here would be something like:
```
auto post_handler = std::get<1>(t);
auto f = run();
if (handler) { 
  f = f.then(post_handler);
}
```
Now, arguably this could be handled just as well outside, so I guess it is not a super strong argument. I however still maintain that having the code outside the context object gets more messy, since it adds a conditional continuation _before_ creating the context object, which means at least some duplicated code.

gleb-cloudius: We are entering personal preferences territory here. IMO the code that do something, set flag to not do it next time and than runs itself recursively may look nice what you write it since it saves typing, but after a year, when you'll forget all about it, will be harder to understand. Code duplication may be less nice, but you did it in `storage_proxy::mutate` (just because you had no place to store a flag :))

elcallio: Ok. Updated to do the deed outside.

slivne:
@elcallio did you address all comments - if so please request gleb to re-review

elcallio:
@gleb-cloudius please re-review.

nyh:
Thanks. Committed to "next".

kbr-:
Please add me to reviewers next time you do CDC-related changes
@haaawk @elcallio

cdc/cdc.cc
```diff
@@ -747,33 +758,62 @@ transform_mutations(std::vector<mutation>& muts, decltype(muts.size()) batch_siz
...
+                if (!s->cdc_options().preimage()) {
+                    mutations.emplace_back(trans.transform(m));
                     return make_ready_future<>();
+                }
+
+                // Note: further improvement here would be to coalesce the pre-image selects into one
+                // iff a batch contains several modifications to the same table. Otoh, batch is rare(?)
+                // so this is premature.
+                auto f = trans.pre_image_select(qs.get_client_state(), db::consistency_level::LOCAL_QUORUM, m);
+                return f.then([trans = std::move(trans), &mutations, idx] (lw_shared_ptr<cql3::untyped_result_set> rs) mutable {
```
kbr-: How can we be sure that `pre_image_select` doesn't use internals of `trans`? If it so happens that it does, and we yield here, then this `move` into lambda might cause undefined behavior...

elcallio: We know because we wrote it to be. But in general I agree with your sentiment. Of course, I would much rather solve the whole issue by moving to coroutines and stop writing ass-backwards code that moves scope local objects around for every call, but...

PR #5581, opened 2020-01-09
Enable general centos RPM (not only centos7)

hagitsegev:
Have one centos RPM repo, both for CentOS7 and CentOS8


syuu1228:
I think it's time to drop "--target" and target variable on mustache template.
Please drop --target and $MUSTACHE_DIST from build_rpm.sh, and {{#centos7}}{{/centos7}} from scylla.spec.mustache.
"Requires: kernel >= 3.10.0-514" can be remain there, since it's valid for all distributions.

dist/redhat/build_rpm.sh
```diff
@@ -56,8 +56,8 @@ if [ ! -e SCYLLA-RELOCATABLE-FILE ]; then
     exit 1
 fi
 
-if [ -z "$TARGET" ] || [ "$TARGET" = "centos" ]; then
-    TARGET=centos7
+if [ -z "$TARGET" ] ; then
+    TARGET=centos
 fi
```
avikivity: Let's keep support for centos7 too so people's scripts don't break. We can just ignore TARGET completely.

dist/redhat/build_rpm.sh
```diff
@@ -119,6 +120,10 @@ ln -fv $RELOC_PKG $RPMBUILD/SOURCES/
 pystache dist/redhat/scylla.spec.mustache "{ \"version\": \"$SCYLLA_VERSION\", \"release\": \"$SCYLLA_RELEASE\", \"housekeeping\": $DIST, \"product\": \"$PRODUCT\", \"$PRODUCT\": true, \"reloc_pkg\": \"$RELOC_PKG_BASENAME\", $MUSTACHE_DIST }" > $RPMBUILD/SPECS/scylla.spec
 if [ "$TARGET" = "centos7" ]; then
     rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" --define "dist .el7" $RPMBUILD/SPECS/scylla.spec
+elif [ "$TARGET" = "centos8" ]; then
+    rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" --define "dist .el8" $RPMBUILD/SPECS/scylla.spec
+elif [ "$TARGET" = "centos" ]; then
+    rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" --define "dist .el" $RPMBUILD/SPECS/scylla.spec
 else
```
avikivity: Why do we need dist at all now? Why not define it to be empty?

dist/redhat/python3/build_rpm.sh
```diff
@@ -91,4 +91,4 @@ mkdir -p "$RPMBUILD"/{BUILD,BUILDROOT,RPMS,SOURCES,SPECS,SRPMS}
 
 ln -fv "$RELOC_PKG" "$RPMBUILD"/SOURCES/
 pystache dist/redhat/python3/python.spec.mustache "{ \"version\": \"${SCYLLA_VERSION}\", \"release\": \"${SCYLLA_RELEASE}\", \"reloc_pkg\": \"${RELOC_PKG_BASENAME}\", \"name\": \"$PRODUCT-python3\", \"target\": \"/opt/scylladb/python3\" }" > "$RPMBUILD"/SPECS/python.spec
-rpmbuild --nodebuginfo -ba --define '_binary_payload w2.xzdio' --define "_build_id_links none" --define "_topdir ${RPMBUILD}" --define "dist .el7" "$RPMBUILD"/SPECS/python.spec
+rpmbuild --nodebuginfo -ba --define '_binary_payload w2.xzdio' --define "_build_id_links none" --define "_topdir ${RPMBUILD}" --define "dist .el" "$RPMBUILD"/SPECS/python.spec
```
avikivity: Here too.

dist/redhat/build_rpm.sh
```diff
@@ -123,7 +123,7 @@ if [ "$TARGET" = "centos7" ]; then
 elif [ "$TARGET" = "centos8" ]; then
     rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" --define "dist .el8" $RPMBUILD/SPECS/scylla.spec
 elif [ "$TARGET" = "centos" ]; then
-    rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" --define "dist .el" $RPMBUILD/SPECS/scylla.spec
+    rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" $RPMBUILD/SPECS/scylla.spec
```
avikivity: Won't we inherit dist for /usr/lib/rpm/macros.d/macros.dist?

dist/redhat/build_rpm.sh
```diff
@@ -68,8 +68,6 @@ if [[ ! $TARGET =~ fedora.*|centos.* ]]; then
     echo "unknown distribution specified."
     echo "can use the following targets:"
     echo "  centos"
-    echo "  centos7"
-    echo "  centos8"
```
avikivity: Why did you add centos8 if you now remove it?

dist/redhat/build_rpm.sh
```diff
@@ -118,12 +116,4 @@ rpm_payload_opts=(--define "_binary_payload w2${xz_thread_param}.xzdio")
...
-if [ "$TARGET" = "centos7" ]; then
-    rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" --define "dist .el7" $RPMBUILD/SPECS/scylla.spec
-elif [ "$TARGET" = "centos8" ]; then
-    rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" --define "dist .el8" $RPMBUILD/SPECS/scylla.spec
-elif [ "$TARGET" = "centos" ]; then
-    rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" $RPMBUILD/SPECS/scylla.spec
-else
-    rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" $RPMBUILD/SPECS/scylla.spec
-fi
+rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" $RPMBUILD/SPECS/scylla.spec
```
avikivity: Why did you add all those lines if you now remove them?

avikivity:
> I think it's time to drop "--target" and target variable on mustache template.

Let's keep --target, just for compatibility. We can ignore the value passed.

> Please drop --target and $MUSTACHE_DIST from build_rpm.sh, and {{#centos7}}{{/centos7}} from scylla.spec.mustache.
> "Requires: kernel >= 3.10.0-514" can be remain there, since it's valid for all distributions.

dist/redhat/build_rpm.sh
hagitsegev: It is an intermediate step (to have a general el instead of el7). I will squash everything when we are satisfied with the solution.

slivne:
@avikivity please re-review

PR #5599, opened 2020-01-15
dist/redhat: install specified version of scylla-enterprise-conf on meta package

syuu1228:
To install specified version of scylla-conf package, we need to add it on Requires.

See scylladb/scylla-enterprise#1203

### DIFF
--- a/dist/redhat/scylla.spec.mustache
+++ b/dist/redhat/scylla.spec.mustache
@@ -7,7 +7,7 @@ Group:          Applications/Databases
 License:        AGPLv3
 URL:            http://www.scylladb.com/
 Source0:        {{reloc_pkg}}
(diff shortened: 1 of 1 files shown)


syuu1228:
ping

dist/redhat/scylla.spec.mustache
```diff
@@ -7,7 +7,7 @@ Group:          Applications/Databases
 License:        AGPLv3
```
avikivity: Please file an issue in the scylla.git repo and reference it instead of the enterprise issue.

syuu1228: Created issue #5639, pushed new commit message https://github.com/scylladb/scylla/pull/5599/commits/e689183e54775e671950c750341bf4b60915e87f

avikivity:
Queued as 83092393701ce41c7e0829c3ef9b0ac6f5374472

avikivity:
Dequeued due to conflicts with the dequeing of #5605 , please respin.

syuu1228:
@avikivity rebased with latest next.

PR #5601, opened 2020-01-16
Update HACKING.md

0xflotus:
fixed typo in --overprovisioned flag

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -191,7 +191,7 @@ The `scylla.yaml` file in the repository by default writes all database data to 
 
 Scylla has a number of requirements for the file-system and operating system to operate ideally and at peak performance. However, during development, these requirements can be relaxed with the `--developer-mode` flag.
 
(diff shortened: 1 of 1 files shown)


bhalevy:
LGTM, just need to change target branch to `next` (we don't merge directly to master)

0xflotus:
> LGTM, just need to change target branch to `next` (we don't merge directly to master)

I opened a new PR #5618 :)

nyh:
@bhalevy it doesn't matter what the target branch is - the committer will take care of it. In fact, it's better to set the target branch to "master"-  that ensures that the pull request is only closed when the patch actually reaches master - and not as soon as it reaches "next".

glommer:
@nyh is this really true? I see a lot of people including @penberg and @avikivity repeatedly asking people to base on top of next

avikivity:
I only ask to base on next under exceptional circumstances. The correct base is master.

bhalevy:
The base should be master and the target branch is next.
This is particularly true is one merges using the github web GUI.
This also helps detect merge conflicts in next before merging.

PR #5604, opened 2020-01-16
Simplify sum overflow

bhalevy:
As a followup to 0bde5906b3337be7691a8fd62d3dfe98e07437e4
This series implements suggestions from @avikivity and @espindola 
It simplifies the template definitions for `accumulator_for`,
adds some debug logging for the overflow values,
and adds unit tests for float and double sum overflow.

Test: unit(dev),
paging_test:TestPagingWithIndexingAndAggregation.test_filter_{indexed,non_indexed,pk}_column(dev)

### DIFF
--- a/cql3/functions/aggregate_fcts.cc
+++ b/cql3/functions/aggregate_fcts.cc
@@ -81,86 +81,33 @@ public:
 template <typename T>
 struct accumulator_for;
 
-template <typename NarrowType, typename AccType>
-static NarrowType checking_narrow(AccType acc) {
-    NarrowType ret = static_cast<NarrowType>(acc);
-    if (static_cast<AccType>(ret) != acc) {
-        throw exceptions::overflow_error_exception("Sum overflow. Values should be casted to a wider type.");
-    }
-    return ret;
-}
-
-template <>
-struct accumulator_for<int8_t> {
+template <typename T>
+struct int128_accumulator_for {
     using type = __int128;
 
-    static int8_t narrow(type acc) {
-        return checking_narrow<int8_t>(acc);
+    static T narrow(type acc) {
+        T ret = static_cast<T>(acc);
+        if (static_cast<type>(ret) != acc) {
+            throw exceptions::overflow_error_exception("Sum overflow. Values should be casted to a wider type.");
+        }
+        return ret;
     }
 };
 
-template <>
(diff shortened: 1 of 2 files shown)


espindola:
Not sure I understand the value in logging on overflow given that we
return an error.

Other than that, just one nit:


+            .with_row({float_type->decompose(3.4028234e+38f)});

Can you use serilized instead of decompose in here and on the other uses?

LGTM otherwise.

Thanks,
Rafael

bhalevy:
> Not sure I understand the value in logging on overflow given that we return an error.

It is helpful to see the actual numbers in the debug log to figure out the overflowed value.

> Other than that, just one nit: + .with_row({float_type->decompose(3.4028234e+38f)}); Can you use serilized instead of decompose in here and on the other uses? LGTM otherwise. Thanks, Rafael

Yes, this is doable.

bhalevy:
In v2:
- use `serialized` instead of `decompose`
- rebased, retested

cql3/functions/aggregate_fcts.cc
```diff
@@ -88,6 +88,10 @@ struct int128_accumulator_for {
     static T narrow(type acc) {
         T ret = static_cast<T>(acc);
         if (static_cast<type>(ret) != acc) {
+            cflogger.debug(format("SUM overflow: 0x{:08x}_{:08x}_{:08x}_{:08x} -> {}",
+                    static_cast<uint32_t>(acc >> 96), static_cast<uint32_t>(acc >> 64),
+                    static_cast<uint32_t>(acc >> 32), static_cast<uint32_t>(acc),
+                    ret).c_str());
             throw exceptions::overflow_error_exception("Sum overflow. Values should be casted to a wider type.");
```
avikivity: This format() call now happens even debug level is disabled in the log (the common case). While it should be rare, it's still not good practice to perform a heavyweight computation that is then discarded. 

It's also fine to drop this completely. If a user gets an exception, it will be easier for them to repeat the query without the sum (and perform the sum by hand) than guess which logger to enable. Logs are more for operators of the database rather than users.

bhalevy: np, will remove in v3

bhalevy:
In v3:
- removed debug logging

espindola:
LGTM

PR #5605, opened 2020-01-16
dist/redhat: override systemd macros to CentOS7 version

syuu1228:
Fedora 31 version of systemd macros does not work correctly on CentOS7,
since CentOS7 does not support "file trigger" feature.
To fix the issue we need to override macros with CentOS7 version.

See scylladb/scylla-jmx#94

### DIFF
--- a/dist/redhat/build_rpm.sh
+++ b/dist/redhat/build_rpm.sh
@@ -93,5 +93,6 @@ fi
 rpm_payload_opts=(--define "_binary_payload w2${xz_thread_param}.xzdio")
 
 ln -fv $RELOC_PKG $RPMBUILD/SOURCES/
+ln -fv dist/redhat/systemd.inc $RPMBUILD/SOURCES/
 pystache dist/redhat/scylla.spec.mustache "{ \"version\": \"$SCYLLA_VERSION\", \"release\": \"$SCYLLA_RELEASE\", \"housekeeping\": $DIST, \"product\": \"$PRODUCT\", \"$PRODUCT\": true, \"reloc_pkg\": \"$RELOC_PKG_BASENAME\" }" > $RPMBUILD/SPECS/scylla.spec
 rpmbuild -ba "${rpm_payload_opts[@]}" --define "_topdir $RPMBUILD" $RPMBUILD/SPECS/scylla.spec
--- a/dist/redhat/scylla.spec.mustache
+++ b/dist/redhat/scylla.spec.mustache
@@ -7,8 +7,11 @@ Group:          Applications/Databases
 License:        AGPLv3
 URL:            http://www.scylladb.com/
 Source0:        {{reloc_pkg}}
+Source1:        systemd.inc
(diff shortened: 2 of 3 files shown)


dist/redhat/centos7-systemd.inc
```diff
@@ -0,0 +1,21 @@
+%undefine systemd_post
+%define systemd_post() \
```
avikivity: I think this comes from the original rpm files. We can't just copy without providing copyright and license information.

Since rpm is GPL/LGPL we can't use it directly, so we have to reimplement it without copying (it's okay to look at the source for inspiration).

syuu1228: You are right, I just pushed re-implemented version. how about this? https://github.com/scylladb/scylla/pull/5605/files#diff-604df91d4cd4c6756e5c01e0e25fd324

syuu1228:
@avikivity ping

avikivity:
Dequeued, I see this in next:

```
12:35:42  Running './dist/redhat/build_rpm.sh  --reloc-pkg /jenkins/workspace/scylla-master/next/scylla/build/release/scylla-package.tar.gz' under build/redhat/scylla-package directory
12:35:42  '/jenkins/workspace/scylla-master/next/scylla/build/redhat/SOURCES/scylla-package.tar.gz' => '/jenkins/workspace/scylla-master/next/scylla/build/release/scylla-package.tar.gz'
12:35:42  '/jenkins/workspace/scylla-master/next/scylla/build/redhat/SOURCES/systemd.inc' => 'dist/redhat/systemd.inc'
12:35:42  error: Unable to open %{SOURCE1}: No such file or directory
```

avikivity:
@syuu1228 please fix and respin

syuu1228:
@avikivity I checked Jenkins build job [0] which where we got the error, and I found that dist/redhat/scylla.spec.mustache is not same with the commit.
Even we can see "Source1: systemd.inc" on the commit [1], I'm not able to find same line on the artifact of build 1552 [2].
Same in the relocatable package of the build [3].
And that's the reason why we getting "error: Unable to open %{SOURCE1}: No such file or directory" on the build job.
I guess we break something on the build job but I don't find what is it..

0. https://jenkins.scylladb.com/view/master/job/scylla-master/job/next/1552/
1. https://github.com/scylladb/scylla/commit/5cf3d9b94979943ef622e5b1288eea3fb4b87b41#diff-69962b3f7242b4af70aeba79ceeb3810R10
2. https://jenkins.scylladb.com/view/master/job/scylla-master/job/next/1552/artifact/scylla/dist/redhat/scylla.spec.mustache/*view*/
3. https://jenkins.scylladb.com/view/master/job/scylla-master/job/next/1552/artifact/scylla/build/release/scylla-package.tar.gz

syuu1228:
@avikivity ping, can we try to merge next branch again? I think PR itself does not have the problem, we need an another build result to find where the issue comes from.
(build 1552 seems already deleted, not able to see the workplace and build log anymore)

PR #5620, opened 2020-01-21
Remove deprecated partitioners

haaawk:
This PR makes named_value respect allowed_values and then use it to transition away from old deprecated RandomPartitioner and ByteOrderedPartitioner. Then it removes the code that's no longer used.

We want to remove deprecated partitioners because, on one hand, they lead to performance problems and hot nodes. Moreover, we're planning to unify the token representation which would allow per table partitioner support. That, in turn, is a feature helpful in multiple efforts like CDC, materialized views, secondary indexes and multi-tenancy. See https://docs.google.com/document/d/16ZRC6Sh3_V_M6z2FTIjyyaHJ-mOlcaLK7MnmFq8Iu48/edit#heading=h.vhrw7dalxq3k for more information. Old partitioners block that effort.

tests: unit(dev)


avikivity:
The cover letter has to make some effort to explain why removing the old partitioners is a good idea.

haaawk:
I added more explanation @avikivity. Is that enough or should I add more?

haaawk:
I rebased on top of master and started BYO here: https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/892/

tzach:
@haaawk can you please explain the UX result of this change?
Are you suggesting to remove the partitioners option, or just warn the partitioners are deprecated?

Since there is no easy way to move from one partitioners to the other, we should give users enough time and tooling to make the move.

haaawk:
I'm suggesting removal of the option. The warning about deprecation was introduced 2.5 year ago -> https://github.com/scylladb/scylla/commit/de011ece5237f294da75fadf454d3b00c92266f2

We already have the tooling for migration. It's called Spark Migrator.

Don't you think @tzach that 2.5 year was enough to migrate away from deprecated partitioners? If we commit this now it will still take months before it reaches users. I will send an email to the user list today to announce that we're planning to remove support for deprecated partitioners.

avikivity:
The option should remain, just the values for random and byteorder should now be unconditionally rejected. Otherwise people's scylla.yaml files can become invalid.

haaawk:
I didn't remove the enable_deprecated_partitioners flag - just deprecated partitioners @avikivity

avikivity:
@haaawk ok, I understood different from your previous comment.

tzach:
> I'm suggesting removal of the option. The warning about deprecation was introduced 2.5 year ago -> de011ec

Unfortunately, assuming users are aware of this warning is optimistic.
I'm all for the change, but we need to actively, explicitly publish it.

haaawk:
> @haaawk ok, I understood different from your previous comment.

I was thinking about option of setting random or byte order partitioners. Sorry for being unclear.

haaawk:
> > I'm suggesting removal of the option. The warning about deprecation was introduced 2.5 year ago -> [de011ec](https://github.com/scylladb/scylla/commit/de011ece5237f294da75fadf454d3b00c92266f2)
> 
> Unfortunately, assuming users are aware of this warning is optimistic.
> I'm all for the change, but we need to actively, explicitly publish it.

I understand - let's communicate on all the channels then. The sooner the better.

avikivity:
The patches look fine but come with no explanations.

haaawk:
> The patches look fine but come with no explanations.

You mean you would like longer commit messages in every the commit?

avikivity:
@tzach it's more than a warning. If you try to use a deprecated partitioner, scylla will refuse to boot until you set the enable_deprecated_partitioners option. The message you get is

```c++
                    startlog.error("The partitioner {} is deprecated and will be removed in a future version."
                            "  To enable it, add \"enable_deprecated_partitioners: true\" to scylla.yaml"
                            "  Contact scylladb-users@googlegroups.com if you are using it in production", cfg->partitioner());
```

You can't possibly miss it, since if you didn't read it you don't know which config option is needed.

avikivity:
@haaawk yes, someone reading the commit should understand why you are doing something (not just what)

avikivity:
you can also merge the "make it impossible to use" and "remove" commits, but I'm okay with their remaining separate.

haaawk:
I will add more explanation.

haaawk:
I added more commit messages. @avikivity please check whether it's enough.

avikivity:
Queued as a80d6d7947a2a1037c411102bf89d5ccbf6a6f28

haaawk:
Thanks @avikivity - btw Parallel Tests and dtest has passed on BYO

PR #5623, opened 2020-01-21
SCYLLA-VERSION-GEN: skip updating version files when git hash unchanged

syuu1228:
On our build system we tries to build relocatable package multiple times on
same revision of the repository, it executes ./SCYLLA-VERSION-GEN for each time.
When the build job invoked at midnight and it did not finished until 12:00AM,
first build and last build has different SCYLLA-RELEASE-FILE, since it contains
current date.
--disable-update-version is the option to hold current SCYLLA-RELEASE-FILE
and prevent update, to build packages with consistent revision number.

Fixes scylladb/scylla-pkg#826

### DIFF
--- a/SCYLLA-VERSION-GEN
+++ b/SCYLLA-VERSION-GEN
@@ -19,6 +19,14 @@ else
 	SCYLLA_RELEASE=$SCYLLA_BUILD.$DATE.$GIT_COMMIT
 fi
 
+if [ -f build/SCYLLA-RELEASE-FILE ]; then
+	RELEASE_FILE=$(cat build/SCYLLA-RELEASE-FILE)
+	GIT_COMMIT_FILE=$(cat build/SCYLLA-RELEASE-FILE |cut -d . -f 3)
+	if [ "$GIT_COMMIT" = "$GIT_COMMIT_FILE" ]; then
+		exit 0
+	fi
+fi
+
 echo "$SCYLLA_VERSION-$SCYLLA_RELEASE"
 mkdir -p build
 echo "$SCYLLA_VERSION" > build/SCYLLA-VERSION-FILE


reloc/build_reloc.sh
```diff
@@ -37,6 +37,7 @@ print_usage() {
     echo "  --clean                 clean build directory"
     echo "  --compiler PATH         C++ compiler path"
     echo "  --c-compiler PATH       C compiler path"
+    echo "  --disable-update-version disable updating SCYLLA-VERSION-FILE and SCYLLA-RELEASE-FILE"
```
hagitsegev: Maybe add here reason? Why should we avoid updating them?
... in order to avoid different revision number on build packages if compiled around midnight.

configure.py
```diff
@@ -1117,10 +1119,11 @@ def setup_first_pkg_of_list(pkglist):
 build_modes =  {m: modes[m] for m in selected_modes}
 build_artifacts = all_artifacts if not args.artifacts else args.artifacts
 
-status = subprocess.call("./SCYLLA-VERSION-GEN")
-if status != 0:
-    print('Version file generation failed')
-    sys.exit(1)
+if not args.update_version:
+    status = subprocess.call("./SCYLLA-VERSION-GEN")
```
hagitsegev: Maybe it is a starter's question.
But, How does the process knows when to create the files (first mode it compiles) and when not to create (not first mode) - files should exist?

hagitsegev:
So when calling build_reloc.sh during build, I need to call it once as is, for the first mode I compile, and after that with the new flag.

configure.py
syuu1228: On first iteration jenkins job should call ./reloc/buid_reloc.sh, then second iteration the job should call ./reloc/build_reloc.sh --disable-update-version.
With the option, build script uses already generated SCYLLA*FILE, we can avoid date difference isue.
Note that running build_reloc.sh --disable-update-version without SCYLLA*FILE, build script will causes file not found error.

reloc/build_reloc.sh
syuu1228: I'll update the description.

configure.py
hagitsegev: Maybe it is better to:
If file exist, use it.
If not - create it.
So you don't need a parameter at all.
On jenkins jobs, we clean the workspace each time.
On developers machines - this is not an issue at all.

syuu1228: @hagitsegev Partially agree, but I suppose we have to update version files when git hash changed (ex: user may run git checkout <branch>)
Otherwise it will be inconsistent, we cannot guarantee SCYLLA-RELEASE-FILE contains correct git hash anymore.

So I think it should be something like this:
```
if exist:
    if hash matches:
        use it
    else:
        update it
else:
    create it
```

syuu1228: @hagitsegev Pushed new version based on the idea described above https://github.com/scylladb/scylla/pull/5623/files

PR #5633, opened 2020-01-22
view_info: Drop partition_ranges()

dekimir:
The method view_info::partition_ranges() is unused.

Tests: unit (dev)

### DIFF
--- a/db/view/view.cc
+++ b/db/view/view.cc
@@ -118,13 +118,6 @@ const query::partition_slice& view_info::partition_slice() const {
     return *_partition_slice;
 }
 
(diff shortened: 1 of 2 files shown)


nyh:
If I understand correctly, this method uses the member "_partition_ranges" to cache its results. Now that you remove this method, the member isn't needed either, right? If so, can you please remove it too? Thanks.

dekimir:
> If I understand correctly, this method uses the member "_partition_ranges" to cache its results. Now that you remove this method, the member isn't needed either, right? If so, can you please remove it too? Thanks.

Done.

PR #5636, opened 2020-01-23
Remove deprecated partitioners

haaawk:
This PR makes named_value respect allowed_values and then use it to transition away from old deprecated RandomPartitioner and ByteOrderedPartitioner. Then it removes the code that's no longer used.

We want to remove deprecated partitioners because, on one hand, they lead to performance problems and hot nodes. Moreover, we're planning to unify the token representation which would allow per table partitioner support. That, in turn, is a feature helpful in multiple efforts like CDC, materialized views, secondary indexes and multi-tenancy. See https://docs.google.com/document/d/16ZRC6Sh3_V_M6z2FTIjyyaHJ-mOlcaLK7MnmFq8Iu48/edit#heading=h.vhrw7dalxq3k for more information. Old partitioners block that effort.

tests: unit(dev)


haaawk:
I've checked locally that all the dtests which failed previously are now passing.

haaawk:
BYO runs here: https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/894/

haaawk:
I found additional breakage. Will update the PR.

haaawk:
I think I've fixed everything. I've run whole auth_test.py and auth_roles_test.py locally and they passed. BYO is here: https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/896/

haaawk:
I got this from jenkins:
`
Fixed SUCCESS: Build Job 'scylla-master/byo/byo_build_tests_dtest [896] on branch/release Personal tests build: Check console output at https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/896/ Please note that tests and dtest build log files are available as build artifacts. Search for *_output.txt
`

so I think we're good to go @avikivity :)

avikivity:
What was broken and what was the fix?

haaawk:
@avikivity allowed values for authorizer and authenticator were wrong. They were missing some values like com.scylladb.auth.TransitionalAuthenticator. It was working before because allowed values were not enforced. Now that I fixed the support for allowed values the wrong list started to fail some dtests. I solved the problem by fixing the lists of allowed values for those flags.

haaawk:
Anything else you would like to see here @avikivity to merge this?

avikivity:
@haaawk no, I was delaying it because next is clogged (due to unrelated reasons). Should have informed you, sorry.

haaawk:
Is next still broken @avikivity ?

avikivity:
Yes

AustEcon:
I really need this feature. How can I use `ByteOrderedPartitioner` with ScyllaDB? Is there an older version of the docker image I could use that would support it?

tzach:
@AustEcon, maybe we can advise on migration from an old cluster to a new one. It's probably more of a [Scylla Forum](https://forum.scylladb.com/) discussion

PR #5655, opened 2020-01-27
config: Remove UDF from experimental_features_t in 3.2

### LINKED ISSUE #5645: Version 3.2 allows `udf` as a valid value of experimental_features

*Installation details*
Scylla version (or git commit hash): 3.2
Cluster size: n/a
OS (RHEL/CentOS/Ubuntu/AWS AMI): n/a

Scylla 3.2 doesn't include the UDF feature, but its configuration code allows `--experimental-features=udf`.  We should remove UDF from `experimental_features_t`, the description of `--experimental-features` flag, and `conf/scylla.yaml`.

dekimir:
Scylla 3.2 doesn't support UDF, so do not accept UDF as a valid option
to experimental_features.

Fixes #5645.

No fix is needed on master, which does support UDF.

Tests: unit (dev)

### DIFF
--- a/conf/scylla.yaml
+++ b/conf/scylla.yaml
@@ -244,7 +244,6 @@ batch_size_fail_threshold_in_kb: 50
 # experimental_features:
 #     - cdc
 #     - lwt
-#     - udf
 
 # The directory where hints files are stored if hinted handoff is enabled.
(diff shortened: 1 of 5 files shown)


avikivity:
Quite a large patch to remove a choice!

PR #5668, opened 2020-01-28
hinted handoff: send orphaned hints on HINT_MUTATION verb

piodul:
When replaying a hint with a destination node that is no longer in the
cluster, it will be sent with cl=ALL to all its new replicas. Before
this patch, the MUTATION verb was used, which causes such hints to be
handled on the same connection and with the same priority as regular
writes. This can cause problems when a large number of hints is
orphaned and they are scheduled to be sent at once. Such situation
may happen when replacing a dead node - all nodes that accumulated hints
for the dead node will now send them with cl=ALL to their new replicas.

This patch changes the verb used to send such hints to HINT_MUTATION.
This verb is handled on a separate connection and with streaming
scheduling group, which gives them similar priority to non-orphaned
hints.

Refs: #4712

Tests: unit(dev)

### DIFF
--- a/database.cc
+++ b/database.cc
@@ -1584,6 +1584,15 @@ future<> database::apply(schema_ptr s, const frozen_mutation& m, db::commitlog::
     return update_write_metrics(_apply_stage(this, std::move(s), seastar::cref(m), timeout, sync));
 }
 
+future<> database::apply_hint(schema_ptr s, const frozen_mutation& m, db::timeout_clock::time_point timeout) {
+    if (dblog.is_enabled(logging::log_level::trace)) {
+        dblog.trace("apply hint {}", m.pretty_printer(s));
+    }
+    return with_scheduling_group(_dbcfg.streaming_scheduling_group, [this, s = std::move(s), &m, timeout] () mutable {
+        return update_write_metrics(_apply_stage(this, std::move(s), seastar::cref(m), timeout, db::commitlog::force_sync::no));
+    });
+}
+
 future<> database::apply_streaming_mutation(schema_ptr s, utils::UUID plan_id, const frozen_mutation& m, bool fragmented) {
     if (!s->is_synced()) {
         throw std::runtime_error(format("attempted to mutate using not synced schema of {}.{}, version={}",
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -282,7 +282,9 @@ public:
     }
     virtual future<> apply_locally(storage_proxy& sp, storage_proxy::clock_type::time_point timeout,
             tracing::trace_state_ptr tr_state) override {
-        return make_exception_future<>(std::runtime_error("Executing hint locally doesn't make sense"));
+        // A hint will be sent to all relevant endpoints when the endpoint it was originally intended for
+        // becomes unavailable - this might include the current node
+        return sp.mutate_hint(_schema, *_mutation, timeout);
     }
     virtual future<> apply_remotely(storage_proxy& sp, gms::inet_address ep, std::vector<gms::inet_address>&& forward,
             storage_proxy::response_id_type response_id, storage_proxy::clock_type::time_point timeout,
@@ -1711,6 +1713,15 @@ storage_proxy::mutate_locally(std::vector<mutation> mutations, clock_type::time_
 }
 
 future<>
+storage_proxy::mutate_hint(const schema_ptr& s, const frozen_mutation& m, clock_type::time_point timeout) {
+    auto shard = _db.local().shard_of(m);
+    get_stats().replica_cross_shard_ops += shard != engine().cpu_id();
+    return _db.invoke_on(shard, _write_smp_service_group, [&m, gs = global_schema_ptr(s), timeout] (database& db) -> future<> {
+        return db.apply_hint(gs, m, timeout);
+    });
+}
+
+future<>
 storage_proxy::mutate_counters_on_leader(std::vector<frozen_mutation_and_schema> mutations, db::consistency_level cl, clock_type::time_point timeout,
                                          tracing::trace_state_ptr trace_state, service_permit permit) {
     get_stats().received_counter_updates += mutations.size();
@@ -1815,6 +1826,12 @@ storage_proxy::create_write_response_handler_helper(schema_ptr s, const dht::tok
 storage_proxy::response_id_type
 storage_proxy::create_write_response_handler(const mutation& m, db::consistency_level cl, db::write_type type, tracing::trace_state_ptr tr_state, service_permit permit) {
     return create_write_response_handler_helper(m.schema(), m.token(), std::make_unique<shared_mutation>(m), cl, type, tr_state,
+            std::move(permit));
+}
+
+storage_proxy::response_id_type
+storage_proxy::create_write_response_handler(const hint_wrapper& h, db::consistency_level cl, db::write_type type, tracing::trace_state_ptr tr_state, service_permit permit) {
+    return create_write_response_handler_helper(h.mut.schema(), h.mut.token(), std::make_unique<hint_mutation>(h.mut), cl, type, tr_state,
             std::move(permit));
 }
 
@@ -2415,6 +2432,16 @@ future<> storage_proxy::send_hint_to_endpoint(frozen_mutation_and_schema fm_a_s,
             db::write_type::SIMPLE,
             get_stats(),
             allow_hints::no);
+}
+
+future<> storage_proxy::mutate_hint_from_scratch(frozen_mutation_and_schema fm_a_s) {
+    const auto timeout = db::timeout_clock::now() + 1h;
+    if (!_features.cluster_supports_hinted_handoff_separate_connection()) {
+        return mutate({fm_a_s.fm.unfreeze(fm_a_s.s)}, db::consistency_level::ALL, timeout, nullptr, empty_service_permit());
+    }
+
+    std::array<hint_wrapper, 1> ms{hint_wrapper { std::move(fm_a_s.fm.unfreeze(fm_a_s.s)) }};
+    return mutate_internal(std::move(ms), db::consistency_level::ALL, false, nullptr, empty_service_permit(), timeout);
 }
 
 /**
--- a/database.hh
+++ b/database.hh
@@ -1478,6 +1478,7 @@ public:
     // Apply the mutation atomically.
     // Throws timed_out_error when timeout is reached.
     future<> apply(schema_ptr, const frozen_mutation&, db::commitlog::force_sync sync, db::timeout_clock::time_point timeout = db::no_timeout);
+    future<> apply_hint(schema_ptr, const frozen_mutation&, db::timeout_clock::time_point timeout = db::no_timeout);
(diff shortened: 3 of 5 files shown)


service/storage_proxy.cc
```diff
@@ -273,7 +273,9 @@ class hint_mutation : public shared_mutation {
     }
     virtual future<> apply_locally(storage_proxy& sp, storage_proxy::clock_type::time_point timeout,
             tracing::trace_state_ptr tr_state) override {
-        return make_exception_future<>(std::runtime_error("Executing hint locally doesn't make sense"));
+        // A hint will be sent to all relevant endpoints when the endpoint it was originally intended for
+        // becomes unavailable - this might include the current node
+        return sp.mutate_hint(_schema, *_mutation, timeout);
```
vladzcloudius: Nice catch!
But I think this deserves to be in a separate patch.

database.cc
```diff
@@ -1534,6 +1534,15 @@ future<> database::apply(schema_ptr s, const frozen_mutation& m, db::commitlog::
     return update_write_metrics(_apply_stage(this, std::move(s), seastar::cref(m), timeout, sync));
 }
 
+future<> database::apply_hint(schema_ptr s, const frozen_mutation& m, db::timeout_clock::time_point timeout) {
+    if (dblog.is_enabled(logging::log_level::trace)) {
+        dblog.trace("apply hint {}", m.pretty_printer(s));
+    }
+    return with_scheduling_group(_dbcfg.streaming_scheduling_group, [this, s = std::move(s), &m, timeout] () {
```
vladzcloudius: If you don't declare a lambda `mutable` std::move(X) on X from the capture list is going to be a copy because X is going to be const. In your case X is an `s`  (schema_ptr).

piodul: Oh, now I see why it was needed in other places. Thanks, fixed.

service/storage_proxy.cc
piodul: Previously, `hint_mutation` would only be used when hint's destination node is in the cluster (namely, `storage_proxy::send_hint_to_endpoint`). In case destination node was no longer the cluster, the path that sent the hint to its new replicas used `shared_mutation`.

This change is not really a bugfix, because it makes `hint_mutation` implement a method that was not used before, but required by `mutation_holder` interface. I did not hit the "Executing hint locally doesn't make sense" exception when I was testing a node replacement scenario without this patch.

service/storage_proxy.cc
```diff
@@ -2319,6 +2336,31 @@ future<> storage_proxy::send_hint_to_endpoint(frozen_mutation_and_schema fm_a_s,
             allow_hints::no);
 }
 
+future<> storage_proxy::mutate_hint_from_scratch(frozen_mutation_and_schema fm_a_s) {
+    // FIXME: using 1h as infinite timeout. If a node is down, we should get an
+    // unavailable exception.
```
gleb-cloudius: A node may go down after we checked for its liveness

service/storage_proxy.cc
```diff
@@ -2319,6 +2336,31 @@ future<> storage_proxy::send_hint_to_endpoint(frozen_mutation_and_schema fm_a_s,
...
+    return mutate_prepare(std::move(ms), db::consistency_level::ALL, db::write_type::SIMPLE, empty_service_permit(),
+            [this] (const mutation& m,
+                db::consistency_level cl,
+                db::write_type type, service_permit permit) {
+        return create_hint_from_scratch_write_response_handler(m);
+    }).then([this, timeout] (std::vector<storage_proxy::unique_response_handler> ids) {
+        return mutate_begin(std::move(ids), db::consistency_level::ALL, timeout);
+    }).then_wrapped([this, p = shared_from_this(), lc] (future<> f) mutable {
+        return p->mutate_end(std::move(f), lc, _stats, nullptr);
+    });
```
gleb-cloudius: You basically copy paste `mutate_internal()` here. The way to extend `mutate_internal()` is to add an override for `storage_proxy::response_id_type storage_proxy::create_write_response_handler` that creates the kind of mutating_holder you want. Than you can use `mutate_internal()`  directly.

service/storage_proxy.cc
piodul: @vladzcloudius , the timeout was already there, but I'm not sure why it was set to a such high value. Do you remember why it was set like that?

Should we change it to the default write timeout (`write_request_timeout_in_ms`)?

service/storage_proxy.cc
piodul: Done - created an overload of `storage_proxy::create_write_response_handler` that takes `hint_wrapper` as the first parameter

service/storage_proxy.cc
piodul: I preemptively changed it so that timeout is based on `write_request_timeout_in_ms`.

slivne:
I think repair has a special mutation as well - can't we reuse that ?

piodul:
@slivne Could you clarify what you mean by "special mutation"?

slivne:
I'll elaborate - yet am not sure I can actually point you to the right
direction

AFAIK one of the updates Asias had todo for the row level repair is move
the repair mutations that used the regular verb ones to a special one so
that they would be tagged correctly in scheduling groups, in addition they
are also using a special connection and not the regular mutation connection.

Hints and Repair are both in the streaming scheduling group, they both can
share the same connections, as such I am not sure if they can not share the
same verb.

Asias is on vacation this week.... he should be back next week and can help
and see if I am making any sense.

piodul:
@slivne I see what you mean. I can see that there is a number of verbs related to repair, but I don't think those can be used here. There is one that seems to write some mutations (`REPAIR_PUT_ROW_DIFF`), but is seems to carry additional information which is not available/applicable here, and its semantic is more complicated than just applying a mutation.

We have already introduced the `HINT_MUTATION` verb in an older patch, which is dedicated for handling hint traffic. It is used already for sending/receiving non-orphaned hints, this patch makes orphaned hints be handled with this verb as well.

service/storage_proxy.cc
vladzcloudius: Got it.

service/storage_proxy.cc
vladzcloudius: @piodul it was @avikivity who added this timeout value in "storage_proxy: implement write timeouts" patch.

@gleb-cloudius this case is not supposed to be hit when the destination node went DOWN but when it's no longer a replica for the token in question.

Going DOWN is not a justification for sending a hint to all other replicas - it's a totally normal scenario. In such a case we should not attempt sending to the node and wait till it becomes up again: look at `sender::can_send()`.

vladzcloudius: > I preemptively changed it so that timeout is based on `write_request_timeout_in_ms`.

@piodul I'm not sure we want to create a correlation between a user defined write request timeout and hints mutation timeout. Having an infinite timeout here should be totally safe as clarified in Avi's comment in code.

gleb-cloudius: No, `mutation` is a one way verb meaning that the sender only waits for message to be sent. Nothing is guarantied about it been processed and reply been eventually received. Sending with infinity timeout here only guaranties that eventually (after running for long enough) we will have infinity hanged writes. Avi lives in a fantasy world where networking is as reliable as shared memory communication.

piodul: I confirm Gleb's statement. I ran a test where I stopped one node and accumulated hints for it. Then, I started a new node which replaced the node that was down, but quickly stopped the new node while hints were still being sent. I could see that there was a small number of writes that lingered for exactly one hour:

linger

(I ran scylla without this patch so that it used the 1h timeout)

vladzcloudius: @gleb-cloudius I see. Thanks for a clarification and @piodul thanks for confirming it. 

I wonder why can't we have one of the following:
 * Connect the TCP socket termination and/or marking the remote node as DOWN/"left the ring (there are a few states that denote this)" to reporting a failure on the pending mutations requests sent to that node and waiting for the timeout. The drawback is that after that we may fail much more such requests if networking is flaky but maybe it's not a bad thing. - In this case you prefer to deliver the message about the flakiness to the user as soon as possible.
 * Make this VERB a two-way. Honestly it would make sense to make it such because the semantics of that timeout is kinda imply two-wayness then why not to make it explicit?

gleb-cloudius: First thing we may do, but we do not. And yes it will introduce false positives (as in message was handled, but failed it)

Second is because we reuse MUTATE code to send hints and the former is not always point-to-point so it cannot  be two-way. Making hints two-way will require writing a lot of new code for a little benefit. 

And if you cannot deliver a hint in to seconds may be the system has better things to do than to try hard and timeout and retry later is not such a bad idea.

vladzcloudius: Well, then probably we do need it (hints timeout) be configurable. Making it equal to `write_request_timeout_in_ms` is a possibility - probably better than introducing yet another config parameter.

slivne:
@asias - can you please have a look at the comment I had on the verbs - if it makes any sense to consolidate the two great , if it doesn't I am simply wrong.

In any case I don't think my comment should block this patchset from going in it can be a followup if it makes any sense

slivne:
@vladzcloudius can you please bless the patchset if all was adressed - and then I'll get a maintainer to push this in

vladzcloudius:
@slivne Sure.

asias:
> @asias - can you please have a look at the comment I had on the verbs - if it makes any sense to consolidate the two great , if it doesn't I am simply wrong.
> 
> In any case I don't think my comment should block this patchset from going in it can be a followup if it makes any sense

The hint can not use the repair verbs. In repair, we use the rpc stream interface for the mutation fragments. 

If there are a lot of such hints, it makes sense to use the rpc stream interface instead of the rpc verb interface, but it is a different issue. It should not block this patch.

asias:
@piodul Can you please use `hinted handoff` instead of hh in the patch title?

psarna:
queued to next

psarna:
@piodul dequeued - fails to compile after a rebase. Please update.

piodul:
@psarna Rebased and updated, it compiles now.

piodul:
@psarna rebased, ping

psarna:
vielen Dank, queued to /next

slivne:
was this merged as 01084a79b8f47f63aeb54bd110135f3850ad3fc8 ?

psarna:
@slivne yes, I'm not sure what's the backporting status though - should I backport to 3.3/3.2?

avikivity:
Merged as 01084a79b8f47f63aeb54bd110135f3850ad3fc8

avikivity:
3.3 has it as 183418f228a28f0e0ad26776b1539656aa716f5b

haaawk:
Just to clarify. The commit Avi mentioned fixes only counter corruption. This PR fixes also HH performance issue. I don't know whether we want performance fix backported as well or corruption fix is enough.

avikivity:
We backport performance fixes for regressions (and can special cases important issues).

This was not a regression fix, was it?

piodul:
@avikivity This patch fixes a performance problem that was always present in the implementation of hinted handoff, so I don't think it was a regression.

PR #5676, opened 2020-01-29
Change token representation to int64_t

haaawk:
After deprecating partitioners other than Murmur3 we can change the representation of tokens to int64_t. This will allow setting custom partitioner on each table. With this change partitioners become just converters from partition keys to tokens (int64_t). Following operations are no longer dependant on partitioner implementation:
1. Tokens comparison
2. Tokens serialization/deserialization to strings
3. Tokens serialization/deserialization to bytes
4. Sharding logic
5. Random token generation

This change will be followed by a PR that enables per table partitioner and then another PR that introduces a special partitioner for CDC tables.

Tests: unit(dev)

Results of memory footprint test:

Differences:

    in cache: 992 vs 984
    in memtable: 750 vs 742
    sizeof(cache_entry) = 112 vs 104
    -- sizeof(decorated_key) = 36 vs 32

MASTER:
mutation footprint:

    in cache: 992
    in memtable: 750
    in sstable: 351
    frozen: 540
    canonical: 827
    query result: 342

sizeof(cache_entry) = 112
-- sizeof(decorated_key) = 36
-- sizeof(cache_link_type) = 32
-- sizeof(mutation_partition) = 96
-- -- sizeof(_static_row) = 8
-- -- sizeof(_rows) = 24
-- -- sizeof(_row_tombstones) = 40

sizeof(rows_entry) = 232
sizeof(lru_link_type) = 16
sizeof(deletable_row) = 168
sizeof(row) = 112
sizeof(atomic_cell_or_collection) = 8

THIS PATCHSET:
mutation footprint:

    in cache: 984
    in memtable: 742
    in sstable: 351
    frozen: 540
    canonical: 827
    query result: 342

sizeof(cache_entry) = 104
-- sizeof(decorated_key) = 32
-- sizeof(cache_link_type) = 32
-- sizeof(mutation_partition) = 96
-- -- sizeof(_static_row) = 8
-- -- sizeof(_rows) = 24
-- -- sizeof(_row_tombstones) = 40

sizeof(rows_entry) = 232
sizeof(lru_link_type) = 16
sizeof(deletable_row) = 168
sizeof(row) = 112
sizeof(atomic_cell_or_collection) = 8


haaawk:
BYO runs here: https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/899/

haaawk:
Unfortunately BYO keeps failing with:
```
19:35:35  FAILED: build/debug/gen/alternator/expressionsLexer.cpp build/debug/gen/alternator/expressionsLexer.hpp build/debug/gen/alternator/expressionsParser.cpp build/debug/gen/alternator/expressionsParser.hpp 
19:35:35  sed -e '/^#if 0/,/^#endif/d' alternator/expressions.g > build/debug/gen/alternator/expressions.g && antlr3 build/debug/gen/alternator/expressions.g && sed -i -e '/^.*On :.*$/d' build/debug/gen/alternator/expressionsLexer.hpp && sed -i -e '/^.*On :.*$/d' build/debug/gen/alternator/expressionsLexer.cpp && sed -i -e '/^.*On :.*$/d' build/debug/gen/alternator/expressionsParser.hpp && sed -i -e 's/^\( *\)\(ImplTraits::CommonTokenType\* [a-zA-Z0-9_]* = NULL;\)$/\1const \2/' -e '/^.*On :.*$/d' -e '1i using ExceptionBaseType = int;' -e 's/^{/{ ExceptionBaseType\* ex = nullptr;/; s/ExceptionBaseType\* ex = new/ex = new/; s/exceptions::syntax_exception e/exceptions::syntax_exception\& e/' build/debug/gen/alternator/expressionsParser.cpp
19:35:35  library initialization failed - unable to allocate file descriptor table - out of memory/bin/sh: line 1:   103 Aborted                 (core dumped) antlr3 build/debug/gen/alternator/expressions.g
```
which is not related.

haaawk:
@kbr- if you rebase your change on top of this PR you will be able to remove the code duplication you have because of the logic you copied out from murmur3_partitioner. All the logic is in free functions located in dht/token-sharding.hh now.

kbr-:
@haaawk but maybe my change gets merged in first :P

haaawk:
If it does then I will clean it up. @kbr-

dht/token.hh
```diff
@@ -65,12 +68,20 @@ public:
...
 
     token() : _kind(kind::before_all_keys) {
     }
 
-    token(kind k, managed_bytes d) : _kind(std::move(k)), _data(std::move(d)) {
+    token(kind k, std::array<uint8_t, 8> d) : _kind(std::move(k)), _data(std::move(d)) {
+    }
+
+    token(kind k, const bytes& b) : _kind(std::move(k)) {
+        std::copy_n(b.begin(), 8, _data.begin());
```
gleb-cloudius: Can we assume 8 bytes here? Can we get more from an old node that may support different kind of tokens. Or less?

haaawk: We assume only Murmur3 is in use on the cluster and it always uses 8 bytes to represent tokens. Support for partitioners other than Murmur3 was dropped here: https://github.com/scylladb/scylla/pull/5636

gleb-cloudius: Assumptions are good, but don't we want to verify that they are true?

haaawk: Sure. I will add a check. Thanks.

dht/token.hh
```diff
@@ -65,12 +68,20 @@ public:
     //     [0x80] == 0.5
     //     [0x00, 0x80] == 1/512
     //     [0xff, 0x80] == 1 - 1/512
-    managed_bytes _data;
+    std::array<uint8_t, 8> _data;
```
kbr-: Why not `int64`?

haaawk: It's an intermediate step to split the change. The following patch turns it into int64_t.

dht/token.hh
```diff
@@ -82,21 +93,25 @@ public:
     }
 
     size_t external_memory_usage() const {
-        return _data.external_memory_usage();
+        return 0;
     }
 
     size_t memory_usage() const {
-        return sizeof(token) + external_memory_usage();
+        return sizeof(token) + sizeof(_data);
```
kbr-: I don't think this is right

dht/token.hh
```diff
@@ -82,21 +93,25 @@ public:
...
     }
 
     explicit token(token_view v) : _kind(v._kind), _data(v._data) {}
 
     operator token_view() const {
         return token_view(*this);
     }
+
+    bytes_view data() const {
+        return bytes_view(reinterpret_cast<int8_t*>(const_cast<uint8_t*>(_data.begin())), 8);
```
kbr-: Is this casting an array iterator to a pointer?

dht/token.hh
```diff
@@ -39,9 +42,9 @@ enum class token_kind {
 class token_view {
 public:
     token_kind _kind;
-    bytes_view _data;
+    std::reference_wrapper<const std::array<uint8_t, 8>> _data;
 
-    token_view(token_kind kind, bytes_view data) : _kind(kind), _data(data) {}
+    token_view(token_kind kind, std::array<uint8_t, 8>& data) : _kind(kind), _data(data) {}
```
kbr-: const?

sstables/types.hh
```diff
@@ -129,7 +129,7 @@ inline std::ostream& operator<<(std::ostream& o, indexable_element e) {
 
 class summary_entry {
 public:
-    dht::token_view token;
+    dht::token token;
```
kbr-: why?

kbr-: Maybe it'd make more sense to just remove the `token_view` class now. The reason this class exists is that it's trivially destructible, so it's easy to dispose giant vectors holding `token_view`s.
But now `token` is trivially destructible. (I'd suggest to add a `static_assert` to check this)

dht/token.hh
```diff
@@ -39,9 +42,9 @@ enum class token_kind {
 class token_view {
```
kbr-: Maybe it'd make more sense to just remove the `token_view` class now. The reason this class exists is that it's trivially destructible, so it's easy to dispose giant vectors holding `token_view`s (https://github.com/scylladb/scylla/commit/dddc7e1676972a3acd34f157b4d430593c4b4099).
But now `token` is trivially destructible. (I'd suggest to add a `static_assert` to check this)

kbr-: Oh, nevermind, I see you do it in a later commit.
https://github.com/scylladb/scylla/pull/5676/commits/bb240792a92004cd726e748a272d2bc960ffd9a6

dht/murmur3_partitioner.cc
```diff
@@ -78,13 +70,7 @@ murmur3_partitioner::get_token(bytes_view key) const {
 
 token
 murmur3_partitioner::get_token(uint64_t value) const {
```
kbr-: This commit does more than just changing `_data` to `int64_t`, and the commit message doesn't mention anything about that, forcing reviewers to guess...

This commit changes `token` to store the data in host byte order, where previously it stored data in network byte order.

dht/token.hh
```diff
@@ -60,28 +60,23 @@ class token {
...
 
-    token(kind k, std::array<uint8_t, 8> d) : _kind(std::move(k)), _data(std::move(d)) {
-    }
+    token(kind k, int64_t d)
+        : _kind(std::move(k))
+        , _data(d == std::numeric_limits<int64_t>::min() ? std::numeric_limits<int64_t>::max() : d) { }
 
     token(kind k, const bytes& b) : _kind(std::move(k)) {
-        std::copy_n(b.begin(), 8, _data.begin());
+        std::copy_n(b.begin(), sizeof(_data), reinterpret_cast<int8_t *>(&_data));
```
kbr-: what if `bytes.size() != sizeof(_data)`?

dht/token.hh
```diff
@@ -60,28 +60,23 @@ class token {
...
+        _data = net::ntoh(_data);
     }
 
     token(kind k, bytes_view b) : _kind(std::move(k)) {
-        std::copy_n(b.begin(), 8, _data.begin());
+        std::copy_n(b.begin(), sizeof(_data), reinterpret_cast<int8_t *>(&_data));
```
kbr-: What if `b.size() != sizeof(_data)`?

dht/token.hh
```diff
@@ -84,6 +84,15 @@ public:
...
         return b;
     }
+
+    /**
+     * Calculate a token representing the approximate "middle" of the given
+     * range.
+     *
+     * @return The approximate midpoint between left and right.
+     */
+    static token midpoint(const token& left, const token& right);
```
kbr-: Why make it a static method instead of just a function?

dht/token.hh
```diff
@@ -93,6 +93,11 @@ public:
      */
     static token midpoint(const token& left, const token& right);
 
+    /**
+     * @return a randomly generated token
+     */
+    static token get_random_token();
```
kbr-: Why make it a static method instead of just a function?

dht/murmur3_partitioner.cc
kbr-: Also, maybe `get_token` should take `int64_t` now. I never understood why it takes `uint64_t`.

dht/token.cc
```diff
@@ -112,4 +112,13 @@ token token::get_random_token() {
     return {kind::key, dht::get_random_number<int64_t>()};
 }
 
+token token::from_sstring(const sstring& t) {
+    auto lp = boost::lexical_cast<long>(t);
+    if (lp == std::numeric_limits<long>::min()) {
+        return minimum_token();
+    } else {
+        return token(kind::key, uint64_t(lp));
```
kbr-: The `uint64_t` conversion is completely unnecessary (it will get converted  back to `int64_t` in the constructor call)

dht/token.cc
```diff
@@ -112,4 +112,13 @@ token token::get_random_token() {
     return {kind::key, dht::get_random_number<int64_t>()};
 }
 
+token token::from_sstring(const sstring& t) {
+    auto lp = boost::lexical_cast<long>(t);
```
kbr-: I wonder why this function uses `long` here and not `int64_t`. Probably because that's what Origin did with Java.
I wonder what happens if we run this on a 32bit machine and the token doesn't fit in 32 bits.

dht/murmur3_partitioner.cc
```diff
@@ -110,9 +96,7 @@ inline int64_t long_token(token_view t) {
         return std::numeric_limits<long>::min();
```
kbr-: This `long` here makes me uncomfortable

dht/token.hh
```diff
@@ -113,6 +113,16 @@ public:
...
     static dht::token from_bytes(bytes_view bytes);
 
+    /**
+     * Calculate the deltas between tokens in the ring in order to compare
+     *  relative sizes.
+     *
+     * @param sortedtokens a sorted List of tokens
+     * @return the mapping from 'token' to 'percentage of the ring owned by that token'.
+     */
+    static std::map<token, float> describe_ownership(const std::vector<token>& sorted_tokens);
```
kbr-: Why make it a static method instead of a normal function?

dht/token.hh
```diff
@@ -122,6 +123,8 @@ public:
      */
     static std::map<token, float> describe_ownership(const std::vector<token>& sorted_tokens);
 
+    static data_type get_token_validator();
```
kbr-: Why make it a static method instead of a normal function?

dht/token.hh
```diff
@@ -125,6 +125,13 @@ public:
 
     static data_type get_token_validator();
 
+    /**
+     * Gets the first shard of the minimum token.
+     */
+    static unsigned shard_of_minimum_token() {
```
kbr-: Why make it a static method instead of a normal function?

dht/i_partitioner.cc
```diff
@@ -22,6 +22,7 @@
 #include "i_partitioner.hh"
```
kbr-: I think we should move sharding logic out of partitioner completely (far from the `i_partitioner` class and the i_partitioner.hh,cc files).

As you said in the PR cover letter, partitioners should be just `key -> token` functions. Sharding logic says which token belongs to which shard, which is independent of the `key -> token` conversion.

dht/token-sharding.hh
```diff
@@ -0,0 +1,40 @@
...
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "dht/token.hh"
+
+namespace dht {
+
+token bias(uint64_t value);      // translate from a zero-baed range
```
kbr-: This is a good opportunity to fix the typo ("zero-baed range")

dht/token-sharding.hh
```diff
@@ -0,0 +1,40 @@
...
+
+uint64_t unbias(const token& t); // translate to a zero-baed range
+
+unsigned zero_based_shard_of(uint64_t zero_based_token, unsigned shards, unsigned sharding_ignore_msb_bits);
+
+std::vector<uint64_t> init_zero_based_shard_start(unsigned shards, unsigned sharding_ignore_msb_bits);
```
kbr-: `bias`, `unbias`, `zero_based_shard_of`, and `init_zero_based_shard_start` were previously private murmur3_partitioner methods, and they are still an implementation detail after this change.

I suggest dropping them from the header file and making them `static` inside dht/token.cc.

configure.py
```diff
@@ -678,6 +678,7 @@ def find_headers(repodir, excluded_dirs):
                 'gms/application_state.cc',
                 'gms/inet_address.cc',
```
avikivity: Please reserve the word "fix" for bug fixes, and use a different word for the pull request title and the changelogs, to reduce confusion.

dht/token.hh
avikivity: It's better not to mix code movement with changes.

dht/murmur3_partitioner.cc
avikivity: Correct, it should be int64_t. Can come in a separate patch.

dht/token.cc
```diff
@@ -29,8 +29,8 @@
 
 namespace dht {
 
-static const token min_token{ token::kind::before_all_keys, {} };
-static const token max_token{ token::kind::after_all_keys, {} };
+static const token min_token{ token::kind::before_all_keys, std::array<uint8_t, 8>{} };
+static const token max_token{ token::kind::after_all_keys, std::array<uint8_t, 8>{} };
```
avikivity: Unrelated note: I'd like to remove token::kind too, and replace it with nonwrapping_range's support for open-ended ranges.

dht/token.hh
avikivity: Huh, github shows me this conversation on every commit that touches this file. Thanks github!

dht/token.hh
avikivity: Yes

dht/token.hh
```diff
@@ -65,12 +68,20 @@ public:
...
     //     [0x00, 0x80] == 1/512
     //     [0xff, 0x80] == 1 - 1/512
-    managed_bytes _data;
+    std::array<uint8_t, 8> _data;
 
     token() : _kind(kind::before_all_keys) {
     }
 
-    token(kind k, managed_bytes d) : _kind(std::move(k)), _data(std::move(d)) {
+    token(kind k, std::array<uint8_t, 8> d) : _kind(std::move(k)), _data(std::move(d)) {
```
avikivity: Note that std::move()ing std::array is copying it. No harm done though.

dht/token.hh
avikivity: Right, it's sizeof(token) + 0.

dht/token.hh
avikivity: Array iterators are a pointer.

But why is the const_cast needed? you can reinterpret_cast to `const int8_t*`, that's all bytes_view needs.

You can use _data.size() instead of 8, nicer not to hard-code assumptions (even though we know they cannot change).

idl/token.idl.hh
```diff
@@ -6,7 +6,7 @@ class token {
         after_all_keys,
     };
     dht::token::kind _kind;
-    bytes _data;
+    bytes data();
 };
```
avikivity: const. I suggest renaming it to make it clear it's a non-native operation now, e.g. `to_bytes()`.

Maybe returning a bytes_view is sufficient.

avikivity: Scratch this comment, it is in the idl, and the code is correct.

avikivity: However, we should rename token::data() to discourage its use, can be done later.

dht/token.hh
```diff
@@ -23,7 +23,10 @@
 
 #include "bytes.hh"
 #include "utils/managed_bytes.hh"
+#include <seastar/net/byteorder.hh>
```
avikivity: I don't think this is needed in this patch.

dht/token.hh
```diff
@@ -23,7 +23,10 @@
...
 
+#include <array>
+#include <functional>
```
avikivity: And possibly this.

dht/i_partitioner.hh
```diff
@@ -848,13 +848,8 @@ template<>
 struct hash<dht::token> {
     size_t operator()(const dht::token& t) const {
         const auto& b = t._data;
-        if (b.size() == sizeof(size_t)) { // practically always
-            return read_le<size_t>(reinterpret_cast<const char*>(b.data()));
-        }
-        return hash_large_token(b);
+        return read_le<size_t>(reinterpret_cast<const char*>(b.data()));
```
avikivity: This breaks on a machine where sizeof(size_t) != 8. Can be fixed by s/size_t/uint64_t/. The fact that the function returns size_t doesn't hurt.

dht/i_partitioner.cc
```diff
@@ -468,14 +468,3 @@ split_ranges_to_shards(const dht::token_range_vector& ranges, const schema& s) {
...
-
-namespace std {
-
-size_t
-hash<dht::token>::hash_large_token(const std::array<uint8_t, 8>& b) const {
-    std::array<uint64_t, 2> result;
-    utils::murmur_hash::hash3_x64_128(b.begin(), 8, 0, result);
-    return result[0];
-}
-
```
avikivity: Changelog should explain why large_hash_token() can be removed.

dht/i_partitioner.hh
```diff
@@ -369,12 +369,6 @@ public:
     bool is_equal(token_view t1, token_view t2) const {
         return tri_compare(t1, t2) == 0;
     }
-    /**
-     * @return true if t1's _data array is less then t2's. _kind comparison should be done separately.
-     */
-    bool is_less(token_view t1, token_view t2) const {
-        return tri_compare(t1, t2) < 0;
-    }
```
avikivity: Changelog.

dht/i_partitioner.hh
```diff
@@ -264,14 +264,6 @@ public:
...
-    /**
-     * @return A token smaller than all others in the range that is being partitioned.
-     * Not legal to assign to a node or key.  (But legal to use in range scans.)
-     */
-    token get_minimum_token() {
-        return dht::minimum_token();
-    }
-
     /**
      * @return a token that can be used to route a given key
```
avikivity: You can squash all those remove unused into one patch.

dht/i_partitioner.hh
```diff
@@ -847,8 +847,7 @@ namespace std {
 template<>
 struct hash<dht::token> {
     size_t operator()(const dht::token& t) const {
-        const auto& b = t._data;
-        return read_le<size_t>(reinterpret_cast<const char*>(b.data()));
+        return t._data;
```
avikivity: This changes the return value of the function, if we run on a big-endian machine. We shouldn't depend on the hash value being stable across nodes, but how can we verify this?

dht/token.hh
avikivity: Hey, from yet another commit!

sstables/sstables.cc
```diff
@@ -573,7 +573,7 @@ future<> parse(sstable_version_types v, random_access_reader& in, summary& s) {
                         auto position = seastar::read_le<uint64_t>(buf.get());
                         auto token = dht::global_partitioner().get_token(key_view(key_data));
                         s.add_summary_data(token.data());
-                        s.entries.push_back({ dht::token(dht::token::kind::key, std::move(token._data)), key_data, position });
+                        s.entries.push_back({ token, key_data, position });
```
avikivity: Why this change? It makes sense, but is unrelated to the patch, no?

dht/token.hh
kbr-: > Array iterators are a pointer.

The C++ standard says that the iterator type is implementation-defined. I don't think we should depend on it (especially if we want to compile Scylla with compilers other than g++ in the future).

In any case, this is an abstraction leak.

sstables/sstables.cc
```diff
@@ -1971,7 +1971,7 @@ void maybe_add_summary_entry(summary& s, const dht::token& token, bytes_view key
         state.next_data_offset_to_write_summary += state.summary_byte_cost * entry_size;
         s.add_summary_data(token.data());
         auto key_data = s.add_summary_data(key);
-        s.entries.push_back({ dht::token(dht::token::kind::key, token._data), key_data, index_offset });
+        s.entries.push_back({ token, key_data, index_offset });
     }
```
avikivity: Here too.

dht/token.hh
avikivity: It is more discoverable. I like free functions myself but agree that discoverability is better as a member.

cql3/functions/token_fct.hh
```diff
@@ -65,7 +65,7 @@ public:
         auto key = partition_key::from_optional_exploded(*_schema, parameters);
         auto tok = dht::global_partitioner().get_token(*_schema, key);
         warn(unimplemented::cause::VALIDATION);
```
avikivity: Changelog should explain why you remove it.

avikivity:
Looks good overall.

avikivity:
Please add the output of memory_footprint_test, before and after, to the cover letter.

cql3/functions/token_fct.hh
haaawk: done.

dht/i_partitioner.cc
haaawk: I wanted to do that but sharding logic depends on shards_count, sharding_ignore_msb_bits and shard_start. At the moment there's no better place to keep this state than partitioner.

dht/i_partitioner.hh
haaawk: done.

dht/token-sharding.hh
haaawk: done

dht/token-sharding.hh
haaawk: Good idea. I moved all of them but init_zero_based_shard_start which is actually called from dht/i_partitioner.cc

dht/token.hh
haaawk: As Avi answered on one of your other comments about free functions, this is about discoverability. All those functions are token related so they are grouped under token namespace. I will resolve all other comments regarding free functions because they have the same answer.

dht/token.hh
haaawk: Without it the compilation fails with:
```
In file included from dht/token.cc:27:
./dht/token.hh:125:12: error: ‘data_type’ does not name a type
dht/token.cc:184:1: error: no declaration matches ‘data_type dht::token::get_token_validator()’
dht/token.cc:184:1: note: no functions named ‘data_type dht::token::get_token_validator()’In file included from dht/token.cc:27:
./dht/token.hh:42:7: note: ‘class dht::token’ defined here
```

dht/token.cc
haaawk: I had exactly the same thought. I will work on this in the following PR.

dht/token.cc
haaawk: I didn't want to change the logic to make it easier for reviewers. I will fix this and your other comment on this commit in a separate patch (but in this PR).

dht/token.cc
haaawk: done.

dht/murmur3_partitioner.cc
haaawk: done.

dht/i_partitioner.hh
haaawk: done.

dht/i_partitioner.hh
haaawk: done

dht/i_partitioner.cc
haaawk: done.

dht/token.hh
haaawk: done

dht/token.hh
haaawk: I removed move and took d by const& to avoid one copy.

dht/token.hh
haaawk: Removed.

dht/token.hh
haaawk: Right. Thanks for catching. Fixed.

sstables/sstables.cc
haaawk: What do you want me to do here @avikivity?

dht/token.hh
haaawk: I added a check.

dht/token.hh
haaawk: I added a check.

dht/murmur3_partitioner.cc
haaawk: Probably but it's out of the scope of this patchset.

dht/token.hh
haaawk: I did bytes_view(reinterpret_cast<const int8_t*>(_data.data()), _data.size()) which is better. Don't worry about this too much though - it goes away in the following patches.

dht/token.hh
haaawk: referrence_wrapper comes from here.

sstables/sstables.cc
haaawk: I moved it out to a separate patch.

sstables/sstables.cc
haaawk: I moved it out to a separate patch.

dht/murmur3_partitioner.cc
haaawk: I added commit message and extracted the normalization logic to a function as before.

dht/i_partitioner.hh
haaawk: I don't understand. previously b.data() was in le and read_le was converting it to host byte ordering. Now t._data is in host byte ordering. Isn't it the same?

configure.py
haaawk: I will.

haaawk:
Rebased on top of master and addressed all the comments.

haaawk:
I wasn't able to get BYO pass but I got two runs that failed with totally different sets of dtests which proves the tests are flakey and it would be possible to get all the tests to pass:
https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/914/
https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/915/

haaawk:
Any more comments @avikivity ?

dht/token.hh
avikivity: @kbr- I was sure that std::array's iterators are always pointers, but can't find evidence.

avikivity:
> Please add the output of memory_footprint_test, before and after, to the cover letter.

@haaawk

avikivity:
@haaawk any significant changes, or just fixing the nitpicks?

haaawk:
Right. Sorry. I missed that one. Will do.

haaawk:
> @haaawk any significant changes, or just fixing the nitpicks?

Just fixing the comments. Nothing revolutionary. Mostly code organization, style, constness, and casts. The biggest was correctness in token::memory_usage.

I answered you comment on hashing - I don't understand what's the problem there.

dht/i_partitioner.hh
avikivity: Actually the data was in big endian order. So we would reverse the byte order unconditionally.

Now you're unconditionally not reversing it.

To be safe, let's unconditionally reverse it now. Something like read_le(read_be()) should work.

avikivity:
I'll re-review tomorrow, so @kbr- can review again if he wishes. Remind me if ("if") i forget.

dht/i_partitioner.hh
haaawk: Hmm. read_le calls le_to_cpu which in turn calls le64toh this will do nothing on little-endian processors and reverse bytes on big-endian processors, no? This means that on little-endian returned value will have bytes in big-endian order and on big-endian processors it will have bytes in little-endian order. Or do I miss something?

Little-endian CPU case:
b.data() -> bytes in network (big-endian) order
read_le calls le64toh which does nothing -> returning int64_t composed of bytes in unchanged order = network (big-endian) order

Big-endian CPU case: 
b.data() -> bytes in network (big-endian) order
read_le calls le64toh which reverses bytes -> returning int64_t composed of bytes in reversed order = little-endian order

This means you are right and we always reverse the order :) - I will use read_le(read_be()). Thanks.

haaawk:
I rebased on top of master and fixed the byte order issue in hashing.

kbr-:
Request: when you do changes and rebase, force-push twice: first with the changes only, then with the rebase only. This way the reviewers can see the changes diff.

haaawk:
> Request: when you do changes and rebase, force-push twice: first with the changes only, then with the rebase only. This way the reviewers can see the changes diff.

ok

sstables/sstables.cc
```diff
@@ -1971,7 +1971,7 @@ void maybe_add_summary_entry(summary& s, const dht::token& token, bytes_view key
         state.next_data_offset_to_write_summary += state.summary_byte_cost * entry_size;
```
kbr-: should be "sstables" instead of "stables" in the commit log

kbr-:
LGTM

sstables/sstables.cc
haaawk: Fixed.

haaawk:
Fixed a commit message.

haaawk:
Could you re-review @avikivity, please? I'm working on getting the memory_footprint_test to work.

haaawk:
Rebased on top of master.

Results of memory footprint test:

Differences:
 - in cache:     992 vs 984
 - in memtable:  750 vs 742
 sizeof(cache_entry) = 112 vs 104
 -- sizeof(decorated_key) = 36 vs 32

MASTER:
mutation footprint:
 - in cache:     992
 - in memtable:  750
 - in sstable:   351
 - frozen:       540
 - canonical:    827
 - query result: 342

 sizeof(cache_entry) = 112
 -- sizeof(decorated_key) = 36
 -- sizeof(cache_link_type) = 32
 -- sizeof(mutation_partition) = 96
 -- -- sizeof(_static_row) = 8
 -- -- sizeof(_rows) = 24
 -- -- sizeof(_row_tombstones) = 40

 sizeof(rows_entry) = 232
 sizeof(lru_link_type) = 16
 sizeof(deletable_row) = 168
 sizeof(row) = 112
 sizeof(atomic_cell_or_collection) = 8

THIS PATCHSET:
mutation footprint:
 - in cache:     984
 - in memtable:  742
 - in sstable:   351
 - frozen:       540
 - canonical:    827
 - query result: 342

 sizeof(cache_entry) = 104
 -- sizeof(decorated_key) = 32
 -- sizeof(cache_link_type) = 32
 -- sizeof(mutation_partition) = 96
 -- -- sizeof(_static_row) = 8
 -- -- sizeof(_rows) = 24
 -- -- sizeof(_row_tombstones) = 40

 sizeof(rows_entry) = 232
 sizeof(lru_link_type) = 16
 sizeof(deletable_row) = 168
 sizeof(row) = 112
 sizeof(atomic_cell_or_collection) = 8

avikivity:
@haaawk thanks (nothing unexpected there, it's just a verification/demonstration).

Please add the results to the cover letter.

haaawk:
done.

avikivity:
Reviewing

dht/i_partitioner.cc
```diff
@@ -472,11 +472,9 @@ split_ranges_to_shards(const dht::token_range_vector& ranges, const schema& s) {
 namespace std {
```
avikivity: Please add an explanation in the changelog about why this is desirable safe. In general changelogs should explain whys, not just whats.

dht/i_partitioner.hh
```diff
@@ -847,8 +848,7 @@ namespace std {
 template<>
 struct hash<dht::token> {
     size_t operator()(const dht::token& t) const {
-        const auto& b = t._data;
-        return read_le<int64_t>(reinterpret_cast<const char*>(b.data()));
+        return bswap_64(t._data);
     }
```
avikivity: Please add a comment here because it now looks odd. It's not the code's fault (it inherited the strangeness), but it should make it less puzzling to readers.

avikivity:
Looks good, I had two comment-related comments.

dht/i_partitioner.hh
haaawk: done.

dht/i_partitioner.cc
haaawk: done

haaawk:
@avikivity 
I rebased on top of master. Fixed conflicts with Kamil's change. Removed some code duplication Kamil had to do because my change was not ready. Added two comments you asked me to do.

PR #5680, opened 2020-01-30
atomic_cell: fix printing counter cells

### LINKED ISSUE #5616: Trace loglevel + counters = abort
Happens on current master, e.g. on 8fde8e36007f8abcc41190f10661f6e26ab48f96
Run scylla with TRACE loglevel on `mutation_data`
```
[julex@localhost scylla]$ ./build/dev/scylla --developer-mode=true --logger-log-level mutation_data=trace
```
Then create and bump a counter:
```
cqlsh> CREATE KEYSPACE ks with replication = {'class':'SimpleStrategy', 'replication_factor': 1};
cqlsh> CREATE TABLE ks.tbl (name text PRIMARY KEY, entities counter);
cqlsh> UPDATE ks.tbl SET entities = entities + 1 where name = 'asdf';
NoHostAvailable:
```
Scylla aborts on line https://github.com/scylladb/scylla/blob/a74a82d4d245d75311c19d0d8fba404d5d75937e/service/storage_proxy.cc#L1897

jul-stas:
Counter cell can represent two things: "delta" (counter update)
and "shards" (counter value). Both are contained in the same
type, but there are differences: e.g. calling `counter_update_value()`
on "shards" cell is a memory corruption. Similarly, call to `value()`
on "delta" cell leads to `abort()` - that's what happens in 
`operator<<(..., atomic_cell_view::printer&)`.

This commit introduces a quickfix: it determines cell type on high
level (from `atomic_cell_view`) and prints counter contents by
`counter_cell_view` or `atomic_cell_view::counter_update_value()`.

Fixes #5616

### DIFF
--- a/atomic_cell.cc
+++ b/atomic_cell.cc
@@ -21,6 +21,7 @@
 
 #include "atomic_cell.hh"
 #include "atomic_cell_or_collection.hh"
+#include "counters.hh"
 #include "types.hh"
 
 /// LSA mirator for cells with irrelevant type
@@ -218,7 +219,9 @@ std::ostream&
 operator<<(std::ostream& os, const atomic_cell_view& acv) {
     if (acv.is_live()) {
         return fmt_print(os, "atomic_cell{{{},ts={:d},expiry={:d},ttl={:d}}}",
-            to_hex(acv.value().linearize()),
+            acv.is_counter_update()
+                    ? "counter_update_value=" + to_sstring(acv.counter_update_value())
+                    : to_hex(acv.value().linearize()),
             acv.timestamp(),
             acv.is_live_and_has_ttl() ? acv.expiry().time_since_epoch().count() : -1,
             acv.is_live_and_has_ttl() ? acv.ttl().count() : 0);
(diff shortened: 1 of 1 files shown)


atomic_cell.cc
```diff
@@ -238,8 +241,21 @@ operator<<(std::ostream& os, const atomic_cell_view::printer& acvp) {
...
+                counter_cell_view::with_linearized(acv, [&cell_value_string_builder] (counter_cell_view& ccv) {
+                    cell_value_string_builder << ::join(", ", ccv.shards());
+                });
+            }
+        } else {
+            cell_value_string_builder << type.to_string(acv.value().linearize());
+        }
         return fmt_print(os, "atomic_cell{{{},ts={:d},expiry={:d},ttl={:d}}}",
-            type.to_string(acv.value().linearize()),
+            cell_value_string_builder.str(),
```
avikivity: Why not fix to_string() for counter types?

Right now it calls fail(unimlemented).

jul-stas: I was thinking about it, but found no easy way. BTW. abort comes from:
https://github.com/scylladb/scylla/blob/a16cb0071946f6a08d77e26c7eaff092ea375a4e/data/cell.hh#L651

Cell here goes as `bytes`, through `deserialize_visitor` and `to_string_impl_visitor`. In serialized representation we don't know whether it's a "delta" or "shards". Maybe it can be worked around (e.g. by guessing from buffer size), but this idea was worrying me.

Moreover, I think `counter_type_impl` is not prepared to store "shards". At least I don't see at this point how to squeeze "shards" and "deltas" into one concrete_type. I'll accept any advice here.

And... it was tempting with simplicity - without hijacking `imr::pod<int64_t>` in `atomic_cell_view::value()`, without serializing back and forth...

avikivity: No, my idea is completely wrong. counter_type_impl is the front-end's view of a counter, which is just an int64_t. atomic_cell uses the back-end's view of a counter, which has all these shard things. Usually the front-end and the back-end share the same view, but for counters, they don't.

atomic_cell.cc
```diff
@@ -238,8 +241,21 @@ operator<<(std::ostream& os, const atomic_cell_view::printer& acvp) {
...
     auto& acv = acvp._cell;
     if (acv.is_live()) {
+        std::ostringstream cell_value_string_builder;
+        if (type.is_counter()) {
+            if (acv.is_counter_update()) {
+                cell_value_string_builder << "counter_update_value=" << acv.counter_update_value();
+            } else {
+                cell_value_string_builder << "shards: ";
+                counter_cell_view::with_linearized(acv, [&cell_value_string_builder] (counter_cell_view& ccv) {
+                    cell_value_string_builder << ::join(", ", ccv.shards());
```
haaawk: There's one thing I don't understand. cell_value_string_builder is captured by reference and this lambda is not mutable so it's a const&. Why does cell_value_string_builder << ... compile then?

jul-stas:
ping, @avikivity - are we ready to merge this?

haaawk:
@avikivity What do we have to do to move forward with this?

haaawk:
@avikivity ping

avikivity:
@haaawk I see there's an unanswered question (from you)

haaawk:
We sorted this out offline. Sorry for not resolving earlier @avikivity .

atomic_cell.cc
haaawk: Resolved offline.

PR #5712, opened 2020-02-04
memory_footprint_test: Make it clear it has to run with -c1

haaawk:
The test fails when run on number of cores different than 1.


avikivity:
Queued as 597e11212046f9e594ab35cc8a16b3c5fb58c584

PR #5713, opened 2020-02-04
schema: store and use the index type internally

amnonh:
This series changes how a view schema decides if it's an index and what kind of an index it is.

Specifically, when running the `describe` method of a view it needs to know if it's an index and if so, what type of index it is.

Instead of taking it from the base table, the index type will be stored as part of the schema and will be deduced when the schema is created.
This will solve the race condition when deleting a table (and implicitly its views) where a view needs to know if it is an index but the base table is already deleted.

### DIFF
--- a/schema.cc
+++ b/schema.cc
@@ -254,7 +254,7 @@ schema::raw_schema::raw_schema(utils::UUID id)
     : _id(id)
 { }
 
-schema::schema(const raw_schema& raw, std::optional<raw_view_info> raw_view_info)
+schema::schema(const raw_schema& raw, std::optional<raw_view_info> raw_view_info, bool is_index, bool is_global_index)
     : _raw(raw)
     , _offsets([this] {
         if (_raw._columns.size() > std::numeric_limits<column_count_type>::max()) {
@@ -340,6 +340,8 @@ schema::schema(const raw_schema& raw, std::optional<raw_view_info> raw_view_info
     rebuild();
     if (raw_view_info) {
         _view_info = std::make_unique<::view_info>(*this, *raw_view_info);
+        _is_index = is_index;
+        _is_global_index = is_global_index;
     }
 }
 
@@ -375,7 +377,7 @@ schema::schema(std::optional<utils::UUID> id,
         build_columns(regular_columns, column_kind::regular_column);
 
         return raw;
-    }(), std::nullopt)
+    }(), std::nullopt, false, false)
 {}
 
 schema::schema(const schema& o)
@@ -385,6 +387,8 @@ schema::schema(const schema& o)
     rebuild();
     if (o.is_view()) {
         _view_info = std::make_unique<::view_info>(*this, o.view_info()->raw());
+        _is_index = o._is_index;
+        _is_global_index = o._is_global_index;
     }
 }
 
@@ -1051,6 +1055,8 @@ schema_builder& schema_builder::with_view_info(utils::UUID base_id, sstring base
 }
 
 schema_builder& schema_builder::with_index(const index_metadata& im) {
+    _is_index = true;
+    _is_global_index = !im.local();
     _raw._indices_by_name.emplace(im.name(), im);
     return *this;
 }
@@ -1105,7 +1111,7 @@ schema_ptr schema_builder::build() {
     }
 
     prepare_dense_schema(new_raw);
-    return make_lw_shared<schema>(schema(new_raw, _view_info));
+    return make_lw_shared<schema>(schema(new_raw, _view_info, _is_index, _is_global_index));
 }
 
 schema_ptr schema_builder::build(compact_storage cp) {
--- a/schema.cc
+++ b/schema.cc
@@ -667,22 +667,14 @@ std::ostream& column_definition_as_cql_key(std::ostream& os, const column_defini
     return os;
 }
 
-static bool is_global_index(const utils::UUID& id, const schema& s) {
-    return  service::get_local_storage_service().db().local().find_column_family(id).get_index_manager().is_global_index(s);
-}
-
-static bool is_index(const utils::UUID& id, const schema& s) {
-    return  service::get_local_storage_service().db().local().find_column_family(id).get_index_manager().is_index(s);
-}
-
 
 std::ostream& schema::describe(std::ostream& os) const {
     os << "CREATE ";
     int n = 0;
 
     if (is_view()) {
-        if (is_index(view_info()->base_id(), *this)) {
-            auto is_local = !is_global_index(view_info()->base_id(), *this);
+        if (is_index()) {
+            auto is_local = !is_global_index();
 
             os << "INDEX " << cql3::util::maybe_quote(secondary_index::index_name_from_table_name(cf_name())) << " ON "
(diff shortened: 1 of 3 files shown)


schema.cc
```diff
@@ -254,7 +254,7 @@ schema::raw_schema::raw_schema(utils::UUID id)
     : _id(id)
 { }
 
-schema::schema(const raw_schema& raw, std::optional<raw_view_info> raw_view_info)
+schema::schema(const raw_schema& raw, std::optional<raw_view_info> raw_view_info, bool is_index, bool is_global_index)
```
nyh: Did we really need to add these as new separate parameters to the constructor? Wouldn't it be simpler if they were inside raw_view_info? The question of whether something is an index or not is a sub-question of whether it is a view (it first has to be a view - and if it's a view, it might be an index).
If this was inside raw_view_info, I think many of the other changes below will not be needed or simplified?

schema.cc
```diff
@@ -1051,6 +1055,8 @@ schema_builder& schema_builder::with_view_info(utils::UUID base_id, sstring base
 }
 
 schema_builder& schema_builder::with_index(const index_metadata& im) {
+    _is_index = true;
```
nyh: Does this work after a reboot, when the tables are reloaded from disk and not created from scratch? 
I thought you might need to modify `view_info::initialize_base_dependent_fields()` to figure out the right is_index from information in its base after a load. But I have to admit I am not very familiar with this code, or what exactly happens when reading schemas from disk. Please test this (you can add printouts to where you set _is_index = true, and see if it actually gets run during boot).

psarna:
Commit fb37394995 (which fixed the order of drop/create schema mutations) handled a quite similar issue, but by ensuring a correct order of applying schema change mutations. In particular, in `merge_tables_and_views`, I see:
```cpp
parallel_for_each(boost::range::join(tables_diff.dropped, views_diff.dropped), [&] (schema_diff::dropped_schema& dt) {
    auto& s = *dt.schema.get();
    return db.drop_column_family(s.ks_name(), s.cf_name(), [&] { return dt.jp.value(); });
}).get();
parallel_for_each(boost::range::join(tables_diff.created, views_diff.created), [&] (global_schema_ptr& gs) {
    return db.add_column_family_and_make_directory(gs);
}).get();
for (auto&& gs : boost::range::join(tables_diff.created, views_diff.created)) {
    db.find_column_family(gs).mark_ready_for_writes();
}
```
... which means that views and tables are dropped in parallel. That doesn't sound entirely correct, since views are better off being deleted before tables - just in case we delete views of the table and the table itself. I think the fix should not add boolean parameters and work around this incorrect order, but just fix the order itself - by dropping all views first, and only then continue with dropping tables. Opinions? This pull request can then be reduced to simply breaking this `parallel_for_each` into two separate stages - one for views, one for tables.

I'm generally not in favour of letting the table know it's an index, we actually strive to decouple these layers as much as possible. I *think* it may also cause problems after reboots, because you'd still need to check if the table backs and index (or, store the information somewhere) - and this check can recursively be prone to races.

nyh:
On Tue, Feb 4, 2020 at 12:48 PM Piotr Sarna <notifications@github.com>
wrote:

> Commit fb37394
> <https://github.com/scylladb/scylla/commit/fb37394995da8051057fbd4c67ab9df05052d189>
> (which fixed the order of drop/create schema mutations) handled a quite
> similar issue, but by ensuring a correct order of applying schema change
> mutations. In particular, in merge_tables_and_views, I see:
>
> parallel_for_each(boost::range::join(tables_diff.dropped, views_diff.dropped), [&] (schema_diff::dropped_schema& dt) {
>     auto& s = *dt.schema.get();
>     return db.drop_column_family(s.ks_name(), s.cf_name(), [&] { return dt.jp.value(); });
> }).get();parallel_for_each(boost::range::join(tables_diff.created, views_diff.created), [&] (global_schema_ptr& gs) {
>     return db.add_column_family_and_make_directory(gs);
> }).get();for (auto&& gs : boost::range::join(tables_diff.created, views_diff.created)) {
>     db.find_column_family(gs).mark_ready_for_writes();
> }
>
> ... which means that views and tables are dropped in parallel. That
> doesn't sound entirely correct, since views are better off being deleted
> before tables - just in case we delete views of the table and the table
> itself. I think the fix should not add boolean parameters and work around
> this incorrect order, but just fix the order itself - by dropping all views
> first, and only then continue with dropping tables. Opinions? This pull
> request can then be reduced to simply breaking this parallel_for_each
> into two separate stages - one for views, one for tables.
>

Can we be sure that the node receiving these mutations will receive them in
the expected order, and together?

psarna:
@nyh It definitely works like that if you send a `DROP TABLE` request - a node will prepare all mutations (table + indexes) and propagate them. I'm also pretty sure that schema reconciliation also works by sending a single batch of mutations in order to be "atomic", but that would be something that @tgrabiec can confirm or deny.

nyh:
@psarna I'm not familiar with the details of schema reconciliation. I thought a node may find itself (e.g., after some network disconnection or something) getting a bunch of differences at once, and needing to act on all of them, without knowing what the "right" order. But perhaps you're right that the exact order doesn't matter as long as we do the view stuff deletions first and only when those are completely done, we do the base deletions?

nyh:
By the way, in the next version of this patch (whatever approach it will take...) in the commit message please mention the issue being fixed. We have https://github.com/scylladb/scylla/issues/5686 - is there another issue on this?

psarna:
I think that even with networking issues, a single node will not send a partial set of mutations, so I would claim that it's enough to maintain the order in every atomic call to `merge_tables_and_views` (atomic only in the sense of not being interleaved with another schema update, since the code yields. Multiple schema merges are still serialized on shard0 with a semaphore though - ref: merge_lock()).

By the way - it would be nice to know how to reproduce this race mentioned in the issue description - either in a form of a test or just an explanation of how it could happen on a cluster - that could help verify if it's indeed only a problem with ordering deletes inside `merge_tables_and_views`.

amnonh:
I've placed the issue in the commits and not the header, I'll update it
once there will be a decision on the approach.

psarna:
Thanks, that adds way more context: https://github.com/scylladb/scylla/issues/5614. Well, if creating a snapshot is interleaved with `merge_tables_and_views`, and `merge_tables_and_views` yields after deleting some tables (but not their views yet), it definitely sounds possible that a base table was deleted before its view. I think that it makes sense to strengthen the ordering by ensuring that all view drops are performed before any table drops, and retest a couple of times. I'd expect the issue to go away.

psarna:
Is this pull request still relevant? I have it on my review list, but there's been no progress lately

amnonh:
I think there is a decision to take a different approach

eliransin:
@amnonh @nyh @psarna, it lookes like we have an agreement right?
We will deparalelize the drops of views and tables into two serial stages.
This is a blocker to 3.3, I would like to know we are aligned so I can submit
a PR for this ( I am going to follow sarnas recipe in https://github.com/scylladb/scylla/pull/5713#issuecomment-581849760)
- @sarna unless you want to be the one that submists it :smile:

psarna:
@eliransin go ahead, let me know when the patch is posted and I'll review it.

amnonh:
I'm closing this PR to get it out of the way

PR #5730, opened 2020-02-05
cdc: Handle non-atomic columns

### LINKED ISSUE #5669: Support collections in CDC
At the moment CDC does not log changes to collections. We should implement that. Maybe except of the list.

elcallio:
Fixes #5669

This implements non-atomic collection and UDT handling for
both cdc preimage + delta.

To be able to express deltas in a meaningful way (and reconstruct
using it), non-atomic values are represented somewhat
differently from regular values:

* maps - stored as is (frozen)
* sets - stored as is (frozen)
* lists - stored as map<timeuuid, value> (frozen)
  this allows reconstructing the list, as otherwise
  things like list[0] = value cannot be represented
  in a meaningful way
* udt - stored as tuple<tuple<field0>, tuple<field1>...> (frozen)
  UDTs are normally just tuples + metadata, but we need to
  distinguish the case of outer tuple element == null, meaning
  "no info/does not partake in mutation" from tuple element
  being a tuple(null) (i.e. empty tuple), meaning "set field to
  null"

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -44,6 +44,9 @@
 #include "cql3/untyped_result_set.hh"
 #include "log.hh"
 #include "json.hh"
+#include "types.hh"
+#include "concrete_types.hh"
+#include "types/listlike_partial_deserializing_iterator.hh"
 
 namespace std {
 
@@ -265,7 +268,29 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID> 
         for (const auto& column : columns) {
             auto type = column.type;
             if (is_data_col) {
-                type = tuple_type_impl::get_instance({ /* op */ data_type_for<column_op_native_type>(), /* value */ type});
+                type = visit(*type, make_visitor(
+                    // lists are represented as map<timeuuid, value_type>. Otherwise we cannot express delta
+                    [] (const list_type_impl& type) {
+                        return map_type_impl::get_instance(type.name_comparator(), type.value_comparator(), false);
+                    },
+                    // user types are expressed as tuple<tuple<field0>, tuple<field1>...>
+                    // the extra tuple allows us to distinguish null == <no info>
+                    // from <null> = set field to null
+                    [] (const user_type_impl& type) -> data_type {
+                        std::vector<data_type> types;
+                        types.reserve(type.size());
+                        for (auto& ft : type.field_types()) {
+                            types.emplace_back(tuple_type_impl::get_instance({ ft }));
+                        }
+                        return tuple_type_impl::get_instance(std::move(types));
+                    },
+                    // everything else is just frozen self
+                    [] (const abstract_type& type) {
+                        return type.freeze();
+                    }
+                ));
+
+                type = tuple_type_impl::get_instance({ /* op */ data_type_for<column_op_native_type>(), /* value */ type->freeze()});
             }
             b.with_column("_" + column.name(), type);
         }
@@ -569,10 +594,10 @@ public:
                     r.for_each_cell([&](column_id id, const atomic_cell_or_collection& cell) {
                         auto& cdef = _schema->column_at(ckind, id);
                         auto* dst = _log_schema->get_column_definition(to_bytes("_" + cdef.name()));
-                        // TODO: collections.
-                        if (cdef.is_atomic()) {
-                            column_op op;
+                        auto has_pirow = pirow && pirow->has(cdef.name_as_text());
+                        auto op = column_op::del;
 
+                        if (cdef.is_atomic()) {
                             values[1] = std::nullopt;
                             auto view = cell.as_atomic_cell(cdef);
                             if (view.is_live()) {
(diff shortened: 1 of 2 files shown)


cdc/log.cc
```diff
@@ -234,7 +237,29 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID>
...
+                    // user types are expressed as tuple<tuple<field0>, tuple<field1>...>
+                    // the extra tuple allows us to distinguish null == <no info>
+                    // from <null> = set field to null
+                    [] (const user_type_impl& type) -> data_type {
+                        std::vector<data_type> types;
+                        types.reserve(type.size());
+                        for (auto& ft : type.field_types()) {
+                            types.emplace_back(tuple_type_impl::get_instance({ ft }));
+                        }
+                        return tuple_type_impl::get_instance(std::move(types));
```
denesb: Will this be interpretable after the original UDT has been altered? Or is that not even a goal?

cdc/log.cc
```diff
@@ -526,36 +551,173 @@ class transformer final {
...
+                            auto in = collection_mutation_input_stream(mv.data);
+                            auto tomb = tombstone{};
+                            auto now = gc_clock::time_point::min();
+
+                            std::optional<column_op> oop;
+                            op = column_op::add;
+                            std::vector<bytes> buf;
+
+                            auto process = [&](auto key_2_type, auto append_value) {
+                                auto has_tomb = in.read_trivial<bool>();
```
denesb: This looks like an open-coded `collection_mutation_view::with_deserialized()`. Why not use that instead?

cdc/log.cc
```diff
@@ -526,36 +551,173 @@ class transformer final {
...
+                            auto mv = cell.as_collection_mutation();
+                            auto in = collection_mutation_input_stream(mv.data);
+                            auto tomb = tombstone{};
+                            auto now = gc_clock::time_point::min();
+
+                            std::optional<column_op> oop;
+                            op = column_op::add;
+                            std::vector<bytes> buf;
+
+                            auto process = [&](auto key_2_type, auto append_value) {
```
denesb: Space between: `]` and `(`. Gratuitous `auto`. Knowing these types at read-time would actually help understand the code.

cdc/log.cc
```diff
@@ -234,7 +237,29 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID>
         for (const auto& column : columns) {
             auto type = column.type;
             if (is_data_col) {
-                type = tuple_type_impl::get_instance({ /* op */ data_type_for<column_op_native_type>(), /* value */ type, /* ttl */long_type});
+                type = visit(*type, make_visitor(
+                    // lists are represented as map<timeuuid, value_type>. Otherwise we cannot express delta
+                    [] (const list_type_impl& type) {
+                        return map_type_impl::get_instance(type.name_comparator(), type.value_comparator(), false);
```
nyh: no need to freeze() here?

cdc/log.cc
```diff
@@ -526,36 +551,173 @@ class transformer final {
                     r.for_each_cell([&](column_id id, const atomic_cell_or_collection& cell) {
                         auto& cdef = _schema->column_at(ckind, id);
                         auto* dst = _log_schema->get_column_definition(to_bytes("_" + cdef.name()));
-                        // TODO: collections.
-                        if (cdef.is_atomic()) {
-                            column_op op;
+                        auto res_type = static_pointer_cast<const tuple_type_impl>(dst->type)->type(1);
```
nyh: I don't understand what this res_type is, or what type(1) does (sorry...) but looking below, it seems it is only used in the non-atomic case? So maybe it can be moved closer to where it's needed?

cdc/log.cc
```diff
@@ -526,36 +551,173 @@ class transformer final {
...
                                 values[1] = view.value().linearize();
                                 if (view.is_live_and_has_ttl()) {
                                     values[2] = long_type->decompose(data_value(view.ttl().count()));
                                 }
-                            } else {
-                                op = column_op::del;
-                            }
+                            } 
+                        } else {
+                            auto mv = cell.as_collection_mutation();
```
nyh: The long part of the code starting here is really difficult for me to understand. Can it be moved to a separate function, with clear inputs and outputs? Maybe even this function could be split into several functions doing different things?

cdc/log.cc
elcallio: No, and no. 
Longer answer: It is interpretable more or less the same way as it is now: You need to know what field 0, 1, 2 etc are in the consumer. The schema here will tell you field type, but no semantic info. Could maybe make the structure tuple<name, value>, but that seems expensive. 
I think it is "buyer beware" when using any non-atomics with cdc. Esp. udt.

cdc/log.cc
elcallio: The first iteration it was more different. It could perhaps be rewritten to use the existing helper now indeed.

cdc/log.cc
elcallio: Last "false" says it is a frozen map (not multicell).

cdc/log.cc
elcallio: No space. You are delusional. :-)
These are lambda callbacks, so types are tricky (don't want to make them std::func here). But maybe better param-names.

cdc/log.cc
elcallio: It is the type of the log value (second element of the column tuple). And, yes, it could be declared lower down now.

cdc/log.cc
elcallio: I think actually one reason was to avoid copying code between the caller lambdas instead, and also slightly less overhead, as we don't need all the views in a vector etc (we have other, funnier overhead vectors instead).

cdc/log.cc
elcallio: Maybe. However, as usual, gratuitous lambda captures makes breaking into functions dull work.

cdc/log.cc
elcallio: Changed to use with_deserialized instead. Not sure it is much smaller, but reuses code so.

cdc/log.cc
elcallio: Redone somewhat.

cdc/log.cc
elcallio: New version is a bit leaner.

cdc/log.cc
elcallio: Moved to local usage points.

elcallio:
Note: updated and rebased. Please re-review.

cdc/log.cc
```diff
@@ -538,23 +563,136 @@ class transformer final {
...
+                            std::vector<bytes> buf;
+                            values[1] = mv.with_deserialized(*cdef.type, [&](collection_mutation_view_description view) -> bytes_opt {
+                                if (view.tomb) {
+                                    // there is a tombstone with timestamp before this mutation.
+                                    // this is how a assign collection = <value> is represented.
+                                    oop = column_op::set;
+                                }
+                                auto process = [&](auto value_callback) {
+                                    for (auto& [key, value] : view.cells) {
+                                        auto live = value.is_live(view.tomb, now, false);
```
kbr-: We cannot refer to the current clock time there, this will give non-deterministic results, and we lose replayability!
If someone writes a live cell to the base table with ttl = 1 say, and the cell expires by the time we process it here in CDC code, we'll show it in the log as a dead cell (null) --- instead it should be a live cell (with a non-null value) with ttl 1.

I guess you want to distinguish here between:
1. a dead cell, i.e. assignment to null; this happens when `!value.is_live()`
2. an alive cell, perhaps with ttl or without, doesn't matter; this happens when `value.is_live()`

cdc/log.cc
```diff
@@ -538,23 +563,136 @@ class transformer final {
...
+                                        if (!live) {
+                                            // while technically mutations can express both add/set/delete
+                                            // in the same mutation, cql statements can't. And alternator
```
kbr-: > cql statements can't

We already know that they can.

The bigger problem is: we're doing a big redesign to use a bunch of columns instead of tuples to represent deleted values, so we should probably wait for that and then update this PR before merging it.

cdc/log.cc
```diff
@@ -538,23 +563,136 @@ class transformer final {
...
+                                        auto tt = static_pointer_cast<const tuple_type_impl>(res_type);
+                                        process([&](const bytes_view& key, const bytes_view& value, bool live) {
+                                            auto idx = deserialize_field_index(key);
+                                            tmp[0] = live ? bytes_view_opt{value} : std::nullopt;
+                                            res[idx].emplace(static_pointer_cast<const tuple_type_impl>(tt->all_types()[idx])->build_value(tmp));
+                                        });
+                                        return tt->build_value(res);
+                                    },
+                                    [&] (const abstract_type& o) -> bytes {
+                                        throw std::runtime_error(format("collection_mutation_view::is_any_live: unknown type {}", o.name()));
```
kbr-: forgot to change the error message

cdc/log.cc
```diff
@@ -538,23 +563,136 @@ class transformer final {
...
+                            op = column_op::add;
+
+                            std::vector<bytes> buf;
+                            values[1] = mv.with_deserialized(*cdef.type, [&](collection_mutation_view_description view) -> bytes_opt {
+                                if (view.tomb) {
+                                    // there is a tombstone with timestamp before this mutation.
+                                    // this is how a assign collection = <value> is represented.
+                                    oop = column_op::set;
+                                }
+                                auto process = [&](auto value_callback) {
```
kbr-: `process` is not a really descriptive name

cdc/log.cc
elcallio: Yes and no. The problem/reason why the tombstone is included is how a collection assignment is done: A tombstone is written for the column (the "tomb" here), and then we add cells. However, technically, it would be "legal" for any of these cells to be timestamped/tombstoned so that they fall _before_ the column tomb. It is obviously a ridiculous case, but we are only operating on raw mutations here... 
I agree with the reasoning that we might race the clock for an actual user provided ttl etc, and perhaps the esoteric case should just be ignored...

cdc/log.cc
elcallio: Yes, sure, push all the rebasing on the Swedes. We never complain.

cdc/log.cc
elcallio: It is a great error message and I support it. But I will also change it.

cdc/log.cc
elcallio: The surrounding function is called "transform". I think the descriptiveness is clear enough given the surrounding scope.

cdc/log.cc
kbr-: If cells are added with timestamp X, the tombstone is added with timestamp X-1. E.g. cql3/maps.cc, `maps::setter::execute`:
```
        mut.tomb = params.make_tombstone_just_before();
```
where `update_parameters::make_tombstone_just_before`:
```
    tombstone make_tombstone_just_before() const {
        return {_timestamp - 1, _local_deletion_time};
    }
```
(`_local_deletion_time` is the `expiry` value, we ignore it)

So if these cells and this tombstone were made in a single `column = value` assignment, the tombstone will never kill the cells.

If `tombstone's timestamp` won't be `cells' timestamp - 1`, I'll split them to separate mutations (I'm working on that splitting stuff right now). Don't worry about that.

Hell, maybe I'll even split the tombstone to a separate mutation, so a `column = value` statement for collection columns will always give two mutations: one of them does `column = null` with timestamp `X - 1`, and the other does a bunch of `column[idx] = val` with timestamp `X`. I have to decide what to do exactly there yet.

cdc/log.cc
kbr-: Come on, I had to rebase my CDC topology support change like 3 times because of your commits. :P

elcallio: Good. At least you're catching up in the pain ranks. :-)

cdc/log.cc
elcallio: Yes, assuming the mutation is well-formed. But otoh, we should probably just assume that, so I agree with your statement.

elcallio:
New version. Review as you see fit.

haaawk:
@nyh please review. We would like to try to merge it in by Thu so please let @elcallio know if you would like to see anything changed.

haaawk:
@denesb Do you have any other comments?

cdc/log.cc
```diff
@@ -265,7 +268,29 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID>
...
+                        std::vector<data_type> types;
+                        types.reserve(type.size());
+                        for (auto& ft : type.field_types()) {
+                            types.emplace_back(tuple_type_impl::get_instance({ ft }));
+                        }
+                        return tuple_type_impl::get_instance(std::move(types));
+                    },
+                    // everything else is just frozen self
+                    [] (const abstract_type& type) {
+                        return type.freeze();
```
haaawk: Do we have to call freeze() here if we call freeze() also in line 293?

cdc/log.cc
```diff
@@ -656,9 +799,13 @@ class transformer final {
                 columns.emplace_back(&cdef);
             });
         }
-
+
```
haaawk: remove whitespace.

cdc/log.cc
```diff
@@ -570,23 +595,141 @@ class transformer final {
...
+                        }
+
+                        values[0] = data_type_for<column_op_native_type>()->decompose(data_value(static_cast<column_op_native_type>(op)));
+                        res.set_cell(log_ck, *dst, atomic_cell::make_live(*dst->type, ts, tuple_type_impl::build_value(values)));
+
+                        if (has_pirow) {
+                            values[0] = data_type_for<column_op_native_type>()->decompose(data_value(static_cast<column_op_native_type>(column_op::set)));
+                            values[1] = cdef.is_atomic()
+                                ? pirow->get_blob(cdef.name_as_text())
+                                : values[1] = visit(*cdef.type, make_visitor(
```
haaawk: Why do we need to do values[1] = visit instead of just visit here? Is this needed because ?: requires same types?

elcallio:
Now rebased on latest master.

nyh:
Pushed to next.

PR #5731, opened 2020-02-06
Make replacing node take writes

asias:
storage_service: Make replacing node take writes

Background:

Replace operation is used to replace a dead node in the cluster.
Currently during replace operation, the replacing node does not take any
writes. As a result, new writes to a range after the sync for that range
is done, e.g., after streaming for that range is finished, will not be
synced to the replacing node. Hinted hand off or repair after the
replacing operation will help. But it is better if we can make the
writes to the replacing node to avoid any post replacing operation
actions.

After this series and repair based node operation series, the replace
operation will guarantee the replacing node has all the latest copy of
data including the new writes during the replace operation. In short, no
more repairs before or after the replacing operation. Just replacing the
node is enough.

Implementation:

1) Filter the node being replaced out of the natural endpoints in
   storage_proxy, so that:

- The node being replaced will not be selected as the target for
  normal write or normal read.

- Do not depend on the gossip liveness to avoid selecting replacing node
  for normal write or normal read when the replacing node has the same
  ip address as the node being replaced. No more special handling for
  hibernate state in gossip which makes it is simpler and more robust.
  Replacing node will be marked as UP.

2) Put the replacing node in the pending list, so that:

- Replacing node will take writes but write to replacing will not be
  counted as CL.

- Replacing node will not take normal read.

Example:

For example, with RF = 3, n1, n2, n3 in the cluster, n3 is dead and
being replaced by node n4. When n4 starts:

- writes to nodes {n1, n2, n3} are changed to
  normal_replica_writes = {n1, n2} and pending_replica_writes= {n4}.

- reads to nodes {n1, n2, n3} are changed to
  normal_replica_reads = {n1, n2} only.

This way, the replacing node n4 now takes writes but does not take reads.

Tests:

1) Measure the number of writes during pending period that is the
   replacing starts and finishes the replace operation.

- Start 5 nodes, n1 to n5.
- Stop n5
- Start write in the background
- Start n6 to replace n5
- Get scylla_database_total_writes metrics when the replacing node announces HIBERNATE (replacing) and NORMAL status.

Before:
2020-02-06 08:35:35.921837 Get metrics when other knows replacing node = HIBERNATE
2020-02-06 08:35:35.939493 scylla_database_total_writes: node1={'scylla_database_total_writes': 15483}
2020-02-06 08:35:35.950614 scylla_database_total_writes: node2={'scylla_database_total_writes': 15857}
2020-02-06 08:35:35.961820 scylla_database_total_writes: node3={'scylla_database_total_writes': 16195}
2020-02-06 08:35:35.978427 scylla_database_total_writes: node4={'scylla_database_total_writes': 15764}
2020-02-06 08:35:35.992580 scylla_database_total_writes: node6={'scylla_database_total_writes': 331}
2020-02-06 08:36:49.794790 Get metrics when other knows replacing node = NORMAL
2020-02-06 08:36:49.809189 scylla_database_total_writes: node1={'scylla_database_total_writes': 267088}
2020-02-06 08:36:49.823302 scylla_database_total_writes: node2={'scylla_database_total_writes': 272352}
2020-02-06 08:36:49.837228 scylla_database_total_writes: node3={'scylla_database_total_writes': 274004}
2020-02-06 08:36:49.851104 scylla_database_total_writes: node4={'scylla_database_total_writes': 262972}
2020-02-06 08:36:49.862504 scylla_database_total_writes: node6={'scylla_database_total_writes': 513}

Writes = 513 - 331

After:
2020-02-06 08:28:56.548047 Get metrics when other knows replacing node = HIBERNATE
2020-02-06 08:28:56.560813 scylla_database_total_writes: node1={'scylla_database_total_writes': 290886}
2020-02-06 08:28:56.573925 scylla_database_total_writes: node2={'scylla_database_total_writes': 310304}
2020-02-06 08:28:56.586305 scylla_database_total_writes: node3={'scylla_database_total_writes': 304049}
2020-02-06 08:28:56.601464 scylla_database_total_writes: node4={'scylla_database_total_writes': 303770}
2020-02-06 08:28:56.615066 scylla_database_total_writes: node6={'scylla_database_total_writes': 604}
2020-02-06 08:29:10.537016 Get metrics when other knows replacing node = NORMAL
2020-02-06 08:29:10.553257 scylla_database_total_writes: node1={'scylla_database_total_writes': 336126}
2020-02-06 08:29:10.567181 scylla_database_total_writes: node2={'scylla_database_total_writes': 358549}
2020-02-06 08:29:10.581939 scylla_database_total_writes: node3={'scylla_database_total_writes': 351416}
2020-02-06 08:29:10.595567 scylla_database_total_writes: node4={'scylla_database_total_writes': 350580}
2020-02-06 08:29:10.610548 scylla_database_total_writes: node6={'scylla_database_total_writes': 45460}

Writes = 45460 - 604

As we can see the replacing node did not take write before and take write after the patch.

2) Check log of writer handler in storage_proxy

storage_proxy - creating write handler for token: -2642068240672386521,
keyspace_name=ks, original_natrual={127.0.0.1, 127.0.0.5, 127.0.0.2},
natural={127.0.0.1, 127.0.0.2}, pending={127.0.0.6}

The node being replaced, n5=127.0.0.5, is filtered out and the replacing
node, n6=127.0.0.6 is in the pending list.

TODO:
- Handle paxos writes

### DIFF
--- a/locator/abstract_replication_strategy.hh
+++ b/locator/abstract_replication_strategy.hh
@@ -95,6 +95,7 @@ public:
                                               token_metadata& token_metadata,
                                               const std::map<sstring, sstring>& config_options);
     virtual std::vector<inet_address> get_natural_endpoints(const token& search_token);
+    virtual std::vector<inet_address> get_natural_endpoints_without_node_being_replaced(const token& search_token);
     virtual void validate_options() const = 0;
     virtual std::optional<std::set<sstring>> recognized_options() const = 0;
     virtual size_t get_replication_factor() const = 0;
--- a/locator/abstract_replication_strategy.hh
+++ b/locator/abstract_replication_strategy.hh
@@ -113,6 +113,10 @@ public:
     // It the analogue of Origin's getAddressRanges().get(endpoint).
     // This function is not efficient, and not meant for the fast path.
     dht::token_range_vector get_ranges(inet_address ep) const;
+
+    // Use the token_metadata provided by the caller instead of _token_metadata
+    dht::token_range_vector get_ranges(inet_address ep, token_metadata& tm) const;
+
     // get_primary_ranges() returns the list of "primary ranges" for the given
     // endpoint. "Primary ranges" are the ranges that the node is responsible
     // for storing replica primarily, which means this is the first node
--- a/locator/abstract_replication_strategy.hh
+++ b/locator/abstract_replication_strategy.hh
@@ -98,6 +98,12 @@ public:
     virtual void validate_options() const = 0;
     virtual std::optional<std::set<sstring>> recognized_options() const = 0;
     virtual size_t get_replication_factor() const = 0;
+    // Decide if the replication strategy allow removing the node being
+    // replaced from the natural endpoints when a node is being replaced in the
+    // cluster. LocalStrategy is the not allowed to do so because it always
+    // returns the node itself as the natural_endpoints and the node will not
+    // appear in the pending_endpoints.
+    virtual bool allow_remove_node_being_replaced_from_natural_endpoints() const = 0;
     uint64_t get_cache_hits_count() const { return _cache_hits_count; }
     replication_strategy_type get_type() const { return _my_type; }
 
--- a/locator/token_metadata.cc
+++ b/locator/token_metadata.cc
@@ -64,6 +64,8 @@ private:
 
     std::unordered_map<token, inet_address> _bootstrap_tokens;
     std::unordered_set<inet_address> _leaving_endpoints;
+    // The map between the existing node to be replaced and the replacing node
+    std::unordered_map<inet_address, inet_address> _replacing_endpoints;
 
     std::unordered_map<sstring, std::unordered_multimap<range<token>, inet_address>> _pending_ranges;
     std::unordered_map<sstring, std::unordered_map<range<token>, std::unordered_set<inet_address>>> _pending_ranges_map;
@@ -359,6 +361,15 @@ public:
 
     bool is_leaving(inet_address endpoint);
 
+    // Is this node being replaced by another node
+    bool is_being_replaced(inet_address endpoint);
+
+    // Is any node being replaced by another node
+    bool is_any_node_being_replaced();
+
+    void add_replacing_endpoint(inet_address existing_node, inet_address replacing_node);
+
+    void del_replacing_endpoint(inet_address existing_node);
 #if 0
     private final AtomicReference<TokenMetadata> cachedTokenMap = new AtomicReference<TokenMetadata>();
 #endif
@@ -487,6 +498,10 @@ public:
         abstract_replication_strategy& strategy,
         lw_shared_ptr<std::unordered_multimap<range<token>, inet_address>> new_pending_ranges,
         lw_shared_ptr<token_metadata> all_left_metadata);
+    future<> calculate_pending_ranges_for_replacing(
+        token_metadata& unpimplified_this,
+        abstract_replication_strategy& strategy,
+        lw_shared_ptr<std::unordered_multimap<range<token>, inet_address>> new_pending_ranges);
 public:
 
     token get_predecessor(token t);
@@ -1221,11 +1236,20 @@ bool token_metadata_impl::is_leaving(inet_address endpoint) {
     return _leaving_endpoints.count(endpoint);
 }
 
+bool token_metadata_impl::is_being_replaced(inet_address endpoint) {
+    return _replacing_endpoints.count(endpoint);
+}
+
+bool token_metadata_impl::is_any_node_being_replaced() {
+    return !_replacing_endpoints.empty();
+}
+
 void token_metadata_impl::remove_endpoint(inet_address endpoint) {
     remove_by_value(_bootstrap_tokens, endpoint);
     remove_by_value(_token_to_endpoint_map, endpoint);
     _topology.remove_endpoint(endpoint);
     _leaving_endpoints.erase(endpoint);
+    _replacing_endpoints.erase(endpoint);
     _endpoint_to_host_id_map.erase(endpoint);
     _sorted_tokens = sort_tokens();
     invalidate_cached_rings();
@@ -1399,6 +1423,31 @@ future<> token_metadata_impl::calculate_pending_ranges_for_leaving(
     });
 }
 
+future<> token_metadata_impl::calculate_pending_ranges_for_replacing(
+        token_metadata& unpimplified_this,
+        abstract_replication_strategy& strategy,
+        lw_shared_ptr<std::unordered_multimap<range<token>, inet_address>> new_pending_ranges) {
+    if (_replacing_endpoints.empty()) {
+        return make_ready_future<>();
+    }
+    return do_with(_replacing_endpoints, strategy.get_address_ranges(unpimplified_this),
+            [this, new_pending_ranges] (std::unordered_map<inet_address, inet_address>& replacing_endpoints,
+                    std::unordered_multimap<inet_address, dht::token_range>& address_ranges) {
+        return do_for_each(replacing_endpoints, [this, new_pending_ranges, &address_ranges]
+                (const std::pair<gms::inet_address, gms::inet_address>& node) {
+            auto existing_node = node.first;
+            auto replacing_node = node.second;
+            return do_for_each(address_ranges, [this, new_pending_ranges, existing_node, replacing_node]
+                    (const std::pair<gms::inet_address, dht::token_range>& x) {
+                if (x.first == existing_node) {
+                    tlogger.debug("Node {} replaces {} for range {}", replacing_node, existing_node, x.second);
+                    new_pending_ranges->emplace(x.second, replacing_node);
+                }
+            });
+        });
+    });
+}
+
 void token_metadata_impl::calculate_pending_ranges_for_bootstrap(
         abstract_replication_strategy& strategy,
         lw_shared_ptr<std::unordered_multimap<range<token>, inet_address>> new_pending_ranges,
@@ -1435,12 +1484,15 @@ future<> token_metadata_impl::calculate_pending_ranges(
         abstract_replication_strategy& strategy, const sstring& keyspace_name) {
     auto new_pending_ranges = make_lw_shared<std::unordered_multimap<range<token>, inet_address>>();
 
-    if (_bootstrap_tokens.empty() && _leaving_endpoints.empty()) {
-        tlogger.debug("No bootstrapping, leaving nodes -> empty pending ranges for {}", keyspace_name);
+    tlogger.debug("calculate_pending_ranges: keyspace_name={}, bootstrap_tokens={}, leaving nodes={}, replacing_endpoints={}",
+        keyspace_name, _bootstrap_tokens, _leaving_endpoints, _replacing_endpoints);
+    if (_bootstrap_tokens.empty() && _leaving_endpoints.empty() && _replacing_endpoints.empty()) {
+        tlogger.debug("No bootstrapping, leaving nodes, replacing nodes -> empty pending ranges for {}", keyspace_name);
         set_pending_ranges(keyspace_name, std::move(*new_pending_ranges));
         return make_ready_future<>();
     }
 
+  return calculate_pending_ranges_for_replacing(unpimplified_this, strategy, new_pending_ranges).then([&unpimplified_this, this, keyspace_name, &strategy, new_pending_ranges] {
     // Copy of metadata reflecting the situation after all leave operations are finished.
     auto all_left_metadata = make_lw_shared<token_metadata>(std::make_unique<token_metadata_impl>(clone_after_all_left()));
 
@@ -1457,6 +1509,7 @@ future<> token_metadata_impl::calculate_pending_ranges(
         }
         return make_ready_future<>();
     });
+  });
 
 }
 
@@ -1485,6 +1538,14 @@ sstring token_metadata_impl::print_pending_ranges() {
 
 void token_metadata_impl::add_leaving_endpoint(inet_address endpoint) {
      _leaving_endpoints.emplace(endpoint);
+}
+
+void token_metadata_impl::add_replacing_endpoint(inet_address existing_node, inet_address replacing_node) {
+    _replacing_endpoints[existing_node] = replacing_node;
+}
+
+void token_metadata_impl::del_replacing_endpoint(inet_address existing_node) {
+    _replacing_endpoints.erase(existing_node);
 }
 
 token_metadata_impl token_metadata_impl::clone_after_all_settled() {
@@ -1754,6 +1815,24 @@ token_metadata::is_member(inet_address endpoint) {
 bool
 token_metadata::is_leaving(inet_address endpoint) {
     return _impl->is_leaving(endpoint);
+}
+
+bool
+token_metadata::is_being_replaced(inet_address endpoint) {
+    return _impl->is_being_replaced(endpoint);
+}
+
+bool
+token_metadata::is_any_node_being_replaced() {
+    return _impl->is_any_node_being_replaced();
+}
+
+void token_metadata::add_replacing_endpoint(inet_address existing_node, inet_address replacing_node) {
+    _impl->add_replacing_endpoint(existing_node, replacing_node);
+}
+
+void token_metadata::del_replacing_endpoint(inet_address existing_node) {
+    _impl->del_replacing_endpoint(existing_node);
 }
 
 token_metadata
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -1844,7 +1844,7 @@ storage_proxy::create_write_response_handler_helper(schema_ptr s, const dht::tok
     auto keyspace_name = s->ks_name();
     keyspace& ks = _db.local().find_keyspace(keyspace_name);
     auto& rs = ks.get_replication_strategy();
-    std::vector<gms::inet_address> natural_endpoints = rs.get_natural_endpoints(token);
+    std::vector<gms::inet_address> natural_endpoints = rs.get_natural_endpoints_without_node_being_replaced(token);
     std::vector<gms::inet_address> pending_endpoints = _token_metadata.pending_endpoints_for(token, keyspace_name);
 
     slogger.trace("creating write handler for token: {} natural: {} pending: {}", token, natural_endpoints, pending_endpoints);
@@ -2156,8 +2156,8 @@ future<> storage_proxy::mutate_counters(Range&& mutations, db::consistency_level
 storage_proxy::paxos_participants
 storage_proxy::get_paxos_participants(const sstring& ks_name, const dht::token &token, db::consistency_level cl_for_paxos) {
     keyspace& ks = _db.local().find_keyspace(ks_name);
-    locator::abstract_replication_strategy& rs = ks.get_replication_strategy();
-    std::vector<gms::inet_address> natural_endpoints = rs.get_natural_endpoints(token);
+    auto& rs = ks.get_replication_strategy();
+    std::vector<gms::inet_address> natural_endpoints = rs.get_natural_endpoints_without_node_being_replaced(token);
     std::vector<gms::inet_address> pending_endpoints = _token_metadata.pending_endpoints_for(token, ks_name);
 
     if (cl_for_paxos == db::consistency_level::LOCAL_SERIAL) {
@@ -4500,7 +4500,7 @@ future<bool> storage_proxy::cas(schema_ptr schema, shared_ptr<cas_request> reque
 
 std::vector<gms::inet_address> storage_proxy::get_live_endpoints(keyspace& ks, const dht::token& token) const {
     auto& rs = ks.get_replication_strategy();
-    std::vector<gms::inet_address> eps = rs.get_natural_endpoints(token);
+    std::vector<gms::inet_address> eps = rs.get_natural_endpoints_without_node_being_replaced(token);
     auto itend = boost::range::remove_if(eps, std::not1(std::bind1st(std::mem_fn(&gms::gossiper::is_alive), &gms::get_local_gossiper())));
     eps.erase(itend, eps.end());
     return eps;
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -719,12 +719,6 @@ void storage_service::join_token_ring(int delay) {
     set_gossip_tokens(_bootstrap_tokens, _cdc_streams_ts);
     set_mode(mode::NORMAL, "node is now in normal status", true);
 
-    // remove the existing info about the replaced node.
-    if (!current.empty()) {
-        for (auto existing : current) {
-            _gossiper.replaced_endpoint(existing);
-        }
-    }
     if (_token_metadata.sorted_tokens().empty()) {
         auto err = format("join_token_ring: Sorted token in token_metadata is empty");
         slogger.error("{}", err);
@@ -1095,34 +1089,28 @@ void storage_service::handle_state_normal(inet_address endpoint) {
 
     std::unordered_set<inet_address> endpoints_to_remove;
 
+    auto do_remove_node = [&] (gms::inet_address node) {
+        _token_metadata.remove_endpoint(node);
+        endpoints_to_remove.insert(node);
+    };
     // Order Matters, TM.updateHostID() should be called before TM.updateNormalToken(), (see CASSANDRA-4300).
     if (_gossiper.uses_host_id(endpoint)) {
         auto host_id = _gossiper.get_host_id(endpoint);
         auto existing = _token_metadata.get_endpoint_for_host_id(host_id);
-        if (db().local().is_replacing() &&
-            db().local().get_replace_address() &&
-                _gossiper.get_endpoint_state_for_endpoint_ptr(db().local().get_replace_address().value())  &&
-            (host_id == _gossiper.get_host_id(db().local().get_replace_address().value()))) {
-            slogger.warn("Not updating token metadata for {} because I am replacing it", endpoint);
-        } else {
-            if (existing && *existing != endpoint) {
-                if (*existing == get_broadcast_address()) {
-                    slogger.warn("Not updating host ID {} for {} because it's mine", host_id, endpoint);
-                    _token_metadata.remove_endpoint(endpoint);
-                    endpoints_to_remove.insert(endpoint);
-                } else if (_gossiper.compare_endpoint_startup(endpoint, *existing) > 0) {
-                    slogger.warn("Host ID collision for {} between {} and {}; {} is the new owner", host_id, *existing, endpoint, endpoint);
-                    _token_metadata.remove_endpoint(*existing);
-                    endpoints_to_remove.insert(*existing);
-                    _token_metadata.update_host_id(host_id, endpoint);
-                } else {
-                    slogger.warn("Host ID collision for {} between {} and {}; ignored {}", host_id, *existing, endpoint, endpoint);
-                    _token_metadata.remove_endpoint(endpoint);
-                    endpoints_to_remove.insert(endpoint);
-                }
-            } else {
+        if (existing && *existing != endpoint) {
+            if (*existing == get_broadcast_address()) {
+                slogger.warn("Not updating host ID {} for {} because it's mine", host_id, endpoint);
+                do_remove_node(endpoint);
+            } else if (_gossiper.compare_endpoint_startup(endpoint, *existing) > 0) {
+                slogger.warn("Host ID collision for {} between {} and {}; {} is the new owner", host_id, *existing, endpoint, endpoint);
+                do_remove_node(*existing);
                 _token_metadata.update_host_id(host_id, endpoint);
+            } else {
+                slogger.warn("Host ID collision for {} between {} and {}; ignored {}", host_id, *existing, endpoint, endpoint);
+                do_remove_node(endpoint);
             }
+        } else {
+            _token_metadata.update_host_id(host_id, endpoint);
         }
     }
 
@@ -1175,10 +1163,6 @@ void storage_service::handle_state_normal(inet_address endpoint) {
 
     for (auto ep : endpoints_to_remove) {
         remove_endpoint(ep);
-        auto replace_addr = db().local().get_replace_address();
-        if (db().local().is_replacing() && replace_addr && *replace_addr == ep) {
-            _gossiper.replacement_quarantine(ep); // quarantine locally longer than normally; see CASSANDRA-8260
-        }
     }
     slogger.debug("handle_state_normal: endpoint={} owned_tokens = {}", endpoint, owned_tokens);
     if (!owned_tokens.empty()) {
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -958,7 +958,7 @@ void storage_service::bootstrap() {
     set_mode(mode::JOINING, "Starting to bootstrap...", true);
     if (is_repair_based_node_ops_enabled()) {
         if (db().local().is_replacing()) {
-            replace_with_repair(_db, _token_metadata).get();
+            replace_with_repair(_db, _token_metadata, _bootstrap_tokens).get();
         } else {
             bootstrap_with_repair(_db, _token_metadata, _bootstrap_tokens).get();
         }
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -288,11 +288,6 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
         }
 
         std::tie(_bootstrap_tokens, _cdc_streams_ts) = prepare_replacement_info(loaded_peer_features).get0();
-        // kbraun: I think we don't have to gossip tokens nor streams timestmap when HIBERNATEd
-        // (any state for this node will be ignored by other nodes anyway), but since TOKENS was already there, I added CDC_STREAMS_TIMESTAMP too.
-        app_states.emplace(gms::application_state::TOKENS, versioned_value::tokens(_bootstrap_tokens));
-        app_states.emplace(gms::application_state::CDC_STREAMS_TIMESTAMP, versioned_value::cdc_streams_timestamp(_cdc_streams_ts));
-        app_states.emplace(gms::application_state::STATUS, versioned_value::hibernate(true));
         auto replace_address = db().local().get_replace_address();
         replacing_a_node_with_same_ip = replace_address && *replace_address == get_broadcast_address();
         replacing_a_node_with_diff_ip = replace_address && *replace_address != get_broadcast_address();
@@ -392,7 +387,12 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
         ss._local_host_id = local_host_id;
     }).get();
     auto features = get_config_supported_features();
-    _token_metadata.update_host_id(local_host_id, get_broadcast_address());
+    if (!replacing_a_node_with_diff_ip) {
+        // Replacing node with a different ip should own the host_id only after
+        // the replacing node becomes NORMAL status. It is updated in
+        // handle_state_normal().
+        _token_metadata.update_host_id(local_host_id, get_broadcast_address());
+    }
 
     // Replicate the tokens early because once gossip runs other nodes
     // might send reads/writes to this node. Replicate it early to make
@@ -937,9 +937,17 @@ void storage_service::bootstrap() {
         set_mode(mode::JOINING, format("sleeping {} ms for pending range setup", get_ring_delay().count()), true);
         _gossiper.wait_for_range_setup().get();
     } else {
-        // Dont set any state for the node which is bootstrapping the existing token...
-        _token_metadata.update_normal_tokens(_bootstrap_tokens, get_broadcast_address());
-        replicate_to_all_cores().get();
+        // Wait until we know tokens of existing node before announcing replacing status.
+        set_mode(mode::JOINING, sprint("Wait until local node knows tokens of peer nodes"), true);
+        _gossiper.wait_for_range_setup().get();
+        set_mode(mode::JOINING, sprint("Announce tokens and status of the replacing node"), true);
+        _gossiper.add_local_application_state({
+            { gms::application_state::TOKENS, versioned_value::tokens(_bootstrap_tokens) },
+            { gms::application_state::CDC_STREAMS_TIMESTAMP, versioned_value::cdc_streams_timestamp(_cdc_streams_ts) },
+            { gms::application_state::STATUS, versioned_value::hibernate(true) },
+        }).get();
+        set_mode(mode::JOINING, format("Wait until peer nodes know the bootstrap tokens of local node"), true);
+        _gossiper.wait_for_range_setup().get();
         auto replace_addr = db().local().get_replace_address();
         if (replace_addr) {
             slogger.debug("Removing replaced endpoint {} from system.peers", *replace_addr);
@@ -1041,6 +1049,27 @@ storage_service::get_range_to_address_map(const sstring& keyspace,
     return construct_range_to_endpoint_map(ks, get_all_ranges(sorted_tokens));
 }
 
+void storage_service::handle_state_replacing(inet_address replacing_node) {
+    slogger.debug("endpoint={} handle_state_replacing", replacing_node);
+    auto host_id = _gossiper.get_host_id(replacing_node);
+    auto existing_node_opt = _token_metadata.get_endpoint_for_host_id(host_id);
+    auto replace_addr = db().local().get_replace_address();
+    if (replacing_node == get_broadcast_address() && replace_addr && *replace_addr == get_broadcast_address()) {
+        existing_node_opt = replacing_node;
+    }
+    if (!existing_node_opt) {
+        slogger.warn("Can not find the existing node for the replacing node {}", replacing_node);
+        return;
+    }
+    auto existing_node = *existing_node_opt;
+    auto existing_tokens = get_tokens_for(existing_node);
+    auto replacing_tokens = get_tokens_for(replacing_node);
+    slogger.info("Node {} is replacing existing node {} with host_id={}, existing_tokens={}, replacing_tokens={}",
+            replacing_node, existing_node, host_id, existing_tokens, replacing_tokens);
+    _token_metadata.add_replacing_endpoint(existing_node, replacing_node);
+    update_pending_ranges().get();
+}
+
 void storage_service::handle_state_bootstrap(inet_address endpoint) {
     slogger.debug("endpoint={} handle_state_bootstrap", endpoint);
     // explicitly check for TOKENS, because a bootstrapping node might be bootstrapping in legacy mode; that is, not using vnodes and no token specified
@@ -1104,12 +1133,19 @@ void storage_service::handle_state_normal(inet_address endpoint) {
             } else if (_gossiper.compare_endpoint_startup(endpoint, *existing) > 0) {
                 slogger.warn("Host ID collision for {} between {} and {}; {} is the new owner", host_id, *existing, endpoint, endpoint);
                 do_remove_node(*existing);
+                slogger.info("Set host_id={} to be owned by node={}, existing={}", host_id, endpoint, *existing);
                 _token_metadata.update_host_id(host_id, endpoint);
+                slogger.info("Remove node {} from pending replacing endpoint", endpoint);
+                _token_metadata.del_replacing_endpoint(endpoint);
             } else {
                 slogger.warn("Host ID collision for {} between {} and {}; ignored {}", host_id, *existing, endpoint, endpoint);
                 do_remove_node(endpoint);
             }
+        } else if (existing && *existing == endpoint) {
+            slogger.info("Remove node {} from pending replacing endpoint", endpoint);
+            _token_metadata.del_replacing_endpoint(endpoint);
         } else {
+            slogger.info("Set host_id={} to be owned by node={}", host_id, endpoint);
             _token_metadata.update_host_id(host_id, endpoint);
         }
     }
@@ -1368,6 +1404,8 @@ void storage_service::on_change(inet_address endpoint, application_state state, 
             handle_state_left(endpoint, pieces);
         } else if (move_name == sstring(versioned_value::STATUS_MOVING)) {
             handle_state_moving(endpoint, pieces);
+        } else if (move_name == sstring(versioned_value::HIBERNATE)) {
+            handle_state_replacing(endpoint);
         } else {
             return; // did nothing.
         }
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -277,6 +277,8 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
     if (get_replace_tokens().size() > 0 || get_replace_node()) {
          throw std::runtime_error("Replace method removed; use replace_address instead");
     }
+    bool replacing_a_node_with_same_ip = false;
+    bool replacing_a_node_with_diff_ip = false;
     if (db().local().is_replacing()) {
         if (db::system_keyspace::bootstrap_complete()) {
             throw std::runtime_error("Cannot replace address with a node that is already bootstrapped");
@@ -291,6 +293,9 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
         app_states.emplace(gms::application_state::TOKENS, versioned_value::tokens(_bootstrap_tokens));
         app_states.emplace(gms::application_state::CDC_STREAMS_TIMESTAMP, versioned_value::cdc_streams_timestamp(_cdc_streams_ts));
         app_states.emplace(gms::application_state::STATUS, versioned_value::hibernate(true));
+        auto replace_address = db().local().get_replace_address();
+        replacing_a_node_with_same_ip = replace_address && *replace_address == get_broadcast_address();
+        replacing_a_node_with_diff_ip = replace_address && *replace_address != get_broadcast_address();
     } else if (should_bootstrap()) {
         check_for_endpoint_collision(loaded_peer_features).get();
     } else {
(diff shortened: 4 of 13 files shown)


asias:
Changes in 4c6ecbd0b475dcb289f1d159ef96ad67cd78c8bb:
- Rebase to master, resolving conflicts due to avi's pimp of token_metadata change.

slivne:
@asias is this ready for review ?

asias:
> @asias is this ready for review ?

I need some help for the paxos part. Other parts in the patch are ready for review.

asias:
@gleb-cloudius Can you please take a look at the storage proxy part?

slivne:
@gleb-cloudius ping

asias:
Ping for review and comments.

slivne:
@gleb-cloudius ping

slivne:
@gleb-cloudius ping

glommer:
Folks - this is sitting here for almost two months!
I will review the storage proxy part myself tomorrow if nobody gets to it (late here now)

In the meantime, @kostja , asias claims to need help understanding how to hook up the paxos part.
Can you please take a look ?

dorlaor:
:(

kostja:
@glommer bad excuse, but I am not really seeing pull requests - too much noise to be able to find them when they are submitted. Using a mailing list Ccing the right person usually works much better.

glommer:
Thanks Kostja.

service/storage_service.cc
```diff
@@ -315,6 +317,9 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
         app_states.emplace(gms::application_state::TOKENS, value_factory.tokens(_bootstrap_tokens));
         app_states.emplace(gms::application_state::CDC_STREAMS_TIMESTAMP, value_factory.cdc_streams_timestamp(_cdc_streams_ts));
         app_states.emplace(gms::application_state::STATUS, value_factory.hibernate(true));
+        auto replace_address = db().local().get_replace_address();
+        replacing_a_node_with_same_ip = replace_address && *replace_address == get_broadcast_address();
+        replacing_a_node_with_diff_ip = replace_address && *replace_address != get_broadcast_address();
```
glommer, line 317: Hi @asias.

The explanation you give in the changelog is not enough for a person who doesn't understand this code that well (me) to understand why we really need this differentiation.

It feels weird to me that we have to make this distinction. What would be the effect of us not differentiating between nodes with the same IP and different IP?

auth/service.cc
```diff
@@ -157,7 +157,7 @@ future<> service::create_keyspace_if_missing(::service::migration_manager& mm) c
     auto& db = _qp.db();
 
     if (!db.has_keyspace(meta::AUTH_KS)) {
-        std::map<sstring, sstring> opts{{"replication_factor", "1"}};
+        std::map<sstring, sstring> opts{{"replication_factor", "3"}};
```
glommer, line 160: This may not work well with clusters with fewer than 3 nodes.
It's also a problem that will happen for any user keyspace that has RF=1, no ?

I see the problem but I think this is not the right way to go. I will try to suggest something

locator/token_metadata.cc
```diff
@@ -1399,6 +1423,31 @@ future<> token_metadata_impl::calculate_pending_ranges_for_leaving(
     });
 }
 
+future<> token_metadata_impl::calculate_pending_ranges_for_replacing(
+        token_metadata& unpimplified_this,
```
glommer, line 1427: `unpimplified_this` is a horrible name. I don't even know what it is supposed to mean

service/storage_proxy.cc
```diff
@@ -1750,6 +1750,31 @@ storage_proxy::create_write_response_handler_helper(schema_ptr s, const dht::tok
     std::vector<gms::inet_address> natural_endpoints = rs.get_natural_endpoints(token);
     std::vector<gms::inet_address> pending_endpoints = _token_metadata.pending_endpoints_for(token, keyspace_name);
 
+    if (get_local_storage_service().get_token_metadata().is_any_node_being_replaced() &&
+        rs.get_type() != locator::replication_strategy_type::local) {
```
glommer, line 1754: how is this different from any user keyspace with RF=1 ?

locator/token_metadata.cc
```diff
@@ -64,6 +64,8 @@ class token_metadata_impl final {
 
     std::unordered_map<token, inet_address> _bootstrap_tokens;
     std::unordered_set<inet_address> _leaving_endpoints;
+    // The map between the existing node to be replaced and the replacing node
+    std::unordered_map<inet_address, inet_address> _replacing_endpoints;
```
gleb-cloudius, line 68: Can there be more than one simultaneously?

locator/token_metadata.cc
```diff
@@ -1399,6 +1423,31 @@ future<> token_metadata_impl::calculate_pending_ranges_for_leaving(
...
+            return do_for_each(address_ranges, [this, new_pending_ranges, existing_node, replacing_node]
+                    (const std::pair<gms::inet_address, dht::token_range>& x) {
+                if (x.first == existing_node) {
+                    tlogger.debug("Node {} replaces {} for range {}", replacing_node, existing_node, x.second);
+                    new_pending_ranges->emplace(x.second, replacing_node);
+                }
+            });
+        });
+    });
+}
```
gleb-cloudius, line 1449: I do not see that this function use futures internally, so first it can return `void` and second it can be simplified using `for()` loops. `do_with()` is also not needed.

service/storage_proxy.cc
```diff
@@ -1750,6 +1750,31 @@ storage_proxy::create_write_response_handler_helper(schema_ptr s, const dht::tok
...
+        // otherwise all the query to keyspace with LocalStrategy will fail,
+        // e.g., the system keyspace.
+        auto original_natural_endpoints = natural_endpoints;
+        auto it = boost::range::remove_if(natural_endpoints, [] (gms::inet_address& p) {
+            return get_local_storage_service().get_token_metadata().is_being_replaced(p);
+        });
+        natural_endpoints.erase(it, natural_endpoints.end());
+        if (original_natural_endpoints.size() != natural_endpoints.size()) {
+            slogger.trace("creating write handler for token: {}, ks={}, original_natrual={}, natural={}, pending={}",
+                    token, keyspace_name, original_natural_endpoints, natural_endpoints, pending_endpoints);
```
gleb-cloudius, line 1774: Why log twice here and below? But more importantly I do not understand what is going on here. Weren't node that is been replaced should have been filtered on gossiper level? get_natural_endpoints  does not suppose to return it.

auth/service.cc
gleb-cloudius: I do not even see the problem

gms/gossiper.cc
```diff
@@ -2173,6 +2173,11 @@ future<> gossiper::wait_for_range_setup() {
     return wait_for_gossip(ring_delay);
 }
 
+void gossiper::set_node_to_be_replaced(gms::inet_address node) {
+    logger.info("Set node to be replaced = {}", node);
+    _node_to_be_replaced = node;
```
gleb-cloudius, line 2178: Where is this used? github gui is so bad I cannot find it.

gleb-cloudius:
I reviewed, but someone who understand gossiper more need to look as well.

locator/token_metadata.cc
asias: It is invented by avi when impl token_metadata.

locator/token_metadata.cc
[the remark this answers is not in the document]
asias: We will when we support replacing nodes in parallel. The work is in progress.

locator/token_metadata.cc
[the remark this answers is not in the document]
asias: It is used to avoid stalls. We see a lot of them in the past.

service/storage_proxy.cc
asias: I guess you forget what we have discussed: https://github.com/scylladb/scylla/issues/5482#issuecomment-576165303

asias: It is mainly used for debug for this early version. I can drop it.

gms/gossiper.cc
asias: It is used in storage_service: Update tokens and replace address for replace operation.

service/storage_proxy.cc
asias: With local strategy, the replacing node will not be in the pending_endpoints. 
With other strategy, the replacing node will be in the pending_endpoints. 

If we filter the replacing node out of natural_endpoints for local strategy, no node will be found for the write.

auth/service.cc
asias: The problem is before this patch, we allow replace operation with RF = 1, which we lose data. After this patch, we correctly detect that a source node can not be found for keyspace with RF = 1.

service/storage_service.cc
asias: The key point is we want the replacing node has the same view of the token ring as the result of the cluster. In case the replacing node is started with a new ip, the result of the cluster will put the new ip into the token_metadata (ip token map) after the replacing is done. In case the replacing node is started with the same IP ad the replaced node, the rest of the cluster as the old ip in token_metadata, so should the replacing node.

auth/service.cc
gleb-cloudius: It is practically the same as disallowing RF=1. With RF=1 there is not HA, so whoever uses it may be well aware of it.

gms/gossiper.cc
gleb-cloudius: I see it is only set there, never used. If it is indeed used anywhare sending a proper patchset to a mailing list would made it much easier for me to find.

locator/token_metadata.cc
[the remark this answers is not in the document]
gleb-cloudius: As long as it will not create a situation where more than one pending node exists for a range it should be fine I guess.

locator/token_metadata.cc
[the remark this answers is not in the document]
gleb-cloudius: It is a pity we need to obfuscate code like this just to avoid stalls. How often does this run? Can it run in a thread?

gms/gossiper.cc
asias: The actual user is in https://github.com/scylladb/scylla/pull/5154. I got some of the patches from that PR because they all are handling the replacing operations.  I do not want this PR to depend on 5154.

service/storage_proxy.cc
gleb-cloudius: I did not read it as literal. What is the point to do filtering in sp each time get_natural_endpoints is called? Do it once in 
get_natural_endpoints. Otherwise you will have to do it for all its callers. What about view building code for instance? Did you handle it here?

locator/token_metadata.cc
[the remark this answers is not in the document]
asias: It should not. Because the replaced nodes are supposed to have different ranges.

auth/service.cc
asias: As I mentioned in commit log in auth: Increase system_auth from RF 1 to 3
    
    TODO:
    - Handle existing cluster already has system_auth with RF 1.
    - We may need a way to allow replace operation with keyspace with RF 1
      if user allows to lost the data for that keyspace.

The question is  what do we in case there is a RF 1 keyspace and the source node is dead. Before we silently ignore in replacing operation. What should we do now?

service/storage_proxy.cc
asias: I think it is wrong to filter inside get_natural_endpoints. What about the other callers of get_natural_endpoints that does not in the proxy write path. For example, the api in  api/api-doc/storage_service.json, repair to get neighbour in repair/repair.cc. 

The dead node to be replaced is still part of the ring. It should be in the get_natural_endpoints until we finish the replacing operation.

asias: Can you elaborate what should be done for the view building code? Is it in proxy service/storage_proxy.cc?

locator/token_metadata.cc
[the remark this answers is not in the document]
asias: It is called everything a topology is changed for each keyspace.  Make it running in a thread will require refactoring all the way to the caller.  I do not want to do the refactor in this series.

service/storage_proxy.cc
gleb-cloudius: Why would you want to get a dead and never comming back node in repair/repair.cc (can we even run repair in parallel with replace?)? Same with API. Why not return replacing node instead since it is the range owned now. Without knowing what the API is used for it is hard to tell what is best there.

And view building is the same as a write path. Just cl=one write. So I do not see why it is not suppose to be handled the same.

asias: Nothing says the node will never come back. The user can abort the replace operation and bring back the dead node back. 

The API is here:
```
 472          "path":"/storage_service/natural_endpoints/{keyspace}",
 473          "operations":[
 474             {
 475                "method":"GET",
 476                "summary":"This method returns the N endpoints that are responsible for storing the specified key i.e for replication. the endpoint responsible for this key",
 477                "type":"array",
 478                "items":{
 479                   "type":"string"
 480                },
```

Is the dead node still part of the ring when replacing? Yes. 
Should we tell user that the node being replaced is not part of the ring? No.

Can you point to me the relevant code for view build code?

gleb-cloudius: So report the old ring topology here by using some other function or introduce some other function to use everywhere else, though I would say that while replacing the dead node is not longer responsible for a range since it is dead. What about reapair code? Can it even run during replacement?

For view related code search for all uses of get natural and pending endpoints and look at those in view related files. I see some in `db/view/view.cc`. And if all writing code has to do the same than it has to be done in one place.

gms/gossiper.cc
gleb-cloudius: First of all it is good that I am not going crazy by not seeing its use here. But why not introduce it in the series it is used in than?

auth/service.cc
glommer: What's the problem with losing data during a replace with RF=1 ?
I see no other way. The problem I see here is that we're trying to handle auth differently. I am not sure it should be handled differently for the purpose of this patch.

I have been saying for a while that auth should use EverywhereStrategy (which we don't have), to automatically shrink and grow so this becomes automatic. But until that's done, why treat auth differently ? If you had it in RF=1 you lose data upon a lost node, same as you would with any RF=1 keyspace.

locator/token_metadata.cc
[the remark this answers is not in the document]
gleb-cloudius: A range becomes pending not only for primary replicas, but secondaries as well. IOW if you replace one node in 3 node cluster all ranges will have pending node and if you will try to replace two they will have two pending nodes which will not work. I think we are too far from been able to support more that one pending node. LWT outright check for that and throws an error. So lest code for one replacing node (at least per DC) for now.

locator/token_metadata.cc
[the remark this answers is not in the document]
gleb-cloudius: You can start a thread at the lowest level if it is rare operation.

gms/gossiper.cc
asias: I can get rid of this patch from this series to reduce the noise.

service/storage_proxy.cc
```diff
@@ -1750,6 +1750,31 @@ storage_proxy::create_write_response_handler_helper(schema_ptr s, const dht::tok
...
+        // However, we can not apply the filter for LocalStrategy because
+        // LocalStrategy always has RF = 1 and returns the node itself as the
+        // natural_endpoints. If the replacing node had the same ip address as
+        // the node being replaced, we can not filter out the local node,
+        // otherwise all the query to keyspace with LocalStrategy will fail,
+        // e.g., the system keyspace.
+        auto original_natural_endpoints = natural_endpoints;
+        auto it = boost::range::remove_if(natural_endpoints, [] (gms::inet_address& p) {
+            return get_local_storage_service().get_token_metadata().is_being_replaced(p);
+        });
```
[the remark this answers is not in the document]
asias: What do you mean by confused with ip? What do you suggest we do?  What do you disagree with this approach?

Here is what Gleb and I discussed and agreed on:

https://github.com/scylladb/scylla/issues/5482#issuecomment-576165303

service/storage_proxy.cc
asias: sorage_proxy..c is the only place needs this filtering out replacing node thing. I think it is better to introduce a new function like get_natural_endpoints_without_node_being_replaced. Use this new function in  storage_proxy. 

We do not reject repair during replace operation. Two cases: 1) the replacing node is using the same IP address, the repair would repair with the node being replaced, should be ok 2) if the replacing node is using a different IP, the repair would repair with the old dead node, the repair with ignore the dead node.

service/storage_service.cc
```diff
@@ -397,6 +402,21 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
         }
     }
 
+    if (replacing_a_node_with_same_ip) {
+        slogger.info("Replacing a node with same IP address, my address={}, node being replaced={}",
+                    get_broadcast_address(), get_broadcast_address());
+        slogger.info("Update tokens for replacing node early, replacing node has the same IP address of the node being replaced");
+        _token_metadata.update_normal_tokens(_bootstrap_tokens, get_broadcast_address());
```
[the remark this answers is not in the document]
asias: We discussed this in the https://github.com/scylladb/scylla/pull/5154. The idea is to make the replacing node has the same  view of the token ring as the rest of the cluster.

service/storage_service.cc
```diff
@@ -1116,11 +1110,24 @@ void storage_service::handle_state_normal(inet_address endpoint) {
     if (_gossiper.uses_host_id(endpoint)) {
         auto host_id = _gossiper.get_host_id(endpoint);
         auto existing = _token_metadata.get_endpoint_for_host_id(host_id);
+        auto replace_addr = db().local().get_replace_address();
         if (db().local().is_replacing() &&
```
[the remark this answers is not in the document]
asias: Because existing code handles replacing node differently. I thought it is easier to handle them in different code path. We probably can combine most of the logic. I will work on it.

service/storage_proxy.cc
asias: > Did you look into the use of pending endpoint in the view code?

I haven't.  I wanted to settle down a version without considering MV first.

> About repair and replacing simultaneously in the first case two different streaming operation will run. Is it even supported? Tested? Also it will stream the same data, so even if it will work this is not something that should be done.

We do not have this test in dtest or sct. Yes, I agree the repair is not helpfu in this case.

gleb-cloudius: > > Did you look into the use of pending endpoint in the view code?
> 
> I haven't. I wanted to settle down a version without considering MV first.
> 
How can you say that storage_proxy is the only place that needs it than?

service/storage_service.cc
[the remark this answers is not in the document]
tgrabiec: Yes, but it's not clear to the view would not be the same without this change.

service/storage_proxy.cc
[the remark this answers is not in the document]
tgrabiec: I mentioned this problem in the previous comment: https://github.com/scylladb/scylla/pull/5731#discussion_r403124032

"Node being confused with ip" refers to the fact that we have two different nodes here, replaced and replacing, but both are identified in the system as one thing because they share the IP. That's why you need to filter natural endpoints, assuming that the IP in the natural endpoints refers to the replaced node, and assume that the IP in the pending endpoints refers to the replacing node. If those two nodes wouldn't share the identified, you wouldn't have to filter natural endpoints.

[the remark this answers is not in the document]
tgrabiec: It would be better if we didn't identify nodes by IP, but something else which uniquely identifies them:

```
class endpoint {
    class impl;
    lw_shared_ptr<impl> _impl;
public:
    gms::inet_address ip() const;
    UUID host_id() const;
    bool is_up() const;
    bool operator==(const endpoint&) const;
};
```

Could we use hostID to distinguish replacing from replaced?

service/storage_proxy.cc
tgrabiec: > sorage_proxy..c is the only place needs this filtering out replacing node thing. I think it is better to introduce a new function like get_natural_endpoints_without_node_being_replaced. Use this new function in storage_proxy.
> 
> We do not reject repair during replace operation. Two cases: 1) the replacing node is using the same IP address, the repair would repair with the node being replaced, should be ok 2) if the replacing node is using a different IP, the repair would repair with the old dead node, the repair with ignore the dead node.

Repair should not ignore the dead node. If it cannot repair every node then it should fail. In both cases above it should fail. In case 1, the repair should not succeed because replacing may be aborted and the old replica set is not repaired.

asias:
I pushed a new branch.  View update is not considered yet in this patch set.  

Changes in 2983627583bf236dc7175bab63fc18b1cb4873f6:

- Rebased to master
- Added get_natural_endpoints_without_node_being_replaced helper
- Simplified handle_normal_state for replacing node
- Adjusted replace_with_repair since the updating of tokens for replacing node are changed
- Dropped patch "auth: Increase system_auth from RF 1 to 3" in this series
- Dropped patch "gossiper: Add set_node_to_be_replaced" in this series

service/storage_proxy.cc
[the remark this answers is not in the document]
asias: No. Replacing node uses the same host id and tokens of the replaced node. Replacing node could use a different IP address or same IP address.

service/storage_proxy.cc
asias: The repair will fail because it is a partial repair, some of the nodes are down. Repair offers best effort, i.e. continue to repair even if some of the repair followers are down. In the end, it will report the repair failed.

auth/service.cc
asias: The question is should we silently make the replace operation succeed with RF = 1 (losing data) or fail the replace operation in this case and ask user to decide what to do. 

For example, user might want to fix the dead node to avoid losing the RF data.

In the new version, I dropped the "auth: Increase system_auth from RF 1 to 3" patch.

locator/token_metadata.cc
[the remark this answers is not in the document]
asias: Yes, it will take a while to support replacing with multiple nodes. It is already coded as supporting multiple replacing node in token_metadata. It is not much more complicated than supporting a single node.  Btw, We have this multiple nodes support in token_metadata for bootstrap too, std::unordered_map<token, inet_address> _bootstrap_tokens.

service/storage_service.cc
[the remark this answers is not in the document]
asias: because the replacing node adds the tokens before the replacing operation is done while the other nodes adds the tokens after the replacing operation is done.

service/storage_proxy.cc
asias: Well, it is nicer to hide it in abstract_replication_strategy. But let's have a basic version that works and we agree on, then go for the nicer thing.

slivne:
@gleb-cloudius / @glommer / @tgrabiec 
   - can we please do a re-review.
  - we are still missing the "abstract_replication_strategy::get_natural_endpoint_without_pending()" that tomek requested, yet it would be good to understand if all the rest are handled like you wanted.

service/storage_service.cc
```diff
@@ -392,7 +387,12 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
         ss._local_host_id = local_host_id;
     }).get();
     auto features = get_config_supported_features();
-    _token_metadata.update_host_id(local_host_id, get_broadcast_address());
+    if (!replacing_a_node_with_diff_ip) {
+        // Replacing node with a different ip should own the host_id only after
+        // the replacing node becomes NORMAL status. It is updated in
+        // handle_state_normal().
+        _token_metadata.update_host_id(local_host_id, get_broadcast_address());
```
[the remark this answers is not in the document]
asias: Host id is the identify of a node. With replacing operation, the node identified by the host id is replaced by a new node. From the rest of the cluster's view, the node has no different before or after the replace operation. 

If the replacing node uses a different host id, the rest of the cluster can not know who the replacing node is replacing with. Look at storage_service::handle_state_normal how we use the host id to figure out the node becomes NORMAL is replacing another node.

service/storage_service.cc
```diff
@@ -392,7 +387,12 @@ void storage_service::prepare_to_join(std::vector<inet_address> loaded_endpoints
         ss._local_host_id = local_host_id;
     }).get();
     auto features = get_config_supported_features();
-    _token_metadata.update_host_id(local_host_id, get_broadcast_address());
+    if (!replacing_a_node_with_diff_ip) {
+        // Replacing node with a different ip should own the host_id only after
+        // the replacing node becomes NORMAL status. It is updated in
+        // handle_state_normal().
```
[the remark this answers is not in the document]
asias: We set the host_id in gossip a few lines below. We just can't set it in _token_metadata yet because the host_id is still owned by the old dead node as the result of the cluster sees.

service/storage_proxy.cc
glommer: Agree with tgrabiec.

service/storage_service.cc
```diff
@@ -1104,12 +1133,21 @@ void storage_service::handle_state_normal(inet_address endpoint) {
...
                 _token_metadata.update_host_id(host_id, endpoint);
+                slogger.info("Remove node {} from pending replacing endpoint", endpoint);
+                _token_metadata.del_replacing_endpoint(endpoint);
             } else {
                 slogger.warn("Host ID collision for {} between {} and {}; ignored {}", host_id, *existing, endpoint, endpoint);
                 do_remove_node(endpoint);
             }
+        } else if (existing && *existing == endpoint) {
+            slogger.info("Set host_id={} to be owned by node={}, case2", host_id, endpoint);
+            _token_metadata.update_host_id(host_id, endpoint);
```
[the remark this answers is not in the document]
asias: Yes, in this case the update_host_id() does not change anything. I kept it there to make the code look the same for all the 3 cases. I can remove it.

locator/token_metadata.cc
```diff
@@ -64,6 +64,8 @@ class token_metadata_impl final {
 
     std::unordered_map<token, inet_address> _bootstrap_tokens;
```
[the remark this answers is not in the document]
asias: It is part of the fixes for 5482. If you prefer, I can use Refs instead for other commits in the series and only use fix for the last patch.

auth/service.cc
glommer: Asias, I thought about this a bit more and I think I matured my preference now.

If we think about it, the replace operation does not "lose data" when RF=1. Data is already lost. There is no need to fail the replace operation, or warn, or anything.

If we want to be nice and somehow warn users about RF=1, then this should be done at create / alter time. Replace is not the place for it.

The auth thing is a problem indeed because it is a keyspace we create, not the user. Ideally auth would have Everywhere replication but I don't want to make this patch depend on @tgrabiec work on that area.

However as much as this sucks we're making this no worse than before, and our documentation does claim that users using auth should increase the RF. I don't want to trust them to do it, but I don't want to special case anything here either.

asias:
Changes in 1ab2f55c07afc56c2e60519b50b75cc75d6ce5b3 (v4):
- Do not check local strategy in storage_proxy
- Do not call _token_metadata.update_host_id when existing == endpoint
- Use Refs instead of Fixes for some of the commits
- Rebase to new master branch

service/storage_service.cc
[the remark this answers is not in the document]
tgrabiec: But, it is a different node in reality. The fact that we try to reuse the same host id on a different node generates special cases all around. Like in this hunk.

So I wonder, why not use a different host id like in bootstrap. The end result would be the same as if you decomissioned and then added a new node with the same tokens.

The rest of the cluster can know that we're replacing from the fact that tokens collide.

locator/token_metadata.cc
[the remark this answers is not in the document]
tgrabiec: Yes, I'd prefer that. 

"Fixes" here means that having this commit alone (and all its dependencies) fixes the issue. If it's part of the fix but is just a supporting cleanup, it doesn't fix. We may decide to backport just this commit, and it doesn't mean that the backport fixes the issue in the target branch.

service/storage_service.cc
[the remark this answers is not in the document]
tgrabiec: So token_metadata only has to know about host ids of tokens which form the old ring? We don't have to have a host if for pending or replacing tokens?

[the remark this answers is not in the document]
asias: The pending and replacing tokens are the tokens of the node being replaced. The host id to ip mapping is still being host id <-> old_ip during the replace operation.

service/storage_service.cc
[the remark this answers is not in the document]
asias: This colliding can happen if we intend to bootstrap a node but unfortunately it generates tokens which belongs to an existing node. The other nodes would think it is replacing  while it is not. I think it is better to be explicit. 

Also, replacing node uses the same host id is something existing Cassandra user would expect.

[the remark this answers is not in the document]
asias: And, when the existing node sees the replacing node advertises to be replacing status, the existing nodes needs to look at the host id of the replacing node to know which node this replacing node is replacing with.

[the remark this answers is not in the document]
glommer: Long term when we move to multi-node topology change operations I think replace should not exist, and just be a special case of remove + add. For instance if you have the cluster A, B, C, you can commit to add D E and F and then {A, B, C} -> {A, B, C, D, F}

If C is dead this is just a special case where you do {A, B, C} -> {A, B, D, F}. So at least in theory I am in favor of what Tomek suggests that maybe we should just handle this case now: let the node enters with a different ID, and we can just figure out a way to accept the fact that it will reuse the same tokens. If you think about it: bootstrap nodes can take writes, so we'll get that for free.

Can we maybe move token selection to a coordinator node instead of allowing the new node to propose its own tokens?  That's how multi-node would do anyway. If the coordinator says it is fine that the token collide, then it is fine that the tokens collide.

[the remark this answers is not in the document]
tgrabiec: > This colliding can happen if we intend to bootstrap a node but unfortunately it generates tokens which belongs to an existing node. The other nodes would think it is replacing while it is not. I think it is better to be explicit.

The bootstrapping node should not chose such tokens in the first place.

If we still want to catch this, we can look at the status of the new node and allow only "hibernate" to replace.


> And, when the existing node sees the replacing node advertises to be replacing status, the existing nodes needs to look at the host id of the replacing node to know which node this replacing node is replacing with.

This can be derived from token metadata, you replace the node which owns any token you replace in the old ring.


> 
> Also, replacing node uses the same host id is something existing Cassandra user would expect.

Do you know of any good reason a user may expect that?

I'm interested in this topic not only for this PR. It will make generic topology changes easier.

I think we should switch to identifying nodes by host id so that we don't have to special case the case of replacing a node with the same ip.

Having two nodes in the system with the same host id precludes that, because host id is still not unique (which is ironic).

If users wants the same id after replace, it could be an extra step after whole operation where you rename a node, but after you remove the old id from the system altogether. So host id is still unique. It'd be better to avoid this, still.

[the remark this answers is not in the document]
glommer: The direction we are moving towards, with multi-add, is something that departs from Cassandra abruptly anyway.

What Cassandra users expect in this case is not important. We just have to document it.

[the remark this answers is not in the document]
asias: When looking at user reported cluster issue, I usually check nodetool status to see if the host_id is owned by the ip address of the new replacing node. If that is the case, I know it is a  replace operation. If the host id changes, it is a bootstrap operation.  
Sometimes, due to operation error, we might find the replacing node uses the same host id but with new and different tokens. If the replacing node would use different host_id, we do not even know it is supposed to be an replace operation but ended up using wrong tokens.


The mapping between host id and tokens never changes. It is a very good property in my view. The host id is added and removed from the cluster when new nodes are added or removed from the cluster. The whole code base assumes this. Do you have a very good reason to change this?

@tgrabiec Can you please point out what are the special cases you are talking about  "The fact that we try to reuse the same host id on a different node generates special cases all around" beyond  this
```
+    if (!replacing_a_node_with_diff_ip) {
+        // Replacing node with a different ip should own the host_id only after
+        // the replacing node becomes NORMAL status. It is updated in
+        // handle_state_normal().
+        _token_metadata.update_host_id(local_host_id, get_broadcast_address());
+    }
```

This special case is needed to avoid more special cases to happen (handling replacing of the dead node differently on the replacing node than the rest of the cluster)

Look how unified  it is with this series: 

```
1) node6 (replacing node)
INFO  2020-04-24 09:36:34,331 [shard 0] storage_service - Set host_id=159d50be-2e2f-41c5-83ec-06d0e02681f8 to be owned by node=127.0.0.2
INFO  2020-04-24 09:36:34,331 [shard 0] storage_service - Set host_id=a855d88a-41ca-47ac-857e-a0ceb41adc6c to be owned by node=127.0.0.3
INFO  2020-04-24 09:36:34,331 [shard 0] storage_service - Set host_id=19784af8-6991-445c-a1f2-fa1e6568c73d to be owned by node=127.0.0.4
INFO  2020-04-24 09:36:34,332 [shard 0] storage_service - Set host_id=faf19e83-d97c-4267-941a-9a9fc1d3cbd2 to be owned by node=127.0.0.1
INFO  2020-04-24 09:36:34,332 [shard 0] storage_service - Set host_id=80135248-ec23-4c48-8701-47cdda0abe99 to be owned by node=127.0.0.5
 

 
WARN  2020-04-24 09:38:32,142 [shard 0] storage_service - Host ID collision for 80135248-ec23-4c48-8701-47cdda0abe99 between 127.0.0.5 and 127.0.0.6; 127.0.0.6 is the new owner
INFO  2020-04-24 09:38:32,142 [shard 0] storage_service - Set host_id=80135248-ec23-4c48-8701-47cdda0abe99 to be owned by node=127.0.0.6, existing=127.0.0.5
INFO  2020-04-24 09:38:32,142 [shard 0] storage_service - Remove node 127.0.0.6 from pending replacing endpoint


2) node1 (existing node):

INFO  2020-04-24 09:36:06,191 [shard 0] storage_service - Set host_id=159d50be-2e2f-41c5-83ec-06d0e02681f8 to be owned by node=127.0.0.2
INFO  2020-04-24 09:36:08,204 [shard 0] storage_service - Set host_id=a855d88a-41ca-47ac-857e-a0ceb41adc6c to be owned by node=127.0.0.3
INFO  2020-04-24 09:36:09,329 [shard 0] storage_service - Set host_id=19784af8-6991-445c-a1f2-fa1e6568c73d to be owned by node=127.0.0.4
INFO  2020-04-24 09:36:12,342 [shard 0] storage_service - Set host_id=80135248-ec23-4c48-8701-47cdda0abe99 to be owned by node=127.0.0.5


WARN  2020-04-24 09:38:32,941 [shard 0] storage_service - Host ID collision for 80135248-ec23-4c48-8701-47cdda0abe99 between 127.0.0.5 and 127.0.0.6; 127.0.0.6 is the new owner
INFO  2020-04-24 09:38:32,941 [shard 0] storage_service - Set host_id=80135248-ec23-4c48-8701-47cdda0abe99 to be owned by node=127.0.0.6, existing=127.0.0.5
INFO  2020-04-24 09:38:32,941 [shard 0] storage_service - Remove node 127.0.0.6 from pending replacing endpoint

```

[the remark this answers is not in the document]
asias: > The direction we are moving towards, with multi-add, is something that departs from Cassandra abruptly anyway.
> 
> What Cassandra users expect in this case is not important. We just have to document it.

We can still keep the property that host_id does not change in replacing operation with multi-add. If breaking compatibility is a must we consider it, if it is not why break it.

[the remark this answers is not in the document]
asias: > > This colliding can happen if we intend to bootstrap a node but unfortunately it generates tokens which belongs to an existing node. The other nodes would think it is replacing while it is not. I think it is better to be explicit.
> 
> The bootstrapping node should not chose such tokens in the first place.
> 
> If we still want to catch this, we can look at the status of the new node and allow only "hibernate" to replace.

Bug happens and we should be defensive. Yes, the extra check would help in this case. 


> > And, when the existing node sees the replacing node advertises to be replacing status, the existing nodes needs to look at the host id of the replacing node to know which node this replacing node is replacing with.
> 
> This can be derived from token metadata, you replace the node which owns any token you replace in the old ring.

Yes, this is doable. More complicated though. We need to get the tokens of the replacing node from gossip. Then choose one of the tokens and look into token_metadata for the ip address to find out the old node. we may ask: Is it enough to look at just one token? What if a bootstrap node happen to get a same token. If simply looking at a host id is enough, why not.


> > Also, replacing node uses the same host id is something existing Cassandra user would expect.
> 
> Do you know of any good reason a user may expect that?
> 
> I'm interested in this topic not only for this PR. It will make generic topology changes easier.
> 
> I think we should switch to identifying nodes by host id so that we don't have to special case the case of replacing a node with the same ip.
> 
> Having two nodes in the system with the same host id precludes that, because host id is still not unique (which is ironic).
> 
> If the user wants the same id after replace, it could be an extra step after the whole ring operation is complete, where we rename a node, but after you remove the old id from the system altogether. So host id is still unique. It'd be better to avoid this, still.

I explained in the other comments why keeping the host_id is useful.  

Regarding to use host_id or whatever id to identify a node, we have been talking about this for a very long time. For example use some id instead of ip in gossip as well. It is far beyond what this PR solves. If you think it is important please open an github issue we can discuss there. 
I hope you do not insist it as a dependency of this PR and block this PR for 1 year and a half.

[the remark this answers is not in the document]
tgrabiec: > When looking at user reported cluster issue, I usually check nodetool status to see if the host_id is owned by the ip address of the new replacing node. If that is the case, I know it is a replace operation. If the host id changes, it is a bootstrap operation.
> Sometimes, due to operation error, we might find the replacing node uses the same host id but with new and different tokens. If the replacing node would use different host_id, we do not even know it is supposed to be an replace operation but ended up using wrong tokens.

We could satisfy this need in a different manner. Rather than reuse host id, we can make the replacing node advertise through gossip the host id which it replaces and nodetool can pick it up and present.

Even if we don't do it, we can still get this information from the log of the replacing node.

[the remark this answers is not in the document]
tgrabiec: > 
> The mapping between host id and tokens never changes. It is a very good property in my view. The host id is added and removed from the cluster when new nodes are added or removed from the cluster. The whole code base assumes this. Do you have a very good reason to change this?

I think host id is supposed to identify the host, not its tokens. If you decommission a node and add a new node with the same tokens, it will have a different host id despite having the same tokens. So a host has tokens, host id identifies a host, but host id != tokens. 

Which code do you think assumes that host id == tokens? 

If we needed a hash of host's tokens, we should introduce a new variable for that.

> 
> @tgrabiec Can you please point out what are the special cases you are talking about "The fact that we try to reuse the same host id on a different node generates special cases all around" beyond this
> 
> ```
> +    if (!replacing_a_node_with_diff_ip) {
> +        // Replacing node with a different ip should own the host_id only after
> +        // the replacing node becomes NORMAL status. It is updated in
> +        // handle_state_normal().
> +        _token_metadata.update_host_id(local_host_id, get_broadcast_address());
> +    }
> ```

It generates special cases indirectly because if we don't have a unique identifier of a node, the coordinator code which routes requests has to special case the replace operation by filtering replaced node from the pending endpoint list.

> 
> This special case is needed to avoid more special cases to happen (handling replacing of the dead node differently on the replacing node than the rest of the cluster)

Why can't we have unified handling if replacing node uses a different id?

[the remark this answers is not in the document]
tgrabiec: > > > This colliding can happen if we intend to bootstrap a node but unfortunately it generates tokens which belongs to an existing node. The other nodes would think it is replacing while it is not. I think it is better to be explicit.
> > 
> > 
> > The bootstrapping node should not chose such tokens in the first place.
> > If we still want to catch this, we can look at the status of the new node and allow only "hibernate" to replace.
> 
> Bug happens and we should be defensive. Yes, the extra check would help in this case.
> 
> > > And, when the existing node sees the replacing node advertises to be replacing status, the existing nodes needs to look at the host id of the replacing node to know which node this replacing node is replacing with.
> > 
> > 
> > This can be derived from token metadata, you replace the node which owns any token you replace in the old ring.
> 
> Yes, this is doable. More complicated though. 

Another option is to advertise the replaced host id using a separate gossip property.

> We need to get the tokens of the replacing node from gossip. Then choose one of the tokens and look into token_metadata for the ip address to find out the old node. we may ask: Is it enough to look at just one token? What if a bootstrap node happen to get a same token. If simply looking at a host id is enough, why not.

Are you thinking about concurrent bootstrap and replace here? 


> 
> > > Also, replacing node uses the same host id is something existing Cassandra user would expect.
> > 
> > 
> > Do you know of any good reason a user may expect that?
> > I'm interested in this topic not only for this PR. It will make generic topology changes easier.
> > I think we should switch to identifying nodes by host id so that we don't have to special case the case of replacing a node with the same ip.
> > Having two nodes in the system with the same host id precludes that, because host id is still not unique (which is ironic).
> > If the user wants the same id after replace, it could be an extra step after the whole ring operation is complete, where we rename a node, but after you remove the old id from the system altogether. So host id is still unique. It'd be better to avoid this, still.
> 
> I explained in the other comments why keeping the host_id is useful.
> 
> Regarding to use host_id or whatever id to identify a node, we have been talking about this for a very long time. For example use some id instead of ip in gossip as well. It is far beyond what this PR solves. If you think it is important please open an github issue we can discuss there.
> I hope you do not insist it as a dependency of this PR and block this PR for 1 year and a half.

I'm not insisting to make this a dependency yet. I'm still trying to understand if this PR is correct. This problem makes it harder to prove it correct.

locator/abstract_replication_strategy.hh
```diff
@@ -98,6 +98,12 @@ public:
     virtual void validate_options() const = 0;
     virtual std::optional<std::set<sstring>> recognized_options() const = 0;
     virtual size_t get_replication_factor() const = 0;
+    // Decide if the replication strategy allow removing the node being
+    // replaced from the natural endpoints when a node is being replaced in the
+    // cluster. LocalStrategy is the not allowed to do so because it always
+    // returns the node itself as the natural_endpoints and the node will not
+    // appear in the pending_endpoints.
```
tgrabiec, line 105: I still think that having `get_natural_endpoints_without_pending()` at this level would be better, because you wouldn't have to understand what storage_proxy does in order to understand what this means.

service/storage_proxy.cc
```diff
@@ -2179,8 +2178,7 @@ future<> storage_proxy::mutate_counters(Range&& mutations, db::consistency_level
 storage_proxy::paxos_participants
 storage_proxy::get_paxos_participants(const sstring& ks_name, const dht::token &token, db::consistency_level cl_for_paxos) {
     keyspace& ks = _db.local().find_keyspace(ks_name);
-    locator::abstract_replication_strategy& rs = ks.get_replication_strategy();
-    std::vector<gms::inet_address> natural_endpoints = rs.get_natural_endpoints(token);
+    std::vector<gms::inet_address> natural_endpoints = get_natural_endpoints_without_node_being_replaced(ks, token);
```
tgrabiec, line 2181: Don't we also need to change storage_proxy::get_live_endpoints() so that reads don't try to read from the pending (replacing) node as if it was the replaced node?

service/storage_service.cc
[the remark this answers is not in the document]
asias: So are you planing to do so? What part of the host_id related in this series do you think it is incorrect? I am happy to hear.

locator/abstract_replication_strategy.hh
```diff
@@ -98,6 +98,12 @@ public:
...
+    virtual bool allow_remove_node_being_replaced_from_natural_endpoints() const { return true; };
```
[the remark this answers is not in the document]
asias: If it is unsafe we should set it to false when we add new strategies.

locator/abstract_replication_strategy.hh
asias: It is worse in my view because the abstract_replication_strategy.hh needs to understand pending nodes which only token_metadata knows.

service/storage_proxy.cc
asias: I do not know for sure. Let's ask people who is familiar with service/storage_proxy.cc @gleb-cloudius ?

locator/abstract_replication_strategy.hh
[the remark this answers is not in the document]
tgrabiec: It is unsafe to have a default implementation. "false" is also not right for all strategies. This should be pure virtual so that each strategy has to decide for itself.

locator/abstract_replication_strategy.hh
tgrabiec: I think the concept of pending nodes belongs to replication strategy. We currently get it from  token_metadata, but it doesn't belong there. Notice how `get_pending_ranges()` takes a keyspace name, because each replication strategy may have a different idea about what pending ranges should be. `token_metadata_impl::calculate_pending_ranges_for_bootstrap()` takes an `abstract_replication_strategy&`. Eventually we should clean this up.

gleb-cloudius: I agree with Tomke here and this is what I proposed on the first version of the patchset. The function does not belong in the storage_proxy since it is not the one who is doing all the vnode calculations.

service/storage_proxy.cc
gleb-cloudius: A read should only contact those replicas that at least potentially have up-to-date data. Does a node been replaced is reported as dead or alive? If the later it should not be contacted during read.

asias: The node being replaced can be dead (when the replacing node is started with a different ip) or alive (when the replacing node is started  with the same IP address as the node being replaced). I think we should use get_natural_endpoints_without_node_being_replaced in get_live_endpoints to filter out the node being replaced.

locator/abstract_replication_strategy.hh
asias: > I think the concept of pending nodes belongs to replication strategy. We currently get it from token_metadata, but it doesn't belong there. Notice how `get_pending_ranges()` takes a keyspace name, because each replication strategy may have a different idea about what pending ranges should be. `token_metadata_impl::calculate_pending_ranges_for_bootstrap()` takes an `abstract_replication_strategy&`. Eventually we should clean this up.

I think you are mixing up pending node and pending ranges. Pending node is a per cluster.  Pending ranges are per keyspace that's why get_pending_ranges takes keyspace.

I think we can move the helper out of storage_proxy.

locator/abstract_replication_strategy.hh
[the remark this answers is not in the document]
asias: If the strategy does not allow it, it should set to false. If it does not care, it should be the default which is true. I think it is overkill to make it pure virtual and ask every strategy to set it explicitly. It adds more code.

asias:
Changes in a3e2b461b98d2c2760ca36e9312da304a38af296 (v5)
- Move get_natural_endpoints_without_node_being_replaced out of storage_proxy
- Make storage_proxy::get_live_endpoints get_natural_endpoints_without_node_being_replaced 
- Rebase

locator/abstract_replication_strategy.hh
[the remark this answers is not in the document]
glommer: Asias, I think Tomek is right on this one.

The goal is not to minimize code, it is to maximize safety. This is safe in the current code but as soon as you start adding more strategies, deriving from existing strategies, and end up forgetting to add this, then we just got ourselves a nasty bug.

Best to write a few extra lines of code now

locator/abstract_replication_strategy.hh
tgrabiec: > > I think the concept of pending nodes belongs to replication strategy. We currently get it from token_metadata, but it doesn't belong there. Notice how `get_pending_ranges()` takes a keyspace name, because each replication strategy may have a different idea about what pending ranges should be. `token_metadata_impl::calculate_pending_ranges_for_bootstrap()` takes an `abstract_replication_strategy&`. Eventually we should clean this up.
> 
> I think you are mixing up pending node and pending ranges. Pending node is a per cluster. Pending ranges are per keyspace that's why get_pending_ranges takes keyspace.

Pending ranges contain pending nodes. In this context, "pending node" is the node which is added to the replica set for a given token range. We can have no pending node in the ring (no physical node added), but can have pending nodes in replication strategy, e.g. when we decommission a node. This is the meaning of "pending node" (node = endpoint) which `token_metadata::pending_endpoints_for()` uses.

Pending ranges are actually per replication strategy instance. We currently have one replication strategy per keyspace, but when we introduce tablet partitioning, each table will have its own replication strategy and it will be per table.

asias:
Changes in (7d86a3b208659e92489216e71df13fb33b1a4c3a) v6:
- Make allow_remove_node_being_replaced_from_natural_endpoints pure virtual

locator/abstract_replication_strategy.hh
asias: If you are talking about future it is a different story, we can change anything in the future. At the moment, the pending range is per keyspace. Take a look at _pending_ranges, _pending_ranges_interval_map.

asias: > Pending ranges contain pending nodes. In this context, "pending node" is the node which is added to the replica set for a given token range. We can have no pending node in the ring (no physical node added), but can have pending nodes in replication strategy, e.g. when we decommission a node. This is the meaning of "pending node" (node = endpoint) which token_metadata::pending_endpoints_for() uses.

Of course pending ranges have a node associated with the ranges, otherwise the ranges are meaningless. Then you should have said "pending ranges" not "pending node" belong to replication strategy.  However, the replication strategy does not need to know the pending ranges because, we only need replication strategy to tell what ranges belong to what node given a token ring, we can calculate the pending ranges. The given token ring is a parameter to replication strategy. If we move pending ranges to replication strategy, replication strategy needs to store a temporary token ring which is unnecessary and less flexible.

locator/abstract_replication_strategy.hh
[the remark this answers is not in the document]
asias: To have less code is not the goal but it is nice to have. It is trivial to change, I modified it in v6.

locator/abstract_replication_strategy.hh
tgrabiec: > > Pending ranges contain pending nodes. In this context, "pending node" is the node which is added to the replica set for a given token range. We can have no pending node in the ring (no physical node added), but can have pending nodes in replication strategy, e.g. when we decommission a node. This is the meaning of "pending node" (node = endpoint) which token_metadata::pending_endpoints_for() uses.
> 
> Of course pending ranges have a node associated with the ranges, otherwise the ranges are meaningless. Then you should have said "pending ranges" not "pending node" belong to replication strategy. 

I think I can say that the concept of a pending node belongs to replication strategy at the API level. It doesn't matter how you internally organize data. You ask replication strategy for pending nodes for a given key, so the user needs to understand what a pending node is, so this concept belongs to the idea of replication strategy.

> However, the replication strategy does not need to know the pending ranges because, we only need replication strategy to tell what ranges belong to what node given a token ring, we can calculate the pending ranges. The given token ring is a parameter to replication strategy. If we move pending ranges to replication strategy, replication strategy needs to store a temporary token ring which is unnecessary and less flexible.

The fact that replication strategy relies on token ring for replication is a detail of the strategy. That's why I think pending ranges should belong to the replication strategy and not token ring. 

There could be strategies which don't rely on the ring, like tablet partitioning, and they would obtain pending nodes from somewhere else. 

The user of the strategy should not be concerned with where the strategy gets its replication information from, he should just be able to ask it for old and new replica sets for a given key.

So it's more flexible if we don't store pending ranges in token metadata but in the strategy.

PR #5733, opened 2020-02-06
Reduce usage of global_partitioner()

haaawk:
In many places we use global_partitioner() to obtain parameters that are available in config. This PR replaces number of global_partitioner() calls with equivalent non-global ways.

tests: unit(dev)

### DIFF
--- a/dht/i_partitioner.hh
+++ b/dht/i_partitioner.hh
@@ -224,10 +224,6 @@ public:
         return _shard_count;
     }
 
-    sstring cpu_sharding_algorithm_name() const {
-        return "biased-token-round-robin";
-    }
-
     unsigned sharding_ignore_msb() const {
         return _sharding_ignore_msb_bits;
     }
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -51,7 +51,6 @@
 #include "gms/i_endpoint_state_change_subscriber.hh"
 #include "service/storage_service.hh"
 #include "message/messaging_service.hh"
-#include "dht/i_partitioner.hh"
 #include "database.hh"
 #include "log.hh"
 #include "db/system_keyspace.hh"
@@ -60,7 +59,6 @@
 #include <seastar/core/metrics.hh>
 #include <seastar/util/defer.hh>
 #include <chrono>
-#include "dht/i_partitioner.hh"
 #include "db/config.hh"
 #include <boost/range/algorithm/set_algorithm.hpp>
 #include <boost/range/adaptors.hpp>
@@ -90,7 +88,7 @@ void gossiper::set_cluster_name(sstring name) {
 }
 
 sstring gossiper::get_partitioner_name() {
-    return dht::global_partitioner().name();
+    return _cfg.partitioner();
 }
 
 std::set<inet_address> gossiper::get_seeds() {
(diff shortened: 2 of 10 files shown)


gms/gossiper.cc
```diff
@@ -90,7 +88,7 @@ void gossiper::set_cluster_name(sstring name) {
 }
 
 sstring gossiper::get_partitioner_name() {
-    return dht::global_partitioner().name();
+    return _cfg.partitioner();
```
denesb: This looks weird, is `_cfg.partitioner()` an `sstring` already? In that case it should be called `partitioner_name()`, no?
Or does the partitioner have `operator sstring()`?

haaawk: It is an sstring. I agree the name could be better but it is part of yaml config so changing it is not easy. It would break compatibility with old configs.

denesb: Ok, no need to change it then.

thrift/handler.cc
```diff
@@ -767,7 +767,7 @@ class thrift_handler : public CassandraCobSvIf {
     }
 
     void describe_partitioner(thrift_fn::function<void(std::string const& _return)> cob) {
-        cob(dht::global_partitioner().name());
+        cob(_db.local().get_config().partitioner());
```
nyh: I'm confused why this is better... I'm worried that it is worse: The partitioner is still "global" in the sense we have just one _db, and if later we want to have a per-table partitioner, this code will still be broken - but will now be harder to find because it no longer has that easy-to-find global_partitioner() call.

This concern also applies to several other changes in other patches.

service/storage_service.cc
```diff
@@ -1934,7 +1934,7 @@ future<> storage_service::gossip_snitch_info() {
 future<> storage_service::gossip_sharding_info() {
     return _gossiper.add_local_application_state({
         { gms::application_state::SHARD_COUNT, value_factory.shard_count(smp::count) },
-        { gms::application_state::IGNORE_MSB_BITS, value_factory.ignore_msb_bits(dht::global_partitioner().sharding_ignore_msb()) },
+        { gms::application_state::IGNORE_MSB_BITS, value_factory.ignore_msb_bits(_db.local().get_config().murmur3_partitioner_ignore_msb_bits()) },
```
nyh: Why is the method called here named "murmur3_partitioner_ignore_msb_bits()"?

dht/i_partitioner.hh
```diff
@@ -224,10 +224,6 @@ public:
         return _shard_count;
     }
 
-    sstring cpu_sharding_algorithm_name() const {
-        return "biased-token-round-robin";
-    }
-
     unsigned sharding_ignore_msb() const {
```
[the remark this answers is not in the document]
haaawk: Coming in the future PR for now we can't because it's still used in one place.

service/storage_service.cc
haaawk: I agree that it would be better if murmur3 wasn't there but it's hard to change it now.

thrift/handler.cc
haaawk: I'm actually having a following PR that does the switch to per table partitioner. This PR is a prelude for that. 

To address your concern. Even with per table partitioner there will be a default partitioner and this method will return the name of that default partitioner. The default partitioner will be the one used when a table won't have it's own partitioner set.

We will have to extend the protocol to inform drivers about partitioners used by different tables but that's another story.

Bottom line is, if you're worried this PR will make stuff worse for per table partitioners - stop. It's part of the effort to make it happen.

haaawk:
@nyh do you have any more concerns?

PR #5755, opened 2020-02-09
Reduce #include bloat around cql3 internals from non-cql3 users

avikivity:
This series removes some #include dependencies around cql3. It results in 30k line (6.6%) reduction in the preprocessed size of database.i, mainly due to elimination of boost::regex (which was brought in in turn by like_matcher). This should result in fewer and faster recompiles.

### DIFF
--- a/table_helper.hh
+++ b/table_helper.hh
@@ -24,10 +24,14 @@
 
 #include <seastar/util/gcc6-concepts.hh>
 #include <seastar/core/apply.hh>
-#include "cql3/statements/modification_statement.hh"
 #include "cql3/statements/prepared_statement.hh"
 #include "cql3/query_processor.hh"
 #include "service/migration_manager.hh"
+
+
+namespace cql3::statements {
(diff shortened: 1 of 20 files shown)


table_helper.hh
```diff
@@ -24,11 +24,15 @@
 
 #include <seastar/util/gcc6-concepts.hh>
 #include <seastar/core/apply.hh>
-#include "cql3/statements/modification_statement.hh"
```
haaawk: Please make sure it does not break audit in enterprise. If I remember correctly table_helper was used there.

nyh: You're right, the "enterprise" branch no longer builds with this change. I get a bizaare compilation error from do_with(). @avikivity can you please look into this and if needed, submit a new version of this series? Thanks.

avikivity: When I merge master into enterprise I will fix things up in the merge commit. In general contributors do not need to consider invisible branches.

Issue #5800, opened 2020-02-13
Active nodes wrongfully marked as DOWN 


asias:
User reported an issue that after a node restart, the restarted node
is marked as DOWN by other nodes in the cluster while the node is up
and running normally.

pidiaquez:
Hi Team

     Customer asks if will it be delivered as a patch for 2019.1.2 Enterprise version they have? Or will it go to 2020? Or else...

Thanks in advance

Pablo

avikivity:
Will backport, but later after it gets some more mileage in master.

PR #5801, opened 2020-02-13
gossiper: Always use the new generation number

asias:
User reported an issue that after a node restart, the restarted node
is marked as DOWN by other nodes in the cluster while the node is up
and running normally.

Consier the following:

- n1, n2, n3 in the cluster
- n3 shutdown itself
- n3 send shutdown verb to n1 and n2
- n1 and n2 set n3 in SHUTDOWN status and force the heartbeat version to
  INT_MAX
- n3 restarts
- n3 sends gossip shadow rounds to n1 and n2, in
  storage_service::prepare_to_join,
- n3 receives response from n1, in gossiper::handle_ack_msg, since
  _enabled = false and _in_shadow_round == false, n3 will apply the
  application state in fiber1, filber 1 finishes faster filber 2, it
  sets _in_shadow_round = false
- n3 receives response from n2, in gossiper::handle_ack_msg, since
  _enabled = false and _in_shadow_round == false, n3 will apply the
  application state in fiber2, filber 2 yields
- n3 finishes the shadow round and continues
- n3 resets gossip endpoint_state_map with
  gossiper.reset_endpoint_state_map()
- n3 resumes fiber 2, apply application state about n3 into
  endpoint_state_map, at this point endpoint_state_map contains
  information including n3 itself from n2.
- n3 calls gossiper.start_gossiping(generation_number, app_states, ...)
  with new generation number generated correctly in
  storage_service::prepare_to_join, but in
  maybe_initialize_local_state(generation_nbr), it will not set new
  generation and heartbeat if the endpoint_state_map contains itself
- n3 continues with the old generation and heartbeat learned in fiber 2
- n3 continues the gossip loop, in gossiper::run,
  hbs.update_heart_beat() the heartbeat is set to the number starting
  from 0.
- n1 and n2 will not get update from n3 because they use the same
  generation number but n1 and n2 has larger heartbeat version
- n1 and n2 will mark n3 as down even if n3 is alive.

To fix, always use the the new generation number.

Fixes: #5800
Backports: 3.0 3.1 3.2

### DIFF
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -1728,9 +1728,9 @@ future<> gossiper::start_gossiping(int generation_nbr, std::map<application_stat
         g.init_messaging_service_handler(do_bind);
     }).then([this, generation_nbr, preload_local_states] {
         build_seeds_list();
-        /* initialize the heartbeat state for this localEndpoint */
-        maybe_initialize_local_state(generation_nbr);
         endpoint_state& local_state = endpoint_state_map[get_broadcast_address()];
+        local_state.set_heart_beat_state_and_update_timestamp(heart_beat_state(generation_nbr));
+        local_state.mark_alive();
         for (auto& entry : preload_local_states) {
             local_state.add_application_state(entry.first, entry.second);
         }


eliransin:
@tgrabiec did you had a chance to review this? (ping if you didn't :) )

tgrabiec:
@asias I wonder if there's a bigger problem here. We clear the endpoint state map after the shadow round, so that we have a state as if the shadow round didn't happen. If we don't wait for pending ACKs, it's as if we didn't clean the endpoint state, which is causing this issue, but maybe it would cause even more issues. So isn't the main problem that the shadow round doesn't wait for pending ACKs? If not, is there any reason to clear the endopoint state after the shadow round now?

asias:
> @asias I wonder if there's a bigger problem here. We clear the endpoint state map after the shadow round, so that we have a state as if the shadow round didn't happen. If we don't wait for pending ACKs, it's as if we didn't clean the endpoint state, which is causing this issue, but maybe it would cause even more issues. So isn't the main problem that the shadow round doesn't wait for pending ACKs? If not, is there any reason to clear the endopoint state after the shadow round now?

We do not wait for all the shadow round responses, we only wait for one of the responses. Currently, the shadow round message reuses syn message which is one way message, it is hard to know if the receiver will send back reponses or not. E.g., it could be the node is slow to repsonse the response will come eventually, or the node is dead the response will never come. Making it two way message will help so we can reliably wait but it is a major change I want to leave it to the get rid of seed project. 

The real reason to clear endpoint state map is that it prevents using the old applcation state of the node itself from other nodes. The node itself knows better about itself than anyone else. However, not to clear endpoint state map is not a problem. Because there are two cases, 1) the shadow round response is applied before gossiper::start_gossiping where we start to use the new generation, the applcation state of the local state will be replaced by the new state in gossiper::start_gossiping. 2) the shadow round response is applied after gossiper::start_gossiping, the application state in the shadow round  repsonse for the node itself will be ignored because it has smaller geneartion number.

slivne:
@tgrabiec did you merge this ?

asias:
I see it is in master now:
```
commit 62774ff88277406de46316e9aadba9a16f270882
Author: Asias He <asias@scylladb.com>
Date:   Thu Feb 13 11:54:14 2020 +0800

    gossiper: Always use the new generation number
```

Not sure why it is not closed automatically here.

asias:
Closing. The patch is  already merged in master.

PR #5803, opened 2020-02-13
CDC: split mutations by timestamp and ttl

kbr-:
Add a bunch of new structs describing a change made
to a table, and an `extract_changes` function which takes a mutation and
returns the set of changes contained in this mutation, separated by
timestamp and ttl.

Add a separate "insert" operation to distinguish inserts (which create a row marker)
from updates (which don't).

Refactor the `transform` function.
It doesn't inspect the mutation manually now. Instead, it calls
`extract_changes` which returns an easily accesible (structured) set of
changes, split by timestamp and ttl.

It then adds a CDC log table row for each of those changes.

As a byproduct:
1. the CDC log now distinguishes between inserts, updates, and row deletes.
2. static row changes appear even if no clustering rows were affected by
   the base mutation (the static row change is always a separate
   change).

Fixes https://github.com/scylladb/scylla/issues/5719.
Fixes https://github.com/scylladb/scylla/issues/5723.
Fixes https://github.com/scylladb/scylla/issues/5744.
Ref: https://github.com/scylladb/scylla/issues/5709: we distinguish row deletes now as separate operations, but preimages for row deletes do not appear yet.

Tests: unit (dev)

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -416,6 +416,200 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
     return utils::UUID_gen::get_random_time_UUID_from_micros(t);
 }
 
+struct atomic_column_update {
+    column_id id;
+    atomic_cell cell;
+};
+
+struct nonatomic_column_deletion {
+    column_id id;
+};
+
+struct nonatomic_column_update {
+    column_id id;
+    std::vector<std::pair<bytes, atomic_cell>> cells;
+};
+
+struct static_row_update {
+    gc_clock::duration ttl;
+    std::vector<atomic_column_update> atomic_entries;
+    std::vector<nonatomic_column_deletion> nonatomic_deletions;
+    std::vector<nonatomic_column_update> nonatomic_updates;
+};
+
+struct clustered_row_insert {
+    gc_clock::duration ttl;
+    clustering_key key;
+    std::vector<atomic_column_update> atomic_entries;
+    std::vector<nonatomic_column_deletion> nonatomic_deletions;
+    // INSERTs can't express updates of individual cells inside a non-atomic
+    // (without deleting the entire field first), so no `nonatomic_updates` field
+    // overwriting a nonatomic column inside an INSERT will be split into two changes:
+    // one with a nonatomic deletion, and one with a nonatomic update
+};
+
+struct clustered_row_update {
+    gc_clock::duration ttl;
+    clustering_key key;
+    std::vector<atomic_column_update> atomic_entries;
+    std::vector<nonatomic_column_deletion> nonatomic_deletions;
+    std::vector<nonatomic_column_update> nonatomic_updates;
+};
+
+struct clustered_row_deletion {
+    clustering_key key;
+};
+
+struct clustered_range_deletion {
+    clustering_key_prefix start;
+    bound_kind start_kind;
+    clustering_key_prefix end;
+    bound_kind end_kind;
+};
+
+struct batch {
+    std::vector<static_row_update> static_updates;
+    std::vector<clustered_row_insert> clustered_inserts;
+    std::vector<clustered_row_update> clustered_updates;
+    std::vector<clustered_row_deletion> clustered_row_deletions;
+    std::vector<clustered_range_deletion> clustered_range_deletions;
+    bool has_partition_deletion = false;
+};
+
+using set_of_changes = std::map<api::timestamp_type, batch>;
+
+struct row_update {
+    std::vector<atomic_column_update> atomic_entries;
+    std::vector<nonatomic_column_deletion> nonatomic_deletions;
+    std::vector<nonatomic_column_update> nonatomic_updates;
+};
+
+std::map<std::pair<api::timestamp_type, gc_clock::duration>, row_update>
+extract_row_updates(const row& r, column_kind ckind, const schema& schema) {
+    std::map<std::pair<api::timestamp_type, gc_clock::duration>, row_update> result;
+    r.for_each_cell([&] (column_id id, const atomic_cell_or_collection& cell) {
+        auto& cdef = schema.column_at(ckind, id);
+        if (cdef.is_atomic()) {
+            auto view = cell.as_atomic_cell(cdef);
+            auto timestamp_and_ttl = std::pair(
+                    view.timestamp(),
+                    view.is_live_and_has_ttl() ? view.ttl() : gc_clock::duration(0)
+                );
+            result[timestamp_and_ttl].atomic_entries.push_back({id, atomic_cell(*cdef.type, view)});
+            return;
+        }
+
+        cell.as_collection_mutation().with_deserialized(*cdef.type, [&] (collection_mutation_view_description mview) {
+            auto desc = mview.materialize(*cdef.type);
+            for (auto& [k, v]: desc.cells) {
+                auto timestamp_and_ttl = std::pair(
+                        v.timestamp(),
+                        v.is_live_and_has_ttl() ? v.ttl() : gc_clock::duration(0)
+                    );
+                auto& updates = result[timestamp_and_ttl].nonatomic_updates;
+                if (updates.empty() || updates.back().id != id) {
+                    updates.push_back({id, {}});
+                }
+                updates.back().cells.push_back({std::move(k), std::move(v)});
+            }
+
+            if (desc.tomb) {
+                auto timestamp_and_ttl = std::pair(desc.tomb.timestamp, gc_clock::duration(0));
+                result[timestamp_and_ttl].nonatomic_deletions.push_back({id});
+            }
+        });
+    });
+    return result;
+};
+
+set_of_changes extract_changes(const mutation& base_mutation, const schema& base_schema) {
+    set_of_changes res;
+    auto& p = base_mutation.partition();
+
+    auto sr_updates = extract_row_updates(p.static_row().get(), column_kind::static_column, base_schema);
+    for (auto& [k, up]: sr_updates) {
+        auto [timestamp, ttl] = k;
+        res[timestamp].static_updates.push_back({
+                ttl,
+                std::move(up.atomic_entries),
+                std::move(up.nonatomic_deletions),
+                std::move(up.nonatomic_updates)
+            });
+    }
+
+    for (const rows_entry& cr : p.clustered_rows()) {
+        auto cr_updates = extract_row_updates(cr.row().cells(), column_kind::regular_column, base_schema);
+
+        const auto& marker = cr.row().marker();
+        auto marker_timestamp = marker.timestamp();
+        auto marker_ttl = marker.is_expiring() ? marker.ttl() : gc_clock::duration(0);
+        if (marker.is_live()) {
+            // make sure that an entry corresponding to the row marker's timestamp and ttl is in the map
+            (void)cr_updates[std::pair(marker_timestamp, marker_ttl)];
+        }
+
+        auto is_insert = [&] (api::timestamp_type timestamp, gc_clock::duration ttl) {
+            if (!marker.is_live()) {
+                return false;
+            }
+
+            return timestamp == marker_timestamp && ttl == marker_ttl;
+        };
+
+        for (auto& [k, up]: cr_updates) {
+            auto [timestamp, ttl] = k;
+
+            if (is_insert(timestamp, ttl)) {
+                res[timestamp].clustered_inserts.push_back({
+                        ttl,
+                        cr.key(),
+                        std::move(up.atomic_entries),
+                        std::move(up.nonatomic_deletions)
+                    });
+                if (!up.nonatomic_updates.empty()) {
+                    // nonatomic updates cannot be expressed with an INSERT.
+                    res[timestamp].clustered_updates.push_back({
+                            ttl,
+                            cr.key(),
+                            {},
+                            {},
+                            std::move(up.nonatomic_updates)
+                        });
+                }
+            } else {
+                res[timestamp].clustered_updates.push_back({
+                        ttl,
+                        cr.key(),
+                        std::move(up.atomic_entries),
+                        std::move(up.nonatomic_deletions),
+                        std::move(up.nonatomic_updates)
+                    });
+            }
+        }
+
+        auto row_tomb = cr.row().deleted_at().regular();
+        if (row_tomb) {
+            res[row_tomb.timestamp].clustered_row_deletions.push_back({cr.key()});
+        }
+    }
+
+    for (const auto& rt: p.row_tombstones()) {
+        if (rt.tomb.timestamp != api::missing_timestamp) {
+            res[rt.tomb.timestamp].clustered_range_deletions.push_back({
+                rt.start, rt.start_kind,
+                rt.end, rt.end_kind
+            });
+        }
+    }
+
+    auto partition_tomb_timestamp = p.partition_tombstone().timestamp;
+    if (partition_tomb_timestamp != api::missing_timestamp) {
+        res[partition_tomb_timestamp].has_partition_deletion = true;
+    }
+
+    return res;
+}
+
 class transformer final {
 private:
     db_context _ctx;
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -254,7 +254,6 @@ sstring log_name(const sstring& table_name) {
 
 static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID> uuid) {
     schema_builder b(s.ks_name(), log_name(s.cf_name()));
-    b.set_default_time_to_live(gc_clock::duration{s.cdc_options().ttl()});
     b.set_comment(sprint("CDC log for %s.%s", s.ks_name(), s.cf_name()));
     b.with_column("stream_id_1", long_type, column_kind::partition_key);
     b.with_column("stream_id_2", long_type, column_kind::partition_key);
@@ -424,6 +423,7 @@ private:
     schema_ptr _log_schema;
     const column_definition& _op_col;
     const column_definition& _ttl_col;
+    ttl_opt _cdc_ttl_opt;
 
     clustering_key set_pk_columns(const partition_key& pk, api::timestamp_type ts, bytes decomposed_tuuid, int batch_no, mutation& m) const {
         const auto log_ck = clustering_key::from_exploded(
@@ -435,7 +435,8 @@ private:
             auto cdef = m.schema()->get_column_definition(to_bytes("_" + column.name()));
             auto value = atomic_cell::make_live(*column.type,
                                                 ts,
-                                                bytes_view(pk_value[pos]));
+                                                bytes_view(pk_value[pos]),
+                                                _cdc_ttl_opt);
             m.set_cell(log_ck, *cdef, std::move(value));
             ++pos;
         }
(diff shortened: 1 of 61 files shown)


avikivity:
@kbr- did you use the mutation splitter we already have, or did you roll a new one?

kbr-:
@avikivity 
I didn't use any existing mutation splitter.

I don't actually make a bunch of `mutation`s out of a `mutation`, because we don't need a bunch of `mutation`s. I introduced new structures that describe the set of changes the `mutation` makes in a way that's easy to consume in `transform`:

```
struct static_row_update;
struct clustered_row_insert;
struct clustered_row_update;
struct clustered_row_deletion;
struct clustered_range_deletion;

struct batch {
    std::vector<static_row_update> static_updates;
    std::vector<clustered_row_insert> clustered_inserts;
    std::vector<clustered_row_update> clustered_updates;
    std::vector<clustered_row_deletion> clustered_row_deletions;
    std::vector<clustered_range_deletion> clustered_range_deletions;
    bool has_partition_deletion = false;
};

using set_of_changes = std::map<api::timestamp_type, batch>;

set_of_changes extract_changes(const mutation& base_mutation, const schema& base_schema);
```

kbr-:
v2:
- rebased on next.
- removed `static_row` struct leftover from my experiments.

denesb:
@kbr- the mutation splitter doesn't produce mutations. It takes a `flat_mutation_reader` and produces readers for each timestamp class.

cdc/log.cc
```diff
@@ -416,6 +416,194 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+    std::vector<atomic_column_update> atomic_entries;
+    std::vector<nonatomic_column_deletion> nonatomic_deletions;
+    std::vector<nonatomic_column_update> nonatomic_updates;
+};
+
+using set_of_changes = std::map<api::timestamp_type, batch>;
+
+set_of_changes extract_changes(const mutation& base_mutation, const schema& base_schema) {
+    auto extract_row_updates = [&base_schema] (const row& r, column_kind ckind) {
+        std::map<std::pair<api::timestamp_type, gc_clock::rep>, row_update> ups;
```
denesb: What us `ups`?

cdc/log.cc
```diff
@@ -416,6 +416,194 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+    column_id id;
+};
+
+struct nonatomic_column_update {
+    column_id id;
+    std::vector<std::pair<bytes, atomic_cell>> cells;
+};
+
+struct static_row_update {
+    gc_clock::rep ttl;
```
denesb: Why `rep` and not `time_point`?

cdc/log.cc
```diff
@@ -416,6 +416,194 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+    std::vector<atomic_column_update> atomic_entries;
+    std::vector<nonatomic_column_deletion> nonatomic_deletions;
+};
+
+struct clustered_row_update {
+    gc_clock::rep ttl;
+    clustering_key key;
+    std::vector<atomic_column_update> atomic_entries;
+    std::vector<nonatomic_column_deletion> nonatomic_deletions;
+    std::vector<nonatomic_column_update> nonatomic_updates;
```
denesb: How come insert doesn't have a `nonatomic_inserts` or equivalent?

cdc/log.cc
```diff
@@ -416,6 +416,194 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+                            v.is_live_and_has_ttl() ? v.ttl().count() : 0
+                        );
+                    auto& updates = ups[timestamp_and_ttl].nonatomic_updates;
+                    if (updates.empty() || updates.back().id != id) {
+                        updates.push_back({id, {}});
+                    }
+                    updates.back().cells.push_back({std::move(k), std::move(v)});
+                }
+
+                auto timestamp_and_ttl = std::make_pair(desc.tomb.timestamp, 0);
```
denesb: No need for `make_pair`, we have class template deduction now, so `std::pair()` works too.

cdc/log.cc
```diff
@@ -416,6 +416,194 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+using set_of_changes = std::map<api::timestamp_type, batch>;
+
+set_of_changes extract_changes(const mutation& base_mutation, const schema& base_schema) {
+    auto extract_row_updates = [&base_schema] (const row& r, column_kind ckind) {
+        std::map<std::pair<api::timestamp_type, gc_clock::rep>, row_update> ups;
+        r.for_each_cell([&] (column_id id, const atomic_cell_or_collection& cell) {
+            auto& cdef = base_schema.column_at(ckind, id);
+            if (cdef.is_atomic()) {
+                auto view = cell.as_atomic_cell(cdef);
+                auto timestamp_and_ttl = std::make_pair(
```
denesb: No need for make_pair, we have class template deduction now, so std::pair() works too.

cdc/log.cc
```diff
@@ -416,6 +416,194 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+                        view.timestamp(),
+                        view.is_live_and_has_ttl() ? view.ttl().count() : 0
+                    );
+                ups[timestamp_and_ttl].atomic_entries.push_back({id, atomic_cell(*cdef.type, view)});
```
denesb: Is `entries` meant to contain both insertions, updates and deletions? If yes, why is this handled separately for nonatomic cells (into different fields)?

cdc/log.cc
```diff
@@ -416,6 +416,194 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+                return;
+            }
+
+            cell.as_collection_mutation().with_deserialized(*cdef.type, [&] (collection_mutation_view_description mview) {
+                auto desc = mview.materialize(*cdef.type);
```
denesb: Why is this necessary. This will copy the entire collection.

cdc/log.cc
denesb: The tomb might be missing, in which case you don't need to store it.

cdc/log.cc
```diff
@@ -416,6 +416,194 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+struct row_update {
+    std::vector<atomic_column_update> atomic_entries;
+    std::vector<nonatomic_column_deletion> nonatomic_deletions;
+    std::vector<nonatomic_column_update> nonatomic_updates;
+};
+
+using set_of_changes = std::map<api::timestamp_type, batch>;
+
+set_of_changes extract_changes(const mutation& base_mutation, const schema& base_schema) {
+    auto extract_row_updates = [&base_schema] (const row& r, column_kind ckind) {
```
denesb: Why not make this a proper function? This is lambda abuse.

cdc/log.cc
```diff
@@ -416,6 +416,194 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+                        ttl,
+                        cr.key(),
+                        std::move(up.atomic_entries),
+                        std::move(up.nonatomic_deletions),
+                        std::move(up.nonatomic_updates)
+                    });
+            }
+        }
+
+        auto row_tomb_timestamp = cr.row().deleted_at().regular().timestamp;
```
denesb: You also need to consider the shadowable tombstone if cdc is ever to be enabled on a materialized view.

denesb: Hint: tombs have an `operator bool()`, which returns true if `timestamp != missing_timestamp`.

cdc/log.hh
```diff
@@ -123,7 +123,7 @@ struct db_context final {
 enum class operation : int8_t {
     // note: these values will eventually be read by a third party, probably not privvy to this
     // enum decl, so don't change the constant values (or the datatype).
-    pre_image = 0, update = 1, row_delete = 2, range_delete_start = 3, range_delete_end = 4, partition_delete = 5
+    pre_image = 0, insert = 1, update = 2, row_delete = 3, range_delete_start = 4, range_delete_end = 5, partition_delete = 6
```
denesb: Won't this cause mis-interpretation of existing data? Maybe append to the end instead?

cdc/log.cc
```diff
@@ -654,154 +715,111 @@ class transformer final {
...
-                    // TODO: separate inclusive/exclusive range
-                    set_operation(log_ck, ts, operation::range_delete_end, res);
-                    ++batch_no;
+
+            for (auto& change: btch.static_updates) {
+                // TODO: find a way to get a preimage for the static row, even if there are no rows
+                // corresponding to clustering keys affected by the mutation (in particular,
+                // even if the mutation affects no clustering rows at all)
+                // from CQL static columns can be queried specifying only the partition key in the WHERE clause
+                // --- it will return values for static columns even if there are no clustering rows
```
denesb: Any read to a partition using the internal APIs, using any slice will also return the static row.

test/boost/cdc_test.cc
```diff
@@ -48,87 +48,9 @@ static cql_test_config mk_cdc_test_config() {
 };
 
 namespace cdc {
-api::timestamp_type find_timestamp(const schema&, const mutation&);
 utils::UUID generate_timeuuid(api::timestamp_type);
 }
 
-SEASTAR_THREAD_TEST_CASE(test_find_mutation_timestamp) {
```
denesb: Why is this deleted?

denesb:
A general comment: please refrain from using acronyms outside the already established ones, like `up`, `ups`. These are hard to parse for those who didn't write the code.

cdc/log.cc
```diff
@@ -635,10 +542,164 @@ class transformer final {
...
     }
 
+    void set_regular_columns(
+            api::timestamp_type ts,
+            row_update&& row_up, const clustering_key& delta_ck,
+            const cql3::untyped_result_set_row* preimage_row, const std::optional<clustering_key>& preimage_ck,
+            column_kind ckind, mutation& log_mut) const {
+        for (auto& up: row_up.atomic_entries) {
+            auto& cdef = _schema->column_at(ckind, up.id);
+            auto& dst = *_log_schema->get_column_definition(to_bytes("_" + cdef.name()));
```
denesb: Shouldn't this be done in a central place to ensure consistency?

cdc/log.cc
kbr-: `extract_row_updates` suggests that `ups` = `updates`, but ok, I'll remove the abbreviation

cdc/log.cc
kbr-: `ttl` is a `duration`, not a `time_point`.
I used `rep` because we need to call `count()` eventually, but ok, `duration` will be better.

kbr-: Ah, I used `rep` so I don't need to write `gc_clock::duration(0)` in places like this:
```
                        view.is_live_and_has_ttl() ? view.ttl().count() : 0
```
this would instead have to be
```
view.is_live_and_has_ttl() ? view.ttl() : gc_clock::duration(0)
```

cdc/log.cc
kbr-: `nonatomic_updates` are for adding cells to nonatomic columns. E.g. suppose you have a `map<int, int> map_col`. You can perform:
```
update ... set map_col = map_col + {1:2}
```
This cannot be expressed in an insert.

In insert you can replace the entire value
```
insert into ... (..., map_col) values (..., {1:2, 3:4})
```
which underneath is translated to a tombstone with timestamp `X-1` and cells with timestamp `X`. So I split the above to two changes with different timestamps: one will be an insert with `nonatomic_deletion`, the other will be an update with `nonatomic_updates`.

cdc/log.cc
kbr-: true

cdc/log.cc
kbr-: I'd have to store an `atomic_cell_or_collection` there, and dispatch+unpack again in `transform` (i.e. call `as_atomic_cell` or `as_collection_mutation`, depending on `is_atomic()`). No need to do that, the proposed way is more convenient.

And I have to unpack nonatomics right there because the contained cells might have different timestamps and ttls.

The idea of `extract_changes` is to do all the unpacking, and later in `transform` we just consume the unpacked changes.

cdc/log.cc
kbr-: I didn't find a way to copy an `atomic_cell_view`. And even if I could, I would then want to return these views outside of `extract_changes`, which is unsafe.

The data has to eventually be copied anyway (when we put it into log table row cells in `set_regular_columns`). Might as well copy it right here.

cdc/log.cc
kbr-: true, will change

cdc/log.cc
kbr-: We won't use CDC on materialized views.
(and if we will, the code would probably need drastic changes anyway)

kbr-: Good point about the `operator bool()`

cdc/log.hh
kbr-: We're doing backward incompatible changes anyway. Note we're still `experimental`.
Even if I appended it at the end, it would still be a backward-incompatible change (because previously `insert`s were treated as `update`s)

test/boost/cdc_test.cc
kbr-: there is no `find_timestamp` function anymore

cdc/log.cc
denesb: :) Personally I'd still use `duration`, but I won't insist.

kbr-: I will change it to `duration`, using `rep` is an abstraction leak

test/boost/cdc_test.cc
denesb: Yes, but can't this test somehow be modified to test the new `extract_changes()` instead?

cdc/log.hh
denesb: We will still need to be able to update from a previous experimental version. Although maybe it is acceptable to ask the user to truncate.

test/boost/cdc_test.cc
kbr-: Yes, I'm planning to add a new test. It will test much more than just whether timestamps are extracted properly.

cdc/log.hh
kbr-: Even without this PR we won't be able to update 3.4 from 3.3 without disabling CDC on all tables first. (even truncate is not enough)

cdc/log.cc
kbr-: I checked experimentally: if you specify a clustering key in the bounds, and there is no clustered row with this clustering key, the static row won't be returned.

At least not from
```
    future<coordinator_query_result> storage_proxy::query(schema_ptr,
        lw_shared_ptr<query::read_command> cmd,
        dht::partition_range_vector&& partition_ranges,
        db::consistency_level cl,
        coordinator_query_options optional_params);
```

kbr-:
v3:
- use `gc_clock::duration` instead of `gc_clock::rep`
- add a comment in `clustered_row_insert` why there is not `nonatomic_updates` field
- make `extract_row_updates` a normal function, not a lambda
- query pre_image even if the mutation contains pratition/range deletions
- some minor naming changes

see [diff](https://github.com/scylladb/scylla/compare/ec5ff5d3e5a4351f6853cf4c1f84e470f2f65281..e8d8ed93bdc4c608f2af76090e5b017ac65b2b44).

cdc/log.cc
kbr-: What do you mean?

denesb: The `"_" + cdef.name()` computation. Isn't this the standard way to get the cdc table column name from the base column one?

cdc/log.cc
```diff
@@ -310,110 +310,211 @@ db_context db_context::builder::build() {
...
+    return utils::UUID_gen::get_random_time_UUID_from_micros(t);
+}
 
-    t = p.partition_tombstone().timestamp;
-    if (t != api::missing_timestamp) {
-        return t;
-    }
+struct atomic_column_update {
+    column_id id;
+    atomic_cell cell;
```
elcallio: Can't this be an atomic_cell_view? We are keeping the original mutation alive and non-moved during the operations, no?

cdc/log.cc
```diff
@@ -310,110 +310,211 @@ db_context db_context::builder::build() {
...
+struct nonatomic_column_deletion {
+    column_id id;
+};
 
-    auto walk_row = [&t, &s] (const row& r, column_kind ckind) {
-        r.for_each_cell_until([&t, &s, ckind] (column_id id, const atomic_cell_or_collection& cell) {
-            auto& cdef = s.column_at(ckind, id);
+struct nonatomic_column_update {
+    column_id id;
+    std::vector<std::pair<bytes, atomic_cell>> cells;
```
elcallio: Same view question here.

cdc/log.cc
```diff
@@ -310,110 +310,211 @@ db_context db_context::builder::build() {
...
-                t = cell.as_atomic_cell(cdef).timestamp();
-                if (t != api::missing_timestamp) {
-                    return stop_iteration::yes;
-                }
-                return stop_iteration::no;
-            }
+struct static_row_update {
+    gc_clock::duration ttl;
+    std::vector<atomic_column_update> atomic_entries;
+    std::vector<nonatomic_column_deletion> nonatomic_deletions;
```
elcallio: So this will not work as you hope I think. An update vector can still have individual cell deletions in it. So the very neat code you have below will eventually need to break things apart further. This is not a showstopper for this patch set though.

elcallio:
General comments: 
1.) There are a bunch of non-related changes in this PR it seems. You need to rebase and prune. 
2.) Is this approach (build a huge set of maps) the best way to approach this? While I do appreciate that there is literally no code from before left (sic), would not a forward-generator approach also work? I.e. create new mutations when discovered needed in the "original" loop, trying to minimize the amount of new structures allocated and copied (there will be a lot of data copying right now - esp since you do not use cell views). 
Since the need to break up things will probably be less super common, and when encountered, fairly easy (one new row or something), could you not just create new log rows iterating forward? It might create more resulting log rows in the complicated cases, but will maybe be quicker/cheaper in the common ones.

haaawk:
General comment. If you want to rename variables don't do that while you're changing the logic of the function. Now every single line is marked as changed even though many of them just have the name of variable changed. This makes the review very hard - the reviewer can't focus on important changes and has to decide for each line whether it was really changed or just the name of static_column was changed to static_column_ids.

cdc/log.cc
```diff
@@ -820,7 +838,7 @@ class transformer final {
             const mutation& m)
     {
         auto& p = m.partition();
-        if (p.partition_tombstone() || !p.row_tombstones().empty() || p.clustered_rows().empty()) {
+        if (p.clustered_rows().empty() && p.static_row().empty()) {
```
haaawk: Why is this good? I think we shouldn't generate preimage for partition deletion.

cdc/log.cc
```diff
@@ -845,21 +863,27 @@ class transformer final {
         std::transform(pc.begin(), pc.end(), std::back_inserter(columns), [](auto& c) { return &c; });
         std::transform(cc.begin(), cc.end(), std::back_inserter(columns), [](auto& c) { return &c; });
 
-        query::column_id_vector static_columns, regular_columns;
+        query::column_id_vector static_column_ids, regular_column_ids;
```
haaawk: Please don't rename variables like this when you change logic. If you really want to change the name do it afterward/beforehand.

cdc/log.cc
```diff
@@ -845,21 +863,27 @@ class transformer final {
...
 
-        auto sk = column_kind::static_column;
-        auto rk = column_kind::regular_column;
         // TODO: this assumes all mutations touch the same set of columns. This might not be true, and we may need to do more horrible set operation here.
-        for (auto& [r, cids, kind] : { std::tie(p.static_row().get(), static_columns, sk), std::tie(p.clustered_rows().begin()->row().cells(), regular_columns, rk) }) {
-            r.for_each_cell([&](column_id id, const atomic_cell_or_collection&) {
-                auto& cdef =_schema->column_at(kind, id);
-                cids.emplace_back(id);
+        if (!p.static_row().empty()) {
```
haaawk: again - why so intrusive change when it's not needed? Correct me if I'm wrong but you don't seem to change the logic here. If that's true why are you wasting reviewers time? You want to refactor the code without changing the behavior - do this in a separate commit and make it clear what it is. Don't blend it into a complex change.

PR #5816, opened 2020-02-16
dist/common/scripts/scylla_coredump_setup: bind-mount coredump directory, add coredump test

### LINKED ISSUE #5753: scylla_coredump_setup should bind-mount coredump dir on Ubuntu 18.04
Seen on enterprise release 2019.1.4

systemd-coredump on Ubuntu 18.04 does not work properly when using a symbolic link to replace `/var/lib/systemd/coredump` to `/var/lib/scylla/coredump`.
These lines are printed on the system log:
```
systemd-coredump@0-20226-0.service: Failed to set up special execution directory in /var/lib: Not a directory 
systemd-coredump@0-20226-0.service: Failed at step STATE_DIRECTORY spawning /lib/systemd/systemd-coredump: Not a directory
```

This does not happen when using `mount --bind  /var/lib/systemd/coredump /var/lib/systemd/coredump`.

This method can be applied on all linux distros but it requires either adding a corresponding line in /etc/fstab to bind-mount the directory on boot, or run `mount --bind` when the scylla service starts.

syuu1228:
On some environment systemd-coredump does not work with symlink directory,
we can use bind-mount instead.
Also, it's better to check systemd-coredump is working by generating coredump.

Fixes #5753

### DIFF
--- a/dist/common/scripts/scylla_coredump_setup
+++ b/dist/common/scripts/scylla_coredump_setup
@@ -24,6 +24,9 @@ import os
 import sys
 import argparse
 import subprocess
+import time
+import tempfile
+import subprocess
 from scylla_util import *
 
 if __name__ == '__main__':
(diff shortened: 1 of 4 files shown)


dist/common/scripts/scylla_coredump_setup
```diff
@@ -62,13 +65,35 @@ ExternalSizeMax=1024G
...
+                fstab = f.readlines()
+            for f in fstab:
+                if re.match(r'^/var/lib/scylla/coredump', f):
+                    print('/etc/fstab already has bind-mount entry', file=sys.stderr)
+                    sys.exit(1)
             makedirs('/var/lib/scylla/coredump')
-            os.symlink('/var/lib/scylla/coredump', '/var/lib/systemd/coredump')
+            with open('/etc/fstab', 'a') as f:
+                f.write('/var/lib/scylla/coredump /var/lib/systemd/coredump none defaults,bind 0 0\n')
+            run('mount /var/lib/systemd/coredump')
```
avikivity: It's safer to use a [systemd.mount](https://www.freedesktop.org/software/systemd/man/systemd.mount.html) file, we don't risk corrupting /etc/fstab. We can also set Before dependencies so it is set up before scylla is launched.

dist/common/scripts/scylla_coredump_setup
```diff
@@ -62,13 +65,35 @@ ExternalSizeMax=1024G
...
         run('systemctl daemon-reload')
         if is_debian_variant():
             run('sysctl -p /usr/lib/sysctl.d/50-coredump.conf')
         else:
             with open('/etc/sysctl.d/99-scylla-coredump.conf', 'w') as f:
                 f.write('kernel.core_pattern=|/usr/lib/systemd/systemd-coredump %p %u %g %s %t %e"')
             run('sysctl -p /etc/sysctl.d/99-scylla-coredump.conf')
+
+        fp = tempfile.NamedTemporaryFile()
+        fp.write(b'kill -11 $$')
```
avikivity: kill --SEGV is more portable, some archs have different signal numbers.

avikivity: Or maybe python's os.kill() is even simpler.

dist/common/scripts/scylla_coredump_setup
syuu1228: Updated the code using systemd.mount.

dist/common/scripts/scylla_coredump_setup
syuu1228: Updated the code using "-SEGV".

penberg:
Merged as 65aadad9a67a9a0868a912045495a3b413548f94.

PR #5822, opened 2020-02-17
Introduce schema::get_partitioner

### LINKED ISSUE #5493: Replace dht:global_partitioner() by a schema::get_partitioner().
Right now, the **partitioner** (i.e., function to convert partition keys into tokens) is global - it's not per-table or even per-keyspace. This is rarely a problem today, because the only recommended partitioner is `Murmur3Partitioner`. But for CDC log tables we would like to use a different partitioner (basically one which uses a 8-byte key directly as an 8-byte token) - and today cannot do it.

We should eliminate `dht::global_partitioner()` and replace it by a `schema::get_partitioner()` method. Making this a method of `schema` makes sense, in particular since what most code does with the returned partitioner is to call its `decorate_key()` or `get_token()`  methods, and both already require the schema as a parameter.

Since CQL does not (and initially, will not) have any way to set a non-default partitioner, we can begin the transition with the new `schema::get_partitioner()` simply always returning the default partitioner - what dht::global_partitioner() returns today. At a later stage `schema::get_partitioner()` can check for CDC log tables (for example) and return a different partitioner. Finally, if we ever want to support arbitrary per-table partitioners, we could add a way to configure (and persist the configuration) of a per-schema partitioner.

Related Cassandra issue: https://jira.apache.org/jira/browse/CASSANDRA-8143

haaawk:
Introduce schema::get_partitioner and use it instead of dht::global_partitioner.

The motivation for this change is a desire to have the ability to set a special partitioner for a selected table. To be able to do this, we need to be able to tell for each table what partitioner it uses. The most convenient way for this is to obtain such a partitioner from the schema. Even though a table can have multiple schemas over time, they all have to use the same partitioner. That would mean column_family is a better place for storing partitioner but the schema is much easier to access from all the places in the code.

The first step is to introduce schema::get_partitioner that still calls dht::global_partitioner internally and use this new function everywhere instead of dht::global_partitioner. Thanks to this, when later we change the implementation of schema::get_partitioner to something that allows setting different partitioners for different tables, it will already be used in all the places that depend on partitioner.

Fixes #5493

Tests: unit(dev, release, debug)

### DIFF
--- a/alternator/executor.cc
+++ b/alternator/executor.cc
@@ -1132,7 +1132,7 @@ std::optional<shard_id> rmw_operation::shard_for_execute(bool needs_read_before_
     // If we're still here, cas() *will* be called by execute(), so let's
     // find the appropriate shard to run it on:
     auto token = dht::global_partitioner().get_token(*_schema, _pk);
-    auto desired_shard = service::storage_proxy::cas_shard(token);
+    auto desired_shard = service::storage_proxy::cas_shard(*_schema, token);
     if (desired_shard == engine().cpu_id()) {
         return {};
     }
@@ -1496,7 +1496,7 @@ static future<> do_batch_write(service::storage_proxy& proxy,
         }
         return parallel_for_each(std::move(key_builders), [&proxy, &client_state, &stats, trace_state] (auto& e) {
             stats.write_using_lwt++;
-            auto desired_shard = service::storage_proxy::cas_shard(e.first.dk.token());
+            auto desired_shard = service::storage_proxy::cas_shard(*e.first.schema, e.first.dk.token());
             if (desired_shard == engine().cpu_id()) {
                 return cas_write(proxy, e.first.schema, e.first.dk, std::move(e.second), client_state, trace_state);
             } else {
--- a/alternator/executor.cc
+++ b/alternator/executor.cc
@@ -1068,7 +1068,7 @@ static lw_shared_ptr<query::read_command> read_nothing_read_command(schema_ptr s
 }
 
 static dht::partition_range_vector to_partition_ranges(const schema& schema, const partition_key& pk) {
-    return dht::partition_range_vector{dht::partition_range(dht::global_partitioner().decorate_key(schema, pk))};
+    return dht::partition_range_vector{dht::partition_range(dht::decorate_key(schema, pk))};
 }
 static dht::partition_range_vector to_partition_ranges(const dht::decorated_key& pk) {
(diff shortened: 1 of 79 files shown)


haaawk:
Removed a usage of global_partitioner added to alternator since last rebase.

avikivity:
What do I have to do to get you to put the motivation on the cover letter?

haaawk:
> What do I have to do to get you to put the motivation on the cover letter?

Is that enough? @avikivity

avikivity:
> > What do I have to do to get you to put the motivation on the cover letter?
> 
> Is that enough? @avikivity

It reads like someone forced you to do it rather than as if you're genuinely interested in explaining the work to a reader who isn't familiar with it.

haaawk:
> > > What do I have to do to get you to put the motivation on the cover letter?
> > 
> > 
> > Is that enough? @avikivity
> 
> It reads like someone forced you to do it rather than as if you're genuinely interested in explaining the work to a reader who isn't familiar with it.

I'm not much of a writer but I will try harder.

haaawk:
> > > > What do I have to do to get you to put the motivation on the cover letter?
> > > 
> > > 
> > > Is that enough? @avikivity
> > 
> > 
> > It reads like someone forced you to do it rather than as if you're genuinely interested in explaining the work to a reader who isn't familiar with it.
> 
> I'm not much of a writer but I will try harder.

Meeting the standards now, @avikivity ?

avikivity:
I already queued it yesterday as 6c7aa1823826817b38268871a0633a818cd98495. This is more for the future.

avikivity:
But yes, the new wording is much better. A patch description is not just communication between contributor and maintainer, but between a contributor and other reviewers who are not aware of the motivation, and between a contributor and future contributors who encounter a piece of code and which to understand how it came to be (perhaps as part of a bug hunt).

haaawk:
> I already queued it yesterday as [6c7aa18](https://github.com/scylladb/scylla/commit/6c7aa1823826817b38268871a0633a818cd98495). This is more for the future.

I know - it does not mean I run away and never touch the PR again :)

haaawk:
> But yes, the new wording is much better. A patch description is not just communication between contributor and maintainer, but between a contributor and other reviewers who are not aware of the motivation, and between a contributor and future contributors who encounter a piece of code and which to understand how it came to be (perhaps as part of a bug hunt).

I know. Sometimes I want to go too fast.

PR #5835, opened 2020-02-18
hinted handoff: send orphaned counter hints correctly

### LINKED ISSUE #5833: Hinted handoff sends counter hints as counter updates when node targeted by the hint does not exist
Counter mutations have two flavors: counter update and the total state of the counter.

Hints for counters are mutations that store the total state of the counter and should be treated as normal mutations. There is a corner case when hinted handoff misinterprets the counter hint mutation and treats it as counter update. As a result, a counter is updated with a value that is his state. This means the value is corrupted.

The problematic case happens when a node that's a destination for a hint is no longer in the cluster. In such a case, the hint is sent to all the remaining replicas. This code path wrongly uses storage_proxy::mutate which expects counter update mutation.

This problem can cause corruption in any counter shard because storage_proxy::mutate picks a leader for the update and it can be any replica.

The problem was partially fixed on master by https://github.com/scylladb/scylla/commit/01084a79b8f47f63aeb54bd110135f3850ad3fc8
We need to fix it fully with an additional fix.

piodul:
This patch fixes a bug that appears because of an incorrect interaction between counters and hinted handoff.

When a counter is updated on the leader, it sends mutations to other replicas that contain all counter shards from the leader. If consistency level is achieved but some replicas are unavailable, a hint with mutation containing counter shards is stored.

When a hint's destination node is no longer its replica, it is attempted to be sent to all its current replicas. Previously, storage_proxy::mutate was used for that purpose. It was incorrect because that function treats mutations for counter tables as mutations containing only a delta (by how much to increase/decrease the counter). These two types of mutations have different serialization format, so in this case a "shards" mutation is reinterpreted as "delta" mutation, which can cause data corruption to occur.

This patch backports `storage_proxy::mutate_hint_from_scratch` function, which bypasses special handling of counter mutations and treats them as regular mutations - which is the correct behavior for "shards" mutations.

Refs #5833.
Tests: unit(dev), dtest(hintedhandoff_additional_test)

### DIFF
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -2319,6 +2319,14 @@ future<> storage_proxy::send_hint_to_endpoint(frozen_mutation_and_schema fm_a_s,
             allow_hints::no);
 }
 
+future<> storage_proxy::mutate_hint_from_scratch(frozen_mutation_and_schema fm_a_s) {
+    // FIXME: using 1h as infinite timeout. If a node is down, we should get an
+    // unavailable exception.
+    const auto timeout = db::timeout_clock::now() + 1h;
+    std::array<mutation, 1> ms{fm_a_s.fm.unfreeze(fm_a_s.s)};
+    return mutate_internal(std::move(ms), db::consistency_level::ALL, false, nullptr, empty_service_permit(), timeout);
+}
+
 /**
  * Send the mutations to the right targets, write it locally if it corresponds or writes a hint when the node
  * is not available.
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -405,11 +405,8 @@ future<> manager::end_point_hints_manager::sender::do_send_one_mutation(frozen_m
             return _proxy.send_hint_to_endpoint(std::move(m), end_point_key());
         } else {
             manager_logger.trace("Endpoints set has changed and {} is no longer a replica. Mutating from scratch...", end_point_key());
-            // FIXME: using 1h as infinite timeout. If a node is down, we should get an
-            // unavailable exception.
-            auto timeout = db::timeout_clock::now() + 1h;
             //FIXME: Add required frozen_mutation overloads
-            return _proxy.mutate({m.fm.unfreeze(m.s)}, consistency_level::ALL, timeout, nullptr, empty_service_permit());
+            return _proxy.mutate_hint_from_scratch(std::move(m));
         }
     });
 }
--- a/service/storage_proxy.hh
+++ b/service/storage_proxy.hh
@@ -471,6 +471,8 @@ public:
     */
     future<> mutate_atomically(std::vector<mutation> mutations, db::consistency_level cl, clock_type::time_point timeout, tracing::trace_state_ptr tr_state, service_permit permit);
 
+    future<> mutate_hint_from_scratch(frozen_mutation_and_schema fm_a_s);
+
     // Send a mutation to one specific remote target.
     // Inspired by Cassandra's StorageProxy.sendToHintedEndpoints but without
     // hinted handoff support, and just one target. See also


service/storage_proxy.cc
```diff
@@ -2319,6 +2319,13 @@ future<> storage_proxy::send_hint_to_endpoint(frozen_mutation_and_schema fm_a_s,
             allow_hints::no);
 }
 
+future<> storage_proxy::mutate_hint_from_scratch(frozen_mutation_and_schema fm_a_s) {
```
gleb-cloudius: What does "from_scratch" means?

service/storage_proxy.cc
```diff
@@ -2319,6 +2319,13 @@ future<> storage_proxy::send_hint_to_endpoint(frozen_mutation_and_schema fm_a_s,
...
+    // FIXME: using 1h as infinite timeout. If a node is down, we should get an
+    // unavailable exception.
+    const auto timeout = db::timeout_clock::now() + 1h;
```
gleb-cloudius: 1 hour timeout again?

service/storage_proxy.cc
piodul: In this context it means that the whole effort of writing this mutation to all replicas is repeated. I used the terminology from the log message above the place where this function is called.

gleb-cloudius: I think more descriptive name that make it clear that a mutation is sent to all owning replicas would be better.

service/storage_proxy.cc
piodul: I'm not sure I want to change the timeout in this patch. This is a backport, and the intent is to fix a data-corrupting behavior.

I agree that the timeout should be shorter, though, and I can see that I forgot to change it during review of https://github.com/scylladb/scylla/pull/5668 . How about a separate patch that can be backported after this?

gleb-cloudius: OK.

haaawk: That's the behavior before this PR. This change does not change the timeout length and since it's a bugfix for other problems we probably should not fix change timeouts here.

service/storage_proxy.cc
haaawk: Any particular name you have in mind @gleb-cloudius ?

gleb-cloudius: `send_hint_to_all_replicas()`

piodul: Like with the timeout, I think we can rename this function on master and then backport the change (on master it is called the same as here). Leaving the function name as it is here will make life easier when backporting.

gleb-cloudius: Backporting what? This is a backport already.

haaawk: This is a limited backport of https://github.com/scylladb/scylla/commit/01084a79b8f47f63aeb54bd110135f3850ad3fc8

If we change the name here then it will cause conflicts for future backports. As you correctly noticed @gleb-cloudius, this is already a backport so maybe we shouldn't change too much. If we want to change something unrelated then it's better to submit it to master and backport again.

gleb-cloudius: We will definitely not backport function name change. Well I guess it is my fault I did not complaine about the name on the original submission (although I do remember scratching my head when I saw the function name).

haaawk: You have a fair point so let's rename on master.

piodul: Understood. In that case, I changed the name to `send_hint_to_all_replicas`, as you suggested.

gleb-cloudius: But please send the rename to the master as well.

slivne:
Please add branches - so it will be clear which branch needs this fix

avikivity:
@piodul I see you renamed something, this is not correct practice. The patch should be a minimal backport of 434347195416cc79211e4b7df001f90b0f1aded7, any deviation is a risk of regression (or perhaps an additional fix that is not present in master).

Please change it to be a minimal backport. Further work like renaming for clarity, which is welcome, should be done on master.

Also reference the original commit by including its hash with an explanation in the patch changelog. `git cherry-pick -x` does this automatically by including the message "(cherry-picked-from <hash>)".

gleb-cloudius:
On Wed, Feb 19, 2020 at 05:42:43AM -0800, Avi Kivity wrote:
> @piodul I see you renamed something, this is not correct practice. The patch should be a minimal backport of 434347195416cc79211e4b7df001f90b0f1aded7, any deviation is a risk of regression (or perhaps an additional fix that is not present in master).
> 
> Please change it to be a minimal backport. Further work like renaming for clarity, which is welcome, should be done on master.
> 
> Also reference the original commit by including its hash with an explanation in the patch changelog. `git cherry-pick -x` does this automatically by including the message "(cherry-picked-from <hash>)".
> 
I asked for rename, but I asked to rename it on the master first.

--
			Gleb.

PR #5845, opened 2020-02-19
dist/docker: Add VERSION argument to Dockerfile

yaronkaikov:
Currently, the `Dockerfile` installs the latest version of Scylla. Let's
add a VERSION argument to Dockerfile, which explicitly specifies the
version to ensure scripts, for example, always build the expected
version. If no VERSION is specified for "docker build", use the default
value of "666.development", which is the version number for latest
nightly.

### DIFF
--- a/dist/docker/redhat/Dockerfile
+++ b/dist/docker/redhat/Dockerfile
@@ -6,7 +6,7 @@ ENV container docker
 
 # The SCYLLA_REPO_URL argument specifies the URL to the RPM repository this Docker image uses to install Scylla. The default value is the Scylla's unstable RPM repository, which contains the daily build.
 ARG SCYLLA_REPO_URL=http://downloads.scylladb.com/rpm/unstable/centos/master/latest/scylla.repo
-
+ARG VERSION=666.development
 ADD scylla_bashrc /scylla_bashrc
 
 # Scylla configuration:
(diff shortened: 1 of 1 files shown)


yaronkaikov:
Merge only after https://github.com/scylladb/scylla-pkg/pull/1015

dist/docker/redhat/Dockerfile
```diff
@@ -6,7 +6,7 @@ ENV container docker
 
 # The SCYLLA_REPO_URL argument specifies the URL to the RPM repository this Docker image uses to install Scylla. The default value is the Scylla's unstable RPM repository, which contains the daily build.
 ARG SCYLLA_REPO_URL=http://downloads.scylladb.com/rpm/unstable/centos/master/latest/scylla.repo
-
+ARG VERSION
```
avikivity: Good idea. But please add a default (666) so it also works without arguments.

yaronkaikov: @avikivity It idea was to make sure the docker will always get the right Scylla version that match the relevant centos-rpm build. We can put a default, but then, we can have a situation that we have a new docker image with older Scylla version inside.

avikivity: I want to preserve the ability to do a "docker build -f dist/docker/redhat .", not everyone can go through jenkins.

penberg: I agree, @yaronkaikov can you please update the pull request?

yaronkaikov: Yes. so default value will be `666.development`?

yaronkaikov: Changed and tested locally

penberg:
Merged as https://github.com/scylladb/scylla/commit/de19496ff7a1aeeb22096006ef2309270e9255e2

Issue #5495, opened 2019-12-18
Performance dropped in twice after node decomission


dkropachev:
Performance dropped in twice after node decomission
-----------------------------------------------------------------------------------------
Scylla version (or git commit hash): scylla-3.2.rc3-0.20191215.39b17be5620.el7.x86_64
Jenkins job: https://jenkins.scylladb.com/view/scylla-3.2/job/scylla-3.2/view/scylla-3.2-test/job/longevity/job/longevity-multi-keyspaces-60h/2/
test_name: 'longevity_test.LongevityTest.test_custom_time'
test_config: 'test-cases/longevity/longevity-multi-keyspaces.yaml'
Cluster size: 10
OS (RHEL/CentOS/Ubuntu/AWS AMI): (AWS region=eu-west-1) jenkins-scylla-3.2-ami-test-24-24 (ami-0cab4c6eceb3249ac)
Graphana snapshot: https://snapshot.raintank.io/dashboard/snapshot/pKaK9IJmWyMwaI6vlG0nl5y52xOqtz5X?orgId=2

backtrace:
`
0x0000000002accbd2
0x0000000002a4579b
0x0000000002a45cc2
0x0000000002a45ff7
0x00007ff0a609be7f
0x0000000001b0b500
0x0000000001b03185
0x0000000001af0d41
0x0000000001af027a
0x0000000001f7e89a
0x0000000001f9f55a
0x0000000001fc9c09
0x0000000001fcac08
0x00000000007dfee3
`
translated backtrace:
`
void seastar::backtrace<seastar::backtrace_buffer::append_backtrace()::{lambda(seastar::frame)#1}>(seastar::backtrace_buffer::append_backtrace()::{lambda(seastar::frame)#1}&&) at /usr/include/c++/9/ext/new_allocator.h:105
seastar::backtrace_buffer::append_backtrace() at /usr/include/c++/9/ext/new_allocator.h:105
 (inlined by) print_with_backtrace at /jenkins/slave/workspace/scylla-3.2/build/scylla/seastar/src/core/reactor.cc:731
seastar::internal::cpu_stall_detector::generate_trace() at /usr/include/c++/9/ext/new_allocator.h:105
seastar::internal::cpu_stall_detector::maybe_report() at /usr/include/c++/9/ext/new_allocator.h:105
[...]

Issue #5226, opened 2019-10-27
Off-Strategy Compaction


slivne:
From: @avikivity 

A compaction strategy attempts to serve user workloads with certain characteristics by maintaining certain invariants that guarantee some performance level or other metric. For example, Time Window Compaction Strategy is tuned for time series workloads, maintains the invariant that no sstables cross time windows, and guarantees that the write amplification is O(log(window size / memory size)) rather than size-tiered’s O(log(data size / memory size)). The general-purpose size-tiered compaction strategy maintains the invariant that the number of sstables of similar size is low, and so compaction strategy component of read amplification and write amplification are both O(log(disk size / memory size)).

We define Off-Strategy Compactions as a compaction happening at a state where the invariant is broken. The invariant can be broken for the following reasons:

* Resharding
* Bootstrap
* Repair
* Importing sstables (“nodetool refresh”)
* Upgrade SSTables (“nodetool upgradesstables”)
* Altering the compaction strategy
* User load that does not fit the compaction strategy (for example, non-time-series data with time windowed compaction strategy)

While the last case is a user error, the other cases happen during normal or semi-normal operation and will result in degraded performance, without any ability on the user’s part to mitigate the effects.

**Types of off-strategy compactions**
* Resharding - Resharding happens when a node is restarted with a different number of active processors (shards), or if some other partitioner parameter has changed. Since partitioning to shards is static, data must be moved to the new shards; usually requiring a full read and write of the entire data set. While resharding is happening, all reads have to consider all sstables, not just the shard’s sstables, resulting in O(number of shards) additional read amplification. While the bloom filter will reduce this for single-partition reads, range scans will have to read all of those sstables, resulting in lower throughput and lower concurrency (due to increased memory consumption).
* Bootstrap/repair - The bootstrap and repair processes are vnode-oriented, so after they complete, the node will have 256 additional sstables. Those new sstables participate in reads, causing read amplification and increased memory consumption. 
* Changing compaction strategy - When the user changes compaction strategy, the existing sstables do not conform to the new compaction strategy’s invariants, and therefore the new compaction strategy may perform poorly.
* Importing sstables - In the general case importing sstables is similar to resharding and changing the compaction strategy. The imported sstables are not properly sharded, and may have been generated using a different compaction strategy (or no compaction strategy at all).

**Proposed Solution:**

We will enhance the usage of the staging  directory (currently used by Materialized Views code_ - so that all this cases will place the ssatbles in the staging directory.
[...]

Issue #5199, opened 2019-10-22
Better compaction for sstables generated by maintenance operations


asias:
When we run a node maintenance operation, e.g., bootstrap, repair, a set of new sstables will be created. For a given token range for a given table, typically a vnode token range, streaming will create 1 new sstables per shard, repair will create 1 new sstables per shard on repair followers and RF - 1 new sstables per shard on repair master. Those sstables do not have the invariant properties defined by the compaction strategy. 

Currently, we add those new sstables to the main sstables set directly and let the normal compaction to compact and "reshape" those sstables. As a result, we effectively push the maintenance work to the compaction scheduling group, not running under maintenance scheduling group, having impact on the user read performance. In addition, those sstables without the invariant properties defined by the compaction strategy may cause bad read performance.

slivne:
merged with https://github.com/scylladb/scylla/issues/5226

Issue #5109, opened 2019-09-26
Streaming during bootstrapping is very slow due to compactions


vladzcloudius:
*Installation details*
Scylla version: 2019.1.2
Cluster size: 4
Platform: AWS
Hardware: i3el.3xl

*Description*
When a new node is joining it naturally becomes a bottleneck since it's a one-to-many scenario. However compactions that are kicking in due to incoming streaming add to the insult and streaming that could have taken less than two hours takes 12 hours. The reason is compactions taking most of CPU runtime on the joining node.

@gnumoreno Please, could you add some screenshots, please?

If we could reduce the compactions load joining would take much less time. 
In most cases the speed of adding a new node is much more important to users than the probably some higher disk utilization during a limited amount of time or some reasonably higher read latency as long as it's temporary.

We could start with estimating how big the impact is going to be if we, for instance, disable compactions during streaming and enable it back right after node joins.

If we see that it doesn't bring a terribly high impact on cluster's performance we may want to implement some automatic scheme that would, for instance, reduce compactions' shares to a bare minimum as long as a node in a JOINING state.

@avikivity @glommer What do you think?

glommer:
During bootstrap and replace, the node receiving the data should increase its streaming priority to be much higher than compactions.

This should be a relatively easy patch to write.

mailmahee:
and also the priority of compactions while bootstrapping can be reduced to as low as possible.

glommer:
Mahee, is my son getting bigger or am I getting smaller?

gnumoreno:
@vladzcloudius I will send you via e-mail (the screenshots).

asias:
Here is a minimal PoC patch that reduces the compaction io priority. 

```
From 13f63d10576961a4b6e6120007a6864ca84a45f6 Mon Sep 17 00:00:00 2001
Message-Id: <13f63d10576961a4b6e6120007a6864ca84a45f6.1570780243.git.asias@scylladb.com>
From: Asias He <asias@scylladb.com>
Date: Wed, 9 Oct 2019 20:48:46 +0800
Subject: [PATCH scylla] range_streamer: Less compaction during bootstrap
To: scylladb-dev@googlegroups.com

Fixes #5109
---
 dht/range_streamer.hh |  2 +-
 dht/boot_strapper.cc  |  3 ++-
 dht/range_streamer.cc | 17 +++++++++++++++--
 3 files changed, 18 insertions(+), 4 deletions(-)

[...]

PR #5866, opened 2020-02-21
table: Do not trigger compaction in add_sstable_and_update_cache

asias:
The table::add_sstable_and_update_cache is used by streaming and repair to
add sstables generated during the operation.

When we run a node maintenance operation, e.g., bootstrap, repair, a set
of new sstables will be created. For a given token range for a given
table, typically a vnode token range, streaming will create 1 new
sstables per shard, repair will create 1 new sstables per shard on
repair followers and RF - 1 new sstables per shard on repair master.

Currently, we add those new sstables to the main sstables set directly
and let the normal compaction to compact and "reshape" those sstables.
As a result, we effectively push the maintenance work to the compaction
scheduling group, not running under maintenance scheduling group, having
impact on the user read performance. In addition, those sstables without
the invariant properties defined by the compaction strategy may cause
bad read performance.

Before we have the Off-Strategy Compaction which will take a while to be
there, we can mitigate the problem by not triggering a compaction
every time we add a stream/repair generated sstable.

Refs: #5495 #5226 #5199 #5109 #4884

### DIFF
--- a/dht/range_streamer.cc
+++ b/dht/range_streamer.cc
@@ -47,6 +47,7 @@
 #include "streaming/stream_state.hh"
 #include "service/storage_service.hh"
 #include "db/config.hh"
+#include "message/messaging_service.hh"
 #include <seastar/core/semaphore.hh>
 #include <boost/range/adaptors.hpp>
 
@@ -310,8 +311,12 @@ future<> range_streamer::do_stream_async() {
                             description, source, keyspace, nr_ranges_streamed, nr_ranges_total, ranges_to_stream.size());
                     nr_ranges_streamed += ranges_to_stream.size();
                     if (_nr_rx_added) {
+                        // This node is the stream receiver, trigger compaction on local node only
+                        _trigger_compaction[keyspace].insert(_address);
                         sp.request_ranges(source, keyspace, ranges_to_stream);
                     } else if (_nr_tx_added) {
+                        // This node is the stream sender, trigger compaction on remote node.
+                        _trigger_compaction[keyspace].insert(source);
                         sp.transfer_ranges(source, keyspace, ranges_to_stream);
                     }
                     sp.execute().discard_result().get();
@@ -342,6 +347,22 @@ future<> range_streamer::do_stream_async() {
                 logger.info("{} with {} for keyspace={} succeeded, took {} seconds", description, source, keyspace, t);
               });
           });
+        });
+    }).then([this, start] {
+        return do_for_each(_trigger_compaction, [] (auto& trigger) {
+            const sstring& keyspace = trigger.first;
+            const std::unordered_set<gms::inet_address>& nodes = trigger.second;
+            return parallel_for_each(nodes, [&keyspace] (const gms::inet_address& node) {
+                logger.info("Trigger compaction on node {} for keyspace={}", node, keyspace);
+                return netw::get_local_messaging_service().send_stream_mutation_done(netw::msg_addr(node), {}, {}, {}, 0, keyspace
+                        ).handle_exception([node, keyspace] (std::exception_ptr ep) {
+                    // In a mixed cluster, node that does not support trigger
+                    // compaction will fail the verb to trigger the compaction.
+                    // It is fine because they already trigger the compaction
+                    // when the stream session is finished in the old way.
+                    logger.warn("Trigger compaction on node {} for keyspace={} failed:{}", node, keyspace, ep);
+                });
+            });
         });
     }).finally([this, start] {
         auto t = std::chrono::duration_cast<std::chrono::seconds>(lowres_clock::now() - start).count();
--- a/database.hh
+++ b/database.hh
@@ -546,7 +546,7 @@ private:
     db_clock::time_point _truncated_at = db_clock::time_point::min();
 
 public:
-    future<> add_sstable_and_update_cache(sstables::shared_sstable sst);
+    future<> add_sstable_and_update_cache(sstables::shared_sstable sst, bool trigger_compaction = true);
     future<> move_sstables_from_staging(std::vector<sstables::shared_sstable>);
     sstables::shared_sstable get_staging_sstable(uint64_t generation) {
         auto it = _sstables_staging.find(generation);
--- a/dht/range_streamer.hh
+++ b/dht/range_streamer.hh
@@ -173,6 +173,7 @@ private:
     streaming::stream_reason _reason;
     std::unordered_multimap<sstring, std::unordered_map<inet_address, dht::token_range_vector>> _to_stream;
     std::unordered_set<std::unique_ptr<i_source_filter>> _source_filters;
+    std::unordered_map<sstring, std::unordered_set<inet_address>> _trigger_compaction;
     stream_plan _stream_plan;
     // Retry the stream plan _nr_max_retry times
     unsigned _nr_retried = 0;
--- a/message/messaging_service.cc
+++ b/message/messaging_service.cc
@@ -890,17 +890,17 @@ future<> messaging_service::send_stream_mutation(msg_addr id, UUID plan_id, froz
 
 // STREAM_MUTATION_DONE
 void messaging_service::register_stream_mutation_done(std::function<future<> (const rpc::client_info& cinfo,
-        UUID plan_id, dht::token_range_vector ranges, UUID cf_id, unsigned dst_cpu_id)>&& func) {
+        UUID plan_id, dht::token_range_vector ranges, UUID cf_id, unsigned dst_cpu_id, rpc::optional<sstring> trigger_compaction)>&& func) {
     register_handler(this, messaging_verb::STREAM_MUTATION_DONE,
             [func = std::move(func)] (const rpc::client_info& cinfo,
                     UUID plan_id, std::vector<wrapping_range<dht::token>> ranges,
-                    UUID cf_id, unsigned dst_cpu_id) mutable {
-        return func(cinfo, plan_id, ::compat::unwrap(std::move(ranges)), cf_id, dst_cpu_id);
+                    UUID cf_id, unsigned dst_cpu_id, rpc::optional<sstring> trigger_compaction) mutable {
+        return func(cinfo, plan_id, ::compat::unwrap(std::move(ranges)), cf_id, dst_cpu_id, trigger_compaction);
     });
 }
-future<> messaging_service::send_stream_mutation_done(msg_addr id, UUID plan_id, dht::token_range_vector ranges, UUID cf_id, unsigned dst_cpu_id) {
+future<> messaging_service::send_stream_mutation_done(msg_addr id, UUID plan_id, dht::token_range_vector ranges, UUID cf_id, unsigned dst_cpu_id, sstring trigger_compaction) {
     return send_message<void>(this, messaging_verb::STREAM_MUTATION_DONE, id,
-        plan_id, std::move(ranges), cf_id, dst_cpu_id);
+        plan_id, std::move(ranges), cf_id, dst_cpu_id, std::move(trigger_compaction));
 }
 
 // COMPLETE_MESSAGE
--- a/message/messaging_service.hh
+++ b/message/messaging_service.hh
@@ -304,8 +304,8 @@ public:
     rpc::sink<repair_hash_with_cmd> make_sink_for_repair_get_full_row_hashes_with_rpc_stream(rpc::source<repair_stream_cmd>& source);
     void register_repair_get_full_row_hashes_with_rpc_stream(std::function<future<rpc::sink<repair_hash_with_cmd>> (const rpc::client_info& cinfo, uint32_t repair_meta_id, rpc::source<repair_stream_cmd> source)>&& func);
 
-    void register_stream_mutation_done(std::function<future<> (const rpc::client_info& cinfo, UUID plan_id, dht::token_range_vector ranges, UUID cf_id, unsigned dst_cpu_id)>&& func);
-    future<> send_stream_mutation_done(msg_addr id, UUID plan_id, dht::token_range_vector ranges, UUID cf_id, unsigned dst_cpu_id);
+    void register_stream_mutation_done(std::function<future<> (const rpc::client_info& cinfo, UUID plan_id, dht::token_range_vector ranges, UUID cf_id, unsigned dst_cpu_id, rpc::optional<sstring> trigger_compaction)>&& func);
+    future<> send_stream_mutation_done(msg_addr id, UUID plan_id, dht::token_range_vector ranges, UUID cf_id, unsigned dst_cpu_id, sstring trigger_compaction = {});
 
     void register_complete_message(std::function<future<> (const rpc::client_info& cinfo, UUID plan_id, unsigned dst_cpu_id, rpc::optional<bool> failed)>&& func);
     future<> send_complete_message(msg_addr id, UUID plan_id, unsigned dst_cpu_id, bool failed = false);
--- a/streaming/stream_session.cc
+++ b/streaming/stream_session.cc
@@ -218,22 +218,25 @@ void stream_session::init_messaging_service_handler() {
                         make_generating_reader(s, std::move(get_next_mutation_fragment)),
                         [plan_id, estimated_partitions, reason] (flat_mutation_reader reader) {
                             auto& cf = get_local_db().find_column_family(reader.schema());
-                            return db::view::check_needs_view_update_path(_sys_dist_ks->local(), cf, reason).then([cf = cf.shared_from_this(), estimated_partitions, reader = std::move(reader)] (bool use_view_update_path) mutable {
+                            return db::view::check_needs_view_update_path(_sys_dist_ks->local(), cf, reason).then([cf = cf.shared_from_this(), estimated_partitions, reader = std::move(reader), reason] (bool use_view_update_path) mutable {
                                 //FIXME: for better estimations this should be transmitted from remote
                                 auto metadata = mutation_source_metadata{};
                                 auto& cs = cf->get_compaction_strategy();
                                 const auto adjusted_estimated_partitions = cs.adjust_partition_estimate(metadata, estimated_partitions);
                                 auto consumer = cf->get_compaction_strategy().make_interposer_consumer(metadata,
-                                        [cf = std::move(cf), adjusted_estimated_partitions, use_view_update_path] (flat_mutation_reader reader) {
+                                        [cf = std::move(cf), adjusted_estimated_partitions, use_view_update_path, reason] (flat_mutation_reader reader) {
                                     sstables::shared_sstable sst = use_view_update_path ? cf->make_streaming_staging_sstable() : cf->make_streaming_sstable_for_write();
                                     schema_ptr s = reader.schema();
                                     auto& pc = service::get_local_streaming_write_priority();
 
                                     return sst->write_components(std::move(reader), std::max(1ul, adjusted_estimated_partitions), s,
-                                                                 sstables::sstable_writer_config{}, encoding_stats{}, pc).then([sst] {
+                                                                 sstables::sstable_writer_config{}, encoding_stats{}, pc).then([sst, reason] {
                                         return sst->open_data();
-                                    }).then([cf, sst] {
-                                        return cf->add_sstable_and_update_cache(sst);
+                                    }).then([cf, sst, reason] {
+                                        // Trigger compaction for repair. Compaction for other stream reasons
+                                        // will be triggered when the operation is done.
+                                        bool trigger_compaction = reason == stream_reason::repair;
+                                        return cf->add_sstable_and_update_cache(sst, trigger_compaction);
                                     }).then([cf, s, sst, use_view_update_path]() mutable -> future<> {
                                         if (!use_view_update_path) {
                                             return make_ready_future<>();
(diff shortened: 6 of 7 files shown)


glommer:
hi Asias.

In general I like this direction. It essentially means that compactions will happen only through normal writes and at that point, compact with the streamed SSTables.

The next step for off strategy compaction is not to include those SSTables in compactions with the main SSTable set.

However, I fear that in a low-write scenario, we will still be left indefinitely with a high-numbered SSTable set. So we should replace the lack of triggers on individual SSTables with a single trigger at the end of the whole process.

Also, you referenced a performance issue. Do you have numbers on improvements coming from this?
I'd be interested in measuring bootstrap times, as well as repair times concurrent with reads to make sure the high number of SSTables are not affecting reads too much.

asias:
> hi Asias.
> 
> In general I like this direction. It essentially means that compactions will happen only through normal writes and at that point, compact with the streamed SSTables.

I am wondering any other point we could compaction the streamed sstables after this patch beyond the write point? `nodetool compact` anything else?

> The next step for off strategy compaction is not to include those SSTables in compactions with the main SSTable set.

Yes.

> 
> However, I fear that in a low-write scenario, we will still be left indefinitely with a high-numbered SSTable set. So we should replace the lack of triggers on individual SSTables with a single trigger at the end of the whole process.

I thought about this too. However it is tricky to define the end of the whole process. For example, scylla manager wants to repair 100 ranges. It sends 10 ranges a time, creating a repair instance, from the repair point of view, the whole process ends after the 10 ranges are repaired.  The real whole process ends when the whole 100 ranges are repaired. It is similar for bootstrap. 

> 
> Also, you referenced a performance issue. Do you have numbers on improvements coming from this?
> I'd be interested in measuring bootstrap times, as well as repair times concurrent with reads to make sure the high number of SSTables are not affecting reads too much.

I do not have any numbers yet. @amoskong has a SCT test to test bootstrap performance, we can probably reuse it when we want to test.

glommer:
Each range creates one SSTable, right? Or is one SSTable created per group of 10 ranges ?

asias:
> Each range creates one SSTable, right? Or is one SSTable created per group of 10 ranges ?

For repair, one sstable per range. For stream, one sstable per 10 ranges group.

glommer:
So for repair we can at least compact at the end of every range group.
And for bootstrap the end of operation is well defined.

asias:
> So for repair we can at least compact at the end of every range group.
> And for bootstrap the end of operation is well defined.

Isn't it better if we trigger the compaction at the end all the ranges are repaired and streamed. For streaming, scylla knows when is that. For repair, scylla does not know because the next groups of ranges are decided by the user which may come or not. Also, the manager can send 1 range per group as well, as as result, we trigger for each range. 

Perhaps, we can trigger it automatically, e.g., if there are not new streaming or repair sstable writes for some time, we trigger it.

vladzcloudius:
@glommer repair case: compacting at the end of each range will give the same situation as we have today since we generate a single sstable per range according to @asias.

I think that instead of inventing heuristics it would be easier and more robust to make it explicit and send a special RPC verb that would denote moving all excising sstables into the compaction set and trigger a compaction round.

vladzcloudius:
@asias I join @glommer in the question about test results.  
I wonder how much this single-liner helps?

glommer:
Adding rpc verbs defeats the purpose of this patch, which is just a quick fix before we move onwards with the real thing. When we have offstrategy compaction, we shall need no such verb.

For bootstrap, it's fine to call trigger_compaction at the very end of it.
An for repair maybe leave this be for now.

Note that "trigger_compaction", despite what the name says, do not actually trigger a compaction.
It just tells the compaction manager that there is an SSTable that is new and it might compact (most of the time it will)

vladzcloudius:
> Adding rpc verbs defeats the purpose of this patch, which is just a quick fix before we move onwards with the real thing. When we have offstrategy compaction, we shall need no such verb.

I'm not sure I understand how you can avoid having such a verb.
You need to be able to coordinate the end of the repair process for example.
If you follow the implementation proposal in the corresponding github issue how will you know when to move streamed sstables to `data` from that staging directory?

glommer:
We are not ready yet with the design of offstrategy. Maybe we will need a verb, maybe we won't.
Adding one now just so we can add this simple fix for Asias, however, is certainly not acceptable.

vladzcloudius:
@glommer I agree that we can start with something simple: enable this patch only for processes which involve a single node as a streaming receiver and which end state is easily defined, e.g. bootstrap or rebuild.

Extending this to a process like a repair will require more work.

> We are not ready yet with the design of offstrategy. Maybe we will need a verb, maybe we won't.
> Adding one now just so we can add this simple fix for Asias, however, is certainly not acceptable.

I think "acceptable" is too strong of a word for this context. If we wanted to include repairs in the context of this patch adding a new verb (or some other sync machinery)  would definitely be acceptable.

However we can definitely get along without a repair (and this verb) for now because repairs are much less of a problem since they usually don't generate the same amount of sstables like regular streaming (e.g. bootstrapping) .

asias:
> @glommer repair case: compacting at the end of each range will give the same situation as we have today since we generate a single sstable per range according to @asias.
> 
> I think that instead of inventing heuristics it would be easier and more robust to make it explicit and send a special RPC verb that would denote moving all excising sstables into the compaction set and trigger a compaction round.

@vladzcloudius The bigger problem here is not adding a verb or not. The bigger problem is only user knows if more repair requests are going to be executed. The heuristics solves this bigger problem.

asias:
> @glommer I agree that we can start with something simple: enable this patch only for processes which involve a single node as a streaming receiver and which end state is easily defined, e.g. bootstrap or rebuild.
> 
> Extending this to a process like a repair will require more work.
> 
> > We are not ready yet with the design of offstrategy. Maybe we will need a verb, maybe we won't.
> > Adding one now just so we can add this simple fix for Asias, however, is certainly not acceptable.
> 
> I think "acceptable" is too strong of a word for this context. If we wanted to include repairs in the context of this patch adding a new verb (or some other sync machinery) would definitely be acceptable.
> 
> However we can definitely get along without a repair (and that verb) for now because repairs are much less of a problem since they usually don't generate the same amount of sstables like regular streaming (e.g. bootstrapping) .

I agree we can handle streaming separately first. Even if we are going to have repair based streaming very soon. I still wish we could have a simple fix for the older stable releases.

avikivity:
For scylla-manager, we'll need some extra information about the repair, to tell the node that this is only a partial repair and to keep the sstables in a staging area, then a verb to tell the node that repair is complete and we can destage the repaired sstables (through the off-strategy compaction process when we have it).

glommer:
@asias to move this forward, let's remove the repair part of it so we can deal with it later.
Let's discuss this week the best way to make that happen.

Respin the patch please including only streaming if possible. If not possible, as much as this hurts, we should drop it and wait for the proper implementation

asias:
> For scylla-manager, we'll need some extra information about the repair, to tell the node that this is only a partial repair and to keep the sstables in a staging area, then a verb to tell the node that repair is complete and we can destage the repaired sstables (through the off-strategy compaction process when we have it).

This means we need an verb any way even if off strategy compaction is there. The question is do we want to introduce the verb now for this patch which Glauber is against.

asias:
> @asias to move this forward, let's remove the repair part of it so we can deal with it later.
> Let's discuss this week the best way to make that happen.
> 
> Respin the patch please including only streaming if possible. If not possible, as much as this hurts, we should drop it and wait for the proper implementation

I will handle streaming for now. I think I can reuse some of the streaming verbs to mark the end of the streaming and trigger the compaction, adding an optional option to the existing verb, so no verb will be introduced.

asias:
@vladzcloudius @glommer I worked out a version that delay the compaction for regular streaming until the end of the entire bootstrap/decommission procedure. No new verb is added. It can even work with mixed cluster with/without the patch. Repair is not changed. I am talking to @amos to reuse the SCT to run some performance tests.

dht/range_streamer.cc
```diff
@@ -310,8 +311,12 @@ future<> range_streamer::do_stream_async() {
                             description, source, keyspace, nr_ranges_streamed, nr_ranges_total, ranges_to_stream.size());
                     nr_ranges_streamed += ranges_to_stream.size();
                     if (_nr_rx_added) {
+                        // This node is the stream receiver, trigger compaction on local node only
+                        _trigger_compaction[keyspace].insert(_address);
                         sp.request_ranges(source, keyspace, ranges_to_stream);
                     } else if (_nr_tx_added) {
+                        // This node is the stream sender, trigger compaction on remote node.
+                        _trigger_compaction[keyspace].insert(source);
```
vladzcloudius: Why do you need to trigger compactions on the receiver from the sender? 
Wouldn't the branch above it be enough?

asias: Say n1, n2 in the cluster, n3 is bootstrapping, it receives from n1 and n2. n3 needs to trigger the compaction after n1 and n2 finishes streaming.  The first branch.

Say n1, n2, n3 in the cluster. n3 is decommissioning. When n3 finishes the ops, it needs to tell n1 and n2 to trigger the compaction. The second branch. 

Note range_streamer can either send or receive. It does not do both.

asias:
With v2: 6bae010, the bootstrap node will not trigger any compaction during the operation by streaming code. However, if there is a write (no writes from normal cql write, but writes from other coordinator node sending to the joining node as the temporary pending list write) which triggers the compaction, the compaction will start to slow down bootstrap streaming.

dht/range_streamer.cc
vladzcloudius: I see. 
But what if the decommissioning node dies in the process?
I think you need to connect the compaction triggering to the same logic that you have today for deciding when streaming is over even if STREAMING_DONE is not received.

slivne:
@glommer  / @avikivity  can you please bless this direction and code (since we are going to invest in testing thos out prior to merge).

If you disagree with something it would be better to get it now and not have to redo testing afterwords.

glommer:
I don't like the fact that you are adding a new option to an existing verb.
Look, trigger compaction will not actually fire a compaction. It will just tell the compaction manager to see if a compaction is doable. It's not like a major or anything like that.

I think it is safe to just call it on `STREAM_MUTATION_DONE`, no changes to the verbs.

tgrabiec:
I don't think it makes sense to delay maybe-trigger-compaction if you don't delay adding the sstable. maybe-trigger-compaction is there to ensure that LSM is well-balanced, so that for example reads are not crazy expensive, or we don't OOM on sstable metadata. Whenever we add sstable, we should consider those conditions. Whether we add it due to memtable flush or streaming shouldn't matter.

asias:
> I don't like the fact that you are adding a new option to an existing verb.
> Look, trigger compaction will not actually fire a compaction. It will just tell the compaction manager to see if a compaction is doable. It's not like a major or anything like that.

I think we can have a better name for the option to avoid confusing. Adding rpc::optional is the designed way to extend the RPC without breaking compatibility.  Other than the name, anything else you don't like?

> 
> I think it is safe to just call it on `STREAM_MUTATION_DONE`, no changes to the verbs.

It is safe to always call the trigger. But 1) STREAM_MUTATION_DONE is called every time a stream plan is finished. We have multiple stream plans for a single bootstrap. If we trigger the compaction in STREAM_MUTATION_DONE, the rest of the stream plans will have to fight with compaction again. 
2) If we change to trigger from end of the ms().register_stream_mutation_fragments() to STREAM_MUTATION_DONE, nothing changes compared to current code, because STREAM_MUTATION_DONE runs immediately after end of register_stream_mutation_fragments.

asias:
> I don't think it makes sense to delay maybe-trigger-compaction if you don't delay adding the sstable. maybe-trigger-compaction is there to ensure that LSM is well-balanced, so that for example reads are not crazy expensive, or we don't OOM on sstable metadata. Whenever we add sstable, we should consider those conditions. Whether we add it due to memtable flush or streaming shouldn't matter.

The question is if we add an sstable without calling table::trigger_compaction(), if there is no reads request and no nodetool compact, will compaction manager consider the newly added sstable for compaction. If I understand Glabuer correctly, it will not. 

Higher read latency is what we would expect and probably the price the user want to pay for faster bootstrap. Also, during bootstrap there is no read at all. We will compact at the end of the bootstrap. 

The OOM issue without immediate compaction check, can't we do that unconditionally (do it even if user do not call  table::trigger_compaction())? If it will cause OOM, we have no choice we have to compaction.

tgrabiec:
On Wed, Mar 4, 2020 at 2:10 AM Asias He <notifications@github.com> wrote:

> I don't think it makes sense to delay maybe-trigger-compaction if you
> don't delay adding the sstable. maybe-trigger-compaction is there to ensure
> that LSM is well-balanced, so that for example reads are not crazy
> expensive, or we don't OOM on sstable metadata. Whenever we add sstable, we
> should consider those conditions. Whether we add it due to memtable flush
> or streaming shouldn't matter.
>
> The question is if we add an sstable without calling
> table::trigger_compaction(), if there is no reads request and no nodetool
> compact, will compaction manager consider the newly added sstable for
> compaction. If I understand Glabuer correctly, it will not.
>
> Higher read latency is what we would expect and probably the price the
> user want to pay for faster bootstrap. Also, during bootstrap there is no
> read at all. We will compact at the end of the bootstrap.
>
> The OOM issue without immediate compaction check, can't we do that
> unconditionally (do it even if user do not call
> table::trigger_compaction())? If it will cause OOM, we have no choice we
> have to compaction.
>

Doing it unconditionally still means that we have to go through the
compaction manager. trigger_compaction() is a signal to the
compaction_manager that the situation changed and it needs to re-evaluate
it against its goal. Still, it's the job of compaction manager to evaluate
the situation and decide what to compact. So the point is
table::trigger_compaction() needs to stay and if we want to change some of
the goals during rebuild (to take "no reads" into account), we should
communicate that to the compaction manager via separate channel.

glommer:
Given where this is going I'd advise to drop this for now.
We want to do this properly, and the whole point of this patch was to have a quick win. If there is fear of potential regressions on reads like Tomek seems to indicate, we should not move forward.

If we want to do something for the 4.0 timeframe, may I suggest we just bump shares for the streaming class in the bootstrapping nodes during the bootstrap process? This will make sure that even if we do have compactions, they won't dominate.

asias:
> Doing it unconditionally still means that we have to go through the compaction manager. trigger_compaction() is a signal to the compaction_manager that the situation changed and it needs to re-evaluate it against its goal. Still, it's the job of compaction manager to evaluate the situation and decide what to compact. So the point is table::trigger_compaction() needs to stay and if we want to change some of the goals during rebuild (to take "no reads" into account), we should communicate that to the compaction manager via separate channel.

Will something like this work:

table::trigger_compaction(bool hint_delay_compaction_if_possible)

We always call trigger_compaction when adding sstables, but if hint_delay_compaction_if_possible is set, the compaction manager compacts only if it must do.

glommer:
The compaction manager should already be compacting "only if it must do".
This is all heuristics, I don't think this will work

asias:
> The compaction manager should already be compacting "only if it must do".
> This is all heuristics, I don't think this will work

By "must do", I mean if not, the system will be out of memory. If there is possible compaction to make read faster, delay it.

glommer:
The compaction manager simply doesn't have the needed information to make this decision. Another factor is that a compaction will only help many minutes in the future - by the time the state of the system will have changed.

asias:
> The question is if we add an sstable without calling table::trigger_compaction(), if there is no reads request and no nodetool compact, will compaction manager consider the newly added sstable for compaction. If I understand Glabuer correctly, it will not.

@tgrabiec @raphaelsc  I still hope to get an answer for this.

asias:
> The compaction manager simply doesn't have the needed information to make this decision. Another factor is that a compaction will only help many minutes in the future - by the time the state of the system will have changed.

According to https://github.com/scylladb/scylla/pull/5866#issuecomment-594456947, we can have a way to tell the compaction manager the goal of the evaluation after the trigger of the compaction. What information is absent for the compaction manager to decide if adding the meta-data of the sstable to be added will cause OOM?

Adding meta-data of sstable cause OOM is a very rare event, no? If this is the case, probably the bootstrap will fail very soon due to lack of memory anyway. Nothing to lose.

glommer:
The compaction manager doesn't run for a table unless trigger_compaction is called.
Once the sstables are added they are eligible for compaction in case something else happens to trigger it. But without `trigger_compaction()` they will not happen

glommer:
The issue is not due to metadata, but memory used for reads. That memory is proportional to nr_sstables.

asias:
> The compaction manager doesn't run for a table unless trigger_compaction is called.
> Once the sstables are added they are eligible for compaction in case something else happens to trigger it. But without `trigger_compaction()` they will not happen

What are the "something else", a memtable flush for that table due to new writes (does it matter if the ranges covered the newly flushed sstable and ranges covered by the streaming sstable added are disjoint or not). How about reads?

glommer:
In theory "something else" could be anything, including user action.
In practice it will be a memtable flush or some other existing compaction finishing. It doesn't matter which ranges are involved.

Reads currently do not trigger compactions.

asias:
> The issue is not due to metadata, but memory used for reads. That memory is proportional to nr_sstables.

Streaming adds very few and big sstables. 10 sstables at most per table, because each covers 10% of the whole ranges a node owns.

glommer:
One way in which we can cheat a bit is maybe to increase min_threshold during streaming.
If we set it to 10, we're effectively very close to what offstrategy will do

asias:
> The issue is not due to metadata, but memory used for reads. That memory is proportional to nr_sstables.

Also, there is no reads at all during bootstrap, only pending writes. If we fear the number of sstables after streaming is too high to trigger OOM, we can compact and  then allow user normal cql read. But this is not needed, because the total sstable numbers are 10 per table.

asias:
> One way in which we can cheat a bit is maybe to increase min_threshold during streaming.
> If we set it to 10, we're effectively very close to what offstrategy will do

min_threshold = number of sstables to consider compact?

If the number is less than that, no compaction at all, and it does not matter if it is STCS or LCS?

asias:
> One way in which we can cheat a bit is maybe to increase min_threshold during streaming.
> If we set it to 10, we're effectively very close to what offstrategy will do

One thing to note though, it is 10 per table per peer nodes. Bootstrap and replace operation will stream from multiple nodes.

glommer:
This doesn't matter. The goal for now is not to prevent compactions altogether. In fact we don't want that.

We just want to make it less aggressive.

tgrabiec:
On Thu, Mar 5, 2020 at 1:12 AM Asias He <notifications@github.com> wrote:

> Doing it unconditionally still means that we have to go through the
> compaction manager. trigger_compaction() is a signal to the
> compaction_manager that the situation changed and it needs to re-evaluate
> it against its goal. Still, it's the job of compaction manager to evaluate
> the situation and decide what to compact. So the point is
> table::trigger_compaction() needs to stay and if we want to change some of
> the goals during rebuild (to take "no reads" into account), we should
> communicate that to the compaction manager via separate channel.
>
> Will something like this work:
>
> table::trigger_compaction(bool hint_delay_compaction_if_possible)
>
> We always call trigger_compaction when adding sstables, but if
> hint_delay_compaction_if_possible is set, the compaction manager compacts
> only if it must do.
>

I think this hint shouldn't be passed by the one who adds the sstable.
Behavior doesn't depend on who is adding the sstable. We want the same
less-aggressive behavior if writes were going through (streaming)
memtables. The behavior should be tuned by the bootstrapper by setting some
global compaction hints.

tgrabiec:
On Thu, Mar 5, 2020 at 2:34 AM Asias He <notifications@github.com> wrote:

> One way in which we can cheat a bit is maybe to increase min_threshold
> during streaming.
> If we set it to 10, we're effectively very close to what offstrategy will
> do
>
> One thing to note though, it is 10 per table per peer nodes. Bootstrap and
> replace operation will stream from multiple nodes.
>

For users with 10000 tables, 10 per table can still OOM.

>

asias:
> On Thu, Mar 5, 2020 at 2:34 AM Asias He ***@***.***> wrote: One way in which we can cheat a bit is maybe to increase min_threshold during streaming. If we set it to 10, we're effectively very close to what offstrategy will do One thing to note though, it is 10 per table per peer nodes. Bootstrap and replace operation will stream from multiple nodes.
> For users with 10000 tables, 10 per table can still OOM.

This case can OOM with/without streaming. Compaction can introduce more sstables, e.g., each streaming generated sstable is compacted into 5 sstables in LCS.

asias:
> I think this hint shouldn't be passed by the one who adds the sstable. Behavior doesn't depend on who is adding the sstable. We want the same less-aggressive behavior if writes were going through (streaming) memtables. The behavior should be tuned by the bootstrapper by setting some global compaction hints.

How does the compaction manager know if the sstables are from normal writes or streaming? We want less aggressive compaction for streaming, not for normal writes compaction.

tgrabiec:
On Mon, Mar 9, 2020 at 1:35 AM Asias He <notifications@github.com> wrote:

> I think this hint shouldn't be passed by the one who adds the sstable.
> Behavior doesn't depend on who is adding the sstable. We want the same
> less-aggressive behavior if writes were going through (streaming)
> memtables. The behavior should be tuned by the bootstrapper by setting some
> global compaction hints.
>
> How does the compaction manager know if the sstables are from normal
> writes or streaming? We want less aggressive compaction for streaming, not
> for normal writes compaction.
>

Why do we want different compaction behavior if the sstable came from
normal writes?

>

asias:
> Why do we want different compaction behavior if the sstable came from normal writes?

Consider a decommission operation, the node receiving the streaming is an existing node. Should we make all the compaction from normal writes slow and slow down the reads? We only want to slow down the compactions of the sstables generated by streaming. The work to compact the sstables generated by streaming is part of the streaming work, it should be in the same level of the streaming group.

tgrabiec:
On Mon, Mar 9, 2020 at 12:49 PM Asias He <notifications@github.com> wrote:

> Why do we want different compaction behavior if the sstable came from
> normal writes?
>
> Consider a decommission operation, the node receiving the streaming is an
> existing node. Should we make all the compaction from normal writes slow
> and slow down the reads? We only want to slow down the compactions of the
> sstables generated by streaming. The work to compact the sstables generated
> by streaming is part of the streaming work, it should be in the same level
> of the streaming group.
>

If you didn't attach the sstables, then yes, it has lower priority. As soon
as you attach the sstables, the priority is inherited from the reads. The
reason we compact is so that reads are not slow, and it doesn't matter who
created the sstable. If we slow down compaction of attached sstables there
would be priority inversion (reads would suffer).

tgrabiec:
On Mon, Mar 9, 2020 at 1:27 AM Asias He <notifications@github.com> wrote:

> On Thu, Mar 5, 2020 at 2:34 AM Asias He *@*.***> wrote: One way in which
> we can cheat a bit is maybe to increase min_threshold during streaming. If
> we set it to 10, we're effectively very close to what offstrategy will do
> One thing to note though, it is 10 per table per peer nodes. Bootstrap and
> replace operation will stream from multiple nodes.
> For users with 10000 tables, 10 per table can still OOM.
>
> This case can OOM with/without streaming. Compaction can introduce more
> sstables, e.g., each streaming generated sstable is compacted into 5
> sstables in LCS.
>
Yes. For such cases we need another solution, like reducing the number of
tables owned by a node.

Before reaching that boiling point, compaction is supposed to reduce the
number of sstables. We already have a controller which increases the shares
of compaction based on the backlog. I think the reasons behind this
controller's existence should hold also during streaming.

asias:
> If you didn't attach the sstables, then yes, it has lower priority. As soon as you attach the sstables, the priority is inherited from the reads. The reason we compact is so that reads are not slow, and it doesn't matter who created the sstable. If we slow down compaction of attached sstables there would be priority inversion (reads would suffer).

The whole point of compaction delay during bootstrap/decommission is we use slower read to trade for faster node operations. The suffer of reads is unavoidable. However, the suffer is after the operation, not during the operation. 

The read will not hit the new ranges the node is going to own during the operation. So even if we attaches the sstables, there should be no read at the ranges we attached until the node operations is done. Reading will not be slowed down due to lack compaction during the operation. What is the point of to compact the streaming generated at all during the operation?

@glommer I guess this is an important property for the off-strategy compaction.

asias:
Retire this PR. Favor solution https://github.com/scylladb/scylla/pull/6007.

PR #5868, opened 2020-02-21
gossip: Do not talk to unreachable members

### LINKED ISSUE #5867: gossip: Do not talk to unreachable members
Currently, in each gossip round, gossip will try to talk to some
unreachable members with some probability to check if the node is back.

It is an optimization hoping to notice a once dead node up early.
However, when a node is down, other nodes will be full of logs like:
```
   INFO  2020-02-21 10:52:17,234 [shard 0] rpc - client 127.0.0.30:7000:
   fail to connect: Connection refused
   INFO  2020-02-21 10:52:19,235 [shard 0] rpc - client 127.0.0.30:7000:
   fail to connect: Connection refused
   INFO  2020-02-21 10:52:20,235 [shard 0] rpc - client 127.0.0.30:7000:
   fail to connect: Connection refused
```
It persists until a node is up again.

The optimization is not very important and it is safe to remove it because

(Consider n1, n2, n3 in the cluster. The node n3 was dead and became up.)

1) When a dead node is up, the node itself will talk to other nodes. So
   even if existing nodes do not try to talk to the dead node, they will
   notice the dead node is up anyway.

2) Consider an extremely unlucky node n2, the node that became up again,
   n3, never talked to n2. It is still fine because when other nodes
   know n3 is up talk to n2. n2 will know n3 is up.

3) Gossip in scylla talks to more than one live nodes in each round. So
   the propagation of n3 is up should be pretty faster without talking
   to unreachable member.

This patch is part of the get rid of the seed concept work.

asias:
Currently, in each gossip round, gossip will try to talk to some
unreachable members with some probability to check if the node is back.

It is an optimization hoping to notice a once dead node up early.
However, when a node is down, other nodes will be full of logs like:

   INFO  2020-02-21 10:52:17,234 [shard 0] rpc - client 127.0.0.30:7000:
   fail to connect: Connection refused
   INFO  2020-02-21 10:52:19,235 [shard 0] rpc - client 127.0.0.30:7000:
   fail to connect: Connection refused
   INFO  2020-02-21 10:52:20,235 [shard 0] rpc - client 127.0.0.30:7000:
   fail to connect: Connection refused

It persists until a node is up again.

The optimization is not very important and it is safe to remove it because

(Consider n1, n2, n3 in the cluster. The node n3 was dead and became up.)

1) When a dead node is up, the node itself will talk to other nodes. So
   even if existing nodes do not try to talk to the dead node, they will
   notice the dead node is up anyway.

2) Consider an extremely unlucky node n2, the node that became up again,
   n3, never talked to n2. It is still fine because when other nodes
   know n3 is up talk to n2. n2 will know n3 is up.

3) Gossip in scylla talks to more than one live nodes in each round. So
   the propagation of n3 is up should be pretty faster without talking
   to unreachable member.

This patch is part of the get rid of the seed concept work.

Fixes #5867

### DIFF
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -768,12 +768,6 @@ void gossiper::run() {
                     });
                 }
 
-                /* Gossip to some unreachable member with some probability to check if he is back up */
-                // Do it in the background.
-                (void)do_gossip_to_unreachable_member(message).handle_exception([] (auto ep) {
-                    logger.trace("Faill to do_gossip_to_unreachable_member: {}", ep);
-                });
-
                 /* Gossip to a seed if we did not do so above, or we have seen less nodes
                    than there are seeds.  This prevents partitions where each group of nodes
                    is only gossiping to a subset of the seeds.
@@ -1183,30 +1177,6 @@ future<int> gossiper::get_current_heart_beat_version(inet_address endpoint) {
 
 future<> gossiper::do_gossip_to_live_member(gossip_digest_syn message, gms::inet_address ep) {
     return send_gossip(message, {ep});
-}
-
-future<> gossiper::do_gossip_to_unreachable_member(gossip_digest_syn message) {
-    double live_endpoint_count = _live_endpoints.size();
-    double unreachable_endpoint_count = _unreachable_endpoints.size();
-    if (unreachable_endpoint_count > 0) {
-        /* based on some probability */
-        double prob = unreachable_endpoint_count / (live_endpoint_count + 1);
-        std::uniform_real_distribution<double> dist(0, 1);
-        double rand_dbl = dist(_random_engine);
-        if (rand_dbl < prob) {
-            std::set<inet_address> addrs;
-            for (auto&& x : _unreachable_endpoints) {
-                // Ignore the node which is decommissioned
-                if (get_gossip_status(x.first) != sstring(versioned_value::STATUS_LEFT)) {
-                    addrs.insert(x.first);
-                }
-            }
-            logger.trace("do_gossip_to_unreachable_member: live_endpoint nr={} unreachable_endpoints nr={}",
-                live_endpoint_count, unreachable_endpoint_count);
-            return send_gossip(message, addrs);
-        }
-    }
-    return make_ready_future<>();
 }
 
 future<> gossiper::do_gossip_to_seed(gossip_digest_syn prod) {
--- a/gms/gossiper.hh
+++ b/gms/gossiper.hh
@@ -392,9 +392,6 @@ private:
     /* Sends a Gossip message to a live member */
     future<> do_gossip_to_live_member(gossip_digest_syn message, inet_address ep);
 
-    /* Sends a Gossip message to an unreachable member */
-    future<> do_gossip_to_unreachable_member(gossip_digest_syn message);
-
     /* Gossip to a seed for facilitating partition healing */
     future<> do_gossip_to_seed(gossip_digest_syn prod);
 


avikivity:
I agree with the patch.

For the record, I'll note that an a way to reduce log spam is to demote the connection error messages to debug level after they are logged once.

However, perhaps it should be coupled with a patch to make a starting node try to communicate with all its peers, so that information propagation speed is not reduced.

What about this scenario:

 1. nodes n1..n6 are happily living together in a cluster
 2. n1 goes down
 3. nodes n7..n11 are bootstrapped
 4. nodes n2..n5 are decommissioned
 5. n6 is started

After the patch, n7..n11 won't talk to n1, and n1 doesn't know about n7..n11, so it can't talk to them,

asias:
> I agree with the patch.
> 
> For the record, I'll note that an a way to reduce log spam is to demote the connection error messages to debug level after they are logged once.
> 
> However, perhaps it should be coupled with a patch to make a starting node try to communicate with all its peers, so that information propagation speed is not reduced.

I already plan to do this in the get rid of seed work. 

> 
> What about this scenario:
> 
>     1. nodes n1..n6 are happily living together in a cluster
> 
>     2. n1 goes down
> 
>     3. nodes n7..n11 are bootstrapped
> 
>     4. nodes n2..n6 are decommissioned
> 
>     5. n1 is started
> 
> 
> After the patch, n7..n11 won't talk to n1, and n1 doesn't know about n7..n11, so it can't talk to them,

This is a very good example. Currently, n1 won't start in step 5 because it has to talk to an existing seed node during start-up for feature bit check. To set seed node for n1 to one of node in existing cluster will make n1 know n7 and n11. However, if n1 only sets itself as seed node, it will be able to restart but other nodes won't talk to it. 

Let's consider we already have removed the seed concept. In this case, when n1 restarts, it still does not know n7 and n11 even if it talks to all peers it knows before it goes down.  To be safe, we can refuse to allow n1 to boot if zero peers can be talked with, but this is too strong because it will also refuse too boot a node in a 2 nodes cluster when the other node is down.

tzach:
> Let's consider we already have removed the seed concept. In this case, when n1 restarts, it still does not know n7 and n11 even if it talks to all peers it knows before it goes down. 

Can you please explain what removing the seed concept means?
Seeds have two functions:
1. Play a unique role in gossip rounds
2. A list of IPs a node connect to in boot

I do not understand how you can remove 2, since a node still need to find his friends on boot.
If you provide all nodes, or a subset of them, how is this different from "seeds"?
Even now, the seeds per node can be different, they are not more then an initial list of friends.

tgrabiec:
> What about this scenario:
> 
> 1. nodes n1..n6 are happily living together in a cluster
> 2. n1 goes down
> 3. nodes n7..n11 are bootstrapped
> 4. nodes n2..n6 are decommissioned
> 5. n1 is started
> 
> After the patch, n7..n11 won't talk to n1, and n1 doesn't know about n7..n11, so it can't talk to them,

I think step 3 should fail already if n1 is down because if n1 doesn't get the new topology then it will route writes incorrectly (at step 5, or earlier if it's just behind a network partition) and we may lose writes as a result.

asias:
> What happens on a plain network partition:
> 
> n1 n2 n3
> 
>     1. There is a network partition between {n1} and {n2, n3}
> 
>     2. n1 convicts n2 and n3 as down
> 
>     3. n2 and n3 convict n1 as down
> 
>     4. network partition heals
> 
>     5. will n2 and n3 notice that n1 is up again if they don't gossip with it and neither is n1 gossiping with them?

Unfortunately, if the nodes both think each other are DOWN, it will be "deadlock" until one of the nodes reach out. Talking to unreachable is such reach out. 

I came up with a idea that if a node can not talk to any of the live node, it then talks to the unreachable nodes. It solves the problem in your example. However, it will not work if {n1,n2} and {n3,n4} are partitioned because n1 can talk to n2.

It is pretty sad we have to keep it for the corner cases. Making unnecessary attempt to the nodes that shtudown themselves explicitly. E.g., n1, n2, n3. n3 shutdown itself and announce it. n1 and n2 knows n3 is in SHUTDOWN state, yet n1 and n2 have to keep reaching out to n3.

asias:
> > What about this scenario:
> > 
> > 1. nodes n1..n6 are happily living together in a cluster
> > 2. n1 goes down
> > 3. nodes n7..n11 are bootstrapped
> > 4. nodes n2..n6 are decommissioned
> > 5. n1 is started
> > 
> > After the patch, n7..n11 won't talk to n1, and n1 doesn't know about n7..n11, so it can't talk to them,
> 
> I think step 3 should fail already if n1 is down because if n1 doesn't get the new topology then it will route writes incorrectly (at step 5, or earlier if it's just behind a network partition) and we may lose writes as a result.

At step 3, n1 is down, it can not route any writes. Once it is up, it should learn the new nodes have joined and learn the new topology. However, if n1 is in a state that is partitioned from the rest of the cluster, n1 can still take writes from clients.

avikivity:
@asias I think the flaw @tgrabiec found is fatal. The nodes are not down, but the network is, so we need to keep reconnecting in order to detect when the network is back up.

asias:
> @asias I think the flaw @tgrabiec found is fatal. The nodes are not down, but the network is, so we need to keep reconnecting in order to detect when the network is back up.

Yes, it is for sure that we can not simply drop the unreachable members like it is in this patch. I hope we can find out smarter alternatives. Trying to talk to a  shutdown node forever is very sad.

kostja:
One possibility is to assume each node always strives to reach a quorum. As long as a node has quorum for all its token ranges, it is not obliged to talk to a node it considers down. Otherwise it keeps reaching out.

asias:
> One possibility is to assume each node always strives to reach a quorum. As long as a node has quorum for all its token ranges, it is not obliged to talk to a node it considers down. Otherwise it keeps reaching out.

Thanks, @kostja . 

Consider this:  
n1, n2, n3 in the cluster
Partition happens between {n1, n2}  and   {n3}
n4, n5 are added. Partition persists between  {n1, n2}  and {n3, n4, n5}
Both partitions think they reach quorum, so they do not talk to nodes in the other partition.

kostja:
1) I thought such operations are not allowed in a degraded state. Shouldn't the DBA first resolve the unavailability issue (mark the node explicitly as down) and only then add new nodes?
2) This can lead to other bad things too. Imagine Paxos/LWT writes in such a partitioned cluster, they will succeed even though it's a split brain situation!
3) If a new node is added to a minority partition in a degraded state, we can detect that too, and keep pinging.

asias:
> 1. I thought such operations are not allowed in a degraded state. Shouldn't the DBA first resolve the unavailability issue (mark the node explicitly as down) and only then add new nodes?

For example, with RF = 10, we can add node n4, n5 with no problem. n1, n2 do not lose any token ranges,  it is not a must to stream from n1 or n2 when adding n4 and n5.

In practice, user can either fix the partition or remove the nodes in one of the partition from the cluster, then add new node. 

> 2. This can lead to other bad things too. Imagine Paxos/LWT writes in such a partitioned cluster, they will succeed even though it's a split brain situation!

Yes.

> 3. If a new node is added to a minority partition in a degraded state, we can detect that too, and keep pinging.

How does it know it is a minority partition? What if there are multiple partitions?

kostja:
It will require some more accurate definition, for sure, but basically one is in a minority partition if some token range have no quorum.  I agree that to work properly we'll need to keep a history of add/remove/unavailability events, since it stops being a function of the current state, but has to be true for all states since the last unavailability event.

tgrabiec:
On Fri, Feb 28, 2020 at 8:49 AM Asias He <notifications@github.com> wrote:

>
>    1. I thought such operations are not allowed in a degraded state.
>    Shouldn't the DBA first resolve the unavailability issue (mark the node
>    explicitly as down) and only then add new nodes?
>
> For example, with RF = 10, we can add node n4, n5 with no problem. n1, n2
> do not lose any token ranges, it is not a must to stream from n1 or n2 when
> adding n4 and n5.
>
> In practice, user can either fix the partition or remove the nodes in one
> of the partition from the cluster, then add new node.
>

Even if we don't need to stream from unreachable nodes, we should fail
bootstrap if some nodes didn't get the new topology. Not doing so can lead
to inconsistency:

https://github.com/scylladb/scylla/issues/5252



>

asias:
> Even if we don't need to stream from unreachable nodes, we should fail bootstrap if some nodes didn't get the new topology. Not doing so can lead to inconsistency:

This depends on what is the cause of the unreachable issue. If the node is partitioned and the node is still taking writes without get the new topology (tokens), the node will use wrong token ring info to route the writes. If the node is shutdown, we do not have this problem because it is not taking any reads/writes. When the node restart, it should talk to existing nodes to get the latest topology. 

Should we reject adding new node if other nodes are shutdown? I think we should not. It is too strong limitation.

kostja:
Indeed, it's a very strong limitation and we should not reject adding a new node then. But what does it have to do with this ticket?

tgrabiec:
On Wed, Mar 4, 2020 at 1:42 AM Asias He <notifications@github.com> wrote:

> Even if we don't need to stream from unreachable nodes, we should fail
> bootstrap if some nodes didn't get the new topology. Not doing so can lead
> to inconsistency:
>
> This depends on what is the cause of the unreachable issue. If the node is
> partitioned and the node is still taking writes without get the new
> topology (tokens), the node will use wrong token ring info to route the
> writes. If the node is shutdown, we do not have this problem because it is
> not taking any reads/writes. When the node restart, it should talk to
> existing nodes to get the latest topology.
>
> Should we reject adding new node if other nodes are shutdown? I think we
> should not. It is too strong limitation.
>
Why do you think it's a too strong limitation? If a node which is shut down
can't be brought up, the proper path to cluster expansion is to replace it.

We could tolerate one node down with some effort, but not more than that
due to the fact that we need to preserve QUORUM write consistency.

Even if the node which is down contacts the cluster first, there are corner
cases like the one mentioned by Avi, where the node which was down no
longer knows about any nodes which had the accurate view of the topology,
so it has no way to sync itself up on boot.

asias:
> Why do you think it's a too strong limitation? If a node which is shut down can't be brought up, the proper path to cluster expansion is to replace it. We could tolerate one node down with some effort, but not more than that due to the fact that we need to preserve QUORUM write consistency. Even if the node which is down contacts the cluster first, there are corner cases like the one mentioned by Avi, where the node which was down no longer knows about any nodes which had the accurate view of the topology, so it has no way to sync itself up on boot.

Because a single node down prevents adding new nodes. User can replace it then add new nodes. However, a node down at the moment does not mean it can not be up again in the future. For instance, a temporary power failure in a rack, the node can be booted again when the issue is resolved, maybe in 30 minutes.  It is just a waste of time and effort to replace it.  During that down time, user is perfectly fine to add new nodes. After we have adding more nodes in parallel support, requiring replacing node before adding nodes serializes the adding node procedure. Say the replace takes 1h, in that time we can add more nodes to the cluster in parallel. This matters if adding nodes asap to the cluster is the first priority.

asias:
> Indeed, it's a very strong limitation and we should not reject adding a new node then. But what does it have to do with this ticket?

It is related because the example cases we are discussing have adding nodes in the procedure.

slivne:
@asias whats the status of this ? I think there is agreement about https://github.com/scylladb/scylla/pull/5868#issuecomment-591589284 - being a blocker for this going in, unless I am missing something.

Please close if its agreed this should not be meged

slivne:
@asias whats the status of this ? I think there is agreement about https://github.com/scylladb/scylla/pull/5868#issuecomment-591589284 - being a blocker for this going in, unless I am missing something.

Please close if its agreed this should not be meged

asias:
We will not merge this. I am closing it for now. We need a better solution.

PR #5871, opened 2020-02-21
dist/debian: remove unused dependencies

syuu1228:
Since we moved relocatable package, almost all dependencies are not needed now.


avikivity:
Queued as 3d1154272f78e0ae01707fcd027922e4fd2c5f95

PR #5872, opened 2020-02-21
dist/redhat: align dependencies with debian

syuu1228:
On Debian, we don't add xfsprogs/mdadm on package dependency, install on
scylla_raid_setup script instead.
Since xfsprogs/mdadm only needed for constructing RAID, we can move
dependencies to scylla_raid_setup too.

### DIFF
--- a/dist/common/scripts/scylla_raid_setup
+++ b/dist/common/scripts/scylla_raid_setup
@@ -82,6 +82,8 @@ if __name__ == '__main__':
 
     if is_debian_variant():
         run('env DEBIAN_FRONTEND=noninteractive apt-get -y install mdadm xfsprogs')
+    elif is_redhat_variant():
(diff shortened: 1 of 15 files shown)


dist/common/scripts/scylla_raid_setup
```diff
@@ -82,6 +82,8 @@ if __name__ == '__main__':
 
     if is_debian_variant():
         run('env DEBIAN_FRONTEND=noninteractive apt-get -y install mdadm xfsprogs')
+    elif is_redhat_variant():
+        run('yum install -y mdadm xfsprogs')
```
avikivity: Perhaps we can make it a weak dependency.

avikivity:
Queued as 98c182ec67513921d23a80fcd44f88c2b80baf9d

PR #5873, opened 2020-02-21
install.sh: run post-install script just like .rpm/.deb package

### LINKED ISSUE #5830: offline install with relocated packages
*Installation details*
Scylla version (or git commit hash): master
Cluster size: 1
OS (RHEL/CentOS/Ubuntu/AWS AMI): CentOS

I tried to offline install scylla on a GCE (centos7) instance by relocated packages, it's completed, but not a good practice.

It's not good to install scylla by relocated packages in production environment with firewall limit. There are some external setup or attribute change in RPM or Deb installation, it's not enough to run the install.sh in all relocated packages. The independency is also a big problem, we can't solve them one by one manually in production env.

The current recommended offline installer (https://github.com/scylladb/scylla/wiki/Offline-Installer) depends on RPM or deb, and all dependency packages will be prepared.

But the install.sh in relocated packages only copy the files to right directories without setting the attributes (ownership, permission). And the dependencies have to be solved manually.

My Procedure (Offline installation with relocated packages):
- https://docs.google.com/document/d/11eg74ojQs_4Zud7aIND8Grdhd0cE2DLDcNtFY4NMUQc/edit?usp=sharing

It seems original offline install script is better. If we want to switch offline install to use relocate packages, we need to improve the setup and solve the dependency issue.

@syuu1228 @slivne @roydahan

syuu1228:
To install scylla using install.sh easily, we need to run following things:
 - add scylla user/group
 - configure scylla.yaml
 - run scylla_post_install.sh

But we don't want to run them when we build .rpm/.deb package,
we also need to add --packaging option to skip them.

Fixes #5830

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -33,6 +33,7 @@ Options:
   --housekeeping           enable housekeeping service
   --nonroot                install Scylla without required root priviledge
   --sysconfdir /etc/sysconfig   specify sysconfig directory name
+  --packaging               use install.sh for packaging
   --help                   this helpful message
 EOF
     exit 1
@@ -43,6 +44,7 @@ housekeeping=false
 python3=/opt/scylladb/python3/bin/python3
 sysconfdir=/etc/sysconfig
 nonroot=false
+packaging=false
 
 while [ $# -gt 0 ]; do
     case "$1" in
@@ -69,6 +71,10 @@ while [ $# -gt 0 ]; do
         "--sysconfdir")
             sysconfdir="$2"
             shift 2
+            ;;
+        "--packaging")
+            packaging=true
+            shift 1
             ;;
         "--help")
             shift 1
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -321,4 +327,31 @@ if $nonroot; then
     touch $rprefix/SCYLLA-NONROOT-FILE
     systemctl --user daemon-reload
     echo "Scylla non-root install completed."
+elif ! $packaging; then
+    getent passwd scylla || nousr=1
+    getent group scylla || nogrp=1
```
avikivity: These variables are not initialized, so if they are present in the user's environment they will result in incorrect action.

penberg:
@syuu1228 Please rebase the pull request to drop the two extra commits.

install.sh
syuu1228: Added initialize code.

syuu1228:
@penberg rebased with latest next.

Issue #4208, opened 2019-02-09
remove, rebuild, replace, are not safe and can stream stale data, causing QUORUM queries to return wrong results


glommer:
(edited: as per discussion with @slivne , add itself seems safe)

Current operations like ~~add~~, remove and rebuild will stream data from one of the replicas only.

It is possible that if the cluster is not fully repaired that the replica that has stale data ends up transferring.
After a sequence of ~~adds~~ (removes) within the same time window without repair, you can end up with no data at all.
Things are even worse if we consider that the cluster will have ongoing writes. We will repair some ranges, but could very well transfer a range that was written to after it was repaired -- meaning that we now need a repair both before & after streaming.

If we really need to ask for repairs in between streaming operations then it would be easier to just pay the price of consulting more replicas upfront to make sure we have up2date data. An option can still be given to people who want faster repairs to do so, but it should in no way be the default. By default we should touch all live replicas to stream to guarantee those operations are always safe.

tzach:
@glommer till this is fix, I understand that running a repair as a prerequisite to the above operation will mitigate the issue. Please comment.

slivne:
I do not agree with all that is stated:

- Add node is a safe operation - it transfers ownership from one node to another - doesn't stream from a random node  - so inconsistency (if there is any will be kept).
- Rebuild operation does what it claims todo - streams information from one other replica to this replica. If there are inconsistencies they are kept not increased. 

- Remove node is not safe and should have repair added - but that should be done prior to removenode not after !!!!

glommer:
Let's get clarification about the first bullet then.

About the second and third, "it does what it is supposed to do" is pointless if what is supposed to do is unsafe. We should then change what it is supposed to do.

rebuild should stream the most up2date copy.
remove should stream the most up2date copy. 

The only way to get those two operations (for now) to be safe is to run repairs before you trigger them. If that's the case they should do the repair themselves (or just do the transfers with increased consistency-level).

While it doesn't we should definitely state it in the documentation that those are needed. But "they do what they are supposed to do" is a poor excuse for a badly designed operation.

slivne:
If the user does rebuild he will not get the node to a worse state then it was

Consider the current state: 3 nodes RF=3

Node A: correct data
Node B: old copy T1
Node C: old copy T2 (T2 < T1)

If the user reads at this state with QUORUM he can get the correct value if Node A is read from or can get old copy T1 (if Node B,C are used in the read).

If the user runs rebuild on Node C:
[...]

Issue #3003, opened 2017-11-21
removenode operation is less safe than it should be.


glommer:
Running removenode in an unrepaired cluster is potentially dangerous. Consider the following scenario:
all writes and reads are QUORUM. Nodes A and B have the most up2date copy of the data, but C doesn't.

Node A dies, now nodes B and C are alive. Upon calling removenode, a new set of nodes E, F and G will now be replicas for that key. B and C will stream for the three of them, but the streaming selection is based on locality, and not on data timestamps.

It is possible that F and G will receive the data from C, which is old.
QUORUM reads involving E will be still fine, but a QUORUM read that happens to return from F and G will return old data.

Hinted Handoff does not fix this: the hints could very well live in node A and are now gone.
Continuous repair does not fix this: all of the above can happen before repair has a chance to touch that token range.

If there is need for repair for safety, we should do it ourselves somehow.

glommer:
One idea from @avikivity is to change removenode invocation, so we can do:

```
   nodetool removenode --streaming=local_quorum | quorum | local_one
```
And allow the user to select the consistency level expected from that streaming.

@asias has other ideas, one of which he explains as:
```
do not use the stream interface, use modified repair internally in removenode operation, i.e. ask all replica to do a hash, stream all replica with different hash.
less data on wire but adds cpu overhead for hashing. 
```

PR #5885, opened 2020-02-24
Repair based node operation

asias:
Here is a simple introduction to the node operations scylla supports and
some of the issues.

- Replace operation

   It is used to replace a dead node. The token ring does not change. It
   pulls data from only one of the replicas which might not be the
   latest copy.

- Rebuild operation

   It is used to get all the data this node owns form other nodes. It
   pulls data from only one of the replicas which might not be the
   latest copy.

- Bootstrap operation

   It is used to add a new node into the cluster. The token ring
   changes. Do no suffer from the "not the latest replica” issue. New
   node pulls data from existing nodes that are losing the token range.

   Suffer from failed streaming. We split the ranges in 10 groups and we
   stream one group at a time. Restream the group if failed, causing
   unnecessary data transmission on wire.

   Bootstrap is not resumable. Failure after 99.99% of data is streamed.
   If we restart the node again, we need to stream all the data again
   even if the node already has 99.99% of the data.

- Decommission operation

   It is used to remove a live node form the cluster. Token ring
   changes.  Do not suffer “not the latest replica” issue. The leaving
   node pushes data to existing nodes.

   It suffers from resumable issue like bootstrap operation.

- Removenode operation

   It is used to remove a dead node out of the cluster. Existing nodes
   pulls data from other existing nodes for the new ranges it own. It
   pulls from one of the replicas which might not be the latest copy.

To solve all the issues above. We could use repair based node operation.
The idea behind repair based node operations is simple: use repair to
sync data between replicas instead of streaming.

The benefits:

   - Latest copy is guaranteed

   - Resumable in nature

   - No extra data is streamed on wire
     E.g., rebuild twice, will not stream the same data twice

   - Unified code path for all the node operations

   - Free repair operation during bootstrap, replace operation and so on.

Fixes: #3003
Fixes: #4208
Tests: update_cluster_layout_tests.py + replace_address_test.py + manual test


asias:
@avikivity I see nodetool cleanup patchset is in. The previous PR is closed and I can not reopen. Creating a new one.

avikivity:
Dequeued, it doesn't build after a merge. Please rebase (and check commits for bisectability, since this isn't a simple textual conflict).

asias:
@avikivity @slivne @roydahan I created https://github.com/scylladb/scylla-dtest/pull/1329. We need it to pass the update_cluster_layout_tests.py. It looks like update_cluster_layout_tests is not part of the scylla next to mater promoting gate. Anyway, we need the change to make new code works with dtest.

asias:
> Dequeued, it doesn't build after a merge. Please rebase (and check commits for bisectability, since this isn't a simple textual conflict).

The build failure is due to 897bbeabea0901673788db3dd35f12e2627a22f6.  

Changes in aaa1f3ce7bfe986712aacae80ef270a4f33f4794:
- Rebased to master
- Fixed missing of finish_bootstrapping build failure
- Verified each and every commit compiles in this series

asias:
@avikivity I ran the update_cluster_layout_tests.py (https://github.com/scylladb/scylla-dtest/pull/132911) today. All 41 tests passed. I think we are ready to go if update_cluster_layout_tests.py is not part of the gating.

avikivity:
Trying again :crossed_fingers:

Issue #5709, opened 2020-02-03
CDC: row deletes are handled incorrectly


kbr-:
```
cqlsh> create keyspace ks with replication = {'class':'SimpleStrategy', 'replication_factor': 1};
cqlsh> create table ks.t (pk int, ck int, v int, primary key (pk, ck)) with cdc = {'enabled':'true', 'preimage':'true'};
cqlsh> insert into ks.t (pk, ck, v) values (0,0,0);                                                                                                                                                                                                                                                                           
cqlsh> delete from ks.t where pk = 0 and ck = 0;                                                                                                                                                                                                                                                                              
cqlsh> select system.totimestamp(time), batch_seq_no, "_pk", "_ck", "_v", operation from ks.t_scylla_cdc_log;                                                                                                                                                                                                                 

 system.totimestamp(time)        | batch_seq_no | _pk | _ck | _v           | operation
---------------------------------+--------------+-----+-----+--------------+-----------
 2020-02-03 17:47:04.946000+0000 |            0 |   0 |   0 | (0, 0, None) |         1
 2020-02-03 17:47:13.162000+0000 |            0 |   0 |   0 |         null |         0
 2020-02-03 17:47:13.162000+0000 |            1 |   0 |   0 |         null |         1

(3 rows)
```

The first row corresponds to the first insert. There is no preimage because there was no row with `pk = 0, ck = 0` --- good.

But the following two rows, which correspond to the _row delete_
```
delete from ks.t where pk = 0 and ck = 0
```
are wrong?
(note: this is different from `delete v from ks.t where pk = 0 and ck = 0`, which is an update, equivalent to `update ks.t set v = null where pk = 0 and ck = 0`)

1. The `_v` of preimage shows `null`, when it should show `(0, 0, None)` (first `0` because the value was previously _set_, second `0` because it was equal to `0`, and `None` because no ttl).
2. The `operation` of delta is `1`, while it should be `2` (_row delete_).

The problem is that CDC code doesn't recognize row deletes. This can be done by looking at tombstones:
1. inserting a row with `v = null` gives a tombstone under the `v` cell, but doesn't insert a row tombstone
2. deleting a row gives a row tombstone which kills all cells (including the row marker)

@haaawk @elcallio

kbr-:
Not recognizing row deletes is one problem (that's why `operation = 1`) but I don't get why preimage was null.

kbr-:
I updated the problem description with more accurate information

[...]

PR #5897, opened 2020-02-25
cdc: Fix `operation` value for row deletes

jul-stas:
Column `operation` now contains `operation::row_delete` (== 2)
after queries like `delete from tbl where pk=x and ck=y;`. Before
this patch row deletes were treated as updates, which was incorrect
because updates do not contain row tombstones (and row deletes do).

Refs #5709

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -752,7 +752,8 @@ public:
                 process_cells(r.row().cells(), column_kind::regular_column);
                 process_cells(p.static_row().get(), column_kind::static_column);
 
-                set_operation(log_ck, ts, operation::update, res);
+                const operation cdc_op = r.row().deleted_at() ? operation::row_delete : operation::update;
+                set_operation(log_ck, ts, cdc_op, res);
                 if (ttl) {
                     set_ttl(log_ck, ts, *ttl, res);
                 }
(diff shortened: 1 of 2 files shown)


piodul:
@jul-stas Please add a unit test that checks that we do differentiate updates from deletes.

haaawk:
Note that this will work only if mutations will be split by timestamps which is ok.

jul-stas:
Update: rebased onto Calle's patch

haaawk:
@nyh please have a look.

nyh:
Here is a comment I tried to write yesterday when github had a bad hair day:

I guess this assumes that a "normal" (from CQL) update cannot include both a row deletion *and* an update, right? Indeed I can't think of a CQL command that generates both.
This is actually common in collections that guess assigned but I guess the collection code is orthogonal to this  - it's not a "row" but a different sort of prefix.

haaawk:
> Here is a comment I tried to write yesterday when github had a bad hair day:
> 
> I guess this assumes that a "normal" (from CQL) update cannot include both a row deletion _and_ an update, right? Indeed I can't think of a CQL command that generates both.
> This is actually common in collections that guess assigned but I guess the collection code is orthogonal to this - it's not a "row" but a different sort of prefix.

Let's get @elcallio opinion.

nyh:
@haaawk I'll commit this, and he can always improve things further later if necessary.

haaawk:
Go ahead @nyh

PR #5905, opened 2020-02-26
CDC: split mutations by timestamp and ttl

kbr-:
Add a bunch of new structs describing a change made
to a table, and an `extract_changes` function which takes a mutation and
returns the set of changes contained in this mutation, separated by
timestamp and ttl.

Add a `split` function which uses `extract_changes` to split a mutation into separate mutations, each describing a single change.

Static rows are put into separate changes now.

The `pre_image_select` function was fixed to select pre_image data always when there is a static row/clustered row change, even if there were e.g. additional range tombstones.

Fixes: https://github.com/scylladb/scylla/issues/5719.

Tests: unit(dev)


haaawk:
Kamil and I are still testing this PR and adding unit tests for it. Commit messages will be improved as well. Please have a look on the code in the meantime and let us know what you think.

cdc/split.cc
```diff
@@ -0,0 +1,179 @@
...
+            auto& cdef = base_schema.column_at(column_kind::regular_column, id);
+            if (cdef.is_atomic()) {
+                auto view = cell.as_atomic_cell(cdef);
+                if (check_or_set(view.timestamp(), view.is_live_and_has_ttl() ? view.ttl() : gc_clock::duration(0))) {
+                    should_split = true;
+                    return;
+                }
+            }
+
+            cell.as_collection_mutation().with_deserialized(*cdef.type, [&] (collection_mutation_view_description mview) {
```
nyh: Again, don't you need "else" here?

cdc/split.cc
```diff
@@ -0,0 +1,179 @@
...
+            found_ts = row_tomb.timestamp;
+        }
+    }
+
+    if (!p.row_tombstones().empty() && (had_static_row || had_clustered_row)) {
+        return true;
+    }
+
+    for (const auto& rt: p.row_tombstones()) {
+        if (rt.tomb) {
```
nyh: Don't you need to check here had_cells (or had_clustered_row or had_static_row), and if true always return true (like you did above for row tombstones, or below for partition tombstones)?

cdc/split.cc
```diff
@@ -0,0 +1,179 @@
...
+                return true;
+            }
+
+            found_ts = rt.tomb.timestamp;
+        }
+    }
+
+    if (p.partition_tombstone().timestamp != api::missing_timestamp
+            && (!p.row_tombstones().empty() || had_static_row || had_clustered_row)) {
+        return true;
```
nyh: I guess this is very unlikely and un-CQL-like, but there's also the possibility that we have both a partition tombstone and other tombstones (range tombstones?) with different times, and this also can't be represented in a single line?

cdc/split.cc
```diff
@@ -24,6 +24,203 @@
 
 #include "split.hh"
 
+struct atomic_column_update {
+    column_id id;
+    atomic_cell cell;
+};
+
+struct nonatomic_column_deletion {
```
nyh: Just curious why we need specifically nonatomic_column_deletion, how does it differ from an atomic one? Are you including deletion of atomic columns as just an atomic_update? (if so, maybe a comment clarifying this can be helpful).

cdc/split.cc
```diff
@@ -24,6 +24,203 @@
...
+};
+
+struct clustered_row_insert {
+    gc_clock::duration ttl;
+    clustering_key key;
+    row_marker marker;
+    std::vector<atomic_column_update> atomic_entries;
+    std::vector<nonatomic_column_deletion> nonatomic_deletions;
+    // INSERTs can't express updates of individual cells inside a non-atomic
+    // (without deleting the entire field first), so no `nonatomic_updates` field
```
nyh: I'm confused how an "insert" is different from an "update"...

cdc/split.cc
```diff
@@ -24,6 +24,203 @@
...
+            auto view = cell.as_atomic_cell(cdef);
+            auto timestamp_and_ttl = std::pair(
+                    view.timestamp(),
+                    view.is_live_and_has_ttl() ? view.ttl() : gc_clock::duration(0)
+                );
+            result[timestamp_and_ttl].atomic_entries.push_back({id, atomic_cell(*cdef.type, view)});
+            return;
+        }
+
+        cell.as_collection_mutation().with_deserialized(*cdef.type, [&] (collection_mutation_view_description mview) {
```
nyh: Ok, so here comes the code duplication I asked about in the should_split() patch :-)
And again here I have to ask why there's no else() here (I guess if I was wrong asking this before, I'm wrong again...)

cdc/log.cc
```diff
@@ -630,7 +630,7 @@ class transformer final {
             const mutation& m)
     {
         auto& p = m.partition();
-        if (p.partition_tombstone() || !p.row_tombstones().empty() || p.clustered_rows().empty()) {
+        if (p.clustered_rows().empty() && p.static_row().empty()) {
```
nyh: I don't understand this change. A better commit message would have helped to understand what you are fixing here.

cdc/log.cc
```diff
@@ -737,7 +750,15 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
             // so this is premature.
             auto f = trans.pre_image_select(qs.get_client_state(), db::consistency_level::LOCAL_QUORUM, m);
             return f.then([trans = std::move(trans), &mutations, idx] (lw_shared_ptr<cql3::untyped_result_set> rs) mutable {
-                mutations.push_back(trans.transform(mutations[idx], rs.get()));
+                auto& m = mutations[idx];
+                auto s = m.schema();
```
nyh: nitpick: this makes a copy of schema, incrementing and decrementing its reference counter. Do we need to?

haaawk:
Thanks, @nyh. As I mentioned in the comment above, commit messages and tests will be added. We posted the code to be able to get your feedback as soon as possible while we're working on tests.

cdc/log.cc
haaawk: Commit message will be added but in the meantime, I will explain.

We want to generate preimage any time there's a clustered or static row in the mutation.
The mutation can contain multiple operations: partition deletion, range tombstone, insert/update/delete in clustering or static row
Any operation on clustering or static row requires preimage. Other operations (partition deletion and range tombstone) have empty preimage.
That's why we want to obtain preimage when there's at least one clustering or static row in the mutation. It means there's operation on clustering or static row and we need preimage for it.

Please let me know if I failed to explain it, @nyh

cdc/split.cc
```diff
@@ -0,0 +1,179 @@
...
+
+    bool should_split = false;
+    p.static_row().get().for_each_cell([&] (column_id id, const atomic_cell_or_collection& cell) {
+        had_static_row = true;
+
+        auto& cdef = base_schema.column_at(column_kind::static_column, id);
+        if (cdef.is_atomic()) {
+            auto view = cell.as_atomic_cell(cdef);
+            if (check_or_set(view.timestamp(), view.is_live_and_has_ttl() ? view.ttl() : gc_clock::duration(0))) {
+                should_split = true;
```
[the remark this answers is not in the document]
tgrabiec: Better to use `mutation_partition_visitor` because when we extend the mutation model this code will become silently broken.

cdc/log.cc
```diff
@@ -657,13 +658,18 @@ class transformer final {
...
+            p.static_row().get().for_each_cell([&] (column_id id, const atomic_cell_or_collection&) {
+                auto& cdef =_schema->column_at(column_kind::static_column, id);
+                static_columns.emplace_back(id);
+                columns.emplace_back(&cdef);
+            });
+        }
+        if (!p.clustered_rows().empty()) {
+            p.clustered_rows().begin()->row().cells().for_each_cell([&] (column_id id, const atomic_cell_or_collection&) {
+                auto& cdef =_schema->column_at(column_kind::regular_column, id);
+                regular_columns.emplace_back(id);
```
elcallio: Could just have added an `.. cids, kind, include] : { std::tie(..., !p.static_row().empty()), std::tie(..., !p.clustered_rows().empty()))` and avoided the repetition. Remember: if it is worth doing, it is worth doing as a loop.

cdc/log.cc
```diff
@@ -722,8 +728,20 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
...
+
             if (!s->cdc_options().preimage()) {
-                mutations.emplace_back(trans.transform(m));
+                if (should_split(m, *s)) {
+                    for(auto&& mm : split(m, s)) {
+                        mutations.emplace_back(trans.transform(mm));
+                    }
+                } else {
+                    mutations.emplace_back(trans.transform(m));
+                }
```
elcallio: Again, code repetition. Why not just make a ready future `<result_set null>`, and set it to result of call iff preimage is enabled?

cdc/split.cc
```diff
@@ -0,0 +1,449 @@
...
+                auto timestamp_and_ttl = std::pair(
+                        v.timestamp(),
+                        v.is_live_and_has_ttl() ? v.ttl() : gc_clock::duration(0)
+                    );
+                auto& updates = result[timestamp_and_ttl].nonatomic_updates;
+                if (updates.empty() || updates.back().id != id) {
+                    updates.push_back({id, {}});
+                }
+                updates.back().cells.push_back({std::move(k), std::move(v)});
+            }
```
elcallio: Do we really need to care about cell ttls/timestamps? Afaik, we never generate disparate cell timestamps/ttls in any collection code? And the tomstone timestamp should be more or less ignorable also, since it is only a stop marker, guaranteed to be before the cells (which again should be ts:ed to the whole row, no?).

haaawk:
At the moment following tests cdc_test tests are still failing:
test_pre_image_logging
test_primary_key_logging

cdc/log.cc
```diff
@@ -722,8 +728,20 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
 
             transformer trans(_ctxt, s);
 
+            std::vector<mutation> muts;
```
haaawk: This and next 4 lines are leftovers of previous version of the code and should be removed.

haaawk:
> At the moment following tests cdc_test tests are still failing:
> test_pre_image_logging
> test_primary_key_logging

I fixed those two but still the following tests are failing:
test_udt_logging
test_list_logging
test_set_logging
test_map_logging

cdc/split.cc
```diff
@@ -0,0 +1,179 @@
...
+ */
+
+#include "mutation.hh"
+#include "schema.hh"
+
+#include "split.hh"
+
+namespace cdc {
+
+bool should_split(const mutation& base_mutation, const schema& base_schema) {
```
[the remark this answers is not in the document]
kbr-: In the future I would like to refactor everything here and do the same thing I did in my previous version of the PR (https://github.com/scylladb/scylla/pull/5803), where I do not create a vector of mutations which is then `transform`ed in CDC, but instead use the `batch` struct directly in `transform`. This would mean we always call `extract_changes` and `should_split` will be gone.

We decided to abandon the previous PR because it did a lot of changes in `transform` which would conflict with changes made by other people. The plan is to merge this PR before the next release, without any major changes to `transform`, and then to refactor in a later PR, with the final result looking like in the previous abandoned PR.

So while I agree with the code duplication issue, I think we shouldn't worry about it now.

cdc/split.cc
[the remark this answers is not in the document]
kbr-: @tgrabiec won't the type system and/or tests catch the breakage?

Also, if this code would become broken, won't `mutation_partition_visitor` be broken too?

By the way, `transform` in cdc/log.cc also has this problem because it walks around the mutation manually. But I plan to refactor it (we didn't in this PR so to not conflict with changes made concurrently e.g. by @elcallio), and then I'll see if `mutation_partition_visitor` is up to the job. Thanks for the heads-up.

cdc/split.cc
```diff
@@ -0,0 +1,179 @@
...
+        auto& cdef = base_schema.column_at(column_kind::static_column, id);
+        if (cdef.is_atomic()) {
+            auto view = cell.as_atomic_cell(cdef);
+            if (check_or_set(view.timestamp(), view.is_live_and_has_ttl() ? view.ttl() : gc_clock::duration(0))) {
+                should_split = true;
+                return;
+            }
+        }
+
+        cell.as_collection_mutation().with_deserialized(*cdef.type, [&] (collection_mutation_view_description mview) {
```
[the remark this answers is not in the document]
kbr-: Correct, needs fixing.

cdc/split.cc
kbr-: Correct, needs fixing.

cdc/split.cc
kbr-: But this is the `row_tombstones` case, the check you're talking about is right above :)

cdc/split.cc
kbr-: Yes, and I'm checking for all these cases right here:
```
    if (p.partition_tombstone().timestamp != api::missing_timestamp
            && (!p.row_tombstones().empty() || had_static_row || had_clustered_row))
```
What's the issue?

cdc/split.cc
kbr-: The main reason I separated this structure from `nonatomic_column_update` (instead of keeping the tombstone inside `nonatomic_column_update` together with the cells) is for the type system to protect me from accidentally putting cells of a nonatomic column into a `clustered_row_insert`. `clustered_row_insert` doesn't have any `nonatomic_column_update`s, as you can see in the struct definition below.

The explanation why do this is fairly complicated. When in CQL we do
```insert into tab (pk,ck,coll) values (0,0,{...})```
where `coll` is a nonatomic collection column and `{...}` is a collection of this type, this creates a tombstone and a bunch of cells. 
However, if we want to create just a tombstone, there's a way with `insert`:
```insert into tab (pk,ck,coll) values (0,0,null)```
but there is no way to create just the cells with an `insert`. To do that, you have to use `update`:
```update tab set coll = coll + {...} where pk = 0 and ck = 0```

Due to this difference of power of expression between `insert` and `update` I decided that for consistency, I will always split the deletion from the cell addition. Thus, this `insert`:
```insert into tab (pk,ck,coll) values (0,0,{...})```
will be split into two changes in CDC: 
```
insert into tab (pk,ck,coll) values (0,0,null)
update tab set coll = coll + {...}
```
there's another reason to do the split above: the tombstone has a different timestamp than the cells (the difference is `-1`).

I agree this deserves a comment somewhere.

Comment (kbr-) [reply cdc/split.cc]: And now I see I did write a comment inside `clustered_row_insert` definition. But I'll add a reference here too.

File: cdc/split.cc (hunk 5 above)
Comment (kbr-) [reply cdc/split.cc]: `insert` creates a row marker, `update` doesn't.
From the user's perspective, with `insert` you can create a row that will be visible even if all of its non-key columns are `null`. You can't do that with an `update`.

File: cdc/split.cc (hunk 6 above)
Comment (kbr-) [reply cdc/split.cc]: About the `else`: correct, needs fixing.

I plan to remove the duplication later, when refactoring `transform` in cdc/log.cc to use the `batch` struct directly instead of unnecessarily creating a bunch of `mutation`s like we do now in the `split` function.

File: cdc/split.cc (hunk 12 above)
Comment (kbr-) [reply cdc/split.cc]: > Afaik, we never generate disparate cell timestamps/ttls in any collection code?

We need to handle this somehow:
```
begin batch
update tab using timestamp 42 set coll = coll + {0} where pk = 0 and ck = 0;
update tab using timestamp 43 set coll = coll + {1} where pk = 0 and ck = 0;
apply batch;
```

This will merge the two `collection_mutation`s, so we'll have a `collection_mutation` with cells of different timestamps.

> And the tomstone timestamp should be more or less ignorable also, since it is only a stop marker, guaranteed to be before the cells (which again should be ts:ed to the whole row, no?)

A mutation that represents the following:
```
begin batch
update tab using timestamp 42 set coll = null where pk = 0 and ck = 0;
update tab using timestamp 43 set coll = coll + {0} where pk = 0 and ck = 0;
apply batch;
```
cannot be distinguished from a mutation that represents the following:
```
update tab using timestamp 43 set coll = {0} where pk = 0 and ck = 0
```
so should we represent it in CDC as two changes or one? I decided to go for the first option (two changes), and from the top of my head I can give two reasons:
- there are two different timestamps here, so it will be consistent with the fact that we're splitting different timestamps into different changes everywhere else
- another reason is related to `ttl`s. Again for consistency / to reduce the number of edge cases I decided to group dead cells with `ttl = null`. This is because we now have this `ttl` column which can sometimes be non-null, and it doesn't really make sense to have a change with `ttl != null` and a dead cell, for example, because dead cells cannot have ttls. So the first option above makes even more sense if we do
`
update tab using timestamp 43 and ttl 5 set coll = {0} where pk = 0 and ck = 0
`
because now we'll have two changes: one that sets `coll = null` at timestamp `42` and ttl `null`, and one that sets `coll = coll + {0}` (adds the `0` cell to `coll`) at timestamp `43` with ttl `5`.

File: cdc/log.cc (hunk 8 above)
Comment (kbr-) [reply cdc/log.cc]: It will be copied anyway in each `split(m, s)` call, but OK, I changed this to `auto&`, one less copy.

File: cdc/log.cc (hunk 10 above)
Comment (kbr-) [reply cdc/log.cc]: https://www.youtube.com/watch?v=WWaLxFIVX1s

File: cdc/log.cc (hunk 11 above)
Comment (kbr-) [reply cdc/log.cc]: Good idea, will do.

**Comment by kbr-** [comment]
v2:
- rebased
- added a small comment
- wrote the commit log
- fixed tests

I'm still writing new tests. However, I have encountered an issue.

First, the nonatomic column code forgot to set the `ttl` column, so I can't really test whether ttls are split correctly. Oh well.
Second, changes from a single CQL batch will be now put into separate CDC batches even if they have the same partition key and timestamp. They'll be put into different `time` rows. I don't like this, I'd like to have such changes have the same `time` value and different `batch_seq_no` values instead. This maybe deserves discussion.

**Comment by kbr-** [comment]
v2:
- rebased
- added a small comment
- wrote the commit log
- fixed existing tests
- added a new test that checks whether changes are split and represented in the log correctly

There's a small issue which probably deserves discussion: changes from a single CQL batch will be now put into separate CDC batches even if they have the same partition key and timestamp. They'll be put into different `time` rows. I don't like this, I'd like to have such changes have the same `time` value and different `batch_seq_no` values instead. This would require passing a `batch_seq_no` parameter into `transform` and instead of returning a set of mutations from `split`, calling `transform` from inside `split` (so that it would make sure to pass the correct `batch_seq_no`).

File: cdc/log.cc (hunk 10 above)
Comment (elcallio) [reply cdc/log.cc]: Not sure what that is supposed to entail? Would you care to defend your code repetition more verbosely?

Comment (kbr-) [reply cdc/log.cc]: Sure.

>  if it is worth doing, it is worth doing as a loop.

I strongly disagree. The loop that was already there was unreadable, and you're proposing to make it even more unreadable.

I agree that code repetition is bad in general, but sometimes it's better than the alternatives.

File: cdc/split.cc (hunk 6 above)
Comment (elcallio) [reply cdc/split.cc]: Again, I must question that "wisdom". Until you prove to me that ttl/timestamp splitting is frequent enough to warrant always building these huge, contrived structures filled with allocated vectors for every mutation, I still think that even this approach is a bit of an overkill. Why not just change the transform loop to generate a new cdc mutation every time we detect a ttl/timestamp change? Can be done in a forward loop, and should generate zero overhead in (what I imagine to be) the common case.

Comment (haaawk) [reply cdc/split.cc]: Let's not discuss this now. We will refactor the code and we will do it in such a way to favour common case not the corner case. Your comment on optimizing for a common case is right and was well heard @elcallio.

File: test/boost/cdc_test.cc [hunk 16]
```diff
@@ -643,19 +646,29 @@ static void test_collection(cql_test_env& e, data_type val_type, std::vector<col
...
 
-        auto val = *updates.back()[val_index];
-        auto tup = value_cast<tuple_type_impl::native_type>(col_type->deserialize(bytes_view(val)));
+        BOOST_REQUIRE_GE(updates.size(), t.changes.size());
 
-        if (!t.next.is_null()) {
-            BOOST_REQUIRE_EQUAL(t.next, f(tup.at(1)));
-        }
+        for (size_t change_ix = 0; change_ix < t.changes.size(); ++change_ix) {
+            auto& change = t.changes[change_ix];
```
elcallio: Iterators. Indexes are so C89

haaawk:
Forward iteration is one of the options we may take when we will be refactoring @elcallio

kbr-:
@elcallio consider a map consisting of the following cells:
{0:0, ttl=1}, {1:1, ttl=2}, {2:2, ttl=1}, {3:3, ttl=2}, {4:4, ttl=1}, {5:5, ttl=2}, and so on...
such a map can be created using two updates.

However, a forward iterator approach will create n different updates, where n is the number of cells.

It may be better for the most common case, sure. But maybe it won't? It's a micro-optimization, so if we're arguing for needing proofs, then whoever proposes to do the optimization should provide proof (that there's a significant improvement). Otherwise we should aim for maximum readability, maintainability, simplicity, ease of use.

elcallio:
> 
> 
> @elcallio consider a map consisting of the following cells:
> {0:0, ttl=1}, {1:1, ttl=2}, {2:2, ttl=1}, {3:3, ttl=2}, {4:4, ttl=1}, {5:5, ttl=2}, and so on...
> such a map can be created using two updates.
> 
> However, a forward iterator approach will create n different updates, where n is the number of cells.
> 
> It may be better for the most common case, sure. But maybe it won't? It's a micro-optimization, so if we're arguing for needing proofs, then whoever proposes to do the optimization should provide proof (that there's a significant improvement). Otherwise we should aim for maximum readability, maintainability, simplicity, ease of use.

I fail to see how a forward approach would be less readable. And note that it is not certain you would have to generate N mutations either. If you allow referencing back to "did I have a mutation with ttl X", you could generate more or less equally optimal mutations with a forward approach. 

I am _really_ allergic to allocations if they can be avoided, esp. when they, like in this case, cannot be know in size until the source mutation is fully traversed -> either both reallocs and probably try to compensate by over-allocating initial usage -> extra overhead for common (small) case.

kbr-:
> If you allow referencing back to "did I have a mutation with ttl X"

But that's exactly what I'm doing here, that's why the `std::map` is for.
I can't "push" the change to the consumer before I'm sure that I saw all cells with a given TTL. So I potentially need to keep the change allocated for the entire duration of the algorithm.

However, perhaps we could reduce the number of allocations by using intrusive data structures maybe?

test/boost/cdc_test.cc
kbr-: I need the `change_ix` in multiple places below.
Could use an iterator and then subtract `changes.begin()` to get `change_ix`. Is that what you'd prefer?

kbr-:
v3:
- rebased on top of Calle's change https://github.com/scylladb/scylla/pull/5910

haaawk:
I rebased and fixed the test @nyh. You can find it here: https://github.com/scylladb/scylla/pull/5940

cdc/split.cc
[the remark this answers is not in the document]
tgrabiec: The type system won't catch the breakage if it's caused by adding new elements because you're currently explicitly accessing the elements you expect by calling appropriate getters.

`mutation_partition_visitor` is part of the mutation model, so it can be trusted by CDC. The visitor has higher chance of being up to date, since it's the foundation of many algorithms which consume the model.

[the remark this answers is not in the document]
haaawk: @tgrabiec is right. We're planning some serious refactoring after 4.0 is cut. We definitely should use mutation_partition_visitor then.

PR #5910, opened 2020-02-26
CDC: implement new column format and naming

elcallio:
Rename metadata and data columns according to new spec
    
Also use transformation methods for names in all code + tests
to make switching again easier

Break up data column tuple
    
Data column is now pure frozen original type. 

- If column is deleted (set to null), a metadata column cdc$deleted_<name> is set to true, to distinguish null column == not involved in row operation
- For non-atomic collections, a cdc$deleted_elements_<name> column is added, and when removing elements from collection this is where they are shown.
- For non-atomic assign, the "cdc$deleted_<name>" is true, and <name> is set to new value.
    
column_op removed.

v2:
Modified to use uniform "key" set semantics for all non-atomic types, i.e. udts put deleted fields as field index in `deleted_elements` set.

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -237,6 +237,7 @@ using operation_native_type = std::underlying_type_t<operation>;
 using column_op_native_type = std::underlying_type_t<column_op>;
 
 static const sstring cdc_log_suffix = "_scylla_cdc_log";
+static const sstring cdc_meta_column_prefix = "cdc$";
 
 static bool is_log_name(const std::string_view& table_name) {
     return boost::ends_with(table_name, cdc_log_suffix);
@@ -255,15 +256,31 @@ sstring log_name(const sstring& table_name) {
     return table_name + cdc_log_suffix;
 }
 
+sstring log_data_column_name(std::string_view column_name) {
+    return sstring(column_name);
+}
+
+seastar::sstring log_meta_column_name(std::string_view column_name) {
+    return cdc_meta_column_prefix + sstring(column_name);
+}
+
+bytes log_data_column_name_bytes(const bytes& column_name) {
+    return column_name;
+}
+
+bytes log_meta_column_name_bytes(const bytes& column_name) {
+    return to_bytes(cdc_meta_column_prefix) + column_name;
+}
+
 static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID> uuid) {
     schema_builder b(s.ks_name(), log_name(s.cf_name()));
     b.set_comment(sprint("CDC log for %s.%s", s.ks_name(), s.cf_name()));
-    b.with_column("stream_id_1", long_type, column_kind::partition_key);
-    b.with_column("stream_id_2", long_type, column_kind::partition_key);
-    b.with_column("time", timeuuid_type, column_kind::clustering_key);
-    b.with_column("batch_seq_no", int32_type, column_kind::clustering_key);
-    b.with_column("operation", data_type_for<operation_native_type>());
-    b.with_column("ttl", long_type);
+    b.with_column(log_meta_column_name_bytes("stream_id_1"), long_type, column_kind::partition_key);
+    b.with_column(log_meta_column_name_bytes("stream_id_2"), long_type, column_kind::partition_key);
+    b.with_column(log_meta_column_name_bytes("time"), timeuuid_type, column_kind::clustering_key);
+    b.with_column(log_meta_column_name_bytes("batch_seq_no"), int32_type, column_kind::clustering_key);
+    b.with_column(log_meta_column_name_bytes("operation"), data_type_for<operation_native_type>());
+    b.with_column(log_meta_column_name_bytes("ttl"), long_type);
     auto add_columns = [&] (const schema::const_iterator_range_type& columns, bool is_data_col = false) {
         for (const auto& column : columns) {
             auto type = column.type;
@@ -292,7 +309,7 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID> 
 
                 type = tuple_type_impl::get_instance({ /* op */ data_type_for<column_op_native_type>(), /* value */ type->freeze()});
             }
-            b.with_column("_" + column.name(), type);
+            b.with_column(log_data_column_name_bytes(column.name()), type);
         }
     };
     add_columns(s.partition_key_columns());
@@ -457,7 +474,7 @@ private:
         size_t pos = 0;
         for (const auto& column : _schema->partition_key_columns()) {
             assert (pos < pk_value.size());
-            auto cdef = m.schema()->get_column_definition(to_bytes("_" + column.name()));
+            auto cdef = m.schema()->get_column_definition(log_data_column_name_bytes(column.name()));
             auto value = atomic_cell::make_live(*column.type,
                                                 ts,
                                                 bytes_view(pk_value[pos]),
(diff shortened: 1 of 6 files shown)


cdc/log.cc
```diff
@@ -306,10 +323,13 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID>
                         return type.freeze();
                     }
                 ));
-
-                type = tuple_type_impl::get_instance({ /* op */ data_type_for<column_op_native_type>(), /* value */ type->freeze()});
+                type = type->freeze();
             }
             b.with_column(log_data_column_name_bytes(column.name()), type);
+            b.with_column(log_data_column_deleted_name_bytes(column.name()), boolean_type);
+            if (type->is_collection()) {
```
haaawk: Is this distinguishing between frozen and non-frozen mutations and adding column only to non-frozen collections?

cdc/log.cc
```diff
@@ -693,9 +714,7 @@ class transformer final {
                                     [&](const user_type_impl& type) {
                                         std::vector<bytes_opt> res(type.size());
                                         std::array<bytes_view_opt, 1> tmp;
-                                        // type of the actual value in log column tuple.
-                                        auto res_type = static_pointer_cast<const tuple_type_impl>(dst->type)->type(1);
-                                        auto tt = static_pointer_cast<const tuple_type_impl>(res_type);
+                                        auto tt = static_pointer_cast<const tuple_type_impl>(dst->type);
```
haaawk: Is the code candling non-frozen UDTs correctly? Shouldn't it create deleted_elements for non-frozen UDTs as well?

haaawk:
LGTM overall - I left some questions.

piodul:
@elcallio What about merging `stream_id_1` and `stream_id_2` into one field `stream_id`? The new schema we agreed upon changes this, but it seems that it these columns are not merged in this PR (they only have a `cdc$` prefix added)

test/boost/cdc_test.cc
```diff
@@ -644,15 +644,15 @@ SEASTAR_THREAD_TEST_CASE(test_ttls) {
 using translate_func = std::function<data_value(data_value)>;
 struct col_test {
     sstring update;
-    data_value prev;
+    data_value prev;
```
haaawk: Remove whitespace

cdc/log.cc
elcallio: No, it is not. Will fix.

cdc/log.cc
elcallio: No. UDT:s cannot use a "deleted_elements", since UDT parts have different types. This is why we have the dreaded `tuple<tuple<typ1>, tuple<typ2>...>`

test/boost/cdc_test.cc
elcallio: ok

elcallio:
> 
> 
> @elcallio What about merging `stream_id_1` and `stream_id_2` into one field `stream_id`? The new schema we agreed upon changes this, but it seems that it these columns are not merged in this PR (they only have a `cdc$` prefix added)

Not part of this PR. That is for the PK and topology team to deal with.

elcallio:
Do not merge yet. I think rebase caused an issue.

elcallio:
Fixed.

kbr-:
Also you can update the cover letter:
> Note: this PR currently includes the non-atomic support PR, because it has not passed into master yet. Once it is in, comparison should be easier. Until then, please just look at the two latter patches.

cdc/log.cc
kbr-: Sorry, noticed this discussion just now. Copy-pasting my comment from below:

Is there no way to use the deleted_elements_ column for UDTs as well?

E.g.:
-    for deleted elements, just use a set of field indexes. We could also use a set of strings (field names) I guess, but that's not resistant to field renames, so field indexes seem better.
-    for elements set to non-null value, use the original UDT type, but frozen.
-    for fields that were not affected, just don't set them in the frozen UDT.

elcallio:
@kbr True. While I somewhat abhor the structure, it is more consistent with the rest of non-atomics. Will implement.

kbr-:
By the way, why didn't we go for showing `timeuuid` indices for list operations?
I see in the test that for additions, we use just values, and for deletions we don't show anything.
This means there's no way to replay the change.

elcallio:
> 
> 
> By the way, why didn't we go for showing `timeuuid` indices for list operations?
> I see in the test that for additions, we use just values, and for deletions we don't show anything.
> This means there's no way to replay the change.

You misread, but it is more clear in latest version: Lists are stored as map<timeuuid, value>, and deleted_elements are set<timeuuid>. 
The test does not explicitly verify this (yet), because we have no way of knowing the timeuuid of the results without reading preimage or similar, so test becomes ikky.

elcallio:
@haaawk @kbr -  Now updated to use uniform `deleted_elements` semantics for udts and collections alike.

cdc/log.cc
```diff
@@ -287,29 +304,35 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID>
...
+                auto dtype = visit(*type, make_visitor(
+                    // lists are represented as map<timeuuid, value_type>. Otherwise we cannot express delta
+                    [] (const collection_type_impl& type) -> data_type {
+                        return set_type_impl::get_instance(type.name_comparator(), false);
+                    },
+                    // user types deletes are a set of the indices removed
+                    [] (const user_type_impl& type) -> data_type {
+                        return set_type_impl::get_instance(short_type, false);
+                    },
+                    // everything else is just frozen self
```
haaawk: This comment and the exception throw below somehow don't match in my mind. Is the comment really in the right place?

elcallio: doh. fixed.

haaawk: Thanks.

haaawk:
@nyh I think we're ready for you :D

kbr-:
I just noticed one thing: for the `deleted_` columns, if the corresponding base column was not deleted, we leave it to be `null`, instead of setting it to `False`. I'm not saying it's bad, I'm just wondering whether this is a conscious choice and if it's in any way better or worse than the alternative.

haaawk:
> I just noticed one thing: for the `deleted_` columns, if the corresponding base column was not deleted, we leave it to be `null`, instead of setting it to `False`. I'm not saying it's bad, I'm just wondering whether this is a conscious choice and if it's in any way better or worse than the alternative.
> 
> (an upside is that it doesn't create a cell, i.e. less memory/disk consumption)

I noticed that as well but my impression was that it is a conscious choice.

kbr-:
Another thing I noticed just now. When we don't add any cells, only delete the collection, the column which represents adding elements to collection is set to `{}`, instead of being `null`:
```
cqlsh> update ks.t set m = null where pk = 3 and ck = 0;
cqlsh> select m, "cdc$deleted_m", "cdc$deleted_elements_m" from ks.t_scylla_cdc_log where pk = 3 allow filtering;

 m | cdc$deleted_m | cdc$deleted_elements_m
---+---------------+------------------------
  {} |          True |                   null

---MORE---
(1 rows)
```
why not leave it `null` instead @elcallio ? Would be consistent with leaving the `deleted_` column to be equal to `null` when we don't delete the collection.

For user defined types it shows up like this:
```
{a: null, b: null}
```
might be a bit confusing.

elcallio:
Wholly conscious. Absent bool == false.

elcallio:
@kbr - it is just the logic right now, since empty/null represents the same set of elements. But it could be optimized to simply null the value iff empty. 
Can do that in follow up though.

elcallio:
@kbr - updated to null columns iff no elements.

nyh:
Committed to next.

haaawk:
> Committed to next.

Great - thanks Nadav!!!

PR #5931, opened 2020-03-01
cdc: distinguishing update from insert

### LINKED ISSUE #5723: CDC: inserts and updates should be different operations
Because they have different effects. CDC log's "operation" column should provide that information.
(and until yesterday I thought insert and update is the same thing, "upsert"...)

```
cqlsh> update ks.t set v = null where pk = 0 and ck = 0;
cqlsh> select * from ks.t where pk = 0 and ck = 0;

 pk | ck | v
----+----+---

(0 rows)
cqlsh> insert into ks.t (pk,ck,v) values (0,0,null);
cqlsh> select * from ks.t where pk = 0 and ck = 0;

 pk | ck | v
----+----+------
  0 |  0 | null

(1 rows)
cqlsh> 
```

the difference is that `insert` _inserts a live row marker_. `update` doesn't touch the row marker at all.

By the way: `update` with setting a column to `null` is the same as a _column delete_:
```
update ks.t set v = null where pk = 0 and ck = 0;
```
is the same as
```
delete v from ks.t where pk = 0 and ck = 0;
```
column delete also doesn't touch the row marker, it just inserts a dead cell into the selected column, just like `update` with `null`. Thus we don't need to distinguish a column delete from a null update.

In cdc/log.hh we have:
```
enum class operation : int8_t {
    // note: these values will eventually be read by a third party, probably not privvy to this
    pre_image = 0, update = 1, row_delete = 2, range_delete_start = 3, range_delete_end = 4, partition_delete = 5
};
```
we need to add another entry: "insert".

Note that null update/column delete is different from a row delete. This issue is related to https://github.com/scylladb/scylla/issues/5709, but it's different. Here's a summary of differences between inserts, updates, and row deletes:
1. `insert` inserts a live row marker and maybe updates some columns (if specified).
2. `update` doesn't touch the row marker at all, only updates some columns.
3. row `delete` (but not a column `delete`, which is the same as `update`) inserts a _row tombstone_ which kills all cells, including the row marker.

These 3 cases should be distinguished in CDC.

jul-stas:
When incoming mutation contains live row marker the `operation` is
described as "insert", not as an "update".

Test cases that relied on "update" being the same as "insert" are
updated accordingly (`test_pre_image_logging`, `test_cdc_across_shards`,
`test_add_columns`). **It had to be done in one commit to avoid test-breaking
revision**. Also, I extended the test case "test_row_delete" with one insert,
which is expected to log different value of `operation` than update
or delete. Renamed the test case accordingly.

Fixes #5723

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -542,7 +542,7 @@ public:
                 }
             }
         } else {
-            // should be update or deletion
+            // should be insert, update or deletion
             int batch_no = 0;
             for (const rows_entry& r : p.clustered_rows()) {
                 auto ck_value = r.key().explode(*_schema);
@@ -752,8 +752,16 @@ public:
                 process_cells(r.row().cells(), column_kind::regular_column);
                 process_cells(p.static_row().get(), column_kind::static_column);
 
-                const operation cdc_op = r.row().deleted_at() ? operation::row_delete : operation::update;
+                operation cdc_op;
+                if (r.row().deleted_at()) {
+                    cdc_op = operation::row_delete;
+                } else if (marker.is_live()) {
+                    cdc_op = operation::insert;
+                } else {
+                    cdc_op = operation::update;
+                }
                 set_operation(log_ck, ts, cdc_op, res);
+
                 if (ttl) {
                     set_ttl(log_ck, ts, *ttl, res);
                 }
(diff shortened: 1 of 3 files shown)


nyh:
Thanks. Committed to next (as one patch, manually, so github doesn't know so I need to close the PR myself).

PR #5947, opened 2020-03-04
Implement remaining changes in CDC Log format

haaawk:
In https://github.com/scylladb/scylla/pull/5910 @elcallio implemented first part of the new CDC Log format. This PR implements the remaining part. Namely:
1. operation column now contains string representation of operation type instead of meaningless int
2. CDC Log partition key is now a single cdc$stream_id column instead of two columns (cdc$stream_id_1 and cdc$stream_id_2).

Description of the new CDC Log format can be found here: https://docs.google.com/document/d/1riJlkormkzZL0SYqQixxpsDXiURCjG0n_9KJS21zCto/edit

Tests: unit(dev)

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -321,8 +321,7 @@ bytes log_data_column_deleted_elements_name_bytes(const bytes& column_name) {
 static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID> uuid) {
     schema_builder b(s.ks_name(), log_name(s.cf_name()));
     b.set_comment(sprint("CDC log for %s.%s", s.ks_name(), s.cf_name()));
-    b.with_column(log_meta_column_name_bytes("stream_id_1"), long_type, column_kind::partition_key);
-    b.with_column(log_meta_column_name_bytes("stream_id_2"), long_type, column_kind::partition_key);
+    b.with_column(log_meta_column_name_bytes("stream_id"), bytes_type, column_kind::partition_key);
     b.with_column(log_meta_column_name_bytes("time"), timeuuid_type, column_kind::clustering_key);
     b.with_column(log_meta_column_name_bytes("batch_seq_no"), int32_type, column_kind::clustering_key);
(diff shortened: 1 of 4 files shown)


cdc/log.cc
```diff
@@ -63,6 +63,27 @@ using namespace std::chrono_literals;
 
 logging::logger cdc_log("cdc");
 
+sstring cdc::operation_to_text(cdc::operation op) {
+    switch (op) {
```
elcallio: const sstring& / const std::string&? Latter could use string literals.

elcallio: former constants declared before.

elcallio:
Do we really want op to be a string? Seems expensive, just to make cqlsh look better. It will hardly make code reading the cdc log easier - you instead have to keep a secondary table of string constants to match against. Not easier than int constants. 
Can't we just either invent cql::enum or just augment cqlsh if the latter is the problem?

Other than that, I have no objections.

haaawk:
> Do we really want op to be a string? Seems expensive, just to make cqlsh look better. It will hardly make code reading the cdc log easier - you instead have to keep a secondary table of string constants to match against. Not easier than int constants.
> Can't we just either invent cql::enum or just augment cqlsh if the latter is the problem?
> 
> Other than that, I have no objections.

@avikivity the change is on your request. Please comment.

cdc/log.cc
haaawk: Done.

elcallio: or simply std::string_view by value. Guaranteed least code and least overhead (unless gcc is completely useless).

haaawk: Changed to string_view.

haaawk:
I uploaded the non-controversial part of the PR here -> https://github.com/scylladb/scylla/pull/5950

Please review and hopefully merge while I'm checking computed columns for the part 2.

nyh:
@haaawk I don't understand if this PR is still relevant or you split it?

haaawk:
I split this and you seemed to merged the other version so I'm closing this.

PR #5948, opened 2020-03-04
docker: rsyslog configuration fixes

### LINKED ISSUE #5937: nightly docker doesn't work - rsyslog fails
Running our latest docker image, e.g.,
```
docker pull scylladb/scylla-nightly:latest
docker run --name scylla -d -p 8000:8000 scylladb/scylla-nightly:latest --alternator-port=8000 --experimental 1
```
Fails to run Scylla. In the log:
```
2020-03-03 09:11:20,972 INFO spawned: 'rsyslog' with pid 96
2020-03-03 09:11:20,981 INFO exited: rsyslog (exit status 1; not expected)
2020-03-03 09:11:22,986 INFO spawned: 'rsyslog' with pid 97
2020-03-03 09:11:23,005 INFO exited: rsyslog (exit status 1; not expected)
2020-03-03 09:11:26,176 INFO spawned: 'rsyslog' with pid 110
2020-03-03 09:11:26,186 INFO exited: rsyslog (exit status 1; not expected)
2020-03-03 09:11:26,945 INFO gave up: rsyslog entered FATAL state, too many start retries too quickly
```

eliransin:
The introduction of rsyslog had two errors in it.
Both errors are non fatal and the docker still works,
however, the system is left in a wrong state in which
supervisord marks rsyslogd service as failed (after several
failed retry attempts). Another bug in the configuration
causes rsyslog to output an error.

1) An inclusion command from a newer version was used
in rsyslogs main configuration file. This caused to rsyslog
to complain during startup but it didn't do much damage since
rsyslog converts every unrecognised command to a message command.
2) in the supervisord definition of the service, rsyslogd is ran
without the -n option which means it defaults to automatically
switch to the background. Supervisord interpret this as an unexpected
process termination and retries to start the process (unsuccessfully
because rsyslog protects itself from having multiple processes of
itself) and eventually marks it as down although it is fully up and
running.
This commit fixes both configuration problems.

Tests: Build and run docker and validate the errors are gone.
Fixes #5937

### DIFF
--- a/dist/docker/redhat/etc/rsyslog.conf
+++ b/dist/docker/redhat/etc/rsyslog.conf
@@ -13,7 +13,7 @@ global(workDirectory="/var/lib/rsyslog")
 module(load="builtin:omfile" Template="RSYSLOG_TraditionalFileFormat")
 
 # Include all config files in /etc/rsyslog.d/
-include(file="/etc/rsyslog.d/*.conf" mode="optional")
+$IncludeConfig "/etc/rsyslog.d/*.conf"
 
 #### MODULES ####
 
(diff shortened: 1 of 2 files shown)


nyh:
Thanks. Committed to next.

PR #5950, opened 2020-03-04
cdc log: merge stream_id columns into a single column

haaawk:
Previously we had stream_id_1 and stream_id_2 columns
of type long each. They were forming a partition key.

In a new format we want a single stream_id column that
forms a partition key. To be able to still store two
longs, the new column will have type blob and its value
will be concatenated bytes of two longs that
partition key is composed of.

We still want partition key to logically be two longs
because those two values will be used by a custom partitioner
later once we implement it.


elcallio:
lgtm

nyh:
pushed to next.

PR #5952, opened 2020-03-04
CDC+LWT: fix missing CDC entries for successful LWTs

### LINKED ISSUE #5869: cdc: LWT prevents cdc logging
On fresh master, e.g. 4e95b67501c33e3f9ea0b1d2cfbd428a640e494a
Results of successful RMW statements are not stored in CDC log:
```
cqlsh> create keyspace ksp with replication = {'class':'SimpleStrategy', 'replication_factor': 1};
cqlsh> create table ksp.t (pk int, ck int, v int, primary key (pk, ck)) with cdc = {'enabled':'true'};
cqlsh> insert into ksp.t (pk, ck, v) values (0,0,0) if not exists;

 [applied] | pk   | ck   | v
-----------+------+------+------
      True | null | null | null

cqlsh> select * from ksp.t; 

 pk | ck | v
----+----+---
  0 |  0 | 0

(1 rows)
cqlsh> select * from ksp.t_scylla_cdc_log; 

 stream_id_1 | stream_id_2 | time | batch_seq_no | _ck | _pk | _v | operation | ttl
-------------+-------------+------+--------------+-----+-----+----+-----------+-----

(0 rows)
```

jul-stas:
Now, if CDC is enabled, `paxos_response_handler::learn_decision()`
augments the base table mutation. The differences in logic between:
(1) `mutate_internal<std::vector<mutation>>()`
and
(2) `mutate_internal<std::vector<std::tuple<paxos::proposal, schema_ptr, ...>>>()`
make it necessary to separate "CDC mutations" from "base mutation"
and send them, respectively, to (1) and (2).

Test will come in a separate PR, because LWT queries from test suite are not routed to correct shards (IIUC `bounce_to_shard` is not handled in test env).

Fixes #5869

### DIFF
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -1177,10 +1177,47 @@ future<bool> paxos_response_handler::accept_proposal(const paxos::proposal& prop
 future<> paxos_response_handler::learn_decision(paxos::proposal decision, bool allow_hints) {
     tracing::trace(tr_state, "learn_decision: committing {} with cl={}", decision, _cl_for_learn);
     paxos::paxos_state::logger.trace("CAS[{}] learn_decision: committing {} with cl={}", _id, decision, _cl_for_learn);
-    std::array<std::tuple<paxos::proposal, schema_ptr, dht::token>, 1> m{std::make_tuple(std::move(decision), _schema, _key.token())};
     // FIXME: allow_hints is ignored. Consider if we should follow it and remove if not.
     // Right now we do not store hints for when committing decisions.
-    return _proxy->mutate_internal(std::move(m), _cl_for_learn, false, tr_state, _permit, _timeout);
+
+    // `mutate_internal` behaves differently when its template parameter is a range of mutations and when it's
+    // a range of (decision, schema, token)-tuples. Both code paths diverge on `create_write_response_handler`.
+    // We use the first path for CDC mutations (if present) and the latter for "paxos mutations".
+    // Attempts to send both kinds of mutations in one shot caused an infinite loop.
+    future<> f_cdc = make_ready_future<>();
+    if (_schema->cdc_options().enabled()) {
+        auto update_mut = decision.update.unfreeze(_schema);
+        const auto base_tbl_id = update_mut.column_family_id();
+        std::vector<mutation> update_mut_vec{std::move(update_mut)};
+
+        if (_proxy->get_cdc_service()->needs_cdc_augmentation(update_mut_vec)) {
+            f_cdc = _proxy->get_cdc_service()->augment_mutation_call(_timeout, std::move(update_mut_vec))
+                    .then([this, base_tbl_id] (std::tuple<std::vector<mutation>, cdc::result_callback>&& t) {
+                auto mutations = std::move(std::get<0>(t));
+                auto end_func = std::move(std::get<1>(t));
+
+                // Pick only the CDC ("augmenting") mutations
+                mutations.erase(std::remove_if(mutations.begin(), mutations.end(), [base_tbl_id = std::move(base_tbl_id)] (const mutation& v) {
+                    return v.schema()->id() == base_tbl_id;
+                }), mutations.end());
+                if (mutations.empty()) {
+                    return make_ready_future<>();
+                }
+
+                auto f = _proxy->mutate_internal(std::move(mutations), _cl_for_learn, false, tr_state, _permit, _timeout);
+                if (end_func) {
+                    f = f.then(std::move(end_func));
+                }
+                return f;
+            });
+        }
+    }
+
+    // Path for the "base" mutations
+    std::array<std::tuple<paxos::proposal, schema_ptr, dht::token>, 1> m{std::make_tuple(std::move(decision), _schema, _key.token())};
+    future<> f_lwt = _proxy->mutate_internal(std::move(m), _cl_for_learn, false, tr_state, _permit, _timeout);
+
+    return when_all_succeed(std::move(f_cdc), std::move(f_lwt));
 }
 
 static std::vector<gms::inet_address>


service/storage_proxy.cc
```diff
@@ -1177,10 +1177,40 @@ future<bool> paxos_response_handler::accept_proposal(const paxos::proposal& prop
 future<> paxos_response_handler::learn_decision(paxos::proposal decision, bool allow_hints) {
     tracing::trace(tr_state, "learn_decision: committing {} with cl={}", decision, _cl_for_learn);
     paxos::paxos_state::logger.trace("CAS[{}] learn_decision: committing {} with cl={}", _id, decision, _cl_for_learn);
-    std::array<std::tuple<paxos::proposal, schema_ptr, dht::token>, 1> m{std::make_tuple(std::move(decision), _schema, _key.token())};
     // FIXME: allow_hints is ignored. Consider if we should follow it and remove if not.
     // Right now we do not store hints for when committing decisions.
-    return _proxy->mutate_internal(std::move(m), _cl_for_learn, false, tr_state, _permit, _timeout);
+    auto update_mut = decision.update.unfreeze(_schema);
+    std::vector update_mut_vec = {update_mut};
```
haaawk: This copies update_mut, no? Maybe it's better to create vector and then push_back with std::move?
That would require capturing base_tbl_id before moving from update_mut.

jul-stas: done

service/storage_proxy.cc
```diff
@@ -1177,10 +1177,42 @@ future<bool> paxos_response_handler::accept_proposal(const paxos::proposal& prop
 future<> paxos_response_handler::learn_decision(paxos::proposal decision, bool allow_hints) {
     tracing::trace(tr_state, "learn_decision: committing {} with cl={}", decision, _cl_for_learn);
     paxos::paxos_state::logger.trace("CAS[{}] learn_decision: committing {} with cl={}", _id, decision, _cl_for_learn);
-    std::array<std::tuple<paxos::proposal, schema_ptr, dht::token>, 1> m{std::make_tuple(std::move(decision), _schema, _key.token())};
     // FIXME: allow_hints is ignored. Consider if we should follow it and remove if not.
     // Right now we do not store hints for when committing decisions.
-    return _proxy->mutate_internal(std::move(m), _cl_for_learn, false, tr_state, _permit, _timeout);
+    auto update_mut = decision.update.unfreeze(_schema);
```
gleb-cloudius: Why to unfreeze even if cdc is disabled?

haaawk: Fair point. @jul-stas please use schema to check whether cdc is enabled on the table.

jul-stas: done

service/storage_proxy.cc
```diff
@@ -1177,10 +1177,44 @@ future<bool> paxos_response_handler::accept_proposal(const paxos::proposal& prop
...
-    return _proxy->mutate_internal(std::move(m), _cl_for_learn, false, tr_state, _permit, _timeout);
+    future<> f_cdc = make_ready_future<>();
+    if (_schema->cdc_options().enabled()) {
+        auto update_mut = decision.update.unfreeze(_schema);
+        const auto base_tbl_id = update_mut.column_family_id();
+        std::vector<mutation> update_mut_vec;
+        update_mut_vec.push_back(std::move(update_mut));
+
+        if (_proxy->get_cdc_service() && _proxy->get_cdc_service()->needs_cdc_augmentation(update_mut_vec)) {
+            // CDC mutations are sent down another route
```
kostja: I have hard time figuring out from this comment what's going on here. AFAICS there are no CDC mutations yet, they are created by augment_mutation_call. Then they are erased for whatever reason (why erase the augmentations, isn't it easier to move out the original mutation?). Then the code duplicates mutate_internal branch, thus any time we will need to extend this logic, we'll need to do it twice.

service/storage_proxy.cc
```diff
@@ -1177,10 +1177,44 @@ future<bool> paxos_response_handler::accept_proposal(const paxos::proposal& prop
...
+                auto f = _proxy->mutate_internal(std::move(mutations), _cl_for_learn, false, tr_state, _permit, _timeout);
+                if (end_func) {
+                    f = f.then(std::move(end_func));
+                }
+                return f;
+            });
+        }
+    }
+
+    // Path for the "base" mutations
```
nyh: Can this part of the code - the "base" (normal) write - be moved before the special CDC code. I think it will be clearer, and in any case, both operations happen in parallel. I think it would make the code clearer - you see the normal code path first, then the special cdc path. 
UPDATE: this suggestion is an alternative to the other idea, of not having separate with-cdc and without-cdc paths - both will generate a vector of mutations with a single mutate_internal() call at the end.

service/storage_proxy.cc
haaawk: "Then they are erased for whatever reason (why erase the augmentations, isn't it easier to move out the original mutation?)"
augment_mutation_call takes vector of mutations and returns vector of mutations that includes all initial mutations + cdc log mutations for some of those initial mutation. The one that have CDC enabled. Initial mutations are removed because they are paxos mutations and have to be handled differently.

2. "Then the code duplicates mutate_internal branch"
No it does not. Those two calls are different mutate_internal overloads. It's LWT magic so I feel uncomfortable explaining it to you :)

service/storage_proxy.cc
```diff
@@ -1177,10 +1177,44 @@ future<bool> paxos_response_handler::accept_proposal(const paxos::proposal& prop
...
-    std::array<std::tuple<paxos::proposal, schema_ptr, dht::token>, 1> m{std::make_tuple(std::move(decision), _schema, _key.token())};
     // FIXME: allow_hints is ignored. Consider if we should follow it and remove if not.
     // Right now we do not store hints for when committing decisions.
-    return _proxy->mutate_internal(std::move(m), _cl_for_learn, false, tr_state, _permit, _timeout);
+    future<> f_cdc = make_ready_future<>();
+    if (_schema->cdc_options().enabled()) {
+        auto update_mut = decision.update.unfreeze(_schema);
+        const auto base_tbl_id = update_mut.column_family_id();
+        std::vector<mutation> update_mut_vec;
+        update_mut_vec.push_back(std::move(update_mut));
```
nyh: nitpick: I think (?) you can do these two lines in one, and potentially improve std::vector's guess on how much memory to allocate, with
std::vector<mutation> update_mut_vec{std::move(update_mut)}

service/storage_proxy.cc
```diff
@@ -1177,10 +1177,44 @@ future<bool> paxos_response_handler::accept_proposal(const paxos::proposal& prop
...
+
+        if (_proxy->get_cdc_service() && _proxy->get_cdc_service()->needs_cdc_augmentation(update_mut_vec)) {
```
nyh: You can move the _proxy->get_cdc_service() check earlier (with enabled()), there is no point in unfreezing the mutation if there is none. By the way, is it even possible that cdc_options().enabled() but !get_cdc_service()?

service/storage_proxy.cc
nyh: I agree it's weird. We could have both code paths (with cdc and without cdc) generate a vector of mutations, and then have just one place where mutate_internal() is called for this order.
Alternatively, if for clarity of this code you switch around the order (this is the option I suggested below) - first write the base write, and only then optionally write the cdc one, then maybe there is some logic in this code (which creates a vector and then removes the original cdc element), because it makes the non-cdc case clearer, with the downside of making the cdc-specific code uglier.

haaawk: using single mutate_internal won't work because those are two different mutate_internal overloads. Could you elaborate a bit more @nyh how you would like to merge those two overloads? The overloads have different logic not just different parameters. overload that takes paxos decision is special.

service/storage_proxy.cc
haaawk: It's not possible so let's drop check for get_cdc_service().

service/storage_proxy.cc
haaawk: Good point - let's do it @jul-stas

service/storage_proxy.cc
haaawk: I don't think it's possible to reorder the cases because execution of the lwt logic will destroy the mutation that we need to execute cdc logic. We could copy the mutation before executing lwt and use it later in cdc but then we would have:
1. cdc
2. lwt
3. cdc

which in my opinion is worse. What do you think @nyh ?

service/storage_proxy.cc
```diff
@@ -1177,10 +1177,44 @@ future<bool> paxos_response_handler::accept_proposal(const paxos::proposal& prop
...
+                // Pick only the CDC ("augmenting") mutations
+                mutations.erase(std::remove_if(mutations.begin(), mutations.end(), [base_tbl_id = std::move(base_tbl_id)] (const mutation& v) {
+                    return v.schema()->id() == base_tbl_id;
+                }), mutations.end());
+                if (mutations.empty()) {
+                    return make_ready_future<>();
+                }
+
+                auto f = _proxy->mutate_internal(std::move(mutations), _cl_for_learn, false, tr_state, _permit, _timeout);
+                if (end_func) {
```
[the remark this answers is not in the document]
elcallio: Requires knowing that augment only appends to mutation vector, but...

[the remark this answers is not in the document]
haaawk: I thought about it but this is less maintainable. Current code is more bullet proof. I would keep it the way it is.

service/storage_proxy.cc
kostja: Where do CDC mutations go?

kostja: Sorry, you're right, the branch is not duplicated, overload of mutate_internal() got me confused.

service/storage_proxy.cc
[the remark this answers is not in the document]
kostja: What is end_func?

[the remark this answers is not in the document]
kostja: haawk, if augment_mutation_vec takes two vectors, input and output, it can be both maintainable and avoid unnecessary work.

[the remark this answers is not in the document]
haaawk: Then it will be less convenient in the main CDC uses. This extra work here is std::move of mutation. I wouldn't go crazy because of it :).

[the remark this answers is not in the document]
haaawk: > What is end_func?

It's a hook to provide postimage when we implement it. @elcallio can elaborate if you need it.

service/storage_proxy.cc
jul-stas: done

service/storage_proxy.cc
jul-stas: done

service/storage_proxy.cc
jul-stas: I added a comment WRT that into the code.

service/storage_proxy.cc
nyh: The "1. cdc" part will just be an unfreeze() call, but you're right, that isn't very pretty either. I guess we can think about this later, and this version is acceptable. I'll commit.

haaawk: Thanks!

PR #5954, opened 2020-03-04
CDC: use a single `cdc$time` value for a batch of changes

kbr-:
If a batch update is performed with a sequence of changes with a single timestamp, they will now show up in CDC with a single timeuuid in the `cdc$time` column, distinguished by different `cdc$batch_seq_no` values.

Fixes https://github.com/scylladb/scylla/issues/5953.

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -938,9 +938,11 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, 
                 auto& m = mutations[idx];
                 auto& s = m.schema();
                 if (should_split(m, *s)) {
-                    for (auto&& mm : split(m, s)) {
-                        mutations.push_back(trans.transform(mm, rs.get()));
-                    }
+                    std::vector<mutation> res;
+                    for_each_change(m, s, [&] (mutation mm) {
+                        res.push_back(trans.transform(std::move(mm), rs.get()));
+                    });
+                    mutations.insert(mutations.end(), std::make_move_iterator(res.begin()), std::make_move_iterator(res.end()));
                 } else {
                     mutations.push_back(trans.transform(m, rs.get()));
                 }
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -531,20 +531,18 @@ public:
 
     // TODO: is pre-image data based on query enough. We only have actual column data. Do we need
     // more details like tombstones/ttl? Probably not but keep in mind.
-    mutation transform(const mutation& m, const cql3::untyped_result_set* rs) const {
-        auto ts = find_timestamp(*_schema, m);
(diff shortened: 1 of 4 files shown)


cdc/log.cc
```diff
@@ -938,9 +938,11 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
                 auto& m = mutations[idx];
                 auto& s = m.schema();
                 if (should_split(m, *s)) {
-                    for (auto&& mm : split(m, s)) {
-                        mutations.push_back(trans.transform(mm, rs.get()));
-                    }
+                    std::vector<mutation> res;
```
haaawk: Why can't we just push back straight to the mutations?

haaawk: Is it because insertion to mutations can invalidate m?

cdc/split.cc
```diff
@@ -375,13 +375,12 @@ bool should_split(const mutation& base_mutation, const schema& base_schema) {
     return false;
 }
 
-std::vector<mutation> split(const mutation& base_mutation, const schema_ptr& base_schema) {
+void for_each_change(const mutation& base_mutation, const schema_ptr& base_schema, std::function<void(mutation)> f) {
```
haaawk: Could we use noncopyable_function here to avoid allocation?

cdc/log.cc
haaawk: It seems that base_mutation is only used in for_each_change before the call to f. If you store partition key from base_mutation in a variable you should be able to insert mm straight to mutations.

test/boost/cdc_test.cc
```diff
@@ -1126,6 +1126,13 @@ SEASTAR_THREAD_TEST_CASE(test_change_splitting) {
             }
         }
 
+        {
+            auto result = get_result({int32_type},
+                "select \"cdc$batch_seq_no\" from ks.t_scylla_cdc_log where pk = 0 and ck = 1 allow filtering");
+            std::vector<std::vector<data_value>> expected = {{int32_t(0)}, {int32_t(1)}, {int32_t(2)}, {int32_t(3)}};
+            BOOST_REQUIRE_EQUAL(expected, result);
+        }
+
```
piodul: Can we convert it to just `BOOST_REQUIRE_EQUAL`, just like in line 1181?

test/boost/cdc_test.cc
```diff
@@ -1126,6 +1126,13 @@ SEASTAR_THREAD_TEST_CASE(test_change_splitting) {
             }
```
piodul: Can we convert it to just `BOOST_REQUIRE_EQUAL`, just like in line 1181?

haaawk: I checked and we can't. The order is still arbitrary.

haaawk:
I rebased, resolved conflicts, addressed all the comments and made the PR bisectable -> https://github.com/scylladb/scylla/pull/5961

PR #5955, opened 2020-03-04
storage_service: boost shares for streaming write during bootstrap, r…

glommer:
…eplace

During bootstrap and replace operations the node can't take reads and
we'd like to see the process ending ASAP. This is because until the
process ends, we keep having to duplicate writes to an extended set. Not
to mention, in the case of a cluster expansion users want to use the
added capacity sooner rather than later.

Streaming generates a lot of compaction activity, that competes with the
bootstrap itself, slowing it down.

Long term, we are moving to treat those compactions differently and
maybe postpone them altogether. However in the meantime we can play with
the shares so as to allow streaming writes to proceed with equal
footing.

Streaming has the lowest shares of all, which makes sense when it is a
maintenance operation. But in the case of a bootstrap or replace
streaming is the star of the show.

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -81,6 +81,7 @@
 #include <seastar/core/metrics.hh>
 #include "cdc/generation.hh"
 #include "repair/repair.hh"
+#include "service/priority_manager.hh"
 
 using token = dht::token;
 using UUID = utils::UUID;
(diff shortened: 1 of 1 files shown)


asias:
The bootstrap node does not take reads but it takes writes. After "replacing node takes write " series, the replacing node will take writes as well.  Boost shares of streaming from 200 to 1000, it will slow down the pending writes, which in turn will slow down all the writes in the cluster, because the writes returns to user after the pending writes finishes.

I proposed to lower compaction shares and boost streaming shares, but streaming shares must be lower than the cql workload. 

I think we need to modify the CPU shares as well.

glommer:
I am retiring this PR. @asias is moving in an interesting direction with increasing min_threshold temporarily and we shouldn't do both

PR #5980, opened 2020-03-08
cdc: on row delete, show the whole row as preimage

### LINKED ISSUE #5709: CDC: row deletes are handled incorrectly

haaawk:
If base mutation has at least one row tombstone, its preimage log entry displays all the base columns.

Fixes #5709

Tests: unit(dev)

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -711,26 +711,7 @@ public:
                     }
 
                     if (has_pirow) {
-                        value = cdef.is_atomic()
-                            ? pirow->get_blob(cdef.name_as_text())
-                            : visit(*cdef.type, make_visitor(
-                                // flatten set
-                                [&] (const set_type_impl& type) {
-                                    auto v = pirow->get_view(cdef.name_as_text());
-                                    auto f = cql_serialization_format::internal();
-                                    auto n = read_collection_size(v, f);
-                                    std::vector<bytes_view> tmp;
-                                    tmp.reserve(n);
-                                    while (n--) {
-                                        tmp.emplace_back(read_collection_value(v, f)); // key
-                                        read_collection_value(v, f); // value. ignore.
-                                    }
-                                    return set_type_impl::serialize_partially_deserialized_form(tmp, f);
-                                },
-                                [&] (const abstract_type& o) -> bytes {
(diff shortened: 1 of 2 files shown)


haaawk:
This is https://github.com/scylladb/scylla/pull/5976 with fixed review comments.

PR #5982, opened 2020-03-08
api: /column_family/major_compaction/{keyspace:table} implementation

sitano:
Signed-off-by: Ivan Prisyazhnyy <ivan@scylladb.com>

Manual QA
===

```
$ ninja
$ env SCYLLA_HOME=$(pwd) ./build/dev/scylla --log-to-syslog 0 --log-to-stdout 1 --default-log-level info --network-stack posix --developer-mode=1 --cpuset 0 --listen-address 127.0.0.1 --rpc-address 127.0.0.1 --seed-provider-parameters seeds=127.0.0.1 --blocked-reactor-notify-ms 999999999 --workdir $SCYLLA_HOME/build/data

CREATE KEYSPACE kr1 WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1} AND durable_writes = false;
USE kr1;
CREATE TABLE standard1 (id int PRIMARY KEY);

$ curl -X POST http://127.0.0.1:10000/column_family/major_compaction/kr1:standard1
```

### DIFF
--- a/api/api-doc/column_family.json
+++ b/api/api-doc/column_family.json
@@ -70,7 +70,7 @@
             {
                "method":"POST",
                "summary":"Force a major compaction of this column family",
-               "type":"string",
+               "type":"void",
                "nickname":"force_major_compaction",
                "produces":[
                   "application/json"
--- a/api/column_family.cc
+++ b/api/column_family.cc
@@ -994,5 +994,15 @@ void set_column_family(http_context& ctx, routes& r) {
         });
     });
 
+    cf::force_major_compaction.set(r, [&ctx](std::unique_ptr<request> req) {
+        if (req->get_query_param("split_output") != "") {
+            fail(unimplemented::cause::API);
+        }
+        return foreach_column_family(ctx, req->param["name"], [](column_family &cf) {
+            return cf.compact_all_sstables();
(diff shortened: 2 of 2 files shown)


sitano:
Speaking of the implementation:
- I don't know what was meant by `split_output` so I left it for the future
- I don't know what POST to this method was expected to return, so I changed it to void

penberg:
Looks good to me, but please add a dtest for this. @glommer, the "split output" option seems to be specific to STCS, but we don't really support it, right? IOW, we should just disallow that for now like this pull request does.

sitano:
@penberg thanks! I will add dtest for it. You are right about the split output - we don't support it in STCS. Yes it is related to the STCS. It is disallowed by throwing an exception (or I can just ignore it).
btw, shall I add dtest for the similar calls to the disable minor compactions PR?

glommer:
We don't support splits, nor do I think we should.

There is a command (with an open issue) that we don't implement to compact just some SStables together. That one I think is worth implementing, the split thing is not

penberg:
Merged to next as https://github.com/scylladb/scylla/commit/5ec7e77b2eb7c21ab64e6340a0007f4091b32591

PR #5991, opened 2020-03-09
cdc: wait for all seeds' tokens when joining the ring without bootstrapping

kbr-:
It's a common scenario in tests where a new cluster is created by starting
multiple nodes as seeds concurrently, even though the recommended way to
start nodes in production is sequential (and only the first node should be joined as seed).

When a node joins a Scylla cluster for the first time (i.e. it joins the
token ring), it proposes a new CDC generation which is supposed to
include tokens of all nodes: all nodes that were already present plus
the new node. The new node learns the tokens of other nodes by
contacting a seed node (other than itself, if it's also a seed).

However, in the concurrent startup scenario, the above strategy doesn't really
work. Example: if there are 3 nodes joining the ring concurrently, then the
second and third might not see each other's tokens when proposing their
CDC generations, so we're left with some generation which doesn't
include all tokens.

Fortunately, (most of) these tests use a certain pattern of specifying
seed lists:
- node 1 specifies {1}
- node 2 specifies {1, 2}
- node 3 specifies {1, 2, 3}

etc.
Exploiting that fact, we can prevent the race by forcing node N to wait
until it sees the tokens of nodes {1, ..., N - 1} (which are all seeds).

The first commit does that. Obviously, this is a heuristic, but it's all we
can have before implementing a real solution, which is transactional
joining of nodes.

But there's another problem. It's possible that the generation proposed by the last node, which saw all other nodes, will not have the greatest timestamp due to desynchronized clocks. The second commit attempts to prevent this as well. Again, this is a temporary heuristic.

Refs: https://github.com/scylladb/scylla/issues/5812. It doesn't truly fix the issue, for that we'll need to finish the CDC topology change support roadmap (i.e. implement transactional node joining), but it'll work most of the time.

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -471,29 +471,33 @@ void storage_service::maybe_start_sys_dist_ks() {
     _sys_dist_ks.invoke_on_all(&db::system_distributed_keyspace::start).get();
 }
 
-/* If we're not a seed node or there are seeds other than us,
- * wait until we contact any of them and learn what tokens are used by other nodes in the cluster,
- * with a timeout.
+/* Wait until we retrieve the tokens of all seeds in the cluster, with a timeout.
  *
- * Returns true if we're the only seed or we've managed to learn the other nodes' tokens before timeouting.
+ * Returns true if we've managed to learn all seeds' tokens before timeouting.
  */
 // Run in seastar::async context.
-static bool wait_for_other_tokens(const locator::token_metadata& tm,
-        const std::set<inet_address>& seeds, inet_address my_ep, abort_source& as) {
-    if (seeds.size() == 1 && seeds.count(my_ep)) {
-        return true;
-    }
+static bool wait_for_seed_tokens(const locator::token_metadata& tm,
+        const std::set<inet_address>& seeds, abort_source& as) {
+    std::unordered_set<inet_address> known_endpoints;
+    known_endpoints.reserve(tm.get_token_to_endpoint().size());
 
-    int retries = 30;
+    int retries = 60;
     while (retries--) {
-        auto& tok_to_ep = tm.get_token_to_endpoint();
-        if (std::any_of(tok_to_ep.begin(), tok_to_ep.end(),
-                [my_ep] (const std::pair<token, inet_address>& p) { return p.second != my_ep; })) {
+        for (const auto& [t, ep]: tm.get_token_to_endpoint()) {
+            known_endpoints.insert(ep);
+            thread::maybe_yield();
+        }
+
+        if (std::all_of(seeds.begin(), seeds.end(), [&known_endpoints] (const inet_address& s) {
+            return known_endpoints.count(s);
+        })) {
             return true;
         }
 
-        slogger.warn("Couldn't retrieve other nodes' tokens. Retrying again in one second...");
+        slogger.warn("Couldn't retrieve tokens of all seeds. Retrying again in one second...");
         sleep_abortable(std::chrono::seconds(1), as).get();
+
+        known_endpoints.clear();
     }
 
     return false;
@@ -681,7 +685,9 @@ void storage_service::join_token_ring(int delay) {
             // We might have not due to misconfiguration, e.g. skipping the "wait for gossip to settle" phase,
             // or due to a network partition which didn't allow us to contact other seeds.
             // But CDC requires that we know other nodes' tokens - they are needed to make a new generation of streams.
-            if (!wait_for_other_tokens(_token_metadata, _gossiper.get_seeds(), get_broadcast_address(), _abort_source)) {
+            // This is a heuristic: it assumes that if we managed to retrieve the tokens of all seeds,
+            // then we also managed to retrieve all other nodes' tokens.
+            if (!wait_for_seed_tokens(_token_metadata, _gossiper.get_seeds(), _abort_source)) {
                 throw std::runtime_error(format(
                     "Can't boot this node because we weren't able to retrieve other nodes' tokens,"
                     " which is required for CDC to work. Make sure that:\n"
--- a/cdc/generation.cc
+++ b/cdc/generation.cc
@@ -327,6 +327,18 @@ db_clock::time_point make_new_cdc_generation(
     auto ts = db_clock::now() + (
             for_testing ? std::chrono::milliseconds(0) : (
                 2 * ring_delay + std::chrono::duration_cast<std::chrono::milliseconds>(generation_leeway)));
+
+    // If we notice that another node has proposed a new generation concurrently,
+    // make sure that the timestamp of our generation is greater.
+    // Our generation is most likely at least as good as theirs because we've observed their proposition.
+    for (const auto& ep: g.get_endpoint_states()) {
+        auto other_ts = get_streams_timestamp_for(ep.first, g);
+        if (other_ts && *other_ts >= ts) {
+            cdc_log.warn("CDC generations proposed concurrently: {} and {}", ts, *other_ts);
+            ts = *other_ts + std::chrono::seconds(1);
+        }
+    }
+
     sys_dist_ks.insert_cdc_topology_description(ts, std::move(gen), { tm.count_normal_token_owners() }).get();
 
     return ts;


service/storage_service.cc
```diff
@@ -471,28 +471,24 @@ void storage_service::maybe_start_sys_dist_ks() {
...
-static bool wait_for_other_tokens(const locator::token_metadata& tm,
-        const std::set<inet_address>& seeds, inet_address my_ep, abort_source& as) {
-    if (seeds.size() == 1 && seeds.count(my_ep)) {
-        return true;
-    }
-
+static bool wait_for_seed_tokens(const locator::token_metadata& tm,
+        const std::set<inet_address>& seeds, abort_source& as) {
     int retries = 30;
     while (retries--) {
```
haaawk: Shouldn't we break this while if about_source says so?

piodul: `sleep_abortable` will throw a `sleep_aborted` exception if that happens.

haaawk: ok. thanks @piodul

service/storage_service.cc
```diff
@@ -471,28 +471,24 @@ void storage_service::maybe_start_sys_dist_ks() {
...
-        auto& tok_to_ep = tm.get_token_to_endpoint();
-        if (std::any_of(tok_to_ep.begin(), tok_to_ep.end(),
-                [my_ep] (const std::pair<token, inet_address>& p) { return p.second != my_ep; })) {
+        if (std::all_of(seeds.begin(), seeds.end(), [&tok_to_ep = tm.get_token_to_endpoint()] (const inet_address& s) {
```
haaawk: Isn't this quadratic? I guess the number of seed nodes shouldn't be big but maybe it's better to sort both collections and do the check linearly on them.

asias:
```
Fortunately, (most of) these tests use a certain pattern of specifying
seed lists:

    node 1 specifies {1}
    node 2 specifies {1, 2}
    node 3 specifies {1, 2, 3}

etc.
Exploiting that fact, we can prevent the race by forcing node N to wait
until it sees the tokens of nodes {1, ..., N - 1} (which are all seeds).
```

It is sad we want to modify scylla to workaround bad  practice in dtest. We should stop using the the pattern in dtest.  This is not what user use in practice. I know it is using this to utilize the fact seed node does not bootstrap to make the test fast, with make gossip faster and robust series we do not need this hack.  We are also in the direction to move away from the seed concept. 

Does changing the cdc dtest help?

asias:
A rare case, but it might happen. if we have a node configured with {seed1, seed2}, seed2 is not up yet. The node will fail to boot (it gives up retrying of 30 times in this patch).  In dtest, node 2 specifies {1, 2} as seed might start earlier than node1, so this do happen.

service/storage_service.cc
```diff
@@ -471,28 +471,24 @@ void storage_service::maybe_start_sys_dist_ks() {
...
+            return std::any_of(tok_to_ep.begin(), tok_to_ep.end(), [&s] (const std::pair<token, inet_address>& p) {
```
asias: This is O(N), N = number of tokens in the cluster. This could cause stall. This is under seastar::async context, we could call seastar::thread::maybe_yield() in side the check lambda.

service/storage_service.cc
```diff
@@ -681,7 +677,9 @@ void storage_service::join_token_ring(int delay) {
             // We might have not due to misconfiguration, e.g. skipping the "wait for gossip to settle" phase,
             // or due to a network partition which didn't allow us to contact other seeds.
             // But CDC requires that we know other nodes' tokens - they are needed to make a new generation of streams.
-            if (!wait_for_other_tokens(_token_metadata, _gossiper.get_seeds(), get_broadcast_address(), _abort_source)) {
+            // This is a heuristic: it assumes that if we managed to retrieve the tokens of all seeds,
+            // then we also managed to retrieve all other nodes' tokens.
+            if (!wait_for_seed_tokens(_token_metadata, _gossiper.get_seeds(), _abort_source)) {
```
asias: The wait_for_seed_tokens check is in the branch `!db::system_keyspace::bootstrap_complete()`, this means this is a bootstrapping node or a replacing node. The node should have learned the tokens before bootstrap or replacing, otherwise it will stream the wrong data.  We call `_gossiper.wait_for_range_setup().get()`  before and after the bootstrap node announces its token. We do similar thing for replace operation in the making replace node take writes [series](https://github.com/scylladb/scylla/pull/5731. ).

service/storage_service.cc
kbr-: I assumed that `seeds` is of constant size in practice. At worst its size can be equal to the total number of nodes, and the biggest Scylla clusters out there have ~100 nodes so it won't be a problem.

haaawk: There are Cassandra deployments with 100k nodes so I don't think it's wise to assume there will always be a small number here.

kbr-:
> A rare case, but it might happen. if we have a node configured with {seed1, seed2}, seed2 is not up yet. The node will fail to boot (it gives up retrying of 30 times in this patch). In dtest, node 2 specifies {1, 2} as seed might start earlier than node1, so this do happen.

CCM (which is used by dtests) starts node 1 first, then it waits for ~2 seconds, then it starts node 2. So when node 2 reaches this check, node 1 should be far into the startup procedure. I think it's safe to assume that the failure won't happen, but I can increase the number of retries just in case.

service/storage_service.cc
kbr-: You think one would specify every node as seed?
But okay, it's not a problem, I'll make a `set` out of `seeds`.

haaawk: I don't think that but I think that if some number is beyond our control it's good to be defensive.

kbr-: No wait, this doesn't make sense. I'll have to make a `set` of `inet_address`es containing the nodes.

service/storage_service.cc
kbr-: No, it's not bootstrapping, that's the problem. It's skipping the streaming phase and joining the token ring immediately (it's either a seed or has auto_bootstrap=off). Otherise `_cdc_streams_ts` wouldn't be `nullopt`.

The bootstrapping case is handled in `bootstrap()`. There, `make_new_cdc_generation` is called after `wait_for_range_setup()`, as you said.

kbr-:
v2:
- increase the number of retries to 60 just in case
- copy the set of endpoints for which we know the tokens to an `unordered_set`, so the seeds check has (average) linear complexity instead of quadratic
- add a `thread::maybe_yield()` when iterating over `tm.get_token_to_endpoint()`

service/storage_service.cc
```diff
@@ -471,29 +471,33 @@ void storage_service::maybe_start_sys_dist_ks() {
...
+    std::unordered_set<inet_address> known_endpoints;
+    known_endpoints.reserve(tm.get_token_to_endpoint().size());
 
-    int retries = 30;
+    int retries = 60;
     while (retries--) {
-        auto& tok_to_ep = tm.get_token_to_endpoint();
-        if (std::any_of(tok_to_ep.begin(), tok_to_ep.end(),
-                [my_ep] (const std::pair<token, inet_address>& p) { return p.second != my_ep; })) {
+        for (const auto& [t, ep]: tm.get_token_to_endpoint()) {
```
haaawk: Why not known_endpoints(tm.get_token_to_endpoint().begin(), tm.get_token_to_endpoint().end())?

kbr-: `thread::maybe_yield()`

service/storage_service.cc
```diff
@@ -471,29 +471,33 @@ void storage_service::maybe_start_sys_dist_ks() {
...
+            known_endpoints.insert(ep);
+            thread::maybe_yield();
```
haaawk: I would just do known_endpoints(tm.get_token_to_endpoint().begin(), tm.get_token_to_endpoint().end()) here. It's just linear operation.

service/storage_service.cc
```diff
@@ -471,29 +471,33 @@ void storage_service::maybe_start_sys_dist_ks() {
...
+        }
+
+        if (std::all_of(seeds.begin(), seeds.end(), [&known_endpoints] (const inet_address& s) {
```
haaawk: Seems like std::includes

service/storage_service.cc
haaawk: Also it's probably more efficient than adding one by one.

kbr-: `known_endpoints(tm.get_token_to_endpoint().begin(), tm.get_token_to_endpoint().end())` is adding one by one.

@asias suggested in other comment:
> This is O(N), N = number of tokens in the cluster. This could cause stall. This is under seastar::async context, we could call seastar::thread::maybe_yield() in side the check lambda.

and even if in practice it won't stall, I don't see a problem with being defensive (as you suggested earlier) and including `maybe_yield()`.

service/storage_service.cc
```diff
@@ -471,29 +471,33 @@ void storage_service::maybe_start_sys_dist_ks() {
...
 // Run in seastar::async context.
-static bool wait_for_other_tokens(const locator::token_metadata& tm,
-        const std::set<inet_address>& seeds, inet_address my_ep, abort_source& as) {
-    if (seeds.size() == 1 && seeds.count(my_ep)) {
-        return true;
-    }
+static bool wait_for_seed_tokens(const locator::token_metadata& tm,
+        const std::set<inet_address>& seeds, abort_source& as) {
+    std::unordered_set<inet_address> known_endpoints;
+    known_endpoints.reserve(tm.get_token_to_endpoint().size());
```
haaawk: Wouldn't it be better to call tm.get_leaving_endpoints() ?

service/storage_service.cc
kbr-: Not really

service/storage_service.cc
kbr-: `get_leaving_endpoints()` is completely unrelated to what we're doing here.

service/storage_service.cc
haaawk: Fair enough.

service/storage_service.cc
haaawk: Sorry, I misread it and thought they are living_endpoinds :D. In general, I was looking for a way to get the set/list of endpoints not a map from token to endpoint.

kbr-: Living endpoints = endpoints which we recently contacted

but here, we don't care whether a node is alive or not. We care whether it's tokens are in the token ring.

haaawk: Sure. That's why I was asking. Question about living endpoints isn't as stupid as about leaving endpoints :D

service/storage_service.cc
asias: Yes, this node can skip streaming (be a seed or auto_bootstrap = false), but it does not change the fact it is a bootstrapping node or replacing node, it is a new node joins the cluster because only new node will have bootstrap_complete() == false.

New node know should let other nodes know its token and learn other nodes token before it joins the cluster.

asias:
> CCM (which is used by dtests) starts node 1 first, then it waits for ~2 seconds, then it starts node 2. So when node 2 reaches this check, node 1 should be far into the startup procedure. I think it's safe to assume that the failure won't happen, but I can increase the number of retries just in case.

Again, we should fix ccm/ccm. It is using the wrong patten. 

1) It violates the rule we can not bootstrap a seed node into the cluster. 
2) we should always use auto_bootstrap = true, even remove the option.

We can fix scylla to get rid of the unnecessary sleeps during bootstrap so the boot speed if fast. This way, non fist node in the cluster will always bootstrap. We will not have the problem at all with this. And we do not need this hack for CDC in scylla. 

We should prevent the wrong things to happen instead of workaround it in scylla.

@slivne @avikivity

service/storage_service.cc
kbr-: > but it does not change the fact it is a bootstrapping node

Then I'm just using a different nomenclature. I was convinced that
- joining with streaming is called "bootstrapping"
- joining without streaming is not bootstrapping

Some parts of the code suggest that naming. Streaming is done inside the `bootstrap()` function, and if we join without streaming we skip the `bootstrap()` function.
Also, the "auto_bootstrap" flag suggest that bootstrapping = joining with streaming, because if we set it to `off`, the node will join without streaming.
But then there's the `bootstrap_complete()` function which says whether we have joined, not whether we have joined with streaming. IMO it should be called `joining_complete()`.
Inconsistent naming.

> New node know should let other nodes know its token

Maybe, but until now it wouldn't do that in the joining-without-streaming case. I'm not trying to change that in this PR, I'm just trying to make CDC work.

But I don't want to add another 30 second wait to the code. Also, relying on real time is not a correct solution anyway.

kbr-:
@asias I agree in general that the current state of things (the existance of auto_bootstrap flag, possibility of joining nodes and seeds, nodes not synchronizing when a new node joins) is bad and we need to fix all those issues. But the goal of this PR is not to change it, it's to make CDC work before we design and implement transactional node joining, as described in the PR cover letter. Please let me merge this.

asias:
> @asias I agree in general that the current state of things (the existence of auto_bootstrap flag, possibility of joining nodes as seeds, nodes not synchronizing when a new node joins) is bad and we need to fix all those issues. But the goal of this PR is not to change it, it's to make CDC work before we design and implement transactional node joining, as described in the PR cover letter. Please let me merge this.

The problem is:  what you are trying to workaround and "make cdc works" is not something we really support. Why do we want to add hacks in the code to pretend this is working and supported.  

Waiting for a node knows the seed node's token does not guarantee any thing.

n1 uses n1 as seed
n2 uses n1 as seed
n3 uses n1 as seed

When n2 and n3 started in parallel without bootstrap, n2 and n3 can still miss each other's token when generating cdc.

kbr-:
> The problem is: what you are trying to workaround and "make cdc works" is not something we really support. Why do we want to add hacks in the code to pretend this is working and supported.
> 
> Waiting for a node knows the seed node's token does not guarantee any thing.
> 
> n1 uses n1 as seed
> n2 uses n1 as seed
> n3 uses n1 as seed
> 
> When n2 and n3 started in parallel without bootstrap, n2 and n3 can still miss each other's token when generating cdc.

I'm aware that it's a workaround and a heuristic, I said it in the PR cover letter too.
I'm also aware of your example.

We're not pretending that parallel node joining is "working and supported", because it's not and I keep saying it everywhere. However, dtests, which use CCM, have been doing this parallel startup since forever (since we inherited those tools from Cassandra). We can't pretend that they don't.

This is not a long-term solution, *I am working on making it correct* but that'll take much more time.

Besides, the workaround already exists (the `wait_for_other_tokens` function), I'm just improving it a bit.

asias:
> I'm aware that it's a workaround and a heuristic, I said it in the PR cover letter too.
> I'm also aware of your example.
> 
> We're not pretending that parallel node joining is "working and supported", because it's not and I keep saying it everywhere. However, dtests, which use CCM, have been doing this parallel startup since forever (since we inherited those tools from Cassandra). We can't pretend that they don't.

Take a look at ccmlib/cluster.py, to modify ccm to stop starting node in parallel is trivial.  

> 
> This is not a long-term solution, _I am working on creating a correct solution_ but that'll take much more time.

Kamil, we are very close to the fast boot support. With it, we can start nodes properly and fast. 

> 
> Besides, the workaround already exists (the `wait_for_other_tokens` function), I'm just improving it a bit.

Existing  wait_for_other_tokens is also a hack to me.  We probably should not add it in the first place.

asias:
Even though I do not like the workaround, I think we should not block this patch.

1) It is better than current version of wait_for_other_tokens
2) We can remove this logic easily later when we have proper support.

So, ACK.

kbr-:
Thanks.

Unfortunately, even if we add fast sequential startup and use it in our tests, the option to use `auto_bootstrap=off` or join nodes as seeds will stay (so customers will be able to use it), so we won't be able to guarantee safety until we remove those options.

kbr-:
Not relevant anymore.

PR #5995, opened 2020-03-10
gossip: Make is_safe_for_bootstrap more strict

### LINKED ISSUE #5172: Node replace procedure may end up losing replaced node tokens
*Installation details*
Scylla version: 2.3.5
Cluster size: 5
OS: Ubuntu 16.04 + 18.04

*Hardware details*
Platform: bare metal

*Description*
1) Provision a cluster of 4 nodes.
2) Populate it with some data.
3) Stop and clean one node (nodeX): delete its `data` and `commitlog` directories.
4) Make nodeX replace itself: set `replace_address_first_boot: <IP of NodeX>` in its `scylla.yaml` and start it.
5) Kill -9 nodeX in the middle of streaming.
6) Remove replace_address_first_boot from the scylla.yaml
7) Clean nodeX'es data and commitlog.
7) Start nodeX .

See that nodeX starts bootstrapping as it's a new node in the cluster and other nodes don't have any information about previous tokens of nodeX as they had after step (3) above.

This effectively equivalent to `nodetool removenode` without streaming.

asias:
Consider

1. Start n1, n2 in the cluster
2. Stop n2 and delete all data for n2
3. Start n2 to replace itself with replace_address_first_boot: n2
4. Kill n2 before n2 finishes the replace operation
5. Remove replace_address_first_boot: n2 from scylla.yaml of n2
6. Delete all data for n2
7. Start n2

At step 7, n2 will be allowed to bootstrap as a new node, because the
application state of n2 in the cluster is HIBERNATE which is not
rejected in the check of is_safe_for_bootstrap. As a result, n2 will
replace n2 with a different tokens and a different host_id, as if the
old n2 node was removed from the cluster silently.

Fixes #5172

### DIFF
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -2174,25 +2174,24 @@ future<> gossiper::wait_for_range_setup() {
 }
 
 bool gossiper::is_safe_for_bootstrap(inet_address endpoint) {
+    // We allow to bootstrap a new node in only two cases:
+    // 1) The node is a completely new node and no state in gossip at all
+    // 2) The node has state in gossip and it is already removed from the
+    // cluster either by nodetool decommission or nodetool removenode
     auto* eps = get_endpoint_state_for_endpoint_ptr(endpoint);
-
-    // if there's no previous state, or the node was previously removed from the cluster, we're good
-    if (!eps || is_dead_state(*eps)) {
-        return true;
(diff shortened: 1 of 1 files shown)


avikivity:
Already committed as cdcedf5eb9c1d093b8053837fd3a23577253260b, closing

PR #5996, opened 2020-03-10
Add postimage implementation

### LINKED ISSUE #4992: Implement post-image of data observed in CDC
This is an enhancement of the minimal version of the Change Data Capture that logs only primary keys. It should make post image of data available in CDC.

This is a subtask of #4985.

elcallio:
Fixes #4992
    
Implements post-image support by synthesizing it from
pre-image + delta.
    
Post-image data differs from the delta data in two ways:
    
1.) It merges non-atomics into an actual result value
2.) It contains _all_ columns of the row, not just
     those affected by the update.
    
For a non-atomic field, the post-image value of a column
is either the pre-image or the delta (maybe null)
    
Tested by adding post-image checks to pre-image test
and collection/udt tests

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -479,6 +479,134 @@ api::timestamp_type find_timestamp(const schema& s, const mutation& m) {
     throw std::runtime_error("cdc: could not find timestamp of mutation");
 }
 
+// iterators for collection merge
+template<typename T>
+class collection_iterator : public std::iterator<std::input_iterator_tag, const T> {
+    bytes_view _v, _next;
+    size_t _rem = 0;
+    T _current;
+public:
+    collection_iterator(bytes_view_opt v = {}) 
+        : _v(v.value_or(bytes_view{}))
+        , _rem(_v.empty() ? 0 : read_collection_size(_v, cql_serialization_format::internal()))
+    {
+        if (_rem != 0) {
+            parse();
+        }
+    }
+    collection_iterator(const collection_iterator&) = default;
+    const T& operator*() const {
+        return _current;
+    }
+    const T* operator->() const {
+        return &_current;
+    }
+    collection_iterator& operator++() {
+        next();
+        if (_rem != 0) {
+            parse();
+        } else {
+            _current = {};
+        }
+        return *this;
+    }
+    collection_iterator operator++(int) {
+        auto v = *this;
+        ++(*this);
+        return v;
+    }
+    bool operator==(const collection_iterator& x) const {
+        return _v == x._v;
+    }
+    bool operator!=(const collection_iterator& x) const {
+        return !(*this == x);
+    }
+private:
+    void next() {
+        --_rem;
+        _v = _next;
+    }
+    void parse();
+};
+
+template<>
+void collection_iterator<std::pair<bytes_view, bytes_view>>::parse() {
+    assert(_rem > 0);
+    _next = _v;
+    auto k = read_collection_value(_next, cql_serialization_format::internal());
+    auto v = read_collection_value(_next, cql_serialization_format::internal());
+    _current = std::make_pair(k, v);
+}
+
+template<>
+void collection_iterator<bytes_view>::parse() {
+    assert(_rem > 0);
+    _next = _v;
+    auto k = read_collection_value(_next, cql_serialization_format::internal());
+    _current = k;
+}
+
+template<typename Container, typename T>
+class maybe_back_insert_iterator : public std::back_insert_iterator<Container> {
+    const abstract_type& _type;
+    collection_iterator<T> _s, _e;
+public:
+    using value_type = typename Container::value_type;
+    maybe_back_insert_iterator(Container& c, const abstract_type& type, collection_iterator<T> s)
+        : std::back_insert_iterator<Container>(c)
+        , _type(type)
+        , _s(s)
+    {}
+    maybe_back_insert_iterator& operator*() {
+        return *this;
+    }
+    maybe_back_insert_iterator& operator=(const value_type& v) {
+        if (!find(v)) {
+            std::back_insert_iterator<Container>::operator=(v);
+        }
+        return *this;
+    }
+    maybe_back_insert_iterator& operator=(value_type&& v) {
+        if (!find(v)) {
+            std::back_insert_iterator<Container>::operator=(std::move(v));
+        }
+        return *this;
+    }
+private:
+    bool find(const value_type& v) {
+       // cheating - reducing search span, because we know we only append unique values (see below).
+       while (_s != _e) {
+           auto n = compare(*_s, v);
+           if (n <= 0) {
+               ++_s;
+           }
+           if (n == 0) {
+               return true;
+           }
+           if (n > 0) {
+               break;
+           }
+       }
+       return false;
+    }
+    bool compare(const T&, const value_type& v);
+};
+
+template<>
+bool maybe_back_insert_iterator<std::vector<std::pair<bytes_view, bytes_view>>, bytes_view>::compare(const bytes_view& t, const value_type& v) {
+    return _type.compare(t, v.first);
+}
+
+template<>
+bool maybe_back_insert_iterator<std::vector<bytes_view>, bytes_view>::compare(const bytes_view& t, const value_type& v) {
+    return _type.compare(t, v);
+}
+
+template<typename Container, typename T>
+auto make_maybe_back_inserter(Container& c, const abstract_type& type, collection_iterator<T> s) {
+    return maybe_back_insert_iterator<Container, T>(c, type, s);
+}
+
 /* Given a timestamp, generates a timeuuid with the following properties:
  * 1. `t1` < `t2` implies timeuuid_type->less(timeuuid_type->decompose(generate_timeuuid(`t1`)),
  *                                            timeuuid_type->decompose(generate_timeuuid(`t2`))),
@@ -538,13 +666,72 @@ public:
             _cdc_ttl_opt = std::chrono::seconds(_schema->cdc_options().ttl());
         }
     }
+    static size_t collection_size(const bytes_opt& bo) {
+        if (bo) {
+            bytes_view bv(*bo);
+            return read_collection_size(bv, cql_serialization_format::internal());
+        }
+        return 0;
+    }
+    template<typename Func>
+    static void udt_for_each(const bytes_opt& bo, Func&& f) {
+        if (bo) {
+            bytes_view bv(*bo);
+            std::for_each(tuple_deserializing_iterator::start(bv), tuple_deserializing_iterator::finish(bv), std::forward<Func>(f));
+        }
+    }
+    static bytes merge(const collection_type_impl& ctype, const bytes_opt& prev, const bytes_opt& next, const bytes_opt& deleted) {
+        std::vector<std::pair<bytes_view, bytes_view>> res;
+        res.reserve(collection_size(prev) + collection_size(next));
+        auto type = ctype.name_comparator();
+        auto cmp = [&type = *type](const std::pair<bytes_view, bytes_view>& p1, const std::pair<bytes_view, bytes_view>& p2) {
+            return type.compare(p1.first, p2.first) < 0;
+        };
+        collection_iterator<std::pair<bytes_view, bytes_view>> e, i(prev), j(next);
+        // note order: set_union, when finding doubles, use value from first1 (j here). So 
+        // since this is next, it has prio
+        std::set_union(j, e, i, e, make_maybe_back_inserter(res, *type, collection_iterator<bytes_view>(deleted)), cmp);
+        return map_type_impl::serialize_partially_deserialized_form(res, cql_serialization_format::internal());
+    }
+    static bytes merge(const set_type_impl& ctype, const bytes_opt& prev, const bytes_opt& next, const bytes_opt& deleted) {
+        std::vector<bytes_view> res;
+        res.reserve(collection_size(prev) + collection_size(next));
+        auto type = ctype.name_comparator();
+        auto cmp = [&type = *type](bytes_view k1, bytes_view k2) {
+            return type.compare(k1, k2) < 0;
+        };
+        collection_iterator<bytes_view> e, i(prev), j(next), d(deleted);
+        std::set_union(j, e, i, e, make_maybe_back_inserter(res, *type, d), cmp);
+        return set_type_impl::serialize_partially_deserialized_form(res, cql_serialization_format::internal());
+    }
+    static bytes merge(const user_type_impl& type, const bytes_opt& prev, const bytes_opt& next, const bytes_opt& deleted) {
+        std::vector<bytes_view_opt> res(type.size());
+        udt_for_each(prev, [&res, i = res.begin()](bytes_view_opt k) mutable {
+            *i++ = k;
+        });
+        udt_for_each(next, [&res, i = res.begin()](bytes_view_opt k) mutable {
+            if (k) {
+                *i = k;
+            }
+            ++i;
+        });
+        collection_iterator<bytes_view> e, d(deleted);
+        std::for_each(d, e, [&res](bytes_view k) {
+            auto index = deserialize_field_index(k);
+            res[index] = std::nullopt;
+        });
+        return type.build_value(res);
+    }
+    static bytes merge(const abstract_type& type, const bytes_opt& prev, const bytes_opt& next, const bytes_opt& deleted) {
+        throw std::runtime_error(format("cdc merge: unknown type {}", type.name()));
+    }
 
     // TODO: is pre-image data based on query enough. We only have actual column data. Do we need
     // more details like tombstones/ttl? Probably not but keep in mind.
     mutation transform(const mutation& m, const cql3::untyped_result_set* rs, api::timestamp_type ts, bytes tuuid, int& batch_no) const {
         auto stream_id = _ctx._cdc_metadata.get_stream(ts, m.token());
         mutation res(_log_schema, stream_id.to_partition_key(*_log_schema));
-
+        const auto postimage = _schema->cdc_options().postimage();
         auto& p = m.partition();
         if (p.partition_tombstone()) {
             // Partition deletion
@@ -591,15 +778,20 @@ public:
             }
         } else {
             // should be insert, update or deletion
-            auto process_cells = [&](const row& r, column_kind ckind, const clustering_key& log_ck, std::optional<clustering_key> pikey, const cql3::untyped_result_set_row* pirow) -> std::optional<gc_clock::duration> {
-                bytes_opt value;
+            auto process_cells = [&](const row& r, column_kind ckind, const clustering_key& log_ck, std::optional<clustering_key> pikey, const cql3::untyped_result_set_row* pirow, std::optional<clustering_key> poikey) -> std::optional<gc_clock::duration> {
+                if (postimage && !poikey) {
+                    poikey = set_pk_columns(m.key(), ts, tuuid, ++batch_no, res);
+                    set_operation(*poikey, ts, operation::post_image, res);
+                }
                 std::optional<gc_clock::duration> ttl;
+                std::unordered_set<column_id> columns_assigned;
                 r.for_each_cell([&](column_id id, const atomic_cell_or_collection& cell) {
                     auto& cdef = _schema->column_at(ckind, id);
                     auto* dst = _log_schema->get_column_definition(log_data_column_name_bytes(cdef.name()));
                     auto has_pirow = pirow && pirow->has(cdef.name_as_text());
                     bool is_column_delete = true;
-
+                    bytes_opt value;
+                    bytes_opt deleted_elements = std::nullopt;
                     if (cdef.is_atomic()) {
                         value = std::nullopt;
                         auto view = cell.as_atomic_cell(cdef);
@@ -613,7 +805,6 @@ public:
                     } else {
                         auto mv = cell.as_collection_mutation();
                         is_column_delete = false;
-                        bytes_opt deleted_elements = std::nullopt;
                         std::vector<bytes> buf;
                         value = mv.with_deserialized(*cdef.type, [&](collection_mutation_view_description view) -> bytes_opt {
                             if (view.tomb) {
@@ -722,20 +913,51 @@ public:
                         res.set_cell(log_ck, *dst, atomic_cell::make_live(*dst->type, ts, *value, _cdc_ttl_opt));
                     }
 
-                    if (has_pirow) {
-                        value = get_preimage_col_value(cdef, pirow);
+                    bytes_opt prev;
 
+                    if (has_pirow) {
+                        prev = get_preimage_col_value(cdef, pirow);
                         assert(std::addressof(res.partition().clustered_row(*_log_schema, *pikey)) != std::addressof(res.partition().clustered_row(*_log_schema, log_ck)));
                         assert(pikey->explode() != log_ck.explode());
-                        res.set_cell(*pikey, *dst, atomic_cell::make_live(*dst->type, ts, *value, _cdc_ttl_opt));
+                        res.set_cell(*pikey, *dst, atomic_cell::make_live(*dst->type, ts, *prev, _cdc_ttl_opt));
+                    }
+
+                    if (postimage) {
+                        // keep track of actually assigning this already
+                        columns_assigned.emplace(id);
+                        // don't merge with pre-image iff column delete
+                        if (is_column_delete) {
+                            prev = std::nullopt;
+                        }
+                        if (cdef.is_atomic() && !is_column_delete && value) {
+                            res.set_cell(*poikey, *dst, atomic_cell::make_live(*dst->type, ts, *value, _cdc_ttl_opt));
+                        } else if (!cdef.is_atomic() && (value || (deleted_elements && prev))) {
+                            auto v = visit(*cdef.type, [&] (const auto& type) -> bytes {
+                                return merge(type, prev, value, deleted_elements);
+                            });
+                            res.set_cell(*poikey, *dst, atomic_cell::make_live(*dst->type, ts, v, _cdc_ttl_opt));
+                        }
                     }
                 });
+
+                // fill in all columns not already processed. Note that column nulls are also marked.
+                if (postimage && pirow) {
+                    for (auto& cdef : _schema->columns(ckind)) {
+                        if (!columns_assigned.count(cdef.id)) {
+                            auto v = pirow->get_view_opt(cdef.name_as_text());
+                            if (v) {
+                                auto dst = _log_schema->get_column_definition(log_data_column_name_bytes(cdef.name()));
+                                res.set_cell(*poikey, *dst, atomic_cell::make_live(*dst->type, ts, *v, _cdc_ttl_opt));
+                            }
+                        }
+                    }
+                }
 
                 return ttl;
             };
 
             if (!p.static_row().empty()) {
-                std::optional<clustering_key> pikey;
+                std::optional<clustering_key> pikey, poikey;
                 const cql3::untyped_result_set_row * pirow = nullptr;
 
                 if (rs && !rs->empty()) {
@@ -747,7 +969,13 @@ public:
                 }
 
                 auto log_ck = set_pk_columns(m.key(), ts, tuuid, batch_no, res);
-                auto ttl = process_cells(p.static_row().get(), column_kind::static_column, log_ck, pikey, pirow);
+
+                if (postimage) {
+                     poikey = set_pk_columns(m.key(), ts, tuuid, ++batch_no, res);
+                     set_operation(*poikey, ts, operation::post_image, res);
+                }
+
+                auto ttl = process_cells(p.static_row().get(), column_kind::static_column, log_ck, pikey, pirow, poikey);
 
                 set_operation(log_ck, ts, operation::update, res);
 
@@ -759,7 +987,7 @@ public:
                 for (const rows_entry& r : p.clustered_rows()) {
                     auto ck_value = r.key().explode(*_schema);
 
-                    std::optional<clustering_key> pikey;
+                    std::optional<clustering_key> pikey, poikey;
                     const cql3::untyped_result_set_row * pirow = nullptr;
 
                     if (rs) {
@@ -785,6 +1013,11 @@ public:
 
                     auto log_ck = set_pk_columns(m.key(), ts, tuuid, batch_no, res);
 
+                    if (postimage) {
+                        poikey = set_pk_columns(m.key(), ts, tuuid, ++batch_no, res);
+                        set_operation(*poikey, ts, operation::post_image, res);
+                    }
+
                     size_t pos = 0;
                     for (const auto& column : _schema->clustering_key_columns()) {
                         assert (pos < ck_value.size());
@@ -794,6 +1027,9 @@ public:
                         if (pirow) {
                             assert(pirow->has(column.name_as_text()));
                             res.set_cell(*pikey, *cdef, atomic_cell::make_live(*column.type, ts, bytes_view(ck_value[pos]), _cdc_ttl_opt));
+                        }
+                        if (poikey) {
+                            res.set_cell(*poikey, *cdef, atomic_cell::make_live(*column.type, ts, bytes_view(ck_value[pos]), _cdc_ttl_opt));
                         }
 
                         ++pos;
@@ -811,7 +1047,7 @@ public:
                             }
                         }
                     } else {
-                        auto ttl = process_cells(r.row().cells(), column_kind::regular_column, log_ck, pikey, pirow);
+                        auto ttl = process_cells(r.row().cells(), column_kind::regular_column, log_ck, pikey, pirow, poikey);
                         const auto& marker = r.row().marker();
                         if (marker.is_live() && marker.is_expiring()) {
                             ttl = marker.ttl();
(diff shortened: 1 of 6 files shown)


cdc/log.cc
```diff
@@ -539,12 +539,109 @@ class transformer final {
...
+        return type.build_value(res);
+    }
+
     // TODO: is pre-image data based on query enough. We only have actual column data. Do we need
     // more details like tombstones/ttl? Probably not but keep in mind.
     mutation transform(const mutation& m, const cql3::untyped_result_set* rs, api::timestamp_type ts, bytes tuuid, int& batch_no) const {
         auto stream_id = _ctx._cdc_metadata.get_stream(ts, m.token());
         mutation res(_log_schema, stream_id.to_partition_key(*_log_schema));
-
+        // should we generate post-image data? note: we can only do this if both pre- and postimage
```
haaawk: It's probably better for postimage=true to implicate preimage=true and check it when the user set's that. It can be done in separate patch though.

cdc/log.cc
```diff
@@ -591,15 +688,21 @@ class transformer final {
...
         } else {
             // should be insert, update or deletion
+            std::optional<clustering_key> poikey;
             auto process_cells = [&](const row& r, column_kind ckind, const clustering_key& log_ck, std::optional<clustering_key> pikey, const cql3::untyped_result_set_row* pirow) -> std::optional<gc_clock::duration> {
+                if (postimage && !poikey) {
+                    poikey = set_pk_columns(m.key(), ts, tuuid, ++batch_no, res);
+                    set_operation(*poikey, ts, operation::post_image, res);
+                }
                 bytes_opt value;
+                bytes_opt deleted_elements = std::nullopt;
```
haaawk: If I'm not mistaken, this variable can be in smaller scope of the lambda passed to r.for_each_cell in line 597/701

cdc/log.cc
```diff
@@ -722,15 +824,57 @@ class transformer final {
...
+                        // keep track of actually assigning this already
+                        columns_assigned.emplace(id);
+                        // don't merge with pre-image iff column delete
+                        if (is_column_delete) {
+                            prev = std::nullopt;
+                        }
+                        if (cdef.is_atomic() && !is_column_delete && value) {
+                            res.set_cell(*poikey, *dst, atomic_cell::make_live(*dst->type, ts, *value, _cdc_ttl_opt));
+                        } else if (!cdef.is_atomic() && (value || (deleted_elements && prev))) {
+                            auto v = visit(*cdef.type, make_visitor(
```
haaawk: Would it be possible to reduce code duplication by using a generic lambda visitor here? That would require having merge with overloads for collection_type_impl, set_type_impl and user_type_impl instead of having merege_maps, merge_sets and merge_udt.

cdc/log.cc
```diff
@@ -539,12 +539,109 @@ class transformer final {
...
+
     // TODO: is pre-image data based on query enough. We only have actual column data. Do we need
     // more details like tombstones/ttl? Probably not but keep in mind.
     mutation transform(const mutation& m, const cql3::untyped_result_set* rs, api::timestamp_type ts, bytes tuuid, int& batch_no) const {
         auto stream_id = _ctx._cdc_metadata.get_stream(ts, m.token());
         mutation res(_log_schema, stream_id.to_partition_key(*_log_schema));
-
+        // should we generate post-image data? note: we can only do this if both pre- and postimage 
+        // is enabled.
+        const auto postimage = _schema->cdc_options().preimage() && _schema->cdc_options().postimage();
```
haaawk: This is inconsistent with line 988/1140. There you fetch preimage when postimage is set to true no matter whether preimage is set to true or not. Why is there OR in that line and AND in this?

cdc/log.hh
```diff
@@ -126,6 +126,7 @@ enum class operation : int8_t {
     // enum decl, so don't change the constant values (or the datatype).
     pre_image = 0, update = 1, insert = 2, row_delete = 3, partition_delete = 4,
     range_delete_start_inclusive = 5, range_delete_start_exclusive = 6, range_delete_end_inclusive = 7, range_delete_end_exclusive = 8,
+    post_image = 9
```
haaawk: add ',' at the end to avoid editing this line when new enum value is added.

haaawk:
@nyh please have a look

kbr-:
> 2.) It contains all columns of the row, not just
those affected by the update.

Does it mean we change the preimage to always show all columns?

cdc/log.hh
```diff
@@ -126,6 +126,7 @@ enum class operation : int8_t {
     // enum decl, so don't change the constant values (or the datatype).
```
piodul: (Commenting here, because I can't attach this comment to a line that is not a part of the diff)

Line 74:

```c++
// Callback to be invoked on mutation finish to fix
// the whole bit about post-image.
// TODO: decide on what the parameters are to be for this.
using result_callback = std::function<future<>()>;
```

What about this? Are we still going to use it?

haaawk:
> > 2.) It contains all columns of the row, not just
> > those affected by the update.
> 
> Does it mean we change the preimage to always show all columns?

Which in my opinion would be good.

cdc/log.cc
```diff
@@ -539,12 +539,109 @@ class transformer final {
...
+            while (size--) {
+                f(read_collection_value(bv, cql_serialization_format::internal()));
+            }
+        }
+    }
+    template<typename Func>
+    static void udt_for_each(const bytes_opt& bo, Func&& f) {
+        if (bo) {
+            bytes_view bv(*bo);
+            std::for_each(tuple_deserializing_iterator::start(bv), tuple_deserializing_iterator::finish(bv), std::move(f));
```
kbr-: I'm not sure, but shouldn't this be `std::forward<Func>(f)` instead of `std::move(f)`?

cdc/log.cc
elcallio: I don't agree. Pre-image will add more data to the log table. If the user only wants post-image, he will have to pay the read-before-write price, but he should not be forced to suffer more log space usage than requested.

haaawk: Makes sense.

cdc/log.cc
elcallio: Right.

cdc/log.cc
elcallio: Not sure what you are getting at?

cdc/log.cc
```diff
@@ -539,12 +539,109 @@ class transformer final {
...
+    }
+    static bytes merge_maps(const abstract_type& type, const bytes_opt& prev, const bytes_opt& next, const bytes_opt& deleted) {
+        std::vector<std::pair<bytes_view, bytes_view>> res;
+        map_for_each(prev, [&res](bytes_view k, bytes_view v) {
+            res.emplace_back(k, v);
+        });
+        map_for_each(next, [&res, &type](bytes_view k, bytes_view v) {
+            auto i = std::find_if(res.begin(), res.end(), [&k, &type](std::pair<bytes_view, bytes_view>& p) {
+                return type.compare(p.first, k) >= 0;
+            });
```
kbr-: This is a quadratic algorithm, which means it will be unnecessarily slow even for quite small maps (I assume that 10000 keys is "small").
I'd suggest using a `map` instead of `vector` to construct the result so we get `O(n logn)` complexity.

Same comment applies to `merge_sets` below.

cdc/log.cc
elcallio: Oops. Mistake. First draft thing.

cql3/untyped_result_set.hh
```diff
@@ -77,6 +77,12 @@ public:
     std::optional<T> get_opt(const sstring& name) const {
         return has(name) ? get_as<T>(name) : std::optional<T>{};
     }
+    bytes_view_opt get_view_opt(const sstring& name) const {
+        if (has(name)) {
+            return get_view(name);
+        }
+        return std::nullopt;
```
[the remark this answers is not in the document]
elcallio: Compiler will complain because of differing types -> no.

cdc/log.hh
elcallio: Probably not, but will remove in different change set.

haaawk: We may want to keep it for CDC metrics. @piodul please comment on this.

cdc/log.cc
haaawk:
```
auto v = visit(*cdef.type, make_visitor([&] (const auto& type) -> bytes {
  return merge(type, prev, value, deleted_elements);
}));

and then create merge for each type (collection_type_impl, set_type_impl, user_type_impl).
```

kbr-: This could be
``make_visitor([&] (const auto& type) -> bytes { return merge(*type.name_comparator(), prev, value, deleted_elements); })``
where `merge` is overloaded in the first parameter (type). See also my comment below `merge_maps`.

But there's a problem that we have the `const abstract_type&` branch which throws an error. But maybe we could make two branches: one generic like shown in the example above for collections and UDTs, and one specific for `const abstract_type&`. Dunno if the compiler will handle this monster.

elcallio: With all the template<..> lines added it will not cut down the size of the file. I enjoy overloads as much as the other guy, but I think it does not really help here (it is 8 lines)

cdc/log.cc
```diff
@@ -539,12 +539,109 @@ class transformer final {
...
+        }
+    }
+    template<typename Func>
+    static void udt_for_each(const bytes_opt& bo, Func&& f) {
+        if (bo) {
+            bytes_view bv(*bo);
+            std::for_each(tuple_deserializing_iterator::start(bv), tuple_deserializing_iterator::finish(bv), std::move(f));
+        }
+    }
+    static bytes merge_maps(const abstract_type& type, const bytes_opt& prev, const bytes_opt& next, const bytes_opt& deleted) {
```
kbr-: It would probably be more kosher to take `const map_type_impl&` here.
Also, this could then be just called `merge` cuz the `_maps` suffix would be unnecessary due to the type.

Same comment applies to `merge_sets` below.

cdc/log.cc
kbr-: This wouldn't add any additional `template<...>` lines, you would just need to change name of `merge_maps` to `merge`, name of `merge_sets` to `merge` etc.

Also in this case it's less about number of lines, more about kosherness :P

cdc/log.cc
haaawk: It should.

schema.cc
```diff
@@ -1374,6 +1374,12 @@ schema::regular_columns() const {
             , _raw._columns.end());
 }
 
+schema::const_iterator_range_type
+schema::columns(column_kind kind) const {
+    return boost::make_iterator_range(_raw._columns.begin() + column_offset(kind)
+            , kind == column_kind::regular_column ? _raw._columns.end() : _raw._columns.begin() + column_offset(column_kind(int(kind) + 1)));
```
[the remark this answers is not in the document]
elcallio: Given that the code controls the enum, I think it is fine, but otoh it makes little difference, so why not. Matches the similar column_at method above.

cdc/log.cc
elcallio: If it was a general function, yes, but since we know always movable/copyable, it matters little.

haaawk: Still better to do perfect forwarding :) or maybe just use noncopyable_function and avoid templates?

cdc/log.cc
elcallio: I disagree. Using maps will create a plethora of nodes for every element, causing massive memory fragmentation. Since the objects are small, potential move-inserts should be relatively cheap, and actual collisions should be rate - if we overwrite, this is just a series of emplace_back, if we set an element, it is most likely only a few elements.

cdc/log.hh
piodul: I'm not sure yet. I initially thought we could use it to update the count of successful log row writes. Now, I don't think it could be used for accurate tracking of that metric, especially if multiple-mutation batches are involved. After `mutate_internal` finishes, we only are informed if everything succeeded or not (either it threw an exception, or not).

cdc/log.cc
elcallio: The type is the key type, not the map.

cdc/log.cc
```diff
@@ -900,18 +1044,26 @@ class transformer final {
...
+            }
         }
         if (!p.clustered_rows().empty()) {
             const bool has_row_delete = std::any_of(p.clustered_rows().begin(), p.clustered_rows().end(), [] (const rows_entry& re) {
                 return re.row().deleted_at();
             });
-
-            if (has_row_delete) {
+            // for postimage we need everything...
+            if (has_row_delete || _schema->cdc_options().postimage()) {
```
kbr-: Hmm, maybe we should just make it consistent and always fetch the entire row for preimage?
It would be good conceptually (probably), but with collections it could reduce performance significantly... so I'm not sure.

cdc/log.hh
haaawk: In that case, @elcallio please don't remove it. @piodul please see if it's useful for you and if not then remove it.

cdc/log.cc
haaawk: In my opinion, it's better for preimage to work the same way as postimage and always show full row.

cdc/log.cc
kbr-: > potential move-inserts should be relatively cheap

It doesn't matter when you have quadratic complexity. It is easy to create an example where a quadratic algorithm will create a 1 second stall, a 1 day stall, a 1 week stall etc.

cdc/log.cc
elcallio: I think pre-image should be minimal. It basically means the user says he will trade some less storage overhead for doing the merge/interpret merge into postimage himself.

cdc/log.cc
kbr-: For ~ `10^5` elements you'll have ~ `10^10` operations, which will already take about a second if I remember correctly (from my competitive programming days I remember that a quadratic solution wouldn't pass even for input size of `10^4` :P)

haaawk: I agree with Kamil and seeing his wise words I hope he won't dispute with https://github.com/scylladb/scylla/pull/5991#discussion_r390929962 :D

haaawk: BTW @elcallio I see your point too and I'm not a great fan of maps either. Sometimes the solution is to sort the collections and then base the efficient solution on that.

cdc/log.cc
haaawk: I guess having one of preimage/postimage being full and the other being partial is a reasonable midpoint. If user wants whole record for a change then they use postimage if they are interested only in changed bits then preimage + delta deliver just that. So I guess, I agree with @elcallio here after all :)

schema.cc
```diff
@@ -1374,6 +1374,22 @@ schema::regular_columns() const {
...
+    case column_kind::partition_key:
+        return partition_key_columns();
+    case column_kind::clustering_key:
+        return clustering_key_columns();
+    case column_kind::static_column:
+        return static_columns();
+    case column_kind::regular_column:
+        return regular_columns();
+    default:
+        std::abort();
```
haaawk: Remove this please. Without the default compiler should complain when switch does not cover all the values of the enum. With default the check is postponed to runtime.

cdc/log.cc
kbr-: Oh yes, this can be made in a cache-and-memory-friendly way using a [merge algorithm](https://en.wikipedia.org/wiki/Merge_algorithm). I think there's even a standard library function for that.

Btw, recommended read:
https://news.ycombinator.com/item?id=21743424

schema.cc
kbr-: You mean the `default:` branch, not the entire switch

haaawk:
It seems that @elcallio has addressed all our comments but the last one about the quadratic behavior. @elcallio are you planning to do it or shall we merge as is? I would vote for improving it since we know how to make it performant but it's your call.

haaawk:
Well I actually have one more comment about the default in switch which I believe we shouldn't do.

haaawk:
and let's wait for @kbr- to finish reviewing.

kbr-:
> It seems that @elcallio has addressed all our comments but the last one about the quadratic behavior. @elcallio are you planning to do it or shall we merge as is? I would vote for improving it since we know how to make it performant but it's your call.

Well I veto that, I believe this *will* become a problem in production :(
I'm willing to agree on one of the two options here:
- either use `map` for `n logn`, but Calle is right that it's not memory-friendly etc., so instead
- use a merge algorithm

cdc/log.cc
elcallio: That is the point. The vector is sorted. A better option is to already switch to use lower_bound, i.e. binary search.

kbr-: `lower_bound` would fix the quadraticity introduced by calling `find_if` here I agree, but there's still the problem with `emplace`s

haaawk: emplaces can be fixed with reserve.

kbr-: No, it won't fix the issue.
It will *maybe* cause allocations not to happen, but `emplace` will still be linear (it shifts elements to the right), so the entire algorithm will be quadratic.

elcallio:
I'm moving to lower_bound usage, but compile is slow.

haaawk:
> I'm moving to lower_bound usage, but compile is slow.

Would std::merge work? @elcallio

kbr-:
Nvm I don't veto, we can fix this in a follow-up PR (but I hope it will be fixed soon!)

elcallio:
`merge` is basically the same, but would require exploding the second collection from bytes first. I think it is better to use the callback model to minimize storage used.

haaawk:
> `merge` is basically the same, but would require exploding the second collection from bytes first. I think it is better to use the callback model to minimize storage used.

or create special iterators, I guess. It should require only linear access to the collections

elcallio:
Updated.

cdc/log.cc
```diff
@@ -538,13 +538,119 @@ class transformer final {
...
+        auto type = ctype.name_comparator();
+        auto cmp = [&type = *type](std::pair<bytes_view, bytes_view>& p, const bytes_view& k) {
+            return type.compare(p.first, k) < 0;
+        };
+        map_for_each(next, [&res, &cmp](bytes_view k, bytes_view v) {
+            auto i = std::lower_bound(res.begin(), res.end(), k, cmp);
+            if (i != res.end() && i->first == k) {
+                i->second = v;
+            } else {
+                res.emplace(i, k, v);
```
haaawk: This is still sad. Maybe we could collect those elements in another vector and at the end do std::merge of res and this new vector that keeps elements which live only in next?

elcallio: Why would that be better? Again, in most cases "next" will be small-ish, and merge would not really have better performance (barring a few memmoves), but we would instead need two temporaries more (one for dest, one for tmp source).

haaawk: I agree that next should be small but the problem is that complexity depends both on the size of next and the size of prev. If every element from next gets inserted before every element in prev, we will be moving elements size(next) * size(prev) times. I agree that std::merge with such temporary won't be much better though - just size(next) times faster. I think the best solution would be to implement a special iterator that computes elements of map from bytes and then use such iterators with std::merge. The question is whether you would like to do it yourself or should we do it in the following PR?

kbr-: say `prev` has `n` elements, `next` has `m` elements
for the first element in `prev`, this `emplace` has to shift `n` elements to the right in the worst case, `n/2` on average
for the second element, it has to shift `n + 1` elements in the worst case, `(n+1)/2` on average
...
for the `k`th element, it has to shift `n+k` elements in the worst case, `(n+k)/2` on average
and so on until `m`

so on average we have to perform `1/2 * ((n+1) + (n + 2) + ... + (n + m))` operations, in the worst case remove the `1/2`

so in total, the average complexity and the worst case complexity of this algorithm is `O((n+m) * m)`

kbr-: we can solve the problem using `std::merge`:
1. compute a sorted vector of elements from `prev`
2. compute a sorted vector of elements from `next`
3. merge the vectors using the merge algorithm, which has `O((n+m) * log (n+m))` complexity

haaawk: > say `prev` has `n` elements, `next` has `m` elements
> for the first element in `prev`, this `emplace` has to shift `n` elements to the right in the worst case, `n/2` on average
> for the second element, it has to shift `n + 1` elements in the worst case, `(n+1)/2` on average
> ...
> for the `k`th element, it has to shift `n+k` elements in the worst case, `(n+k)/2` on average
> and so on until `m`
> 
> so on average we have to perform `1/2 * ((n+1) + (n + 2) + ... + (n + m))` operations, in the worst case remove the `1/2`
> 
> so in total, the average complexity and the worst case complexity of this algorithm is `O((n+m) * m)`

This is not true because both maps are sorted. You're at most moving n*m times IMO

haaawk: > we can solve the problem using `std::merge`:
> 
> 1. compute a sorted vector of elements from `prev`
> 2. compute a sorted vector of elements from `next`
> 3. merge the vectors using the merge algorithm, which has `O((n+m) * log (n+m))` worst case complexity

This is is worse than having special iterator that extracts maps from bytes and use merge on those iterators. It requires 3 allocations instead of 1 and using 2 * (n + m) memory instead of just (n+m).

kbr-: The "sorted vector" is conceptual, it doesn't have to reside in memory all at once of course. You're right, an iterator can be used because we just need to walk over the vector once

kbr-: But the very least thing we need to do is reduce the complexity, **then** we can fight for improving the constants (2*x instead of x etc.)

jul-stas:
@elcallio One thing: I don't see clustering columns being set for `poikey`. Is it missing, done on purpose, or I'm missing something?

cdc/log.cc
kbr-: Argh, stupid error.
The merge algorithm has `O(n+m)` complexity, not `O((n+m) log (n+m))`.
So it will be linear because our maps are already sorted :)

kbr-: > This is not true because both maps are sorted. You're at most moving n*m times IMO

Right. You never shift elements from the `next` map, because you walk it in sorted order. So on average it's `n/2` elements shifted for every element in `next`, i.e. `n/2 * m` shifts total on average. `n * m` shifts total in the worst case.

haaawk: @elcallio Do you want to do this or should Kamil or me do it?

cdc/log.cc
```diff
@@ -591,15 +697,21 @@ class transformer final {
             }
         } else {
             // should be insert, update or deletion
+            std::optional<clustering_key> poikey;
             auto process_cells = [&](const row& r, column_kind ckind, const clustering_key& log_ck, std::optional<clustering_key> pikey, const cql3::untyped_result_set_row* pirow) -> std::optional<gc_clock::duration> {
-                bytes_opt value;
+                if (postimage && !poikey) {
+                    poikey = set_pk_columns(m.key(), ts, tuuid, ++batch_no, res);
+                    set_operation(*poikey, ts, operation::post_image, res);
+                }
```
kbr-: Why are we setting `poikey` here and not near the place where `pikey` and `log_ck` are set? This choice of placement of this code seems pretty random to me.

elcallio: To avoid code duplication. Since we only either do static cols or regular cols, this is fine. There will be a separate transform call for each.

elcallio:
New version with iterator-based in-place merge of collections.

cdc/log.cc
elcallio: Now iterator-based.

cdc/log.cc
```diff
@@ -479,6 +479,134 @@ api::timestamp_type find_timestamp(const schema& s, const mutation& m) {
     throw std::runtime_error("cdc: could not find timestamp of mutation");
 }
 
+// iterators for collection merge
+template<typename T>
+class collection_iterator : public std::iterator<std::input_iterator_tag, const T> {
```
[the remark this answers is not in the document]
elcallio: It is deprecated in C++20, but we generally use c++17-- semantics for iterator classification. I suggest doing a global sweep upgrading the tagging once the dust finally settles on c++20 (i.e. I trust gcc stdlib is fully implemented in this regard).

cdc/log.cc
```diff
@@ -479,6 +479,134 @@ api::timestamp_type find_timestamp(const schema& s, const mutation& m) {
...
+template<>
+void collection_iterator<bytes_view>::parse() {
+    assert(_rem > 0);
+    _next = _v;
+    auto k = read_collection_value(_next, cql_serialization_format::internal());
+    _current = k;
+}
+
+template<typename Container, typename T>
+class maybe_back_insert_iterator : public std::back_insert_iterator<Container> {
```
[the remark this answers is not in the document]
elcallio: Why? No standard iterators are deprecated afaik, only the tagging system. Back inserter is never gonna be removed. Way to convenient.

cdc/log.cc
```diff
@@ -479,6 +479,134 @@ api::timestamp_type find_timestamp(const schema& s, const mutation& m) {
...
+        return *this;
+    }
+    maybe_back_insert_iterator& operator=(value_type&& v) {
+        if (!find(v)) {
+            std::back_insert_iterator<Container>::operator=(std::move(v));
+        }
+        return *this;
+    }
+private:
+    bool find(const value_type& v) {
```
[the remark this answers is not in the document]
elcallio: Later.

cdc/log.cc
```diff
@@ -479,6 +479,134 @@ api::timestamp_type find_timestamp(const schema& s, const mutation& m) {
...
+
+template<>
+void collection_iterator<bytes_view>::parse() {
+    assert(_rem > 0);
+    _next = _v;
+    auto k = read_collection_value(_next, cql_serialization_format::internal());
+    _current = k;
+}
+
+template<typename Container, typename T>
```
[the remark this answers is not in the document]
elcallio: Here: not really, but after some ironing and verification maybe these should be promoted to global scope and generalized usage, and then it will be needed in some way or another. Also hurts noone.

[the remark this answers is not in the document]
haaawk: +1 to generalizing. They seem useful for collections.

haaawk:
@nyh please review. The code is now shinny and as one president would say "great again" :D

haaawk:
> @elcallio One thing: I don't see clustering columns being set for `poikey`. Is it missing, done on purpose, or I'm missing something?

@elcallio what about this?

@jul-stas checked that ck is not set in postimage:
```
cqlsh> insert into ks.tbl_postimg (pk, ck, val) values (4,11,111);
cqlsh> select * from ks.tbl_postimg_scylla_cdc_log  ;
 cdc$stream_id                      | cdc$time                             | cdc$batch_seq_no | cdc$deleted_ck | cdc$deleted_pk | cdc$deleted_val | cdc$operation | cdc$ttl | ck   | pk | val
------------------------------------+--------------------------------------+------------------+----------------+----------------+-----------------+---------------+---------+------+----+------
 0xc5b7c7cda04517d55971351e7ef9e93b | ce7bc72e-63bb-11ea-f3a6-8a2ec3898046 |                0 |           null |           null |            null |             2 |    null |   11 |  4 |  111
 0xc5b7c7cda04517d55971351e7ef9e93b | ce7bc72e-63bb-11ea-f3a6-8a2ec3898046 |                1 |           null |           null |            null |             9 |    null | null |  4 |  111
```

cdc/log.cc
```diff
@@ -479,6 +479,134 @@ api::timestamp_type find_timestamp(const schema& s, const mutation& m) {
...
+    }
+    maybe_back_insert_iterator& operator=(value_type&& v) {
+        if (!find(v)) {
+            std::back_insert_iterator<Container>::operator=(std::move(v));
+        }
+        return *this;
+    }
+private:
+    bool find(const value_type& v) {
+       // cheating, because we know we only append unique values (see below).
```
nyh: I didn't understand what this "cheat" means, nor why find() modifies _s. Is this the cheat? And where below should I see?

test/boost/cdc_test.cc
```diff
@@ -427,27 +427,32 @@ SEASTAR_THREAD_TEST_CASE(test_primary_key_logging) {
...
             auto rows = select_log(e, "tbl");
 
             BOOST_REQUIRE(to_bytes_filtered(*rows, cdc::operation::pre_image).empty());
+            BOOST_REQUIRE(to_bytes_filtered(*rows, cdc::operation::post_image).empty() == !enabled);
 
             auto first = to_bytes_filtered(*rows, cdc::operation::update);
 
             auto ck2_index = column_index(*rows, cdc::log_data_column_name("ck2"));
             auto val_index = column_index(*rows, cdc::log_data_column_name("val"));
+            auto val2_index = column_index(*rows, cdc::log_data_column_name("val2"));
```
nyh: There's something I don't understand here. If we have both a post-image and a delta, like I think we have here, what does this search find, the post-image or the delta? Should the test search both, and verify that they are both correct? The code below seems to do exactly this, but this code doesn't.

test/boost/cdc_test.cc
```diff
@@ -460,18 +465,28 @@ SEASTAR_THREAD_TEST_CASE(test_pre_image_logging) {
...
                 auto pre_image = to_bytes_filtered(*rows, cdc::operation::pre_image);
                 auto second = to_bytes_filtered(*rows, cdc::operation::update);
+                auto post_image = to_bytes_filtered(*rows, cdc::operation::post_image);
 
                 if (!enabled) {
                     BOOST_REQUIRE(pre_image.empty());
+                    BOOST_REQUIRE(post_image.empty());
                 } else {
                     sort_by_time(*rows, second);
+                    sort_by_time(*rows, pre_image);
```
nyh: This change fixes a bug in the pre_image test?

test/boost/cdc_test.cc
```diff
@@ -460,18 +465,28 @@ SEASTAR_THREAD_TEST_CASE(test_pre_image_logging) {
...
+                    sort_by_time(*rows, post_image);
                     BOOST_REQUIRE_EQUAL(pre_image.size(), i + 1);
 
                     val = *pre_image.back()[val_index];
```
nyh: nitpick: now that you added val2, the preimage should contain it too, you can check that...

test/boost/cdc_test.cc
```diff
@@ -492,25 +507,30 @@ SEASTAR_THREAD_TEST_CASE(test_pre_image_logging) {
...
 
             auto rows = select_log(e, "tbl");
 
             BOOST_REQUIRE(to_bytes_filtered(*rows, cdc::operation::pre_image).empty());
+            BOOST_REQUIRE(to_bytes_filtered(*rows, cdc::operation::post_image).empty() == !enabled);
 
             auto first = to_bytes_filtered(*rows, cdc::operation::update);
 
             auto s_index = column_index(*rows, cdc::log_data_column_name("s"));
+            auto s2_index = column_index(*rows, cdc::log_data_column_name("s2"));
```
nyh: All my comments/questions for the previous test also applies to this one. I wonder if both tests could be folded into one test instead of duplicating most of the code...

schema.cc
```diff
@@ -1374,6 +1374,21 @@ schema::regular_columns() const {
...
+    case column_kind::partition_key:
+        return partition_key_columns();
+    case column_kind::clustering_key:
+        return clustering_key_columns();
+    case column_kind::static_column:
+        return static_columns();
+    case column_kind::regular_column:
+        return regular_columns();
+    }
+    std::abort();
```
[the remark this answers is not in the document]
kbr-: Maybe the `abort` can be removed because the compiler already deduced that code after the `switch` is unreachable.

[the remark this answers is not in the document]
haaawk: Probably it can't because the compiler will complain that there's a flow that exits the function without returning anything.

cdc/log.cc
```diff
@@ -722,15 +914,46 @@ class transformer final {
...
+                        // keep track of actually assigning this already
+                        columns_assigned.emplace(id);
+                        // don't merge with pre-image iff column delete
+                        if (is_column_delete) {
+                            prev = std::nullopt;
+                        }
+                        if (cdef.is_atomic() && !is_column_delete && value) {
+                            res.set_cell(*poikey, *dst, atomic_cell::make_live(*dst->type, ts, *value, _cdc_ttl_opt));
+                        } else if (!cdef.is_atomic() && (value || (deleted_elements && prev))) {
+                            auto v = visit(*cdef.type, make_visitor([&] (const auto& type) -> bytes {
```
piodul: `make_visitor` is not needed here, the generic lambda is a valid visitor here by itself.

schema.cc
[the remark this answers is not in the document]
elcallio: Yup. This is straight from the `columns_count?` playbook. If I change here I should change the preceding case as well. Symmetry.

cdc/log.cc
elcallio: The cheat is modifying _s, yes. We know we only use the filter with `set_union` of unique (no repeats) sets, so it is safe to reduce the filter range like I do. If this was a generalized type though, it would not be. 
For generalizing, it should instead use a predicate operator callback or something, but I opted away that in this case, because that has higher overhead.

test/boost/cdc_test.cc
elcallio: The code above just figures out column index in result sets for our cdc log queries. This is idempotent across all queries in a single test run, but might vary across builds/versions.

test/boost/cdc_test.cc
elcallio: It pre-empts a potential bug.

test/boost/cdc_test.cc
elcallio: It probably could, but that is not the scope of this PR.

cdc/log.cc
elcallio: Quite.

elcallio:
@haaawk - ugh. Ok, will check where it borks. Nice catch. We probably need to update tests to check for this.

schema.cc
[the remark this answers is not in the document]
elcallio: `on_internal_error`requires a logger. We don't have a logger in schema (nor should we - it is more or less a pure data object). Does not fit well.

test/boost/cdc_test.cc
elcallio: No, pre-image only contains columns modified. val2 is not touched by update, so not present.

schema.cc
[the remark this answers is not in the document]
elcallio: I've changed it to simply throw an `invalid_argument`. It seems more appropriate anyway.

elcallio:
@haaawk , @nyh - all issues/remarks addressed.

test/boost/cdc_test.cc
nyh: Oh, I didn't realize this until now! Wow. I'll send a separate mail about this issue.

nyh:
Thanks. Committed to next. I sent a separate email regarding the (unrelated) preimage issue that I suddenly realized.

haaawk:
Thanks @nyh for both committing and sending an email with DynamoDB perspective on preimage.

PR #6007, opened 2020-03-11
compaction: use a larger min_threshold during bootstrap, replace

glommer:
During bootstrap and replace operations the node can't take reads and
we'd like to see the process ending ASAP. This is because until the
process ends, we keep having to duplicate writes to an extended set. Not
to mention, in the case of a cluster expansion users want to use the
added capacity sooner rather than later.

Streaming generates a lot of compaction activity, that competes with the
bootstrap itself, slowing it down.

Long term, we are moving to treat those compactions differently and
maybe postpone them altogether. However for now we can reduce the amount
of compactions by increasing the minimum threshold of SSTables that have
to accumulate before they are selected for compactions. The default is
2, meaning we will trigger a compaction every time 2 SSTables of about
the same size are found (for STCS, others follow a similar pattern).

Until we have offstrategy infrastructure we don't want the compactions
to stop happening altogether so the reads, when they start, don't suffer.
This patch sets the minimum threshold to 16, meaning we will generate a
lot less compaction activity during streaming. Once streaming is done we
revert it to its original.

Unfortunately there isn't much we can do at the moment about decommission.
During decommission the nodes receiving data are also taking reads and
we don't want SSTables to accumulate.

### DIFF
--- a/database.hh
+++ b/database.hh
@@ -546,6 +546,7 @@ private:
     // The master resides in system.truncated table
     db_clock::time_point _truncated_at = db_clock::time_point::min();
 
+    bool _is_bootstrap_or_replace = false;
 public:
     future<> add_sstable_and_update_cache(sstables::shared_sstable sst);
     future<> move_sstables_from_staging(std::vector<sstables::shared_sstable>);
@@ -563,6 +564,14 @@ public:
     }
     db_clock::time_point get_truncation_record() {
         return _truncated_at;
+    }
+
+    void notify_bootstrap_or_replace_start() {
+        _is_bootstrap_or_replace = true;
+    }
+
+    void notify_bootstrap_or_replace_end() {
+        _is_bootstrap_or_replace = false;
     }
 private:
     void update_stats_for_new_sstable(uint64_t disk_space_used_by_sstable, const std::vector<unsigned>& shards_for_the_sstable) noexcept;
@@ -870,7 +879,20 @@ public:
     }
 
     bool compaction_enforce_min_threshold() const {
-        return _config.compaction_enforce_min_threshold;
+        return _config.compaction_enforce_min_threshold || _is_bootstrap_or_replace;
+    }
+
+    unsigned min_compaction_threshold() {
+        // During receiving stream operations, the less we compact the faster streaming is. For
+        // bootstrap and replace thereThere are no readers so it is fine to be less aggressive with
+        // compactions as long as we don't ignore them completely (this could create a problem for
+        // when streaming ends)
+        if (_is_bootstrap_or_replace) {
+            auto target = std::min(schema()->max_compaction_threshold(), 16);
+            return std::max(schema()->min_compaction_threshold(), target);
+        } else {
+            return schema()->min_compaction_threshold();
+        }
     }
 
     /*!
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -81,6 +81,7 @@
 #include <seastar/core/metrics.hh>
 #include "cdc/generation.hh"
 #include "repair/repair.hh"
+#include "service/priority_manager.hh"
 
 using token = dht::token;
 using UUID = utils::UUID;
@@ -935,6 +936,12 @@ void storage_service::bootstrap() {
 
     _gossiper.check_seen_seeds();
 
+    _db.invoke_on_all([this] (database& db) {
+        for (auto& cf : db.get_non_system_column_families()) {
+            cf->notify_bootstrap_or_replace_start();
+        }
+    }).get();
+
     set_mode(mode::JOINING, "Starting to bootstrap...", true);
     if (is_repair_based_node_ops_enabled()) {
         if (db().local().is_replacing()) {
@@ -947,6 +954,13 @@ void storage_service::bootstrap() {
         // Does the actual streaming of newly replicated token ranges.
         bs.bootstrap().get();
     }
+    _db.invoke_on_all([this] (database& db) {
+        for (auto& cf : db.get_non_system_column_families()) {
+            cf->notify_bootstrap_or_replace_end();
+        }
+    }).get();
+
+
     slogger.info("Bootstrap completed! for the tokens {}", _bootstrap_tokens);
 }
 
--- a/sstables/leveled_manifest.hh
+++ b/sstables/leveled_manifest.hh
@@ -49,6 +49,7 @@
 #include <boost/range/algorithm/partial_sort.hpp>
 
 class leveled_manifest {
+    table& _table;
     schema_ptr _schema;
     std::vector<std::vector<sstables::shared_sstable>> _generations;
     uint64_t _max_sstable_size_in_bytes;
@@ -82,7 +83,8 @@ public:
     static constexpr float TARGET_SCORE = 1.001f;
 private:
     leveled_manifest(column_family& cfs, int max_sstable_size_in_MB, const sstables::size_tiered_compaction_strategy_options& stcs_options)
-        : _schema(cfs.schema())
+        : _table(cfs)
+        , _schema(cfs.schema())
         , _max_sstable_size_in_bytes(max_sstable_size_in_MB * 1024 * 1024)
         , _stcs_options(stcs_options)
     {
@@ -245,7 +247,7 @@ public:
             // TODO: we shouldn't proceed with size tiered strategy if cassandra.disable_stcs_in_l0 is true.
             if (get_level_size(0) > MAX_COMPACTING_L0) {
                 auto most_interesting = sstables::size_tiered_compaction_strategy::most_interesting_bucket(get_level(0),
-                    _schema->min_compaction_threshold(), _schema->max_compaction_threshold(), _stcs_options);
+                    _table.min_compaction_threshold(), _schema->max_compaction_threshold(), _stcs_options);
                 if (!most_interesting.empty()) {
                     logger.debug("L0 is too far behind, performing size-tiering there first");
                     return sstables::compaction_descriptor(std::move(most_interesting));
@@ -430,7 +432,7 @@ private:
             // do STCS in L0 when max_sstable_size is high compared to size of new sstables, so we'll
             // avoid quadratic behavior until L0 is worth promoting.
             candidates = sstables::size_tiered_compaction_strategy::most_interesting_bucket(get_level(0),
-                _schema->min_compaction_threshold(), _schema->max_compaction_threshold(), _stcs_options);
+                _table.min_compaction_threshold(), _schema->max_compaction_threshold(), _stcs_options);
         }
         return { std::move(candidates), can_promote };
     }
--- a/sstables/compaction_manager.cc
+++ b/sstables/compaction_manager.cc
@@ -153,7 +153,7 @@ int compaction_manager::trim_to_compact(column_family* cf, sstables::compaction_
     }
 
     uint64_t total_size = get_total_size(descriptor.sstables);
-    int min_threshold = cf->schema()->min_compaction_threshold();
+    int min_threshold = cf->min_compaction_threshold();
     auto compacting_run_identifiers = boost::copy_range<std::unordered_set<utils::UUID>>(descriptor.sstables
         | boost::adaptors::transformed(std::mem_fn(&sstables::sstable::run_identifier)));
 
--- a/sstables/size_tiered_compaction_strategy.cc
+++ b/sstables/size_tiered_compaction_strategy.cc
@@ -135,7 +135,7 @@ size_tiered_compaction_strategy::most_interesting_bucket(std::vector<std::vector
 compaction_descriptor
 size_tiered_compaction_strategy::get_sstables_for_compaction(column_family& cfs, std::vector<sstables::shared_sstable> candidates) {
     // make local copies so they can't be changed out from under us mid-method
-    int min_threshold = cfs.schema()->min_compaction_threshold();
+    int min_threshold = cfs.min_compaction_threshold();
     int max_threshold = cfs.schema()->max_compaction_threshold();
     auto gc_before = gc_clock::now() - cfs.schema()->gc_grace_seconds();
 
@@ -177,7 +177,7 @@ size_tiered_compaction_strategy::get_sstables_for_compaction(column_family& cfs,
 }
 
 int64_t size_tiered_compaction_strategy::estimated_pending_compactions(column_family& cf) const {
-    int min_threshold = cf.schema()->min_compaction_threshold();
+    int min_threshold = cf.min_compaction_threshold();
     int max_threshold = cf.schema()->max_compaction_threshold();
     std::vector<sstables::shared_sstable> sstables;
     int64_t n = 0;
--- a/sstables/time_window_compaction_strategy.hh
+++ b/sstables/time_window_compaction_strategy.hh
@@ -220,9 +220,9 @@ private:
         // Update the highest window seen, if necessary
         _highest_window_seen = std::max(_highest_window_seen, p.second);
 
-        update_estimated_compaction_by_tasks(p.first, cf.schema()->min_compaction_threshold());
+        update_estimated_compaction_by_tasks(p.first, cf.min_compaction_threshold());
 
-        return newest_bucket(std::move(p.first), cf.schema()->min_compaction_threshold(), cf.schema()->max_compaction_threshold(),
+        return newest_bucket(std::move(p.first), cf.min_compaction_threshold(), cf.schema()->max_compaction_threshold(),
             _options.sstable_window_size, _highest_window_seen, _stcs_options);
     }
 public:


glommer:
@slivne FYI

raphaelsc:
Increasing min threshold will barely have any effect in LCS, given that compaction on levels higher than 0 is about getting one sstable from level N and compacting it with overlapping sstables in level N+1. Min threshold is only used when level 0 falls behind and LCS uses STCS to reduce number of sstables in level 0.

That being said, I'd suggest that we have a strategy specific change in behavior when we're in the bootstrap mode. 
For STCS and TWCS, increasing the min threshold will work just fine. 
But for LCS, we'll need to decrease activity on levels > 0 or even disable it altogether in addition to increasing the min threshold.

service/storage_service.cc
```diff
@@ -937,6 +938,12 @@ void storage_service::bootstrap() {
 
     _gossiper.check_seen_seeds();
 
+    _db.invoke_on_all([this] (database& db) {
+        for (auto& cf : db.get_non_system_column_families()) {
+            cf->notify_stream_recv_start();
+        }
```
raphaelsc: Do we really need this? I was under the impression that storage_service::_is_bootstrap_mode could be used to determine when bootstrap phase is done.

glommer:
I agree it won't have much an effect on LCS, but this is a stop gap temporarily.
We are fixing this for real with offstrategy compaction and I am not willing to add strategy heuristics of any kind at this moment.

service/storage_service.cc
glommer: We do, because the places in which we check don't have access to the storage service.

raphaelsc:
> I agree it won't have much an effect on LCS, but this is a stop gap temporarily.
> We are fixing this for real with offstrategy compaction and I am not willing to add strategy heuristics of any kind at this moment.

Understood. But we could easily get this done by converting this LCS code, look: https://github.com/scylladb/scylla/blob/f3f86876c20c232e83f071343dcf92f64d41c36c/sstables/leveled_manifest.hh#L188

If we want this temporary fix to have an actual effect for LCS users, then I think we should at least do what I suggested above. Otherwise, this temporary fix will be limited to TWCS and STCS users.

service/storage_service.cc
raphaelsc: Understood. thanks

glommer:
I don't know what to "look" for, github just takes me to the actual file. Too much context there.
What exactly is it that you propose?

I am totally fine with this being only for STCS and TWCS. This is temporary, and the min_threshold for STCS and TWCS used to be 4. In a sense we regressed our bootstraps when we reduced it.

I am only touching LCS if it is trivial. But I can't see to which change you are pointing to

asias:
> Until we have offstrategy infrastructure we don't want the compactions
> to stop happening altogether so the reads, when they start, don't suffer.
> This patch sets the minimum threshold to 16, meaning we will generate a
> lot less compaction activity during streaming. Once streaming is done we
> revert it to its original.

In the old regular based streaming, 10 sstables  will be generated per peer node per table, so consider a 2 node cluster, 3rd node is joining. With threshold 16, we will have compaction, 80% is done streaming the table (each finishes 8 sstables out of 10).

I propose to make the minimum threshold more aggressive. 

Let N the number of peers to stream from for a table, the total sstables generated by streaming would be 10 * N for this table. 

Making minimum threshold =  10 * N is too aggressive which is  O(N).

Perhaps 64, so it will cover 6 peer nodes streaming case. 

If we worry about too many sstables for read workload once bootstrap is done, we can compact until a threshold (maximum sstables to allow) is meet then mark the stream is done

glommer:
@asias we don't want to be too aggressive. If streaming ends and we haven't compacted anything, we're in trouble.

The maximum is 32, so I set it to half of that

asias:
> @asias we don't want to be too aggressive. If streaming ends and we haven't compacted anything, we're in trouble.
> 
> The maximum is 32, so I set it to half of that

How about: compact at the end until threshold is meet then finish streaming?

glommer:
All of that is too complex and needs testing and validation. That's close to what offstrategy will do, but we can't do any of that for the current cycle in 4.0

We can decide to live with the problem for longer, which is fine, but if we want to do something to alleviate it now we'll have to settle for something simple and effective and stop trying to be perfect.

asias:
> All of that is too complex and needs testing and validation. That's close to what offstrategy will do, but we can't do any of that for the current cycle in 4.0

It looks trivial to check the number of sstables and mark the bootstrap done.  We need to test and validate the changing of threshold from 2 to 16 anyway. What happens if the threshold exceeds the current max 32?

> 
> We can decide to live with the problem for longer, which is fine, but if we want to do something to alleviate it now we'll have to settle for something simple and effective and stop trying to be perfect.

I totally agree we can offer a simple but not perfect solution. I hope we can make it a little better without too much effect.

asias:
Another idea is to make streaming produces even less sstables. Currently, we stream 10% of ranges per stream plan, we can make it 20% so that it produces 5 sstables instead of 10 per table, as if we had increased the threshold 2 times.  The downside is we need to restream more if error happens, but considering it happens rarely, I think it is worth it.

glommer:
If you are talking about errors that retry a range, I don't think they are that rare.
Errors that are permanent and have to restart the bootstrap are more rare but if we're only talking about those why not create a single sstable per stream plan ?

asias:
> If you are talking about errors that retry a range, I don't think they are that rare.

Yes, I mean retry a range. This can happen when a node is marked as DOWN by gossip, e.g., due to a temp network issues or the node is not responding, lasting more than ~30 seconds. Yes, it happens, what's why I am not recommending to stream 100%  ranges at a time.  If it is hard to reason about increasing the max sstables threshold,  it is much easier to reason about the number of sstables generated. 

> Errors that are permanent and have to restart the bootstrap are more rare but if we're only talking about those why not create a single sstable per stream plan ?

We do create single sstable per stream plan. We use multiple stream plan to stream the whole vnode token ranges, each stream plan covers 10% of those ranges.

glommer:
Asias, it is not at all hard to reason about min_threshold and it is the least risky option by far.

Here's what's important:
- The max_threshold is 32. We hit it sometimes in practice in busy nodes, and as long as we don't raise it we don't risk new OOM issues arising from compacting too much together
- Any number below 32 is good. The higher the number the better because that means we'll compact fewer times.
- 16 is close to the 10 SSTables we will generate from a single node guaranteeing that we'll always do at least one intermediary compaction even in the case where a single node streams.

That's it.

asias:
> Asias, it is not at all hard to reason about min_threshold and it is the least risky option by far.

I meant to reason about setting min_threshold beyond current max 32. 

> 
> Here's what's important:
> 
>     * The max_threshold is 32. We hit it sometimes in practice in busy nodes, and as long as we don't raise it we don't risk new OOM issues arising from compacting too much together
> 
>     * Any number below 32 is good. The higher the number the better because that means we'll compact fewer times.

OK.  Then I would choose 30 instead of 16 to make it more aggressive and safe and not reaching current maximum. 

> 
>     * 16 is close to the 10 SSTables we will generate from a single node guaranteeing that we'll always do at least one intermediary compaction even in the case where a single node streams.

If there is one node streaming and we choose 16, we will not trigger intermediary compaction at all. I do not think we need to choose the number close to what sstables generated by streaming. We just need to choose a number that is guaranteed to be safe. Once the sstables generated by streaming exceeds, a intermediary compaction happens.

> 
> 
> That's it.

glommer:
Asias, you are thinking too much about how to extract the maximum about this without taking into consideration the need to be conservative with a change that comes without the supporting infrastructure (offstrategy).

16 is the conservative choice.

raphaelsc:
> I don't know what to "look" for, github just takes me to the actual file. Too much context there.
> What exactly is it that you propose?
> 
> I am totally fine with this being only for STCS and TWCS. This is temporary, and the min_threshold for STCS and TWCS used to be 4. In a sense we regressed our bootstraps when we reduced it.
> 
> I am only touching LCS if it is trivial. But I can't see to which change you are pointing to

Please see if this link works for you. It will be relatively simple to convert that code.
https://github.com/scylladb/scylla/blob/f3f86876c20c232e83f071343dcf92f64d41c36c/sstables/leveled_manifest.hh#L188 (Line 188)

glommer:
Thanks @raphaelsc . Okay, I see.

I see drawbacks of doing that too. We can end up with many SSTables - over 16 - that an LCS reader will have to add to their reader set. That means that once bootstrap is done, reads on LCS will be way more expensive than what they would otherwise be.

Guys, let's all keep in mind that we are working on a solution for this that is more robust so we want to do things that are low risk and high payoff now and are not too complex.

asias:
> Asias, you are thinking too much about how to extract the maximum about this without taking into consideration the need to be conservative with a change that comes without the supporting infrastructure (offstrategy).
> 
> 16 is the conservative choice.

You said any number below 32 is good, no?  Yes, smaller number is more conservative at the cost of more compactions. We'd better to get some test numbers. If setting to 16 already improves a lot compared to current version and setting to 30 does not improve much compared to 16, we can definitely set to safer value.  If 30 is much much better than 16, we can reconsider to be less conservative. 

Also, I still think we'd better reduce the number of stream plans (from 10 to 5) for bootstrap and replace.

asias:
> Thanks @raphaelsc . Okay, I see.
> 
> I see drawbacks of doing that too. We can end up with many SSTables - over 16 - that an LCS reader will have to add to their reader set. That means that once bootstrap is done, reads on LCS will be way more expensive than what they would otherwise be.
> 
> Guys, let's all keep in mind that we are working on a solution for this that is more robust so we want to do things that are low risk and high payoff now and are not too complex.

So, do you plan to leave LCS as it is and bootstap with LCS won't be improved by this patch?

glommer:
LCS still benefits, as far as I understand, because we will wait more before moving SSTables to the invariant and will move more at the time. It just doesn't benefit as much as it could because keeping the invariant is expensive.

@asias Also indeed I said any number is "good", in the sense that any number is valid, but the larger the number the more SSTables may be left uncompacted when the reads start.

raphaelsc:
On Fri, Mar 13, 2020 at 11:45 AM Glauber Costa <notifications@github.com>
wrote:

> LCS still benefits, as far as I understand, because we will wait more
> before moving SSTables to the invariant and will move more at the time. It
> just doesn't benefit as much as it could because keeping the invariant is
> expensive.
>
Actually LCS does L0->L1 compaction as soon as it has data in level which
size is greater than max_sstable_size_in_mb, min threshold is not taken
into account[1]. If level 0 falls behind, meaning it has more than 32
sstables, then LCS switches to STCS mode which takes min threshold into
account.

[1] I wonder why LCS doesn't wait for min_threshold sstables in L0 before
compacting them with L1. That could increase read amplification a little
bit (additional min_threshold-1 sstables to consider) but would save tons
of write amplification. Right now, we'd recompact the whole L1 every time
only one new L0 sstable is introduced.

glommer:
That's good to know, but still that doesn't change my opinion about the code piece you pointed me to. We can't risk getting to read phase with hundreds of SSTables piled up.

asias:
Ping.

@raphaelsc if this looks good to me, please approve.

glommer:
Maintainer ping

raphaelsc:
@glommer, @asias could it happen that a bootstrapping node will store redundant data coming from multiple nodes across different sstables? I am afraid about transient increased space amplification, but actually it's a node being bootstrapped so I think it doesn't matter at all. Each size tier will have 4 times more sstables but if the way streaming for bootstrapping, causes no data redundancy stored across different sstables, read amplification will not be an issue because of bloom filter.

asias:
> @glommer, @asias could it happen that a bootstrapping node will store redundant data coming from multiple nodes across different sstables? I am afraid about transient increased space amplification, but actually it's a node being bootstrapped so I think it doesn't matter at all. Each size tier will have 4 times more sstables but if the way streaming for bootstrapping, causes no data redundancy stored across different sstables, read amplification will not be an issue because of bloom filter.

The bootstrapping node does not stream redundant data from multiple nodes. A token range of data comes from only one source node.

asias:
Please add the issue number this PR fixes.

glommer:
I have added a Fixes tag and forced pushed a clean master rebase

Maintainer ping @penberg @nyh randomly selected

database.hh
```diff
@@ -870,7 +879,20 @@ public:
     }
 
     bool compaction_enforce_min_threshold() const {
-        return _config.compaction_enforce_min_threshold;
+        return _config.compaction_enforce_min_threshold || _is_receiving_streaming;
+    }
+
+    unsigned min_compaction_threshold() {
+        // During receiving stream, (bootstrap or replace) the less we
```
avikivity: Decommission is also receiving streaming.

database.hh
```diff
@@ -870,7 +879,20 @@ public:
...
+        // compact the faster streaming is. There are no readers so it is
+        // fine to be less aggressive with compactions as long as we don't
+        // ignore them completely (this could create a problem for when
+        // streaming ends)
+        if (_is_receiving_streaming) {
+            return 16;
```
avikivity: Should be: std::max(schema()->min_compaction_threshold(), 16)

avikivity:
Looks ok, left a comment.

database.hh
glommer: It does, but it also receive reads that can be negatively affected by this.

For now I'd like to tackle just boostrap, where the risk is low.

database.hh
glommer: indeed, will fix.

database.hh
glommer: I have changed the names to reflect the fact that this is only used for bootstrap and replace.

glommer:
@avikivity force-pushed.

glommer:
@avikivity I have addressed your comments and force pushed a version that is more explicit about the fact that this is done for bootstrap only, and respects min_threshold in case it is larger than 16 already.

Let me know if there is anything else needed

glommer:
@avikivity anything else you would like to see here ?

raphaelsc:
On Sun, Mar 22, 2020, 8:28 AM Avi Kivity <notifications@github.com> wrote:

>
> In database.hh
> <https://github.com/scylladb/scylla/pull/6007#discussion_r396082545>:
>
> > @@ -870,7 +879,20 @@ public:
>      }
>
>      bool compaction_enforce_min_threshold() const {
> -        return _config.compaction_enforce_min_threshold;
> +        return _config.compaction_enforce_min_threshold || _is_receiving_streaming;
> +    }
> +
> +    unsigned min_compaction_threshold() {
> +        // During receiving stream, (bootstrap or replace) the less we
> +        // compact the faster streaming is. There are no readers so it is
> +        // fine to be less aggressive with compactions as long as we don't
> +        // ignore them completely (this could create a problem for when
> +        // streaming ends)
> +        if (_is_receiving_streaming) {
> +            return 16;
>
> Should be: std::max(schema()->min_compaction_threshold(), 16)
>
Should we make it respect max threshold given that 16 works like a hint?
perhaps change it to max(min_c_t, max_c_t / 2)

glommer:
if the user has set a higher than 16 min_threshold, I don't see why we wouldn't respect it.

raphaelsc:
On Wed, Mar 25, 2020, 11:52 AM Glauber Costa <notifications@github.com>
wrote:

> if the user has set a higher than 16 min_threshold, I don't see why we
> wouldn't respect it.
>
If somebody sets a max threshold lower than 16, unlikely but possible, so
what about something like min(max(min threshold, 16), max threshold)?

glommer:
What's the goal with that?
This is only used during bootstrap, I can't see why we'd be bothered with max_threshold at all

raphaelsc:
On Wed, Mar 25, 2020, 12:01 PM Glauber Costa <notifications@github.com>
wrote:

> What's the goal with that?
> This is only used during bootstrap, I can't see why we'd be bothered with
> max_threshold at all
>
I was only afraid a strategy could misbehave if min threshold > max
threshold, even if temporarily. But nothing specific comes to mind.

glommer:
It's a fair concern, so I have just tested this.

What happens if max_threshold is lower than 16 is that we end up compacting `max_threshold` sstables instead.

I think it's a sane behavior and I see no reason to change the patch because of that.

glommer:
ping maintainer.

Raphael's concern is addressed (by testing, not code change)

database.hh
```diff
@@ -870,7 +879,19 @@ public:
...
+        return _config.compaction_enforce_min_threshold || _is_bootstrap_or_replace;
+    }
+
+    unsigned min_compaction_threshold() {
+        // During receiving stream operations, the less we compact the faster streaming is. For
+        // bootstrap and replace thereThere are no readers so it is fine to be less aggressive with
+        // compactions as long as we don't ignore them completely (this could create a problem for
+        // when streaming ends)
+        if (_is_bootstrap_or_replace) {
+            return std::max(schema()->min_compaction_threshold(), 16);
```
avikivity: I think @raphaelsc is right that we should consider max_threshold here. A reason to set it is if we have very large rows and therefore can run out of memory; consider 32MiB rows, with 16 sstables we need 512MiB per compaction.

database.hh
```diff
@@ -870,7 +879,19 @@ public:
     }
 
     bool compaction_enforce_min_threshold() const {
-        return _config.compaction_enforce_min_threshold;
+        return _config.compaction_enforce_min_threshold || _is_bootstrap_or_replace;
+    }
+
+    unsigned min_compaction_threshold() {
```
avikivity: This is kind of wonky. For the purpose of the config variable compaction_enforce_min_threshold, we don't touch the min_compaction_threshold() and set/clear a flag whether to ignore it or not. But for bootstrap we do touch the variable, and set the "ignore" flag to false.

It's not easy to simplify, and now it's going to be even harder.

database.hh
glommer: I still don't see what the problem is.
His concern was that "something wrong" (unspecified) would happen if max_threshold was lower than min_threshold. I tested that, and in that case max_threshold is still respected: we compact a `max_threshold` set.

So if max_threshold is higher than min_threshold, it does what everyone expects and caps the amount of SSTables in the set.

If max_threshold is lower than min_threshold, it is still respected since that is how many SSTables we will compact. Whichever way, max_threshold is respected.

So what do you have in mind here? Are you asking to make 16 even lower when max_threshold is lower?

database.hh
glommer: It is not that hard to simplify.
All we have to do is delete this code altogether when offstrategy is ready.

I agree this is a bit weird for now, but after your comment I am not left to guess again what is on your mind.
- Are you against this patch because of that?
- Would you like to pursue a different strategy?
- Are you just saying that is ugly but there's nothing to do?

database.hh
avikivity: Yes, is (say) max_threshold was set to 4 to avoid problems with 32Gib rows.

database.hh
avikivity: I'm saying it's making things worse but as it's "temporary" we can live with it.

database.hh
glommer: Ok, I will use max_threshold / 2 as a goal instead of 16.

glommer: I force-pushed a version with this

avikivity: And if max_threshold == 2, you will not make progress.

I suggest min(16, max_threshold)

asias:
The commit log in version bdd39a6 does not match what the code does:  "This patch sets the minimum threshold to 16".

database.hh
glommer: Ok, will change.

setting the target to `min(16, max_threshold)`, and then using that target as `min_threshold 
= max(target, min_threshold)`

This is only ill-defined if `max_threshold < min_threshold`. However that is not a valid configuration (although we don't enforce it), and as I tested, all that means is that max_threshold becomes both min and max thresholds.

An interesting exercise to the reader is to set `min_threshold` to 1 and see how fun that is.

glommer:
force-pushed, with the change suggested by avi to do `target = min(16, max_threshold)`

sstables/leveled_manifest.hh
```diff
@@ -49,6 +49,7 @@
 #include <boost/range/algorithm/partial_sort.hpp>
 
 class leveled_manifest {
+    table& _table;
```
avikivity, line 52: This is de-modularizing. We now have compaction manifests knowing about tables. This is spaghettifing the code.

I'll let it go this time but I'm sure I'll come to regret it as someone will have to disentangle it.

database.hh
```diff
@@ -870,7 +879,20 @@ public:
...
+        // During receiving stream operations, the less we compact the faster streaming is. For
+        // bootstrap and replace thereThere are no readers so it is fine to be less aggressive with
+        // compactions as long as we don't ignore them completely (this could create a problem for
+        // when streaming ends)
+        if (_is_bootstrap_or_replace) {
+            auto target = std::min(schema()->max_compaction_threshold(), 16);
+            return std::max(schema()->min_compaction_threshold(), target);
```
avikivity, line 892: Note you could have simplified this with std::clamp().

avikivity:
Queued as 87dd23db0

sstables/leveled_manifest.hh
glommer: That sucks indeed. However I think that making the schema be aware of whether or not we're bootstrapping would be much worse.

Hopefully we'll be able to revert this entire patch soon, so nobody will have to disentangle anything.

database.hh
glommer: I was not aware of it. Thanks, Avi.

PR #6019, opened 2020-03-13
When altering table, keep old values of unchanged extensions

kbr-:
When the user performed
```
alter ks.t with compaction = {...}
```
the values of most other options, which were not specified in the
statement, e.g. compression, were left unchanged. That wasn't true for
extension options however: for example, the "cdc" option was removed.

This commit fixes the behavior to keep the old values of extension
options not specified in the alter statement.

Fixes https://github.com/scylladb/scylla/issues/6015.

### DIFF
--- a/cql3/statements/cf_prop_defs.cc
+++ b/cql3/statements/cf_prop_defs.cc
@@ -290,6 +290,10 @@ void cf_prop_defs::apply_to_builder(schema_builder& builder, schema::extensions_
         cfm.caching(cachingOptions);
 #endif
 
+    // for extensions that are not altered, keep the old ones
+    auto& old_exts = builder.get_extensions();
+    schema_extensions.insert(old_exts.begin(), old_exts.end());
+
     builder.set_extensions(std::move(schema_extensions));
 }
 
--- a/schema_builder.hh
+++ b/schema_builder.hh
@@ -193,6 +193,9 @@ public:
         _raw._extensions = std::move(exts);
         return *this;
     }
+    const schema::extensions_map& get_extensions() const {
+        return _raw._extensions;
+    }
     schema_builder& set_compaction_strategy(sstables::compaction_strategy_type type) {
         _raw._compaction_strategy = type;
         return *this;


cql3/statements/cf_prop_defs.cc
```diff
@@ -290,6 +290,10 @@ void cf_prop_defs::apply_to_builder(schema_builder& builder, schema::extensions_
         cfm.caching(cachingOptions);
 #endif
 
+    // for extensions that are not altered, keep the old ones
+    for (auto& old_ext: builder.get_extensions()) {
+        schema_extensions.insert(old_ext);
```
haaawk: The comment says that only extensions that are not altered should be kept but I don't understand how is it determined which extension has been changed and which not.

kbr-: Those that are in `schema_extensions` are changed by the statement.
The old values are in `builder.get_extensions()`.

haaawk: aren't the code overriding the changed extensions then? or will insert fail if the element is already in the collection?

haaawk: OK. It seems to work. Why not schema_extensions.insert(builder.get_extensions().begin(), builder.get_extensions().end())?

cql3/statements/cf_prop_defs.cc
```diff
@@ -290,6 +290,10 @@ void cf_prop_defs::apply_to_builder(schema_builder& builder, schema::extensions_
         cfm.caching(cachingOptions);
 #endif
 
+    // for extensions that are not altered, keep the old ones
+    auto& old_exts = builder.get_extensions();
+    schema_extensions.insert(old_exts.begin(), old_exts.end());
```
nyh: What about extensions that *were* altered, does this do this alteration correctly?

I won't insist now, but a general request for this kind of patch: please add a test (or mention some pre-existing test which tests this). This will ensure that you test all the cases, and also tell the reviewer which cases you remembered to check, and which you forgot to check.

haaawk: I asked exactly the same thing @nyh. the extensions that were altered are already in schema_extensions and insert to a map fails if there is already an element under a given key so this code won't override altered extensions.

nyh: Oh, I didn't see your question (I guess the code changed so your comment disappeared from this version).
I guessed this might be the explanation, but I'm still very unclear about the details: Is the only type of alteration completely replacing the old extension - will there never be a type of "merging" needed? What about *removing* an extension (e.g., previously you had CDC and now you want to remove it) - how will this work? 
If I saw tests that tested these cases, I wouldn't need to doubt that they are handled correctly, but without such tests, I have to raise doubts :-(

haaawk: Those are fair questions. Let's wait for @kbr- to answer them.

kbr-: >  Is the only type of alteration completely replacing the old extension - will there never be a type of "merging" needed?

We do the same thing with extension options (such as `cdc`) as with other table options (`compression`, `compaction`, etc.) do: when you set the option to something, you completely replace what was before. E.g.:
```
alter table ... with compression = {...}
```
replaces the `compression` option that was used before with whatever's in `{...}`.

We could add some merging, but that's a new feature which could be implemented in the future if people express the desire to have it.

> What about removing an extension (e.g., previously you had CDC and now you want to remove it) - how will this work?

You can't remove `compression`, you can't remove `compaction` (whatever you set it to, `describe table` will show it), and similarly, you won't be able to remove `cdc`. But you can disable cdc by e.g. using `with cdc = {}`.

I'll try adding some tests tomorrow.

nyh:
Thanks. Committed to next (without a merge, so github doesn't know).
You can (and perhaps should) add tests later as a followup patch.

PR #6026, opened 2020-03-16
redis: add lolwut command

syuu1228:
Add lolwut command that shows redis version and ascii art.

see: https://redis.io/commands/lolwut

### DIFF
--- a/redis/commands.cc
+++ b/redis/commands.cc
@@ -30,6 +30,7 @@
 #include "redis/options.hh"
 #include "redis/query_utils.hh"
 #include "redis/mutation_utils.hh"
+#include "redis/lolwut.hh"
 
 namespace redis {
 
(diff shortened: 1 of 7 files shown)


redis/commands.cc
```diff
@@ -129,6 +130,48 @@ future<redis_message> echo::execute(service::storage_proxy&, redis::redis_option
...
+    int squares_per_col = 12;
+    try {
+        if (req.arguments_size() >= 1) {
+            cols = std::stoi(std::string(reinterpret_cast<const char*>(req._args[0].data()), req._args[0].size()));
+            if (cols < 1) {
+                cols = 1;
+            } else if (cols > 1000) {
+                cols = 1000;
+            }
+        }
```
avikivity: You can use cols = std::clamp(cols, 1, 1000) to simplify

redis/commands.cc
```diff
@@ -129,6 +130,48 @@ future<redis_message> echo::execute(service::storage_proxy&, redis::redis_option
...
+        if (req.arguments_size() >= 2) {
+            squares_per_row = std::stoi(std::string(reinterpret_cast<const char*>(req._args[1].data()), req._args[1].size()));
+            if (squares_per_row < 1) {
+                squares_per_row = 1;
+            } else if (squares_per_row > 200) {
+                squares_per_row = 200;
+            }
```
avikivity: Here and below too.

redis/commands.cc
```diff
@@ -129,6 +130,48 @@ future<redis_message> echo::execute(service::storage_proxy&, redis::redis_option
...
+            squares_per_col = std::stoi(std::string(reinterpret_cast<const char*>(req._args[2].data()), req._args[2].size()));
+            if (squares_per_col < 1) {
+                squares_per_col = 1;
+            } else if (squares_per_col > 200) {
+                squares_per_col = 200;
+            }
+        }
+    }
+    catch (...) {
+        throw wrong_arguments_exception(1, req.arguments_size(), req._command);
```
avikivity: We use `} catch` in a single line in our coding style.

redis/lolwut.cc
```diff
@@ -0,0 +1,202 @@
...
+#include "redis/lolwut.hh"
+#include "redis/version.hh"
+
+using namespace seastar;
+
+namespace redis {
+
+class canvas {
+    const int _width, _height;
+    char *_pixels;
```
avikivity: Better to use `std::unique_ptr<char[]> _pixels;`:

 - you get the destructor defined
 - with the current canvas class, it it is copied, then the destructor will run on both copies resulting in double-free.

redis/lolwut.cc
```diff
@@ -0,0 +1,202 @@
...
+            if (e2 > -dy) {
+                err -= dy;
+                x1 += sx;
+            }
+            if (e2 < dx) {
+                err += dx;
+                y1 += sy;
+            }
+        }
+    }
```
avikivity: Hey, Bresenham's algorithm!

redis/lolwut.cc
```diff
@@ -0,0 +1,202 @@
...
+                    byte |= (1 << 5);
+                }
+                if (get_pixel(x, y + 3)) {
+                    byte |= (1 << 6);
+                }
+                if (get_pixel(x + 1, y + 3)) {
+                    byte |= (1 << 7);
+                }
+                translate_pixels_group(byte, unicode);
+                txt.append(unicode, 3);
```
avikivity: This has quadratic behavior. Better to allocate txt up-front and copy to the correct location.

redis/lolwut.cc
```diff
@@ -0,0 +1,202 @@
...
+            }
+            if (y != _height - 1) {
+                const bytes nr {"\n"};
+                txt += nr;
+            }
+        }
+        return txt;
+    }
+
```
avikivity: If this command proves popular we may need to run it in a thread and add maybe_yield calls, but for now we can leave it as is.

redis/lolwut.cc
avikivity: Another option is to accept a bytes_ostream parameter which knows how to grow without quadratic behavior.

syuu1228:
@avikivity new version pushed, please review again.

Issue #6022, opened 2020-03-15
member call on null pointer of type 'const struct schema' during decommission_node_during_mv_insert_4_nodes_test


bhalevy:
Seen in https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug-random/91/artifact/logs-debug.2/1584227523710_materialized_views_test.TestMaterializedViews.decommission_node_during_mv_insert_4_nodes_test/node2.log
Scylla version 6b747f4673c0bce50248c3a8208b9b854cda499f
```
INFO  2020-03-14 23:12:01,162 [shard 0] repair - decommission_with_repair: finished with keyspace=keyspace1, leaving_node=127.0.95.2, nr_ranges=762, nr_ranges_synced=762, nr_ranges_skipped=0
INFO  2020-03-14 23:12:01,221 [shard 0] repair - decommission_with_repair: started with keyspace=ks, leaving_node=127.0.95.2, nr_ranges=762
INFO  2020-03-14 23:12:01,439 [shard 0] repair - repair id 3 to sync data for keyspace=ks, status=started
repair/repair.cc:662:36: runtime error: member call on null pointer of type 'const struct schema'
Aborting on shard 0.
Backtrace:
  /jenkins/workspace/scylla-master/dtest-debug-random/scylla/.ccm/scylla-repository/6b747f4673c0bce50248c3a8208b9b854cda499f/scylla//opt/scylladb/libreloc/libasan.so.5+0x000000000006ce7f
  0x00000000156c80d4
  0x00000000156a66b8
  0x000000001557ef53
  0x000000001557f0b1
  0x00000000155d271a
  0x000000001560b697
  0x000000001560b718
  0x00007f1f2caddb1f
  /jenkins/workspace/scylla-master/dtest-debug-random/scylla/.ccm/scylla-repository/6b747f4673c0bce50248c3a8208b9b854cda499f/scylla//opt/scylladb/libreloc/libc.so.6+0x000000000003c624
  /jenkins/workspace/scylla-master/dtest-debug-random/scylla/.ccm/scylla-repository/6b747f4673c0bce50248c3a8208b9b854cda499f/scylla//opt/scylladb/libreloc/libc.so.6+0x00000000000258d8
[...]

PR #6028, opened 2020-03-17
repair: Handle keyspace with zero table

asias:
The following error was seen in
materialized_views_test.py:TestMaterializedViews.decommission_node_during_mv_insert_4_nodes_test

INFO [shard 0] repair - repair id 3 to sync data for
keyspace=ks, status=started repair/repair.cc:662:36: runtime error: member call
on null pointer of type 'const struct schema'
Aborting on shard 0.

The problem is in the test a keyspace was created without creating any
table. In get_partitioner_for_tables, we access nullptr when no table is
present.

	schema_ptr last_s;
	for (auto t: tables) {
	    // set last_s
	}
	last_s->get_partitione()
To fix:
1) Skip the repair in sync_data_using_repair if there is no table in the keyspace
2) Throw if no schema_ptr is found in get_partitioner_for_tables. Be defensive.

After:
INFO [shard 0] repair - decommission_with_repair: started with keyspace=ks, leaving_node=127.0.0.2, nr_ranges=744
INFO [shard 0] repair - repair id 3 to sync data for keyspace=ks, status=started
WARN [shard 0] repair - repair id 3 to sync data for keyspace=ks, no table in this keyspace
INFO [shard 0] repair - repair id 3 completed successfully
INFO [shard 0] repair - repair id 3 to sync data for keyspace=ks, status=succeeded

Tests: materialized_views_test.py:TestMaterializedViews.decommission_node_during_mv_insert_4_nodes_test
Fixes: #6022

### DIFF
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -659,6 +659,10 @@ get_partitioner_for_tables(seastar::sharded<database>& db, const sstring& keyspa
         }
         last_s = std::move(s);
     }
+    if (!last_s) {
+        throw std::runtime_error(format("Failed to find partitioner for keyspace={}, tables={}, no table in this keyspace",
+                keyspace, names));
+    }
     return last_s->get_partitioner();
 }
 
@@ -1529,6 +1533,10 @@ future<> sync_data_using_repair(seastar::sharded<database>& db,
         rlogger.info("repair id {} to sync data for keyspace={}, status=started", id, keyspace);
         return repair_tracker().run(id, [id, &db, keyspace, ranges = std::move(ranges), neighbors = std::move(neighbors)] () mutable {
             auto cfs = list_column_families(db.local(), keyspace);
+            if (cfs.empty()) {
+                rlogger.warn("repair id {} to sync data for keyspace={}, no table in this keyspace", id, keyspace);
+                return make_ready_future<>();
+            }
             std::vector<future<>> repair_results;
             repair_results.reserve(smp::count);
             for (auto shard : boost::irange(unsigned(0), smp::count)) {


asias:
@haaawk can you please review the change in get_partitioner_for_tables, you added this function.

repair/repair.cc
```diff
@@ -1529,6 +1533,10 @@ future<> sync_data_using_repair(seastar::sharded<database>& db,
         rlogger.info("repair id {} to sync data for keyspace={}, status=started", id, keyspace);
         return repair_tracker().run(id, [id, &db, keyspace, ranges = std::move(ranges), neighbors = std::move(neighbors)] () mutable {
             auto cfs = list_column_families(db.local(), keyspace);
+            if (cfs.empty()) {
+                rlogger.warn("repair id {} to sync data for keyspace={}, no table in this keyspace", id, keyspace);
```
bhalevy: Why is it a warning?
If this can happen under normal circumstances then info or even debug should be enough.

asias: In normal cases, user would probably create at least one table when creating the keyspace. I put it as warn level because I worry the case that the keyspace has zero table because the table was dropped in the middle of the operation, to give the user a warning that this happens. It is up to the user to to decided.

bhalevy: But the user may do that, right?
Do we recommend against it or try to prevent it in any way?

asias: Yes, user may do that. The recommendation is not to perform such operation during the topology changes.

bhalevy: If we recommend against it than a warning is justified. thanks.

bhalevy:
@haaawk ping review please

bhalevy:
@scylladb/scylla-maint please merge

avikivity:
Queued as be1a196988da6872192f838c4c94eec510aa7320

PR #6030, opened 2020-03-17
cdc: add metrics

### LINKED ISSUE #6002: Add metricts for CDC
We need to add some metrics to CDC to be able to monitor it.

The most important is the number of successful/unsuccessful CDC augmentations.

Additionally, the number of successful/unsuccessful preimage selects.

It would be also useful to have number of successful/unsuccessful CDC augmentations that required mutation splitting.

It might be also useful to track the number of CDC augmentations that:
* touch static row
* touch map
* touch set
* touch list
* touch udt
* touch range tombstone
* touch partition delete
* touch row delete

separately for successful and unsuccessful operations

piodul:
Adds CDC-related metrics.

Following counters are added, both for total and failed operations:
- Total number of CDC operations that did/did not perform splitting,
- Total number of CDC operations  that touched a particular mutation part.
- Total number of preimage selects.

Fixes #6002.
Tests: unit(dev, debug)

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -126,6 +126,28 @@ cdc::stats::stats() {
     register_counters(counters_failed, "failed");
 }
 
+cdc::operation_result_tracker::~operation_result_tracker() {
+    auto update_stats = [this] (stats::counters& counters) {
+        if (_details.was_split) {
+            counters.split_count++;
+        } else {
+            counters.unsplit_count++;
+        }
+        counters.touches.apply(_details.touched_parts);
+        if (_details.had_preimage) {
+            counters.with_preimage_count++;
+        }
+        if (_details.had_postimage) {
+            counters.with_postimage_count++;
+        }
+    };
+
+    update_stats(_stats.counters_total);
+    if (_failed) {
+        update_stats(_stats.counters_failed);
+    }
+}
+
 class cdc::cdc_service::impl : service::migration_listener::empty_listener {
     friend cdc_service;
     db_context _ctxt;
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -236,7 +236,7 @@ public:
         }
     }
 
-    future<std::vector<mutation>> augment_mutation_call(
+    future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>> augment_mutation_call(
         lowres_clock::time_point timeout,
         std::vector<mutation>&& mutations,
         tracing::trace_state_ptr tr_state
@@ -1299,7 +1299,7 @@ transform_mutations(std::vector<mutation>& muts, decltype(muts.size()) batch_siz
 
 } // namespace cdc
 
-future<std::vector<mutation>>
+future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>>
 cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, std::vector<mutation>&& mutations, tracing::trace_state_ptr tr_state) {
     // we do all this because in the case of batches, we can have mixed schemas.
     auto e = mutations.end();
@@ -1308,16 +1308,15 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, 
     });
 
     if (i == e) {
-        return make_ready_future<std::vector<mutation>>(std::move(mutations));
+        return make_ready_future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>>(std::make_tuple(std::move(mutations), lw_shared_ptr<cdc::operation_result_tracker>()));
     }
 
     tracing::trace(tr_state, "CDC: Started generating mutations for log rows");
     mutations.reserve(2 * mutations.size());
 
-    return do_with(std::move(mutations), service::query_state(service::client_state::for_internal_calls(), empty_service_permit()),
-            [this, timeout, i, tr_state = std::move(tr_state)] (std::vector<mutation>& mutations, service::query_state& qs) {
-        
-        return transform_mutations(mutations, 1, [this, &mutations, timeout, &qs, tr_state = tr_state] (int idx) mutable {
+    return do_with(std::move(mutations), service::query_state(service::client_state::for_internal_calls(), empty_service_permit()), operation_details{},
+            [this, timeout, i, tr_state = std::move(tr_state)] (std::vector<mutation>& mutations, service::query_state& qs, operation_details& details) {
+        return transform_mutations(mutations, 1, [this, &mutations, timeout, &qs, tr_state = tr_state, &details] (int idx) mutable {
             auto& m = mutations[idx];
             auto s = m.schema();
 
@@ -1345,18 +1344,22 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, 
                 tracing::trace(tr_state, "CDC: Preimage not enabled for the table, not querying current value of {}", m.decorated_key());
             }
 
-            return f.then([trans = std::move(trans), &mutations, idx, tr_state = std::move(tr_state)] (lw_shared_ptr<cql3::untyped_result_set> rs) {
+            return f.then([trans = std::move(trans), &mutations, idx, tr_state = std::move(tr_state), &details] (lw_shared_ptr<cql3::untyped_result_set> rs) {
                 auto& m = mutations[idx];
                 auto& s = m.schema();
+                details.had_preimage |= s->cdc_options().preimage();
+                details.had_postimage |= s->cdc_options().postimage();
                 tracing::trace(tr_state, "CDC: Generating log mutations for {}", m.decorated_key());
                 int generated_count;
                 if (should_split(m, *s)) {
                     tracing::trace(tr_state, "CDC: Splitting {}", m.decorated_key());
+                    details.was_split = true;
                     generated_count = 0;
                     for_each_change(m, s, [&] (mutation mm, api::timestamp_type ts, bytes tuuid, int& batch_no) {
                         auto [mut, parts] = trans.transform(std::move(mm), rs.get(), ts, tuuid, batch_no);
                         mutations.push_back(std::move(mut));
                         ++generated_count;
+                        details.touched_parts.add(parts);
                     });
                 } else {
                     tracing::trace(tr_state, "CDC: No need to split {}", m.decorated_key());
@@ -1366,13 +1369,15 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, 
                     auto [mut, parts] = trans.transform(m, rs.get(), ts, tuuid, batch_no);
                     mutations.push_back(std::move(mut));
                     generated_count = 1;
+                    details.touched_parts.add(parts);
                 }
                 // `m` might be invalidated at this point because of the push_back to the vector
                 tracing::trace(tr_state, "CDC: Generated {} log mutations from {}", generated_count, mutations[idx].decorated_key());
             });
-        }).then([tr_state](std::vector<mutation> mutations) {
+        }).then([this, tr_state, &details](std::vector<mutation> mutations) {
             tracing::trace(tr_state, "CDC: Finished generating all log mutations");
-            return make_ready_future<std::vector<mutation>>(std::move(mutations));
+            auto tracker = make_lw_shared<cdc::operation_result_tracker>(_ctxt._proxy.get_cdc_stats(), details);
+            return make_ready_future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>>(std::make_tuple(std::move(mutations), std::move(tracker)));
         });
     });
 }
@@ -1383,7 +1388,7 @@ bool cdc::cdc_service::needs_cdc_augmentation(const std::vector<mutation>& mutat
     });
 }
 
-future<std::vector<mutation>>
+future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>>
 cdc::cdc_service::augment_mutation_call(lowres_clock::time_point timeout, std::vector<mutation>&& mutations, tracing::trace_state_ptr tr_state) {
     return _impl->augment_mutation_call(timeout, std::move(mutations), std::move(tr_state));
 }
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -25,6 +25,7 @@
 #include <boost/range/irange.hpp>
 #include <seastar/util/defer.hh>
 #include <seastar/core/thread.hh>
+#include <seastar/core/metrics.hh>
 
 #include "cdc/log.hh"
 #include "cdc/generation.hh"
@@ -49,6 +50,7 @@
 #include "concrete_types.hh"
 #include "types/listlike_partial_deserializing_iterator.hh"
 #include "tracing/trace_state.hh"
+#include "stats.hh"
 
 namespace std {
 
@@ -66,6 +68,62 @@ logging::logger cdc_log("cdc");
 
 namespace cdc {
 static schema_ptr create_log_schema(const schema&, std::optional<utils::UUID> = {});
+}
+
+static constexpr auto cdc_group_name = "cdc";
+
+void cdc::stats::parts_touched_stats::register_metrics(seastar::metrics::metric_groups& metrics, std::string_view suffix) {
+    namespace sm = seastar::metrics;
+    auto register_part = [&] (part_type part, sstring part_name) {
+        metrics.add_group(cdc_group_name, {
+                sm::make_total_operations(format("operations_on_{}_performed_{}", part_name, suffix), count[(size_t)part],
+                        sm::description(format("number of {} CDC operations that processed a {}", suffix, part_name)),
+                        {})
+            });
+    };
+
+    register_part(part_type::STATIC_ROW, "static_row");
+    register_part(part_type::CLUSTERING_ROW, "clustering_row");
+    register_part(part_type::MAP, "map");
+    register_part(part_type::SET, "set");
+    register_part(part_type::LIST, "list");
+    register_part(part_type::UDT, "udt");
+    register_part(part_type::RANGE_TOMBSTONE, "range_tombstone");
+    register_part(part_type::PARTITION_DELETE, "partition_delete");
+    register_part(part_type::ROW_DELETE, "row_delete");
+}
+
+cdc::stats::stats() {
+    namespace sm = seastar::metrics;
+
+    auto register_counters = [this] (counters& counters, sstring kind) {
+        const auto split_label = sm::label("split");
+        _metrics.add_group(cdc_group_name, {
+                sm::make_total_operations("operations_" + kind, counters.unsplit_count,
+                        sm::description(format("number of {} CDC operations", kind)),
+                        {split_label(false)}),
+
+                sm::make_total_operations("operations_" + kind, counters.split_count,
+                        sm::description(format("number of {} CDC operations", kind)),
+                        {split_label(true)}),
+
+                sm::make_total_operations("preimage_selects_" + kind, counters.preimage_selects,
+                        sm::description(format("number of {} preimage queries performed", kind)),
+                        {}),
+
+                sm::make_total_operations("operations_with_preimage_" + kind, counters.with_preimage_count,
+                        sm::description(format("number of {} operations that included preimage", kind)),
+                        {}),
+
+                sm::make_total_operations("operations_with_postimage_" + kind, counters.with_postimage_count,
+                        sm::description(format("number of {} operations that included postimage", kind)),
+                        {})
+            });
+
+        counters.touches.register_metrics(_metrics, kind);
+    };
+    register_counters(counters_total, "total");
+    register_counters(counters_failed, "failed");
 }
 
 class cdc::cdc_service::impl : service::migration_listener::empty_listener {
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -1301,7 +1301,14 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, 
                 // iff a batch contains several modifications to the same table. Otoh, batch is rare(?)
                 // so this is premature.
                 tracing::trace(tr_state, "CDC: Selecting preimage for {}", m.decorated_key());
-                f = trans.pre_image_select(qs.get_client_state(), db::consistency_level::LOCAL_QUORUM, m);
+                f = trans.pre_image_select(qs.get_client_state(), db::consistency_level::LOCAL_QUORUM, m).then_wrapped([this] (future<lw_shared_ptr<cql3::untyped_result_set>> f) {
+                    auto& cdc_stats = _ctxt._proxy.get_cdc_stats();
+                    cdc_stats.counters_total.preimage_selects++;
+                    if (f.failed()) {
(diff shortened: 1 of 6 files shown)


cdc/log.cc
```diff
@@ -68,6 +69,70 @@ namespace cdc {
...
+    register_part(part_type::STATIC_ROW, part_label("static_row"));
+    register_part(part_type::CLUSTERING_ROW, part_label("clustering_row"));
+    register_part(part_type::MAP, part_label("map"));
+    register_part(part_type::SET, part_label("set"));
+    register_part(part_type::LIST, part_label("list"));
+    register_part(part_type::UDT, part_label("udt"));
+    register_part(part_type::RANGE_TOMBSTONE, part_label("range_tombstone"));
+    register_part(part_type::PARTITION_DELETE, part_label("partition_delete"));
+    register_part(part_type::ROW_DELETE, part_label("row_delete"));
+}
```
piodul: I used labels here to differentiate counters that track the "touches" of different parts of mutation. I admit that the value they sum up to doesn't have much sense in itself, so I'm open to ideas on how to improve the representation of those counters.

cdc/log.cc
```diff
@@ -68,6 +69,70 @@ namespace cdc {
...
+                sm::make_histogram("preimage_select_latency",
+                        [&histograms] { return histograms.preimage_select_latency.get_histogram(16, 20); },
+                        sm::description("latency of preimage queries"),
+                        {succeeded_instance})
+            });
+    };
+
+    const auto succeeded_label = sm::label("succeeded");
+    register_latencies(histograms_succeeded, succeeded_label(false));
+    register_latencies(histograms_failed, succeeded_label(true));
```
piodul: I observed that we have a convention that for a given operation, we define two metrics: total and failed. I was not sure it made sense for histograms, so I created "successful" and "failed" histograms. I can convert the "successful" into "total", if that's OK.

cdc/log.cc
```diff
@@ -68,6 +69,70 @@ namespace cdc {
...
+    register_counters(counters_total, "total");
+    register_counters(counters_failed, "failed");
+
+    auto register_latencies = [this] (histograms& histograms, sm::label_instance succeeded_instance) {
+        _metrics.add_group(cdc_group_name, {
+                sm::make_histogram("ops_latency",
+                        [&histograms] { return histograms.write_latency.get_histogram(16, 20); },
+                        sm::description("latency of combined write operations, one operation includes "
+                                "write of all cdc rows generated from single mutation"),
+                                {succeeded_instance}),
```
piodul: Should the time spent in preimage select and `transformer::transform` be included here as well?

cdc/stats.hh
```diff
@@ -0,0 +1,118 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <seastar/core/metrics.hh>
```
amnonh: do not include metrics.hh in your header file, use metrics_registration.hh instead

metrics.hh should be included in your source file

cdc/log.cc
```diff
@@ -68,6 +69,70 @@ namespace cdc {
 static schema_ptr create_log_schema(const schema&, std::optional<utils::UUID> = {});
 }
 
+static constexpr auto cdc_group_name = "cdc";
+
+void cdc::stats::parts_touched_stats::register_metrics(seastar::metrics::metric_groups& metrics, sstring suffix) {
+    namespace sm = seastar::metrics;
+    auto register_part = [&] (part_type part, sm::label_instance part_instance) {
+        metrics.add_group(cdc_group_name, {
+                sm::make_total_operations("components_touched_" + suffix, count[(size_t)part],
```
amnonh: isn't the suffix equal to the label?
if so, it's redundant.
you should either differ by name or by label, not both

cdc/log.cc
```diff
@@ -68,6 +69,70 @@ namespace cdc {
...
+                        sm::description("number of times a mutation part type have been touched during cdc mutation augmentation"),
```
amnonh: the description is not clear to me, if that's for internal use only, it's fine, if you think users should look at it, it should be clearer.

cdc/log.cc
amnonh: Again, ops name is misleading. 
I don't think the description is clear

cdc/log.cc
piodul: They are different. The suffix is either "total" or "failed". The label determines which part of the mutation was touched (e.g. "static_row", "row_tombstone", "udt", etc.).

service/storage_proxy.cc
```diff
@@ -330,6 +331,43 @@ class cas_mutation : public mutation_holder {
     }
 };
 
+// Tracks the success of a set of CDC mutations generated from the same mutation
```
nyh: Is it really "tracks the success"? It sounds like a when_all_succeeds() could have done this, and here you have much more - _was_split, _touches and all sorts of other stuff... So I don't think I understand the purpose of this structure, or again (and I'm repeating myself...) why the code can't just update the statistics immediately when something happens (e.g., a split) instead of remembering _was_split.

service/storage_proxy.cc
```diff
@@ -330,6 +331,43 @@ class cas_mutation : public mutation_holder {
...
+        , _touches(touches)
+        , _lc(lc)
+    {}
+
+    ~cdc_augmentation_tracker() {
+        auto& histograms = _had_error ? _stats.histograms_failed : _stats.histograms_succeeded;
+        histograms.write_latency.add(_lc.stop().latency(), histograms.write_latency._count + 1);
+        if (_had_error) {
+            if (_was_split) {
+                _stats.counters_failed.split_count++;
```
nyh: Here is an example of what I asked above... Something sets _was_split, and now we do split_count++. Why couldn't the split function just do the split_count++ itself?

service/storage_proxy.cc
```diff
@@ -1245,7 +1245,12 @@ future<> paxos_response_handler::learn_decision(paxos::proposal decision, bool a
                 auto meta = std::move(std::get<1>(t));
                 // Pick only the CDC ("augmenting") mutations
                 auto range = boost::make_iterator_range(mutations.end() - meta.total_cdc_mutations, mutations.end());
-                return _proxy->mutate_internal(std::move(range), _cl_for_learn, false, tr_state, _permit, _timeout);
+                // Need to adjust ranges in the metadata, because we are going to skip non-CDC mutations
+                for (auto& range : meta.generated_ranges) {
```
nyh: Please don't use the variable name "range" here, it's very confusing because you have a different variable "range" just one line up!

service/storage_proxy.cc
```diff
@@ -1245,7 +1245,12 @@ future<> paxos_response_handler::learn_decision(paxos::proposal decision, bool a
...
+                    range.begin -= meta.total_cdc_mutations;
+                    range.end -= meta.total_cdc_mutations;
+                }
+                return _proxy->mutate_internal(std::move(range), _cl_for_learn, false, tr_state, _permit, _timeout, std::move(meta));
```
nyh: I have to say its strange that mutate_internal takes some "meta" which I guess is augmentation_meta. Would have been nice if we didn't have to add yet another argument to mutate_internal().

cdc/log.cc
```diff
@@ -49,6 +49,7 @@
 #include "concrete_types.hh"
 #include "types/listlike_partial_deserializing_iterator.hh"
 #include "tracing/trace_state.hh"
+#include "stats.hh"
```
haaawk: I'm commenting here because I can't on the commit message:
s/cdc_service it is also going modify it/cdc_service it is also going to modify it/

cdc/stats.hh
```diff
@@ -0,0 +1,96 @@
...
+        part_type::SET,
+        part_type::LIST,
+        part_type::UDT,
+        part_type::RANGE_TOMBSTONE,
+        part_type::PARTITION_DELETE,
+        part_type::ROW_DELETE
+    >>;
+
+    struct parts_touched_stats {
+        uint64_t count[(size_t)part_type::MAX] = {};
```
haaawk: Can we use std::array?

cdc/stats.hh
```diff
@@ -0,0 +1,96 @@
...
+                count[(size_t)idx]++;
+            }
+        }
+
+        void register_metrics(seastar::metrics::metric_groups& metrics, sstring suffix);
+    };
+
+    struct counters {
+        uint64_t unsplit_count = 0;
+        uint64_t split_count = 0;
```
haaawk: It would be also interesting to know the number of operations that had pre/post image

cdc/log.hh
```diff
@@ -43,6 +43,7 @@
 #include "tracing/trace_state.hh"
 #include "cdc_options.hh"
 #include "utils/UUID.hh"
+#include "stats.hh"
```
haaawk: Why is this include needed here if stats are only used in log.cc?

cdc/log.cc
```diff
@@ -156,7 +156,7 @@ class cdc::cdc_service::impl : service::migration_listener::empty_listener {
         }
     }
 
-    future<std::tuple<std::vector<mutation>, result_callback>> augment_mutation_call(
+    future<std::vector<mutation>> augment_mutation_call(
```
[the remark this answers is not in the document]
piodul: Submitted separately: https://github.com/scylladb/scylla/pull/6037

cdc/log.cc
haaawk: In my opinion, operations_performed would be better than components_touched

cdc/log.cc
haaawk: I'm not a big fan of 'touched'. What does it mean? If an operation changes static row and two clustering rows how many touches are there? I would expect to count one static row operation and one clustering row operation.

cdc/log.cc
```diff
@@ -1316,20 +1325,25 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
...
                 } else {
                     tracing::trace(tr_state, "CDC: No need to split {}", m.decorated_key());
                     int batch_no = 0;
                     auto ts = find_timestamp(*s, m);
                     auto tuuid = timeuuid_type->decompose(generate_timeuuid(ts));
-                    mutations.push_back(trans.transform(m, rs.get(), ts, tuuid, batch_no));
+                    auto [mut, parts] = trans.transform(m, rs.get(), ts, tuuid, batch_no);
+                    mutations.push_back(std::move(mut));
                     generated_count = 1;
+                    touched_parts = parts;
```
haaawk: touched_parts = std::move(parts) ?

cdc/log.cc
```diff
@@ -1308,25 +1308,33 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
...
 
             transformer trans(_ctxt, s);
+            auto& stats = _ctxt._proxy.get_cdc_stats();
 
             auto f = make_ready_future<lw_shared_ptr<cql3::untyped_result_set>>(nullptr);
             if (s->cdc_options().preimage() || s->cdc_options().postimage()) {
                 // Note: further improvement here would be to coalesce the pre-image selects into one
                 // iff a batch contains several modifications to the same table. Otoh, batch is rare(?)
                 // so this is premature.
+                utils::latency_counter lc;
```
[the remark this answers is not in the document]
haaawk: It's interesting to see how log CDC takes. If it takes longer than other parts then it's a bottle neck. Plus it's useful for performance improvements of CDC itself.

cdc/log.cc
```diff
@@ -1327,13 +1329,15 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
...
                 auto& m = mutations[idx];
                 auto& s = m.schema();
                 tracing::trace(tr_state, "CDC: Generating log mutations for {}", m.decorated_key());
                 int generated_count;
+                size_t size_before_appends = mutations.size();
+                const bool do_split = should_split(m, *s);
                 stats::part_type_set touched_parts;
-                if (should_split(m, *s)) {
+                if (do_split) {
                     stats.counters_total.split_count++;
```
haaawk: Do we still need this? Can't generate_range take care of this as well?

service/storage_proxy.cc
```diff
@@ -1192,16 +1192,12 @@ future<> paxos_response_handler::learn_decision(paxos::proposal decision, bool a
...
                 // Pick only the CDC ("augmenting") mutations
-                mutations.erase(std::remove_if(mutations.begin(), mutations.end(), [base_tbl_id = std::move(base_tbl_id)] (const mutation& v) {
-                    return v.schema()->id() == base_tbl_id;
-                }), mutations.end());
-                if (mutations.empty()) {
-                    return make_ready_future<>();
-                }
-
-                return _proxy->mutate_internal(std::move(mutations), _cl_for_learn, false, tr_state, _permit, _timeout);
+                auto range = boost::make_iterator_range(mutations.end() - meta.total_cdc_mutations, mutations.end());
```
haaawk: Assuming the order in mutations is bad. Let's keep the old code which was safer.

service/storage_proxy.cc
haaawk: Because the place that splits does not know whether the operation will succeed or fail.

cdc/log.cc
piodul: @haaawk Your understanding is correct, this counter will be incremented as you described.

I'm not sure if `operations_performed` is a good name by itself here. There will be many `operations_performed` metrics, differing by the label that determines which part of the mutation was "touched". If we sum them all up, a number greater then the actual total number of operations performed will appear.

Perhaps we should abandon labels and give different name for those counters, e.g. `operations_on_static_row_performed`, `operations_on_clustering_row_performed`?

haaawk: It might be a solution. Labels may not be a good fit here.

cdc/log.cc
piodul: It should not matter here, `touched_parts` is of `enum_set` type, which is essentially a bit-set that uses a single `size_t` under the hood. It is very cheap to copy.

haaawk: OK. Makes sense.

piodul:
v2 (big update):

- Metrics for "cdc operations" are more coarse now. Now, `augment_mutation_call` returns a tracker object that is attached to all response handlers of cdc mutations generated by the same `augment_mutation_call` call. As a result of this change, `augmentation_metadata` is removed.
- Removed histograms - they need more discussion, and will be added in a later PR.
- Made metric names and descriptions (hopefully) more clear.

service/storage_proxy.cc
piodul: Done, new version does not depend on the order.

cdc/log.cc
piodul: Now both "failed" and "total" metrics are updated after all cdc mutation writes finish.

cdc/log.hh
piodul: Removed

cdc/stats.hh
piodul: Added counters for preimage - I'll add counters for post image shortly

cdc/stats.hh
piodul: Yes, we can! Now we do.

service/storage_proxy.cc
piodul: I don't know how to avoid it other than by copy pasting it and adjusting slightly for the purpose of CDC... Do you have any ideas how we could avoid cluttering it with arguments and introducing copy-paste at the same time?

cdc/log.cc
```diff
@@ -799,18 +799,21 @@ class transformer final {
 
     // TODO: is pre-image data based on query enough. We only have actual column data. Do we need
     // more details like tombstones/ttl? Probably not but keep in mind.
-    mutation transform(const mutation& m, const cql3::untyped_result_set* rs, api::timestamp_type ts, bytes tuuid, int& batch_no) const {
+    std::tuple<mutation, stats::part_type_set> transform(const mutation& m, const cql3::untyped_result_set* rs, api::timestamp_type ts, bytes tuuid, int& batch_no) const {
```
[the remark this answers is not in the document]
piodul: If writing of mutations generated by transform() fails, we will need to update the counters of "failed" operations. At this point we don't know if writing of the mutations that are generated will succeed, so this information needs to be stored, and the update must be deferred after we know the result of all of the writes.

cdc/stats.hh
piodul: Done

cdc/log.cc
```diff
@@ -68,6 +69,70 @@ namespace cdc {
...
+    register_part(part_type::ROW_DELETE, part_label("row_delete"));
+}
+
+cdc::stats::stats() {
+    namespace sm = seastar::metrics;
+
+    auto register_counters = [this] (counters& counters, sstring kind) {
+        const auto split_label = sm::label("split");
+        _metrics.add_group(cdc_group_name, {
+                sm::make_total_operations("ops_" + kind, counters.unsplit_count,
```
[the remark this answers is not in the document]
piodul: The names and descriptions should be more clear now.

piodul:
v3: there are now 3 metrics related to preimage/postimage:
1. number of preimage selects performed,
1. number of CDC operations with preimage enabled,
1. number of CDC operations with postimage enabled.

cdc/stats.hh
piodul: Done, see description for v3

piodul:
Rebased onto master.

ping @haaawk @nyh @amnonh

cdc/log.cc
```diff
@@ -68,6 +70,62 @@ namespace cdc {
 static schema_ptr create_log_schema(const schema&, std::optional<utils::UUID> = {});
 }
 
+static constexpr auto cdc_group_name = "cdc";
+
+void cdc::stats::parts_touched_stats::register_metrics(seastar::metrics::metric_groups& metrics, sstring suffix) {
```
haaawk: Nit: suffix can be taken by const sstring& or string_view, no?

cdc/stats.hh
```diff
@@ -0,0 +1,92 @@
...
+        inline void apply(part_type_set parts_set) {
+            for (part_type idx : parts_set) {
+                count[(size_t)idx]++;
+            }
+        }
+
+        void register_metrics(seastar::metrics::metric_groups& metrics, sstring suffix);
+    };
+
+    struct counters {
```
haaawk: Super nit: I think we can mark it final.

cdc/stats.hh
```diff
@@ -0,0 +1,92 @@
...
+#include <array>
+#include <cstdint>
+#include <seastar/core/metrics_registration.hh>
+#include "enum_set.hh"
+#include "utils/histogram.hh"
+#include "utils/estimated_histogram.hh"
+
+namespace cdc {
+
+struct stats {
```
haaawk: Nit: maybe change struct to class and drop the private: label below.

haaawk: Also mark final.

cdc/log.cc
```diff
@@ -1299,7 +1299,14 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
...
                 // so this is premature.
                 tracing::trace(tr_state, "CDC: Selecting preimage for {}", m.decorated_key());
-                f = trans.pre_image_select(qs.get_client_state(), db::consistency_level::LOCAL_QUORUM, m);
+                f = trans.pre_image_select(qs.get_client_state(), db::consistency_level::LOCAL_QUORUM, m).then_wrapped([this] (future<lw_shared_ptr<cql3::untyped_result_set>> f) {
+                    auto& cdc_stats = _ctxt._proxy.get_cdc_stats();
+                    cdc_stats.counters_total.preimage_selects++;
+                    if (f.failed()) {
+                        cdc_stats.counters_failed.preimage_selects++;
+                    }
+                    return f;
```
haaawk: Does it make sense to return std::move(f) here or will the compiler complain about RVO being impossible then?

cdc/log.cc
```diff
@@ -1320,15 +1330,17 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
                     tracing::trace(tr_state, "CDC: Splitting {}", m.decorated_key());
                     generated_count = 0;
                     for_each_change(m, s, [&] (mutation mm, api::timestamp_type ts, bytes tuuid, int& batch_no) {
-                        mutations.push_back(trans.transform(std::move(mm), rs.get(), ts, tuuid, batch_no));
+                        auto [mut, parts] = trans.transform(std::move(mm), rs.get(), ts, tuuid, batch_no);
```
haaawk: Isn't compiler complain about unused variable parts?

cdc/log.cc
```diff
@@ -1320,15 +1330,17 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
...
+                        mutations.push_back(std::move(mut));
                         ++generated_count;
                     });
                 } else {
                     tracing::trace(tr_state, "CDC: No need to split {}", m.decorated_key());
                     int batch_no = 0;
                     auto ts = find_timestamp(*s, m);
                     auto tuuid = timeuuid_type->decompose(generate_timeuuid(ts));
-                    mutations.push_back(trans.transform(m, rs.get(), ts, tuuid, batch_no));
+                    auto [mut, parts] = trans.transform(m, rs.get(), ts, tuuid, batch_no);
```
haaawk: Same here.

cdc/stats.hh
```diff
@@ -89,4 +89,33 @@ public:
     stats();
 };
 
+// Contains the details on what happened during a CDC operation.
+struct operation_details {
```
haaawk: Super nit: final

cdc/stats.hh
```diff
@@ -89,4 +89,33 @@ public:
...
+    stats::part_type_set touched_parts;
+    bool was_split = false;
+    bool had_preimage = false;
+    bool had_postimage = false;
+};
+
+// This object tracks the lifetime of write handlers related to one CDC operation. After all
+// write handlers for the operation finish, CDC metrics are updated.
+struct operation_result_tracker {
```
haaawk: Maybe make it a class and drop private: label below. Also mark final.

cdc/log.cc
elcallio: This is the correct syntax.

cdc/log.cc
piodul: Changed to string_view

cdc/stats.hh
piodul: Done

cdc/stats.hh
piodul: Done

cdc/log.cc
piodul: Nope, I tried compiling it and it does not complain.

cdc/stats.hh
piodul: Fixed

cdc/stats.hh
piodul: Done

service/storage_proxy.cc
piodul: No longer relevant to the newest version

service/storage_proxy.cc
piodul: (Now this object is called `operation_result_tracker`)

Currently, all CDC counters come in two flavors: "total" and "failed", where "total" means a cumulative value for all operations performed, and "failed" means a value for all operations that failed.

We cannot determine in advance if writing cdc mutations will succeed or fail, so we need to store details about what was happening in the operation. When the operation fails, these details will be used to update relevant "failed" counters.

PR #6031, opened 2020-03-17
Separate sharding and partitioning logic

haaawk:
Currently, both sharding and partitioning logic is encapsulated into partitioners. This is not desirable because these two concepts are totally independent and shouldn't be coupled together in such a way.

This PR separates sharding and partitioning. Partitioning will still live in i_partitioner class and its subclasses. Sharding is extracted to a new class called sharding_info. Both partitioners and sharding_info are still managed by schema class. Partitioner can be accessed with schema::get_partitioner while sharding_info can be accessed with schema::get_sharding_info.

The transition is done in steps:
1. sharding_info class is defined and all the sharding logic is extracted from partitioner to the new class. Temporarily sharding_info is still embedded into i_partitioner and all sharding related functions in i_partitioner call delegate to the embedded sharding_info object.
2. All calls to i_partitioner functions that are related to sharding are gradually switched to calls to sharding_info equivalents. sharding_info.
3. Once everything uses sharding_info, all sharding logic is dropped from i_partitioner.

Tests: unit(dev, release)


haaawk:
@asias could you please review? Repair is one of the heavy users of sharding.

haaawk:
@nyh could you please review? I had to touch a few places in the alternator. Changes are relatively simple but it would be great if you could have a look.

haaawk:
@avikivity could you please have a look? The change spans over the whole codebase so it might deserve a CTO attention ;).

message/messaging_service.hh
```diff
@@ -341,9 +341,9 @@ public:
     future<> send_repair_put_row_diff(msg_addr id, uint32_t repair_meta_id, repair_rows_on_wire row_diff);
 
     // Wrapper for REPAIR_ROW_LEVEL_START
-    void register_repair_row_level_start(std::function<future<> (const rpc::client_info& cinfo, uint32_t repair_meta_id, sstring keyspace_name, sstring cf_name, dht::token_range range, row_level_diff_detect_algorithm algo, uint64_t max_row_buf_size, uint64_t seed, unsigned remote_shard, unsigned remote_shard_count, unsigned remote_ignore_msb, sstring remote_partitioner_name, table_schema_version schema_version)>&& func);
+    void register_repair_row_level_start(std::function<future<> (const rpc::client_info& cinfo, uint32_t repair_meta_id, sstring keyspace_name, sstring cf_name, dht::token_range range, row_level_diff_detect_algorithm algo, uint64_t max_row_buf_size, uint64_t seed, unsigned remote_shard, unsigned remote_shard_count, unsigned remote_ignore_msb, table_schema_version schema_version)>&& func);
```
asias: This change will cause compatibility issue. We can only add new parameters ad rpc::optional and we can not remove old parameters. If  remote_partitioner_name is not needed any more, leave as it is in the rpc verb.

haaawk: Thanks @asias - I was expecting difficulties around this place :). I will do as you suggested. Leave the parameter but just ignore it.

asias: Yes. After this change, we still have the partitioner name, right? If this is the case, we need to send the partitioner name. We need to consider the case when the remote is a old node does not support this change. Will mixed node cluster work?

haaawk: That's a fair point. I will always send "Murmur3" from "new" nodes to be compatible with old nodes in mixed cluster. That should do the trick.

asias:
> @asias could you please review? Repair is one of the heavy users of sharding.

I skimmed the change for repair and streaming subsystem. It looks reasonable.

Thanks.

haaawk:
I addressed @asias comment. Put back the remote_partitioner_name argument in the repair message. Ignored it on the receiving side and always send Murmur3 on the sending side in the new code.

haaawk:
dev and release unit tests are passing. dtests and debug unit tests are running. https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/938/

haaawk:
All debug unit tests are passing as well. Unfortunately, BYO task is broken. @yaronkaikov has restarted it here https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/944/

Let's see if it works this time.

dht/i_partitioner.cc
```diff
@@ -234,60 +234,6 @@ ring_position_range_sharder::next(const schema& s) {
     return ring_position_range_and_shard{std::move(_range), shard};
 }
 
-
-ring_position_exponential_sharder::ring_position_exponential_sharder(const i_partitioner& partitioner, partition_range pr)
```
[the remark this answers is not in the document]
haaawk: @avikivity please comment

repair/row_level.cc
```diff
@@ -592,8 +590,8 @@ class repair_meta {
     uint32_t _repair_meta_id;
     // Repair master's sharding configuration
     shard_config _master_node_shard_config;
-    // Partitioner of repair master
-    std::unique_ptr<dht::i_partitioner> _remote_partitioner;
+    // sharding info of repair master
+    dht::sharding_info _remote_sharding_info;
```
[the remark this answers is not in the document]
haaawk: To be honest I wasn't planning for multiple sharding algorithms. Are there any plans for such? If so then I would have to adjust the PR for that. @avikivity what are the plans?

repair/repair.cc
```diff
@@ -681,6 +657,10 @@ repair_info::repair_info(seastar::sharded<database>& db_,
     , _row_level_repair(db.local().features().cluster_supports_row_level_repair()) {
 }
 
+const dht::sharding_info& repair_info::get_sharding_info() const {
```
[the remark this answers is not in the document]
haaawk: Asias already did and said it's ok. See https://github.com/scylladb/scylla/pull/6031#issuecomment-599991465

mutation_reader.cc
```diff
@@ -1477,6 +1477,7 @@ class multishard_combining_reader : public flat_mutation_reader::impl {
         }
     };
 
+    const dht::sharding_info& _sharding_info;
```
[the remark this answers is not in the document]
haaawk: The lifetime is guaranteed to be longer than any operation because we're using sharding_info object that's static and managed by database. I agree that probably it will be better in the previous patch to also take sharding_info by const& to be consistent.

alternator/executor.cc
```diff
@@ -1181,7 +1181,7 @@ rmw_operation::write_isolation rmw_operation::get_write_isolation_for_schema(sch
 // a read-before-write, but not just on it - depending on configuration,
 // execute() may unconditionally use cas() for every write. Unfortunately,
 // this requires duplicating here a bit of logic from execute().
-std::optional<shard_id> rmw_operation::shard_for_execute(bool needs_read_before_write) {
+std::optional<shard_id> rmw_operation::shard_for_execute(service::storage_proxy& proxy, bool needs_read_before_write) {
```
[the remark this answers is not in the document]
haaawk: Makes sense. I will pass just sharding_info.

service/storage_proxy.cc
```diff
@@ -4883,7 +4883,7 @@ void storage_proxy::init_messaging_service() {
         return get_schema_for_read(cmd.schema_version, src_addr).then([this, cmd = std::move(cmd), key = std::move(key), ballot,
                          only_digest, da, timeout, tr_state = std::move(tr_state), src_ip] (schema_ptr schema) mutable {
             dht::token token = dht::get_token(*schema, key);
-            unsigned shard = dht::shard_of(*schema, token);
+            unsigned shard = _db.local().get_sharding_info().shard_of(token);
```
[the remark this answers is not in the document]
haaawk: Will add such function.

database.hh
```diff
@@ -364,6 +364,7 @@ struct table_stats {
 class table : public enable_lw_shared_from_this<table> {
 public:
     struct config {
+        const dht::sharding_info& sharding_info;
```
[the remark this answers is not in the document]
haaawk: It felt more appropriate to pass sharding_info down from database to table instead of putting it into the schema. I agree it would be simpler but it will create a coupling that is not justified in any way. schema and sharding_info are totally separate. That would create a false impression that sharding info can be set per table. I didn't expect we want multiple sharding algorithms per node. If that's true then it changes my perspective a bit.

dht/token-sharding.hh
```diff
@@ -81,4 +81,6 @@ public:
 
 };
 
+const sharding_info& default_sharding_info();
```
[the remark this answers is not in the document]
haaawk: Good idea. Will add comment.

sstables/sstables.cc
```diff
@@ -3443,7 +3443,8 @@ mutation_source sstable::as_mutation_source() {
     });
 }
 
-sstable::sstable(schema_ptr schema,
+sstable::sstable(const dht::sharding_info& sinfo,
```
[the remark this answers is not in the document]
haaawk: I'm reluctant to join schema and sharding_info unless we want to make it possible to set sharding per table. If sharding_info is meant to be common for the whole node then putting it into schema would couple two unrelated things just for making the change easier which in the long term is probably a negligible advantage.

haaawk:
> I think we plan to move in the opposite direction, to let partitioners do the sharding as well (for tablets for example).
> 
> What's the motivation for this set?

There were two motivations:
1. Without separating sharding and partitioning logic, the repair wasn't possible for tables with different partitioners. When I implemented CDCPartitioner, tests started to fail on node addition/removal because streaming was trying to repair both base table and cdc log and they had different partitioners.
2. Sharding and partitioning logic are independent and all the partitioners actually had the same sharding logic. This PR actually proves that different places in the code use either sharding or partitioning but never both which proves they don't belong together and the coupling is not necessary (actually harmful if we count loose coupling as a characteristic of a good software).

Is it really important for tablets that partitioners do sharding? Wouldn't it be enough to allow setting both independently? That would be even more flexible than partitioners doing sharding. The benefit is that the code isn't so much tightly coupled. Unless I'm missing something, there's nothing using partitioning and sharding at the same time.

sstables/sstables.cc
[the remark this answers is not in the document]
haaawk: I was basing my understanding on @avikivity comment on my previous PR -> https://github.com/scylladb/scylla/pull/5750#discussion_r390981368
```What does it mean to associate a schema with shard_count and sharding_ignore_msb_bits? These are per-node settings. They aren't shared between different nodes for the same table.```

File: service/storage_proxy.cc (hunk 7 above)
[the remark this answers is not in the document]
Comment (haaawk) [reply service/storage_proxy.cc]: done.

File: alternator/executor.cc (hunk 6 above)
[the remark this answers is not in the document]
Comment (haaawk) [reply alternator/executor.cc]: done.

File: dht/token-sharding.hh (hunk 9 above)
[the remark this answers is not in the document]
Comment (haaawk) [reply dht/token-sharding.hh]: I added a comment and renamed to default_sharding_info_for_tests.

File: repair/row_level.cc (hunk 3 above)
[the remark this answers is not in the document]
Comment (haaawk) [reply repair/row_level.cc]: This object is created based on parameters obtained from the other node. As long as we have a single sharding algorithm, keeping sharding_info by value here works fine. If we see the need for different sharding algorithms and subclasses here then we will have to store a unique_ptr here in repair_metadata. It all depends on our plans regarding various sharding algorithms. BTW this is the only place that needs changing. Everything else sharding info is passed by const&.

File: mutation_reader.cc (hunk 5 above)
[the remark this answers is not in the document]
Comment (haaawk) [reply mutation_reader.cc]: I double-checked and sharding_info is always passed by const&. The only place that's different is repair_metadata because there we create sharding_info based on parameters from another node. It can be easily fixed by storing sharding_info as unique_ptr if we need that so in my opinion introducing inheritance later if we need it won't be a problem at all.

**Comment by haaawk** [comment]
I rebased on master and addressed @nyh comments. I answered half-baked thoughts in comments and haven't changed the code for them as it wasn't yet clear what to do.

**Comment by haaawk** [comment]
Rebased on next and resolved conflict.

**Comment by haaawk** [comment]
@tgrabiec my goal is not to fix sharding logic per node but to separate it from partitioning logic. If we find it desirable to enable different sharding logic for different entities in the database we can still do that. The thing is that it is probably not a good idea to set sharding and partitioning logic on the same level. Partitioners have to be set per table because of CDC while sharding is better set on the keyspace level because of streaming/repair that operates on all the tables in the keyspace.

**Comment by haaawk** [comment]
If our plans require the ability to set different sharding logic for different entities in database then I can rework the PR and make it possible to set sharding_logic per keyspace. That would probably be in line with @nyh idea of accessing sharding_info through schema. @avikivity and @tgrabiec could you please share your view on our plans and needs regarding sharding logic in the future?

**Comment by tgrabiec** [comment]
> > I think we plan to move in the opposite direction, to let partitioners do the sharding as well (for tablets for example).
> > What's the motivation for this set?
> 
> There were two motivations:
> 
> 1. Without separating sharding and partitioning logic, the repair wasn't possible for tables with different partitioners. When I implemented CDCPartitioner, tests started to fail on node addition/removal because streaming was trying to repair both base table and cdc log and they had different partitioners.

I don't understand what was the problem with repair and how separating i_partitioner and sharding_info helps. If sharding_info is the same for all partitioners, why separating it helps?


> 2. Sharding and partitioning logic are independent and all the partitioners actually had the same sharding logic. This PR actually proves that different places in the code use either sharding or partitioning but never both which proves they don't belong together and the coupling is not necessary (actually harmful if we count loose coupling as a characteristic of a good software).
> 
> Is it really important for tablets that partitioners do sharding? Wouldn't it be enough to allow setting both independently? That would be even more flexible than partitioners doing sharding. The benefit is that the code isn't so much tightly coupled. Unless I'm missing something, there's nothing using partitioning and sharding at the same time.

Partitioning is about assigning keys to locations in the cluster, where location is a particular shard ( `(node, shard)`). So semantically sharding is part of the partitioner. Currently, all tables use the same mapping for `token -> shard`, but that will change.

Part of partitioning is also the method of assigning tokens to keys. This method is confusingly called the partitioner as well. Maybe you wanted to separate that?

Partitioning in the most general meaning aggregates both `key -> token` and `token -> (node, shard)`.

**Comment by tgrabiec** [comment]
> If our plans require the ability to set different sharding logic for different entities in database then I can rework the PR and make it possible to set sharding_logic per keyspace. That would probably be in line with @nyh idea of accessing sharding_info through schema. @avikivity and @tgrabiec could you please share your view on our plans and needs regarding sharding logic in the future?

In the long term, we want sharding to be per-table.

**Comment by haaawk** [comment]
> > If our plans require the ability to set different sharding logic for different entities in database then I can rework the PR and make it possible to set sharding_logic per keyspace. That would probably be in line with @nyh idea of accessing sharding_info through schema. @avikivity and @tgrabiec could you please share your view on our plans and needs regarding sharding logic in the future?
> 
> In the long term, we want sharding to be per-table.

This will require changes to streaming and repair.

**Comment by haaawk** [comment]
> > > I think we plan to move in the opposite direction, to let partitioners do the sharding as well (for tablets for example).
> > > What's the motivation for this set?
> > 
> > 
> > There were two motivations:
> > 
> > 1. Without separating sharding and partitioning logic, the repair wasn't possible for tables with different partitioners. When I implemented CDCPartitioner, tests started to fail on node addition/removal because streaming was trying to repair both base table and cdc log and they had different partitioners.
> 
> I don't understand what was the problem with repair and how separating i_partitioner and sharding_info helps. If sharding_info is the same for all partitioners, why separating it helps?

Repair uses sharding/partitioning parameters from the other node. Before this PR, it has to create a partitioner with those parameters to emulate remote node. Repair works in batches. Each batch is composed of all the tables in the keyspace but only one set of sharding/partitioning parameters is sent between nodes.

If there are two tables A and B that are in the same keyspace and use different partitioners, repair depends on the fact that they have the same sharding logic. It can not create a partitioner to emulate it because there are two different partitioners present (of A and of B). When I separated partitioners from sharding logic, partitioner is no longer needed to use sharding logic.

> 
> > 1. Sharding and partitioning logic are independent and all the partitioners actually had the same sharding logic. This PR actually proves that different places in the code use either sharding or partitioning but never both which proves they don't belong together and the coupling is not necessary (actually harmful if we count loose coupling as a characteristic of a good software).
> > 
> > Is it really important for tablets that partitioners do sharding? Wouldn't it be enough to allow setting both independently? That would be even more flexible than partitioners doing sharding. The benefit is that the code isn't so much tightly coupled. Unless I'm missing something, there's nothing using partitioning and sharding at the same time.
> 
> Partitioning is about assigning keys to locations in the cluster, where location is a particular shard ( `(node, shard)`). So semantically sharding is part of the partitioner. Currently, all tables use the same mapping for `token -> shard`, but that will change.
> 
> Part of partitioning is also the method of assigning tokens to keys. This method is confusingly called the partitioner as well. Maybe you wanted to separate that?
> 
> Partitioning in the most general meaning aggregates both `key -> token` and `token -> (node, shard)`.

I agree. Probably I used wrong naming. As you correctly suspected, I wanted to separate key->token mapping from token->shard mapping. And that's what this PR does. It also assumes that we have single sharding logic per node but I can change it per table or per keyspace if that's what we will need for tablets or some other features. How does this sound?

**Comment by kostja** [comment]
I disagree with the rationale for this patch (while I don't mind the patch). The implementation is currently indeed orthogonal, but the concepts are very similar. Sharding *is* a way to partition data currently among CPU cores. What is the rationale for this besides cleaning up the implementation (which may change with tablets)?

**Comment by haaawk** [comment]
> I disagree with the rationale for this patch (while I don't mind the patch). The implementation is currently indeed orthogonal, but the concepts are very similar. Sharding _is_ a way to partition data currently among CPU cores. What is the rationale for this besides cleaning up the implementation (which may change with tablets)?

There are two things in sharding/partitioning: mapping from the key to the token and mapping from the token to the shard. In my opinion, those two work together but are two separate things so don't have to live together. Most places in the code use one or the other.

**Comment by kostja** [comment]
> > I disagree with the rationale for this patch (while I don't mind the patch). The implementation is currently indeed orthogonal, but the concepts are very similar. Sharding _is_ a way to partition data currently among CPU cores. What is the rationale for this besides cleaning up the implementation (which may change with tablets)?
> 
> There are two things in sharding/partitioning: mapping from the key to the token and mapping from the token to the shard. In my opinion, those two work together but are two separate things so don't have to live together. Most places in the code use one or the other.

A key could map direclty to the shard, bypassing the mapping to the token. I.e. each token could belong to a single shard only.

File: message/messaging_service.cc [hunk 11]
```diff
@@ -1185,7 +1185,13 @@ void messaging_service::register_repair_row_level_start(std::function<future<> (
 future<> messaging_service::unregister_repair_row_level_start() {
     return unregister_handler(messaging_verb::REPAIR_ROW_LEVEL_START);
 }
-future<> messaging_service::send_repair_row_level_start(msg_addr id, uint32_t repair_meta_id, sstring keyspace_name, sstring cf_name, dht::token_range range, row_level_diff_detect_algorithm algo, uint64_t max_row_buf_size, uint64_t seed, unsigned remote_shard, unsigned remote_shard_count, unsigned remote_ignore_msb, sstring remote_partitioner_name, table_schema_version schema_version) {
+future<> messaging_service::send_repair_row_level_start(msg_addr id, uint32_t repair_meta_id, sstring keyspace_name, sstring cf_name, dht::token_range range, row_level_diff_detect_algorithm algo, uint64_t max_row_buf_size, uint64_t seed, unsigned remote_shard, unsigned remote_shard_count, unsigned remote_ignore_msb, table_schema_version schema_version) {
```
asias: I think it is better to ask the user of send_repair_row_level_start to pass the default partition name in repair/row_level.cc, instead of changing messaging service.  messaging service should not know what is the default remote_partitioner_name.

```
1414     // RPC API
1415     future<>
1416     repair_row_level_start(gms::inet_address remote_node, sstring ks_name, sstring cf_name, dht::token_range range, table_schema_version schema_version) {
1417         if (remote_node == _myip) {
1418             return make_ready_future<>();
1419         }
1420         stats().rpc_call_nr++;
1421         return netw::get_local_messaging_service().send_repair_row_level_start(msg_addr(remote_node),
1422                 _repair_meta_id, std::move(ks_name), std::move(cf_name), std::move(range), _algo, _max_row_buf_size, _seed,
1423                 _master_node_shard_config.shard, _master_node_shard_config.shard_count, _master_node_shard_config.ignore_msb, _master_node_shard_config.partitioner_name, std::move(schema_version));
1424     }
```

haaawk: Done.

haaawk:
Rebased on top of master. Introduced schema::get_sharding_info to make it easy in the future to set it per table. @nyh @tgrabiec @avikivity please have a look.

database.hh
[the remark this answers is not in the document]
haaawk: done - sharding_info is in schema now.

sstables/sstables.cc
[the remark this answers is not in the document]
haaawk: done - sharding_info lives in schema now.

alternator/executor.cc
```diff
@@ -1190,7 +1190,7 @@ std::optional<shard_id> rmw_operation::shard_for_execute(bool needs_read_before_
     // If we're still here, cas() *will* be called by execute(), so let's
     // find the appropriate shard to run it on:
     auto token = dht::get_token(*_schema, _pk);
-    auto desired_shard = service::storage_proxy::cas_shard(*_schema, token);
+    auto desired_shard = proxy.cas_shard(token);
```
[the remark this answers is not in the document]
haaawk: this change went away once sharding_info was moved to schema.

haaawk:
unit tests in dev, release and debug are passing

haaawk:
ping

haaawk:
ping

avikivity:
> > > I disagree with the rationale for this patch (while I don't mind the patch). The implementation is currently indeed orthogonal, but the concepts are very similar. Sharding _is_ a way to partition data currently among CPU cores. What is the rationale for this besides cleaning up the implementation (which may change with tablets)?
> > 
> > 
> > There are two things in sharding/partitioning: mapping from the key to the token and mapping from the token to the shard. In my opinion, those two work together but are two separate things so don't have to live together. Most places in the code use one or the other.
> 
> A key could map direclty to the shard, bypassing the mapping to the token. I.e. each token could belong to a single shard only.

Each token already belongs to a single shard. Please clarify your intent.

Having two partitioning paths [key -> token -> node/shard] and  [key -> node/shard] will be very difficult as it duplicates code paths. Instead we can have partitioners that set token(key) = key or token(key) = key.field (with suitable types for key and key.field).

avikivity:
> > > > I disagree with the rationale for this patch (while I don't mind the patch). The implementation is currently indeed orthogonal, but the concepts are very similar. Sharding _is_ a way to partition data currently among CPU cores. What is the rationale for this besides cleaning up the implementation (which may change with tablets)?
> > > 
> > > 
> > > There are two things in sharding/partitioning: mapping from the key to the token and mapping from the token to the shard. In my opinion, those two work together but are two separate things so don't have to live together. Most places in the code use one or the other.
> > 
> > 
> > A key could map direclty to the shard, bypassing the mapping to the token. I.e. each token could belong to a single shard only.
> 
> Each token already belongs to a single shard. Please clarify your intent.
> 
> Having two partitioning paths [key -> token -> node/shard] and [key -> node/shard] will be very difficult as it duplicates code paths. Instead we can have partitioners that set token(key) = key or token(key) = key.field (with suitable types for key and key.field).

Also. "partitioner" doesn't deal with all partitioning. It deals with converting keys to tokens, and with converting tokens to shards (this part is being split by the patchset). i_partitioner should really be called tokenizer.

If we rename it, we'll have
 - tokenizer - maps keys to tokens
 - abstract_replication_strategy - maps tokens to nodes
 - sharding_info - maps tokens to shards

Note we also have classes named "*sharder" that also map tokens to shards in more complicated ways.

For tablets sharding should become part of the replication strategy.

dht/i_partitioner.cc
```diff
@@ -36,24 +36,37 @@
...
     , _shard_start(init_zero_based_shard_start(_shard_count, _sharding_ignore_msb_bits))
 {}
 
+i_partitioner::i_partitioner(unsigned shard_count, unsigned sharding_ignore_msb_bits)
+    : _sharding_info(shard_count, sharding_ignore_msb_bits)
+{}
+
 unsigned
-i_partitioner::shard_of(const token& t) const {
+sharding_info::shard_of(const token& t) const {
```
avikivity: As I said in a top-level comment, this is more consistently named "sharded", but we can delay renaming until after the dust settles.

dht/token-sharding.hh
```diff
@@ -35,4 +35,56 @@ unsigned shard_of(unsigned shard_count, unsigned sharding_ignore_msb_bits, const
...
+protected:
+    unsigned _shard_count;
+    unsigned _sharding_ignore_msb_bits;
+    std::vector<uint64_t> _shard_start;
+public:
+    sharding_info(unsigned shard_count = smp::count, unsigned sharding_ignore_msb_bits = 0);
+    /**
+     * Calculates the shard that handles a particular token.
+     */
+    virtual unsigned shard_of(const token& t) const;
```
avikivity: Why virtual?

dht/i_partitioner.cc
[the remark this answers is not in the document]
avikivity: ring_position_exponential_sharder was only used by ring_position_exponential_vector_sharder (removed by a5b6374398bf3ed4ca896c5c94ee1c32ad6bf290). The latter's user was removed in 2f66bde26f0ff52afdf22905047a5073311da180, switching to a new mechanism. So we can remove ring_position_exponential_sharder.

dht/token-sharding.hh
haaawk: because there's dummy_sharding_info in tests that has its own logic for testing.

avikivity: Seems wasteful to make this heavily-called function virtual. Can't dummy_sharding_info use the standard logic?

I'll comment there when I reach that commit.

repair/repair.cc
```diff
@@ -640,8 +640,8 @@ future<uint64_t> estimate_partitions(seastar::sharded<database>& db, const sstri
 }
```
avikivity: Very good, this is simplifying things already.

repair/row_level.cc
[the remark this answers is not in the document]
avikivity: The long term plan is to merge sharding into general topology, which is managed by abstract_replication_strategy/token_metadata (inet_address token_metadata::get_endpoint()). I expect that tablets will change that to return a node+shard pair.

repair/row_level.cc
```diff
@@ -1417,9 +1413,16 @@ class repair_meta {
             return make_ready_future<>();
         }
         stats().rpc_call_nr++;
+        // Even though remote partitioner name is ignored in the current version of
+        // repair, we still have to send something to keep compatibility with nodes
+        // that run older versions. This will make it possible to run mixed cluster.
+        // Murmur3 is appropriate because that's the only supported partitioner at
+        // the time this change is introduced.
```
avikivity: We were lucky with this, not to have to send the partitioner name everywhere just for compatibility.

haaawk: Lucky because we removed all other partitioners?

test/lib/dummy_partitioner.hh
```diff
@@ -37,17 +36,11 @@ public:
...
+    dummy_sharding_info(const dht::sharding_info& sharding, const std::map<dht::token, T>& something_by_token)
+        : sharding_info(sharding)
         , _tokens(boost::copy_range<std::vector<dht::token>>(something_by_token | boost::adaptors::map_keys)) {
     }
 
-    virtual dht::token get_token(const schema& s, partition_key_view key) const override { return _partitioner.get_token(s, key); }
-    virtual dht::token get_token(const sstables::key_view& key) const override { return _partitioner.get_token(key); }
-    virtual bool preserves_order() const override { return _partitioner.preserves_order(); }
-    virtual const sstring name() const override { return _partitioner.name(); }
     virtual unsigned shard_of(const dht::token& t) const override;
```
avikivity: Ok. I see that sharding_info::shard_of() preserves the virtualness of i_partitioner::shard_of() and does not introduce it. So it's okay to keep it. though we should try to remove it later. It certainly doesn't seem to be easy.

db/schema_tables.cc
```diff
@@ -2353,6 +2353,7 @@ schema_ptr create_table_from_mutations(const schema_ctxt& ctxt, schema_mutations
 
     if (auto partitioner = sm.partitioner()) {
         builder.with_partitioner(*partitioner, smp::count, ctxt.murmur3_partitioner_ignore_msb_bits());
+        builder.with_sharding_info(smp::count, ctxt.murmur3_partitioner_ignore_msb_bits());
```
avikivity: But doesn't the partitioner already contain a sharding_info? Either we should drop the sharding_info from the partitioner, or use the sharding_info from the partitioner. It doesn't make sense for one table to have two sharding_infos.

avikivity: Ok, I see this later pach removed sharding_info from the partitioner. So it's good.

avikivity: And the changelog explains the intent, so I can't complain.

avikivity:
Looks good to me. Even apart from the changes we want to make, this is a good cleanup.

repair/row_level.cc
avikivity: Lucky that it turned out we don't need to do complicated acrobatics for something that nobody uses.

kostja:
* Avi Kivity <notifications@github.com> [20/03/29 15:54]:
> > > > I disagree with the rationale for this patch (while I don't mind the patch). The implementation is currently indeed orthogonal, but the concepts are very similar. Sharding _is_ a way to partition data currently among CPU cores. What is the rationale for this besides cleaning up the implementation (which may change with tablets)?
> > > 
> > > 
> > > There are two things in sharding/partitioning: mapping from the key to the token and mapping from the token to the shard. In my opinion, those two work together but are two separate things so don't have to live together. Most places in the code use one or the other.

I meant token range, not token.
 
> > A key could map direclty to the shard, bypassing the mapping to the token. I.e. each token could belong to a single shard only.
> 
> Each token already belongs to a single shard. Please clarify your intent.
> 
> Having two partitioning paths [key -> token -> node/shard] and  [key -> node/shard] will be very difficult as it duplicates code paths. Instead we can have partitioners that set token(key) = key or token(key) = key.field (with suitable types for key and key.field).

I am just not seeing benefits (perhaps I'm simply unaware of)
breaking up sharding and partitioning. In the end I want to get
from a key to some routing information. The fewer steps it is, the
better. Given the amount of reviewers this patch has I don't think 
my concerns need to be addressed.

-- 

avikivity:
> * Avi Kivity <notifications@github.com> [20/03/29 15:54]:
> > > > I disagree with the rationale for this patch (while I don't mind the patch). The implementation is currently indeed orthogonal, but the concepts are very similar. Sharding _is_ a way to partition data currently among CPU cores. What is the rationale for this besides cleaning up the implementation (which may change with tablets)? > > > > > > There are two things in sharding/partitioning: mapping from the key to the token and mapping from the token to the shard. In my opinion, those two work together but are two separate things so don't have to live together. Most places in the code use one or the other.
> I meant token range, not token.
> > A key could map direclty to the shard, bypassing the mapping to the token. I.e. each token could belong to a single shard only. Each token already belongs to a single shard. Please clarify your intent. Having two partitioning paths [key -> token -> node/shard] and [key -> node/shard] will be very difficult as it duplicates code paths. Instead we can have partitioners that set token(key) = key or token(key) = key.field (with suitable types for key and key.field).
> I am just not seeing benefits (perhaps I'm simply unaware of) breaking up sharding and partitioning. In the end I want to get from a key to some routing information. The fewer steps it is, the better. Given the amount of reviewers this patch has I don't think my concerns need to be addressed.


Questions should be answered, at a minimum someone will learn something.

In this case we are above the minimum, I became aware that partitioner is a confusing name since it doesn't deal with partitioning, and we should rename it to tokenizer. So we're not splitting sharding from partitioning, we're splitting sharding from tokenizing.

haaawk:
I rebased on top of master, resolved multiple conflicts and added commits that rename sharding_info to sharder. Is there anything else you would like me to do @avikivity?

dev and release unit tests are passing.

haaawk:
@tgrabiec and @nyh are you planning to review this PR again or are we good to merge this?

repair/row_level.cc
```diff
@@ -1417,9 +1413,16 @@ class repair_meta {
...
+        sstring remote_partitioner_name = "org.apache.cassandra.dht.Murmur3Partitioner";
```
tgrabiec: Why can't we send `_master_node_shard_config.partitioner_name` anymore?

multishard_mutation_query.cc
```diff
@@ -585,9 +585,9 @@ static future<reconcilable_result> do_query_mutations(
         tracing::trace_state_ptr trace_state,
         db::timeout_clock::time_point timeout,
         query::result_memory_accounter&& accounter) {
-    return do_with(seastar::make_shared<read_context>(db, s, cmd, ranges, trace_state), [s, &cmd, &ranges, trace_state, timeout,
+    return do_with(seastar::make_shared<read_context>(db, s, cmd, ranges, trace_state), [&db, s, &cmd, &ranges, trace_state, timeout,
             accounter = std::move(accounter)] (shared_ptr<read_context>& ctx) mutable {
-        return ctx->lookup_readers().then([&ctx, s = std::move(s), &cmd, &ranges, trace_state, timeout,
+        return ctx->lookup_readers().then([&db, &ctx, s = std::move(s), &cmd, &ranges, trace_state, timeout,
```
tgrabiec: This change seems unused.

repair/row_level.cc
haaawk: It is removed.

multishard_mutation_query.cc
haaawk: Yes. Thanks for catching will remove.

haaawk:
Removed unused capture Tomek has noticed.

avikivity:
Queued as dee0b6834

PR #6032, opened 2020-03-17
Make compaction cancellation easier.

glommer:
Right now we can cancel all compactions of a specific type (cleanup,
regular, etc). This is not good enough, as in some situations we may
want to cancel a specific compaction.

One example is the upcoming offstrategy compactions that are fired
when the compaction strategy changes. If the user changes the compaction
strategy and then changes it again we will want to cancel just the
compaction that we started. Not all compactions on the column family,
not all compactions of a specific type.

There is already a mechanism to cancel a compaction given a task
pointer, so all we have to do is expose the task pointer to the
caller.

At this moment I don't see the need to cancel regular minor compactions
so that is deferred until and if we need that.

For now we will return the task to major compaction callers.

Note that there is no need to do anything with repect to the management
of the _tasks list. If the task is not added to the list the future will
be ready immediately. If it is, the compaction will have started already
for sure and will stop with an exception - removing itself from the
list, when it sees the stop notification

### DIFF
--- a/sstables/compaction_manager.cc
+++ b/sstables/compaction_manager.cc
@@ -256,12 +256,15 @@ private:
     virtual void remove_sstable(sstables::shared_sstable sst)  override { }
 };
 
-future<> compaction_manager::submit_major_compaction(column_family* cf) {
-    if (_stopped) {
-        return make_ready_future<>();
-    }
+lw_shared_ptr<compaction_manager::task>
+compaction_manager::submit_major_compaction(column_family* cf) {
     auto task = make_lw_shared<compaction_manager::task>();
     task->compacting_cf = cf;
+    if (_stopped) {
+        task->compaction_done = make_ready_future<>();
+        return task;
+    }
+
     _tasks.push_back(task);
 
     // first take major compaction semaphore, then exclusely take compaction lock for column family.
@@ -306,7 +309,7 @@ future<> compaction_manager::submit_major_compaction(column_family* cf) {
             _stats.errors++;
         }
     });
-    return task->compaction_done.get_future().then([task] {});
+    return task;
 }
 
 future<> compaction_manager::run_resharding_job(column_family* cf, std::function<future<>()> job) {
--- a/sstables/compaction_manager.hh
+++ b/sstables/compaction_manager.hh
@@ -54,7 +54,7 @@ public:
         uint64_t active_tasks = 0; // Number of compaction going on.
         int64_t errors = 0;
     };
-private:
+
     struct task {
         column_family* compacting_cf = nullptr;
         shared_future<> compaction_done = make_ready_future<>();
@@ -63,6 +63,7 @@ private:
         sstables::compaction_type type = sstables::compaction_type::Compaction;
         bool compaction_running = false;
     };
+private:
 
     // compaction manager may have N fibers to allow parallel compaction per shard.
     std::list<lw_shared_ptr<task>> _tasks;
@@ -184,7 +185,7 @@ public:
     future<> perform_sstable_scrub(column_family* cf, bool skip_corrupted);
 
     // Submit a column family for major compaction.
-    future<> submit_major_compaction(column_family* cf);
+    lw_shared_ptr<task> submit_major_compaction(column_family* cf);
 
     // Run a resharding job for a given column family.
     // it completes when future returned by job is ready or returns immediately
--- a/table.cc
+++ b/table.cc
@@ -1352,7 +1352,8 @@ future<> table::rewrite_sstables(sstables::compaction_descriptor descriptor) {
 // Note: We assume that the column_family does not get destroyed during compaction.
 future<>
 table::compact_all_sstables() {
-    return _compaction_manager.submit_major_compaction(this);
+    auto task = _compaction_manager.submit_major_compaction(this);
+    return task->compaction_done.get_future().then([task] {});
 }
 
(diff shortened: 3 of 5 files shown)


raphaelsc:
> ere is already a mechanism to cancel a compaction given a task
> pointer, so all we have to do is expose the task pointer to the
> caller.

worth to mention though that task_stop() will not cancel the ongoing compaction, but rather prevent the task from trying compaction again like to satisfy to strategy once the ongoing compaction finishes.

to actually cancel a specific ongoing compaction, we'll probably need to forward to sstable::compaction via descriptor some mechanism (like utils::observable) which will allow the compaction entity to be stopped once the compaction initiator asks for it.

currently, we cancel a compaction on behalf of a table by iterating through _compactions list (respective entry is stored in sstables::compaction) and communicating the need to stop. We do check it whenever emitting a new partition.

glommer:
Thanks for the info. I thought this would be enough to actually do the cancellation.

I see no need to rush a half measure. I am closing this and will re-submit with the compaction_info forwarded (somehow)

raphaelsc:
On Tue, Mar 17, 2020 at 2:09 PM Glauber Costa <notifications@github.com>
wrote:

> Thanks for the info. I thought this would be enough to actually do the
> cancellation.
>
> I see no need to rush a half measure. I am closing this and will re-submit
> with the compaction_info forwarded (somehow)
>
I thought about that just recently. That will be a great move to remove the
cyclic dependence between manager and compaction itself. Let's make task
store lw_shared_ptr<compaction_info> and forward it to compaction via
compaction_descriptor. Now compaction no longer calls compaction manager
methods to register and deregister compactions, manager itself will do it.

sstables/compaction_manager.cc
```diff
@@ -256,12 +256,15 @@ class user_initiated_backlog_tracker final : public compaction_backlog_tracker::
...
-future<> compaction_manager::submit_major_compaction(column_family* cf) {
-    if (_stopped) {
-        return make_ready_future<>();
-    }
+lw_shared_ptr<compaction_manager::task>
+compaction_manager::submit_major_compaction(column_family* cf) {
     auto task = make_lw_shared<compaction_manager::task>();
     task->compacting_cf = cf;
+    if (_stopped) {
+        task->compaction_done = make_ready_future<>();
```
bhalevy: I wonder if that's really what we want to return to the user?
But that's out of the scope of this change.

bhalevy: Also compaction_done seems to be initialized to a (shared) ready future.
So I'm not sure how waiting for it works (even without this change).

bhalevy: @raphaelsc ?

raphaelsc:
On Wed, Mar 18, 2020 at 9:23 AM Benny Halevy <notifications@github.com>
wrote:

> *@bhalevy* commented on this pull request.
>
> In sstables/compaction_manager.cc
> <https://github.com/scylladb/scylla/pull/6032#discussion_r394306219>:
>
> >      auto task = make_lw_shared<compaction_manager::task>();
>      task->compacting_cf = cf;
> +    if (_stopped) {
> +        task->compaction_done = make_ready_future<>();
>
> Also compaction_done seems to be initialized to a (shared) ready future.
> So I'm not sure how waiting for it works (even without this change).
>
3ac22bc0d7e57e explains why it was introduced.

bhalevy:
> [3ac22bc](https://github.com/scylladb/scylla/commit/3ac22bc0d7e57edc9b4cb96305a561c9cf4f7108) explains why it was introduced.

@raphaelsc this commit is about `shared_future`, but even before it `compaction_done` was initialized to `make_ready_future<>()`.  So when was it not ready?

raphaelsc:
On Wed, Mar 18, 2020 at 11:26 AM Benny Halevy <notifications@github.com>
wrote:

> 3ac22bc
> <https://github.com/scylladb/scylla/commit/3ac22bc0d7e57edc9b4cb96305a561c9cf4f7108>
> explains why it was introduced.
>
> @raphaelsc <https://github.com/raphaelsc> this commit is about
> shared_future, but even before it compaction_done was initialized to
> make_ready_future<>(). So when was it not ready?
>
A few lines below in the code you'll see that we replace the ready future
(we only return a ready future if compaction manager was asked to stop or
something) with a non-ready future for the compaction task.

bhalevy:
> A few lines below in the code you'll see that we replace the ready future (we only return a ready future if compaction manager was asked to stop or something) with a non-ready future for the compaction task.

This seems a bit awkward... I think that the natural way of doing it would be to have an non-ready shared_promise on the task instead and call either `set_value` or `set_exception` when compaction is done / aborted, or as you said if the compaction_manager is already stopped then set it to a ready future (and I think that it'd be better to return an error to the user rather than success in this case is more appropriate...)

raphaelsc:
On Wed, Mar 18, 2020 at 12:37 PM Benny Halevy <notifications@github.com>
wrote:

> A few lines below in the code you'll see that we replace the ready future
> (we only return a ready future if compaction manager was asked to stop or
> something) with a non-ready future for the compaction task.
>
> This seems a bit awkward... I think that the natural way of doing it would
> be to have an non-ready shared_promise on the task instead and call either
> set_value or set_exception when compaction is done / aborted,
>
 shared_future is a facility that provides all that. you can use it to have
N fibers waiting on the same future.

PR #6036, opened 2020-03-18
Adjust ranges per stream plan for decommission and bootstrap

asias:
This series adjusts ranges per stream plan for decommission and bootstrap.

1) For decommission

The receiver of the decommission stream plan invalidates the cache for
the sstable that the stream plan generates. A single sstable is
generated covering multiple ranges. The cache between the start of the
first range and the end of the last range is invalidated. This
invalidates more caches ranges than it is necessary.  The receiver of
the decommission node serves read. The cache invalidation has a huge
impact on the read latency.

To mitigate the cache invalidation impact before we fix the cache
invalidation, stream 1 range per stream plan as a temporary fix. This
will cause more sstables to be generated on the receiver and thus
generates more compaction work. It is worth it because the cache
invalidation impact is much worse than the compaction.

2) For bootstrap 

Compaction dominating IO bandwidth makes streaming slow during
bootstrap. To mitigate problem before we have off-strategy compaction
support, we want to reduce the compaction work during bootstrap. The
patch 'compaction: use a larger min_threshold during bootstrap, replace'
reduces the compaction by increasing the min threshold of sstables for
compaction.

In this patch, we reduce the number of sstables generated for bootstrap,
by increasing the ranges per stream plan. The bootstrap node does not
server read, so it does not have cache invalidation problem as it has in
decommission.

Refs: #5226 #5199 #5109 #4884 #5495

### DIFF
--- a/dht/range_streamer.cc
+++ b/dht/range_streamer.cc
@@ -317,6 +317,19 @@ future<> range_streamer::do_stream_async() {
                     // compaction work. It is worth it because the cache
                     // invalidation impact is much worse than the compaction.
                     nr_ranges_per_stream_plan = 1;
+                } else if (_reason == streaming::stream_reason::bootstrap) {
+                    // Compaction dominating IO bandwidth makes streaming slow
+                    // during bootstrap. To mitigate problem before we have
+                    // off-strategy compaction support, we want to reduce the
+                    // compaction work during bootstrap. The patch 'compaction:
+                    // use a larger min_threshold during bootstrap, replace'
+                    // reduces the compaction by increasing the min threshold
+                    // of sstables for compaction. Here, we reduce the number
+                    // of sstables generated during bootstrap, by increasing
+                    // the ranges per stream plan. The bootstrap node does not
+                    // server read, so it does not have cache invalidation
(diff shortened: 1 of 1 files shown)


glommer:
The decommission part looks good to me and should go in.
The benefits of not invalidating too many cache ranges outweigh the increased compaction load in my view. It's also an easy to backport patch, which is good.

I think we should be more careful with the trade offs involved with bootstrap and wait to see how the performance will be with min_threshold increasing. For now I'd prefer not to change the boostrap range calculation code.

asias:
The only down side I see with "range_streamer: Generate less sstables for bootstrap " is that in case of streaming error we need to re-stream 20% of data compared to 10% data. It reduces the number of sstables generated by half which make the min_threshold change even more safe. In my view, the extra re-stream in case of error is worth the benefits of less compaction due to less sstables.

It's ok to upstream the decommission change only for now. We can revisit the bootstrap optimization later once we have min_threshold baseline numbers.

asias:
Retire this one.

Created https://github.com/scylladb/scylla/pull/6048

PR #6037, opened 2020-03-18
cdc: remove result_callback

piodul:
The `result_callback` was a callback returned by `augment_mutation_call` that was supposed to be used in the CDC postimage implementation. Because CDC postimage was implemented without using this callback, and currently a no-op function is always returned, this callback can safely be removed.

Tests: unit(dev)

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -156,7 +156,7 @@ public:
         }
     }
 
-    future<std::tuple<std::vector<mutation>, result_callback>> augment_mutation_call(
+    future<std::vector<mutation>> augment_mutation_call(
         lowres_clock::time_point timeout,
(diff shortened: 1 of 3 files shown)


elcallio:
postimage, otherwise lgtm

piodul:
@elcallio Updated both commit and PR descriptions.

piodul:
Actually, this callback __will__ be reused for the purpose of collecting metrics. If we merged it, I would only have to reintroduce it again.

piodul:
It turns out I was wrong, and the callback will not be needed. I closed this PR before thinking enough about my idea of reusing the callback, which turned out to be useless. I apologize for messy behaviour.

PR #6041, opened 2020-03-18
cdc: remove result_callback

piodul:
The `result_callback` was a callback returned by `augment_mutation_call` that was supposed to be used in the CDC postimage implementation. Because CDC postimage was implemented without using this callback, and currently a no-op function is always returned, this callback can safely be removed.

Tests: unit(dev)

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -156,7 +156,7 @@ public:
         }
     }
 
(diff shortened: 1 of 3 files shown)


piodul:
This is reopened #6037

nyh:
Thanks. Committed to "next". Because this is not a merge, Github will not notice so closing this pull-request manually.

Issue #6013, opened 2020-03-12
Make counters work with CDC


jul-stas:
A table with counters cannot have CDC enabled:
```
cqlsh> CREATE TABLE jul.tb1 (id int PRIMARY KEY, cnt counter) with cdc={'enabled':true};
ConfigurationException: Cannot add a non counter column (cdc$deleted_cnt) in a counter column family
```
This happens for a purely technical reason: counter columns are territorial animals and don't allow anyone else into CF, apart from PK and CK (and other counters). If we try to enable CDC for a table-with-counters, CDC tries to create "ordinary" columns like `ttl`, `deleted$*` which are in conflict with (a) counter column(s). The CDC column type (here: `counter_type`) is inferred from the base table, hence the error.

I have a quickpatch with counter columns being cast to bigints, but this would not be reliable - please imagine a situation where two counter-leaders concurrently update their own counter-shards - what CDC should both of them log? Should they reconcile their CDC writes? I don't know, my proposals are:
1) do CDC logging per counter-shard. Every row of CDC "delta" log would have the counter-update and shard ID;
2) keep the track of (shard, update) in a single CDC row, in unfrozen map. Merge the updates within... idk, some time frame?

I'm not sure that clients want to be bothered with "counter shards". I'm even less sure about reconciliation of CDC logs for counter writes.
Preimage/postimage? even more issues may arise.

Please, comment @avikivity @slivne @haaawk @elcallio @kbr- @gleb-cloudius

haaawk:
I think that representing counter as a map<shard_id, int64> would be best. This way we would be able to replicate counter value to another Scylla instance. This map should be non-frozen and each counter shard leader should write only updates to its own shard.

kbr-:
Is it possible to reliably log the actual delta (difference between counter values caused by an update)?
Also, there's the third option:

3. don't allow CDC on counter tables

haaawk:
> Is it possible to reliably log the actual delta (difference between counter values caused by an update)?
> Also, there's the third option:
> 
> 1. don't allow CDC on counter tables

Let's first explore other options.

jul-stas:
> Is it possible to reliably log the actual delta (difference between counter values caused by an update)?
Seems doable to me - we can try augmenting counter update mutations. The cdc-deltas would be propagated alongside corresponding counter-shards-mutations, as reliably as counters do it..?

ATM I'm investigating how C* solved it (yes, they have something). We can think in the meantime on how to do it perfectly.

[...]

PR #6047, opened 2020-03-19
cdc+counters: [RFC] Display updated counter shard in CDC delta log

jul-stas:
This is a pure RFC. I'm asking for comments before I go too deep.

The goal is to make counters work with CDC and this is a fragile thing. Apparently, we'll have to introduce some simple but convoluted codepaths, so I mostly ask for logic/design ideas. Currently, the plan is to:

- make CDC "delta" entries be the maps of all counter-shards `{id:value}` for affected columns
- set preimage as the entire affected row before RMW
- set postimage as the entire affected row after RMW

For sure, there are features missing ATM, e.g. postimage, preimage, deleted counters, disabled TTLs, more. Batches and hints need particular attention.

Refs #6013

### DIFF
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -2144,9 +2144,13 @@ storage_proxy::get_paxos_participants(const sstring& ks_name, const dht::token &
  */
 future<> storage_proxy::mutate(std::vector<mutation> mutations, db::consistency_level cl, clock_type::time_point timeout, tracing::trace_state_ptr tr_state, service_permit permit, bool raw_counters) {
     if (_cdc && _cdc->needs_cdc_augmentation(mutations)) {
-        return _cdc->augment_mutation_call(timeout, std::move(mutations), tr_state).then([this, cl, timeout, tr_state, permit = std::move(permit), raw_counters](std::vector<mutation>&& mutations) mutable {
-            return _mutate_stage(this, std::move(mutations), cl, timeout, std::move(tr_state), std::move(permit), raw_counters);
-        });
+        const bool mutation_is_counter_update = mutations.front().schema()->is_counter() && !raw_counters;
+        // Counter updates (deltas) cannot be augmented.
+        if (!mutation_is_counter_update) {
+            return _cdc->augment_mutation_call(timeout, std::move(mutations), tr_state).then([this, cl, timeout, tr_state, permit = std::move(permit), raw_counters](std::vector<mutation>&& mutations) mutable {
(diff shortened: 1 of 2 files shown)


service/storage_proxy.cc
```diff
@@ -2163,8 +2167,22 @@ future<> storage_proxy::do_mutate(std::vector<mutation> mutations, db::consisten
 
 future<> storage_proxy::replicate_counter_from_leader(mutation m, db::consistency_level cl, tracing::trace_state_ptr tr_state,
                                                       clock_type::time_point timeout, service_permit permit) {
-    // FIXME: do not send the mutation to itself, it has already been applied (it is not incorrect to do so, though)
```
piodul: Does this comment still apply? I suspect it does, and if so we should probably not remove it.

jul-stas: Ouch, yes, it does.

service/storage_proxy.cc
```diff
@@ -2163,8 +2167,22 @@ future<> storage_proxy::do_mutate(std::vector<mutation> mutations, db::consisten
...
+        f_cdc = get_cdc_service()->augment_mutation_call(timeout, std::move(mut_vec), tr_state);
+    }
+
+    return f_cdc.then([this, cl, tr_state, timeout, permit = std::move(permit)]
+            (std::vector<mutation>&& mutations) {
+        auto borderline = boost::range::partition(mutations, [] (auto&& m) {
+            return m.schema()->is_counter();
+        });
+        return seastar::when_all_succeed(
+            mutate_internal(boost::make_iterator_range(mutations.begin(), borderline), cl, true, tr_state, permit, timeout),
```
haaawk: I don't think it's a good idea to assume mutations will be grouped into counter updates + cdc writes.

haaawk:
Looks good in general.

jul-stas:
@haaawk Don't you foresee any apparent disturbance of batches with this code flow?
@piodul Don't you foresee any apparent disturbance of hints with this code flow?
@kbr-  Don't you see any violation of CDC principles with this code flow?

Sorry for pushing - ofc it's my responsibility to get it right - but bad decisions here can do collateral harm to all of us.

piodul:
@jul-stas I think hints should work alright in that flow. It doesn't seem that there is any confusion between update counter mutations and shard counter mutations.

Hints generated for a cdc log table of a base table that has a counter should be safe - there is no possibility for confusion, because you convert counter columns from base table to a non-counter column in log table.

PR #6048, opened 2020-03-20
Adjust ranges per stream plan for decommission

asias:
range_streamer: Stream one range per stream plan for decommission

The receiver of the decommission stream plan invalidates the cache for
the sstable that the stream plan generates. A single sstable is
generated covering multiple ranges. The cache between the start of the
first range and the end of the last range is invalidated. This
invalidates more caches ranges than it is necessary.  The receiver of
the decommission node serves read. The cache invalidation has a huge
impact on the read latency.

To mitigate the cache invalidation impact before we fix the cache
invalidation, stream 1 range per stream plan as a temporary fix. This
will cause more sstables to be generated on the receiver and thus
generates more compaction work. It is worth it because the cache
invalidation impact is much worse than the compaction.

Refs #5495

### DIFF
--- a/dht/range_streamer.cc
+++ b/dht/range_streamer.cc
@@ -302,6 +302,23 @@ future<> range_streamer::do_stream_async() {
                 size_t nr_ranges_total = range_vec.size();
                 size_t groups = 10;
                 size_t nr_ranges_per_stream_plan = (nr_ranges_total + groups - 1) / groups;
+                if (_reason == streaming::stream_reason::decommission) {
+                    // The receiver of the decommission stream plan invalidates
+                    // the cache for the sstable that the stream plan
+                    // generates. A single sstable is generated covering
+                    // multiple ranges. The cache between the start of the first
+                    // range and the end of the last range is invalidated. This
+                    // invalidates more caches ranges than it is necessary.
+                    // The receiver of the decommission node serves read. The
+                    // cache invalidation has a huge impact on the read latency.
+                    // To mitigate the cache invalidation impact before we fix
+                    // the cache invalidation, stream 1 range per stream plan
+                    // as a temporary fix. This will cause more sstables to be
+                    // generated on the receiver and thus generates more
+                    // compaction work. It is worth it because the cache
+                    // invalidation impact is much worse than the compaction.
+                    nr_ranges_per_stream_plan = 1;
+                }
                 dht::token_range_vector ranges_to_stream;
                 auto do_streaming = [&] {
                     auto sp = stream_plan(format("{}-{}-index-{:d}", description, keyspace, sp_index++), _reason);
--- a/dht/range_streamer.cc
+++ b/dht/range_streamer.cc
@@ -300,7 +300,8 @@ future<> range_streamer::do_stream_async() {
                 unsigned sp_index = 0;
                 unsigned nr_ranges_streamed = 0;
                 size_t nr_ranges_total = range_vec.size();
-                size_t nr_ranges_per_stream_plan = nr_ranges_total / 10;
+                size_t groups = 10;
+                size_t nr_ranges_per_stream_plan = (nr_ranges_total + groups - 1) / groups;
                 dht::token_range_vector ranges_to_stream;
                 auto do_streaming = [&] {
                     auto sp = stream_plan(format("{}-{}-index-{:d}", description, keyspace, sp_index++), _reason);


dht/range_streamer.cc
```diff
@@ -300,7 +300,7 @@ future<> range_streamer::do_stream_async() {
                 unsigned sp_index = 0;
                 unsigned nr_ranges_streamed = 0;
                 size_t nr_ranges_total = range_vec.size();
-                size_t nr_ranges_per_stream_plan = nr_ranges_total / 10;
+                size_t nr_ranges_per_stream_plan = nr_ranges_total / 10 + 1;
```
denesb: What if there are 760 ranges (no remainder), won't the estimation be off just the same?

avikivity: Yes, the correct calculation is (nr_rangtes_total + 9) / 10.

dht/range_streamer.cc
```diff
@@ -301,6 +301,23 @@ future<> range_streamer::do_stream_async() {
...
+                    // invalidates more caches ranges than it is necessary.
+                    // The receiver of the decommission node serves read. The
+                    // cache invalidation has a huge impact on the read latency.
+                    // To mitigate the cache invalidation impact before we fix
+                    // the cache invalidation, stream 1 range per stream plan
+                    // as a temporary fix. This will cause more sstables to be
+                    // generated on the receiver and thus generates more
+                    // compaction work. It is worth it because the cache
+                    // invalidation impact is much worse than the compaction.
+                    nr_ranges_per_stream_plan = 1;
```
avikivity: I don't agree that it's worse, they are both worse.

Can we add a vector of token ranges to send along with the stream plan to invalidate?

asias: We do not need the ranges to be transferred from the sender.  We can intersect the ranges the receiver node owns with the range from the sstable. We invalidate the intersected ranges. 

This is close to how we would fix the problem at the cache layer. 

This is more complicated and adds more risks to backport.

asias:
Changes in 1572780c88d6dfb289ad27eaba633f610fd09072:
-  Fix nr_ranges_per_stream_plan off by one issue

dht/range_streamer.cc
asias: @glommer @avikivity Please decide which way to proceed.

glommer: I agree with @avikivity . Let's stream the information about the intervals. That is for two reasons:
- It matches the overall direction we're going
- It reduces the risk of increased compaction.

Backportability is a factor, but between that and performance we'll pick performance. The main drawback of this approach was the high number of SSTables generated, and with Avi's suggestion we can get rid of it.

asias: We had the discussion how to fix before I sent this patch 1) more complicated fix that change how we invalidate the cache or 2) simple fix that does not send cross vnode ranges in stream.

You wanted the simple fix. You also did not want to add the new parameter to existing RPC verb in the solution. 

Now, you want both the cache fix and the new parameter. 

See my comments above, https://github.com/scylladb/scylla/pull/6048#discussion_r396166189, we do not really need to send the ranges. We can intersect with the ranges the receiver node owns.

I am fine with fix the cache invalidation, but let's decide if we want to send the ranges over the rpc verb or not .

glommer: Hi Asias. I don't recall discussing adding an rpc verb with you in the context of this patch.
I recall doing so in the context of bootstrap.
Bootstrap is different, because that is a verb we wouldn't necessarily use with offstrategy so I felt it was premature.

I don't feel a verb here is premature. They are different situations.

About cache, as far as I understand the work is still not touching the cache per se, which is what I want to avoid. We still invalidate entire ranges and don't change anything in the cache, except that we decide those ranges based on the information coming in the wire.

Also note that it doesn't necessarily have to be an RPC verb: if we could encode this information in the SSTable metadata component, that works as well.

However between the two, I suspect RPC will be cleaner.

asias: I meant this discussion: https://github.com/scylladb/scylla/pull/5866#issuecomment-593942272

Just be clear, we do not need a new rpc verb here at all.

My question remains. Do we really need to send the ranges from the sender? I think we do not need to do so.

slivne:
@asias did you address all comments if so please requests a review

asias:
@slivne we are still discussing which way to proceed: https://github.com/scylladb/scylla/pull/6048#discussion_r399249868

asias:
I had a chat with Glauber: 

A summary:

1) No need to send the ranges from the sender to the receiver. The receiver can intersect the range from the sstable with the ranges it owns.

2) We prefer to fix at the cache layer to not invalidate what it has to invalidate. 

I am retiring this PR.

glommer:
@asias @slivne it's fine to retire this, but need understanding on who's going to tackle this at the cache side. Is this something that asias is expected to do ? Someone else ?

PR #6057, opened 2020-03-22
Use inline variables to reduce code duplication and complexity

haaawk:
C++17 introduced inline variables and now we don't have to declare a global extern in a header and then define it in .cc file. It is better to use inline variables because it reduces code duplication and complexity. Defining a variable in a single place is simpler and more maintainable.

This PR switches multiple places to use inline variables.

Tests: unit (dev)

### DIFF
--- a/auth/common.cc
+++ b/auth/common.cc
@@ -32,15 +32,6 @@
 
 namespace auth {
 
-namespace meta {
-
-const sstring DEFAULT_SUPERUSER_NAME("cassandra");
-const sstring AUTH_KS("system_auth");
-const sstring USERS_CF("users");
-const sstring AUTH_PACKAGE_NAME("org.apache.cassandra.auth.");
-
-}
-
 static logging::logger auth_log("auth");
 
 // Func must support being invoked more than once.
--- a/auth/common.hh
+++ b/auth/common.hh
@@ -52,10 +52,10 @@ namespace auth {
 
 namespace meta {
 
-extern const sstring DEFAULT_SUPERUSER_NAME;
-extern const sstring AUTH_KS;
-extern const sstring USERS_CF;
-extern const sstring AUTH_PACKAGE_NAME;
+inline const sstring DEFAULT_SUPERUSER_NAME("cassandra");
+inline const sstring AUTH_KS("system_auth");
+inline const sstring USERS_CF("users");
+inline const sstring AUTH_PACKAGE_NAME("org.apache.cassandra.auth.");
 
 }
 
--- a/auth/permission.cc
+++ b/auth/permission.cc
@@ -45,17 +45,6 @@
 
 #include <unordered_map>
 
-const auth::permission_set auth::permissions::ALL = auth::permission_set::of<
-        auth::permission::CREATE,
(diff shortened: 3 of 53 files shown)


avikivity:
@espindola don't such changes increase object sizes and work for the linker?

Work for the linker isn't so important now since lld is plenty fast.

haaawk:
> @espindola don't such changes increase object sizes and work for the linker?
> 
> Work for the linker isn't so important now since lld is plenty fast.

Isn't there still just one instance of the object in the final binary? Do we care about *.o files size that much?

avikivity:
> > @espindola don't such changes increase object sizes and work for the linker?
> > Work for the linker isn't so important now since lld is plenty fast.
> 
> Isn't there still just one instance of the object in the final binary? Do we care about *.o files size that much?

We don't care much about the .o size, but I think it can increase compilation time. Every time someone uses an inline variable the compiler has to generate constructors and destructors for it and add them to the .o, which the linker then eliminates. Similar to inline functions, you don't make all your functions inline just because the linker eliminates the duplicates in the end.

haaawk:
> > > @espindola don't such changes increase object sizes and work for the linker?
> > > Work for the linker isn't so important now since lld is plenty fast.
> > 
> > 
> > Isn't there still just one instance of the object in the final binary? Do we care about *.o files size that much?
> 
> We don't care much about the .o size, but I think it can increase compilation time. Every time someone uses an inline variable the compiler has to generate constructors and destructors for it and add them to the .o, which the linker then eliminates. Similar to inline functions, you don't make all your functions inline just because the linker eliminates the duplicates in the end.

Would you like me to measure compile time with and without this PR @avikivity ?

avikivity:
It can't hurt, but I also want to hear from @espindola

haaawk:
> It can't hurt, but I also want to hear from @espindola

Sure. I wasn't suggesting otherwise :).

haaawk:
I run the clean compilation before and after the PR in dev and release

dev:

before
real    14m12.106s
user    197m3.745s
sys     11m20.727s

after
real    13m53.203s
user    199m25.441s
sys     11m23.304s

release:

before
real    27m28.252s
user    382m17.279s
sys     16m4.258s

after
real    28m46.035s
user    387m6.294s
sys     16m15.970s

The user time is ~1% more after this PR, in both cases.

haaawk:
@espindola ping

espindola:

> @espindola don't such changes increase object sizes and work for the linker?

Yes. They also mean that the initialization is seen by every translation
unit. I would prefer using inline variables only for simplifications
(templates for instance) and avoid it when a plain extern in .hh + definition
in .cc will do.

I know it is tempting, but I think the real replacement for extern in
.hh + definition in .cc is modules, and we just have to wait for it.

Cheers,
Rafael

haaawk:
So what should we do with this PR @avikivity? Should I close it or is there a chance for it to be merged considering a little impact it has on compilation time?

haaawk:
ping

avikivity:
Seems like a net loss.  Is it worth 18 more seconds of build time? not IMO.

PR #6061, opened 2020-03-22
tests: move token_generation_for_shard to common code

glommer:
We now have a utils file for SSTables. This is potentially useful for
other tests.

As a matter of fact, this function is repeated right now for the
resharding test. And to add insult to injury, the version in the
resharding test has the parameters shard and number of tokens flipped,
which although extremely confusing is the predictable outcome of
such repetition


avikivity:
Queued as dd65f7dcbb

PR #6065, opened 2020-03-23
redis: add exists command

syuu1228:
Add exists command that returns key availablitiy.
We only support single argument for now, but upstream
does support multiple keys since Redis 3.0.3.

see: https://redis.io/commands/exists

### DIFF
--- a/redis/command_factory.cc
+++ b/redis/command_factory.cc
@@ -35,6 +35,7 @@ shared_ptr<abstract_command> command_factory::create(service::storage_proxy& pro
         { "ping",  [] (service::storage_proxy& proxy, request&& req) { return commands::ping::prepare(proxy, std::move(req)); } }, 
         { "select",  [] (service::storage_proxy& proxy, request&& req) { return commands::select::prepare(proxy, std::move(req)); } }, 
         { "get",  [] (service::storage_proxy& proxy, request&& req) { return commands::get::prepare(proxy, std::move(req)); } }, 
+        { "exists", [] (service::storage_proxy& proxy, request&& req) { return commands::exists::prepare(proxy, std::move(req)); } },
         { "set",  [] (service::storage_proxy& proxy, request&& req) { return commands::set::prepare(proxy, std::move(req)); } }, 
         { "del",  [] (service::storage_proxy& proxy, request&& req) { return commands::del::prepare(proxy, std::move(req)); } }, 
         { "echo",  [] (service::storage_proxy& proxy, request&& req) { return commands::echo::prepare(proxy, std::move(req)); } },
--- a/redis/commands.cc
+++ b/redis/commands.cc
@@ -53,6 +53,27 @@ future<redis_message> get::execute(service::storage_proxy& proxy, redis::redis_o
     });
 }
 
+exists::exists(bytes&& name, std::vector<bytes>&& keys) : abstract_command(std::move(name)) , _keys(std::move(keys)) {}
+
+shared_ptr<abstract_command> exists::prepare(service::storage_proxy& proxy, request&& req) {
+    if (req.arguments_size() < 1) {
+        throw wrong_arguments_exception(1, req.arguments_size(), req._command);
+    }
+    return seastar::make_shared<exists>(std::move(req._command), std::move(req._args));
+}
+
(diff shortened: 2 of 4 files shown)


redis/commands.cc
```diff
@@ -53,6 +53,22 @@ future<redis_message> get::execute(service::storage_proxy& proxy, redis::redis_o
...
 
+shared_ptr<abstract_command> exists::prepare(service::storage_proxy& proxy, request&& req) {
+    if (req.arguments_size() != 1) {
+        throw wrong_arguments_exception(1, req.arguments_size(), req._command);
+    }
+    return seastar::make_shared<exists> (std::move(req._command), std::move(req._args[0]));
+}
+
+future<redis_message> exists::execute(service::storage_proxy& proxy, redis::redis_options& options, service_permit permit) {
+    return redis::read_strings(proxy, options, _key, permit).then([] (auto result) {
```
avikivity: Please don't use generic lambas (`auto`), they are harder to use with an IDE because it cannot guess the type of `result`.

redis/commands.cc
```diff
@@ -53,6 +53,22 @@ future<redis_message> get::execute(service::storage_proxy& proxy, redis::redis_o
...
+        if (result->has_result()) {
+            return redis_message::one();
```
avikivity: According to https://redis.io/commands/exists you should return the number of keys that exist, not 0/1.

syuu1228:
@avikivity Fixed 'auto' on result, implemented the command to return number of keys exist (previously it just return single key availability).
Please review it again.

penberg:
@syuu1228 Please add a test in `redis-test` for this and verify that it passes against both Redis and Scylla. This is critical in ensuring that as we grow the Redis API over time, we don't introduce subtle compatibility bugs, which are hard to track down later as the user base of the API grows.

redis/command_factory.cc
```diff
@@ -35,6 +35,7 @@ shared_ptr<abstract_command> command_factory::create(service::storage_proxy& pro
         { "ping",  [] (service::storage_proxy& proxy, request&& req) { return commands::ping::prepare(proxy, std::move(req)); } }, 
         { "select",  [] (service::storage_proxy& proxy, request&& req) { return commands::select::prepare(proxy, std::move(req)); } }, 
         { "get",  [] (service::storage_proxy& proxy, request&& req) { return commands::get::prepare(proxy, std::move(req)); } }, 
+        { "exists", [] (service::storage_proxy& proxy, request&& req) { return commands::exists::prepare(proxy, std::move(req)); } },
```
penberg, line 38: Not something for this pull request, but this command lookup table looks pretty verbose, especially since the number of commands is in the hundreds. Perhaps we should simplify it somehow?

redis/commands.hh
```diff
@@ -39,6 +39,18 @@ public:
...
 
+class exists : public abstract_command {
+    std::vector<bytes> _keys;
+    size_t _count = 0;
+public:
+    static shared_ptr<abstract_command> prepare(service::storage_proxy& proxy, request&& req);
+    exists(bytes&& name, std::vector<bytes>&& keys)
+        : abstract_command(std::move(name))
+        , _keys(std::move(keys)) {
+    }
```
penberg, line 50: Let's move the constructor definition to the `commands.cc` file to keep the implementation in one place.

redis/commands.cc
```diff
@@ -53,6 +53,25 @@ future<redis_message> get::execute(service::storage_proxy& proxy, redis::redis_o
     });
 }
 
+shared_ptr<abstract_command> exists::prepare(service::storage_proxy& proxy, request&& req) {
+    if (req.arguments_size() < 1) {
+        throw wrong_arguments_exception(1, req.arguments_size(), req._command);
+    }
+    return seastar::make_shared<exists> (std::move(req._command), std::move(req._args));
```
penberg, line 60: Extra space between `>` and `(`.

syuu1228:
@penberg added test and also fixed minor fixes on the code, please review again.

penberg:
Queued as c8679ce4e5469e76a4880382bee503097dd21463, thanks @syuu1228!

PR #6076, opened 2020-03-24
move rewrite_sstables to the compaction_manager

glommer:
move rewrite_sstables to the compaction_manager
    
There is no reason why the table code has to be aware of the efforts of
rewriting (cleanup, scrub, upgrade) an SSTable versus compacting it.
    
Rewrite is special, because we need to do it one SSTable at a time,
without lumping it together. However, the compaction manager is totally
capable of doing that itself. If we do that, the special
"table::rewrite_sstables" can be killed.
    
This code would maybe be better off as a thread, where we wouldn't need
to keep state. However there are some methods like maybe_stop_on_error()
that expect a future so I am leaving this be for now. This is a cleanup
that can be done later.

### DIFF
--- a/sstables/compaction_manager.cc
+++ b/sstables/compaction_manager.cc
@@ -612,44 +612,44 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
         sstables_ptr->pop_back();
 
         return repeat([this, task, options, sst = std::move(sst)] () mutable {
-        column_family& cf = *task->compacting_cf;
-        auto sstable_level = sst->get_sstable_level();
-        auto run_identifier = sst->run_identifier();
-        auto descriptor = sstables::compaction_descriptor({ sst }, sstable_level,
-                    sstables::compaction_descriptor::default_max_sstable_bytes, run_identifier, options);
+            column_family& cf = *task->compacting_cf;
+            auto sstable_level = sst->get_sstable_level();
+            auto run_identifier = sst->run_identifier();
+            auto descriptor = sstables::compaction_descriptor({ sst }, sstable_level,
+                        sstables::compaction_descriptor::default_max_sstable_bytes, run_identifier, options);
 
-        auto compacting = make_lw_shared<compacting_sstable_registration>(this, descriptor.sstables);
-        // Releases reference to cleaned sstable such that respective used disk space can be freed.
-        descriptor.release_exhausted = [compacting] (const std::vector<sstables::shared_sstable>& exhausted_sstables) {
-            compacting->release_compacting(exhausted_sstables);
-        };
+            auto compacting = make_lw_shared<compacting_sstable_registration>(this, descriptor.sstables);
+            // Releases reference to cleaned sstable such that respective used disk space can be freed.
+            descriptor.release_exhausted = [compacting] (const std::vector<sstables::shared_sstable>& exhausted_sstables) {
+                compacting->release_compacting(exhausted_sstables);
+            };
 
-        _stats.pending_tasks--;
-        _stats.active_tasks++;
-        task->compaction_running = true;
-        compaction_backlog_tracker user_initiated(std::make_unique<user_initiated_backlog_tracker>(_compaction_controller.backlog_of_shares(200), _available_memory));
-        return do_with(std::move(user_initiated), [this, &cf, descriptor = std::move(descriptor)] (compaction_backlog_tracker& bt) mutable {
-            return with_scheduling_group(_scheduling_group, [this, &cf, descriptor = std::move(descriptor)] () mutable {
-                return cf.run_compaction(std::move(descriptor));
-            });
-        }).then_wrapped([this, task, compacting = std::move(compacting)] (future<> f) mutable {
-            task->compaction_running = false;
-            _stats.active_tasks--;
-            if (!can_proceed(task)) {
-                maybe_stop_on_error(std::move(f), stop_iteration::yes);
-                return make_ready_future<stop_iteration>(stop_iteration::yes);
-            }
-            if (maybe_stop_on_error(std::move(f))) {
-                _stats.errors++;
-                _stats.pending_tasks++;
-                return put_task_to_sleep(task).then([] {
-                    return make_ready_future<stop_iteration>(stop_iteration::no);
+            _stats.pending_tasks--;
+            _stats.active_tasks++;
+            task->compaction_running = true;
+            compaction_backlog_tracker user_initiated(std::make_unique<user_initiated_backlog_tracker>(_compaction_controller.backlog_of_shares(200), _available_memory));
+            return do_with(std::move(user_initiated), [this, &cf, descriptor = std::move(descriptor)] (compaction_backlog_tracker& bt) mutable {
+                return with_scheduling_group(_scheduling_group, [this, &cf, descriptor = std::move(descriptor)] () mutable {
+                    return cf.run_compaction(std::move(descriptor));
                 });
-            }
-            _stats.completed_tasks++;
-            reevaluate_postponed_compactions();
-            return make_ready_future<stop_iteration>(stop_iteration::yes);
-        });
+            }).then_wrapped([this, task, compacting = std::move(compacting)] (future<> f) mutable {
+                task->compaction_running = false;
+                _stats.active_tasks--;
+                if (!can_proceed(task)) {
+                    maybe_stop_on_error(std::move(f), stop_iteration::yes);
+                    return make_ready_future<stop_iteration>(stop_iteration::yes);
+                }
+                if (maybe_stop_on_error(std::move(f))) {
+                    _stats.errors++;
+                    _stats.pending_tasks++;
+                    return put_task_to_sleep(task).then([] {
+                        return make_ready_future<stop_iteration>(stop_iteration::no);
+                    });
+                }
+                _stats.completed_tasks++;
+                reevaluate_postponed_compactions();
+                return make_ready_future<stop_iteration>(stop_iteration::yes);
+            });
         });
     }).finally([this, task, sstables = std::move(sstables)] {
         _stats.pending_tasks -= sstables->size();
--- a/sstables/compaction_manager.cc
+++ b/sstables/compaction_manager.cc
@@ -596,17 +596,28 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
     task->compacting_cf = cf;
     task->type = options.type();
     _tasks.push_back(task);
-    _stats.pending_tasks++;
 
-    task->compaction_done = repeat([this, task, options, get_func = std::move(get_func)] () mutable {
+    auto sstables = std::make_unique<std::vector<sstables::shared_sstable>>(get_func(*cf));
+    auto sstables_ptr = sstables.get();
+    _stats.pending_tasks += sstables->size();
+
+    task->compaction_done = do_until([sstables_ptr] { return sstables_ptr->empty(); }, [this, task, options, sstables_ptr] () mutable {
+
         // FIXME: lock cf here
         if (!can_proceed(task)) {
-            _stats.pending_tasks--;
-            return make_ready_future<stop_iteration>(stop_iteration::yes);
+            return make_ready_future<>();
         }
+
+        auto sst = sstables_ptr->back();
+        sstables_ptr->pop_back();
+
+        return repeat([this, task, options, sst = std::move(sst)] () mutable {
         column_family& cf = *task->compacting_cf;
-        sstables::compaction_descriptor descriptor = sstables::compaction_descriptor(get_func(cf));
-        descriptor.options = options;
+        auto sstable_level = sst->get_sstable_level();
+        auto run_identifier = sst->run_identifier();
+        auto descriptor = sstables::compaction_descriptor({ sst }, sstable_level,
+                    sstables::compaction_descriptor::default_max_sstable_bytes, run_identifier, options);
+
         auto compacting = make_lw_shared<compacting_sstable_registration>(this, descriptor.sstables);
         // Releases reference to cleaned sstable such that respective used disk space can be freed.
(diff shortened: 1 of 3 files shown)


avikivity:
Please use descriptive titles.

sstables/compaction_manager.cc
```diff
@@ -596,39 +596,60 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
...
+    auto wrk = std::make_unique<work>(get_func(*cf));
+    auto w = wrk.get();
+    _stats.pending_tasks += w->total_pending;
+
+    task->compaction_done = do_for_each(w->sstables.begin(), w->sstables.end(),
+            [this, task, options, w] (sstables::shared_sstable sst) mutable {
+
         // FIXME: lock cf here
         if (!can_proceed(task)) {
-            _stats.pending_tasks--;
```
denesb: I think it would be better to decrement `_stats.pending_tasks` here as before, instead of at the end. It is easier to keep track of all the increments/decrements.

sstables/compaction_manager.cc
```diff
@@ -596,39 +596,60 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
...
+        auto descriptor = sstables::compaction_descriptor({ sst }, sstable_level,
+                    sstables::compaction_descriptor::default_max_sstable_bytes, run_identifier, options);
+
         auto compacting = make_lw_shared<compacting_sstable_registration>(this, descriptor.sstables);
         // Releases reference to cleaned sstable such that respective used disk space can be freed.
         descriptor.release_exhausted = [compacting] (const std::vector<sstables::shared_sstable>& exhausted_sstables) {
             compacting->release_compacting(exhausted_sstables);
         };
 
+        w->total_pending--;
```
denesb: This additional counter is confusing.

sstables/compaction_manager.cc
```diff
@@ -596,39 +596,60 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
...
         _stats.pending_tasks--;
         _stats.active_tasks++;
         task->compaction_running = true;
+
         compaction_backlog_tracker user_initiated(std::make_unique<user_initiated_backlog_tracker>(_compaction_controller.backlog_of_shares(200), _available_memory));
         return do_with(std::move(user_initiated), [this, &cf, descriptor = std::move(descriptor)] (compaction_backlog_tracker& bt) mutable {
             return with_scheduling_group(_scheduling_group, [this, &cf, descriptor = std::move(descriptor)] () mutable {
-                return cf.rewrite_sstables(std::move(descriptor));
+                return cf.run_compaction(std::move(descriptor));
```
denesb: It's confusing that `table` has both `run_compaction()` and `compact_sstables()`, the former calling the latter. Not relevant to this patch.

sstables/compaction_manager.cc
```diff
@@ -639,7 +660,9 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
             reevaluate_postponed_compactions();
             return make_ready_future<stop_iteration>(stop_iteration::yes);
         });
-    }).finally([this, task] {
+        });
+    }).finally([this, task, wrk = std::move(wrk)] {
+        _stats.pending_tasks -= wrk->total_pending;
```
denesb: Shouldn't this be 0 at this point?

sstables/compaction_manager.cc
glommer: why do you think this is unnecessary ? If we are pushing 100 SSTables to be rewritten, there should be 100 pending tasks.

I can change it for the sake of "was like this before", but I don't think this is correct.

sstables/compaction_manager.cc
glommer: we'll get there.

sstables/compaction_manager.cc
glommer: if we ran all tasks yes, if it was stopped in the middle then no.

This is the finally block

sstables/compaction_manager.cc
glommer: If we agree that all SSTables in the job are "one task", then this is not needed.
I don't think that is correct, and then this is needed.

The reason this is needed is because we can stop processing the list in the middle (for instance if the compaction_manager was stopped temporarily)  in which case the finally block has to keep track of what is left to decrement accordingly

sstables/compaction_manager.cc
denesb: I'm trying to understand what is the reason we cannot use `_stats.pending_tasks` directly to account the number of pending tasks. Why we need an additional variable for the same purpose?

glommer: That works well if all tasks are eventually processed. The compaction manager can be stopped in which case tasks are dropped in the middle. If we rely on just `_stats.pending_tasks` we will never be able to subtract those

denesb: In that case can I propose to replace `work` with a container and just pop sstables from it as we process them? Then it is much more clear what is happening. In the `finally()` block we can just do `_pending_task -= pending_sstables.size()`.

denesb: I just also realized that since we use `do_for_each()`, we will have an effective concurrency of one, just before. This is even better as we don't create reactor tasks for N-1 sstables.

glommer: I am okay with that, except I would not be able to use `do_for_each`, since popping from live containers that are iterated usually confuses `begin()` and `end()`. Maybe repeat will work, although I don't quite like the nested `repeat`

This really should've been a thread.

glommer: nah, `do_until` works well

denesb: `do_until([&sstables] { return sstables.empty(); }, [&sstables] {
    sst = sstables.front();
    sstables.pop_front();
    //...
});
```

Comment (denesb) [reply sstables/compaction_manager.cc]: `back()` and `pop_back()` work just as well and are friendly to `std::vector`.

Comment (glommer) [reply sstables/compaction_manager.cc]: yeah, I was just rusty on remembering all primitives.
I coded that already

**Comment by glommer** [comment]
The debug test is failing for sstable_datafile_test. It is pre-existing, so I will force push this for reviewers. Will investigate in the meantime

File: sstables/compaction_manager.cc [hunk 5]
```diff
@@ -596,17 +596,28 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
...
         if (!can_proceed(task)) {
-            _stats.pending_tasks--;
-            return make_ready_future<stop_iteration>(stop_iteration::yes);
+            return make_ready_future<>();
         }
+
+        auto sst = w->back();
+        w->pop_back();
+
+        return repeat([this, task, options, sst = std::move(sst), w] () mutable {
```
denesb: Don't see w used here.

sstables/compaction_manager.cc
```diff
@@ -619,9 +630,9 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
         compaction_backlog_tracker user_initiated(std::make_unique<user_initiated_backlog_tracker>(_compaction_controller.backlog_of_shares(200), _available_memory));
         return do_with(std::move(user_initiated), [this, &cf, descriptor = std::move(descriptor)] (compaction_backlog_tracker& bt) mutable {
             return with_scheduling_group(_scheduling_group, [this, &cf, descriptor = std::move(descriptor)] () mutable {
-                return cf.rewrite_sstables(std::move(descriptor));
+                return cf.run_compaction(std::move(descriptor));
             });
-        }).then_wrapped([this, task, compacting = std::move(compacting)] (future<> f) mutable {
+        }).then_wrapped([this, task, compacting = std::move(compacting), w] (future<> f) mutable {
```
denesb: Don't see w used here.

sstables/compaction_manager.cc
```diff
@@ -596,17 +596,28 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
     task->compacting_cf = cf;
     task->type = options.type();
     _tasks.push_back(task);
-    _stats.pending_tasks++;
 
-    task->compaction_done = repeat([this, task, options, get_func = std::move(get_func)] () mutable {
+    auto sstables = std::make_unique<std::vector<sstables::shared_sstable>>(get_func(*cf));
+    auto w = sstables.get();
```
denesb: `w` is not a relevant name anymore.

glommer: you are right, because now we no longer update the counters. Patch will be simpler too.

Let me spend some time to track down the failure in debug mode and then I'll resend what is hopefully the final version

glommer:
Ok, the debug mode test is broken for a while - I was trying a bisect but I don't know when it broke.
I am off now for my fatherly duties so I will just force push this since it seems ready to go.

I'll go back to the test later, also letting @raphaelsc know about it in case he has any idea

sstables/compaction_manager.cc
```diff
@@ -596,50 +596,63 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
...
-                    return make_ready_future<stop_iteration>(stop_iteration::no);
+
+        auto sst = w->back();
+        w->pop_back();
+
+        return repeat([this, task, options, sst = std::move(sst)] () mutable {
+            column_family& cf = *task->compacting_cf;
+            auto sstable_level = sst->get_sstable_level();
+            auto run_identifier = sst->run_identifier();
+            auto descriptor = sstables::compaction_descriptor({ sst }, sstable_level,
```
raphaelsc: nitpick: move sst into descriptor?

glommer: can't move anything here because it will be retried in case of compaction failure.

sstables/compaction_manager.cc
```diff
@@ -596,50 +596,63 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
...
     task->type = options.type();
     _tasks.push_back(task);
-    _stats.pending_tasks++;
 
-    task->compaction_done = repeat([this, task, options, get_func = std::move(get_func)] () mutable {
+    auto sstables = std::make_unique<std::vector<sstables::shared_sstable>>(get_func(*cf));
+    auto w = sstables.get();
+    _stats.pending_tasks += sstables->size();
+
+    task->compaction_done = do_until([w] { return !w->empty(); }, [this, task, options, w] () mutable {
```
raphaelsc: do_until's interface is sometime counter-intuitive because its first parameter is stop condition, not the "keep going" condition, which is what we're most familiar with in programming, ex: while(condition) { do }
do_until([w] { return !w->empty(); } must be do_until([w] { return w->empty(); })

raphaelsc: I go on saying we should perhaps deprecate do_until(stop_condition, action) in favor of a new do_while(condition, action). I prefer the later much more.

raphaelsc: @glommer That's probably the reason behind the test failure in debug mode, right?

glommer: Maybe, thanks a lot Raphael! An update : I was probably compiling and running different binaries because Raphael and I found out that indeed the test was not failing before. Thinking it did thew me off track

(that's probably an artifact of the `test_g` vs `test` thing)

Updating

glommer:
Force-pushed. I fixed the condition that Raphael found and  tests pass in debug mode

sstables/compaction_manager.cc
```diff
@@ -639,7 +650,9 @@ future<> compaction_manager::rewrite_sstables(column_family* cf, sstables::compa
             reevaluate_postponed_compactions();
             return make_ready_future<stop_iteration>(stop_iteration::yes);
         });
-    }).finally([this, task] {
+        });
+    }).finally([this, task, sstables = std::move(sstables)] {
+        _stats.pending_tasks -= sstables->size();
```
raphaelsc: cool

sstables/compaction_manager.cc
denesb: This is still not solved.

denesb:
Why is there no way to un-resolve a comment, marked as resolved? This UI is so half-baked.

glommer:
@denesb how come? There is only one instance left where `w` is passed to the lambda, and that's where it is used so we can pop from the array. Please point out in the code again where do you see `w` being passed to the lambda without being used (maybe I pushed the wrong version?)

denesb:
> @denesb how come? There is only one instance left where `w` is passed to the lambda, and that's where it is used so we can pop from the array. Please point out in the code again where do you see `w` being passed to the lambda without being used (maybe I pushed the wrong version?)

It's not about to which lambda it is passed, my problem is with the name `w` itself. It used to be short for `work`, but not it is just a pointer to a list of sstables.

glommer:
I have not seen any comment from you complaining about that.

I will fix, but I think this is an over the top level of nitpicking. It's not like it's passed all over (anymore)

denesb:
I *did* comment on this https://github.com/scylladb/scylla/pull/6076#discussion_r397229044.

Since when are comments about naming are over-the-top level nitpicking?

glommer:
I did not see that comment, which is why I haven't addressed.
I don't think comments (plural, general) about naming are that. I think this one (singular, particular) is, because it is a temporary variable passed over and consumed very close to the define site.

I am finishing testing it again and will push an updated version soon

denesb:
We accept abbreviated, even single letter variable names in reduced scope. But abbreviated doesn't mean arbitrary. It should still be associable with the purpose.

I don't think such a simple change needs a complete re-test. I think just recompiling the .cc is enough to typo-check.

glommer:
I refuse to push anything without a unit test retest.
I have seen much simpler changes causing problems in the past.

glommer:
ping maintainers

bhalevy:
> ping maintainers

@glommer: ping @scylladb/scylla-maint
so they get a more targeted email message

glommer:
ping2 @scylladb/scylla-maint

PR #6079, opened 2020-03-25
migration_manager: Make sync_schema return error when node is down

### LINKED ISSUE #4791: migration_manager::sync_schema() may succeed even if schema was not synced
Currently used only when repair starts

It calls `migration_manager::submit_migration_task()` which calls `service::migration_task::run_may_throw()`, which has this:

```
    if (!gms::get_local_gossiper().is_alive(endpoint)) {
        mlogger.warn("Can't send migration request: node {} is down.", endpoint);
        return make_ready_future<>();
    }
```

So if one of the requested nodes is down we will succeed the sync, while we should fail. According to the doc:

```
// Makes sure that this node knows about all schema changes known by "nodes" that were made prior to this call.
```

asias:
sync_schema is supposed to make sure that this node knows about all
schema changes known by "nodes" that were made prior to this call.

Currently, when a node is down, the sync is sliently skipped.

To fix, add a flag to migration_task::run_may_throw to indicate that it
should fail if a node is down.

Fixes #4791

### DIFF
--- a/service/migration_task.cc
+++ b/service/migration_task.cc
@@ -51,11 +51,12 @@ namespace service {
 
 static logging::logger mlogger("migration_task");
 
-future<> migration_task::run_may_throw(const gms::inet_address& endpoint)
+future<> migration_task::run_may_throw(const gms::inet_address& endpoint, bool can_ignore_down_node)
 {
     if (!gms::get_local_gossiper().is_alive(endpoint)) {
-        mlogger.warn("Can't send migration request: node {} is down.", endpoint);
-        return make_ready_future<>();
+        auto msg = format("Can't send migration request: node {} is down.", endpoint);
+        mlogger.warn("{}", msg);
+        return can_ignore_down_node ? make_ready_future<>() : make_exception_future<>(std::runtime_error(msg));
     }
     netw::messaging_service::msg_addr id{endpoint, 0};
     return service::get_local_migration_manager().merge_schema_from(id).handle_exception([](std::exception_ptr e) {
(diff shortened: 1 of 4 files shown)


service/migration_task.cc
```diff
@@ -51,11 +51,12 @@ namespace service {
...
 {
     if (!gms::get_local_gossiper().is_alive(endpoint)) {
-        mlogger.warn("Can't send migration request: node {} is down.", endpoint);
-        return make_ready_future<>();
+        auto msg = format("Can't send migration request: node {} is down.", endpoint);
+        mlogger.warn("{}", msg);
+        return can_ignore_down_node ? make_ready_future<>() : make_exception_future<>(std::runtime_error(msg));
     }
     netw::messaging_service::msg_addr id{endpoint, 0};
     return service::get_local_migration_manager().merge_schema_from(id).handle_exception([](std::exception_ptr e) {
```
tgrabiec: handle_exception() below also swallows errors, we should fix that too. 

I wonder why we need to swallow `configuration_exception` at all.

asias: I have no idea. How about: 

```
--- a/service/migration_task.cc
+++ b/service/migration_task.cc
@@ -64,6 +64,7 @@ future<> migration_task::run_may_throw(const gms::inet_address& endpoint, bool c
             std::rethrow_exception(e);
         } catch (const exceptions::configuration_exception& e) {
             mlogger.error("Configuration exception merging remote schema: {}", e.what());
+            return make_exception_future<>(e);
         }
     });
 }
```

asias:
Changes in ef64f52152746bb3b70f4fc757e2d9ba1309fb34:
- Fix migration_task::run_may_throw. Do not swallow exception.

PR #6081, opened 2020-03-25
dist/docker: update SCYLLA_REPO_URL and VERSION

yaronkaikov:
update params for release 4.0

### DIFF
--- a/dist/docker/redhat/Dockerfile
+++ b/dist/docker/redhat/Dockerfile
@@ -5,8 +5,8 @@ MAINTAINER Avi Kivity <avi@cloudius-systems.com>
 ENV container docker
 
 # The SCYLLA_REPO_URL argument specifies the URL to the RPM repository this Docker image uses to install Scylla. The default value is the Scylla's unstable RPM repository, which contains the daily build.
-ARG SCYLLA_REPO_URL=http://downloads.scylladb.com/rpm/unstable/centos/master/latest/scylla.repo
-ARG VERSION=666.development
+ARG SCYLLA_REPO_URL=http://downloads.scylladb.com/rpm/unstable/centos/branch-4.0/latest/scylla.repo
+ARG VERSION=4.0.*
 
 ADD scylla_bashrc /scylla_bashrc
 


yaronkaikov:
@penberg @avikivity Need your advice here please
on master i created new parameter in Dockerfile with default value: `VERSION=666.development`
and later when we install Scylla during docker we have the command `yum install scylla-$VERSION`
Now for release (first time we have it) we don't have a dev version like master. so what i did is to change Dockerfile default parameter to: `VERSION=scylla` and then when we actually install scylla we will run `yum install $VERSION` 
This will also make us change the publish-docker.sh script in scylla-pkg to send proper version with the name scylla.
so i have 2 questions now. 
1) Maybe we should rename the parameter to SCYLLA_VERSION?
2) Should we change master to be identical?

What do you think?

dist/docker/redhat/Dockerfile
```diff
@@ -5,8 +5,8 @@ MAINTAINER Avi Kivity <avi@cloudius-systems.com>
 ENV container docker
 
 # The SCYLLA_REPO_URL argument specifies the URL to the RPM repository this Docker image uses to install Scylla. The default value is the Scylla's unstable RPM repository, which contains the daily build.
-ARG SCYLLA_REPO_URL=http://downloads.scylladb.com/rpm/unstable/centos/master/latest/scylla.repo
-ARG VERSION=666.development
+ARG SCYLLA_REPO_URL=http://downloads.scylladb.com/rpm/unstable/centos/branch-4.0/latest/scylla.repo
+ARG VERSION=scylla
```
penberg: If you set `VERSION` to `4.0.*`, we will always install the _latest_ from the unstable repository, which seems like a reasonable default to me.

The release pipelines will pass the stable repository URL and a specific URL.

yaronkaikov: Great. Thanks @penberg

penberg:
Did you verify this with `docker build`? I only tested `yum install scylla-4.0.*` in bash so let's make sure it works from `Dockerfile` too.

yaronkaikov:
Yes, i run locally `docker build --no-cache -t scylladb/scylla:4.0.rc0 .` so default ARGs should take place. 
during the build i saw it is trying to install 
`scylla               x86_64 4.0.rc0-0.20200325.78f5afec30 scylla         6.5 k`

Once it finished i run the commands:

`docker run -d --name scylla scylladb/scylla:4.0.rc0` and  
`docker exec scylla scylla --version` then got the following output:
4.0.rc0-0.20200325.78f5afec30

penberg:
Queued to `next-4.0` as 282cd0df7c2f670d1222f84ea1eba50b0ef552e6

PR #6090, opened 2020-03-27
cdc+counters: make counters work in CDC

jul-stas:
This is a simplified approach - without shards info. Postimage is calculated as sum of preimage and delta.

Unit test is on the way, this is a moment for last RFC.


jul-stas:
I'm closing this because fundamental flaw was found (logs are being generated for deleted counters). A heavyweight solution is on the way.

PR #6107, opened 2020-03-31
Resharding through compact sstables

glommer:
This patchseries is part of my effort to make resharding less special - and hopefully less problematic.
The next steps are a bit heavy, so I'd like to, if possible, get this out of the way.

After these two patches, there is no more need to ever call reshard_sstables: compact_sstables will
do, and it will be able to recognize resharding compactions.

To do that we need to unify the creator function, which is trivially done by adding a shard parameter
to regular compactions as well: they can just ignore it. I have considered just making the compaction_descriptor have a virtual create() function and specializing it, but because we have to store the creator in the compaction object I decided to keep the virtual function for now. 

In a later cleanup step, if we can for instance store the entire compaction_descriptor object in the compaction object we could do that.

Tests: unit tests (dev), dtest (resharding.py)

### DIFF
--- a/sstables/compaction.hh
+++ b/sstables/compaction.hh
@@ -84,6 +84,18 @@ namespace sstables {
         compaction_type type() const;
     };
 
+    struct compaction_completion_desc {
+        std::vector<shared_sstable> input_sstables;
+        std::vector<shared_sstable> output_sstables;
+        // Set of compacted partition ranges that should be invalidated in the cache.
+        dht::partition_range_vector ranges_for_cache_invalidation;
+    };
+
+    // creates a new SSTable for a given shard
+    using creator_fn = std::function<shared_sstable(shard_id shard)>;
+    // Replaces old sstable(s) by new one(s) which contain all non-expired data.
+    using replacer_fn = std::function<void(compaction_completion_desc)>;
+
     struct compaction_descriptor {
         // List of sstables to be compacted.
         std::vector<sstables::shared_sstable> sstables;
@@ -100,6 +112,9 @@ namespace sstables {
         // The options passed down to the compaction code.
         // This also selects the kind of compaction to do.
         compaction_options options = compaction_options::make_regular();
+
+        creator_fn creator;
+        replacer_fn replacer;
 
         compaction_descriptor() = default;
 
@@ -190,16 +205,6 @@ namespace sstables {
         }
     };
 
-    struct compaction_completion_desc {
-        std::vector<shared_sstable> input_sstables;
-        std::vector<shared_sstable> output_sstables;
-        // Set of compacted partition ranges that should be invalidated in the cache.
-        dht::partition_range_vector ranges_for_cache_invalidation;
-    };
-
-    // Replaces old sstable(s) by new one(s) which contain all non-expired data.
-    using replacer_fn = std::function<void(compaction_completion_desc)>;
-
     // Compact a list of N sstables into M sstables.
     // Returns info about the finished compaction, which includes vector to new sstables.
     //
@@ -212,8 +217,7 @@ namespace sstables {
     // If descriptor.cleanup is true, mutation that doesn't belong to current node will be
     // cleaned up, log messages will inform the user that compact_sstables runs for
     // cleaning operation, and compaction history will not be updated.
-    future<compaction_info> compact_sstables(sstables::compaction_descriptor descriptor, column_family& cf,
-        std::function<shared_sstable()> creator, replacer_fn replacer);
+    future<compaction_info> compact_sstables(sstables::compaction_descriptor descriptor, column_family& cf);
 
     // Compacts a set of N shared sstables into M sstables. For every shard involved,
     // i.e. which owns any of the sstables, a new unshared sstable is created.
--- a/sstables/compaction.hh
+++ b/sstables/compaction.hh
@@ -48,9 +48,11 @@ namespace sstables {
         struct scrub {
             bool skip_corrupted;
         };
+        struct reshard {
+        };
 
     private:
-        using options_variant = std::variant<regular, cleanup, upgrade, scrub>;
+        using options_variant = std::variant<regular, cleanup, upgrade, scrub, reshard>;
 
     private:
         options_variant _options;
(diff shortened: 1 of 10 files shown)


sstables/compaction.hh
```diff
@@ -84,6 +90,18 @@ namespace sstables {
...
 
+    struct compaction_completion_desc {
+        std::vector<shared_sstable> input_sstables;
+        std::vector<shared_sstable> output_sstables;
+        // Set of compacted partition ranges that should be invalidated in the cache.
+        dht::partition_range_vector ranges_for_cache_invalidation;
+    };
+
+    // creates a new SSTable for a given shard
+    using creator_fn = std::function<shared_sstable(unsigned shard)>;
```
bhalevy: why not `shard_id shard` as it used to be?

test/boost/sstable_resharding_test.cc
```diff
@@ -81,7 +81,10 @@ void run_sstable_resharding_test() {
     auto filter_fname = sstables::test(sst).filename(component_type::Filter);
     uint64_t bloom_filter_size_before = file_size(filter_fname).get0();
 
-    auto creator = [&env, &cf, &tmp, version] (shard_id shard) mutable {
+
```
bhalevy: nit: extraneous empty line

test/boost/sstable_resharding_test.cc
```diff
@@ -81,7 +81,10 @@ void run_sstable_resharding_test() {
...
+    auto descriptor = sstables::compaction_descriptor({sst}, 0, std::numeric_limits<uint64_t>::max());
+    descriptor.options = sstables::compaction_options::make_reshard();
```
bhalevy: How about adding `options` to `compaction_descriptor` constructor, along with `sstables`, `level`, and `max_sstable_bytes`? (and default to `make_regular` as it is now)?
It looks like it could come handy in several places and it can save initializing options and then overwriting it.

test/lib/sstable_utils.cc
```diff
@@ -121,6 +121,14 @@ token_generation_for_shard(unsigned tokens_to_generate, unsigned shard,
     return key_and_token_pair;
 }
 
+future<compaction_info> compact_sstables(sstables::compaction_descriptor descriptor, column_family& cf, std::function<shared_sstable()> creator, replacer_fn replacer) {
+    descriptor.creator = [creator = std::move(creator)] (unsigned dummy) mutable {
```
bhalevy: `shard_id` dummy

test/lib/sstable_utils.hh
```diff
@@ -96,6 +96,8 @@ token_generation_for_shard(unsigned tokens_to_generate, unsigned shard,
 std::vector<std::pair<sstring, dht::token>>
 token_generation_for_current_shard(unsigned tokens_to_generate);
 
+future<compaction_info> compact_sstables(sstables::compaction_descriptor descriptor, column_family& cf, std::function<shared_sstable()> creator, replacer_fn replacer);
```
bhalevy: I thought of suggesting defining `replacer = replacer_fn_no_op()` as the default value to save it in most call sites, but you might be against default values.  Up to you...

sstables/compaction.hh
```diff
@@ -45,12 +45,14 @@ namespace sstables {
...
         };
+        struct reshard {
+        };
         struct scrub {
             bool skip_corrupted;
         };
 
     private:
-        using options_variant = std::variant<regular, cleanup, upgrade, scrub>;
+        using options_variant = std::variant<reshard, regular, cleanup, upgrade, scrub>;
```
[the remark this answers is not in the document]
glommer: ok, makes sense

test/boost/sstable_3_x_test.cc
```diff
@@ -2999,12 +2999,14 @@ static flat_mutation_reader compacted_sstable_reader(test_env& env, schema_ptr s
     tmpdir tmp;
     auto sstables = open_sstables(env, s, format("test/resource/sstables/3.x/uncompressed/{}", table_name), generations);
     auto new_generation = generations.back() + 1;
-    auto new_sstable = [s, &tmp, &env, new_generation] {
+
+    auto desc = sstables::compaction_descriptor(std::move(sstables));
+    desc.creator = [s, &tmp, &env, new_generation] (unsigned dummy) {
```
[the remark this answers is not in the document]
glommer: thanks, makes sense

sstables/compaction.hh
glommer: oversight.

test/boost/sstable_resharding_test.cc
glommer: tnx!

test/lib/sstable_utils.hh
glommer: I am not against a default in this case. Suggestion makes sense.

test/boost/sstable_resharding_test.cc
glommer: I considered, the main problem is that there are many callers we'd have to patch. I am not sure it is worth the effort.

glommer: I tried and indeed there are many sites. @bhalevy I don't dislike the idea.
But how about we do this in a follow up patch? It's technically not related to what I am trying to do here. I am happy to clean that up separately but I'd like to get those in

glommer:
v2: merged comments from Benny
One comment was not merged: I am not touching the constructor for compaction_descriptor as I consider that to be worthy of a separate patch: it's unrelated to what I am doing here and touches way too many sites.

Force-pushed the branch

test/boost/sstable_resharding_test.cc
raphaelsc: Just realized last argument of c__d's constructor is c__o default'd to regular, so we could do:
sstables::compaction_descriptor({sst}, 0, std::numeric_limits<uint64_t>::max(), utils::make_random_uuid(), sstables::compaction_options::make_reshard());
I think we can postpone this to a followup cleanup patch.

bhalevy: @glommer @raphaelsc sure, this is mostly syntactic sugar. let's do that in a follow-up patch then.

PR #6115, opened 2020-04-01
fixing --experimental flag option from "on" to "1"

guy9:
When trying to run with "--experimental=on" I get:
File "/usr/lib64/python3.6/argparse.py", line 2318, in _check_value
   raise ArgumentError(action, msg % args)
argparse.ArgumentError: argument --experimental: invalid choice: 'on' (choose from '0', '1')

### DIFF
--- a/docs/alternator/alternator.md
+++ b/docs/alternator/alternator.md
@@ -23,7 +23,7 @@ Alternator on port 8000 - the traditional port used by DynamoDB.
 
 Alternator uses Scylla's LWT feature, which is currently considered
 experimental and needs to be seperately enabled as well, e.g. with the
-"`--experimental=on`" option.
+"`--experimental=1`" option.
 
 By default, Scylla listens on this port on all network interfaces.
 To listen only on a specific interface, pass also an "`alternator-address`"


tzach:
https://docs.scylladb.com/operating-scylla/scylla-yaml/#enable-all-experimental-features

nyh:
We have in both dist/docker/redhat/scyllasetup.py and alternator-test/run the "--experimental=on", and it works! What changed?
By the way, LWT is no longer experimental,so when we commit that change (if I haven't done it, I'll do it now) we can completely drop this line instead of fixing it.

nyh:
I just realized in the issue description above you quoted a Python error message... But why Python? Scylla is a C++ program... What exactly did you try to run?

nyh:
Ok, I know what the problem is... You didn't run Scylla, you ran the docker image and it ran the script docs/alternator/alternator.md. This script has:

        if self._experimental == "1":
            args += ["--experimental=on"]

I don't know why this script has a different convention for experimental... @penberg ?
In any case, since LWT is no longer experimental, I think we should close this PR and drop any mention of experimental in the Alternator documentation. I'm opening an issue for this right now.

nyh:
Opened https://github.com/scylladb/scylla/issues/6118.
Closing this PR because it's invalid. Scylla indeed uses "--experimental=on" and alternator.md, and it shouldn't be fixed. The docker image indeed uses "--experimental=1" but that's correctly explained in getting-started.md.

PR #6117, opened 2020-04-01
Use special partitioner for CDC Log

haaawk:
CDC has to create CDC streams that are co-located with corresponding BaseTable data. This is not always easy. Especially for small vnodes. This PR introduces new partitioner which allows us to easily find such stream ids that the stream belongs to a given vnode and shard.

The idea is that a partitioner accepts only keys that are a blob composed of two int64 numbers. The first number is the token of the key.

Tests: unit(dev), dtests(CDC)


cdc/generation.cc
```diff
@@ -173,118 +167,34 @@ topology_description generate_topology_description(
...
-            int last_set_stream_idx = ss.size() - 1;
-            while (last_set_stream_idx > -1 && !ss[last_set_stream_idx].is_set()) {
-                --last_set_stream_idx;
-            }
-
-            if (last_set_stream_idx == -1) {
+        static thread_local std::mt19937_64 rand_gen(std::random_device().operator()());
+        static thread_local std::uniform_int_distribution<int64_t> rand_dist(std::numeric_limits<int64_t>::min());
+        dht::sharder sharder(entry.streams.size(), entry.sharding_ignore_msb);
+        int64_t next_token_for_unallocated_shard = dht::token::to_int64(prev_token) + 1;
```
avikivity, line 173: Why do you need the random stuff? You can just pick the vnode's end token and then use walk backwards from there.

cdc/generation.cc
```diff
@@ -173,118 +167,34 @@ topology_description generate_topology_description(
...
-            }
-            // Replace 'unset' stream ids with indexes above last_set_stream_idx
-            for (int s_idx = ss.size() - 1; s_idx > last_set_stream_idx; --s_idx) {
-                if (ss[s_idx].is_set()) {
-                    some_stream = ss[s_idx];
+                        "CDC wasn't able to generate any stream on shard {} on vnode ({}, {}]. "
+                        "This shard will have its stream located on another shard in this vnode instead. "
+                        "This might lead to worse performance.",
+                        shard_idx, prev_token, entry.token_range_end);
+                token = next_token_for_unallocated_shard;
```
avikivity, line 184: How is it possible that we'll fail here?

cdc/generation.cc
haaawk: It's not random. It's the other end of the range. We need both lower and upper bound when we're walking so that we can wrap around if needed. I'm currently walking upwards but if you prefer I can walk downwards from upper_bound to dht::token::to_int64(prev_token) + 1 which is the end of the previous vnode.

cdc/generation.cc
haaawk: This happens when a vnode is small and it does not have tokens for each shard.

cdc/generation.cc
haaawk: Maybe I can rename it to end_of_previous_vnode to make it more clear?

haaawk: Or are you asking about rand_dist?

kbr-: `rand_dist` is probably used to avoid stream ID collisions between different generations.
However, it's weird that it's initialized here, when it's not used until much later below.

haaawk: It's used in a loop and I didn't want to initialize static variable in a loop.

dht/cdc_partitioner.cc
```diff
@@ -0,0 +1,57 @@
+/*
```
nyh, line 1: Question on the commit message (because I can't put it anywhere else... github...):
Can you please explain better what's the idea of having two-integer partition keys, where only the first integer is the token? Until now I was under the impression that what we're trying to do is that the partition key - the stream id - will be the same as the token? What do we need that second integer for and if we do, why is it not part of the token?

configure.py
```diff
@@ -699,6 +699,7 @@ def find_headers(repodir, excluded_dirs):
                 'dht/i_partitioner.cc',
                 'dht/token.cc',
                 'dht/murmur3_partitioner.cc',
+                'dht/cdc_partitioner.cc',
```
nyh, line 702: To me it makes more sense to put cdc_partitioner in cdc/. Yes, it's a partitioner, but it's a partitioner to be used *only* by CDC so will be nice to put it with the rest of the cdc-specific code. If you think it could be used outside CDC, then we should find a better name for it (maybe "identity_partitioner"?), not "cdc_partitioner".

dht/cdc_partitioner.hh
```diff
@@ -0,0 +1,41 @@
...
+#pragma once
+
+#include "i_partitioner.hh"
+
+namespace dht {
+
+class cdc_partitioner final : public i_partitioner {
+public:
+    cdc_partitioner() = default;
+    virtual const sstring name() const { return "com.scylladb.dht.CDCPartitioner"; }
```
nyh, line 31: This is a duplicating of the string in "registrator". Can we have name() call registrator, or the other way around, or put this string once in a string constant?

cdc/generation.cc
```diff
@@ -173,118 +167,34 @@ topology_description generate_topology_description(
...
+                } else if (next_token_for_unallocated_shard == std::numeric_limits<int64_t>::max()){
+                    next_token_for_unallocated_shard = std::numeric_limits<int64_t>::min();
                 } else {
-                    ss[s_idx] = some_stream;
+                    ++next_token_for_unallocated_shard;
                 }
+            } else {
+                token = dht::token::to_int64(t);
             }
+            entry.streams[shard_idx] = {token, rand_dist(rand_gen)};
```
nyh, line 195: Continuing Avi's question above: Why do we need this random number here? Why can't we just take "token"?
This is also related to my question on the previous patch - why do we need this random second half of the stream id if anyway the token is just the first part? Can't we just lose the second part? Or if we can't, what does its randomness achieve?

haaawk:
Sorry - I accidentally removed the branch and that closed the PR. Reopening.

dht/cdc_partitioner.cc
haaawk: I will improve the commit message and add comments. Thanks.

configure.py
haaawk: I will move it.

dht/cdc_partitioner.hh
haaawk: Will give it a try.

cdc/generation.cc
haaawk: There are multiple generations of CDC streams. Every time a topology change appears the new generation is created. There can be two streams, one is old generation and another in new generation, that share the same token but we want their stream_ids to be different because those are different streams. They just happen to be represented by the same token. Does that make sense @nyh?

cdc/generation.cc
avikivity: Then it doesn't need a stream for each shard. Failure is impossible.

cdc/generation.cc
avikivity: > It's not random. 

I'm referring to 

```c++
        static thread_local std::mt19937_64 rand_gen(std::random_device().operator()());
        static thread_local std::uniform_int_distribution<int64_t> rand_dist(std::numeric_limits<int64_t>::min());
```

configure.py
avikivity: Agree

cdc/generation.cc
haaawk: I guess you're right because the Base Table won't have any data belonging to such shards as well. Will change this. Thanks.

cdc/generation.cc
nyh: I don't think I understand. If the topology change did not involve a specific vnode (changed neither of its ends), then we might very well continue the same stream, and don't need a new one. If it *did* involve a specific vnode, then if it changed the beginning of the range it will change the begin token itself - and if it changed the end of the range, isn't it fine that one stream will remain with the same name, and the second half of the vnode will get a new stream?
Even if this second part is necessary, if the second half is about a "generation" number, why not use it directly as the second half - not a random number (which, thanks to the birthday paradox, can get a collision more frequently than you might think).

configure.py
haaawk: done.

dht/cdc_partitioner.hh
haaawk: done.

cdc/generation.cc
haaawk: done.

cdc/generation.cc
haaawk: When the topology changes, we have to close all the streams and open a set of new streams that reflect new topology. This is because the client has to read all the records that appeared before the topology change prior to reading records that appear after the change. It has to be a clear cut if we want to maintain the order of the messages.

Do you really think that the chance for the collision is so big?

haaawk:
I rebased on top of master and:
1. Moved partitioner to cdc/ directory
2. Simplified handling of shards that don't own any token in a vnode
3. Removed the duplication of the partitioner name.

haaawk:
I updated the CDC dtests to check that every vnode has a stream now:
https://github.com/scylladb/scylla-dtest/pull/1417

haaawk:
@avikivity @nyh @kbr- please review

haaawk:
QA just got an error that was caused by too big number of cross shard writes https://github.com/scylladb/scylla/issues/6127
This PR should fix this.

cdc/generation.cc
```diff
@@ -173,118 +174,17 @@ topology_description generate_topology_description(
...
-                } else {
-                    ss[s_idx] = some_stream;
-                }
+        dht::sharder sharder(entry.streams.size(), entry.sharding_ignore_msb);
+        for (size_t shard_idx = 0; shard_idx < entry.streams.size(); ++shard_idx) {
+            auto t = sharder.token_for_next_shard(prev_token, shard_idx);
+            if (t > entry.token_range_end) {
+                // This means a shard does not own any tokens in this vnode
+                // and it's ok to pick any token so we pick end of vnode.
+                t = entry.token_range_end;
```
kbr-, line 183: Hmm.
Remember that this only ensures shard-colocation on one of the replicas. It does so for all of them only if all replicas have the same number of shards.

For the case where replicas have different numbers of shards, we can be defensive and take some random token so other replicas don't get too imbalanced.

cdc/cdc_partitioner.cc
```diff
@@ -0,0 +1,66 @@
...
+
+#include "cdc_partitioner.hh"
+#include "dht/token.hh"
+#include "schema.hh"
+#include "sstables/key.hh"
+#include "utils/class_registrator.hh"
+#include "cdc/generation.hh"
+#include "keys.hh"
+
+static const sstring cdc_partitioner_name = "com.scylladb.dht.CDCPartitioner";
```
kbr-, line 30: what do we need those `com.scylladb.dht` prefixes for?

cdc/generation.cc
kbr-: If we create 1 billion CDC generations such that there will be some token that ends a vnode in all of those generations, the probability of collision for this token will be 0.03.

cdc/generation.cc
```diff
@@ -173,118 +174,17 @@ topology_description generate_topology_description(
...
-                }
-            }
-            // Replace 'unset' stream ids with indexes above last_set_stream_idx
-            for (int s_idx = ss.size() - 1; s_idx > last_set_stream_idx; --s_idx) {
-                if (ss[s_idx].is_set()) {
-                    some_stream = ss[s_idx];
-                } else {
-                    ss[s_idx] = some_stream;
-                }
+        dht::sharder sharder(entry.streams.size(), entry.sharding_ignore_msb);
```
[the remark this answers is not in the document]
avikivity: Yes. Note you can't return a sharder by value (as we do here) if it has derived classes. We'll have to convert sharder to a value type (with a pimpl implementation) if we ever want to do that.

Agree it's out of scope.

cdc/generation.cc
avikivity: Hmm. This is very sad since I wanted to be able to rely on collocation.

I don't see a way out though.

I don't think picking a random token will be any better than the end token though (and also think that we can allow less than perfect behavior when the shard counts are different, at least until we have tablets)

cdc/cdc_partitioner.cc
```diff
@@ -0,0 +1,66 @@
...
+
+namespace cdc {
+
+const sstring cdc_partitioner::name() const {
+    return cdc_partitioner_name;
+}
+
+dht::token
+cdc_partitioner::get_token(bytes key) const {
```
avikivity, line 39: This is suboptimal, moving bytes around is expensive since it has to check if it's using internal or external storage each time.

As a follow up, please change stream_id to store two int64_t instead of bytes, and change the stream_id constructor to accept a bytes_view (and check during construction that the size is 16). Then this can also use bytes_view.

cdc/cdc_partitioner.hh
```diff
@@ -0,0 +1,52 @@
...
+
+class cdc_partitioner final : public dht::i_partitioner {
+    dht::token get_token(bytes key) const;
+    dht::token get_token(int64_t value) const;
+public:
+    cdc_partitioner() = default;
+    virtual const sstring name() const override;
+    virtual dht::token get_token(const schema& s, partition_key_view key) const override;
+    virtual dht::token get_token(const sstables::key_view& key) const override;
+    virtual bool preserves_order() const override { return false; }
```
avikivity, line 48: Note: we can kill preserves_order(). It has only one user which can be trivially removed.

cdc/generation.cc
```diff
@@ -154,8 +154,9 @@ topology_description generate_topology_description(
     tokens.erase(std::unique(tokens.begin(), tokens.end()), tokens.end());
 
     std::vector<token_range_description> entries(tokens.size());
-    int spots_to_fill = 0;
 
+    assert(tokens.size() > 1);
```
avikivity, line 158: Why > 1? What if we have a single node with one token?

cdc/generation.cc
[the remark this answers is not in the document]
avikivity: Scratch that, you're not returning a dht::sharder by value.

cdc/generation.cc
```diff
@@ -173,118 +174,17 @@ topology_description generate_topology_description(
...
-            // Replace 'unset' stream ids with indexes above last_set_stream_idx
-            for (int s_idx = ss.size() - 1; s_idx > last_set_stream_idx; --s_idx) {
-                if (ss[s_idx].is_set()) {
-                    some_stream = ss[s_idx];
-                } else {
-                    ss[s_idx] = some_stream;
-                }
+        dht::sharder sharder(entry.streams.size(), entry.sharding_ignore_msb);
+        for (size_t shard_idx = 0; shard_idx < entry.streams.size(); ++shard_idx) {
+            auto t = sharder.token_for_next_shard(prev_token, shard_idx);
```
avikivity, line 179: This breaks for the wrap-around case (when prev_token == tokens.back()). token_for_next_shard does not wrap around, so you'll get maximum_token if prev_token is very close to th end.

cdc/generation.cc
```diff
@@ -173,118 +174,17 @@ topology_description generate_topology_description(
...
+            if (t > entry.token_range_end) {
```
avikivity, line 180: In addition, this test can fail even though we got a good token. Consider min_token = 0 and max_token = 1000. Consider prev_token = 990 and t = 10 (a wraparound range). Suppose token_for_next_shard returns 995. Then this test will compare it against 10 and fail, even though 995 is inside the wraparound range (990, 10].

avikivity: I think you can fix both problems by using ring_position_range_vector_sharder. Use a vector of ranges instead of a range; the wraparound range translates to a two-element vector and the rest translate to a single element vector. You can keep calling ring_position_range_vector_sharder_oh_why_is_this_so_long::next() until you have enough for smp::count or until it return nullopt, indicating no more shards own data.

avikivity:
Reviewed, I think I found some bugs.

cdc/generation.cc
haaawk: Fair point. Will fix.

cdc/cdc_partitioner.hh
```diff
@@ -0,0 +1,52 @@
...
+namespace sstables {
+
+class key_view;
+
+}
+
+namespace cdc {
+
+class cdc_partitioner final : public dht::i_partitioner {
+    dht::token get_token(bytes key) const;
```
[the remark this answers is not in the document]
haaawk: If they are a pure functions then I will make them a static functions.

cdc/cdc_partitioner.cc
haaawk: That's just a convention partitioners use.

cdc/cdc_partitioner.hh
haaawk: I will work on that in a separate PR.

cdc/generation.cc
haaawk: This assumption just simplified the code. I thought that we always use vnodes but if you want to cover the case of a node having just one token then I will fix the code.

avikivity: I don't think you depend on it.

And yes, we always use vnodes, but we don't want to add secret dependencies on vnodes.

haaawk: Sure. Will fix.

cdc/cdc_partitioner.hh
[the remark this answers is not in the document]
haaawk: done.

cdc/cdc_partitioner.cc
haaawk: It would be suboptimal because then we would have to transform those two ints into blob for every CDC write. It's way more common than this streams generation. I will introduce a static function to stream_id than takes bytes_view and returns token and will use it here instead.

haaawk: done

cdc/generation.cc
haaawk: done.

kbr-:
Would it be possible to include this PR in 4.0 release?

People that use CDC and upgrade from 4.0 to 4.1 are going to have a problem, because their CDC log tables will use the old partitioner while the CDC generation algorithm will use the new CDC partitioner, so colocation will be completely broken.

haaawk:
> Would it be possible to include this PR in 4.0 release?
> 
> People that use CDC and upgrade from 4.0 to 4.1 are going to have a problem, because their CDC log tables will use the old partitioner while the CDC generation algorithm will use the new CDC partitioner, so colocation will be completely broken.

As much as I would like to see this PR in 4.0, it shouldn't be critical. We will just ask the users to disabled CDC before the upgrade.

cdc/generation.cc
haaawk: That would require some unnecessary allocations. I solved both issues by just special casing vnode (tokens.back(), tokens.front()].

haaawk:
I addressed all the comments. Thanks @avikivity and @kbr-

cdc/generation.cc
avikivity: This happens once per bootstrap, allocations here are not a problem, but correctness is. I hope this isn't causing more complications.

avikivity:
> Would it be possible to include this PR in 4.0 release?
> 
> People that use CDC and upgrade from 4.0 to 4.1 are going to have a problem, because their CDC log tables will use the old partitioner while the CDC generation algorithm will use the new CDC partitioner, so colocation will be completely broken.

No, 4.0 only receives fixes for regressions now.

cdc/generation.cc
haaawk: It's actually quite simple. I even think that the resulting version is easier to understand.

haaawk:
It turned out in testing that for some reason the code does not generate what I was expecting for smp = 1. I'm investigating so please don't merge before I get back to you.

haaawk:
It seems that token_for_next_shard does not work for smp == 1. It always returns maxumum_token in such case.

haaawk:
I think I understand the problem now. token_for_next_shard has a comment:
```
Gets the first token greater than `t` that is in shard `shard`, and is a shard boundary (its first token).
```

`and is a shard boundary (its first token)` is crucial here. I need something else for this PR. First token for shard but not necessarily the shard boundary.

haaawk:
It seems that ring_position_range_vector_sharder has the same problem so it's not the solution as well.

haaawk:
I fixed the issue and the latest version passes tests with both smp == 1 and smp == 3.
@avikivity @kbr- @nyh please have a look

cdc/cdc_partitioner.hh
haaawk: I added a commit that removes preserves_order to this PR.

haaawk:
@avikivity I added a commit that removes i_partitioner::preserves_order

cdc/generation.cc
```diff
@@ -154,137 +158,69 @@ topology_description generate_topology_description(
...
-                    some_stream = s;
-                    break;
-                }
-            }
-            if (some_stream.is_set()) {
-                break;
-            }
+        t = dht::token::from_int64(std::numeric_limits<int64_t>::min());
+        if (sharder.shard_of(t) != shard_idx) {
+            t = sharder.token_for_next_shard(t, shard_idx);
```
kbr-, line 197: I have absolutely no idea what this token/sharder wizardry does.

If I understand correctly, the problem statement is:
given token `start`, find the smallest token "after" `start` which falls into shard `shard_idx`.

Can you:
1. take this piece of code and name it (inside a pure static function), e.g. with the following signature
```
static dht::token first_token_for_shard(dht::token start, shard_id shard, sharder&)
```
2. write a comment outside the function explaining what the function does and how it's different from `token_for_next_shard`
3. write a comment inside the function explaining how the function does it (why do we need all this magic)?

cdc/generation.cc
```diff
@@ -154,137 +158,69 @@ topology_description generate_topology_description(
...
-                }
+    // Then we cover all vnodes that don't wrap around
+    for (size_t i = 1; i < tokens.size(); ++i) {
+        auto start = tokens[i - 1];
+        auto end = tokens[i];
+        setup_vnode(entries[i], start, end, [&] (const dht::sharder& sharder, size_t shard_idx) {
+            // It is guaranteed that start is not MAX_INT64 because end is still greater
+            auto t = dht::token::from_int64(dht::token::to_int64(start) + 1);
+            if (sharder.shard_of(t) != shard_idx) {
+                t = sharder.token_for_next_shard(start, shard_idx);
```
kbr-, line 215: this should also probably be part of `first_token_for_shard` (or however you name it)

cdc/generation.cc
haaawk: I will extract the logic to free functions and add some comments.

haaawk:
@kbr- I cleaned up the code, split it into more meaningful functions and added more comments. I hope you will find it easier now.

I also enclosed the logic into a class to avoid passing around parameters.

kbr-:
> I also enclosed the logic into a class to avoid passing around parameters.

This is a very opinionated change, I'm not sure this PR is a place for such changes... I like when parameters are being passed explicitly to functions, because it's easy to see what the function uses and what it doesn't (assuming it doesn't use global variables).
But I won't protest...

cdc/generation.cc
```diff
@@ -119,176 +123,213 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
+     * Finds the first token in this vnode which is a wrap around.
+     *
+     * Such vnode has to be considered as two ranges.
+     * Range (|start|, 2^63-1] first and only when no token for shard
+     * |shard_idx| is found there, a second range of [-2^63, |end|] is searched.
+     */
+    dht::token find_first_token_for_shard_in_wrap_around(
+            const dht::sharder& sharder, size_t shard_idx) const;
+    /*
+     * Finds the first token in this vnode which is a wrap around.
```
kbr-, line 180: "which is not a wrap around" or "which is a non-wrap around"

haaawk: thanks.

cdc/generation.cc
```diff
@@ -119,176 +123,213 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
-            auto endpoint = token_metadata.get_endpoint(entry.token_range_end);
-            if (!endpoint) {
-                throw std::runtime_error(format("Can't find endpoint for token {}", entry.token_range_end));
-            }
-            auto sc = get_shard_count(*endpoint, gossiper);
-            entry.streams.resize(sc > 0 ? sc : 1);
-            entry.sharding_ignore_msb = get_sharding_ignore_msb(*endpoint, gossiper);
-        }
+static
+dht::token next_token_wrap_around(dht::token t) {
```
kbr-, line 204: This name is a bit confusing, if we contrast it with `next_token_no_wrap_around`, which assumes that the token is not the maximum token.
Maybe simple `next_token` would suffice?

cdc/generation.cc
```diff
@@ -119,176 +123,213 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
+        const dht::sharder& sharder, size_t shard_idx) const {
+    auto t = next_token_wrap_around(start);
+    if (sharder.shard_of(t) != shard_idx) {
+        t = sharder.token_for_next_shard(start, shard_idx);
     }
+    if (!t.is_maximum()) {
+        // This means we have found a token for shard |shard_idx| before 2^63-1
+        return t;
+    }
+    // No token owned by shard |shard_idx| was found in (|start|, 2^63 - 1]
```
kbr-, line 219: what if `t == 2^63 - 1` and `t` belongs to `shard_idx`?
Then `t.is_maximum()` is true, and we can return `t`.

cdc/generation.cc
```diff
@@ -119,176 +123,213 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
-
-        // For each range that doesn't have any stream,
-        // use streams of the first range to the left which does have a stream.
+dht::token vnode::find_first_token_for_shard_in_non_wrap_around(
+        const dht::sharder& sharder, size_t shard_idx) const {
+    // It is guaranteed that start is not MAX_INT64 because end is greater
+    auto t = next_token_no_wrap_around(start);
+    if (sharder.shard_of(t) != shard_idx) {
+        t = sharder.token_for_next_shard(start, shard_idx);
+    }
```
kbr-, line 234: can't the above 4 lines be simplified to
`t = sharder.token_for_next_shard(start, shard_idx);`
?

cdc/generation.cc
```diff
@@ -119,176 +123,213 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
+    const locator::token_metadata& _token_metadata;
+    const gms::gossiper& _gossiper;
+
+    // Compute a set of tokens that split the token ring into vnodes
+    auto get_tokens() const {
+        auto tokens = _token_metadata.sorted_tokens();
+        auto it = tokens.insert(
+                tokens.end(), _bootstrap_tokens.begin(), _bootstrap_tokens.end());
+        std::sort(it, tokens.end());
+        std::inplace_merge(tokens.begin(), it, tokens.end());
```
kbr-, line 250: You had to do this, didn't you :P

cdc/generation.cc
```diff
@@ -119,176 +123,213 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
-            for (auto s: entries[idx].streams) {
-                if (s.is_set()) {
-                    some_stream = s;
-                    break;
-                }
-            }
-            if (some_stream.is_set()) {
-                break;
+    // Fetch sharding parameters for a node that owns vnode ending with |end|
+    std::pair<size_t, uint8_t> get_sharding_info(dht::token end) const {
```
kbr-, line 256: The comment could explain what the parameters actually are, `pair` doesn't say much. (although the (only) caller makes it obvious)

cdc/generation.cc
haaawk: No. That would not pick start+1 if [start-Y, start + X] for X > 1 and Y > 0 was owned by shard_idx

cdc/generation.cc
haaawk: no - is_maximum return true only for the special kind of token that's after_all_tokens.

cdc/generation.cc
haaawk: Will change.

cdc/generation.cc
haaawk: Yes :)

cdc/generation.cc
kbr-: Would it be possible to completely get rid of those other token kinds that are not represented by integers? What do we need them for?

haaawk:
@kbr- I improved the code a bit more. It's now shorter, more performant and hopefully easier to understand.

cdc/generation.cc
kbr-: ah, because `token_for_next_shard` only picks a boundary.

cdc/generation.cc
```diff
@@ -119,176 +123,197 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
-
-            auto shard_id = dht::shard_of(entry.streams.size(), entry.sharding_ignore_msb, token);
-            assert(shard_id < entry.streams.size());
-
-            if (!entry.streams[shard_id].is_set()) {
-                --spots_to_fill;
-                entry.streams[shard_id] = stream_id;
-            }
+        if (!t.is_maximum()) {
+            // This means we have found a token for shard |shard_idx| before 2^63-1
```
kbr-, line 179: So the comment does not say the whole truth.
it could be that `t == 2^63-1` here.

haaawk: I guess I wasn't precise. I meant <= 2^63-1. Will remove the -1.

cdc/generation.cc
```diff
@@ -123,176 +123,197 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
-            auto endpoint = token_metadata.get_endpoint(entry.token_range_end);
-            if (!endpoint) {
-                throw std::runtime_error(format("Can't find endpoint for token {}", entry.token_range_end));
-            }
-            auto sc = get_shard_count(*endpoint, gossiper);
-            entry.streams.resize(sc > 0 ? sc : 1);
-            entry.sharding_ignore_msb = get_sharding_ignore_msb(*endpoint, gossiper);
+/*
+ * Represents a wrap around vnode..
+ * This means a token range (|start|, |end|].
```
avikivity, line 152: Why absolute value?

cdc/generation.cc
avikivity: This is already implemented by ring_position_range_vector_sharder. The duplication isn't helping anything.

cdc/generation.cc
haaawk: It's not the absolute value just a way of marking those are fields in the class. I will change to this.start and this.end to avoid confusion.

cdc/generation.cc
haaawk: For ranges. I will look at removing them but not in this PR.

cdc/generation.cc
haaawk: added comment.

haaawk:
@avikivity Is the usage of ring_position_range_vector_sharder a requirement without which you won't merge this PR? In my opinion, implementation using ring_position_range_vector_sharder will be not only less performant but also more complicated and definitely harder to understand due to the total lack of any documentation for sharder.

haaawk:
ping @avikivity

avikivity:
> @avikivity Is the usage of ring_position_range_vector_sharder a requirement without which you won't merge this PR? In my opinion, implementation using ring_position_range_vector_sharder will be not only less performant but also more complicated and definitely harder to understand due to the total lack of any documentation for sharder.

It's not a requirement; I just think it will be simpler since your code has to duplicate its work. Since you say you simplified it, I'll review again.

Agree about the lack of documentation, we should improve both on documenting, and requiring documentation when reviewing.

cdc/generation.cc
```diff
@@ -123,176 +123,198 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
-
-    auto tokens = token_metadata.sorted_tokens();
-    tokens.insert(tokens.end(), bootstrap_tokens.begin(), bootstrap_tokens.end());
-    std::sort(tokens.begin(), tokens.end());
-    tokens.erase(std::unique(tokens.begin(), tokens.end()), tokens.end());
+// Assumes t is smaller than 2^63 - 1
+static
+dht::token next_token_no_wrap_around(dht::token t) {
+    return dht::token::from_int64(dht::token::to_int64(t) + 1);
+}
```
avikivity, line 137: These kinds of token calculations should be in token.hh.

cdc/generation.cc
```diff
@@ -123,176 +123,198 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
 
-    std::vector<token_range_description> entries(tokens.size());
-    int spots_to_fill = 0;
+static
+dht::token smallest_token() {
+    return dht::token::from_int64(std::numeric_limits<int64_t>::min());
```
avikivity, line 141: This is actually an illegal token. The first legal token is numeric_limits::min() + 1.

cdc/generation.cc
```diff
@@ -123,176 +123,198 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
-        auto passed = std::chrono::duration_cast<std::chrono::seconds>(now - start_time);
-        if (passed > quota) {
-            return stop_iteration::yes;
+/*
+ * Represents a non wrap around vnode.
+ * This means a token range (this.start, this.end].
+ * For example (10, 100].
+ */
+struct non_wrap_around_vnode final {
+    dht::token start;
```
avikivity, line 198: I have to say I still think this can be simplified by using the partition_range_vector_sharder. You're recreating that machinery, with bugs.

You can write something like this.

```c++
auto wrapping_vnode = wrapping_range(start_token, end_token);
auto unwrapped_vnode = wrapping_vnode.unwrap(); // this is a vector
auto token_sharder = ring_position_range_vector_sharder(.., unwrapped_vnode);
while (auto next = token_sharder.next(); nr_tokens < nr_cpus && next) {
    tokens[next->shard] (next->rangebegin());
    ++nr_tokensl
}
```

Unfortunately, what we have now works with ring_positions, not tokens. So you'll need a schema. Is one handy here?

haaawk: Nope. No schema because the generation is shared across all tables.

cdc/generation.cc
haaawk: You're right. I will fix. BTW using ring_position_range_sharder would have the same problem because the sharder would return range starting with dht::minimum_token().

cdc/generation.cc
haaawk: The code you proposed is incorrect the same way current implementation is. One of the returned ranges will start with dht::minnimum_token which is not valid.

If you insist then I will just use the sharder. I don't care about the matter that much.

haaawk: BTW you're way to optimistic about how simple the code will be:
1. wrapping_range.unwrap() returns pair of wrapping_ranges (not nonwrapping_ranges) and if you call it on a range that's actually not wrapping it will return (-inf, start] + [end, inf) so we can't call the same code for wrapping and nonwrapping vnodes.

avikivity: @haaawk, I don't insist on using the range vector sharder, but I insist on using simple code. The code you presented isn't simple. Either simplify it, or use more infrastructure, or add more infrastructure and use it. This is not the place to play token games.

I agree the partition_range_vector_sharder isn't optimal due to its use of infinities and since it requires a schema. Perhaps a new sharder can be used. This separates the concerns.

haaawk: I dealt with the problem of schema by turning ring_position_range_sharder into token_range_sharder but the code does not look simple at all:

```
        dht::sharder sharder(shard_count, ignore_msb);
        dht::token_range_vector_sharder token_sharder(sharder, std::move(vnode));
        while (auto next = token_sharder.next()) {
            auto stream_token = end;
            auto& start = next->ring_range.start();
            if (!start || start->value().is_minimum()) {
                auto next_token = dht::from_int64(std::numeric_limits<int64_t>::min() + 1);
                if (next->ring_range.contains(next_token, dht::tri_compare)) {
                    desc.streams[next->shard] = next_token;
                }
            } else {
                if (start->is_inclusive()) {
                    desc.streams[next->shard] = start->value();
                } else {
                    int64_t token_value = dht::to_int64(start->value());
                    if (token_value < std::numeric_limits<int64_t>::max())) {
                        auto next_token = dht::from_int64(token_value + 1);
                        if (next->ring_range.contains(next_token, dht::tri_compare)) {
                            desc.streams[next->shard] = next_token;
                        }
                    }
                }
            }
        }
```

haaawk: I think I will just try to simplify my previous attempt.

haaawk:
I extracted infrastructure (dht::first_token_for_shard) and added test for it. The code in cdc/generation is simpler now. Unit tests in dev and release are passing. CDC dtests as well. @avikivity could you please have another look?

haaawk:
@avikivity ping

avikivity:
Looking

dht/token-sharding.hh
```diff
@@ -87,4 +87,20 @@ inline std::ostream& operator<<(std::ostream& os, const sharder& sharder) {
...
+ *
+ * The first token means the one that appears first on the ring when going
+ * from `start` to `end`.
+ * 'first token' is not always the smallest.
+ * For example, if in vnode (100, 10] only tokens 110 and 1 belong to
+ * shard shard_idx then token 110 is the first because it appears first
+ * when going from 100 to 10 on the ring.
+ */
+dht::token find_first_token_for_shard(
+        const dht::sharder& sharder, dht::token start, dht::token end, size_t shard_idx);
```
avikivity, line 104: s/size_t/shard_id/, we have a type for it.

Let's add _wrapping to the name, and _exclusive and _inclusive to start and end. It is well documented, but this is tricky and making the names stand out helps a little.

dht/token.cc
```diff
@@ -286,4 +286,35 @@ dht::token token::from_int64(int64_t i) {
...
 
+static
+dht::token find_first_token_for_shard_in_not_wrap_around_range(const dht::sharder& sharder, dht::token start, dht::token end, size_t shard_idx) {
+    // Invariant start < end
+    // It is guaranteed that start is not MAX_INT64 because end is greater
+    auto t = dht::token::from_int64(dht::token::to_int64(start) + 1);
+    if (sharder.shard_of(t) != shard_idx) {
+        t = sharder.token_for_next_shard(t, shard_idx);
+    }
+    return std::min(t, end);
```
avikivity, line 297: This is returning the wrong result, no? Maybe return std::nullopt instead (and std::optional<token> for the type).

dht/token-sharding.hh
```diff
@@ -87,4 +87,20 @@ inline std::ostream& operator<<(std::ostream& os, const sharder& sharder) {
     return os;
 }
 
+/*
+ * Finds the first token in token range (`start`, `end`] that belongs to shard shard_idx.
+ *
+ * If there is no token that belongs to shard shard_idx in this range,
+ * `end` is returned.
```
avikivity, line 94: How does one distingish between `end` used as a failure marker and `end` being a valid result?

Better return std::optional instead.

dht/token.cc
```diff
@@ -286,4 +286,35 @@ dht::token token::from_int64(int64_t i) {
...
+}
+
+dht::token find_first_token_for_shard(
+        const dht::sharder& sharder, dht::token start, dht::token end, size_t shard_idx) {
+    if (start < end) { // Not a wrap around token range
```
avikivity, line 302: start == end _is_ a wrap around range, so this is correct. But please add a comment noting it.

dht/token.cc
```diff
@@ -286,4 +286,35 @@ dht::token token::from_int64(int64_t i) {
...
+        return find_first_token_for_shard_in_not_wrap_around_range(sharder, start, end, shard_idx);
+    } else { // A wrap around token range
+        dht::token t;
+        if (dht::token::to_int64(start) != std::numeric_limits<int64_t>::max()) {
+            t = find_first_token_for_shard_in_not_wrap_around_range(sharder, start, dht::maximum_token(), shard_idx);
+            if (!t.is_maximum()) {
+                // This means we have found a token for shard shard_idx before 2^63
```
avikivity, line 309: This works, but I think it's better to return not found via std::nullopt. This is tricky code, let's make it less tricky.

cdc/generation.cc
```diff
@@ -123,176 +123,110 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
+    const gms::gossiper& _gossiper;
+
+    // Compute a set of tokens that split the token ring into vnodes
+    auto get_tokens() const {
+        auto tokens = _token_metadata.sorted_tokens();
+        auto it = tokens.insert(
+                tokens.end(), _bootstrap_tokens.begin(), _bootstrap_tokens.end());
+        std::sort(it, tokens.end());
+        std::inplace_merge(tokens.begin(), it, tokens.end());
+        tokens.erase(std::unique(tokens.begin(), tokens.end()), tokens.end());
```
avikivity, line 146: How can there be duplicate tokens?

avikivity: Never mind, this is existing code.

cdc/generation.cc
```diff
@@ -123,176 +123,110 @@ const std::vector<token_range_description>& topology_description::entries() cons
...
-                "cdc: bootstrap tokens is empty in generate_topology_description");
+class topology_description_generator final {
+    const db::config& _cfg;
+    const std::unordered_set<dht::token>& _bootstrap_tokens;
+    const locator::token_metadata& _token_metadata;
+    const gms::gossiper& _gossiper;
+
+    // Compute a set of tokens that split the token ring into vnodes
+    auto get_tokens() const {
+        auto tokens = _token_metadata.sorted_tokens();
```
avikivity, line 141: How does this work with multiple data centers and racks? A vnode with multiple data centers and racks ignores adjacent tokens that are on the same rack or same dc.

I think I found a problem: let's say two adjacent tokens are (300001, 300002]. Then the code to find the first token for a shard will fail for all but one shard (the one that olds 300002). But if the token 300001 is held by a node on the same rack, or a different data center, then the beginning of the vnode is not 300001 but somewhere before, let's say 100001. Now you can find shards for all the vnode range.

avikivity:
Please use --ignore-date on the next rebase because otherwise github displays the commits out of order.

dht/token-sharding.hh
haaawk: done.

dht/token.cc
haaawk: Returning optional instead.

dht/token-sharding.hh
haaawk: done.

dht/token.cc
haaawk: Yes. Done.

dht/token.cc
haaawk: done.

cdc/generation.cc
haaawk: Aren't different datacenters having separate token rings?

haaawk: I don't know how racks affect the ring - will check it out.

haaawk: @kbr- Have you considered multi DC and racks when you were doing the initial code? Could you comment, please?

avikivity: No, there's one big ugly ring and the replication strategy ignores racks in the same dc and nodes in different dc:s.

kbr-: > @kbr- Have you considered multi DC and racks when you were doing the initial code? Could you comment, please?

Tokens and DCs/Racks are orthogonal concepts. We're talking apples and oranges, so there's really nothing to consider here.

> let's say two adjacent tokens are (300001, 300002]. Then the code to find the first token for a shard will fail for all but one shard (the one that olds 300002). But if the token 300001 is held by a node on the same rack, or a different data center, then the beginning of the vnode is not 300001 but somewhere before, let's say 100001.

Each node chooses a set of 255 (by default) tokens. These tokens are **chosen** by this node, and this node is the **owner** of these tokens.

The token ring is the sum of chosen tokens for all nodes. There is 255 * num_nodes chosen tokens in total.

A vnode is a smallest token range `(x, y]` such that `x` and `y` are chosen. "smallest" means that there is no chosen token in between.

If 300001 and 300002 are chosen, then `(300001, 300002]` is a vnode, and `(100001, 300002]` is not a vnode.

The definition of vnode does not include "DC" or "rack" anywhere.

A replication strategy iterates over all vnodes in the token ring in the clockwise direction -- or equivalently, it iterates over all chosen tokens in the clockwise direction -- and looks at the owner of the end token of this vnode (for vnode `(x, y]`, it looks at the owner of `y`). It doesn't look at tokens `z` such that `x < z <= y`, because they give no information: the concept of "owner" is not defined for these tokens. In the case of NTS, it will sometimes decide to not include an owner in the calculated replica set if it already included an owner in the same DC/rack.

For two tokens in the same vnode, the set of calculated replicas will be the same, because the sequence of chosen tokens "to the right" of these tokens is the same. Intuitively, the replication strategy (any replication strategy) cannot distinguish between two tokens in the same vnode.

We choose streams so that every vnode has at least one stream whose token lies in this vnode. When we consider a partition key (when making a write), we take its token and we take a stream whose token lies in the same vnode as the token of the key. Because they lie in the same vnode, the replica set for both of them will be the same.

kbr-: Also I made this document

https://docs.google.com/document/d/1_p5mJUSWb_CoCD9Fg7mum2KH893FuvM2uRquQrXZmGY/edit?usp=sharing

with a detailed explanation of tokens, token ring, replication strategies, and how these concepts play with each other.

(it's not published yet because it's part of a bigger document that I'm working on from time to time which is not finished yet)

haaawk: @avikivity we can't do the optimization you mentioned. We need stream for exactly each vnode because writes will be going to all replicas (in all datacenters).

haaawk:
@avikivity What's the status of this PR? You had one comment but I believe it is not correct. Do you agree and can we merge this or do you want more explanation?

cdc/generation.cc
avikivity: I think @kbr- is correct. But let's make sure. Where is the code that uses the narrow definition of vnode (two adjacent tokens) rather than the wide definition (two tokens that define a range according to the replication strategy)?

avikivity:
> @avikivity What's the status of this PR? You had one comment but I believe it is not correct. Do you agree and can we merge this or do you want more explanation?

It's not an optimization, but a correctness issue. I believe the code is correct, but I asked for more info to prove this.

haaawk:
> > @avikivity What's the status of this PR? You had one comment but I believe it is not correct. Do you agree and can we merge this or do you want more explanation?
> 
> It's not an optimization, but a correctness issue. I believe the code is correct, but I asked for more info to prove this.

Is @kbr- wall of text enough?

avikivity:
> > > @avikivity What's the status of this PR? You had one comment but I believe it is not correct. Do you agree and can we merge this or do you want more explanation?
> > 
> > 
> > It's not an optimization, but a correctness issue. I believe the code is correct, but I asked for more info to prove this.
> 
> Is @kbr- wall of text enough?

I asked for a pointer to code (see conversation). I agree with his reasoning but would like to see that the code agrees too.

cdc/generation.cc
kbr-: For NTS, this is the loop that walks over the vnodes (their end tokens) and decides whether or not to take an owner of this token into the set of replicas:
```
    for (auto& next : tm.ring_range(search_token)) {
        if (dcs_to_fill == 0) {
            break;
        }

        inet_address ep = *tm.get_endpoint(next);
        auto& loc = tp.get_location(ep);
        auto i = dcs.find(loc.dc);
        if (i != dcs.end() && i->second.add_endpoint_and_check_if_done(ep, loc)) {
            --dcs_to_fill;
        }
    }
```
(in `network_topology_strategy::calculate_natural_replicas`).

`tm` is token_metadata. `tm.ring_range(search_token)` takes the first chosen token "equal or to the right, clockwise" of `search_token`. So `next` walks over chosen tokens in clockwise direction, and decides whether or not we should include its owner to the replica set.

`ring_range` is implemented using `token_iterator_impl` in token_metadata, which uses `tm.sorted_tokens()`. It doesn't look at DC/rack placement.

The important fact for us is that for two values of `search_token` for which `tm.ring_range(search_token)` will be the same, the set of calculated replicas will be the same.

In other words, for two tokens that are in the same vnode, the set of calculated replicas will be the same.

For SimpleStrategy we have a similar, but simpler loop:
```
    for (auto& token : tm.ring_range(t)) {
        if (endpoints.size() == replicas) {
           break;
        }
        auto ep = tm.get_endpoint(token);
        assert(ep);

        endpoints.push_back(*ep);
    }
```
(in `simple_strategy::calculate_natural_endpoints`).

avikivity: To complete my understanding, where does cdc make the connection between the mutation's token and the chosen stream?

kbr-: cdc/log.cc:
```
    std::tuple<mutation, stats::part_type_set> transform(const mutation& m, const cql3::untyped_result_set* rs, api::timestamp_type ts, bytes tuuid, int& batch_no) {
        auto stream_id = _ctx._cdc_metadata.get_stream(ts, m.token());
```

Where `get_stream` is a method of `cdc_metadata`, which is defined in cdc/metadata.hh.

And this is the implementation of `get_stream`:
```
static cdc::stream_id get_stream(
        const cdc::token_range_description& entry,
        dht::token tok) {
    // The ith stream is the stream for the ith shard.
    auto shard_cnt = entry.streams.size();
    auto shard_id = dht::shard_of(shard_cnt, entry.sharding_ignore_msb, tok);

    if (shard_id >= shard_cnt) {
        on_internal_error(cdc_log, "get_stream: shard_id out of bounds");
    }

    return entry.streams[shard_id];
}

static cdc::stream_id get_stream(
        const std::vector<cdc::token_range_description>& entries,
        dht::token tok) {
    if (entries.empty()) {
        on_internal_error(cdc_log, "get_stream: entries empty");
    }

    auto it = std::lower_bound(entries.begin(), entries.end(), tok,
            [] (const cdc::token_range_description& e, dht::token t) { return e.token_range_end < t; });
    if (it == entries.end()) {
        it = entries.begin();
    }

    return get_stream(*it, tok);
}
```

avikivity: Thanks for the detailed explanations.

avikivity:
Queued.

PR #6137, opened 2020-04-06
install.sh: suppoprt --upgrade

### LINKED ISSUE #5874: install.sh does not provide a way to upgrade scylla
Related with #5830, current install.sh only provides a way to install scylla, but not upgrade scylla.
To support upgrade, we need to prevent overwrite config files and existing data.

syuu1228:
To use install.sh as Scylla install script w/o using .rpm/.deb package,
we need to provide a way to upgrade Scylla version, not just install.

With --upgrade option, install.sh does not overwrite config files.
It will install <filename>.new file on same directory, when old config file and
new config file does not contain same data.
If old one and new one is exactly same, it will nothing.

To implement this, rewriting api_ui_dir/api_doc_dir path on scylla.yaml
moved from .rpm/.deb scriptlet to install.sh.

Fixes #5874

### DIFF
--- a/dist/debian/debian/scylla-server.postinst
+++ b/dist/debian/debian/scylla-server.postinst
@@ -25,11 +25,6 @@ fi
 
 ln -sfT /etc/scylla /var/lib/scylla/conf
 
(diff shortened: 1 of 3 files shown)


avikivity:
Queued as 3ce6cdc6d

PR #6144, opened 2020-04-06
[DRAFT v4] cdc+counters: make counters work in CDC

jul-stas:
[draft, because I left tests at home, unpushed]

This patch augments counter "deltas" on leader node. Before that,
"deltas" are "sanitized", which means that they are set to 0 if
they try to update a deleted counter, so CDC will not contain
spurious (ineffective) updates.

On the CDC side, counter columns are represented as `long_type`,
with unchanged logic behind pre-/postimage (preimage query returns
ordinary integer, postimage is calculated from delta and preimg).

### DIFF
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -1818,11 +1818,32 @@ storage_proxy::mutate_counter_on_leader_and_replicate(const schema_ptr& s, froze
     auto shard = _db.local().shard_of(fm);
     bool local = shard == this_shard_id();
     get_stats().replica_cross_shard_ops += !local;
-    return _db.invoke_on(shard, {_write_smp_service_group, timeout}, [gs = global_schema_ptr(s), fm = std::move(fm), cl, timeout, gt = tracing::global_trace_state_ptr(std::move(trace_state)), permit = std::move(permit), local] (database& db) {
+    const bool get_sanitized_deltas = s->cdc_options().enabled();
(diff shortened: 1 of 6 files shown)


service/storage_proxy.cc
```diff
@@ -2223,11 +2244,16 @@ storage_proxy::get_paxos_participants(const sstring& ks_name, const dht::token &
  */
 future<> storage_proxy::mutate(std::vector<mutation> mutations, db::consistency_level cl, clock_type::time_point timeout, tracing::trace_state_ptr tr_state, service_permit permit, bool raw_counters) {
     if (_cdc && _cdc->needs_cdc_augmentation(mutations)) {
-        return _cdc->augment_mutation_call(timeout, std::move(mutations), tr_state).then([this, cl, timeout, tr_state, permit = std::move(permit), raw_counters](std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>&& t) mutable {
-            auto mutations = std::move(std::get<0>(t));
-            auto tracker = std::move(std::get<1>(t));
-            return _mutate_stage(this, std::move(mutations), cl, timeout, std::move(tr_state), std::move(permit), raw_counters, std::move(tracker));
-        });
+        const bool is_one_counter_mutation = mutations.size() == 1 && mutations.front().schema()->is_counter();
```
haaawk, line 2247: This should be part of needs_cdc_augmentation, I think.

haaawk, line 2247: I would rename to just is_counter_mutation and add a comment that at the moment we don't accept counters in batches.

PR #6152, opened 2020-04-07
Support SSL Certificate Hot Reloading

### LINKED ISSUE #6067: Support SSL Certificate Hot Reloading
Scylla version (or git commit hash): 3.4

Support hot reloading of SSL Certificates, perdiacly and using `nodetool reloadssl` command
See http://cassandra.apache.org/doc/latest/operating/security.html#ssl-certificate-hot-reloading

elcallio:
Fixes #6067

Requires seastar patch set "Reloadable TLS certificates"

Makes the scylla endpoint initializations that support TLS use reloadable certificate stores, watching used cert + key files for changes, and reload iff modified. 

Tests in separate dtest set.

### DIFF
--- a/alternator/server.cc
+++ b/alternator/server.cc
@@ -414,7 +414,13 @@ future<> server::init(net::inet_address addr, std::optional<uint16_t> port, std:
             if (https_port) {
                 set_routes(_https_server._routes);
                 _https_server.set_content_length_limit(server::content_length_limit);
-                _https_server.set_tls_credentials(creds->build_server_credentials());
+                _https_server.set_tls_credentials(creds->build_reloadable_server_credentials([](const std::unordered_set<sstring>& files, std::exception_ptr ep) {
+                    if (ep) {
+                        slogger.warn("Exception loading {}: {}", files, ep);
+                    } else {
+                        slogger.info("Reloaded {}", files);
+                    }
+                }).get0());
                 _https_server.listen(socket_address{addr, *https_port}).get();
                 _enabled_servers.push_back(std::ref(_https_server));
                 slogger.info("Alternator HTTPS server listening on {} port {}", addr, *https_port);
--- a/message/messaging_service.cc
+++ b/message/messaging_service.cc
@@ -277,9 +277,9 @@ future<> messaging_service::unregister_handler(messaging_verb verb) {
     return _rpc->unregister_handler(verb);
 }
 
-messaging_service::messaging_service(gms::inet_address ip, uint16_t port, bool listen_now)
+messaging_service::messaging_service(gms::inet_address ip, uint16_t port)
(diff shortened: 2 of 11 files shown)


message/messaging_service.hh
```diff
@@ -254,6 +255,7 @@ public:
             memory_config mcfg, scheduling_config scfg, bool sltba = false, bool listen_now = true);
     ~messaging_service();
 public:
+    future<> start_listen_ex();
```
avikivity, line 258: Can we avoid having two entry points? e.g. rename start_listen() to something private and internal?

message/messaging_service.cc
```diff
@@ -379,7 +399,9 @@ messaging_service::messaging_service(gms::inet_address ip
         ci.attach_auxiliary("max_result_size", max_result_size.value_or(query::result_memory_limiter::maximum_result_size));
         return rpc::no_wait;
     });
-
+    if (listen_now && credentials) {
+        throw std::invalid_argument("Cannot start from contructor when TLS is enabled");
+    }
     if (listen_now) {
```
avikivity, line 405: Let's drop listen_now() to avoid having two code paths.

alternator/server.cc
```diff
@@ -414,7 +414,13 @@ future<> server::init(net::inet_address addr, std::optional<uint16_t> port, std:
             if (https_port) {
                 set_routes(_https_server._routes);
                 _https_server.set_content_length_limit(server::content_length_limit);
-                _https_server.set_tls_credentials(creds->build_server_credentials());
+                _https_server.set_tls_credentials(creds->build_reloadable_server_credentials([](const std::unordered_set<sstring>& files, std::exception_ptr ep) {
+                    if (ep) {
+                        slogger.warn("Exception loading {}: {}", files, ep);
+                    } else {
+                        slogger.info("Reloaded {}", files);
```
avikivity, line 421: You can use info0() instead to avoid having 98 repetitions of the message.

avikivity:
Please rebase on top of master so we pick up the seastar changes.

alternator/server.cc
elcallio: Yes, but that would generate potential confusion with the warning on exception. What can happen is a user only replaces the PEM and not the KEY file of a cert pair, than you will get an exception "mismatched certs". Then he completes his file mods and replaces KEY as well, so reload succeeds, and we print the joyful message. 
But if this was in a race condition where some shard is lagging, we would get could get a shard0 "success", but a shardX fail (followed by a silence success). Me thinks confusing to log reader...

elcallio:
Updated: 

- Rebased on newest master
- Made "messaging_service::start_listen" non-optional (remove "listen_now" option)

Did not change the logging, because of reason above. Could optionally create shared wrapper type that keeps state of whether it was called with errors and outputs info even on non-zero shard to make logging consistent for user. Maybe follow-up?

avikivity:
Queued.

PR #6153, opened 2020-04-07
dist/docker: make docker-entrypoint.py pass signals to supervisord

### LINKED ISSUE #6150: docker-entrypoint.py doesn't gracefully shutdown scylla when the container is stopped
Looks like the current **docker-entrypoint.py** is not able to handle graceful shutdown of the DB.
when I stop the container I don't see any logs indicating the DB is shutting down, and eventually it gets killed after the termination grace period.

to correctly propagate the SIGTERM signal shouldn't supervisord be running as PID 1?
As this is currently not the case.

*Installation details*
Scylla version: 3.0.8
Cluster size: 5
OS: Ubuntu

fruch:
Stopping docker currently didn't pass the signals to supervisord,
hence scylla wasn't gracefully shutdown.

Fixes #6150

### DIFF
--- a/dist/docker/redhat/docker-entrypoint.py
+++ b/dist/docker/redhat/docker-entrypoint.py
@@ -3,12 +3,21 @@
 
 import os
 import sys
+import signal
+import subprocess
 import scyllasetup
 import logging
 import commandlineparser
 
 logging.basicConfig(stream=sys.stdout, level=logging.DEBUG, format="%(message)s")
 
+supervisord = None
+
+def signal_handler(signum, frame):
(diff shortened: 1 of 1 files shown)


penberg:
Queued as https://github.com/scylladb/scylla/commit/8ea4caec1f46e3b3e51859053faca8a1265741c4, thanks @fruch!

fruch:
> Queued as [8ea4cae](https://github.com/scylladb/scylla/commit/8ea4caec1f46e3b3e51859053faca8a1265741c4), thanks @fruch!

@penberg your'e welcome

PR #6163, opened 2020-04-08
When reading legacy SSTables, freeze types nested in collection types

### LINKED ISSUE #6130: Upgrade from 3.1.1 to 3.2.4 corrupt an SStable with frozen UDT

*Installation details*
Scylla version (or git commit hash): 3.2.4
Cluster size: 1
OS (RHEL/CentOS/Ubuntu/AWS AMI): CentOS

I upgrade ScyllaDB from 3.1.1 to 3.2.4. After restarting into 3.2.4, I found an error message indicating that the table "config" cannot be read from the SStable file, with the problem related to the serialization of a column with UDT:

Apr 04 05:35:32 uat-data1 scylla[2174]:  [shard 0] compaction_manager - compaction failed: sstables::malformed_sstable_exception (fee definition in serialization header does not match schema. Expected org.apache.cassandra.db.marshal.MapType(org.apache.cassandra.db.marshal.UTF8Type,org.apache.cassandra.db.marshal.FrozenType(org.apache.cassandra.db.marshal.UserType(pacificdb,6665655f3032,6d6178:org.apache.cassandra.db.marshal.DoubleType,6d696e:org.apache.cassandra.db.marshal.DoubleType,74797065:org.apache.cassandra.db.marshal.UTF8Type,666565:org.apache.cassandra.db.marshal.FrozenType(org.apache.cassandra.db.marshal.ListType(org.apache.cassandra.db.marshal.FrozenType(org.apache.cassandra.db.marshal.UserType(pacificdb,6665655f30315f76616c7565,616d6f756e74:org.apache.cassandra.db.marshal.DoubleType,616d6f756e745f66726f6d:org.apache.cassandra.db.marshal.DoubleType,616d6f756e745f746f:org.apache.cassandra.db.marshal.DoubleType,666565:org.apache.cassandra.db.marshal.DoubleType,636f6465:org.apache.cassandra.db.marshal.UTF8Type))))))) but got org.apache.cassandra.db.marshal.MapType(org.apache.cassandra.db.marshal.UTF8Type,org.apache.cassandra.db.marshal.UserType(pacificdb,6665655f3032,6d6178:org.apache.cassandra.db.marshal.DoubleType,6d696e:org.apache.cassandra.db.marshal.DoubleType,74797065:org.apache.cassandra.db.marshal.UTF8Type,666565:org.apache.cassandra.db.marshal.FrozenType(org.apache.cassandra.db.marshal.ListType(org.apache.cassandra.db.marshal.UserType(pacificdb,6665655f30315f76616c7565,616d6f756e74:org.apache.cassandra.db.marshal.DoubleType,616d6f756e745f66726f6d:org.apache.cassandra.db.marshal.DoubleType,616d6f756e745f746f:org.apache.cassandra.db.marshal.DoubleType,666565:org.apache.cassandra.db.marshal.DoubleType,636f6465:org.apache.cassandra.db.marshal.UTF8Type))))) in sstable /var/lib/scylla/data/pacificdb/config-16efe27091fe11e7ab87000000000000/mc-19-big-Data.db): retrying
Apr 04 05:35:32 uat-data1 scylla[2174]:  [shard 0] compaction_manager - compaction task handler sleeping for 300 seconds

The "config" table has the following structure:

CREATE TABLE pacificdb.config (
    name text PRIMARY KEY,
    apps text,
    fee map<text, frozen<fee_02>>,
    language text,
    mode map<text, text>,
    screen text,
    software map<text, text>,
    splash text,
    stuff map<text, text>,
    test text,
    type text
);

And the fee_02 is defined as:

CREATE TYPE pacificdb.fee_02 (
    max double,
    min double,
    type text,
    fee frozen<list<frozen<fee_01_value>>>
);

CREATE TYPE pacificdb.fee_01_value (
    amount double,
    amount_from double,
    amount_to double,
    fee double,
    code text
);

The result is I can no longer access this table.

It seems Scylla 3.2 cannot correctly interpret SStable from 3.1 with implied frozen type, or maybe it should migrate this SStable when upgrade.

Also is there a way I can migrate this SStable to the correct format for Scylla 3.2?

kbr-:
Some legacy "mc" SSTables (created in Scylla 3.0) may contain incorrect
serialization headers, which don't wrap frozen UDTs nested inside collections
with the FrozenType<...> tag. When reading such SSTable,
Scylla would detect a mismatch between the schema saved in schema
tables (which correctly wraps UDTs in the FrozenType<...> tag) and the schema
from the serialization header (which doesn't have these tags).

SSTables created in Scylla versions 3.1 and above, in particular in
Scylla versions that contain this commit, create correct serialization
headers (which wrap UDTs in the FrozenType<...> tag).

This commit does two things:
1. for all SSTables created after this commit, include a new feature
   flag, CorrectUDTsInCollections, presence of which implies that frozen
   UDTs inside collections have the FrozenType<...> tag.
2. when reading a Scylla SSTable without the feature flag, we assume that UDTs
   nested inside collections are always frozen, even if they don't have
   the tag. This assumption is safe to be made, because at the time of
   this commit, Scylla does not allow non-frozen (multi-cell) types inside
   collections or UDTs, and because of point 1 above.

There is one edge case not covered: if we don't know whether the SSTable
comes from Scylla or from C*. In that case we won't make the assumption
described in 2. Therefore, if we get a mismatch between schema and
serialization headers of a table which we couldn't confirm to come from
Scylla, we will still reject the table. If any user encounters such an
issue (unlikely), we will have to use another solution, e.g. using a
separate tool to rewrite the SSTable.

Fixes #6130.

tests: unit(dev)


sstables/mp_row_consumer.cc
```diff
@@ -80,9 +81,28 @@ atomic_cell make_counter_cell(api::timestamp_type timestamp, bytes_view value) {
     return ccb.build(timestamp);
 }
 
+// See #6130.
+data_type freeze_types_in_collections(data_type t) {
```
kbr-, line 85: I need to recurse here...

kbr-:
v2:
- make `freeze_types_in_collections` static.
- recursion inside `freeze_types_in_collections`.

kbr-:
Forgot to add the test sstable files...
v3: added the test sstable files

sstables/mp_row_consumer.cc
```diff
@@ -80,9 +81,35 @@ atomic_cell make_counter_cell(api::timestamp_type timestamp, bytes_view value) {
     return ccb.build(timestamp);
 }
 
+// See #6130.
+static data_type freeze_types_in_collections(data_type t) {
+    return ::visit(*t, make_visitor(
+    [] (const map_type_impl& typ) -> data_type {
```
haaawk, line 87: typ? type would be more English less Polish I guess :)

haaawk: Same in all the other lambdas here.

sstables/mp_row_consumer.cc
```diff
@@ -80,9 +81,35 @@ atomic_cell make_counter_cell(api::timestamp_type timestamp, bytes_view value) {
...
+                freeze_types_in_collections(typ.get_elements_type()->freeze()),
+                typ.is_multi_cell());
+    },
+    [] (const list_type_impl& typ) -> data_type {
+        return list_type_impl::get_instance(
+                freeze_types_in_collections(typ.get_elements_type()->freeze()),
+                typ.is_multi_cell());
+    },
+    [&] (const abstract_type& typ) -> data_type {
+        return std::move(t);
```
haaawk, line 104: Doesn't the lambda have to be mutable for this std::move to not become a copy?

test/boost/sstable_3_x_test.cc
```diff
@@ -5259,3 +5259,48 @@ SEASTAR_THREAD_TEST_CASE(test_sstable_log_too_many_rows) {
     test_sstable_log_too_many_rows_f(random, (random + 1), false);
     test_sstable_log_too_many_rows_f((random + 1), random, true);
 }
+
+// The following test runs on test/resource/sstables/3.x/uncompressed/legacy_udt_in_collection
+// It was created using Scylla 3.0.x using the following CQL statements:
+//
+// CREATE KEYSPACE ks WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1};
+// CREATE TYPE ks.ut (a int, b int);
+// CREATE TABLE ks.t (pk int PRIMARY KEY, m map<int, frozen<ut>>) WITH compression = {};
```
haaawk, line 5268: Could you please also add a test that checks frozen<map<int, frozen<ut>>>?

haaawk: It would be better if the test has covered all the collection types.

haaawk: Also nested collection please.

sstables/mp_row_consumer.cc
```diff
@@ -80,9 +81,35 @@ atomic_cell make_counter_cell(api::timestamp_type timestamp, bytes_view value) {
...
 }
 
+// See #6130.
+static data_type freeze_types_in_collections(data_type t) {
+    return ::visit(*t, make_visitor(
+    [] (const map_type_impl& typ) -> data_type {
+        return map_type_impl::get_instance(
+                freeze_types_in_collections(typ.get_keys_type()->freeze()),
+                freeze_types_in_collections(typ.get_values_type()->freeze()),
+                typ.is_multi_cell());
```
haaawk, line 91: Does the bug affect only non-frozen collections or frozen as well? If it affects only non-frozen collections then maybe we could break out of the recursion earlier.

haaawk:
I also can't find the place where you're setting CorrectUDTsInCollections for new sstables @kbr-. Is it done automatically?

haaawk:
> I also can't find the place where you're setting CorrectUDTsInCollections for new sstables @kbr-. Is it done automatically?

Nevermind. I see that sstables writer for mc takes sstable_enabled_features::all() so I guess it is automatically included in every new sstable.

sstables/mp_row_consumer.cc
kbr-: Typ is an abbreviation for type ;)

test/boost/sstable_3_x_test.cc
kbr-: Testing other collection types is redundant, this test + a quick glance at `freeze_types_in_collections` should be enough evidence that this thing works IMHO.

But I agree that a nested frozen collection test could be good.

sstables/mp_row_consumer.cc
kbr-: affects everything

test/boost/sstable_3_x_test.cc
haaawk: I meant both nested collection and a top-level frozen collection since you said it's affected as well. I don't think it's redundant to test all the types of collections. If you made a mistake in the set then this test wouldn't catch that. Same situation with someone changing the code later. The person would run a test and thought all is good even when the mistake was introduced into the handling of a list.

kbr-: I mean, there is really no difference in `freeze_types_in_collection` between list/set/map handling.
If someone broke sets but not maps, they would have to modify just the set part of this function. Which means they would immediately notice that something is wrong. (or they would have to be really careless)

we can make a short proof:
1. all other code is independent of the particular collection affected
2. test checks that maps are handled correctly
3. `freeze_types_in_collection` handles sets and lists the same way it handles maps

=> sets and lists are handled correctly too

(and I'm saying that without actually checking sets and lists manually. But I'll do this and I'll be really ashamed if they won't work)

kbr-: But I can add the test, no problem ;)

haaawk: The tests are also there to help when a programmer fails to notice - even the obvious stuff ;).

haaawk: Thanks!

sstables/mp_row_consumer.cc
kbr-: The lambda's constness applies to the reference, not the thing being referred to.
Remember that `x& const` is the same as `x&`, and `const x& const` is the same as `const x&`. References are always implicitly const.
Thus lambda's adding of const to the reference to `t` is a no-op.

I checked on a small example to confirm:
```
#include <iostream>

class A {
public:
    A() = default;
    A(const A&) {
        std::cout << "copy cons" << std::endl;
    }
    A(A&&) {
        std::cout << "move cons" << std::endl;
    }
};

int main() {
    A x;
    auto l = [&] () {
        //return x;
        return std::move(x);
    };
    auto y = l();
}
```

it prints "move cons".

kbr-:
v4: extended tests to cover udts inside collections nested inside collections, collections other than map, and udts nested inside collections which are top-level frozen.

sstables/mp_row_consumer.cc
haaawk: Makes sense. Thanks for checking.

slivne:
@kbr- 

I am still missing something  (only based on the patch comments not the code)

What if the schema tables are carried over from Scylla 3.0 and have not been overwritten since then  - would they include the FrozenType<..> wrapper for collections of UDTs. 

If they would not include the FrozenType<...> wrapper for collections of UDTs, would the code still work ...

C* upgrade for schema is the following: https://github.com/thobbs/cassandra/commit/fac91dd06b891fd0387ee2fb92a86d9fa7e127bf#diff-956374139fd7a8bc3587fa95efaf0e6a

What do the schema tables (not the sstables) contain, in my view we need to fix the schema not carry on a schema that has a mistake and keep the workaround forever ?

avikivity:
> * when reading an SSTable without the feature flag, we assume that UDTs
>   nested inside collections are always frozen, even if they don't have
>   the tag. This assumption is safe to be made, because at the time of
>   this commit, Scylla does not allow non-frozen (multi-cell) types inside
>   collections or UDTs, and because of point 1 above.

What about sstables imported from Cassandra? They will not have the flag set.

We can use a more complicated condition

    !have_scylla_metadata || correct_udts_in_collections

to isolate old versions of Scylla.

sstables/column_translation.hh
```diff
@@ -72,47 +72,7 @@ private:
         static std::vector<column_info> build(
                 const schema& s,
                 const utils::chunked_vector<serialization_header::column_desc>& src,
-                bool is_static) {
```
avikivity, line 75: I queued this patch as 5a454663f

sstables/mp_row_consumer.cc
avikivity: typical :)

sstables/mp_row_consumer.cc
```diff
@@ -101,6 +128,10 @@ std::vector<column_translation::column_info> column_translation::state::build(
         for (auto&& desc : src) {
             const bytes& type_name = desc.type_name.value;
             data_type type = db::marshal::type_parser::parse(to_sstring_view(type_name));
+            if (!features.is_enabled(CorrectUDTsInCollections)) {
+                // See #6130.
+                type = freeze_types_in_collections(std::move(type));
+            }
```
avikivity, line 134: So here you should check that scylla metadata is present, otherwise it could be a Cassandra sstable. A lazy way to do it is to check whether any bit is enabled in features, hoping there is no sstable so old that it has no bits set (a safe bet since mc format came after feature bits) and that we don't start retiring old feature bits (less safe).

A less lazy way is to add a separate boolean to sstable_enabled_features indicating that the metadata itself was present.

avikivity:
Looks good overall, but need not to break Cassandra imports.

kbr-:
> I am still missing something (only based on the patch comments not the code)
> 
> What if the schema tables are carried over from Scylla 3.0 and have not been overwritten since then - would they include the FrozenType<..> wrapper for collections of UDTs.
> 
> If they would not include the FrozenType<...> wrapper for collections of UDTs, would the code still work ...
> 
> C* upgrade for schema is the following: [thobbs/cassandra@fac91dd#diff-956374139fd7a8bc3587fa95efaf0e6a](https://github.com/thobbs/cassandra/commit/fac91dd06b891fd0387ee2fb92a86d9fa7e127bf#diff-956374139fd7a8bc3587fa95efaf0e6a)
> 
> What do the schema tables (not the sstables) contain, in my view we need to fix the schema not carry on a schema that has a mistake and keep the workaround forever ?

@slivne 

Looks like a different issue.

It seems that we don't have a problem with the schema tables, only with mc sstables associated to user created tables (they had incorrect serialization headers). I created a table in Scylla 3.0 with frozen UDTs and checked `system_schema.columns`, it showed the frozen<> tags correctly. So it seems like we didn't have the problem that the linked Cassandra patch shows, or we had it but it was fixed somewhere in Scylla 3.0 or earlier.
I'm compiling Scylla 3.0.0 to check if there's gonna be a problem with upgrading the schema tables. There wasn't with latest minor version of Scylla 3.0.

kbr-:
Upgrading from 3.0.0 also works. So there is no issue with schema tables, at least the ones created in Scylla 3.x. Don't know about previous versions of Scylla, though.

kbr-:
v5: freeze types nested in collections only if we know for certain that we're dealing with a Scylla-created SSTable. See [diff](https://github.com/scylladb/scylla/compare/f9942edf1b9f24f25d6e53f1f8b40d28e49ad50d..4b4a88dc7021d629f15027a04042dc0702c46826).

kbr-:
v6: rebased onto next (still conflicts with master though).

sstables/mp_row_consumer.cc
kbr-: Done. I also updated the PR cover letter to explain what happens.

PR #6165, opened 2020-04-08
Match Cassandra's null prohibitions

dekimir:
We currently allow null on the right-hand side of certain relations, while Cassandra prohibits it.  Since our handling of these null values is mostly incorrect, it's better to match Cassandra in prohibiting it.

See the discussion [here](https://github.com/scylladb/scylla/pull/5763#discussion_r405557323).

NB: any reverse mismatch (Scylla prohibiting something that Cassandra allows) is left remaining.  For example, we forbid null bounds on clustering columns, which Cassandra allows.

Tests: unit (dev)

### DIFF
--- a/cql3/restrictions/statement_restrictions.cc
+++ b/cql3/restrictions/statement_restrictions.cc
@@ -682,7 +682,7 @@ bool single_column_restriction::IN::is_satisfied_by(bytes_view data, const query
     });
 }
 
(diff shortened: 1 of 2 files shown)


avikivity:
Queued as a4a5b77bd

bhalevy:
Corresponding dtest fix is here: https://github.com/scylladb/scylla-dtest/pull/1430

PR #6167, opened 2020-04-09
cdc+counters: make counters work in CDC

jul-stas:
Introducing CDC log columns for counters as plain bigints (64-bit,
`long_type`). They cannot be of `counter_type` because counter
columns are mutually exclusive with other types (e.g. `bool`
in cdc$deleted_*).

Correct generation of CDC log requires knowledge about tombstones
because updating dead counter cell has no effect (and thus should
not trigger CDC). This is why counter shard mutations are
transferred from DB code to CDC code.

Preimage values for counters are calulated from counter-shards.
Postimage is queried from the DB, because when CDC is being
calculated, counter updates are already applied. What remains
unresolved is printing the full preimage (all columns) upon row
delete, beacuse CDC code for counters sees the DB in the state
of postimage.

Also, this assumes that counter mutations always have only 1 row.
It was intended to simplify the rewrite of `transformer::transform`.
The logic of splitting mutations and tracing is re-used.

EDIT: because this is a heavy change, I'm taking time to write more "basic" tests. They are about to appear here.
Anyway, I expect to see a lot of remarks here and rebasing 10x is... no pleasure.

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -239,6 +239,7 @@ public:
     future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>> augment_mutation_call(
         lowres_clock::time_point timeout,
         std::vector<mutation>&& mutations,
+        mutation_opt&& counter_preimage,
         tracing::trace_state_ptr tr_state
     );
 
@@ -404,6 +405,10 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID> 
                     // lists are represented as map<timeuuid, value_type>. Otherwise we cannot express delta
                     [] (const list_type_impl& type) -> data_type {
                         return map_type_impl::get_instance(type.name_comparator(), type.value_comparator(), false);
+                    },
+                    // counters and their deltas are represented as bigints
+                    [] (const counter_type_impl& type) -> data_type {
+                        return long_type;
                     },
                     // everything else is just frozen self
                     [] (const abstract_type& type) {
@@ -828,7 +833,7 @@ public:
     }
 
     // TODO: is pre-image data based on query enough. We only have actual column data. Do we need
-    // more details like tombstones/ttl? Probably not but keep in mind.
+    // more details like tombstones/ttl? Probably not (except counters) but keep in mind.
     std::tuple<mutation, stats::part_type_set> transform(const mutation& m, const cql3::untyped_result_set* rs, api::timestamp_type ts, bytes tuuid, int& batch_no) {
         auto stream_id = _ctx._cdc_metadata.get_stream(ts, m.token());
         mutation res(_log_schema, stream_id.to_partition_key(*_log_schema));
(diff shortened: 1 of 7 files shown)


kbr-:
> Postimage is queried from the DB, because when CDC is being
calculated, counter updates are already applied. 

I don't get this part. Could you elaborate?

I don't understand: if we have the pre-image and the delta, we should be able to calculate the post-image.

haaawk:
> > Postimage is queried from the DB, because when CDC is being
> > calculated, counter updates are already applied.
> 
> I don't get this part. Could you elaborate?
> 
> I don't understand: if we have the pre-image and the delta, we should be able to calculate the post-image.

This is not a full pre-image so we can't construct postimage that contains all columns.

kbr-:
> This is not a full pre-image

Again, please elaborate.
We don't know the entire counter's value? We don't know what the value is for each shard?
What exactly does the pre-image contain?

haaawk:
> > This is not a full pre-image
> 
> Again, please elaborate.
> We don't know the entire counter's value? We don't know what the value is for each shard?
> What exactly does the pre-image contain?

This is not counter specific. Row can contain multiple columns. Not every operation touches each column. Thus, preimage obtained for the operation does not contain value for each column in the row and we can't compose a full postimage from it. 

Counter value or shard has nothing to do with it.

kbr-:
@haaawk if postimage is enabled, we query all columns in `pre-image_select`.

haaawk:
> @haaawk if postimage is enabled, we query all columns in `pre_image_select`.
> That's how we calculate postimage in the general case.

No we don't :p.
For counters preimage is obtained in a different way.

cdc/log.cc
```diff
@@ -1315,7 +1552,7 @@ transform_mutations(std::vector<mutation>& muts, decltype(muts.size()) batch_siz
 } // namespace cdc
 
 future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>>
-cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, std::vector<mutation>&& mutations, tracing::trace_state_ptr tr_state) {
+cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, std::vector<mutation>&& mutations, mutation_opt&& counter_preimage, tracing::trace_state_ptr tr_state) {
```
kbr-, line 1555: Please write a comment above this function which explains why preimages for counters are obtained from the caller, while preimages for different types are obtained inside the function.

This seems super weird.

Maybe we could unify this and always pass preimage from the caller?

cdc/log.hh
```diff
@@ -93,7 +93,13 @@ public:
         std::vector<mutation>&& mutations,
         tracing::trace_state_ptr tr_state
         );
+
+    future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>> calc_counter_mutation_augmentation(
```
kbr-, line 97: comments?

cdc/log.cc
haaawk: There should be two different augment methods one for counters and one for non-counters.

kbr-: There are two different functions on the top level.
I wouldn't go as far as having two different methods inside `impl`, because it would introduce a shitton of code repetition.

We could, however, make the `is_counter` check compile-time using some template magic.

haaawk: I think it should be separated all the way. If there are common parts they should be free functions used by both.

haaawk: Counters are never mixed with any other type in a table so there's no reason to mangle the logic of unrelated stuff.

database.cc
```diff
@@ -1282,8 +1282,35 @@ std::ostream& operator<<(std::ostream& out, const database& db) {
...
 }
 
-future<mutation> database::do_apply_counter_update(column_family& cf, const frozen_mutation& fm, schema_ptr m_schema,
-                                                   db::timeout_clock::time_point timeout,tracing::trace_state_ptr trace_state) {
+static bool has_deleted_counter_cell(const mutation& state) {
+    bool result = false;
+    for (auto&& cr : state.partition().clustered_rows()) {
+        cr.row().cells().for_each_cell_until([&] (auto id, const atomic_cell_or_collection& ac_o_c) {
+            const auto& cdef = state.schema()->column_at(column_kind::regular_column, id);
+            if (ac_o_c.as_atomic_cell(cdef).is_dead(gc_clock::now())) {
```
kbr-, line 1290: Using the local clock here introduces nondeterminism... Depending on how long it takes for the process to reach this place, the cell will be considered deleted or not.

You said that "updating a dead counter cell has no effect". How does the updating code check whether the cell is dead or not?

database.cc
```diff
@@ -1282,8 +1282,35 @@ std::ostream& operator<<(std::ostream& out, const database& db) {
...
+    }
+    state.partition().static_row().for_each_cell_until([&] (auto id, const atomic_cell_or_collection& ac_o_c) {
+        const auto& cdef = state.schema()->column_at(column_kind::static_column, id);
+        if (ac_o_c.as_atomic_cell(cdef).is_dead(gc_clock::now())) {
+            result = true;
+            return stop_iteration::yes;
+        }
+        return stop_iteration::no;
+    });
+    return result;
```
kbr-, line 1308: This code repetition should be very easy to deal with.

database.cc
```diff
@@ -1325,15 +1353,22 @@ future<mutation> database::do_apply_counter_update(column_family& cf, const froz
...
                 // ...now, that we got existing state of all affected counter
                 // cells we can look for our shard in each of them, increment
                 // its clock and apply the delta.
                 transform_counter_updates_to_shards(m, mopt ? &*mopt : nullptr, cf.failed_counter_applies_to_memtable());
                 tracing::trace(trace_state, "Applying counter update");
+
+                // If current state has some deleted cells we should return the preimage to CDC.
+                // That's how CDC will know about ineffective writes, so it can omit them (don't log them).
+                if (return_previous_state || (mopt && m_schema->cdc_options().enabled() && has_deleted_counter_cell(*mopt))) {
+                    mopt_persist = std::move(mopt);
```
kbr-, line 1366: what is `mopt_persist`?

service/storage_proxy.cc
```diff
@@ -1817,11 +1817,32 @@ storage_proxy::mutate_counter_on_leader_and_replicate(const schema_ptr& s, froze
...
+            auto cdc_ptr = service::get_local_storage_proxy()._cdc;
+            if (cdc_ptr && cdc_ptr->counter_needs_cdc_augmentation({delta})) {
+                cdc_fut = cdc_ptr->calc_counter_mutation_augmentation(timeout, std::move(delta), std::move(old_state), trace_state).then(
+                        [cl, timeout, trace_state, p](std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>> t) mutable {
+                    auto cdc_mutations = std::move(std::get<0>(t));
+                    if (cdc_mutations.empty()) {
+                        return make_ready_future<>();
+                    }
+                    auto cdc_tracker = std::move(std::get<1>(t));
+                    return service::get_local_storage_proxy().mutate_internal(std::move(cdc_mutations), cl, false, trace_state, p, timeout, std::move(cdc_tracker));
```
kbr-, line 1839: So we perform 2 round trips: one to apply the cdc mutation, the other to apply the counter mutation.
Can't we attach the cdc mutation to the `replicate_counter_from_leader` path?

database.cc
jul-stas: Yes, this is the state of counters before update (kinda 'preimage'), which used to be discarded, but now we need it, so it's moved to a variable with greater scope (more "persistent").

service/storage_proxy.cc
jul-stas: There was such an attempt (and this patch consisted of ~10 LOC), but augmenting counters needs the context of counter tombstones. This context is reliably available only at the moment of counter update. Otherwise we cannot get rid of updates of dead counters in cdc log.

database.cc
jul-stas: true, will do

database.cc
jul-stas: I'm no expert here but... counters have no TTL..? So I assumed that tombstone is a definitive tombstone (deliberately introduced, independent on time). If a node sees it = dead cell; if it doesn't = there's not much we can do.
If you're thinking about inter-node exchange of counter data (incl. tombstones) and synchronization thereof - well, that opens a lot of other questions...

jul-stas:
> @haaawk if postimage is enabled, we query all columns in `pre_image_select`.
> That's how we calculate postimage in the general case.
> 
> My question was regarding the counter column - why don't we calculate postimage for the counter column using the standard way.

At the time of CDC calculation counter updates are already applied. So for counters `preimage_select` yields the postimage (yes). This is for  simplicity and better sync - there's no better place to read the "counter preimage" than in the function that updates them.
Anyway, CQL layer knows nothing about deleted counters (it silently returns OK when you update them, but does nothing in reality). So we need the raw counter-shard "preimage" anyway, with tombstones.

service/storage_proxy.cc
kbr-: I don't see how those two things are related.
There is a moment where you have two (sets of) mutations ready to be sent: `new_state`, and `cdc_mutations`. I'm simply suggesting to send these two mutations in a single vector / in parallel, so we don't do 2 round trips.

database.cc
kbr-: Can it be given a better name or at least a comment near its definition?

service/storage_proxy.cc
```diff
@@ -2222,11 +2243,36 @@ storage_proxy::get_paxos_participants(const sstring& ks_name, const dht::token &
...
+            mutations.erase(cnt_begin, mutations.end());
+
+            return _cdc->augment_mutation_call(timeout, std::move(mutations), tr_state).then(
+                    [this, cl, timeout, tr_state, permit = std::move(permit), raw_counters, cmut = std::move(counter_mutations)]
+                    (std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>&& t) mutable {
+                auto mutations_with_cdc = std::move(std::get<0>(t));
+                auto tracker = std::move(std::get<1>(t));
+                cmut.insert(cmut.end(),
+                        std::make_move_iterator(mutations_with_cdc.begin()),
+                        std::make_move_iterator(mutations_with_cdc.end()));
```
kbr-, line 2266: this is a lot of cdc-specific logic cluttering this `mutate` function whose purpose should be simple to explain: "augment the set of mutations if needed by adding more mutations (e.g. cdc-related), then send those mutations".

The whole "augment" logic should be moved out to a separate function which doesn't know about the "send" part.

database.cc
kbr-: There is no such thing as a "tombstone independent of time". Each tombstone has a timestamp.

That's why the `is_dead(T)` can even make sense: it compares the tombstone's timestamp with `T`. If `T < tombstone_ts`, the cell is not dead, otherwise it is.

We don't want to compare the tombstone with `now`, where `now` is defined according to the local clock of one of the nodes (`gc_clock::now()`), because that doesn't make any sense! The node's clock might be a week into the future compared to the client's clock, and the update's timestamp might use the client's clock. So he wouldn't see the update in the CDC log even though according to his time, the counter is still alive, and will be alive for one more week.

service/storage_proxy.cc
jul-stas: Well, one could argue that `mutate` does not hit close to "augment the set of mutations if needed by adding more mutations (e.g. cdc-related), then send those mutations". But otherwise I agree.

database.cc
kbr-: Sorry, I made a mistake in the second paragraph which resulted in my post not making any sense.

A tombstone cell with timestamp `tombstone_ts` is considered dead at time `T` iff `T >= tombstone_ts`.

Suppose that the update's timestamp is `U`, node A's local clock is `NOW = U + week`, and node B's local clock is `NOW = U + 1 day`. Suppose there is a tombstone `T = U + 6 days`.

If node B reads the counter, it will show a value (for 5 more days). But if node A reads the counter, it will show that it's dead.
Therefore, when we perform the update with timestamp `U`, we should still log it because `U < T`. There are (or may be) nodes that still consider the counter alive.

jul-stas:
CDC underwent substantial changes in the meantime. This is no longer relevant.

PR #6176, opened 2020-04-10
api: enable/disable auto compaction

### LINKED ISSUE #440: nodetool disableautocompaction
Disables autocompaction for a keyspace and one or more tables.
http://docs.datastax.com/en/cassandra/2.1/cassandra/tools/toolsDisableAutoCompaction.html

Can be useful when compaction take too much space
http://nieksand.tumblr.com/

### LINKED ISSUE #1488: Add support to compaction property to enable/disable background compaction
Details of the property:

name: enabled   default value: true Set to enable background compaction.

### LINKED ISSUE #1808: non-intrusive way for disabling automatic compaction for a cf
currently, it's possible to disable automatic compaction for a column family by setting its strategy to NullCompactionStrategy, but that's too intrusive. It would be nice to introduce an API that would allow user to enable/disable automatic compaction.

sitano:
The patch implements:

- /storage_service/auto_compaction API endpoint
- /column_family/autocompaction/{name} API endpoint

Those APIs allow to control and request the status of background
compaction jobs for the existing tables.

The implementation introduces the table::_compaction_disabled_by_user.
Then the CompactionManager checks if it can push the background
compaction job for the corresponding table.

New members
===

    table::enable_auto_compaction();
    table::disable_auto_compaction();
    bool table::is_auto_compaction_disabled_by_user() const

Test
===
Tests: unit(sstable_datafile_test autocompaction_control_test), manual

    $ ninja build/dev/test/boost/sstable_datafile_test
    $ ./build/dev/test/boost/sstable_datafile_test --run_test=autocompaction_control_test -- -c1 -m2G --overprovisioned --unsafe-bypass-fsync 1 --blocked-reactor-notify-ms 2000000

The test tries to submit a compaction job after playing
with autocompaction control table switch. However, there is
no reliable way to hook pending compaction task. The code
assumed that with_scheduling_group() closure will never
preempt execution of the stats check.

Revert
===
Reverts commit c8247ac. In previous version the execution
sometimes resulted into the following error:

    test/boost/sstable_datafile_test.cc(1076): fatal error: in "autocompaction_control_test":
    critical check cm->get_stats().pending_tasks == 1 || cm->get_stats().active_tasks == 1 has failed

This version adds a few sstables to the cf, starts
the compaction and awaits until it is finished.

API change
===

- `/column_family/autocompaction/` always returned `true` while answering to the question: if the autocompaction disabled (see https://github.com/scylladb/scylla-jmx/blob/master/src/main/java/org/apache/cassandra/db/ColumnFamilyStore.java#L321). now it answers to the question: if the autocompaction for specific table is enabled. The question logic is inverted. The patch to the JMX is required. However, the change is decent because all old values were invalid (it always reported all compactions are disabled).
- `/column_family/autocompaction/` got support for POST/DELETE per table 

Fixes
===

Fixes #1488
Fixes #1808
Fixes #440

### DIFF
--- a/api/storage_service.cc
+++ b/api/storage_service.cc
@@ -89,6 +89,23 @@ static auto wrap_ks_cf(http_context &ctx, ks_cf_func f) {
     };
 }
 
+future<> set_tables_autocompaction(http_context& ctx, const sstring &keyspace, std::vector<sstring> tables, bool enabled) {
+    if (tables.empty()) {
+        tables = map_keys(ctx.db.local().find_keyspace(keyspace).metadata().get()->cf_meta_data());
+    }
+    return ctx.db.invoke_on_all([keyspace, tables, enabled] (database& db) {
+        return parallel_for_each(tables, [&db, keyspace, enabled](const sstring& table) mutable {
+            column_family& cf = db.find_column_family(keyspace, table);
+            if (enabled) {
+                cf.enable_auto_compaction();
+            } else {
+                cf.disable_auto_compaction();
+            }
+            return make_ready_future<>();
+        });
+    });
+}
+
 void set_storage_service(http_context& ctx, routes& r) {
     ss::local_hostid.set(r, [](std::unique_ptr<request> req) {
         return db::system_keyspace::get_local_host_id().then([](const utils::UUID& id) {
@@ -703,19 +720,19 @@ void set_storage_service(http_context& ctx, routes& r) {
     });
 
     ss::enable_auto_compaction.set(r, [&ctx](std::unique_ptr<request> req) {
-        //TBD
-        unimplemented();
         auto keyspace = validate_keyspace(ctx, req->param);
-        auto column_family = req->get_query_param("cf");
-        return make_ready_future<json::json_return_type>(json_void());
+        auto tables = split_cf(req->get_query_param("cf"));
+        return set_tables_autocompaction(ctx, keyspace, tables, true).then([]{
+            return make_ready_future<json::json_return_type>(json_void());
+        });
     });
 
     ss::disable_auto_compaction.set(r, [&ctx](std::unique_ptr<request> req) {
-        //TBD
-        unimplemented();
         auto keyspace = validate_keyspace(ctx, req->param);
-        auto column_family = req->get_query_param("cf");
-        return make_ready_future<json::json_return_type>(json_void());
+        auto tables = split_cf(req->get_query_param("cf"));
+        return set_tables_autocompaction(ctx, keyspace, tables, false).then([]{
+            return make_ready_future<json::json_return_type>(json_void());
+        });
     });
 
     ss::deliver_hints.set(r, [](std::unique_ptr<request> req) {
--- a/test/boost/sstable_datafile_test.cc
+++ b/test/boost/sstable_datafile_test.cc
@@ -5702,3 +5702,78 @@ SEASTAR_TEST_CASE(twcs_major_compaction_test) {
         BOOST_REQUIRE(ret.new_sstables.size() == 2);
     });
 }
+
+SEASTAR_TEST_CASE(autocompaction_control_test) {
+    return test_env::do_with_async([] (test_env& env) {
+        storage_service_for_tests ssft;
+        cell_locker_stats cl_stats;
+        cache_tracker tracker;
+
+        compaction_manager cm; cm.start();
+
+        auto s = schema_builder(some_keyspace, some_column_family)
+                .with_column("id", utf8_type, column_kind::partition_key)
+                .with_column("value", int32_type)
+                .build();
+
+        auto tmp = tmpdir();
+        column_family::config cfg = column_family_test_config();
+        cfg.datadir = tmp.path().string();
+        cfg.enable_commitlog = false;
+        cfg.enable_disk_writes = true;
+
+        auto cf = make_lw_shared<column_family>(s, cfg, column_family::no_commitlog(), cm, cl_stats, tracker);
+        cf->set_compaction_strategy(sstables::compaction_strategy_type::size_tiered);
+        cf->mark_ready_for_writes();
+
+        // no compactions done yet
+        auto& ss = cm.get_stats();
+        BOOST_REQUIRE(ss.pending_tasks == 0 && ss.active_tasks == 0 && ss.completed_tasks == 0);
+        // auto compaction is enabled by default
+        BOOST_REQUIRE(!cf->is_auto_compaction_disabled_by_user());
+        // disable auto compaction by user
+        cf->disable_auto_compaction();
+        // check it is disabled
+        BOOST_REQUIRE(cf->is_auto_compaction_disabled_by_user());
+
+        // generate a few sstables
+        auto sst_gen = [&env, s, &tmp, gen = make_lw_shared<unsigned>(1)] () mutable {
+            return env.make_sstable(s, tmp.path().string(), (*gen)++, la, big);
+        };
+        auto make_insert = [&] (partition_key key) {
+            mutation m(s, key);
+            m.set_clustered_cell(clustering_key::make_empty(), bytes("value"), data_value(int32_t(1)), 1 /* ts */);
+            return m;
+        };
+        auto min_threshold = cf->schema()->min_compaction_threshold();
+        auto tokens = token_generation_for_current_shard(1);
+        for (auto i = 0; i < 2 * min_threshold; ++i) {
+            auto key = partition_key::from_exploded(*s, {to_bytes(tokens[0].first)});
+            auto mut = make_insert(key);
+            auto sst = make_sstable_containing(sst_gen, {mut});
+            cf->add_sstable_and_update_cache(sst).wait();
+        }
+
+        // check compaction manager does not receive background compaction submissions
+        cf->start();
+        cf->trigger_compaction();
+        cf->get_compaction_manager().submit(cf.get());
+        BOOST_REQUIRE(ss.pending_tasks == 0 && ss.active_tasks == 0 && ss.completed_tasks == 0);
+        // enable auto compaction
+        cf->enable_auto_compaction();
+        // check enabled
+        BOOST_REQUIRE(!cf->is_auto_compaction_disabled_by_user());
+        // trigger background compaction
+        cf->trigger_compaction();
+        // wait until compaction finished
+        do_until([&ss] { return ss.pending_tasks == 0 && ss.active_tasks == 0; }, [] {
+            return sleep(std::chrono::milliseconds(100));
+        }).wait();
+        // test compaction successfully finished
+        BOOST_REQUIRE(ss.errors == 0);
+        BOOST_REQUIRE(ss.completed_tasks == 1);
+
+        cf->stop().wait();
+        cm.stop().wait();
+    });
+}
--- a/api/api-doc/column_family.json
+++ b/api/api-doc/column_family.json
@@ -380,16 +380,54 @@
          "operations":[
             {
                "method":"GET",
-               "summary":"check if the auto compaction disabled",
+               "summary":"check if the auto_compaction property is enabled for a given table",
                "type":"boolean",
-               "nickname":"is_auto_compaction_disabled",
+               "nickname":"get_auto_compaction",
                "produces":[
                   "application/json"
                ],
                "parameters":[
                   {
                      "name":"name",
-                     "description":"The column family name in keyspace:name format",
+                     "description":"The table name in keyspace:name format",
+                     "required":true,
+                     "allowMultiple":false,
+                     "type":"string",
+                     "paramType":"path"
+                  }
+               ]
+            },
+            {
+               "method":"POST",
+               "summary":"Enable table auto compaction",
+               "type":"void",
+               "nickname":"enable_auto_compaction",
+               "produces":[
+                  "application/json"
+               ],
+               "parameters":[
+                  {
+                     "name":"name",
+                     "description":"The table name in keyspace:name format",
+                     "required":true,
+                     "allowMultiple":false,
+                     "type":"string",
+                     "paramType":"path"
+                  }
+               ]
+            },
+            {
+               "method":"DELETE",
+               "summary":"Disable table auto compaction",
+               "type":"void",
+               "nickname":"disable_auto_compaction",
+               "produces":[
+                  "application/json"
+               ],
+               "parameters":[
+                  {
+                     "name":"name",
+                     "description":"The table name in keyspace:name format",
                      "required":true,
                      "allowMultiple":false,
                      "type":"string",
--- a/api/column_family.cc
+++ b/api/column_family.cc
@@ -839,11 +839,26 @@ void set_column_family(http_context& ctx, routes& r) {
         return make_ready_future<json::json_return_type>(res);
     });
 
-    cf::is_auto_compaction_disabled.set(r, [] (const_req req) {
-        // FIXME
-        // currently auto compaction is disable
-        // it should be changed when it would have an API
-        return true;
+    cf::get_auto_compaction.set(r, [&ctx] (const_req req) {
+        const utils::UUID& uuid = get_uuid(req.param["name"], ctx.db.local());
+        column_family& cf = ctx.db.local().find_column_family(uuid);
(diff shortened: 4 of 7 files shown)


raphaelsc:
Erasing the condition makes the unit test weak.

I think your work on sitano@6d51c0d is what I expect. BTW, make_sstable_containing() can be used to simplify that code. Look for examples in sstable_datafile_test. Only 1 sstable is enough for the test purposes but you need to set schema's min threshold to 1.

sitano:
@raphaelsc thanks! I will look into that and resubmit the patch.

sitano:
@raphaelsc ok, the work is done. `make_sstable_containing()` helped. setting min threshold did not help, because setting it to 1 makes compaction task never stop. It tries continually to compact 1 sstable into 1 sstable and never succeeds. so I add 2 sstables, submit compaction, wait for the completion and check for errors.

sitano:
maybe settings `cf->set_compaction_strategy(sstables::compaction_strategy_type::size_tiered);` is excessive but everyone doing that.

sitano:
there is great potential for simplifying neighbor tests by making helper functions global and unifying CF Schemas used across tests. besides that, I could not reuse `column_family_for_tests` because it builds non-durable table (maybe there is a workaround for that but I didn't have time to dig any further).

penberg:
@raphaelsc @glommer Please have a look at this.

@sitano, please include a revert of the commit in master and amend these changes to it so that when it's time to merge, I can just pull and not any manual steps (and introduce bugs).

sitano:
@penberg ok, a question: shall I put PRs target branch for `next`, not for the `master`? (to allow merged changes to live a bit before getting into main branch) (do we have any github flow suggestions?)

penberg:
You should always target `next`. That's where we merge new code, which is then tested by our Jenkins infrastructure, and pushed to `master` if gating tests pass.

sitano:
@penberg ok, I am sorry my prev patch went into the master bypassing next. I didn't know that.

penberg:
@sitano No it didn't, I merged it by hand to the correct branch :)

test/boost/sstable_datafile_test.cc
```diff
@@ -1036,6 +1036,76 @@ static flat_mutation_reader make_normalizing_sstable_reader(shared_sstable sst,
     return make_normalizing_sstable_reader(sst, s, query::full_partition_range);
 }
 
+SEASTAR_TEST_CASE(autocompaction_control_test) {
+    return test_env::do_with_async([] (test_env& env) {
+        storage_service_for_tests ssft;
+        BOOST_REQUIRE(smp::count == 1);
```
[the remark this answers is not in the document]
glommer: BTW, I know you copied this from another test that probably does this, but it's best not to add more occurrences of this mistake

test/boost/sstable_datafile_test.cc
```diff
@@ -1036,6 +1036,76 @@ static flat_mutation_reader make_normalizing_sstable_reader(shared_sstable sst,
...
+            return env.make_sstable(s, tmp.path().string(), (*gen)++, la, big);
+        };
+
+        auto make_insert = [&] (partition_key key) {
+            mutation m(s, key);
+            m.set_clustered_cell(clustering_key::from_exploded(*s, {to_bytes("c1")}), bytes("r1"), data_value(int32_t(1)), 1);
+            return m;
+        };
+
+        auto alpha = partition_key::from_exploded(*s, {to_bytes("alpha")});
```
glommer, line 1075: this is the reason you need smp ==1 in the first place, and we currently have a way to work around it.

There is a function in `test/lib/sstable_utils.cc` called `token_generation_shard`

Please see an example of its usage in the resharding test:

```
       auto key_token_pair = token_generation_for_shard(keys_per_shard, i, cfg->murmur3_partitioner_ignore_msb_bits());
            BOOST_REQUIRE(key_token_pair.size() == keys_per_shard);
```

It is important to pass the correct msb. There is another function btw that is `token_generation_for_shard` it is probably broken as it doesn't pass the right msb. That one is actually used in the sstable_datafile_test which is why we don't notice, but don't use that.

test/boost/sstable_datafile_test.cc
```diff
@@ -1036,6 +1036,76 @@ static flat_mutation_reader make_normalizing_sstable_reader(shared_sstable sst,
...
+
+        column_family_test(cf).add_sstable(make_sstable_containing(sst_gen, {make_insert(alpha)}));
+        column_family_test(cf).add_sstable(make_sstable_containing(sst_gen, {make_insert(alpha)}));
```
glommer, line 1078: By adding two SSTables, there is a risk a compaction will happen. In particular, you set the threshold to two, so I have no idea why it doesn't happen. Maybe it does happen and that's another point where the test can get flaky.

It's best to disable the compaction first, then add the SSTables.

test/boost/sstable_datafile_test.cc
```diff
@@ -1036,6 +1036,76 @@ static flat_mutation_reader make_normalizing_sstable_reader(shared_sstable sst,
...
+        // check CompactionManager does not receive background compaction submissions
+        cf->trigger_compaction();
+        cf->get_compaction_manager().submit(cf.get());
+        BOOST_REQUIRE(cm->get_stats().pending_tasks == 0 && cm->get_stats().active_tasks == 0 && cm->get_stats().completed_tasks == 0);
+        // enable auto compaction
+        cf->enable_auto_compaction();
+        // check enabled
+        BOOST_REQUIRE(!cf->is_auto_compaction_disabled_by_user());
+        // trigger background compaction
+        cf->trigger_compaction();
```
glommer, line 1095: Can we check the stats here to make sure the compaction indeed happened ?

glommer: I see that this is indeed something you had and removed due to flakiness.

Take a look at `test/lib/eventually.hh`

It can help you both here and in the `do_until` look below

glommer:
@sitano as the patch is already now solidly reverted, let's just add the changelog from scratch.

You can add, at the bottom, a description saying it was formerly reverted before and what is different.

test/boost/sstable_datafile_test.cc
sitano: Sadly saying, this:
```
    auto key_and_token_pair = token_generation_for_current_shard(50);
    auto min_key = key_and_token_pair[0].first;
    auto max_key = key_and_token_pair[key_and_token_pair.size()-1].first;
    auto max_sstable_size_in_mb = 1;
    auto max_sstable_size = max_sstable_size_in_mb*1024*1024;

    // Creating two sstables which key range overlap.
    add_sstable_for_leveled_test(env, cf, /*gen*/1, max_sstable_size, /*level*/0, min_key, max_key);
    BOOST_REQUIRE(cf->get_sstables()->size() == 1);

    add_sstable_for_leveled_test(env, cf, /*gen*/2, max_sstable_size, /*level*/0, key_and_token_pair[1].first, max_key);
    BOOST_REQUIRE(cf->get_sstables()->size() == 2);
```
did not help. table compaction fails with an assertion `sstables/sstables.cc:2538: uint64_t sstables::sstable::bytes_on_disk() const: Assertion `_bytes_on_disk > 0' failed.`
I will have to add ss tables manually with mutations.

sitano:
```
auto alpha = partition_key::from_exploded(*s, {to_bytes("alpha")});
```

> this is the reason you need smp ==1 in the first place, and we currently have a way to work around it.

@glommer Why? I don't see the connection between `smp` number and what I put into sstable. As far as I see `partition_key::from_exploded` only serializes a key. I don't know why it is called `exploded` btw. My guess is that you meant that maybe I need properly generated tokens for building overlapping sstables? It seems like everything works without it. However, I am not sure if it is right.
 
> There is a function in `test/lib/sstable_utils.cc` called `token_generation_shard`

No test uses this function. However, almost all of them use an alias to it `token_generation_for_current_shard(tokens_to_generate) = token_generation_for_shard(tokens_to_generate, this_shard_id(), unsigned ignore_msb = 0, unsigned smp_count = smp::count);` either for generating overlapping sstables or for uses tokens instead of the keys. The rest half of the tests just uses `from_explode` to generate keys for mutations.

Basically here what they do:

```
        auto make_insert = [&] (partition_key key) {
            mutation m(s, key);
            // TODO: replace this with token_generation_shard
            m.set_clustered_cell(clustering_key::from_exploded(*s, {to_bytes("c1")}), bytes("r1"), data_value(int32_t(1)), 1);
            return m;
        };

        // TODO: replace this with token_generation_shard
        auto alpha = partition_key::from_exploded(*s, {to_bytes("alpha")});

        column_family_test(cf).add_sstable(make_sstable_containing(sst_gen, {make_insert(alpha)}));
        column_family_test(cf).add_sstable(make_sstable_containing(sst_gen, {make_insert(alpha)}));
```

or 

```
 auto tokens = token_generation_for_current_shard(4);
        auto make_insert = [&] (auto p) {
            auto key = partition_key::from_exploded(*s, {to_bytes(p.first)});
            mutation m(s, key);
            m.set_clustered_cell(clustering_key::make_empty(), bytes("value"), data_value(int32_t(1)), 1 /* ts */);
            return m;
        };
        auto mut1 = make_insert(tokens[0]);
        auto mut2 = make_insert(tokens[1]);
        auto mut3 = make_insert(tokens[2]);
        auto mut4 = make_insert(tokens[3]);
        std::vector<shared_sstable> ssts = {
                make_sstable_containing(sst_gen, {mut1, mut2}),
                make_sstable_containing(sst_gen, {mut3, mut4})
        };
```

from my perspective of very limited knowledge of tables internals and my goals, it looks almost the same. So the question remains, how SMP is connected to the from_exploded.

sitano: > this is the reason you need smp ==1 in the first place, and we currently have a way to work around it.

My guess is it is because we want to put proper tokens into the sstables relative to shard?

test/boost/sstable_datafile_test.cc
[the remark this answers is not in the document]
sitano: This one is removed

test/boost/sstable_datafile_test.cc
sitano: :+1:

test/boost/sstable_datafile_test.cc
sitano: It is there

sitano:
@glommer I reworked the test. Take a look.

sitano:
I was thinking that maybe it could have sense to extend `column_family::config` with `enable_background_compations = true` but I am not sure everyone are gonna like it.

api/storage_service.cc
```diff
@@ -89,6 +89,23 @@ static auto wrap_ks_cf(http_context &ctx, ks_cf_func f) {
     };
 }
 
+future<> set_tables_autocompaction(http_context& ctx, const sstring &keyspace, std::vector<sstring> tables, bool enabled) {
```
glommer, line 92: @amnonh why do we need two ways two control auto compaction - one in the storage service and another at the column family level ?

Which ones does scylla-jmx use ? I am okay keeping both if there is a compat reason but I'd prefer to have just one

test/boost/sstable_datafile_test.cc
```diff
@@ -5702,3 +5702,78 @@ SEASTAR_TEST_CASE(twcs_major_compaction_test) {
...
+            return env.make_sstable(s, tmp.path().string(), (*gen)++, la, big);
+        };
+        auto make_insert = [&] (partition_key key) {
+            mutation m(s, key);
+            m.set_clustered_cell(clustering_key::make_empty(), bytes("value"), data_value(int32_t(1)), 1 /* ts */);
+            return m;
+        };
+        auto min_threshold = cf->schema()->min_compaction_threshold();
+        auto tokens = token_generation_for_current_shard(1);
+        for (auto i = 0; i < min_threshold; i ++) {
```
glommer, line 5750: `i++`, no space between i and `+`,
`++i` is even better as it will match the style of the rest of Scylla

However more importantly, if you look until `< min_threshold` you will likely never trigger a compaction. You need to go over it. Let's set `2 * min_threshold` as the loop boundary.

test/boost/sstable_datafile_test.cc
glommer: well yes, but there is now a much better way to do it - as you did. This can be resolved.

glommer:
> I was thinking that maybe it could have sense to extend `column_family::config` with `enable_background_compations = true` but I am not sure everyone are gonna like it.

You're right, I don't like it.

api/storage_service.cc
sitano: `scylla-jmx` only uses 1 API for the column_family
```
    @Override
    public boolean isAutoCompactionDisabled() {
        log(" isAutoCompactionDisabled()");
        return client.getBooleanValue("column_family/autocompaction/" + getCFName());
    }
```
but I thought it would be cool to have it consistent with the storage service. The only difference in the naming is the underscore:

`/column_family/autocompaction/` vs `/storage_service/auto_compaction` which I would happy to reduce to 1 variant, however, it would make jmx incompat

glommer: I would prefer not to expose two APIs for that.

If scylla-jmx is using only column family, let's please add just that one.
It looks simpler than storage service anyway

test/boost/sstable_datafile_test.cc
sitano: :+1: done

glommer:
I approved the PR, it looks good.

However it seems to me from talking to you privately that you had run these tests in dev mode only.
Please do a debug mode run and let us know when it is done. Assuming no issues this is good to go

sitano:
@glommer I have executed tests in both Debug and Release mode for some time. They worked.

```
./build/debug/test/boost/sstable_datafile_test --run_test=autocompaction_control_test -- -c1 -m2G --overprovisioned --unsafe-bypass-fsync 1 --blocked-reactor-notify-ms 2000000
WARNING: debug mode. Not for benchmarking or production
random-seed=2622011243
Running 1 test case...
==1580444==WARNING: ASan doesn't fully support makecontext/swapcontext functions and may produce false positives in some cases!
INFO  2020-04-29 17:02:52,370 [shard 0] features - Feature RANGE_TOMBSTONES is enabled
...
INFO  2020-04-29 17:02:52,370 [shard 0] features - Feature LWT is enabled
INFO  2020-04-29 17:02:52,370 [shard 0] features - Feature PER_TABLE_PARTITIONERS is enabled
INFO  2020-04-29 17:02:52,563 [shard 0] compaction - Compacting [/tmp/scylla-7a99b28d-c327-4320-a858-4a49706b6add/la-5-big-Data.db:level=0, /tmp/scylla-7a99b28d-c327-4320-a858-4a49706b6add/la-4-big-Data.db:level=0, /tmp/scylla-7a99b28d-c327-4320-a858-4a49706b6add/la-3-big-Data.db:level=0, /tmp/scylla-7a99b28d-c327-4320-a858-4a49706b6add/la-7-big-Data.db:level=0, /tmp/scylla-7a99b28d-c327-4320-a858-4a49706b6add/la-2-big-Data.db:level=0, /tmp/scylla-7a99b28d-c327-4320-a858-4a49706b6add/la-8-big-Data.db:level=0, /tmp/scylla-7a99b28d-c327-4320-a858-4a49706b6add/la-6-big-Data.db:level=0, /tmp/scylla-7a99b28d-c327-4320-a858-4a49706b6add/la-1-big-Data.db:level=0, ]
INFO  2020-04-29 17:02:52,608 [shard 0] compaction - Compacted 8 sstables to [/tmp/scylla-7a99b28d-c327-4320-a858-4a49706b6add/mc-1-big-Data.db:level=0, ]. 38592 bytes to 6236 (~16% of original) in 27ms = 0.22MB/s. ~1024 total partitions merged to 1.
INFO  2020-04-29 17:02:52,663 [shard 0] compaction_manager - Asked to stop
INFO  2020-04-29 17:02:52,663 [shard 0] compaction_manager - Stopped

*** No errors detected
```

glommer:
Thanks @sitano 

@penberg do you mind taking a look at this ? Same patch as last time pretty much, with the test fixed.

penberg:
Queued as 84e25e8ba4a40e41a483bec2f28b9fa068b14258.

PR #6194, opened 2020-04-14
repair: Ignore table removed in sync_data_using_repair

asias:
Commit 75cf255c67641db7255286908c8b4f5a67590f9c (repair: Ignore keyspace
that is removed in sync_data_using_repair) is not enough to fix the
issue because when the repair master checks if the table is dropped, the
table might not be dropped yet on the repair master.

To fix, the repair master should check if the follower failed the repair
because the table is dropped by checking the error returned from
follower.

With this patch, we would see

WARN  2020-04-14 11:19:00,417 [shard 0] repair - repair id 1 on shard 0
completed successfully, keyspace=ks, ignoring dropped tables={cf}

when the table is dropped during bootstrap.

Tests: update_cluster_layout_tests.py:TestUpdateClusterLayout.simple_add_new_node_while_schema_changes_test

Fixes: #5942

### DIFF
--- a/message/messaging_service.cc
+++ b/message/messaging_service.cc
@@ -1193,14 +1193,14 @@ future<> messaging_service::send_repair_put_row_diff(msg_addr id, uint32_t repai
 }
 
 // Wrapper for REPAIR_ROW_LEVEL_START
-void messaging_service::register_repair_row_level_start(std::function<future<> (const rpc::client_info& cinfo, uint32_t repair_meta_id, sstring keyspace_name, sstring cf_name, dht::token_range range, row_level_diff_detect_algorithm algo, uint64_t max_row_buf_size, uint64_t seed, unsigned remote_shard, unsigned remote_shard_count, unsigned remote_ignore_msb, sstring remote_partitioner_name, table_schema_version schema_version, rpc::optional<streaming::stream_reason> reason)>&& func) {
+void messaging_service::register_repair_row_level_start(std::function<future<repair_row_level_start_response> (const rpc::client_info& cinfo, uint32_t repair_meta_id, sstring keyspace_name, sstring cf_name, dht::token_range range, row_level_diff_detect_algorithm algo, uint64_t max_row_buf_size, uint64_t seed, unsigned remote_shard, unsigned remote_shard_count, unsigned remote_ignore_msb, sstring remote_partitioner_name, table_schema_version schema_version, rpc::optional<streaming::stream_reason> reason)>&& func) {
     register_handler(this, messaging_verb::REPAIR_ROW_LEVEL_START, std::move(func));
 }
 future<> messaging_service::unregister_repair_row_level_start() {
     return unregister_handler(messaging_verb::REPAIR_ROW_LEVEL_START);
 }
-future<> messaging_service::send_repair_row_level_start(msg_addr id, uint32_t repair_meta_id, sstring keyspace_name, sstring cf_name, dht::token_range range, row_level_diff_detect_algorithm algo, uint64_t max_row_buf_size, uint64_t seed, unsigned remote_shard, unsigned remote_shard_count, unsigned remote_ignore_msb, sstring remote_partitioner_name, table_schema_version schema_version, streaming::stream_reason reason) {
-    return send_message<void>(this, messaging_verb::REPAIR_ROW_LEVEL_START, std::move(id), repair_meta_id, std::move(keyspace_name), std::move(cf_name), std::move(range), algo, max_row_buf_size, seed, remote_shard, remote_shard_count, remote_ignore_msb, std::move(remote_partitioner_name), std::move(schema_version), reason);
+future<rpc::optional<repair_row_level_start_response>> messaging_service::send_repair_row_level_start(msg_addr id, uint32_t repair_meta_id, sstring keyspace_name, sstring cf_name, dht::token_range range, row_level_diff_detect_algorithm algo, uint64_t max_row_buf_size, uint64_t seed, unsigned remote_shard, unsigned remote_shard_count, unsigned remote_ignore_msb, sstring remote_partitioner_name, table_schema_version schema_version, streaming::stream_reason reason) {
+    return send_message<rpc::optional<repair_row_level_start_response>>(this, messaging_verb::REPAIR_ROW_LEVEL_START, std::move(id), repair_meta_id, std::move(keyspace_name), std::move(cf_name), std::move(range), algo, max_row_buf_size, seed, remote_shard, remote_shard_count, remote_ignore_msb, std::move(remote_partitioner_name), std::move(schema_version), reason);
 }
 
 // Wrapper for REPAIR_ROW_LEVEL_STOP
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -740,7 +740,11 @@ void repair_info::check_failed_ranges() {
         rlogger.info("repair id {} on shard {} failed - {} ranges failed", id, shard, nr_failed_ranges);
         throw std::runtime_error(format("repair {:d} on shard {:d} failed to do checksum for {:d} sub ranges", id, shard, nr_failed_ranges));
     } else {
-        rlogger.info("repair id {} on shard {} completed successfully", id, shard);
+        if (dropped_tables.size()) {
+            rlogger.warn("repair id {} on shard {} completed successfully, keyspace={}, ignoring dropped tables={}", id, shard, keyspace, dropped_tables);
+        } else {
+            rlogger.info("repair id {} on shard {} completed successfully, keyspace={}", id, shard, keyspace);
+        }
     }
 }
 
@@ -1027,7 +1031,16 @@ static future<> repair_range(repair_info& ri, const dht::token_range& range) {
         return do_for_each(ri.cfs.begin(), ri.cfs.end(), [&ri, &neighbors, range] (auto&& cf) {
             ri._sub_ranges_nr++;
             if (ri.row_level_repair()) {
-                return repair_cf_range_row_level(ri, cf, range, neighbors);
+                if (ri.dropped_tables.count(cf)) {
+                    return make_ready_future<>();
+                }
+                return repair_cf_range_row_level(ri, cf, range, neighbors).handle_exception_type([&ri, cf] (no_such_column_family&) mutable {
+                    ri.dropped_tables.insert(cf);
+                    return make_ready_future<>();
+                }).handle_exception([&ri] (std::exception_ptr ep) mutable {
+                    ri.nr_failed_ranges++;
+                    return make_exception_future<>(std::move(ep));
+                });
             } else {
                 return repair_cf_range(ri, cf, range, neighbors);
             }
--- a/repair/repair.hh
+++ b/repair/repair.hh
@@ -205,6 +205,7 @@ public:
     repair_stats _stats;
     bool _row_level_repair;
     uint64_t _sub_ranges_nr = 0;
+    std::unordered_set<sstring> dropped_tables;
 public:
     repair_info(seastar::sharded<database>& db_,
             const sstring& keyspace_,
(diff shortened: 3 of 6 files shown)


slivne:
@nyh  / @avikivity this is related to repair based operations - we need this for 4.0 - please prioritize

repair/repair.cc
```diff
@@ -1025,7 +1029,22 @@ static future<> repair_range(repair_info& ri, const dht::token_range& range) {
...
-                return repair_cf_range_row_level(ri, cf, range, neighbors);
+                if (ri.dropped_tables.count(cf)) {
+                    return make_ready_future<>();
+                }
+                return repair_cf_range_row_level(ri, cf, range, neighbors).handle_exception([&ri, cf] (std::exception_ptr ep) mutable {
+                    try {
+                        std::rethrow_exception(ep);
+                    } catch (const no_such_column_family& e) {
+                        if (!ri.ignore_dropped_tables) {
+                            ri.nr_failed_ranges++;
```
bhalevy, line 1040: The ignore_dropped_tables logic seems overly complicated.
Why not return exception future in this case?

repair/row_level.cc
```diff
@@ -2486,10 +2487,13 @@ class row_level_repair {
                     send_missing_rows_to_follower_nodes(master);
                 }
             } catch (std::exception& e) {
+                std::string err(e.what());
+                if (err.find("Can't find a column family") == 0) {
```
bhalevy, line 2491: string matching is brittle. why not catch the no_such_column_family exception?

bhalevy: or rethrow here and catch it to handle the specific exception

asias: I hate the string match here too, but the error is from rpc. rpc loses the exception type, it is std::runtime_error.

repair/repair.cc
asias: Because we want  the repair to continue to repair the result of the tables if one of the tables is dropped. If we return exception future here, the rest of the tables will be ignored.

asias: The logic is this:  if we see no_such_column_family exception, we always continue to repair the rest of the tables.  The ignore_dropped_tables controls if we treat the repair as a failure or success. 

For normal repair operations, ignore_dropped_tables = false, the repair will be treated an a failure. 

For repair based node operations, ignore_dropped_tables = true, the repair will be treated as a success.

repair/repair.cc
```diff
@@ -1025,7 +1029,22 @@ static future<> repair_range(repair_info& ri, const dht::token_range& range) {
         return do_for_each(ri.cfs.begin(), ri.cfs.end(), [&ri, &neighbors, range] (auto&& cf) {
             ri._sub_ranges_nr++;
             if (ri.row_level_repair()) {
-                return repair_cf_range_row_level(ri, cf, range, neighbors);
+                if (ri.dropped_tables.count(cf)) {
+                    return make_ready_future<>();
+                }
+                return repair_cf_range_row_level(ri, cf, range, neighbors).handle_exception([&ri, cf] (std::exception_ptr ep) mutable {
+                    try {
```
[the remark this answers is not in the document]
asias: ok.

asias:
Changes in e00d01bc935db825b2f1db620e3740ceb8a6565b:
- Use handle_exception_type
- Wrap creation of row_level_repair object inside futurize_invoke since row_level_repair can throw if the  table is dropped

repair/row_level.cc
bhalevy: got it, thanks

slivne:
@nyh / @avikivity this is related to repair based operations - we need this for 4.0 - please prioritize

repair/repair.cc
```diff
@@ -1025,7 +1029,19 @@ static future<> repair_range(repair_info& ri, const dht::token_range& range) {
         return do_for_each(ri.cfs.begin(), ri.cfs.end(), [&ri, &neighbors, range] (auto&& cf) {
             ri._sub_ranges_nr++;
             if (ri.row_level_repair()) {
-                return repair_cf_range_row_level(ri, cf, range, neighbors);
+                if (ri.dropped_tables.count(cf)) {
+                    return make_ready_future<>();
```
avikivity, line 1033: The whole thing is vulerable to drop/create races. If we drop a table and re-create it with the same name, will repair notice? Note the new table can have a completely different schema.

We should work with table uuids, not names. A drop/create will result in a new uuid even if they have the same name. It's also better for rename, if we ever support it.

repair/repair.hh
```diff
@@ -204,6 +204,8 @@ public:
     repair_stats _stats;
     bool _row_level_repair;
     uint64_t _sub_ranges_nr = 0;
+    std::unordered_set<sstring> dropped_tables;
+    bool ignore_dropped_tables = false;
```
avikivity, line 208: When would we want this to be false?

avikivity: Note it can also refer to a newly added table, if the repair master saw it but the follower hasn't seen it yet.

repair/row_level.cc
avikivity, line 2491: This is extermely clunky. Let's find a better way to communicate this.

asias: I do not like this neither. Unfortunately, RPC losts the exception type info.  The proper fix would be in rpc layter.  @gleb-cloudius what do you think?

repair/repair.hh
asias: Repair master and followers will use the same schema version choose by the repair master. So the master and follower will see the same schema. 

```
commit 3db136f81e2f70593d95d13a28c65281287c7492
Author: Asias He <asias@scylladb.com>
Date:   Fri Jun 14 09:15:26 2019 +0800

    repair: Use the same schema version for repair master and followers
    
    Before this patch, repair master and followers use their own schema
    version at the point repair starts independently. The schemas can be
    different due to schema change. Repair uses the schema to serialize
    mutation_fragment and deserialize the mutation_fragment received from
    peer nodes. Using different schema version to serialize and deserialize
    cause undefined behaviour.
    
    To fix, we use the schema the repair master decides for all the repair
    nodes involved.
    
    On top of this patch, we could do another step to make sure all nodes
    has the latest schema. But let's do it in a separate patch.
    
    Fixes: #4549
    Backports: 3.1
```

asias: When ignore_dropped_tables to false? The regular repair should set this to false. It is fine to fail the regular repair and report to user.

repair/repair.cc
asias: No. Repair loops over the ranges it asked to repair, for each range, it loops over the tables. So, if the a table is dropped and a new table is created with the same name, it is possible for some ranges the repair is performed with the old tables and some with the new tables. 

However, user specifies the table name not the uuid when user starts the repair. It is not wrong to repair the new tables for some of the ranges.  What can we do in this case? I think it is worse to fix the whole ranges with the old tables when the old table is dropped. 

Btw, for a given range, repair master and followers will always repair the same table because they repair the same schema id.

nyh: Do we have a dtest of what happens if during a repair the table is dropped and/or recreated during a repair?
Because I'm not seeing the big picture here (I guess the big picture isn't in this patch) - does a repair use the table name, or its UUID, to communicate with its inner loop, and also with repair slaves? It would be good to have a dtest to demonstrate that. If the situation is already correct in this respect (is it?) it would be shame for this patch to break it.

asias: We do have a dtest to test the table is dropped during repair: update_cluster_layout_tests.py:TestUpdateClusterLayout.simple_add_new_node_while_schema_changes_test. 

Repair master and followers already use the same schema version for a given table and range we repair. 

The problem here is what do we do if a table is dropped during repair.  Should we fail the whole bootstrap operation because the table is dropped and the repair failed (bootstrap uses repair now)?

repair/row_level.cc
slivne: @asias - we discussed this a bit - we do not want to extend rpc for 4.0 but would rather have the server side and return as a normal return and not as an exception - can this be done ?

asias: I think it is even more ugly than comparing the error msg in the exception. We need to patch all the verbs that might return the can't find a column family exception. Some of verbs return nothing but a future<>, there is no way to extend such rpc verb to send back the error message.  We have verbs that returns future<uint64_t>, no way to extend it in idl.

repair/repair.cc
avikivity: The new table can have an incompatible schema (different partition key and clustering key types). It can only end in disaster.

To avoid the race, convert the table names to uuids exactly once. If the drop/create happens before the lookup, we repair the new table. If it happens during or after the lookup, we won't repair (but that's fine, since it's a new table that did not exist when repair started).

avikivity: If a table is dropped during repair, we should ignore it. A dropped table is fully synchronized.

repair/repair.hh
avikivity: It's better not to fail the repair. Different people can be in charge of the schema and repair, they should not have to synchronize in order to manage the schema or repair. In fact either or both can be automation.

repair/row_level.cc
avikivity: How about creating new verbs with better returns and using them if they exist?

asias: This is pushing the job of the rpc framework to the users of rpc. Our rpc is pretty hard to use already, it is better not to make it even harder. It is not better returns, it is worse returns with an error code. We need to change all the verbs. We need to figure out if the new verb is supported. IMO, it does not deserve all the extra complexity introduced because we want to delay the fix at the rpc layer. 

Alternatively, we can just fail the repair for both normal repair and bootstrap operations if a table is dropped. We document do not change the schema during a node operation.

repair/repair.cc
asias: Yes, new table can have incompatible schema. But currently, it won't cause any problem because repair master and followers use the same schema id for a given range we repair. 

I do agree that it is more consistent to ask all ranges inside a repair request to use the same schema id too. I will add this change.

asias: Ok, so we ignore dropped tables and report the repair as success.

repair/row_level.cc
slivne: We recommend users not to have DDL changes but can't really force that, some systems have this done programaticlly on the user end. Adding node is a long running operation, there is no way to grantee their will not be DDL.

asias:
Changes in 6775c9e2bfd406bb520b9a2e3320dc25baf95988 (v3):
- Found a way to extend rpc response, using future<rpc::optional<new_response>>. Now the no_such_column_family error is send back with a error code. 
- Ignore dropped tables and report the repair as success for both normal repair and node operations repair.

message/messaging_service.cc
```diff
@@ -1193,14 +1193,14 @@ future<> messaging_service::send_repair_put_row_diff(msg_addr id, uint32_t repai
 }
 
 // Wrapper for REPAIR_ROW_LEVEL_START
-void messaging_service::register_repair_row_level_start(std::function<future<> (const rpc::client_info& cinfo, uint32_t repair_meta_id, sstring keyspace_name, sstring cf_name, dht::token_range range, row_level_diff_detect_algorithm algo, uint64_t max_row_buf_size, uint64_t seed, unsigned remote_shard, unsigned remote_shard_count, unsigned remote_ignore_msb, sstring remote_partitioner_name, table_schema_version schema_version, rpc::optional<streaming::stream_reason> reason)>&& func) {
+void messaging_service::register_repair_row_level_start(std::function<future<repair_row_level_start_response> (const rpc::client_info& cinfo, uint32_t repair_meta_id, sstring keyspace_name, sstring cf_name, dht::token_range range, row_level_diff_detect_algorithm algo, uint64_t max_row_buf_size, uint64_t seed, unsigned remote_shard, unsigned remote_shard_count, unsigned remote_ignore_msb, sstring remote_partitioner_name, table_schema_version schema_version, rpc::optional<streaming::stream_reason> reason)>&& func) {
```
avikivity, line 1196: Did you check that cross-version repair works with this change?

asias: Yes, I did. It worked.

slivne:
@avikivity asias has responded to your question he ran a test - can we please merge this

slivne:
@avikivity ping

PR #6209, opened 2020-04-16
UUID: Stop caching clock_seq_and_node

### LINKED ISSUE #6208: currentTimeUUID creates duplicates when called at the same point in time
The issue was reported on StackOverflow [here](https://stackoverflow.com/questions/61223391/how-to-guarantee-monotonically-increasing-timeuuid-when-selecting-from-scylla/61243270)

Internally currentTimeUUID calls [get_time_UUID](https://github.com/scylladb/scylla/blob/d13492485f1772fc2350b09595ca1fc57e783f8a/utils/UUID_gen.hh#L94) which creates UUID using current time and a [clock_seq_and_node](https://github.com/scylladb/scylla/blob/d13492485f1772fc2350b09595ca1fc57e783f8a/utils/UUID_gen.hh#L94) constant [generated once per shard](https://github.com/scylladb/scylla/blob/67a07e9cbc498bbac21a9bb545993232f5a164cb/utils/UUID_gen.cc#L99).

```
    static UUID get_time_UUID()
    {
        auto uuid = UUID(instance->create_time_safe(), clock_seq_and_node);
        assert(uuid.is_timestamp());
        return uuid;
    }
```

```
const thread_local int64_t UUID_gen::clock_seq_and_node = make_clock_seq_and_node();
```

haaawk:
Previously, clock_seq_and_node was generated once at startup and was used for
generation of TimeUUIDs. This is problematic because as a result creating
TimeUUID at the same time on the same node always creates exactly the same
value. It's bad because the whole purpose of using TimeUUID is to create
unique identifier for a timestamp. Not a duplicate. Otherwise a timestamp
alone could be used as well.

This patch drops the cache of clock_seq_and_node and calls
make_clock_seq_and_node() every time TimeUUID is generated. This creates new
unique clock_seq_and_node for each generation and reduces the chance for
collision.

Fixes #6208

Tests: unit(dev)

### DIFF
--- a/utils/UUID_gen.cc
+++ b/utils/UUID_gen.cc
@@ -57,14 +57,15 @@ static int64_t make_node()
 }
 
 
-static int64_t make_clock_seq_and_node()
+int64_t make_clock_seq_and_node()
 {
+    thread_local std::mt19937_64 engine(std::random_device().operator()());
+    thread_local std::uniform_int_distribution<int> dist;
     // The original Java code did this, shuffling the number of millis
     // since the epoch, and taking 14 bits of it. We don't do exactly
     // the same, but the idea is the same.
     //long clock = new Random(System.currentTimeMillis()).nextLong();
-    unsigned int seed = std::chrono::system_clock::now().time_since_epoch().count();
-    int clock = rand_r(&seed);
+    int clock = dist(engine);
 
     long lsb = 0;
     lsb |= 0x8000000000000000L;                 // variant (2 bits)
@@ -96,7 +97,6 @@ UUID UUID_gen::get_name_UUID(const unsigned char *s, size_t len) {
     return get_UUID(digest);
 }
 
-const thread_local int64_t UUID_gen::clock_seq_and_node = make_clock_seq_and_node();
 thread_local const std::unique_ptr<UUID_gen> UUID_gen::instance (new UUID_gen());
 
 
--- a/utils/UUID_gen.hh
+++ b/utils/UUID_gen.hh
@@ -50,6 +50,8 @@
 
 namespace utils {
 
+int64_t make_clock_seq_and_node();
+
 /**
  * The goods are here: www.ietf.org/rfc/rfc4122.txt.
  */
@@ -58,7 +60,6 @@ class UUID_gen
 private:
     // A grand day! millis at 00:00:00.000 15 Oct 1582.
     static constexpr int64_t START_EPOCH = -12219292800000L;
-    static thread_local const int64_t clock_seq_and_node;
 
     /*
      * The min and max possible lsb for a UUID.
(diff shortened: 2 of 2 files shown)


haaawk:
> LGTM

Thanks, @bhalevy!

tgrabiec:
Do you have a unit test reproducing this?

Seems to me that the bug is not possible because `create_time_safe()` ensures that the timestamp is strictly monotonic.

bhalevy:
> Do you have a unit test reproducing this?
> 
> Seems to me that the bug is not possible because `create_time_safe()` ensures that the timestamp is strictly monotonic.

@tgrabiec shouldn't `last_nanos` be either thread_local or atomic? (maybe that's the actual bug)

tgrabiec:
On Thu, Apr 16, 2020 at 5:09 PM Benny Halevy <notifications@github.com>
wrote:

> Do you have a unit test reproducing this?
>
> Seems to me that the bug is not possible because create_time_safe()
> ensures that the timestamp is strictly monotonic.
>
> @tgrabiec <https://github.com/tgrabiec> shouldn't last_nanos be either
> thread_local or atomic? (maybe that's the actual bug)
>
The instance of UUID_gen through which last_nanos is accessed is
thread-local.

>

bhalevy:
> The instance of UUID_gen through which last_nanos is accessed is thread-local.

I see.  (`UUID_gen::instance`)

bhalevy:
https://en.cppreference.com/w/cpp/chrono/system_clock says that system_clock may not be monotonic, I think this could happen with ntp with coarse adjustments.

haaawk:
> Do you have a unit test reproducing this?
> 
> Seems to me that the bug is not possible because `create_time_safe()` ensures that the timestamp is strictly monotonic.

I don't. Maybe I read the code wrong. I will try to have another look.

kostja:
* Piotr Jastrzębski <notifications@github.com> [20/04/16 09:56]:
> Previously, clock_seq_and_node was generated once at startup and was used for
> generation of TimeUUIDs. This is problematic because as a result creating
> TimeUUID at the same time on the same node always creates exactly the same
> value. It&#39;s bad because the whole purpose of using TimeUUID is to create
> unique identifier for a timestamp. Not a duplicate. Otherwise a timestamp
> alone could be used as well.
> 
> This patch drops the cache of clock_seq_and_node and calls
> make_clock_seq_and_node() every time TimeUUID is generated. This creates new
> unique clock_seq_and_node for each generation and reduces the chance for
> collision.

What's the performance impact?


-- 
https://scylladb.com

haaawk:
I'm not convineced either @nyh :). I'm not sure it should be merged at the moment.

haaawk:
This is probably a wrong fix.

PR #6210, opened 2020-04-16
dht: document incremental partition_range and token_range sharders

avikivity:


avikivity:
@haaawk please review. This is not a subliminal message that you should switch, just a response to your complaint about lack of documentation.

dht/sharder.hh
```diff
@@ -28,28 +28,58 @@
 
 namespace dht {
 
+
+// Utilities for sharding ring partition_range:s
```
haaawk, line 32: What's this ':'? Is it some sort of convention?

dht/sharder.hh
```diff
@@ -28,28 +28,58 @@
...
+
+// A ring_position range's data is divided into sub-ranges, where each sub-range's data
+// is owned by a single range. Note that multiple non-overlapping sub-ranges may map to a
```
haaawk, line 35: 'owned by a single range'? Did you mean 'owned by a single shard'?

dht/sharder.hh
```diff
@@ -28,28 +28,58 @@
...
+// single range, and some shards may not receive any sub-range.
```
haaawk, line 36: 'single range'? Did you mean 'single shard'?

dht/sharder.hh
avikivity: I learned it from @elcallio so I don't know the origin story. It's useful to keep an identifer seprated from the plural marker, so it's clear the s isn't part of it, in case you want to search for it.

dht/sharder.hh
avikivity: Yes, will fix

dht/sharder.hh
avikivity: yes

dht/sharder.hh
```diff
@@ -28,28 +28,58 @@
...
+//
+// This module provides utilities for determining the sub-ranges to shard mapping. The utilities
+// generate optimal mappings: you will each range that you get is the largest possible, so you
```
haaawk, line 39: 'you will each range'? Missing a verb, I think. Probably 'you will' should just be removed.

dht/sharder.hh
```diff
@@ -28,28 +28,58 @@
...
+// get the minimum number of ranges possible. You can get many ranges, so operate on them
+// one (or a few) at a time, rather than accumulating them.
+
+// A mapping between a partition_range and a shard. All positions within `ring_range` are
+// owned by `shard`.
```
haaawk, line 44: It might be worth repeating here that ring_range is maximal which means start(ring_range) - 1 and end(ring_range) + 1 don't belong to `shard`.

dht/sharder.hh
avikivity: yes

dht/sharder.hh
avikivity: ok. Too much repetition is annoying, too little repetition creates bugs.

dht/sharder.hh
```diff
@@ -28,28 +28,58 @@
...
             : ring_position_range_and_shard(std::move(rpras)), element(element) {
     }
     unsigned element;
 };
 
+// Incrementally divides several non-overlapping `partition_range`:s into sub-ranges wholly owned by
+// a single shard.
+//
+// Similar to ring_position_range_sharder, but instead of stopping when the input range is exhauseted,
+// moves on to the next input range (input ranges are supplied ina vector).
```
haaawk, line 78: s/ina/in a/ ?

avikivity: yes

avikivity:
Updated:
- various thinkos: s/single range/single shard/
- typos
- repeated that shard-ranges are maximal, so that their count is minimal
- added another use case (the only one in-tree) - sstable-shard mapping calculation

dht/sharder.hh
haaawk: This is still 'single range' instead of 'single shard'

dht/sharder.hh
```diff
@@ -28,28 +28,67 @@
...
+//
+// This module provides utilities for determining the sub-ranges to shard mapping. The utilities
+// generate optimal mappings: each range that you get is the largest possible, so you
+// get the minimum number of ranges possible. You can get many ranges, so operate on them
+// one (or a few) at a time, rather than accumulating them.
+
+// A mapping between a partition_range and a shard. All positions within `ring_range` are
+// owned by `shard`.
+//
+// The classes that return ring_position_range_and_shard make them as large as possible,
```
haaawk, line 46: It would be more clear to say 'make ring_range as large as possible' instead of 'make them as large as possible'.

haaawk: I believe that it would be valuable to explain what it means for a range to be as large as possible. It means that enlarging it at any end will mean the range is no longer owned by a single shard. I know this can be deduced but let's not put too much pressure on the reader ;)

avikivity: Done.

avikivity:
Updated:
 - made it clear what it means to have a maximally large range

dht/sharder.hh
haaawk: What about this @avikivity

avikivity: Sorry, I thought I fixed it, I'll send a follow up.

avikivity: Ah, it wasn't merged. Fixing and pushing.

avikivity:
Updated:
 - rebased
 - a missed s/single range/single shard/ from the previous iteration

dht/sharder.hh
```diff
@@ -77,6 +132,7 @@ class selective_token_range_sharder {
     dht::token _start_token;
     std::optional<range_bound<dht::token>> _start_boundary;
 public:
+    // Initializes the selective_token_range_sharder with a token range and shard_id of interest.
```
haaawk, line 135: The comment just repeats the code so it would probably be better to just drop it.

dht/sharder.hh
```diff
@@ -86,6 +142,9 @@ public:
             , _start_boundary(_sharder.shard_of(_start_token) == shard ?
                 _range.start() : range_bound<dht::token>(_sharder.token_for_next_shard(_start_token, shard))) {
     }
+    // Returns the next token_range that is both wholly contained within the input range and also
+    // wholly owned by the input shard_id. When the input range is exhausted, std::nullopt is returned.
+    // Note if the range does not intersect the shard at all, it may be returned immediately.
```
haaawk, line 147: Probably cleaner to say "std::nullopt may be returned immediately" instead of "it can be returned immediately"

dht/sharder.hh
```diff
@@ -63,11 +105,24 @@ private:
...
+    //
+    // The result is augmented with an `element` field which indicates the index from the input vector
+    // that the result belongs to.
+    //
+    // Results are returned sorted by index within the vector first, then within each vector item
     std::optional<ring_position_range_and_shard_and_element> next(const schema& s);
 };
 
+// Incrementally divides a `partition_range` into sub-ranges wholly owned by a single shard.
+// Unlike ring_position_range_sharder, it only returns result for a shard number provided the caller.
```
haaawk, line 125: s/provided the caller/provided by the caller/

nyh:
@avikivity @haaawk this is an antique PR, is it still relevant? What was the conclusion of the review?

avikivity:
It's still relevant, I forgot to apply the last review comment.

avikivity:
Updated:
 - rebased
 - document maximal range property for ring_position_range_and_shard_and_element
 - clarified that selective_token_range_sharder can return nullopt immediately if the range does not intersect the range
 - s/provided the caller/provided by the caller/

haaawk (approved):

dht/sharder.hh
```diff
@@ -28,28 +28,76 @@
...
 };
 
+// A mapping between a partition_range and a shard (like ring_position_range_and_shard) extended
+// by having a reference to input range index. See ring_position_range_vector_sharder for use.
+//
+// The classes that return ring_position_range_and_shard_and_element make `ring_range` as large as
+// possible (maximizing the number of tokens), so the total number of such ranges is minimized.
+// Successive ranges therefore always have a different `shard` than the previous return.
+// (classes that return ring_position_range_and_shard_and_element can have the same `shard`
+// in successive returns, if `element` is different).
```
nyh, line 78: I think it is not clear enough what "element" is. Is it an index into something? into what? This is not explained. And if it is an index, why not call it "index"?

dht/sharder.hh
```diff
@@ -28,28 +28,76 @@
...
     ring_position_range_and_shard_and_element(ring_position_range_and_shard&& rpras, unsigned element)
             : ring_position_range_and_shard(std::move(rpras)), element(element) {
     }
     unsigned element;
 };
 
+// Incrementally divides several non-overlapping `partition_range`:s into sub-ranges wholly owned by
+// a single shard.
+//
+// Similar to ring_position_range_sharder, but instead of stopping when the input range is exhauseted,
```
nyh, line 89: exhausted

dht/sharder.hh
```diff
@@ -28,28 +28,76 @@
...
+// moves on to the next input range (input ranges are supplied in a vector).
+//
+// This has two use cases:
+
+// 1. vnodes. A vnode cannot be described by a single range, since
+//    one vnode wraps around from the largest token back to the smallest token. Hence it must be
```
nyh, line 95: I would say "may wrap" instead of "wraps". Obviously not all token ranges wrap around.

nyh (approved):
Looks good. I had some comments (and there is one typo) but it can all be done later as a followup - it's already better than the (no) documentation we had.

PR #6215, opened 2020-04-16
cdc: fix the "NoHostAvailable" client error when CL is not met

### LINKED ISSUE #5746: Misleading error from CDC preimage when RF > # of nodes
Operations on table with `'preimage':'true'` throw `NoHostAvailable` when `replication_factor` is greater than number of nodes. E.g. with a 1-node cluster:
```
cqlsh> create keyspace ks with replication = {'class':'SimpleStrategy', 'replication_factor': 3};
cqlsh> use ks;
cqlsh:ks> create table t1 (a int primary key, b int) with cdc = {'enabled':true, 'preimage':'true'};
cqlsh:ks> insert into t1 (a,b) values (1,1);
NoHostAvailable: 
```

jul-stas:
CL of LOCAL_QUORUM used to be hardcoded into CDC preimage query
and led to an error every time the number of replicas was lower than CL
would require. The solution here is to link the CLs of writes
to base table with the CLs of CDC reads, so the client will get
the (limited) control over the consistency of preimage SELECTs
(instead of constant misleading errors).

The algorithm is as follows:
1. If write that caused CDC activity was done with CL = ANY,
  then do preimage read with CL = ONE.
2. If write that caused CDC activity was done with CL = ALL,
  then do preimage read with CL = QUORUM.
3. SERIAL and LOCAL_SERIAL writes cause preimage read with QUORUM
  and LOCAL_QUORUM, respectively.
4. In other cases do preimage read with the same CL as base write.

To further mitigate the incomprehensible error being sent to client,
I wrapped the preimage's SELECT query in try-catch and
intercept the `unavailable_exception`, which was manifesting as
`NoHostAvailable` in Python and Java drivers. Now client gets a new
error code and a message specific to the issue of CL not being met
by the preimage query.

Fixes #5746

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -1292,6 +1292,7 @@ public:
 
         const auto select_cl = adjust_cl(write_cl);
 
+      try {
         return _ctx._proxy.query(_schema, std::move(command), std::move(partition_ranges), select_cl, service::storage_proxy::coordinator_query_options(default_timeout(), empty_service_permit(), client_state)).then(
                 [s = _schema, partition_slice = std::move(partition_slice), selection = std::move(selection)] (service::storage_proxy::coordinator_query_result qr) -> lw_shared_ptr<cql3::untyped_result_set> {
                     cql3::selection::result_set_builder builder(*selection, gc_clock::now(), cql_serialization_format::latest());
@@ -1302,6 +1303,13 @@ public:
                     }
                     return make_lw_shared<cql3::untyped_result_set>(*result_set);
         });
+      } catch (exceptions::unavailable_exception& e) {
+        // `query` can throw `unavailable_exception`, which is seen by clients as ~ "NoHostAvailable". 
+        // So, we'll translate it to a `read_failure_exception` with custom message.
+        cdc_log.debug("Preimage: translating a (read) `unavailable_exception` to `request_execution_exception` - {}", e);
+        throw exceptions::read_failure_exception("CDC preimage query could not achieve the CL.",
+                e.consistency, e.alive, 0, e.required, false);
+      }
     }
 
     /** For preimage query use the same CL as for base write, except for CLs ANY and ALL. */
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -239,7 +239,8 @@ public:
     future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>> augment_mutation_call(
         lowres_clock::time_point timeout,
         std::vector<mutation>&& mutations,
-        tracing::trace_state_ptr tr_state
+        tracing::trace_state_ptr tr_state,
+        db::consistency_level write_cl
     );
 
     template<typename Iter>
@@ -1208,7 +1209,7 @@ public:
 
     future<lw_shared_ptr<cql3::untyped_result_set>> pre_image_select(
             service::client_state& client_state,
-            db::consistency_level cl,
+            db::consistency_level write_cl,
             const mutation& m)
     {
         auto& p = m.partition();
@@ -1289,7 +1290,9 @@ public:
         auto partition_slice = query::partition_slice(std::move(bounds), std::move(static_columns), std::move(regular_columns), std::move(opts));
         auto command = ::make_lw_shared<query::read_command>(_schema->id(), _schema->version(), partition_slice, row_limit);
 
-        return _ctx._proxy.query(_schema, std::move(command), std::move(partition_ranges), cl, service::storage_proxy::coordinator_query_options(default_timeout(), empty_service_permit(), client_state)).then(
+        const auto select_cl = adjust_cl(write_cl);
+
+        return _ctx._proxy.query(_schema, std::move(command), std::move(partition_ranges), select_cl, service::storage_proxy::coordinator_query_options(default_timeout(), empty_service_permit(), client_state)).then(
                 [s = _schema, partition_slice = std::move(partition_slice), selection = std::move(selection)] (service::storage_proxy::coordinator_query_result qr) -> lw_shared_ptr<cql3::untyped_result_set> {
                     cql3::selection::result_set_builder builder(*selection, gc_clock::now(), cql_serialization_format::latest());
                     query::result_view::consume(*qr.query_result, partition_slice, cql3::selection::result_set_builder::visitor(builder, *s, *selection));
@@ -1299,6 +1302,18 @@ public:
                     }
                     return make_lw_shared<cql3::untyped_result_set>(*result_set);
         });
+    }
+
+    /** For preimage query use the same CL as for base write, except for CLs ANY and ALL. */
+    static db::consistency_level adjust_cl(db::consistency_level write_cl) {
+        if (write_cl == db::consistency_level::ANY) {
+            return db::consistency_level::ONE;
+        } else if (write_cl == db::consistency_level::ALL || write_cl == db::consistency_level::SERIAL) {
+	        return db::consistency_level::QUORUM;
+        } else if (write_cl == db::consistency_level::LOCAL_SERIAL) {
+	        return db::consistency_level::LOCAL_QUORUM;
+        }
+        return write_cl;
     }
 };
 
@@ -1314,7 +1329,7 @@ transform_mutations(std::vector<mutation>& muts, decltype(muts.size()) batch_siz
 } // namespace cdc
 
 future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>>
-cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, std::vector<mutation>&& mutations, tracing::trace_state_ptr tr_state) {
+cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, std::vector<mutation>&& mutations, tracing::trace_state_ptr tr_state, db::consistency_level write_cl) {
     // we do all this because in the case of batches, we can have mixed schemas.
     auto e = mutations.end();
     auto i = std::find_if(mutations.begin(), e, [](const mutation& m) {
@@ -1329,8 +1344,8 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, 
     mutations.reserve(2 * mutations.size());
 
     return do_with(std::move(mutations), service::query_state(service::client_state::for_internal_calls(), empty_service_permit()), operation_details{},
-            [this, timeout, i, tr_state = std::move(tr_state)] (std::vector<mutation>& mutations, service::query_state& qs, operation_details& details) {
-        return transform_mutations(mutations, 1, [this, &mutations, timeout, &qs, tr_state = tr_state, &details] (int idx) mutable {
+            [this, timeout, i, tr_state = std::move(tr_state), write_cl] (std::vector<mutation>& mutations, service::query_state& qs, operation_details& details) {
+        return transform_mutations(mutations, 1, [this, &mutations, timeout, &qs, tr_state = tr_state, &details, write_cl] (int idx) mutable {
             auto& m = mutations[idx];
             auto s = m.schema();
 
@@ -1346,7 +1361,7 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout, 
                 // iff a batch contains several modifications to the same table. Otoh, batch is rare(?)
                 // so this is premature.
                 tracing::trace(tr_state, "CDC: Selecting preimage for {}", m.decorated_key());
-                f = trans.pre_image_select(qs.get_client_state(), db::consistency_level::LOCAL_QUORUM, m).then_wrapped([this] (future<lw_shared_ptr<cql3::untyped_result_set>> f) {
+                f = trans.pre_image_select(qs.get_client_state(), write_cl, m).then_wrapped([this] (future<lw_shared_ptr<cql3::untyped_result_set>> f) {
                     auto& cdc_stats = _ctxt._proxy.get_cdc_stats();
                     cdc_stats.counters_total.preimage_selects++;
                     if (f.failed()) {
@@ -1403,6 +1418,6 @@ bool cdc::cdc_service::needs_cdc_augmentation(const std::vector<mutation>& mutat
 }
 
 future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>>
-cdc::cdc_service::augment_mutation_call(lowres_clock::time_point timeout, std::vector<mutation>&& mutations, tracing::trace_state_ptr tr_state) {
-    return _impl->augment_mutation_call(timeout, std::move(mutations), std::move(tr_state));
+cdc::cdc_service::augment_mutation_call(lowres_clock::time_point timeout, std::vector<mutation>&& mutations, tracing::trace_state_ptr tr_state, db::consistency_level write_cl) {
+    return _impl->augment_mutation_call(timeout, std::move(mutations), std::move(tr_state), write_cl);
 }
--- a/exceptions/exceptions.hh
+++ b/exceptions/exceptions.hh
@@ -197,6 +197,14 @@ protected:
         , failures{failures_}
         , block_for{block_for_}
     {}
+
+    request_failure_exception(exception_code code, const sstring& msg, db::consistency_level consistency_, int32_t received_, int32_t failures_, int32_t block_for_) noexcept
+        : cassandra_exception{code, msg}
+        , consistency{consistency_}
+        , received{received_}
+        , failures{failures_}
+        , block_for{block_for_}
+    {}
 };
 
 struct mutation_write_failure_exception : public request_failure_exception {
@@ -212,6 +220,11 @@ struct read_failure_exception : public request_failure_exception {
 
     read_failure_exception(const sstring& ks, const sstring& cf, db::consistency_level consistency_, int32_t received_, int32_t failures_, int32_t block_for_, bool data_present_) noexcept
         : request_failure_exception{exception_code::READ_FAILURE, ks, cf, consistency_, received_, failures_, block_for_}
+        , data_present{data_present_}
+    { }
+
+    read_failure_exception(const sstring& msg, db::consistency_level consistency_, int32_t received_, int32_t failures_, int32_t block_for_, bool data_present_) noexcept
+        : request_failure_exception{exception_code::READ_FAILURE, msg, consistency_, received_, failures_, block_for_}
         , data_present{data_present_}
     { }
 };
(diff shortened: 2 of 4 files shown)


cdc/log.cc
```diff
@@ -1289,7 +1290,12 @@ class transformer final {
         auto partition_slice = query::partition_slice(std::move(bounds), std::move(static_columns), std::move(regular_columns), std::move(opts));
         auto command = ::make_lw_shared<query::read_command>(_schema->id(), _schema->version(), partition_slice, row_limit);
 
-        return _ctx._proxy.query(_schema, std::move(command), std::move(partition_ranges), cl, service::storage_proxy::coordinator_query_options(default_timeout(), empty_service_permit(), client_state)).then(
+        const auto select_cl = write_cl >= db::consistency_level::QUORUM
+                ? db::consistency_level::QUORUM
+                : db::consistency_level::ONE;
+
+      try {
```
jul-stas, line 1297: This is a gift for you, reviewers, unless we don't do half-stepbacks here..?

cdc/log.cc
```diff
@@ -239,7 +239,8 @@ class cdc::cdc_service::impl : service::migration_listener::empty_listener {
     future<std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>> augment_mutation_call(
```
haaawk, line 239: Please create a proper commit message by paraphrasing the mailing list quoted.

cdc/log.cc
```diff
@@ -1304,6 +1305,15 @@ class transformer final {
                     }
                     return make_lw_shared<cql3::untyped_result_set>(*result_set);
         });
+      } catch (exceptions::unavailable_exception& e) {
```
haaawk, line 1308: Shouldn't it be a handle_exception on the future returned from _ctx._proxy.query. Otherwise, it is possible for this method to return a non-ready future and the exception you want to catch will be thrown somewhere else, no? Or is the exception thrown from _ctx._proxy.query before creating any future?

cdc/log.cc
```diff
@@ -1304,6 +1305,15 @@ class transformer final {
...
+          // `query` can throw `unavailable_exception`, which triggers an error code often interpreted by clients
+          // as ~ "NoHostAvailable". We'll translate it to a `request_execution_exception` and futurize it.
+          cdc_log.debug("Preimage: translating a (read) `unavailable_exception` to `request_execution_exception` - ", e);
+          exceptions::request_execution_exception new_exc(exceptions::exception_code::CDC_PREIMAGE_READ_FAILURE,
+                  format("CDC could not achieve the CL of {}. Required {} responses, got {}.",
+                          e.consistency, e.required, e.alive));
+          return make_exception_future<lw_shared_ptr<cql3::untyped_result_set>>(std::move(new_exc));
```
haaawk, line 1315: Is the futurization really necessary? Wouldn't just throwing the exception have exactly the same effect?

jul-stas: It would be the same (checked experimentally), but I assumed that once futurized it's more "seastary".

haaawk: I think it's better to just throw. It would be simpler and more concise.

cdc/log.cc
jul-stas: I thought about it, but this `throw` is buried quite deep in non-deferred code (`storage_proxy::get_read_executor`) so I assumed that throwing is a part of its contract and better not be futurized. I also experimentally tried `handle_exception` here, but it didn't catch anything.

cdc/log.cc
```diff
@@ -1304,6 +1305,15 @@ class transformer final {
...
                     return make_lw_shared<cql3::untyped_result_set>(*result_set);
         });
+      } catch (exceptions::unavailable_exception& e) {
+          // `query` can throw `unavailable_exception`, which triggers an error code often interpreted by
+          // clients as ~ "NoHostAvailable". We'll translate it to a `request_execution_exception`.
+          cdc_log.debug("Preimage: translating a (read) `unavailable_exception` to `request_execution_exception` - ", e);
+          exceptions::request_execution_exception new_exc(exceptions::exception_code::CDC_PREIMAGE_READ_FAILURE,
+                  format("CDC could not achieve the CL of {}. Required {} responses, got {}.",
+                          e.consistency, e.required, e.alive));
+          throw new_exc;
```
haaawk, line 1315: why not just throw without giving it a name?

cdc/log.cc
```diff
@@ -1289,7 +1290,11 @@ class transformer final {
         auto partition_slice = query::partition_slice(std::move(bounds), std::move(static_columns), std::move(regular_columns), std::move(opts));
         auto command = ::make_lw_shared<query::read_command>(_schema->id(), _schema->version(), partition_slice, row_limit);
 
-        return _ctx._proxy.query(_schema, std::move(command), std::move(partition_ranges), cl, service::storage_proxy::coordinator_query_options(default_timeout(), empty_service_permit(), client_state)).then(
+        const auto select_cl = write_cl >= db::consistency_level::QUORUM
+                ? db::consistency_level::QUORUM
+                : db::consistency_level::ONE;
```
piodul, line 1295: We can't use `>=` for comparing the strength of consistency levels. For example, it won't work for LOCAL_ONE - it has the highest enum value, but should probably be considered weaker than quorum. We should compare it in some other way.

Maybe you could find something useful in `db/consistency_level.hh`?

cdc/log.cc
```diff
@@ -1289,7 +1290,11 @@ class transformer final {
         auto partition_slice = query::partition_slice(std::move(bounds), std::move(static_columns), std::move(regular_columns), std::move(opts));
         auto command = ::make_lw_shared<query::read_command>(_schema->id(), _schema->version(), partition_slice, row_limit);
 
-        return _ctx._proxy.query(_schema, std::move(command), std::move(partition_ranges), cl, service::storage_proxy::coordinator_query_options(default_timeout(), empty_service_permit(), client_state)).then(
+        const auto select_cl = write_cl >= db::consistency_level::QUORUM
```
haaawk, line 1293: @piodul found a bug here. Apparently we can't compare consistency_level with < and >. The order of values in the enum is not semantically correct:
```
enum class consistency_level {
    ANY, MIN_VALUE = ANY,
    ONE,
    TWO,
    THREE,
    QUORUM,
    ALL,
    LOCAL_QUORUM,
    EACH_QUORUM,
    SERIAL,
    LOCAL_SERIAL,
    LOCAL_ONE, MAX_VALUE = LOCAL_ONE
};
```

LOCAL_ONE is after QUORUM

haaawk: Also I the condition should look like this:
`std::min(QUORUM, std::max(ONE, write_cl))`
with a semantically correct comparison. This means if write_cl is LOCAL_ONE, TWO or THREE, we should select with LOCAL_ONE, TWO or THREE as well

haaawk: I think we will have to implement some sort of comparator for CL but the problem is the order is not total and depends on RF. For example with RF=1 order is:
ANY, ONE, LOCAL_ONE, TWO, QUORUM, SERIAL, LOCAL_QUORUM, LOCAL_SERIAL, EACH_QUORUM, THREE, ALL
while for RF=6 the order is:
ANY, ONE, LOCAL_ONE, TWO, THREE, QUORUM, SERIAL, LOCAL_QUORUM, LOCAL_SERIAL, EACH_QUORUM, ALL

so the comparator will have to take RF into account.

haaawk: @kostja Could you please confirm the placement of SERIAL and LOCAL_SERIAL in the order?

haaawk: I'm also not sure whether QUORUM < LOCAL_QUORUM or the other way around. QUORUM will require more ACKs but it will probably be easier to achieve QUORUM than LOCAL_QUORUM because the number of replicas for QUORUM is bigger than number of replicas for LOCAL_QUORUM.

piodul: If we define strength of the consistency level as the number of replicas needed for read/write, then this function might come in handy:

```
size_t db::block_for(keyspace& ks, consistency_level cl);
```

It's defined in `db/consistency_level.cc`. It takes RF of the keyspace into consideration.

haaawk: I saw it but it won't be useful because ONE, LOCAL_ONE and ANY are equal there. It is useful to copy some stuff from it like `ks.get_replication_strategy().get_replication_factor()` and `quorum_for(ks)` though

jul-stas: Nice catch, I missed LOCAL_ONE and deliberately squeezed TWO and THREE into ONE (maybe wrongly, but they are flaky with their guarantees of all or nothing, depending on RF). I'll update.

Speaking of (LOCAL_)SERIAL - I need guidance here.

kostja: SERIAL is QUORUM, LOCAL_SERIAL is LOCAL_QUORUM. Treat them this way please.

kostja: QUORUM and LOCAL_QUORUM are not comparable at all. Imagine you have a cluster of 6 nodes, 3 nodes in dc1 and 3 in dc2. 
If 2 nodes out of 3 in DC1 are down, QUORUM will pass, LOCAL_QUORUM in DC1 will not.

Imagine you have 3 nodes in DC1 down. LOCAL_QUORUM ind DC2 will pass, QUORUM will not.

haaawk: Maybe we can go with a simple solution ANY,ONE, LOCAL_ONE, TWO, THREE -> ONE and other CL -> QUORUM?

haaawk: Because creating comparator clearly makes no sense.

cdc/log.cc
```diff
@@ -1299,6 +1303,35 @@ class transformer final {
...
+          // `query` can throw `unavailable_exception`, which triggers an error code often interpreted by
+          // clients as ~ "NoHostAvailable". We'll translate it to a `request_execution_exception`.
+          cdc_log.debug("Preimage: translating a (read) `unavailable_exception` to `request_execution_exception` - ", e);
+          throw exceptions::request_execution_exception(exceptions::exception_code::CDC_PREIMAGE_READ_FAILURE,
+                  format("CDC could not achieve the CL of {}. Required {} responses, got {}.",
+                          e.consistency, e.required, e.alive));
+      }
+    }
+
+    static db::consistency_level adjust_cl(db::consistency_level write_cl) {
```
jul-stas, line 1316: Here's a new proposal:

- (ANY, ONE) -> ONE, just because we cannot go lower (but see my remark at the end). The client wants flakily better consistency (TWO, THREE)? OK, boost the preimage in the same way.
- Hitting all other cases with one hammer needs trade-offs. The only question is how much clients expect from the preimage (I don't know the answer). I trust you that QUORUM is the Goodenough's Solution.

Any problems with this solution? For sure, when NO replicas are present, write with CL=ANY can succeed by being captured by hinted handoff. Preimage query will fail - maybe it should insert NULLs instead?

haaawk: I think (ANY, ONE) -> ONE and others -> QUORUM is good enough for now.

exceptions/exceptions.hh
```diff
@@ -67,6 +67,7 @@ enum class exception_code : int32_t {
     FUNCTION_FAILURE= 0x1400,
     WRITE_FAILURE   = 0x1500,
     CDC_WRITE_FAILURE = 0x1600,
+    CDC_PREIMAGE_READ_FAILURE = 0x1601,
```
piodul, line 70: Are we allowed to allocate new exception codes like that? This error code is not recognized by Cassandra. It is dangerously close to CDC_WRITE_FAILURE, and I can imagine a possibility where new CDC-related codes could be defined right after CDC_WRITE_FAILURE (`0x1601`, `0x1602` etc.).

Avi suggested to allocate error codes starting from `0x1aff` backwards, see discussion here: https://github.com/scylladb/scylla/pull/5537#discussion_r363084064

cdc/log.cc
jul-stas: As we discussed, I force-pushed a yet another solution:
- ANY -> ONE
- ALL -> QUORUM
- (LOCAL_)SERIAL -> (LOCAL_)QUORUM
- for all other cases: WRITE_CL -> PREIMAGE_CL

exceptions/exceptions.hh
jul-stas: OK, I picked `0x1aef` for `CDC_PREIMAGE_READ_FAILURE`, keeping 15 free slots for more overflow-related codes.
Python driver seems to handle non-standard codes nicely, but I'll get an opinion from someone with more "strategic" knowledge.

exceptions/exceptions.hh
```diff
@@ -71,6 +71,7 @@ enum class exception_code : int32_t {
     // Scylla-specific codes
     // Allocated backwards from 0x1aff-0x1a01 to minimize the chance of collision with Cassandra.
     OVERFLOW_ERROR  = 0x1aff,
+    CDC_PREIMAGE_READ_FAILURE = 0x1aef,
```
gleb-cloudius, line 74: Extending a binary protocol like that is a slippery slope. I see that we alraedy did it even before that patch, but what is it good for if we do not write our own drivers? How do we know that all the drivers will handle it gracefully? At the very least we should not use them unless we negotiated that our own driver is in use.

kostja: Interestingly we need a similar extension for LWT, Gleb :)

kostja: (and I agree that extending the protocol on the server side only is useless)

gleb-cloudius: We do have an ability to negotiate features with our drivers. It is crazy how we are adding those two codes here without any regards with ABI compatibility.  Even Cassandra when they introduce new error codes only use them when a proper binary version of a protocol is negotiated. How are _we_ going to know which of our version supports what errors?

jul-stas: @gleb-cloudius @kostja So, maybe I'll return `READ_FAILURE` or `UNAVAILABLE`?
Not that reusing codes can also lead to a mess (I mean, a less visible mess but potentially far-reaching).

Anyway, please, let's state a definitive opinion here. This is a simple patch and doesn't deserve to hang for that long.

haaawk: Please return READ_FALIURE and let's get this patch merged.

gleb-cloudius: You can report what exactly failed in the message supplied to the error.

jul-stas: > Please return READ_FALIURE and let's get this patch merged.

Ooops, I cannot. Both `READ_FAILURE` and `UNAVAILABLE` impose constraints on the error message. Just checked: `request_execution_exception` with any of these crashes the python client.

My bad, the problem is still open.

gleb-cloudius: What do you mean? How returning READ_FAILURE crashes python driver?

jul-stas:
```
  File "/home/julex/scylla-tools-java/bin/cqlsh.py", line 1053, in perform_simple_statement
    result = future.result()
  File "/home/julex/scylla-tools-java/bin/../lib/cassandra-driver-internal-only-3.11.0-bb96859b.zip/cassandra-driver-3.11.0-bb96859b/cassandra/cluster.py", line 3925, in result
    raise self._final_exception
error: unpack requires a string argument of length 2
```
Same thing with `UNAVAILABLE`. When I throw plain `cassandra_exception` with those codes effect is identical. I didn't investigate what exactly happens on clientside, but it seems that these codes imply some structure in error response.

slivne: Hm ...

If clients will not handle the error - it will be an issue (since error handling will not work as well - e.g. retries)

So looking at https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v5.spec there are some additional extensions for CDC. 

We do not support v5 protocol - so once we do support that we will have the ability to return CDC specific codes (nothing matches the preimage unfortunately as Cassandra does not do pre-image) 

So lets settle on something that does work.

gleb-cloudius: @jul-stas I've seen this before. It meas that the error encoding did not match error code. Make sure that you do not lose exception type between the place it was thrown and cqls server where it is encoded into a protocol message.

bhalevy:
On Tue, Apr 21, 2020 at 5:53 PM Shlomi Livne <notifications@github.com>
wrote:

> *@slivne* commented on this pull request.
>
> In exceptions/exceptions.hh
> <https://github.com/scylladb/scylla/pull/6215#discussion_r412256028>:
>
> > @@ -71,6 +71,7 @@ enum class exception_code : int32_t {
>      // Scylla-specific codes
>      // Allocated backwards from 0x1aff-0x1a01 to minimize the chance of collision with Cassandra.
>      OVERFLOW_ERROR  = 0x1aff,
> +    CDC_PREIMAGE_READ_FAILURE = 0x1aef,
>
> Hm ...
>
> If clients will not handle the error - it will be an issue (since error
> handling will not work as well - e.g. retries)
>

Specific to OVERFLOW_ERROR retry won't help and is not desired.
The error needs to be conveyed to the application.

cdc/log.cc
bhalevy: that's usually accepted.

jul-stas:
An update: now we are throwing `read_failure_exception` with custom description (non-standard error is removed). Specific error messages have specific memory layouts. Clients identify the layout by err code, while Scylla chooses error message based on the exception class hierarchy.
The bottomline is that we cannot mix any exception type with any error code.
This is what we get now, instead of `NoHostAvailable`:
```
ReadFailure: Error from server: code=1300 [Replica(s) failed to execute read] message="CDC preimage query could not achieve the CL." info={'failures': 0, 'received_responses': 1, 'required_responses': 2, 'consistency': 'QUORUM'}
```

cdc/log.cc
```diff
@@ -1302,6 +1303,13 @@ class transformer final {
                     }
                     return make_lw_shared<cql3::untyped_result_set>(*result_set);
         });
+      } catch (exceptions::unavailable_exception& e) {
+        // `query` can throw `unavailable_exception`, which is seen by clients as ~ "NoHostAvailable". 
+        // So, we'll translate it to a `read_failure_exception` with custom message.
+        cdc_log.debug("Preimage: translating a (read) `unavailable_exception` to `request_execution_exception` - ", e);
```
psarna, line 1309: I think you need to add `{}` after the dash, otherwise the `e` parameter will never get printed, right?

exceptions/exceptions.hh
avikivity: It turns out that some drivers convert unknown error numbers into protocol errors, so we can't do this. You can use CDC_WRITE_FAILURE instead.

cdc/log.cc
jul-stas: ooops! done.

haaawk:
@psarna I think @jul-stas fixed the last comment.

Issue #6159, opened 2020-04-08
UX: suggest to clean coredump file at the end of scylla_coredump_setup


amoskong:
*Installation details*
Scylla version (or git commit hash): 666.development-0.20200407.16246d1c99e
Cluster size: 1
OS (RHEL/CentOS/Ubuntu/AWS AMI): ami-0d6a24fe35fdf5dc4 (eu-west-1)

### Description:
Commit https://github.com/scylladb/scylla/commit/5f18964763608b5e5d5a833c0777e253d2c21624 added a coredump test in scylla_setup (scylla_coredump_setup), which will generate a coredump file.

### Suggest:
clean the coredump file at the end of scylla_coredump_setup, it might get people confuse.

### Test info:
Job link: https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-oel76/40/
test id: 4c259718-60ce-4621-ba4a-c99f4369a314

This artifact-test failed for a coredump in scylla_setup (coredump setup), the coredump is generated for 'bash' process as expected.

Logs: 
   - db-cluster | https://cloudius-jenkins-test.s3.amazonaws.com/4c259718-60ce-4621-ba4a-c99f4369a314/20200408_001125/db-cluster-4c259718.zip |
   - sct-runner | https://cloudius-jenkins-test.s3.amazonaws.com/4c259718-60ce-4621-ba4a-c99f4369a314/20200408_001125/sct-runner-4c259718.zip

```
04-08 00:00:59,347: Running command "sudo /usr/lib/scylla/scylla_setup --nic eth0 --disks /dev/nvme0n1 "...
....
04-08 00:01:26,085: Creating XFS volume for scylla using 1 disk(s): /dev/nvme0n1
04-08 00:01:26,304: Created symlink from /etc/systemd/system/multi-user.target.wants/var-lib-systemd-coredump.mount to /etc/systemd/system/var-lib-systemd-coredump.mount.
04-08 00:01:26,364: kernel.core_pattern = |/usr/lib/systemd/systemd-coredump %p %u %g %s %t %e"
04-08 00:01:29,383:            PID: 12356 (bash)
04-08 00:01:29,383:            UID: 0 (root)
04-08 00:01:29,383:            GID: 0 (root)
04-08 00:01:29,383:         Signal: 11 (SEGV)
04-08 00:01:29,383:      Timestamp: Wed 2020-04-08 01:01:26 BST (3s ago)
[...]

PR #6216, opened 2020-04-16
fix(scylla_coredump_setup): remove the coredump create by the check

fruch:
since those coredumps are kept after the setup, this make out testing infra
to identify it as a coredump create by the test and fail it.

once it's working we can delete this coredump

Fixes: #6159


fruch:
@penberg, this fails now the AMI test, and both master and enterprise are blocked by this

bentsi:
this should fix #6159

dist/common/scripts/scylla_coredump_setup
```diff
@@ -99,6 +99,7 @@ WantedBy=multi-user.target
         try:
             run('coredumpctl --no-pager --no-legend info {}'.format(pid))
             print('\nsystemd-coredump is working finely.')
+            run('rm /var/lib/systemd/coredump/core.bash.*')
```
bentsi, line 102: better `rm -f `

roydahan:
@eliransin FYI

dist/common/scripts/scylla_coredump_setup
eliransin, line 102: I think it is better to locate the file to delete the way cluster.py at sdcm does it.
see _notify_backtrace function.

amoskong: > I think it is better to locate the file to delete the way cluster.py at sdcm does it.
> see _notify_backtrace function.

It's not just a problem in testing env, we should not leave a meaningless core file to users, it's not graceful.

amoskong: @fruch do you agree to use '-f' option?

fruch: I agree

dist/common/scripts/scylla_coredump_setup
```diff
@@ -99,6 +99,7 @@ WantedBy=multi-user.target
         try:
             run('coredumpctl --no-pager --no-legend info {}'.format(pid))
             print('\nsystemd-coredump is working finely.')
+            run('rm -f /var/lib/systemd/coredump/core.bash.*')
```
eliransin, line 102: Who said that all core.bash* are meaningless, I think I was misunderstood, I think we should delete **only**
the cores that were created in this script or we might delete unrelated legitimate cores.

slivne:
@fruch ping - can you please address @eliransin comment above we need this merged for 4.1

fruch:
I've address @eliransin concerns in 62a213e,
I've opened PR so we can stop doing this step in SCT, if we want something more of of it, that won't need to be in a different PR/patchset

eliransin:
LGTM thanks @fruch

eliransin:
@penberg this one is just pending merge (and your approval of course :slightly_smiling_face:

penberg:
Queued to `next` as commit 28c3d4f8e8f7f13da1c28bcaa135d312e87d2d80.

PR #6226, opened 2020-04-17
dist: add scylla_memory_setup

### LINKED ISSUE #1393: Scylla gets swapped out if user runs some process, with horrible results
_Installation details_
Scylla version (or git commit hash): 1.1
OS (RHEL/CentOS/Ubuntu/AWS AMI): Ubuntu 14.04

If a user runs some administrative task that requires a lot of memory, the kernel might swap out Scylla.  This later results in Scylla being very unresponsive as it pages itself back in.

syuu1228:
To ask user the host is not shared with another services, then set
"--lock-memory 1" if it's not shared.

Fixes #1393


dist/common/systemd/scylla-server.service
```diff
@@ -14,7 +14,7 @@ LimitNPROC=8096
 EnvironmentFile=/etc/sysconfig/scylla-server
 EnvironmentFile=/etc/scylla.d/*.conf
 ExecStartPre=/opt/scylladb/scripts/scylla_prepare
-ExecStart=/usr/bin/scylla $SCYLLA_ARGS $SEASTAR_IO $DEV_MODE $CPUSET
+ExecStart=/usr/bin/scylla $SCYLLA_ARGS $SEASTAR_IO $DEV_MODE $CPUSET $MEM_CONF
 ExecStopPost=/opt/scylladb/scripts/scylla_stop
 TimeoutSec=900
```
avikivity, line 19: Don't we also have to add CAP_IPC_LOCK?

Note, we put this now in capabilities.conf, but maybe we can move it back to scylla-server.service. systemd-219-33 was released more than 3 years ago, we can add a conflict with earlier versions.

syuu1228:
@avikivity Added CAP_IPC_LOCK on capabilities.conf, please review again. I think mergeing capabilities.conf to scylla-server.service should be done on another PR.

avikivity:
Queued as df4fac284918c53f58eeabd1bacd30273c8a5646

PR #6242, opened 2020-04-20
Fix hang in multishard_writer

### LINKED ISSUE #6241: multishard_writer can dead lock when producer fails
We have this in multishard_writer:
    
    future<uint64_t> multishard_writer::operator()() {
        return distribute_mutation_fragments().finally([this] {
            return wait_pending_consumers();
        }).then([this] {
            return _consumed_partitions;
        });
    }
    
The wait_pending_consumers which waits for the consumers to finish is
called even when distribute_mutation_fragments fails.
    
When distribute_mutation_fragments fails and the failure is due to the
producer fails, consumers can wait for data which will never come because
the producer has failed already. This can cause a deadlock.

This is supposed to be the root cause of stream hang during repair observed by a user.

asias:
This series fix hang in multishard_writer when error happens. It contains
- multishard_writer: Abort the queue attached to consumers when producer fails
- repair: Fix hang when the writer is dead

Fixes #6241
Refs: #6248

### DIFF
--- a/mutation_writer/multishard_writer.cc
+++ b/mutation_writer/multishard_writer.cc
@@ -173,6 +173,13 @@ future<> multishard_writer::distribute_mutation_fragments() {
                 return handle_end_of_stream();
             }
         });
+    }).handle_exception([this] (std::exception_ptr ep) {
+        for (auto& q : _queue_reader_handles) {
+            if (q) {
+                q->abort(ep);
+            }
+        }
+        return make_exception_future<>(std::move(ep));
     });
 }
 
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -441,7 +441,7 @@ class repair_writer {
     uint64_t _estimated_partitions;
     size_t _nr_peer_nodes;
     // Needs more than one for repair master
-    std::vector<std::optional<future<uint64_t>>> _writer_done;
+    std::vector<std::optional<future<>>> _writer_done;
     std::vector<std::optional<seastar::queue<mutation_fragment_opt>>> _mq;
     // Current partition written to disk
     std::vector<lw_shared_ptr<const decorated_key_with_hash>> _current_dk_written_to_sstable;
@@ -521,7 +521,15 @@ public:
                 return consumer(std::move(reader));
             });
         },
-        t.stream_in_progress());
+        t.stream_in_progress()).then([this, node_idx] (uint64_t partitions) {
+            rlogger.debug("repair_writer: keyspace={}, table={}, managed to write partitions={} to sstable",
+                _schema->ks_name(), _schema->cf_name(), partitions);
+        }).handle_exception([this, node_idx] (std::exception_ptr ep) {
+            rlogger.warn("repair_writer: keyspace={}, table={}, multishard_writer failed: {}",
+                    _schema->ks_name(), _schema->cf_name(), ep);
+            _mq[node_idx]->abort(ep);
+            return make_exception_future<>(std::move(ep));
+        });
     }
 
     future<> write_partition_end(unsigned node_idx) {
@@ -548,21 +556,33 @@ public:
         }
     }
 
+    future<> write_end_of_stream(unsigned node_idx) {
+        if (_mq[node_idx]) {
+            // Partition_end is never sent on wire, so we have to write one ourselves.
+            return write_partition_end(node_idx).then([this, node_idx] () mutable {
+                // Empty mutation_fragment_opt means no more data, so the writer can seal the sstables.
+                return _mq[node_idx]->push_eventually(mutation_fragment_opt());
+            });
+        } else {
+            return make_ready_future<>();
+        }
+    }
+
+    future<> do_wait_for_writer_done(unsigned node_idx) {
+        if (_writer_done[node_idx]) {
+            return std::move(*(_writer_done[node_idx]));
+        } else {
+            return make_ready_future<>();
+        }
+    }
+
     future<> wait_for_writer_done() {
         return parallel_for_each(boost::irange(unsigned(0), unsigned(_nr_peer_nodes)), [this] (unsigned node_idx) {
-            if (_writer_done[node_idx] && _mq[node_idx]) {
-                // Partition_end is never sent on wire, so we have to write one ourselves.
-                return write_partition_end(node_idx).then([this, node_idx] () mutable {
-                    // Empty mutation_fragment_opt means no more data, so the writer can seal the sstables.
-                    return _mq[node_idx]->push_eventually(mutation_fragment_opt()).then([this, node_idx] () mutable {
-                        return (*_writer_done[node_idx]).then([] (uint64_t partitions) {
-                            rlogger.debug("Managed to write partitions={} to sstable", partitions);
-                            return make_ready_future<>();
-                        });
-                    });
-                });
-            }
-            return make_ready_future<>();
+            return when_all_succeed(write_end_of_stream(node_idx), do_wait_for_writer_done(node_idx));
+        }).handle_exception([this] (std::exception_ptr ep) {
+            rlogger.warn("repair_writer: keyspace={}, table={}, wait_for_writer_done failed: {}",
+                    _schema->ks_name(), _schema->cf_name(), ep);
+            return make_exception_future<>(std::move(ep));
         });
     }
 };
--- a/test/boost/mutation_writer_test.cc
+++ b/test/boost/mutation_writer_test.cc
@@ -117,6 +117,53 @@ SEASTAR_TEST_CASE(test_multishard_writer) {
     });
 }
 
+SEASTAR_TEST_CASE(test_multishard_writer_producer_aborts) {
+    return do_with_cql_env_thread([] (cql_test_env& e) {
+        auto test_random_streams = [] (random_mutation_generator&& gen, size_t partition_nr, generate_error error = generate_error::no) {
+            auto muts = gen(partition_nr);
+            schema_ptr s = gen.schema();
+            auto source_reader = partition_nr > 0 ? flat_mutation_reader_from_mutations(muts) : make_empty_flat_reader(s);
+            int mf_produced = 0;
+            auto get_next_mutation_fragment = [&source_reader, &mf_produced] () mutable {
+                if (mf_produced++ > 800) {
+                    return make_exception_future<mutation_fragment_opt>(std::runtime_error("the producer failed"));
+                } else {
+                    return source_reader(db::no_timeout);
+                }
+            };
+            auto& sharder = s->get_sharder();
+            try {
+                distribute_reader_and_consume_on_shards(s,
+                    make_generating_reader(s, std::move(get_next_mutation_fragment)),
+                    [&sharder, error] (flat_mutation_reader reader) mutable {
+                        if (error) {
+                            return make_exception_future<>(std::runtime_error("Failed to write"));
+                        }
+                        return repeat([&sharder, reader = std::move(reader), error] () mutable {
+                            return reader(db::no_timeout).then([&sharder,  error] (mutation_fragment_opt mf_opt) mutable {
+                                if (mf_opt) {
+                                    if (mf_opt->is_partition_start()) {
+                                        auto shard = sharder.shard_of(mf_opt->as_partition_start().key().token());
+                                        BOOST_REQUIRE_EQUAL(shard, this_shard_id());
+                                    }
+                                    return make_ready_future<stop_iteration>(stop_iteration::no);
+                                } else {
+                                    return make_ready_future<stop_iteration>(stop_iteration::yes);
+                                }
+                            });
+                        });
+                    }
+                ).get0();
+            } catch (...) {
+                // The distribute_reader_and_consume_on_shards is expected to fail and not block forever
+            }
+        };
+
+        test_random_streams(random_mutation_generator(random_mutation_generator::generate_counters::no, local_shard_only::yes), 1000, generate_error::no);
+        test_random_streams(random_mutation_generator(random_mutation_generator::generate_counters::no, local_shard_only::yes), 1000, generate_error::yes);
+    });
+}
+
 namespace {
 
 class bucket_writer {


mutation_writer/multishard_writer.cc
```diff
@@ -173,6 +173,13 @@ future<> multishard_writer::distribute_mutation_fragments() {
                 return handle_end_of_stream();
             }
         });
+    }).handle_exception([this] (std::exception_ptr ep) {
```
gleb-cloudius, line 176: Why add more continuations? You can replace `final()` clause in `multishard_writer::operator()` with `then_wrappe()` and handle the error there, no?

asias: It is the job of multishard_writer::distribute_mutation_fragments to abort queue. Not the job of the caller.  Also multishard_writer::distribute_mutation_fragments is called once. I do not see a problem attaching a .handle_exception to it.  I also think handle_exception is cleaner than then_wrapper.

gleb-cloudius: It may be conceptually cleaner, but it requires more allocations. There is nothing wrong with using then_wrapped to handle both happy and unhappy path if both has to be handled. In fact this is preferred practice. As of hwos job it is - this is all internal multishard_writer code, so callers and callees are  very intervened here. I do have a point that this is not a fast path, so may be extra allocation is not that matter.

But at least make sure that the error is logged. Looking at the code I actually think it will not be.

asias: The error is now logged in the second version.

asias:
Change in b7e36e80e53c1d4929ab09ea9be3bc9b6c297c63:
- Added "multishard_writer: Fix hang when the writer is dead" to fix another hang issue.

mutation_writer/multishard_writer.cc
```diff
@@ -181,14 +192,18 @@ future<uint64_t> multishard_writer::operator()() {
         return wait_pending_consumers();
     }).then([this] {
         return _consumed_partitions;
+    }).handle_exception([this] (std::exception_ptr ep) {
+        _on_fail(ep);
```
gleb-cloudius, line 196: I do not understand why you need a callback if you propagate the error to the caller by returning it. Juts log it at the caller.

asias: Because the future returned from the writer is accessed only in wait_for_writer_done when we call stop().

I think I can find a way to avoid the callback. 

I will send a new version.

asias:
Changes in 41617e22c0e8eaa73e8c625370718d452d48885b:
- Drop the on_fail callback to simplify the logic

repair/row_level.cc
```diff
@@ -563,6 +567,14 @@ class repair_writer {
                 });
             }
             return make_ready_future<>();
+        }).handle_exception([this] (std::exception_ptr ep) {
+            for (auto& w : _writer_done) {
+                if (w && w->failed()) {
+                    rlogger.warn("repair_writer::wait_for_writer_done: writer failed: {}", w->get_exception());
```
gleb-cloudius, line 573: Didn't you already reported this above?

asias: With w->get_exception() at line 573 above, I saw "Exceptional future ignored" errors. This can happen if error happens before line 563 below.  We can not sallow the exception in mutation_writer::distribute_reader_and_consume_on_shards exception handler. 

```
555     future<> wait_for_writer_done() {
 556         return parallel_for_each(boost::irange(unsigned(0), unsigned(_nr_peer_nodes)), [this] (unsigned node_idx) {
 557             if (_writer_done[node_idx] && _mq[node_idx]) {
 558                 return make_exception_future<>(std::runtime_error("error1"));
 559                 // Partition_end is never sent on wire, so we have to write one ourselves.
 560                 return write_partition_end(node_idx).then([this, node_idx] () mutable {
 561                     // Empty mutation_fragment_opt means no more data, so the writer can seal the sstables.
 562                     return _mq[node_idx]->push_eventually(mutation_fragment_opt()).then([this, node_idx] () mutable {
 563                         return (*_writer_done[node_idx]).then([] (uint64_t partitions) {
 564                             rlogger.debug("Managed to write partitions={} to sstable", partitions);
 565                             return make_ready_future<>();
 566                         });
 567                     });
 568                 });
 569             }
 570             return make_ready_future<>();
 571         }).handle_exception([this] (std::exception_ptr ep) {
 572             for (auto& w : _writer_done) {
 573                 if (w && w->failed()) {
 574                     rlogger.warn("repair_writer::wait_for_writer_done: writer failed: {}", w->get_exception());
 575                 }
 576             }
 577             rlogger.warn("repair_writer::wait_for_writer_done failed: {}", ep);
 578             return make_exception_future<>(std::move(ep));
 579         });
 580     }

```

gleb-cloudius: I think you see unhandled exception because you are propagating it in 527.

asias: Yes, but we can not swallow the exception at line 527, otherwise wait_for_writer_done would see the operation is successful while we have an exception in distribute_reader_and_consume_on_shards. 

Perhaps, we can call w->ignore_ready_future() at live 574 above instead calling get_exception() and log it. This happens rarely, I think it is better to log it here.

gleb-cloudius: Why is it important for wait_for_writer_done to see the exception if it only logs it and it was already logged anyway?

asias: In wait_for_writer_done, we need to either read the exception or call ignore_ready_future, otherwise we see ignored future warnings.  Log is not a must. 

wait_for_writer_done need to see the exception, because we need to propagate the error to the upper layer to know the repair is failed.

gleb-cloudius: The code is like that:
```
_writer_done[node_idx] = mutation_writer::distribute_reader_and_consume_on_shards(...).handle_exception(ex) {
   log(ex)
   return ex;
}
```

If you will not return ex  _writer_done will never be failed exception so there will be no need to call for ignore_ready_future.  wait_for_writer_done() does not propagate errors from _writer_done.

asias: We have
```
 562                         return (*_writer_done[node_idx]).then([] (uint64_t partitions) {
 563                             rlogger.debug("Managed to write partitions={} to sstable", partitions);
 564                             return make_ready_future<>();
 565                         });
```
in wait_for_writer_done. If _writer_done fails, it propagates errors to the caller of wait_for_writer_done. This is what we wanted. That's why we should not swallow the exception in the the exception handler in distribute_reader_and_consume_on_shards .

avikivity: Why both log and return the exception? Whoever gets the exception will log it.

avikivity:
@denesb please review

mutation_writer/multishard_writer.cc
```diff
@@ -173,6 +173,13 @@ future<> multishard_writer::distribute_mutation_fragments() {
...
             }
         });
+    }).handle_exception([this] (std::exception_ptr ep) {
+        for (auto& q : _queue_reader_handles) {
+            if (q) {
+                q->abort(ep);
+            }
+        }
+        return make_exception_future<>(std::move(ep));
     });
```
avikivity, line 183: Please add a unit test.

repair/row_level.cc
gleb-cloudius: So why do you have logging for _writer_done scattered everywhere? Do:
```
_writer_done[node_idx] = mutation_writer::distribute_reader_and_consume_on_shards(...).then_wrapped(f) {
    if (f.failed())
       log(ex)
   else
      log("success");
}
```

Less references you have for _writer_done the easier  to understand the logic.

asias:
``` 
 710     future<> stop() {
 711         auto gate_future = _gate.close();
 712         auto writer_future = _repair_writer.wait_for_writer_done();
 713         auto f1 = _sink_source_for_get_full_row_hashes.close();
 714         auto f2 = _sink_source_for_get_row_diff.close();
 715         auto f3 = _sink_source_for_put_row_diff.close();
 716         return when_all_succeed(std::move(gate_future), std::move(writer_future), std::move(f1), std::move(f2), std::move(f3));
 717     }
```
The consumer of the wait_for_writer_done is when_all_succeed. Only one of the exception will be returned, it might not be exception from wait_for_writer_done.

asias: > So why do you have logging for _writer_done scattered everywhere? Do:
> 
> ```
> _writer_done[node_idx] = mutation_writer::distribute_reader_and_consume_on_shards(...).then_wrapped(f) {
>     if (f.failed())
>        log(ex)
>    else
>       log("success");
> }
> ```
> 
> Less references you have for _writer_done the easier to understand the logic.

@gleb-cloudius As I mentioned above. I agree it is not a must to log here again:

```
 573                 if (w && w->failed()) {
 574                     rlogger.warn("repair_writer::wait_for_writer_done: writer failed: {}", w->get_exception());
 575                 }
```

It is wrong to swallow the exception though. Do you agree?

gleb-cloudius: I do not understand your argument. In your sniped you log and swallow the exception. Since the exception was already logged there is no point in logging it again so the only purpose of the code is to swallow the exception in which case why do you propagate it after logging. 

You pointed out that you have another continuation attached to _writer_done that logs success and I suggest to move that logging to the same continuation that log an exception.

asias: > In your sniped you log and swallow the exception. 

I know the misunderstanding know. Yes,  here we swallow the exception. But it does not matter, because wait_for_writer_done already knows there is something went wrong it could be error from distribute_reader_and_consume_on_shards or other fibers. The thing is wait_for_writer_done will return an exception to the caller. 

```
573                 if (w && w->failed()) {
 574                     rlogger.warn("repair_writer::wait_for_writer_done: writer failed: {}", w->get_exception());
 575                 }
```

I never said we need to log the success.

My argument is that we should never swallow the exception at line 527 below.

````
 494         _writer_done[node_idx] = mutation_writer::distribute_reader_and_consume_on_shards(_schema,
 495                 make_generating_reader(_schema, std::move(get_next_mutation_fragment)),
 496                 [&db, estimated_partitions = this->_estimated_partitions] (flat_mutation_reader reader) {
 497             auto& t = db.local().find_column_family(reader.schema());
 498             return db::view::check_needs_view_update_path(_sys_dist_ks->local(), t, streaming::stream_reason::repair).then([t = t.shared_from_this(), estimated_partitions, reader = std::move(reader)] (bool use_view_update_path     ) mutable {
 499                 //FIXME: for better estimations this should be transmitted from remote
 500                 auto metadata = mutation_source_metadata{};
 501                 auto& cs = t->get_compaction_strategy();
 502                 const auto adjusted_estimated_partitions = cs.adjust_partition_estimate(metadata, estimated_partitions);
 503                 auto consumer = cs.make_interposer_consumer(metadata,
 504                         [t = std::move(t), use_view_update_path, adjusted_estimated_partitions] (flat_mutation_reader reader) {
 505                     sstables::shared_sstable sst = use_view_update_path ? t->make_streaming_staging_sstable() : t->make_streaming_sstable_for_write();
 506                     schema_ptr s = reader.schema();
 507                     auto& pc = service::get_local_streaming_write_priority();
 508                     return sst->write_components(std::move(reader), std::max(1ul, adjusted_estimated_partitions), s,
 509                                                  t->get_sstables_manager().configure_writer(),
 510                                                  encoding_stats{}, pc).then([sst] {
 511                         return sst->open_data();
 512                     }).then([t, sst] {
 513                         return t->add_sstable_and_update_cache(sst);
 514                     }).then([t, s, sst, use_view_update_path]() mutable -> future<> {
 515                         if (!use_view_update_path) {
 516                             return make_ready_future<>();
 517                         }
 518                         return _view_update_generator->local().register_staging_sstable(sst, std::move(t));
 519                     });
 520                 });
 521                 return consumer(std::move(reader));
 522             });
 523         },
 524         t.stream_in_progress()).handle_exception([this, node_idx] (std::exception_ptr ep) {
 525             rlogger.warn("repair_writer::create_writer: multishard_writer failed: {}", ep);
 526             _mq[node_idx]->abort(ep);
 527             return make_exception_future<uint64_t>(std::move(ep));
 528         });
```

gleb-cloudius: > My argument is that we should never swallow the exception at line 527 below.

And I do not see why. The continuation should handle both exception and success and propagate nothing. A continuation that logs success in  wait_for_writer_done should be dropped. BTW if _mq[node_idx]->push_eventually(mutation_fragment_opt()) there fails nobody waits for _writer_done which may probably cause use after free.

mutation_writer/multishard_writer.cc
asias: Done. Will send a new version.

asias:
Chang in 01638735f5b923e892e77d0429b804801596caa0 (v4):
- Dropped a for _writer_done 
- Added unit test to catch the hang

repair/row_level.cc
asias: > The continuation should handle both exception and success and propagate nothing.

How does it tell wait_for_writer_done  something went wrong then? The exception handler only abort the _mq, when it aborts, wait_for_writer_done might have already finished run all the code right before calling *_writer_done[node_idx]).then(). In this case, wait_for_writer_done will wrongly report the operation is success.

```
 524         t.stream_in_progress()).handle_exception([this, node_idx] (std::exception_ptr ep) {
 525             rlogger.warn("repair_writer::create_writer: multishard_writer failed: {}", ep);
 526             _mq[node_idx]->abort(ep);
 527             return make_exception_future<uint64_t>(std::move(ep));
 528         });

```

Regarding wait for _writer_done in case of error, perhaps we can have:

```
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -558,7 +558,7 @@ class repair_writer {
                 // Partition_end is never sent on wire, so we have to write one ourselves.
                 return write_partition_end(node_idx).then([this, node_idx] () mutable {
                     // Empty mutation_fragment_opt means no more data, so the writer can seal the sstables.
-                    return _mq[node_idx]->push_eventually(mutation_fragment_opt()).then([this, node_idx] () mutable {
+                    return _mq[node_idx]->push_eventually(mutation_fragment_opt()).finally([this, node_idx] () mutable {
                         return (*_writer_done[node_idx]).then([] (uint64_t partitions) {
                             rlogger.debug("Managed to write partitions={} to sstable", partitions);
                             return make_ready_future<>();
```

We need to take care of  parallel_for_each because if one of them abort, we still need to wait for the other to run which waits for _writer_done.

repair/row_level.cc
```diff
@@ -563,6 +567,14 @@ class repair_writer {
                 });
             }
             return make_ready_future<>();
+        }).handle_exception([this] (std::exception_ptr ep) {
+            for (auto& w : _writer_done) {
```
gleb-cloudius, line 571: Well you did not really addressed any of my comments. But the discussion is gone. Yeah for pull requests! 
So I will repeat it again: if in line 561 `_mq[node_idx]->push_eventually(mutation_fragment_opt())` fails here a future in `_writer_done[]` may still be unresolved so the code will crash here.

repair/row_level.cc
gleb-cloudius: You abort _mq. This will propagate the error. no?

repair/row_level.cc
asias: The discussion is not gone: see here https://github.com/scylladb/scylla/pull/6242#discussion_r412790718.

Regarding use after free, I think we are still in discussion how to address it, it is not included in v4.  I sent v4 which adds tests that Avi asked for.  I also dropped the double log which you requested.

repair/row_level.cc
asias: It might not propagate because wait_for_writer_done might be done with the _mq when it abort_mq:

```
 The exception handler only abort the _mq, when it aborts, wait_for_writer_done might have already finished run all the code right before calling *_writer_done[node_idx]).then(). In this case, wait_for_writer_done will wrongly report the operation is success.
```

gleb-cloudius: Who do you want to propagate the error to? You already logged it. Also which error are you going to propagate there are a lot of _mq and _writer_done and any of them can have an error?

repair/row_level.cc
gleb-cloudius: > The discussion is not gone: see here [#6242 (comment)](https://github.com/scylladb/scylla/pull/6242#discussion_r412790718).

This is impossible to follow. Suddenly I  need to look at some other place to continue a discussion? It is not enough that I need to refresh the page to see your replies because unlike the email the interface is pull and not push, but now I need to refresh two pages? How is it usable?

repair/row_level.cc
asias: The caller of wait_for_writer_done() which in turn propagate to who started the repair. It does not matter which error to propagate to the upper layer, what important is that we report the error and do not falsely report the repair operation is ok but actually there is error.

denesb:
LGTM.

repair/row_level.cc
gleb-cloudius: If you care about only one error you can store it in a variable in repair_writer class. I think the code needs serious refactoring here because what wait_for_writer_done() tries to do it wait_all_succes() on two set of futures but open codes it.

gleb-cloudius:
> LGTM.

This version cannot be merged since it may crash. At least changing from then() to final() is required like @asias proposed above. I think more refactoring is needed, but final() is a must.

denesb:
> > LGTM.
> 
> This version cannot be merged since it may crash. At least changing from then() to final() is required like @asias proposed above. I think more refactoring is needed, but final() is a must.

I meant apart from the discussion above, which I attempted to follow but failed. I do think Asias has a point, the error has to be propagated up, otherwise the upper layer will declare the repair as successfull.

asias:
> > > LGTM.
> > 
> > 
> > This version cannot be merged since it may crash. At least changing from then() to final() is required like @asias proposed above. I think more refactoring is needed, but final() is a must.
> 
> I meant apart from the discussion above, which I attempted to follow but failed. I do think Asias has a point, the error has to be propagated up, otherwise the upper layer will declare the repair as successfull.

Thanks guys. Sure, let's not merge it.  I am working on a new version to simplify things and  solve waiting for _writer_done issue.

gleb-cloudius:
> > > LGTM.
> > 
> > 
> > This version cannot be merged since it may crash. At least changing from then() to final() is required like @asias proposed above. I think more refactoring is needed, but final() is a must.
> 
> I meant apart from the discussion above, which I attempted to follow but failed. I do think Asias has a point, the error has to be propagated up, otherwise the upper layer will declare the repair as successfull.

Some error has to. I do not think we need to propagate all of them through all the layers and then explicitly frop all but one.

denesb:
> > > > LGTM.
> > > 
> > > 
> > > This version cannot be merged since it may crash. At least changing from then() to final() is required like @asias proposed above. I think more refactoring is needed, but final() is a must.
> > 
> > 
> > I meant apart from the discussion above, which I attempted to follow but failed. I do think Asias has a point, the error has to be propagated up, otherwise the upper layer will declare the repair as successfull.
> 
> Some error has to. I do not think we need to propagate all of them through all the layers and then explicitly frop all but one.

Yes, it is enough to just signal that repair has failed. The exact reason (could be multiple reasons) will be logged already anyway.

asias:
Changes in d2bf64ffff7baec9da6c770e9f3f3b749cb8605b (v5)
- Always wait for _writer_done to avoid use after free
- Drop the check of _writer_done->failed() since we now always wait for it
- Improve the log to contain keyspace and table name

repair/row_level.cc
asias: I tried to use when_all_succeed as an alternative to wait for _writer_done.  What we need to do is not just waiing for it. We need to write some data to the queue then wait for _writer_done.  The parallel_for_each already does the job  to store one of the error, so no need to introduce another class member. wait_for_writer_done actually waits for N set of futures, N is a subset of the vector _writer_done[]. I think parallel_for_each is better than using when_all_succeed in wait_for_writer_done.

gleb-cloudius: Why can't we wait for it and then write into the queue? Why can't we initiate wait (by calling when_all_succeed()) and then write into the queue?

gleb-cloudius: What I mean is that it looks like wait_for_writer_done() start some activity on mq[] and than waits for that activity to complete and also waits for everything in _writer_done. It can be expressed as:
```
auto mq_futures = mq |  boost::adaptors::filtered(is_engaged) | boost::adaptors::transformed( [] (mq) { return write_partition_end().then([] { return mq->push_eventually(); })});
auto writer_futures = _writer_done | boost::adaptors::filtered(is_engaged) ;

return wait_for_succeed(boost::range::join(mg_futures, writer_futures ));
```

From this pseudoscience it is clear that the code waits for bunch of featured and forwards one of the failures if any to the caller.

asias: > Why can't we wait for it and then write into the queue? Why can't we initiate wait (by calling when_all_succeed()) and then write into the queue?

Because the _writer_done won't finish until the write into the queue.  Why do we want to reverse the operation, and for what purpose?  Take a look at  d2bf64f (v5). We do not really need the when_all_succeed.

gleb-cloudius:
> Because the _writer_done won't finish until the write into the queue.
This does not mean we cannot start waiting on it.

--
			Gleb.

asias:
> Because the _writer_done won't finish until the write into the queue.
> This does not mean we cannot start waiting on it.

But why do we want to do that which reversing the actual logic flow.

gleb-cloudius:
> > Because the _writer_done won't finish until the write into the queue.
> > This does not mean we cannot start waiting on it.
> 
> But why do we want to do that which reversing the actual logic flow.
This is not uncommon to start waiting on a future before starting a fiber that will resolve it, so there is nothing confusing about it. But see the pseudo code I proposed. It does not wait on writers before writing to mqs, it writes into mqs and that waits for all futures together.

asias:
> > > Because the _writer_done won't finish until the write into the queue.
> > > This does not mean we cannot start waiting on it.
> > 
> > 
> > But why do we want to do that which reversing the actual logic flow.
> 
> This is not uncommon to start waiting on a future before starting a fiber that will resolve it, so there is nothing confusing about it. But see the pseudo code I proposed. It does not wait on writers before writing to mqs, it writes into mqs and that waits for all futures together.

The code you proposed has lots of boost transformed, filtered , join and is_enaged stuff.   My code use none of them. The logic in my code is 
```
parallel_for_each([] {
  return do_foo().finally([] {
     return do_bar();
  });
})

```
It is pretty standard to call something unconditionally with finally. I am not sure why we want to do all the transformation in order to use when_all_succeed instead of parallel_for_each. parallel_for_each will report one of the failed future which is the  same as when_all_succeed. 

Btw, we also need the node_index to access _mq, we need extra code to get it with your proposed code.

gleb-cloudius:
On Thu, Apr 23, 2020 at 05:56:01PM -0700, Asias He wrote:
> > > > Because the _writer_done won't finish until the write into the queue.
> > > > This does not mean we cannot start waiting on it.
> > > 
> > > 
> > > But why do we want to do that which reversing the actual logic flow.
> > 
> > This is not uncommon to start waiting on a future before starting a fiber that will resolve it, so there is nothing confusing about it. But see the pseudo code I proposed. It does not wait on writers before writing to mqs, it writes into mqs and that waits for all futures together.
> 
> The code you proposed has lots of boost transformed, filtered , join and is_enaged stuff.
This is a good thing because it uses standard idioms to describe what
is going on which makes it easy to understand without need to read the
code closely.

> ```
> parallel_for_each([] {
>   return do_foo().finally([] {
>      return do_bar();
>   });
> })
> 
> ```

The problem with this code is that there is no straightforward dependency between
do_foo() and do_bar(). And since the code does not expresses that clearly
there was a bug that finally() was then() before. Such code would have
made sense if do_bar() did a cleanup after do_foo() but it does not.

> It is pretty standard to call something unconditionally with finally. I am not sure why we want to do all the transformation in order to use when_all_succeed instead of parallel_for_each. parallel_for_each will report one of the failed future which is the  same as when_all_succeed. 
when_all_succeed() express more clearly the intention of the code here -
the fact we want to do a bunch of stuff on a bunch of fibers and do not
care much about what happens to them but forwarding one error to a
caller. I do not _insist_ on not using parallel_for_each of course, I
just showed a way to use when_all_succeed() on a writers which you said
you could not do.
> 
> Btw, we also need the node_index to access _mq, we need extra code to get it with your proposed code. 
> 
I am not sure why since transformed() will provide you with a pointer.
But if you do not like the mq pipeline you can leave mq part in
parallel_for_each and do:

auto f = parallel_for_each(do only mq part);

return wait_for_all_succeed(f, writers | filter(empty));

--
			Gleb.

asias:
> This is a good thing because it uses standard idioms to describe what is going on which makes it easy to understand without need to read the code closely.

In my view, it is a little abuse of boost and is making the code more complicated than it is necessary.

asias:
> The problem with this code is that there is no straightforward dependency between do_foo() and do_bar(). And since the code does not expresses that clearly there was a bug that finally() was then() before. Such code would have made sense if do_bar() did a cleanup after do_foo() but it does not.

finally() is not only invented  to do cleanup. There is a comment in the code now says why we need to call finally in the pipeline.

asias:
> when_all_succeed() express more clearly the intention of the code here - the fact we want to do a bunch of stuff on a bunch of fibers and do not care much about what happens to them but forwarding one error to a caller. I do not _insist_ on not using parallel_for_each of course, I just showed a way to use when_all_succeed() on a writers which you said you could not do.

Good to know that you do not insist on not using parallel_for_each.

gleb-cloudius:
On Thu, Apr 23, 2020 at 11:16:58PM -0700, Asias He wrote:
> > This is a good thing because it uses standard idioms to describe what is going on which makes it easy to understand without need to read the code closely.
> 
> In my view, it is a little abuse of boost and is making the code more complicated than it is necessary. 
> 
This is how it was intended to be used. Using parallel_for_each is not
an abuse even though for() can be used.

--
			Gleb.

asias:
> I am not sure why since transformed() will provide you with a pointer. But if you do not like the mq pipeline you can leave mq part in parallel_for_each and do: auto f = parallel_for_each(do only mq part); return wait_for_all_succeed(f, writers | filter(empty));

The parallel_for_each part would check if  _writer_done[] is empty (we need to write only if the _writer_done[]  is enaged ) and we need to do it again in writers | filter(empty) part.

gleb-cloudius:
On Thu, Apr 23, 2020 at 11:18:59PM -0700, Asias He wrote:
> > The problem with this code is that there is no straightforward dependency between do_foo() and do_bar(). And since the code does not expresses that clearly there was a bug that finally() was then() before. Such code would have made sense if do_bar() did a cleanup after do_foo() but it does not.
> 
> finally() is not only invented  to do cleanup. There is a comment in the code now says why we need to call finally in the pipeline. 
> 
finally() was modeled after the one in java which is literally about
cleanup. Isn't it better to make code express the intention instead of
put it in a comment?

--
			Gleb.

gleb-cloudius:
On Thu, Apr 23, 2020 at 11:20:49PM -0700, Asias He wrote:
> > when_all_succeed() express more clearly the intention of the code here - the fact we want to do a bunch of stuff on a bunch of fibers and do not care much about what happens to them but forwarding one error to a caller. I do not _insist_ on not using parallel_for_each of course, I just showed a way to use when_all_succeed() on a writers which you said you could not do.
> 
> Good to know that you do not insist on not using parallel_for_each.
> 
I am not. But I do want to get rid of dependency between mq completion and
writers waiting that is in the code now but logically does not exist.

But if you think about what code is doing it maps between set of mqs to
a set of futures and this is what transform() allows to express in a
nice functional manner.

--
			Gleb.

gleb-cloudius:
On Thu, Apr 23, 2020 at 11:25:58PM -0700, Asias He wrote:
> > I am not sure why since transformed() will provide you with a pointer. But if you do not like the mq pipeline you can leave mq part in parallel_for_each and do: auto f = parallel_for_each(do only mq part); return wait_for_all_succeed(f, writers | filter(empty));
> 
> The parallel_for_each part would check if  _writer_done[] is empty (we need to write only if the _writer_done[]  is enaged ) and we need to do it again in writers | filter(empty) part. 
>
That another thing I wanted to ask: how can it be that there is an mq
but not corresponding writer or vice versa?

--
			Gleb.

asias:
> finally() was modeled after the one in java which is literally about cleanup. Isn't it better to make code express the intention instead of put it in a comment?

Isn't it pretty clear with the finally. The intentions is we always call the finally part no matter what.  And we preserve the logic sequence we want to call do_foo() then call do_far().

asias:
> That another thing I wanted to ask: how can it be that there is an mq but not corresponding writer or vice versa?

- set _mq
- error happens
- set _write_done is not called

asias:
@gleb-cloudius  Other than you think it would be better if we could refactor with boost adaptors and use when_all,  do you think current version is good enough to be merged? May I remind you this is going to be backported to an ent branch and we have a user waiting for a fix.

gleb-cloudius:
On Thu, Apr 23, 2020 at 11:35:10PM -0700, Asias He wrote:
> > That another thing I wanted to ask: how can it be that there is an mq but not corresponding writer or vice versa?
> 
> - set _mq
> - error happens
If error happened we should not push into any _mq but abort the whole
operation.

> - set _write_done is not called

--
			Gleb.

gleb-cloudius:
On Thu, Apr 23, 2020 at 11:43:36PM -0700, Asias He wrote:
> @gleb-cloudius  Other than you think it would be better if we could refactor with boost adaptors and use when_all,  do you think current version is good enough to be merged? May I remind you this is going to be backported to an ent branch and we have a user waiting for a fix.  
> 
The current code will work as far as I can tell. A local refactoring
like proposed here will not complicate backporting.

--
			Gleb.

asias:
> If error happened we should not push into any _mq but abort the whole operation.

The whole repair stops in case of that error, but as part of the cleanup, we will call stop() which in turn calls wait_for_writer_done().

asias:
> On Thu, Apr 23, 2020 at 11:43:36PM -0700, Asias He wrote: @gleb-cloudius Other than you think it would be better if we could refactor with boost adaptors and use when_all, do you think current version is good enough to be merged? May I remind you this is going to be backported to an ent branch and we have a user waiting for a fix.
> > The current code will work as far as I can tell. A local refactoring like proposed here will not complicate backporting.

Good. Since you do not insist on not using parallel_for_each and you think the code is correct, I assume you are ok to merge this series as it is.

gleb-cloudius:
On Fri, Apr 24, 2020 at 12:12:36AM -0700, Asias He wrote:
> > On Thu, Apr 23, 2020 at 11:43:36PM -0700, Asias He wrote: @gleb-cloudius Other than you think it would be better if we could refactor with boost adaptors and use when_all, do you think current version is good enough to be merged? May I remind you this is going to be backported to an ent branch and we have a user waiting for a fix.
> > > The current code will work as far as I can tell. A local refactoring like proposed here will not complicate backporting.
> 
> Good. Since you do not insist on not using parallel_for_each and you think the code is correct, I assume you are ok to merge this series as it is. 
> 
No. I am asking to wait for writers separately and not intermix it with
mq operations. They are logically independent.

--
			Gleb.

gleb-cloudius:
On Fri, Apr 24, 2020 at 12:04:59AM -0700, Asias He wrote:
> > If error happened we should not push into any _mq but abort the whole operation.
> 
> The whole repair stops in case of that error, but as part of the cleanup, we will call stop() which in turn calls wait_for_writer_done().
> 
It should be communicated to wait_for_writer_done() that an error
already happened. It does not make sense to call push on already
aborted mq and if mq is not aborted a push may hang because of previous
errors.

--
			Gleb.

asias:
Changes in cd72da1b4ac3ad6870b937a5dd6a323bebbd2762 (v6)
- Use when_all_succeed instead of finally to call _writer_done

asias:
> On Fri, Apr 24, 2020 at 12:12:36AM -0700, Asias He wrote: > On Thu, Apr 23, 2020 at 11:43:36PM -0700, Asias He wrote: @gleb-cloudius Other than you think it would be better if we could refactor with boost adaptors and use when_all, do you think current version is good enough to be merged? May I remind you this is going to be backported to an ent branch and we have a user waiting for a fix. > > The current code will work as far as I can tell. A local refactoring like proposed here will not complicate backporting. Good. Since you do not insist on not using parallel_for_each and you think the code is correct, I assume you are ok to merge this series as it is.
> No. I am asking to wait for writers separately and not intermix it with mq operations. They are logically independent.

See v6 below.

asias:
> It should be communicated to wait_for_writer_done() that an error already happened. It does not make sense to call push on already aborted mq and if mq is not aborted a push may hang because of previous errors.

The aborted mq is the flag. Whenever the writer is dead the mq is aborted, the push will not hang. In case of errors other than the writer is dead, we may still want to write the end of the stream to seal the existing data already been written so far for this repair operation.

repair/row_level.cc
```diff
@@ -548,21 +553,31 @@ class repair_writer {
...
+            // Empty mutation_fragment_opt means no more data, so the writer can seal the sstables.
+            return _mq[node_idx]->push_eventually(mutation_fragment_opt());
+        });
+    }
+
+    future<> do_wait_for_writer_done(unsigned node_idx) {
+        return (*_writer_done[node_idx]).then([] (uint64_t partitions) {
+            rlogger.debug("Managed to write partitions={} to sstable", partitions);
+            return make_ready_future<>();
+        });
```
gleb-cloudius, line 568: Why no do success logging in the same continuation as a failure. That you will not even need this function, simply move _writer_done directly to when_all().

gleb-cloudius:
On Sun, Apr 26, 2020 at 06:06:58PM -0700, Asias He wrote:
> > It should be communicated to wait_for_writer_done() that an error already happened. It does not make sense to call push on already aborted mq and if mq is not aborted a push may hang because of previous errors.
> 
> The aborted mq is the flag. Whenever the writer is dead the mq is aborted, the push will not hang. In case of errors other than the writer is dead, we may still want to write the end of the stream to seal the existing data already been written so far for this repair operation. 
> 

Then I do not see why is it important to check _writer_done before
pushing to _mq. As you say pushing into _mq will fail anyway. It would
be nice of course to not even push if operation is already failed, but
getting an error from _mq is also works.

--
			Gleb.

repair/row_level.cc
asias: I did not get you. What do you mean? Are you complaining we should not have the debug log here?

asias:
> Then I do not see why is it important to check _writer_done before pushing to _mq. As you say pushing into _mq will fail anyway. It would be nice of course to not even push if operation is already failed, but getting an error from _mq is also works.

We can move the check into write_end_of_stream and do_wait_for_writer_done.

asias:
Changes in b17aa336c23cd6020d722e6840ba053333ef6d9e (v7):
- Move the check into write_end_of_stream and do_wait_for_writer_done.

gleb-cloudius:
On Mon, Apr 27, 2020 at 02:46:34AM -0700, Asias He wrote:
> @asias commented on this pull request.
> 
> 
> 
> > @@ -548,21 +553,31 @@ class repair_writer {
>          }
>      }
>  
> +    future<> write_end_of_stream(unsigned node_idx) {
> +        // Partition_end is never sent on wire, so we have to write one ourselves.
> +        return write_partition_end(node_idx).then([this, node_idx] () mutable {
> +            // Empty mutation_fragment_opt means no more data, so the writer can seal the sstables.
> +            return _mq[node_idx]->push_eventually(mutation_fragment_opt());
> +        });
> +    }
> +
> +    future<> do_wait_for_writer_done(unsigned node_idx) {
> +        return (*_writer_done[node_idx]).then([] (uint64_t partitions) {
> +            rlogger.debug("Managed to write partitions={} to sstable", partitions);
> +            return make_ready_future<>();
> +        });
> 
> I did not get you. What do you mean? Are you complaining we should not have the debug log here?
> 
No, I am saying we can move it where _writer_done future is created. We
already have there:

_writer_done = create().handle_exception(log error);

Why not do:

_writer_done = create().then_wrapped(log error or success)

instead of splitting it like that?

--
			Gleb.

asias:
Changes in 35c5ef78b9e253c6c05302a377967b1d202a79b0 (v8):
- Move the log of  _writer_done

repair/row_level.cc
```diff
@@ -521,7 +521,15 @@ class repair_writer {
                 return consumer(std::move(reader));
             });
         },
-        t.stream_in_progress());
+        t.stream_in_progress()).then([this, node_idx] (uint64_t partitions) {
```
gleb-cloudius, line 524: I see that you do not like using then_wrapped() :) On a fast path it is important. I hope this one is not such.

asias: It is on a slow path. Right, I am not a fan of then_wrapped. I use it only when I have no choice.

asias:
@avikivity @nyh Please consider merge it.  I have bless from gleb and botond.

repair/row_level.cc
gleb-cloudius: It saves compilation time and the code size as well...

asias:
Ping#2 for merge.

asias:
Ping#3 for merge.

avikivity:
Queued as 8925e00e96

avikivity:
For the record, I prefer then_wrapped too.

PR #6244, opened 2020-04-20
dist/offline_installer/redhat: fix umask error

### LINKED ISSUE #6243: dist/offline_installer/redhat: causes "scylla does not work with current umask setting (0077)"
After installed rpm, the script causes following error while auto-executing scylla_setup.
```
Scylla does not work with current umask setting (0077),
please restore umask to the default value (0022).
```
Because offline install script which generated by "makeself" modifies umask at the beginning of the script:
```
#!/bin/sh
# This script was generated using Makeself 2.2.0

umask 077
```

syuu1228:
Since makeself script changes current umask, scylla_setup causes
"scylla does not work with current umask setting (0077)" error.
To fix that we need use latest version of makeself, and specfiy --keep-umask
option.

FIxes #6243


dist/offline_installer/redhat/build_offline_installer.sh
```diff
@@ -109,5 +106,15 @@ if [ -n "$RELEASEVER" ]; then
...
 fi
 sudo yum -y install $YUMOPTS --downloadonly --installroot=`pwd`/build/installroot --downloaddir=build/offline_installer scylla sudo ntp ntpdate net-tools kernel-tools
+cd build
+if [ ! -x makeself-$MAKESELFVER.run ]; then
+    wget https://github.com/megastep/makeself/releases/download/release-$MAKESELFVER/makeself-$MAKESELFVER.run
+    chmod a+x makeself-$MAKESELFVER.run
+fi
+if [ ! -d ./makeself-$MAKESELFVER ]; then
+    ./makeself-$MAKESELFVER.run
+fi
```
penberg, line 116: Why do we need this? The `makeself` version starting from Fedora 30 is 2.4.0 so at least our `dbuild` should have the correct one.

syuu1228: As I mentioned on debian version https://github.com/scylladb/scylla/pull/6246#discussion_r415639584, current script only support to run in target distribution, CentOS7.
And it does not have makeself-2.4.0.

syuu1228:
@penberg @avikivity pushed new version, witch supports building offline_installer.sh in Fedora 32 using Docker (dbuild), and use makeself-2.4.0 on Fedora 32.
Please review again.

dist/offline_installer/redhat/build_offline_installer.sh
```diff
@@ -92,22 +56,19 @@ if [ ! -f /usr/bin/createrepo ]; then
...
 mkdir -p build/installroot/etc/yum/vars
 sudo sh -c "echo $RELEASE >> build/installroot/etc/yum/vars/releasever"
 
 mkdir -p build/offline_installer
 cp dist/offline_installer/redhat/header build/offline_installer
-if [ -n "$RELEASEVER" ]; then
-    YUMOPTS="--releasever=$RELEASEVER"
-fi
-sudo yum -y install $YUMOPTS --downloadonly --installroot=`pwd`/build/installroot --downloaddir=build/offline_installer scylla sudo ntp ntpdate net-tools kernel-tools
+cat <<EOF > build/construct_offline_repo.sh
```
avikivity, line 72: need -e flag so we notice a failure

dist/offline_installer/redhat/build_offline_installer.sh
```diff
@@ -92,22 +56,19 @@ if [ ! -f /usr/bin/createrepo ]; then
...
+yum -y install epel-release wget yum-plugin-downloadonly
+wget $REPO -O /etc/yum.repos.d/scylla.repo
+yum -y install --releasever=$RELEASE --downloadonly --installroot=`pwd`/build/installroot --downloaddir=build/offline_installer scylla sudo ntp ntpdate net-tools kernel-tools
+EOF
+
+sudo ./tools/toolchain/dbuild --image centos:7 -- bash ./build/construct_offline_repo.sh
```
avikivity, line 78: why sudo?

avikivity: Use docker.io/centos:7 so it works better with podman (podman looks for other registries first)

avikivity:
We need to think about the offline installer's place in the build structure.

Option 1: run it during normal build
Advantage: everything is built ahead of time, no surprises
Disadvantage: need huge number of variants, for each major.minor release and for each distribution (centos and rhel are different distributions here). We'd need separate builds for RHEL 7.4, CentOS 7.4, RHEL 8.1, CentOS 8.1, etc.

Option 2: run it after the fact, outside the normal build
Advantage: no need for huge number of variants
Disadvantage: no standard place, no automatic testing

/cc @slivne

slivne:
@syuu1228 / @penberg is this still needed

syuu1228:
@slivne @avikivity @penberg Should we keep this older version of offline installer, which archives all rpm/deb dependencies?
If not, we can drop the script and focus relocatable package based offline installer.

avikivity:
> @slivne @avikivity @penberg Should we keep this older version of offline installer, which archives all rpm/deb dependencies?

Let's split this question into two:

1. do we want to support offline install of .rpm/.deb?

Yes, it's preferable to tarball install. Tarball install should be used only when you don't have root.

2. do we need to have a script that collects the dependencies?

I think there is no need any more, because the new packages have very few dependencies.

> If not, we can drop the script and focus relocatable package based offline installer.

syuu1228:
> I think there is no need any more, because the new packages have very few dependencies.

Yes, we dropped almost all dependencies from rpm.
And if we don't need to collect dependencies because it's very few, I guess we actually don't need this script because that's the only feature of the script.

syuu1228:
Rebased with next, rewritten cross build implementation to support multiple Docker container image, verified to run
CentOS 7/8, Oracle Linux 7.

dist/offline_installer/redhat/build_offline_installer.sh
syuu1228: fixed on new version.

dist/offline_installer/redhat/build_offline_installer.sh
syuu1228: root user was needed to run yum install, but I find a way to avoid taking root privilege here:
 - don't install package on container running time, build custom docker image instead
 - add fakeroot on `yum install --downloadonly`

penberg:
@syuu1228 There commits in the branch, which should not be there. Please rebase against `master`.

penberg (approved):
Looks good to me. @avikivity, please ACK since you started the review round.

@syuu1228 You need to rebase this against `master` to get rid of the extra commits.

syuu1228:
@penberg rebased.

PR #6246, opened 2020-04-20
dist/offline_installer/debian: fix umask error

syuu1228:
same as redhat, makeself script changes current umask, scylla_setup causes
"scylla does not work with current umask setting (0077)" error.
To fix that we need use latest version of makeself, and specfiy --keep-umask
option.

See #6243

### DIFF
--- a/dist/offline_installer/debian/build_offline_installer.sh
+++ b/dist/offline_installer/debian/build_offline_installer.sh
@@ -88,6 +88,11 @@ if [ ! -f /usr/bin/makeself ]; then
     pkg_add makeself
 fi
 
+if ! makeself --help | grep -q keep-umask; then
+    echo "$(makeself --version) is too old, please install 2.4.0 or later"
+    exit 1
+fi
+
 sudo rm -rf build/chroot build/offline_installer build/scylla_offline_installer.sh
 mkdir -p build/chroot
 sudo debootstrap $SUITE build/chroot
@@ -128,4 +133,4 @@ sudo chroot build/chroot env DEBIAN_FRONTEND=noninteractive apt-get install -d -
 mkdir -p build/offline_installer/debs
 cp dist/offline_installer/debian/header build/offline_installer
 cp build/chroot/var/cache/apt/archives/*.deb build/offline_installer/debs
-(cd build; makeself offline_installer scylla_offline_installer.sh "Scylla offline package" ./header)
+(cd build; makeself --keep-umask offline_installer scylla_offline_installer.sh "Scylla offline package" ./header)
--- a/dist/offline_installer/debian/build_offline_installer.sh
+++ b/dist/offline_installer/debian/build_offline_installer.sh
@@ -30,8 +30,27 @@ print_usage() {
     exit 1
 }
 
+is_redhat_variant() {
+    [ -f /etc/redhat-release ]
+}
 is_debian_variant() {
     [ -f /etc/debian_version ]
+}
+
+pkg_add() {
+    if is_redhat_variant; then
+        if [ -x /usr/bin/dnf ]; then
+            sudo dnf -y install "$@"
+        else
+            sudo yum -y install "$@"
+        fi
+    elif is_debian_variant; then
+        sudo apt-get update
+        sudo apt-get install -y "$@"
+    else
+        echo "Unsupported distribution"
+        exit 1
+    fi
 }
 
 REPO=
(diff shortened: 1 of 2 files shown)


dist/offline_installer/debian/build_offline_installer.sh
```diff
@@ -118,4 +115,14 @@ sudo chroot build/chroot env DEBIAN_FRONTEND=noninteractive apt-get install -d -
 mkdir -p build/offline_installer/debs
 cp dist/offline_installer/debian/header build/offline_installer
 cp build/chroot/var/cache/apt/archives/*.deb build/offline_installer/debs
-(cd build; makeself offline_installer scylla_offline_installer.sh "Scylla offline package" ./header)
+cd build
+if [ ! -x makeself-$MAKESELFVER.run ]; then
+    wget https://github.com/megastep/makeself/releases/download/release-$MAKESELFVER/makeself-$MAKESELFVER.run
+    chmod a+x makeself-$MAKESELFVER.run
+fi
```
avikivity, line 122: Why? Fedora 31 has makeself-2.4.0.

avikivity: Let's add it to install-dependencies instead. Please post it as a separate patch, since I need to regenerate the toolchain.

syuu1228: @avikivity this is because build_offline_install.sh only works in target distribution for now, and CentOS7 does not have makeself-2.4.0.
For same reason, it cannot be add on install-dependencies.
We probably can make build script to use docker like dbuild, anyway.

syuu1228: Oh, I mentioned "CentOS7" but it's on Debian/Ubuntu, sorry. Older Debian distribution does not have makeself-2.4.0.

avikivity: Let's run it in dbuild then. I don't want the build system to download random things.

avikivity: I see it uses debootstrap/chroot, so it should be able to run it from Fedora dbuild. We'll have to add debootstrap to install-dependencies too.

penberg: Yeah, we need builds to remain reproducible (think enterprise release lifecycle of many years), and downloading tools breaks this.

syuu1228:
@avikivity @penberg pushed new version, witch supports building offline_installer.sh in Fedora 32, and use makeself-2.4.0 on Fedora 32.
Please review again.

avikivity:
@syuu1228 there are three commits here

syuu1228:
@avikivity "support cross build" is for support Fedora, "support latest distributions" is for adding support of Ubuntu18/Debian10, "fix umask error" is the umask fix itself (requires Fedora, so it actually depends on "support cross build").
"support latest distributions" is not really related with this PR, so I dropped it now, will send in another PR.

syuu1228:
The commit "support latest distributions" moved to https://github.com/scylladb/scylla/pull/6558

dist/offline_installer/debian/build_offline_installer.sh
```diff
@@ -30,12 +30,26 @@ print_usage() {
...
+is_redhat_variant() {
+    [ -f /etc/redhat-release ]
+}
 is_debian_variant() {
     [ -f /etc/debian_version ]
 }
 
+pkg_add() {
+    if is_redhat_variant; then
+        sudo yum -y install $*
```
avikivity, line 42: Please use proper quoting (`"$@"`).

avikivity: Also, yum or dnf?

dist/offline_installer/debian/build_offline_installer.sh
```diff
@@ -30,12 +30,26 @@ print_usage() {
...
+    elif is_debian_variant; then
+        sudo apt-get update
+        sudo apt-get install -y $*
```
avikivity, line 45: Quoting.

dist/offline_installer/debian/build_offline_installer.sh
```diff
@@ -52,30 +66,21 @@ while [ $# -gt 0 ]; do
...
 fi
 
 if [ ! -f /usr/sbin/debootstrap ]; then
-    sudo apt-get -y install debootstrap
+    pkg_add debootstrap
 fi
 
 if [ ! -f /usr/bin/makeself ]; then
-    sudo apt-get -y install makeself
+    pkg_add makeself
```
avikivity, line 83: Should we add makeself to install-dependencies.sh?

dist/offline_installer/debian/build_offline_installer.sh
```diff
@@ -30,12 +30,26 @@ print_usage() {
...
+        sudo apt-get update
+        sudo apt-get install -y $*
+    else
+        echo "Unsupported distribution"
+        exit 1
+    fi
+}
+
 REPO=
-SUITE=
```
avikivity, line 38: Unrelated?

dist/offline_installer/debian/build_offline_installer.sh
```diff
@@ -82,6 +82,12 @@ fi
 if [ ! -f /usr/bin/makeself ]; then
     pkg_add makeself
 fi
+ 
+makeself_ver=$(makeself --version|cut -d ' ' -f 3|sed -e 's/\.//g')
+if [ $makeself_ver -lt 240 ]; then
```
avikivity, line 87: Nicer: `if makeself --help | grep -q keep-umask; then`

dist/offline_installer/debian/build_offline_installer.sh
avikivity: ?

dist/offline_installer/debian/build_offline_installer.sh
avikivity: I see you did. But let me know when you make changes.

syuu1228:
@avikivity pushed new version, please review it again

avikivity:
Queued as 1e2509ffec632b7332832065eeb39df1fe23180d

PR #6289, opened 2020-04-26
redis: add strlen command

syuu1228:
Add strlen command that returns string length of the key.

see: https://redis.io/commands/strlen

### DIFF
--- a/redis-test/test_strings.py
+++ b/redis-test/test_strings.py
@@ -199,3 +199,32 @@ Georg Nees - schotter, plotter on paper, 1968\. Redis ver\. [0-9]+\.[0-9]+\.[0-9
     assert re.match(pattern2, res)
     res = r.execute_command('LOLWUT', 10, 4, 2)
     assert re.match(pattern3, res)
+
+def test_strlen():
+    r = connect()
+    key1 = random_string(10)
+    val1 = random_string(10)
+    key2 = random_string(10)
+    val2 = random_string(1000)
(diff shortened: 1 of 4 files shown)


redis-test/test_strings.py
```diff
@@ -141,3 +141,18 @@ def test_exists_multiple_existent_key():
...
+    key2 = random_string(10)
+    val2 = random_string(1000)
+    key3 = random_string(10)
+
+    assert r.set(key1, val1) == True
+    assert r.set(key2, val2) == True
+    r.delete(key3)
+    assert r.strlen(key1) == 10
+    assert r.strlen(key2) == 1000
+    assert r.strlen(key3) == 0
```
nyh, line 158: Please add another test of what happens when you have a non-string value (e.g., a list) and you try to run strlen on it. I know what will happen in Scylla (since the list and string values come from a different array, strlen will return 0 exactly like the key is missing) but I want to know if the same happens also in Redis. (If not, we can make this test xfail - and it will remain xfail as long as we keep separate databases for the different types.)

syuu1228:
@nyh added non-string value test, please review again.

(Note: multiple types on redis currently does not implemented. Also see: #6291
#6291 is even not enough to support wrong-type error, we need to feature to type check not to returning incompatible type.
And also need CAS to support for List/Hash commands.)

nyh:
Committed to next: a233b0ab3bbf228f65b5b46a923d42e12a04e069

PR #6292, opened 2020-04-27
build: replace xxhash submodule with OS package

avikivity:
The xxhash library has been packaged by Fedora, so we can use it
instead of carrying the submodule. This reduces allows us to
receive updates as the OS packages are updated. Build time will
not be reduced since it is a header-only library.

xxhash preserves the hash results across versions so rolling
upgrades will still work.

The frozen toolchain is updated with the new package.

Tests: unit (dev)

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1289,10 +1289,12 @@ libs = ' '.join([maybe_static(args.staticyamlcpp, '-lyaml-cpp'), '-latomic', '-l
                  ' -lstdc++fs', ' -lcrypt', ' -lcryptopp', ' -lpthread',
                  maybe_static(args.staticboost, '-lboost_date_time -lboost_regex -licuuc'), ])
 
-xxhash_dir = 'xxHash'
+pkgconfig_libs = [
+    'libxxhash',
+]
 
-if not os.path.exists(xxhash_dir) or not os.listdir(xxhash_dir):
-    raise Exception(xxhash_dir + ' is empty. Run "git submodule update --init".')
+args.user_cflags += ' ' + ' '.join([pkg_config(lib, '--cflags') for lib in pkgconfig_libs])
+libs += ' ' + ' '.join([pkg_config(lib, '--libs') for lib in pkgconfig_libs])
 
 if not args.staticboost:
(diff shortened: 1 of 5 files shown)


configure.py
```diff
@@ -1289,10 +1289,12 @@ def configure_zstd(build_dir, mode):
...
                  maybe_static(args.staticboost, '-lboost_date_time -lboost_regex -licuuc'), ])
 
-xxhash_dir = 'xxHash'
+pkgconfig_libs = [
+    'libxxhash',
+]
 
-if not os.path.exists(xxhash_dir) or not os.listdir(xxhash_dir):
-    raise Exception(xxhash_dir + ' is empty. Run "git submodule update --init".')
+args.user_cflags += ' ' + ' '.join([pkg_config(lib, '--cflags') for lib in pkgconfig_libs])
```
nyh, line 1296: What is this change?

avikivity: xxhash_dir no longer exists so we shouldn't check that it exists.
On the other hand, we should use pkgconfig to add the correct cflags and ldflags.

nyh: Oh, nice. I didn't notice you had this new pkgconfig_libs technique. That makes me wonder, why is xxhash special in that regard? Why don't we use pkgconfig also for zlib and a dozen other libraries? Is it because we actually do the same thing differently in cmake instead of here in configure.py? Shouldn't we do these things the same? Are you starting a move back from cmake to configure.py? :-(

avikivity: It's not special. It comes with pkgconfig support so I used that. We should use it for the other libraries too, but that's not part ofthis patch.

nyh: "We should use it for the other libraries too" - I agree, but my point was that we actively took the opposite direction. For example, instead of just using "pkg-config --libs liblz4" in configure.py as we did in the past, we have cmake/Findlz4.cmake which,  in 70 lines of unintelligable mumbo-jumbo basically uses pkg-config too...
Anyway, this patch is fine. I'll commit.

nyh:
Committed to next. Github will not notice, so closing manually.

PR #6313, opened 2020-04-28
dist: redhat: drop dependency on pystache

avikivity:
We use pystache to parametrize our scylla.spec, but pystache is not
present in Fedora 32. Fortunately rpm provides its own template mechanism,
and this patch switches to using it:

 - no longer install pystache
 - pass parameters via rpm "-D" options
 - use 0/1 for conditionals instead of true/false as per rpm conventions
 - sanitize the "product" variable to not contain dashes
 - change the .spec file to use rpm templating: %{...} and %if ... %endif
   instead of mustache templating

### DIFF
--- a/dist/redhat/build_rpm.sh
+++ b/dist/redhat/build_rpm.sh
@@ -10,12 +10,12 @@ print_usage() {
     echo "  --reloc-pkg specify relocatable package path"
     exit 1
 }
-DIST=false
+DIST=0
 RELOC_PKG=
 while [ $# -gt 0 ]; do
     case "$1" in
         "--dist")
-            DIST=true
+            DIST=1
             shift 1
             ;;
         "--target") # This is obsolete, but I keep this in order not to break people's scripts.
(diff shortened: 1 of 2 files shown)


espindola:
LGTM, but out of curiosity, why do we have

 -D"${PRODUCT/-/_} 1"

If I understand the code correctly, that define a variable named scylla
or scylla_enterprise, but I don't see that being used.

Cheers,
Rafael

avikivity:
> LGTM, but out of curiosity, why do we have -D"${PRODUCT/-/_} 1" If I understand the code correctly, that define a variable named scylla or scylla_enterprise, but I don't see that being used. Cheers, Rafael

I'm guessing it's using in enterprise. We should probably drop it, but I don't want to mix it with this change. I'll drop it from enterprise and then from here.

espindola:

>> LGTM, but out of curiosity, why do we have -D"${PRODUCT/-/_} 1" If I understand the code correctly, that define a variable named scylla or scylla_enterprise, but I don't see that being used. Cheers, Rafael
>
> I'm guessing it's using in enterprise. We should probably drop it, but I don't want to mix it with this change. I'll drop it from enterprise and then from here.

Sounds good to me.

Cheers,
Rafael

avikivity:
@penberg @syuu1228 ping

avikivity:
ping...

espindola:
LGTM in case it was not clear.

avikivity:
ping

avikivity:
...ping

nyh:
Thanks. Committed (sorry about the delay...). Github won't know, so closing the pull request manually.

PR #6316, opened 2020-04-29
Repair use table id instead of table name

asias:
Use table_id instead of table_name in row level repair to find a table. It
guarantees we repair the same table even if a table is dropped and a new
table is created with the same name.

Refs: #5942

### DIFF
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -1041,13 +1041,14 @@ static future<> repair_range(repair_info& ri, const dht::token_range& range) {
             neighbors.swap(live_neighbors);
       }
       return ::service::get_local_migration_manager().sync_schema(ri.db.local(), neighbors).then([&neighbors, &ri, range, id] {
-        return do_for_each(ri.cfs.begin(), ri.cfs.end(), [&ri, &neighbors, range] (auto&& cf) {
+        return do_for_each(ri.table_ids.begin(), ri.table_ids.end(), [&ri, &neighbors, range] (utils::UUID table_id) {
+            auto cf = ri.db.local().find_column_family(table_id).schema()->cf_name();
             ri._sub_ranges_nr++;
             if (ri.row_level_repair()) {
                 if (ri.dropped_tables.count(cf)) {
                     return make_ready_future<>();
                 }
-                return repair_cf_range_row_level(ri, cf, range, neighbors).handle_exception_type([&ri, cf] (no_such_column_family&) mutable {
+                return repair_cf_range_row_level(ri, cf, table_id, range, neighbors).handle_exception_type([&ri, cf] (no_such_column_family&) mutable {
                     ri.dropped_tables.insert(cf);
                     return make_ready_future<>();
                 }).handle_exception([&ri] (std::exception_ptr ep) mutable {
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -652,14 +652,14 @@ future<uint64_t> estimate_partitions(seastar::sharded<database>& db, const sstri
 
 static
 const dht::sharder&
-get_sharder_for_tables(seastar::sharded<database>& db, const sstring& keyspace, const std::vector<sstring>& names) {
+get_sharder_for_tables(seastar::sharded<database>& db, const sstring& keyspace, const std::vector<sstring>& table_names, const std::vector<utils::UUID>& table_ids) {
     schema_ptr last_s;
-    for (auto& name : names) {
+    for (size_t idx = 0 ; idx < table_ids.size(); idx++) {
         schema_ptr s;
         try {
-            s = db.local().find_column_family(keyspace, name).schema();
+            s = db.local().find_column_family(table_ids[idx]).schema();
         } catch(...) {
-            throw std::runtime_error(format("No column family '{}' in keyspace '{}'", name, keyspace));
+            throw std::runtime_error(format("No column family '{}' in keyspace '{}'", table_names[idx], keyspace));
         }
         if (last_s && last_s->get_sharder() != s->get_sharder()) {
             throw std::runtime_error(
@@ -672,7 +672,7 @@ get_sharder_for_tables(seastar::sharded<database>& db, const sstring& keyspace, 
     }
     if (!last_s) {
         throw std::runtime_error(format("Failed to find sharder for keyspace={}, tables={}, no table in this keyspace",
-                keyspace, names));
+                keyspace, table_names));
     }
     return last_s->get_sharder();
 }
@@ -687,7 +687,7 @@ repair_info::repair_info(seastar::sharded<database>& db_,
     const std::vector<sstring>& hosts_,
     streaming::stream_reason reason_)
     : db(db_)
-    , sharder(get_sharder_for_tables(db_, keyspace_, cfs_))
+    , sharder(get_sharder_for_tables(db_, keyspace_, cfs_, table_ids_))
     , keyspace(keyspace_)
     , ranges(ranges_)
     , cfs(cfs_)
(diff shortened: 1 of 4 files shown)


repair/repair.cc
```diff
@@ -670,6 +679,7 @@ repair_info::repair_info(seastar::sharded<database>& db_,
     const sstring& keyspace_,
     const dht::token_range_vector& ranges_,
     const std::vector<sstring>& cfs_,
+    const std::vector<utils::UUID>& table_ids_,
```
denesb, line 682: Make this a value

repair/repair.cc
```diff
@@ -679,6 +689,7 @@ repair_info::repair_info(seastar::sharded<database>& db_,
     , keyspace(keyspace_)
     , ranges(ranges_)
     , cfs(cfs_)
+    , table_ids(table_ids_)
```
denesb, line 692: And move it here, to avoid a copy.

repair/repair.cc
```diff
@@ -1459,11 +1470,12 @@ static int do_repair_start(seastar::sharded<database>& db, sstring keyspace,
...
         std::vector<future<>> repair_results;
         repair_results.reserve(smp::count);
+        auto table_ids = get_table_ids(db.local(), keyspace, cfs);
         for (auto shard : boost::irange(unsigned(0), smp::count)) {
-            auto f = db.invoke_on(shard, [&db, keyspace, cfs, id, ranges,
+            auto f = db.invoke_on(shard, [&db, keyspace, cfs, table_ids, id, ranges,
                     data_centers = options.data_centers, hosts = options.hosts] (database& localdb) mutable {
                 auto ri = make_lw_shared<repair_info>(db,
-                        std::move(keyspace), std::move(ranges), std::move(cfs),
+                        std::move(keyspace), std::move(ranges), std::move(cfs), std::move(table_ids),
```
denesb, line 1479: This move has no effect as the constructor takes a `const&`.

repair/repair.cc
```diff
@@ -1036,9 +1036,10 @@ static future<> repair_range(repair_info& ri, const dht::token_range& range) {
       }
       return ::service::get_local_migration_manager().sync_schema(ri.db.local(), neighbors).then([&neighbors, &ri, range, id] {
         return do_for_each(ri.cfs.begin(), ri.cfs.end(), [&ri, &neighbors, range] (auto&& cf) {
+            auto idx = std::distance(ri.cfs.begin(), std::find(ri.cfs.begin(), ri.cfs.end(), cf));
```
denesb, line 1039: Why not make `_cfs` store both the name and the UUID instead of doing this?

asias: I do not want to mess it up with the old partition level repair for backport reasons. Another reason, at some point we might want to store the uuids only.

asias:
Changes in 9eb06c58f5bb827d220a176104be31d1f4570f27 (v2):
- Make repair_info take a value instead of a reference for table_ids

repair/repair.cc
```diff
@@ -670,6 +679,7 @@ repair_info::repair_info(seastar::sharded<database>& db_,
     const sstring& keyspace_,
     const dht::token_range_vector& ranges_,
     const std::vector<sstring>& cfs_,
+    std::vector<utils::UUID> table_ids_,
     int id_,
```
avikivity, line 683: We should only accept the UUID:s, and let the api level do the conversion, this guarantees it is done only once and there is no inconsistency.

repair/repair.cc
```diff
@@ -1036,9 +1036,10 @@ static future<> repair_range(repair_info& ri, const dht::token_range& range) {
       }
       return ::service::get_local_migration_manager().sync_schema(ri.db.local(), neighbors).then([&neighbors, &ri, range, id] {
         return do_for_each(ri.cfs.begin(), ri.cfs.end(), [&ri, &neighbors, range] (auto&& cf) {
+            auto idx = std::distance(ri.cfs.begin(), std::find(ri.cfs.begin(), ri.cfs.end(), cf));
             ri._sub_ranges_nr++;
             if (ri.row_level_repair()) {
-                return repair_cf_range_row_level(ri, cf, range, neighbors);
+                return repair_cf_range_row_level(ri, cf, ri.table_ids[idx], range, neighbors);
```
avikivity, line 1042: Again we should iterate only in table_ids, and if we want the name, grab it from the schema.

repair/repair.cc
asias: We always pass the tables_ids and the tables_name(cfs) to repair_info converted from the api level. It is done once. I see no consistency problem. 

```
auto table_ids = get_table_ids(db.local(), keyspace, cfs)
repair_info(..., tables_ids, cfs)
```

Even if we only pass tables_ids to repair_info, it is much more convenient to store a list of names, we access the table names a lot. So, with you suggestion, we would do

```
auto table_ids = get_table_ids(db.local(), keyspace, cfs)
repair_info(..., tables_ids) {
     cfs = get_table_names_from_table_ids(table_ids)
}
```
It makes little sense to convert it back and forth.

repair/repair.cc
asias: The tables names and ids are converted once, the are fixed 1:1 mapping. I tried to iterate the index but do_for_each does not take do_for_each(boost::irange(0, size)). 

I will change the code to iterate table_ids and find the table name on the fly.

asias:
Changes in 64f5a8a91361d455732b9bed704a95a4531cbf60 (v3):
- Iterate over tables_ids only in repair_range
- Rebase to master

asias:
@avikivity ping.

repair/repair.cc
```diff
@@ -87,6 +87,15 @@ static std::vector<sstring> list_column_families(const database& db, const sstri
     return ret;
 }
 
+static std::vector<utils::UUID> get_table_ids(const database& db, const sstring& keyspace, const std::vector<sstring>& tables) {
+    std::vector<utils::UUID> table_ids;
+    table_ids.reserve(tables.size());
+    for (auto& table : tables) {
```
glommer, line 93: @asias how many tables total can we expect here?
Every time I see a loop I think "stalls".

If this is running in context where we can futurize it (it seems we are), we should be futurizing this loop.

repair/repair.cc
glommer: Are you two in agreement here? It's hard to tell

asias: I am not sure neither. Let's wait for avi to comment.

repair/repair.cc
asias: It is depends on the number of tables in the keyspace. I agree it is better to futurize. I will do that when we clear other comments.

repair/repair.cc
avikivity: I still don't understand it. uuids and table names can drift out of sync when a table is dropped and re-created. It is dangerous to use a table name twice, since the second time might refer to a different table. There is only one safe thing you can do with a table name - convert it to a uuid.

Why do you need the table names?

asias: It is used for logging in multiple places to tell which tables we are working on. After this patch,we do not use table name to find a table object any more.

asias:
Changes in v4: 6c89cedf0a62242337a0c83f0ee2578d00419aee
- Rebase to master
- Make get_table_ids stall-free
- Pass table ids to repair_info only

repair/repair.cc
asias: I agree it is better to pass it in one way. The downside is we need to convert it back and forth. To make progress, I pushed v4 which passes only the table id to repair_info and repair_info does the convert from table_ids to table_names again.

asias:
@avikivity Ping. I have addressed the "pass table id only" issue you raised.  Can you please take another look?

avikivity:
It would be nicer to have a vector<struct { table_uuid, table_name }> instead of two parallel vectors, but can be done later.

avikivity:
Queued.

PR #6327, opened 2020-04-29
reloc-pkg: move all files under project name directory

### LINKED ISSUE #6315: current relocatable packages are not optimized to unify single tarball
Currently, we only have relocatable packages for each individual components, still missing unified relocatable package.
To make unified relocatable package easily, we may want to merge tarballs to single tarball like this:
```zcat *.tar.gz | gzip -c > scylla-unified.tar.xz```
But it's not possible with current relocatable package format, since there are multiple files conflicts, install.sh, SCYLLA-*-FILE, dist/, README.md, etc..

To support this, we need to archive everything in the directory when building relocatable package.
For example, in scylla repo,
```
ar.add('build/SCYLLA-RELEASE-FILE', arcname='SCYLLA-RELEASE-FILE')
ar.add('build/SCYLLA-VERSION-FILE', arcname='SCYLLA-VERSION-FILE')
ar.add('build/SCYLLA-PRODUCT-FILE', arcname='SCYLLA-PRODUCT-FILE')
ar.add('seastar/scripts')
ar.add('install.sh')
```
should be changed to:
```
ar.add('build/SCYLLA-RELEASE-FILE', arcname='scylla/SCYLLA-RELEASE-FILE')
ar.add('build/SCYLLA-VERSION-FILE', arcname='scylla/SCYLLA-VERSION-FILE')
ar.add('build/SCYLLA-PRODUCT-FILE', arcname='scylla/SCYLLA-PRODUCT-FILE')
ar.add('seastar/scripts', arcname='scylla/seaster/scripts')
ar.add('seastar/dpdk/usertools', arcname='scylla/seastar/dpdk/usertools')
ar.add('install.sh', arcname='scylla/install.sh')
```

Also see: https://github.com/scylladb/scylla-pkg/issues/1218

### PR DESCRIPTION
**Pull request by syuu1228**
To make unified relocatable package easily, we may want to merge tarballs to single tarball like this:
zcat *.tar.gz | gzip -c > scylla-unified.tar.xz
But it's not possible with current relocatable package format, since there are multiple files conflicts, install.sh, SCYLLA-*-FILE, dist/, README.md, etc..

To support this, we need to archive everything in the directory when building relocatable package.

Fixes #6315

### DIFF
--- a/dist/redhat/python3/python.spec
+++ b/dist/redhat/python3/python.spec
@@ -23,7 +23,7 @@ functionality). All shared libraries needed for the interpreter to
 operate are shipped with it.
 
 %prep
-%setup -q -c
+%setup -n scylla-python3-package
 
 %install
 ./install.sh --root "$RPM_BUILD_ROOT"
--- a/dist/redhat/scylla.spec
+++ b/dist/redhat/scylla.spec
@@ -44,7 +44,7 @@ This package installs all required packages for ScyllaDB,  including
 %defattr(-,root,root)
 
 %prep
-%setup -c
+%setup -n scylla-package
 
 %package        server
 Group:          Applications/Databases
--- a/reloc/build_deb.sh
+++ b/reloc/build_deb.sh
@@ -36,7 +36,7 @@ RELOC_PKG=$(readlink -f $RELOC_PKG)
 if [[ ! $OPTS =~ --reloc-pkg ]]; then
     OPTS="$OPTS --reloc-pkg $RELOC_PKG"
 fi
-mkdir -p build/debian/scylla-package
-tar -C build/debian/scylla-package -xpf $RELOC_PKG
+mkdir -p build/debian/
+tar -C build/debian/ -xpf $RELOC_PKG
 cd build/debian/scylla-package
 exec ./dist/debian/build_deb.sh $OPTS
--- a/reloc/build_rpm.sh
+++ b/reloc/build_rpm.sh
@@ -52,8 +52,8 @@ RELOC_PKG=$(readlink -f $RELOC_PKG)
 if [[ ! $OPTS =~ --reloc-pkg ]]; then
     OPTS="$OPTS --reloc-pkg $RELOC_PKG"
 fi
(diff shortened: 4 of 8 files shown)

### CODE REVIEW

**Comment by penberg** [comment]
Thanks, looks good to me. @avikivity @glommer, comments?

**Comment by glommer** [comment]
I want to look at two classes of files differently.
Some of them are actual scylla files. As you mentioned, directory names like dist, README, etc

But the others like the `SCYLLA-` files and install.sh, we're just bringing those conflicts on ourselves.
We should not have those files individually in each project's repository. They should come from the supermodule.

As far as I know we have one now. If we move those files to the supermodule, what are we left with?
If we're left with only a few trivial conflicts maybe it'd be better for us to just rename them instead.

**Comment by penberg** [comment]
@glommer First of all, I agree that a super-module will make many things simpler, but it does not fix this issue. There's no way to prevent different components (even with super-module) from generating artifacts with the same name. The most obvious example is `install.sh`. Furthermore, we don't have a super-module right now, so any technical solution to the offline installer tarball unification has to work with what we have now.

**Comment by glommer** [comment]
It looks okay to me to. I have one minor comment, in that all those strings with sub-directory names are a bit hard to maintain and extent. I would prefer that we stopped using `ar.add()`, then, and used instead a wrapper that adds the directory name instead. Something like:

```
def scylla_add(file):
    ar.add(..., arcname="{}/file".format(dir_name))
```

But i am not insisting on it

slivne:
@avikivity we need your help here - its blocking the offline installer

penberg:
@syuu1228 can you please rebase this pull request? Please also have a look at Glauber's suggestion to improving the code.

syuu1228:
@penberg rebased & implemented wrapper function Glauber suggested, please review again

avikivity:
I dequeued this (and corresponding updates in scylla-jmx and scylla-tools-java), it breaks ccm which calls install.sh. So ccm needs to learn about this.

Also, let's change the directory names to scylla, scylla-python3 and scylla-tools-java. No need to include "package" in the name, when unpacked it's no longer a package.

PR #6338, opened 2020-04-30
Support Snapshot of multiple tables

### LINKED ISSUE #6333: Support mutli-table in take snapshot
The take-spanshot command takes multiple keyspaces it would be helpful to support multiple tables as well.

amnonh:
This series adds support for taking a snapshot of multiple tables.

Fixes #6333

### DIFF
--- a/api/storage_service.cc
+++ b/api/storage_service.cc
@@ -992,12 +992,12 @@ void set_snapshot(http_context& ctx, routes& r) {
 
     ss::take_snapshot.set(r, [](std::unique_ptr<request> req) {
         auto tag = req->get_query_param("tag");
-        auto column_family = req->get_query_param("cf");
+        auto column_families = split(req->get_query_param("cf"), ",");
 
         std::vector<sstring> keynames = split(req->get_query_param("kn"), ",");
 
         auto resp = make_ready_future<>();
-        if (column_family.empty()) {
+        if (column_families.empty()) {
             resp = service::get_local_storage_service().take_snapshot(tag, keynames);
         } else {
             if (keynames.empty()) {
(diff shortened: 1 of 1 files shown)


service/storage_service.cc
```diff
@@ -2076,30 +2076,37 @@ future<> storage_service::take_snapshot(sstring tag, std::vector<sstring> keyspa
 
 }
 
-future<> storage_service::take_column_family_snapshot(sstring ks_name, sstring cf_name, sstring tag) {
+future<> storage_service::take_column_family_snapshot(sstring ks_name, std::vector<sstring> tabels, sstring tag) {
```
avikivity, line 2079: s/tabels/tables/

api/storage_service.cc
```diff
@@ -992,12 +992,12 @@ void set_snapshot(http_context& ctx, routes& r) {
 
     ss::take_snapshot.set(r, [](std::unique_ptr<request> req) {
         auto tag = req->get_query_param("tag");
-        auto column_family = req->get_query_param("cf");
+        auto column_families = split(req->get_query_param("cf"), ",");
```
avikivity, line 996: We should have a separate REST endpoint that takes a list.

amnonh: In the new version of the API, definitely, but this is in sync with how the API looks like today for all other column_family commands and it was a request to keep the naming convention

service/storage_service.cc
amnonh: I'll fix and resend

amnonh:
@avikivity I have forced push a new version with the typo fixed

slivne:
@avikivity ping

api/storage_service.cc
avikivity: ok (yuch)

PR #6344, opened 2020-04-30
forbid tmpdir().path()

sitano:
tmpdir is an object that exploits RAII for providing the user
with the temporary directory. however, it is very easy to miss
that it is must not be used as an rvalue for the path() call.

Implementation
==============

I wanted to forbid `tmpdir` to be an rvalue, but there are a lot
of places that use rvalue to lvalue materialization to the future
with `do_with(tmpdir(), [](){...})`.

The implementation deletes `path() &&` leaving only `path() &`.
It also extracts path generation logic to the separate static
method `tmpdir::temp_directory_path()`.

### DIFF
--- a/test/lib/tmpdir.hh
+++ b/test/lib/tmpdir.hh
@@ -42,9 +42,7 @@ private:
     }
 
 public:
-    tmpdir()
-     : _path(fs::temp_directory_path() /
-             fs::path(fmt::format(FMT_STRING("scylla-{}"), utils::make_random_uuid()))) {
+    tmpdir() : _path(tmpdir::temp_directory_path()) {
         fs::create_directories(_path);
     }
 
@@ -60,5 +58,12 @@ public:
         remove();
     }
 
-    const fs::path& path() const noexcept { return _path; }
+    const fs::path& path() const & noexcept { return _path; }
+    // tmpdir().path() must not be allowed, use tmpdir::temp_directory_path() for this.
+    const fs::path& path() const && = delete;
+
+    static const fs::path temp_directory_path() {
+      return fs::temp_directory_path() /
+             fs::path(fmt::format(FMT_STRING("scylla-{}"), utils::make_random_uuid()));
+    }
 };


glommer:
I don't have much of an opinion on it at first glance.
Adding Benny

test/lib/tmpdir.hh
```diff
@@ -60,5 +58,12 @@ public:
         remove();
     }
 
-    const fs::path& path() const noexcept { return _path; }
+    const fs::path& path() const & noexcept { return _path; }
+    // tmpdir().path() must not be allowed, use tmpdir::gen() for this.
+    const fs::path& path() const && = delete;
```
denesb, line 63: I've never seen `= delete()` used like this. Makes sense though as C++ won't let you provide a `const&` overload specification unless there is also an `&&` one.

avikivity: It will allow you to write it so, but it's not helpful - the rvalue *this will decay to const& and call the const& overload.

avikivity: (so the = delete overload is necessary)

test/boost/database_test.cc
```diff
@@ -337,7 +337,7 @@ SEASTAR_THREAD_TEST_CASE(test_distributed_loader_with_pending_delete) {
 future<> do_with_some_data(std::function<future<> (cql_test_env& env)> func) {
     return seastar::async([func = std::move(func)] () mutable {
         auto db_cfg_ptr = make_shared<db::config>();
-        db_cfg_ptr->data_file_directories(std::vector<sstring>({ tmpdir().path().string() }));
+        db_cfg_ptr->data_file_directories(std::vector<sstring>({ tmpdir::gen().string() }));
```
avikivity, line 340: Why not move tmpdir to the thread scope above? Instead of gen() which leaks a path into the filesystem.

test/lib/tmpdir.hh
sitano: I am wondering why
```
const fs::path& path() const & noexcept { return _path; }
```
allows `tmpdir().path()` if `&&` version is not deleted. spec as if says it must work without explicitly deleting `&&`.

test/boost/database_test.cc
sitano: > Instead of gen() which leaks a path into the filesystem.

Doesn't `tmpdir a(); a.path().string()` leak the path the same way?

> Why not move tmpdir to the thread scope above? 

It depends on the intention of the test author. I don't know if the test requires only having a path to the imaginary temporary directory or it requires having a temporary directory with the referenced path.

I only fixed the wrong usage of tmpdir in place.

avikivity: > > Instead of gen() which leaks a path into the filesystem.
> 
> Doesn't `tmpdir a(); a.path().string()` leak the path the same way?

No, because tmpdir's destructor removes the directory.

> 
> > Why not move tmpdir to the thread scope above?
> 
> It depends on the intention of the test author. I don't know if the test requires only having a path to the imaginary temporary directory or it requires having a temporary directory with the referenced path.
> 
> I only fixed the wrong usage of tmpdir in place.

But now running a test will leave a directory in the filesystem.

elcallio: How is setting data_file_directories to an imaginary temp path that does not exist useful? Fixing the test to actually _have_ a temp dir instead of just referencing a name seems quite more useful. Otherwise, if the intention really is to just have an arbitrary name, the original code will do just fine, just create and delete a folder. No need to forbid path(). Quite the opposite.

sitano: > But now running a test will leave a directory in the filesystem.

`tmpdir::gen()` does not create a directory. maybe test does. I will check.

avikivity: Oh, I assumed it does because of the name. Maybe if it was called generate_name() I would be less confused.

sitano: > How is setting data_file_directories to an imaginary temp path that does not exist useful? 

I don't know. We shall ask the test author. It is was he was doing - getting a path of unexisting directory (that was just deleted).

> Fixing the test to actually _have_ a temp dir instead of just referencing a name seems quite more useful. 

IMHO, it only depends on the test semantics.

> Otherwise, if the intention really is to just have an arbitrary name, the original code will do just fine, just create and delete a folder. No need to forbid path(). Quite the opposite.

That's absolutely confusing to me. It is creating and deleting a directory before the test, and only passes the path of deleted directory to the test in the original version. What's the point of allowing that? That exactly what I was trying to protect users of tmpdir against.

That's the main problem with RAII objects - their semantics is not clear from their names.

If the point to fix the test alone, - accepted, I will look into that one.

elcallio: How is a tmpdir object semantics in c++ not obvious? Create a resource - delete a resource. 
That being said, I am fine with the change itself, though
1.) rename gen() - because confusing
2.) make private - should _never_ be called by anyone. That is just borked. 

But more importantly, fix the offending test in a proper way. My guess is that the intention is to have a temp dir present. Nothing else makes sense, because just using a temp name does not guarantee that by the time we get to something reading the config property, the dir was not created by someone else (unlikely, but possible).

bhalevy: why not do something along these lines:
```diff
diff --git a/test/boost/database_test.cc b/test/boost/database_test.cc
index 59858f4659..dd7cf2e068 100644
--- a/test/boost/database_test.cc
+++ b/test/boost/database_test.cc
@@ -336,8 +336,9 @@ SEASTAR_THREAD_TEST_CASE(test_distributed_loader_with_pending_delete) {
 // Snapshot tests and their helpers
 future<> do_with_some_data(std::function<future<> (cql_test_env& env)> func) {
     return seastar::async([func = std::move(func)] () mutable {
+        tmpdir tmpdir_for_data;
         auto db_cfg_ptr = make_shared<db::config>();
-        db_cfg_ptr->data_file_directories(std::vector<sstring>({ tmpdir().path().string() }));
+        db_cfg_ptr->data_file_directories(std::vector<sstring>({ tmpdir_for_data.path().string() }));
         do_with_cql_env_thread([func = std::move(func)] (cql_test_env& e) {
             e.create_table([](std::string_view ks_name) {
                 return schema({}, ks_name, "cf",
```

sitano: > How is a tmpdir object semantics in c++ not obvious? Create a resource - delete a resource.
> That being said, I am fine with the change itself, though
> 1.) rename gen() - because confusing
> 2.) make private - should _never_ be called by anyone. That is just borked.
> 
> But more importantly, fix the offending test in a proper way. My guess is that the intention is to have a temp dir present. Nothing else makes sense, because just using a temp name does not guarantee that by the time we get to something reading the config property, the dir was not created by someone else (unlikely, but possible).

This is a major problem with the C++ code besides cyclic dependencies in the helper objects that are also use RAII. When you look at this form alone:

```
tmpdir()
```

what do you think of it? I tell what - you think it is a function call, and that is the problem. You can understand only by knowing the implementation of this `tmpdir` what is it, how it behave and what are it's limitations. You can't derive from its name or API that it is a RAII object that creates a temp directory in a constructor and then deletes it in a destructor. And that is exactly the reason why it must NOT be a rvalue. But we have _futurizer_ that materializes rvalues into lvalues futures context so the developers are unable to use this to restrict objects API. 

```
gen()
```

I picked `gen` to denote that it is generating random temporary path name. To my deepest regret C++ does not allow to have `path()` class member and static class member overload (`static path()`).

But agree, that another name could be better. 

> 2.) make private - should _never_ be called by anyone. That is just borked.

I don't get it. What is borking?

elcallio: > 
> 
> > How is a tmpdir object semantics in c++ not obvious? Create a resource - delete a resource.
> > That being said, I am fine with the change itself, though
> > 1.) rename gen() - because confusing
> > 2.) make private - should _never_ be called by anyone. That is just borked.
> > But more importantly, fix the offending test in a proper way. My guess is that the intention is to have a temp dir present. Nothing else makes sense, because just using a temp name does not guarantee that by the time we get to something reading the config property, the dir was not created by someone else (unlikely, but possible).
> 
> This is a major problem with the C++ code besides cyclic dependencies in the helper objects that are also use RAII. When you look at this form alone:
> 
> ```
> tmpdir()
> ```
> 
> what do you think of it? I tell what - you think it is a function call, and that is the problem. 

You have a name, you need to determine if it is a type or a function (yes, functions are also types, but ignoring that for the moment). Tough, but believing that languages guard against having to read and understand the tools used is a pipe dream, so I don't think it is that big a problem. 
Code conventions, such as making types capitalized help somewhat, but neither we or stdlib uses this convention, so moot. 

>You can understand only by knowing the implementation of this `tmpdir` what is it, how it behave and what are it's limitations. You can't derive from its name or API that it is a RAII object that creates a temp directory in a constructor and then deletes it in a destructor. 

Yes you can. By being familiar with the conventions of c++ code design. 

> And that is exactly the reason why it must NOT be a rvalue. 

??? It must always be a rvalue. 

> But we have _futurizer_ that materializes rvalues into lvalues futures context so the developers are unable to use this to restrict objects API.

??? futurizer typically operates on move semantics. 

> I picked `gen` to denote that it is generating random temporary path name. To my deepest regret C++ does not allow to have `path()` class member and static class member overload (`static path()`).

Speaking of names that does not immediately explain function.... gen/generation can mean create, but also group etc. `path` would be equally non-descriptive of what it does. And again, it should not really exist, since it never should be called (below). 

> > 2.) make private - should _never_ be called by anyone. That is just borked.
> 
> I don't get it. What is borking?

We should not have the case where anyone needs a "generated" temp path that is not backed by having actually created it, or that can be forgotten (deleted on object destroy).

sitano: > Yes you can. By being familiar with the conventions of c++ code design.

And how? I am familiar with the conventions of C++, Google conventions, and our codebase uses neither of those. What you are saying is literally that you are ok with the API that can be abused.

sitano: > ??? It must always be a rvalue.

I don't understand why you are thinking this.

sitano: > ??? futurizer typically operates on move semantics.

Yes, and that's what is happening in `do_with(tmpdir(), ...)`

sitano: > `path` would be equally non-descriptive of what it does. 

global convention on read-const-methods are the same across all languages:

- getX()
- x()

only Ruby has it abused, but you are not writing Ruby. and original `path` name was not invented by me. but I do think that it is perpect.

> And again, it should not really exist, since it never should be called (below).

I don't understand 1) why 2) how you see this object must be used than

sitano: > We should not have the case where anyone needs a "generated" temp path that is not backed by having actually created it, or that can be forgotten (deleted on object destroy).

This is the only part I can agree with, but the discussion of it is outside of this PR.

sitano: I really can't understand why each of my contributions turns out into a nightmare of discussions like this. We discussing such an obvious and simple thing with a patch of 1 line of code. And still, I find no understanding.

sitano: Considering, you are not approving it, I think it is worth rejection.

PR #6355, opened 2020-05-03
database: remove variadic future from query() and query_mutations()

avikivity:
Variadic futures are deprecated; replace with future<std::tuple<...>>.

Tests: unit (dev)

### DIFF
--- a/database.cc
+++ b/database.cc
@@ -1166,7 +1166,7 @@ compare_atomic_cell_for_merge(atomic_cell_view left, atomic_cell_view right) {
     return 0;
 }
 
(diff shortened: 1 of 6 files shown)


espindola:
LGTM

denesb:
I would say we should also add a type to return from `database::query()`, but we already returned a tuple before, it is just more obvious now.

avikivity:
ping?

avikivity:
@scylladb/scylla-maint ping

psarna:
queued as beaeda5234f062dc96f6

avikivity:
Closing as merged.

PR #6358, opened 2020-05-04
redis: add EX option for set command

syuu1228:
Add EX option for SET command, to set TTL for the key.
A behavior of SET EX is same as SETEX command, it just different syntax.

see: https://redis.io/commands/set


nyh:
Thanks. Committed to next. Closing manually because github won't notice.

Issue #6148, opened 2020-04-07
CommitLog Not Honoring its Input Size


STRATZ-Ken:
It seems like my cluster is not honoring the limit sizes enforced by the config file.    

Example 1 : I left the default commit log size at `-1` upon investigation, I gave the commitlog disk 75GB of space.  I later found after a small bit of time, my commit log file drive size was at 100%.  How is it possible if my max RAM size is 35GB, that the commit log got over 75GB.

Example 2 : I set in the config a max commit log disk size of 30GB.  I am now seeing the same thing as example 1.

I do find, that if I `nodetool drain` and then restart `scylla-server` the entire drive free's up in space back to nearly nothing.

I did some debugging on the drive and there is commitlog files and recycled-commitlog files in the drive that consume 99% of the drive.

I reported this in slack and was informed to make a ticket here.

*Installation details*
Scylla version (or git commit hash): 3.3 Latest 
Cluster size: 1TB
OS (RHEL/CentOS/Ubuntu/AWS AMI): CentOS

*Hardware details (for performance issues)*          Delete if unneeded
Platform (physical/VM/cloud instance type/docker): VM - Centos 8.2
Hardware: vCPU: 14 memory=35000MB
Disks: (SSD/HDD, count) : 1 x SSD Commit Log, 4 x 6 TB HDD data storage

STRATZ-Ken:
Would anyone on the team have a suggestion on how to handle this temporary?  I am using Scylla in a production environment, but I just have two servers crash over night due to the commitlog/hints file directory usage.  I expanded to 250GB hard drives, in attempt to give it more time, but seems that still is not enough.

elcallio:
You can try disabling recycling of segments, which I suspect are not correctly respecting the disk limits, by setting the option:
`commitlog_reuse_segments: False`
in scylla.yaml.

STRATZ-Ken:
Thanks, I will give it a try, but is there anything I should worry about in backlash for setting this command?

elcallio:
It will cause all commitlog segments to be created new, instead of reused. This in turn will make writes slightly less performant. So, yes, you will get overall less commitlog throughput. 

However, I find it very strange that you should get recycled segment in any large excess over the set disk limit. While it is possible that you would get one or two segment extra, it should in general keep to the set max size. 

Can you provide the relevant parts of your scylla.yaml for the node? I.e. the commitlog settings.

STRATZ-Ken:
[...]

PR #6368, opened 2020-05-04
Make commitlog respect disk limit better

elcallio:
Refs #6148

Separates disk usage into two cases: Allocated and used.
Since we use both reserve and recycled segments, both
which are not actually filled with anything at the point
of waiting.

Also refuses to recycle segments or increase reserve size
if our current disk footprint exceeds threshold.

And finally uses some initial heuristics to determine when
we should suggest flushing, based on disk limit, segment
size, and current usage. Right now, when we only have
a half segment left before hitting used == max.

Some initial tests show an improved adherence to limit
though it will still be exceeded, because we do _not_
force waiting for segments to become cleared or similar
if we need to add data, thus slow flushing can still make
usage create extra segments. We will however attempt to
shrink disk usage when load is lighter.

Somewhat unclear how much this impacts performance
with tight limits, and how much this matters.

v2:
* Add some comments/explanations
v3:
* Made disk footprint subtract happen post delete (non-optimistic)

### DIFF
--- a/db/commitlog/commitlog.cc
+++ b/db/commitlog/commitlog.cc
@@ -202,6 +202,7 @@ public:
     // Divide the size-on-disk threshold by #cpus used, since we assume
     // we distribute stuff more or less equally across shards.
     const uint64_t max_disk_size; // per-shard
+    const uint64_t disk_usage_threshold;
 
     bool _shutdown = false;
     std::optional<shared_promise<>> _shutdown_promise = {};
@@ -258,8 +259,10 @@ public:
         uint64_t segments_destroyed = 0;
         uint64_t pending_flushes = 0;
         uint64_t flush_limit_exceeded = 0;
-        uint64_t total_size = 0;
         uint64_t buffer_list_bytes = 0;
+        // size on disk, actually used - i.e. containing data (allocate+cycle)
+        uint64_t active_size_on_disk = 0;
+        // size allocated on disk - i.e. files created (new, reserve, recycled)
         uint64_t total_size_on_disk = 0;
         uint64_t requests_blocked_memory = 0;
     };
@@ -326,6 +329,7 @@ public:
     future<> do_pending_deletes();
 
     future<> delete_segments(std::vector<sstring>);
+    future<> delete_file(const sstring&);
 
     void discard_unused_segments();
     void discard_completed_segments(const cf_id_type&);
@@ -444,6 +448,7 @@ class db::commitlog::segment : public enable_shared_from_this<segment>, public c
 
     uint64_t _file_pos = 0;
     uint64_t _flush_pos = 0;
+    uint64_t _size_on_disk = 0;
 
     bool _closed = false;
     // Not the same as _closed since files can be reused
@@ -503,10 +508,11 @@ public:
     // TODO : tune initial / default size
     static constexpr size_t default_size = align_up<size_t>(128 * 1024, alignment);
 
-    segment(::shared_ptr<segment_manager> m, descriptor d, file && f)
+    segment(::shared_ptr<segment_manager> m, descriptor&& d, file&& f, uint64_t initial_disk_size)
             : _segment_manager(std::move(m)), _desc(std::move(d)), _file(std::move(f)),
-        _file_name(_segment_manager->cfg.commit_log_location + "/" + _desc.filename()), _sync_time(
-                    clock_type::now()), _pending_ops(true) // want exception propagation
+        _file_name(_segment_manager->cfg.commit_log_location + "/" + _desc.filename()), 
+        _size_on_disk(initial_disk_size),
+        _sync_time(clock_type::now()), _pending_ops(true) // want exception propagation
     {
         ++_segment_manager->totals.segments_created;
         clogger.debug("Created new segment {}", *this);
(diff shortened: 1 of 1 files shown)


db/commitlog/commitlog.cc
```diff
@@ -258,8 +259,8 @@ class db::commitlog::segment_manager : public ::enable_shared_from_this<segment_
         uint64_t segments_destroyed = 0;
         uint64_t pending_flushes = 0;
         uint64_t flush_limit_exceeded = 0;
-        uint64_t total_size = 0;
         uint64_t buffer_list_bytes = 0;
+        uint64_t active_size_on_disk = 0;
         uint64_t total_size_on_disk = 0;
```
gleb-cloudius, line 264: So what is the difference between total and active? Can we have a comment describing which is which? may be rename them to be more clear?

elcallio: active == used by commitlog data, total == allocated (including unused, reserve and recycled)
but yes, needs commenting. And maybe better names in fact.

kostja:
* Calle Wilund <notifications@github.com> [20/05/04 21:18]:

I reported https://github.com/scylladb/scylla/issues/6314

Please close it as a duplicate if it's the same issue.

> Refs #6148
> 
> _Not for merging before comments, opinions and better measuring/benchmarking._ 
> 
> Separates disk usage into two cases: Allocated and used.
> Since we use both reserve and recycled segments, both
> which are not actually filled with anything at the point
> of waiting.
> 
> Also refuses to recycle segments or increase reserve size
> if our current disk footprint exceeds threshold.
> 
> And finally uses some initial heuristics to determine when
> we should suggest flushing, based on disk limit, segment
> size, and current usage. Right now, when we only have
> a half segment left before hitting used == max.
> 
> Some initial tests show an improved adherence to limit
> though it will still be exceeded, because we do _not_
> force waiting for segments to become cleared or similar
> if we need to add data, thus slow flushing can still make
> use create extra segments. We will however attempts to
> shrink disk usage when load is lighter.
> 
> Somewhat unclear how much this impacts performance
> with tight limits, and how much this matters.
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/scylladb/scylla/pull/6368
> 
> -- Commit Summary --
> 
>   * commitlog: Demote buffer write log messages to trace
>   * commitlog: Make commitlog respect disk limit better
> 
> -- File Changes --
> 
>     M db/commitlog/commitlog.cc (79)
> 
> -- Patch Links --
> 
> https://github.com/scylladb/scylla/pull/6368.patch
> https://github.com/scylladb/scylla/pull/6368.diff
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/scylladb/scylla/pull/6368

-- 
https://scylladb.com

elcallio:
Updated with some more comments

slivne:
@gleb-cloudius can you please re-review

db/commitlog/commitlog.cc
```diff
@@ -1593,6 +1621,15 @@ void db::commitlog::segment_manager::add_file_to_close(file f) {
     _files_to_close.emplace_back(std::move(f));
 }
 
+future<> db::commitlog::segment_manager::delete_file(const sstring& filename) {
+    return seastar::file_size(filename).then([this, filename](uint64_t size) {
+        // optimistic. because we don't want fails to prevent using a new segment
+        totals.total_size_on_disk -= size; 
+        clogger.debug("Deleting segment file {}", filename);
+        return commit_io_check(&seastar::remove_file, filename);
```
gleb-cloudius, line 1629: Can this fail after we decreased total_size_on_disk above?

raphaelsc:
should we merge this? last comment is an ack from Gleb.

db/commitlog/commitlog.cc
```diff
@@ -1036,6 +1049,8 @@ db::commitlog::segment_manager::segment_manager(config c)
     , max_size(std::min<size_t>(std::numeric_limits<position_type>::max(), std::max<size_t>(cfg.commitlog_segment_size_in_mb, 1) * 1024 * 1024))
     , max_mutation_size(max_size >> 1)
     , max_disk_size(size_t(std::ceil(cfg.commitlog_total_space_in_mb / double(smp::count))) * 1024 * 1024)
+    // our threshold for trying to force a flush. needs heristics, for now max - segment_size/2.
+    , disk_usage_threshold(max_disk_size - (max_disk_size > (max_size/2) ? (max_size/2) : 0))
```
avikivity, line 1053: I think we'll also want a hard limit, beyond which we block and wait for space to be freed. Can come later.

db/commitlog/commitlog.cc
avikivity: Agree, should adjust after remove is suceessful (note that we'll take the node down if remove fails, but still).

avikivity:
Looks reasonable to me, left a minor comment agreeing with @gleb-cloudius

elcallio:
Addressed comment and removed the "RFC".

avikivity:
@gleb-cloudius please re-review

db/commitlog/commitlog.cc
```diff
@@ -786,6 +794,11 @@ class db::commitlog::segment : public enable_shared_from_this<segment>, public c
                 });
             }).finally([this, buf = std::move(buf), size] {
                     _segment_manager->notify_memory_written(size);
+                _segment_manager->totals.buffer_list_bytes -= buf.size_bytes();
+                if (_size_on_disk < _file_pos) {
+                    _segment_manager->totals.total_size_on_disk += (_file_pos - _size_on_disk);
+                    _size_on_disk = _file_pos;
+                }
```
gleb-cloudius, line 801: Is everything OK here with the indentation?

avikivity:
@elcallio what are the new changes?

elcallio:
Indentation, pessimistic subtract of footprint and rebased.

PR #6373, opened 2020-05-05
test: lib: exception_utils: fix crash with fmt-6.2.0

avikivity:
fmt, the formatting library we use, detects types with conversion
to std::string_view (and formats them as strings) and types that
support operator<<(std::ostream, const T&) (and performs custom
formatting on them). However, if <fmt/ostream.h>, the latter is
not done.

The problem happens with seastar::sstring, which implements both,
and debug mode, which disables inlining. Some translation units
do include <fmt/ostream.h>, and so generate code to do custom
formatting. exception_utils.cc doesn't, and so generates code
to format via string_view conversion. At link time, the
compiler picks one of the generated functions and includes it
in the final binary; it happened to pick one generated outside
exception_utils.cc, using custom formatting.

However, there is also code in fmt to encode which path fmt
chose - string_view or custom. This code is constexpr and so
is evaluated in exception_utils.cc. The result is that the
function to perform formatting of seastar::sstring uses custom
formatting, while the descriptor containing the method used
says it is formatting via string_view. This is enough to cause
a crash.

The problem is limited to debug mode, since in other modes
all this code is inlined, and so is consistent within the
translation unit.

We need a more general fix (hopefully in fmt), but for now a
simple fix is to add the missing include.

Ref https://github.com/fmtlib/fmt/issues/1662

### DIFF
--- a/test/lib/exception_utils.cc
+++ b/test/lib/exception_utils.cc
@@ -23,6 +23,7 @@
 
 #include <boost/test/unit_test.hpp>
 #include <fmt/format.h>
+#include <fmt/ostream.h>
 
 std::function<bool(const std::exception&)> exception_predicate::make(
         std::function<bool(const std::exception&)> check,


espindola:
Don't we need the include in a more central location? From the
description it looks like we have to make sure it is present in all
files.

Another idea is to delete operator<< for sstring. Given that it can be
converted to string_view we can probably do without it, no?

Cheers,
Rafael

avikivity:
> Don't we need the include in a more central location? From the description it looks like we have to make sure it is present in all files. Another idea is to delete operator<< for sstring. Given that it can be converted to string_view we can probably do without it, no? Cheers, Rafael

This is a band-aid while we figure out a better solution. The fmt author is working on a better fix.

avikivity:
> Another idea is to delete operator<< for sstring. Given that it can be converted to string_view we can probably do without it, no?

What if someone does std::cout << an_sstring?

espindola:

>> Don't we need the include in a more central location? From the description it looks like we have to make sure it is present in all files. Another idea is to delete operator<< for sstring. Given that it can be converted to string_view we can probably do without it, no? Cheers, Rafael
>
> This is a band-aid while we figure out a better solution. The fmt author is working on a better fix.

I am tempted to say LGTM, but since the intention is to switch dbuild to
fedora 32, aren't you afraid of having odd crashes that only show up on
next promotion?

Cheers,
Rafael

avikivity:
> Avi Kivity <notifications@github.com> writes:
> > Don't we need the include in a more central location? From the description it looks like we have to make sure it is present in all files. Another idea is to delete operator<< for sstring. Given that it can be converted to string_view we can probably do without it, no? Cheers, Rafael This is a band-aid while we figure out a better solution. The fmt author is working on a better fix.
> I am tempted to say LGTM, but since the intention is to switch dbuild to fedora 32, aren't you afraid of having odd crashes that only show up on next promotion? Cheers, Rafael

All debug tests pass (and only debug mode is vulnerable).

Upstream fix is here: https://github.com/fmtlib/fmt/commit/080e44d0bf9581b27ac655e5ab747e5bb8ab4ccf. I'll work to get it into Fedora.

espindola:
LGTM since a proper fix is on the way and any other solution would be
more disruptive.

Cheers,
Rafael

avikivity:
And here is the proper fix. It will likely take 1-2 weeks to propagate.

https://src.fedoraproject.org/rpms/fmt/pull-request/3

penberg:
Queued as 2b0c317dec9b6f3cba8a40cec7fc9981c3408729 to next.

PR #6383, opened 2020-05-06
Revert "config: Do not enable repair based node operations by default"

asias:
This reverts commit b8ac10c4514d2ff099a288dfa3cb6f938b42d582.

The repair based node operations will be enabled by default in 4.1.
Revert the patch which disables it by default.


slivne:
@avikivity please merge this so we can have repair-based-operations included in 4.1

avikivity:
Queued as 71d0d58f8c

PR #6392, opened 2020-05-07
change AWS URL

lauranovich:
Change URL to one which renders better in docs site

### DIFF
--- a/docs/alternator/alternator.md
+++ b/docs/alternator/alternator.md
@@ -3,8 +3,7 @@
 ## Introduction
 Alternator is a Scylla feature adding compatibility with Amazon DynamoDB(TM).
 DynamoDB's API uses JSON-encoded requests and responses which are sent over
-an HTTP or HTTPS transport. It is described in detail on Amazon's site:
-  https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/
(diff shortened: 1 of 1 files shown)


docs/alternator/alternator.md
```diff
@@ -3,8 +3,7 @@
 ## Introduction
 Alternator is a Scylla feature adding compatibility with Amazon DynamoDB(TM).
 DynamoDB's API uses JSON-encoded requests and responses which are sent over
-an HTTP or HTTPS transport. It is described in detail on Amazon's site:
-  https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/
+an HTTP or HTTPS transport. It is described in detail on [Amazon's site](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/ "AWS API Reference").
```
nyh, line 6: Laura, what does it mean to have two different descriptions of the same URL - "Amazon's site" and "AWS API Reference"? Can we drop the second one? Because this text should also be readable as plain text, and the string "AWS API Reference" doesn't really fit the flow of the text. Maybe you want to change the string "Amazon's site" to be "AWS API Reference"?

annastuchlik:
The link has already been fixed with https://github.com/scylladb/scylladb/pull/8910.
See the published dos: https://docs.scylladb.com/stable/alternator/alternator.html
This PR is now redundant. Closing.

PR #6400, opened 2020-05-08
range_streamer: Handle table of RF 1 in get_range_fetch_map

### LINKED ISSUE #6351: Error when replacing node: Startup failed: std::runtime_error (unable to find sufficient sources for streaming range ...
Seen in multiple dtests at scylla version e44b2826abba0d1c6ae91b7d40f0af112a24517d
I suspect it could be caused by d78fbf7c1677a6d30d9f4c24796147c45b582871
With or without relationship to b8ac10c4514d2ff099a288dfa3cb6f938b42d582

https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/471/artifact/logs-release.2/1588467306499_counter_tests.TestCountersOnMultipleNodes.counter_consistency_node_replace_test/node4.log
```
ERROR 2020-05-03 00:54:56,295 [shard 0] init - Startup failed: std::runtime_error (unable to find sufficient sources for streaming range (9166892199407561264, +inf) in keyspace system_auth)
```

https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/471/artifact/logs-release.2/1588473761239_replace_address_test.TestReplaceAddress.replace_node_no_hibernate_state_test/node4.log
```
ERROR 2020-05-03 02:42:41,124 [shard 0] init - Startup failed: std::runtime_error (Cannot replace token 3444658265712363823 which does not exist!)
```

https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/471/artifact/logs-release.2/1588473890707_replace_address_test.TestReplaceAddress.replace_shutdown_node_test/node4.log
```
ERROR 2020-05-03 02:44:50,626 [shard 0] init - Startup failed: std::runtime_error (unable to find sufficient sources for streaming range (-8679473293557364538, -8617587428664198610] in keyspace system_auth)
```

https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/471/artifact/logs-release.2/1588479514271_update_cluster_layout_tests.TestUpdateClusterLayout.verify_latest_copy_replace_node_test/node4.log
```
ERROR 2020-05-03 04:18:33,992 [shard 0] init - Startup failed: std::runtime_error (unable to find sufficient sources for streaming range (9184757305069671073, +inf) in keyspace system_auth)
```

asias:
After "Make replacing node take writes" series, with repair based node
operations disabled, we saw the replace operation fail like:

```
[shard 0] init - Startup failed: std::runtime_error (unable to find
sufficient sources for streaming range (9203926935651910749, +inf) in
keyspace system_auth)
```
The reason is the system_auth keyspace has default RF of 1. It is
impossible to find a source node to stream from for the ranges owned by
the replaced node.

In the past, the replace operation with keyspace of RF 1 passes, because
the replacing node calls token_metadata.update_normal_tokens(tokens,
ip_of_replacing_node) before streaming. We saw:

```
[shard 0] range_streamer - Bootstrap : keyspace system_auth range
(-9021954492552185543, -9016289150131785593] exists on {127.0.0.6}
```

Node 127.0.0.6 is the replacing node 127.0.0.5. The source node check in
range_streamer::get_range_fetch_map will pass if the source is the node
itself. However, it will not stream from the node itself. As a result,
the system_auth keyspace will not get any data.

After the "Make replacing node take writes" series, the replacing node
calls token_metadata.update_normal_tokens(tokens, ip_of_replacing_node)
after the streaming finishes. We saw:

```
[shard 0] range_streamer - Bootstrap : keyspace system_auth range
(-9049647518073030406, -9048297455405660225] exists on {127.0.0.5}
```

Since 127.0.0.5 was dead, the source node check failed, so the bootstrap
operation.

Ta fix, we ignore the table of RF 1 when it is unable to find a source
node to stream.

Fixes #6351

### DIFF
--- a/dht/range_streamer.cc
+++ b/dht/range_streamer.cc
@@ -91,7 +91,16 @@ range_streamer::get_range_fetch_map(const std::unordered_map<dht::token_range, s
         }
 
         if (!found_source) {
-            throw std::runtime_error(format("unable to find sufficient sources for streaming range {} in keyspace {}", range_, keyspace));
+            auto& ks = _db.local().find_keyspace(keyspace);
+            auto rf = ks.get_replication_strategy().get_replication_factor();
+            // When a replacing node replaces a dead node with keyspace of RF
+            // 1, it is expected that replacing node could not find a peer node
+            // that contains data to stream from.
+            if (_reason == streaming::stream_reason::replace && rf == 1) {
+                logger.warn("Unable to find sufficient sources to stream range {} for keyspace {} with RF = 1 for replace operation", range_, keyspace);
+            } else {
+                throw std::runtime_error(format("unable to find sufficient sources for streaming range {} in keyspace {}", range_, keyspace));
+            }
         }
     }
 
--- a/dht/boot_strapper.cc
+++ b/dht/boot_strapper.cc
@@ -43,15 +43,23 @@
 #include "log.hh"
 #include "db/config.hh"
 #include "database.hh"
+#include "streaming/stream_reason.hh"
 
 static logging::logger blogger("boot_strapper");
 
 namespace dht {
 
-future<> boot_strapper::bootstrap() {
+future<> boot_strapper::bootstrap(streaming::stream_reason reason) {
     blogger.debug("Beginning bootstrap process: sorted_tokens={}", _token_metadata.sorted_tokens());
-
-    auto streamer = make_lw_shared<range_streamer>(_db, _token_metadata, _abort_source, _tokens, _address, "Bootstrap", streaming::stream_reason::bootstrap);
+    sstring description;
+    if (reason == streaming::stream_reason::bootstrap) {
+        description = "Bootstrap";
+    } else if (reason == streaming::stream_reason::replace) {
+        description = "Replace";
+    } else {
+        return make_exception_future<>(std::runtime_error("Wrong stream_reason provided: it can only be replace or bootstrap"));
+    }
+    auto streamer = make_lw_shared<range_streamer>(_db, _token_metadata, _abort_source, _tokens, _address, description, reason);
     streamer->add_source_filter(std::make_unique<range_streamer::failure_detector_source_filter>(gms::get_local_gossiper().get_unreachable_members()));
     auto keyspaces = make_lw_shared<std::vector<sstring>>(_db.local().get_non_system_keyspaces());
     return do_for_each(*keyspaces, [this, keyspaces, streamer] (sstring& keyspace_name) {
--- a/dht/boot_strapper.hh
+++ b/dht/boot_strapper.hh
@@ -41,6 +41,7 @@
 #include "dht/i_partitioner.hh"
 #include <unordered_set>
 #include "database_fwd.hh"
+#include "streaming/stream_reason.hh"
 #include <seastar/core/distributed.hh>
 #include <seastar/core/abort_source.hh>
 
@@ -66,7 +67,7 @@ public:
         , _token_metadata(tmd) {
     }
 
-    future<> bootstrap();
+    future<> bootstrap(streaming::stream_reason reason);
 
     /**
      * if initialtoken was specified, use that (split on comma).
--- a/dht/range_streamer.hh
+++ b/dht/range_streamer.hh
@@ -146,7 +146,7 @@ private:
      *                      here, we always exclude ourselves.
      * @return
      */
-    static std::unordered_map<inet_address, dht::token_range_vector>
+    std::unordered_map<inet_address, dht::token_range_vector>
     get_range_fetch_map(const std::unordered_map<dht::token_range, std::vector<inet_address>>& ranges_with_sources,
                         const std::unordered_set<std::unique_ptr<i_source_filter>>& source_filters,
                         const sstring& keyspace);
--- a/idl/streaming.idl.hh
+++ b/idl/streaming.idl.hh
@@ -49,6 +49,7 @@ enum class stream_reason : uint8_t {
     removenode,
     rebuild,
     repair,
+    replace,
 };
 
 enum class stream_mutation_fragments_cmd : uint8_t {
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -1945,7 +1945,7 @@ future<> rebuild_with_repair(seastar::sharded<database>& db, locator::token_meta
 future<> replace_with_repair(seastar::sharded<database>& db, locator::token_metadata tm, std::unordered_set<dht::token> replacing_tokens) {
     auto op = sstring("replace_with_repair");
     auto source_dc = get_local_dc();
-    auto reason = streaming::stream_reason::bootstrap;
+    auto reason = streaming::stream_reason::replace;
     tm.update_normal_tokens(replacing_tokens, utils::fb_utilities::get_broadcast_address());
     return do_rebuild_replace_with_repair(db, std::move(tm), std::move(op), std::move(source_dc), reason);
 }
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -973,7 +973,11 @@ void storage_service::bootstrap() {
     } else {
         dht::boot_strapper bs(_db, _abort_source, get_broadcast_address(), _bootstrap_tokens, _token_metadata);
         // Does the actual streaming of newly replicated token ranges.
-        bs.bootstrap().get();
+        if (db().local().is_replacing()) {
+            bs.bootstrap(streaming::stream_reason::replace).get();
+        } else {
+            bs.bootstrap(streaming::stream_reason::bootstrap).get();
+        }
     }
     _db.invoke_on_all([this] (database& db) {
         for (auto& cf : db.get_non_system_column_families()) {
--- a/streaming/stream_reason.hh
+++ b/streaming/stream_reason.hh
@@ -32,6 +32,7 @@ enum class stream_reason : uint8_t {
     removenode,
     rebuild,
     repair,
+    replace,
 };
 
 }


dht/range_streamer.cc
```diff
@@ -91,7 +91,13 @@ range_streamer::get_range_fetch_map(const std::unordered_map<dht::token_range, s
         }
 
         if (!found_source) {
-            throw std::runtime_error(format("unable to find sufficient sources for streaming range {} in keyspace {}", range_, keyspace));
+            auto& ks = _db.local().find_keyspace(keyspace);
+            auto rf = ks.get_replication_strategy().get_replication_factor();
+            if (rf == 1) {
```
avikivity, line 96: This needs a comment, it is not trivial.

Also, should this be qualified for replace node operations only? For bootstrap/decommission, it means the a node is down, and we will end up with unnecessary data loss. If the node comes back up, we can bootstrap/decomission without data loss.

asias: I will add a comment. 

The "abort the operation and bring back the node" applies to replace operation as well. E.g, we can avoid data lose to abort the replace operation and bring back the dead node then do the replace operation. 

Glauber had the argument that if  the rf  is 1 and a node is down, it is fine to lose data. 

The safest way is to abort the operation and let user decide what to do with the keyspace with rf one. I am not sure we want to do this.

glommer: No asias, I didn't have the argument that if the *source* node is down it is fine to lose data, which seems to be Avi's concern.

What I said is that if we are using RF=1, and you lost a node, then it is fine to accept data loss in this case: no need to error out.

My understanding about what @avikivity is saying is this:
- let's say you are bootstrapping a new node into an RF=1 keyspace.
- Let's say one of your nodes is temporarily down while you bootstrap.

You will not find a source node and bail, right? In that case you'll end up with unnecessary data loss.

asias: I understand the bootstrap case. My point is even if with replace operation, we could avoid unnecessary data loss if we abort the replace operation and user can fix the dead node and bring it back.  It is not obvious for the user that she/he is going to lose data if she/he performs the replace operation with rf one keyspace. We do not even give a big warning about this and give a chance for the user to reconsider.

asias: So, to summary, I think we have two choices: 

Choice 1) 
Abort if no sufficient source are found. Ask user to decide if she/he wants to proceed the operation. A new config option is needed. 

Choice 2)
Do not abort  if no sufficient source and  rf == 1 and operation == replacing

What do yo think?

glommer: Asias, there is no way to prevent data loss in a replace operation with RF=1.
The data is already gone , so what is the user supposed to do?
You will error out, and then what?
You will add an option, and then what? If the user decides not to proceed because of data loss, is the cluster dead forever?

replace with RF=1 should keep going - because the data is already lost.
bootstrap or decommission with RF=1 should not, because the data may not be lost yet.
Depends on whether the failure is temporary or permanent.

asias: Not really, the dead node might still have the data. For example, the network is broken, we can fix the network and bring back the dead node, to avoid the replace operation.

glommer: @asias the replace operation is user-initiated.
If the user initiated the operation, this means they have already decided the node is not recoverable and the data is lost from the point of view of the cluster.

asias: They might want to reconsider the decision if they know there is a data lose. They might forget there is a keyspace with rf one if there  are a lot of keyspaces or not be aware the replace operation will lose data at all.

glommer: Asias, you are overthinking this.

Let's say the user reconsiders the decision now that they know there is a keyspace with RF=1.
What do they do? go back in time?

For the common case of replace, the node is dead already. There is nothing the user can possibly do.

You are proposing making this more complex - with flags and such - because there is a possibility (which I never seen) that the node is not really dead.

slivne: @glommer the user can upload data from a backup if he chooses todo that.

slivne:
@glommer / @avikivity / @asias we need to move with this - its blocking repair-based-operations - as with regular streaming nodes do start AFAIK.

dht/range_streamer.cc
glommer: then what difference does it make if the data was streamed to the new node or not?

Look: from the point of view of streaming, if you are RF=1 and you lost a node, there is nothing to replace. What is done about that later has nothing to do with replace.

glommer:
Agree shlomi, but this discussion is currently is not productive.
The semantics are clear: on replace, RF=1 means the data is already lost.
On bootstrap the donor node can come back online, so we should be retrying and not ignoring.

Let's not overcomplicate this

vladzcloudius:
@glommer @slivne @avikivity @asias 
I think we should definitely fail replace if there are tables with RF=1. For the exactly same reason as we want to fail it if there are two nodes down and they are both replicas for the same token.

If there is a data loss possibility we should inform the user about it - otherwise all this repair-based effort is worth little.

We should also have an option to force the replace: a parameter in scylla.yaml would do nicely.

Think of the scenario when we lost two nodes. With today's master you won't be able to replace any of them and that's actually good although it's a regression and it's annoying.

One of the main benefits  of having a repair-based streaming is not to have to run a repair after operations like 'replace' or 'rebuild' and if there are two nodes DOWN and if these nodes are both replicas for the same token then there is a chance that there was a data loss.

In this case we should at least inform the user about this and failing the streaming is a good and easy way to do so - with an appropriate scary warning in the log of course.

Once we do the user would decide whether he/she wants to proceed using that extra parameter that we'll add or if he/she should go ahead and start restoring the data from the backup.

vladzcloudius:
@avikivity @slivne @asias @glommer 
The problem is more generic: think of the case where you have a KS with RF=3, minimum write CL=2 (QUORUM), two nodes DOWN and these two nodes are replica for the same token (e.g. they are in different RACKs in the same DC).

This means that there is a possible data loss situation - similar to what's discussed here for a table with RF=1.

I think that in each such case (including the one discussed here) we should inform the user about the problem and then let the user decide what to do:
  * Start restoring the data from the backup.
  * Force rebuild despite the data loss possibility: we should add a parameter to allow that.

Failing the streaming is an easy and obvious way to do so (inform the user). 
In order to keep things simple we should also scan all tables before we start streaming and report all tables where the data loss may have occurred.

In order to be able to reliably detect the data loss case we should also know the minimum CL used for writes for each table.
The best would be to keep track of that automagically and store it in some system_distributed.xxx table on a per-table basis.

If we have all this the use case described in this issue would also be covered.

With today's master if we loose two nodes and of they are replicas for the same token we won't be able to use replace procedure for any of them.

This is a regression and we need to address this better sooner than later.

glommer:
> @avikivity @slivne @asias @glommer
> The problem is more generic: think of the case where you have a KS with RF=3, minimum write CL=2 (QUORUM), two nodes DOWN and these two nodes are replica for the same token (e.g. they are in different RACKs in the same DC).
> 
> This means that there is a possible data loss situation - similar to what's discussed here for a table with RF=1.

The model tolerates RF-1 failures. That's 2 for  RF=3, and that's 0 for RF=1. The user makes a choice of what they can tolerate when the RF is chosen. Nothing new here.

> 
> If the minimum write CL=1 then even with a single node DOWN we are in a data loss case.

Once more, nothing new here. There is nothing we can do in this situation and adding flags won't make anything better. It will just overcomplicate things.

As you noticed yourself, this actually depends on the consistency level. So now we need to take into account something we have no idea about - because it's per request - when we are replacing a node.

Since a request with CL=LOCAL_ONE is always possible, that means that there is always the possibility of data loss.

So for any replace to succeed this flag must always be set. What's the point of the flag, then ?

> 
> I think that in each such case (including the one discussed above) we should inform the user about the problem and then let the user decide what to do:
> 
> * Start restoring the data from the backup.
> * Force rebuild despite the data loss possibility: we should add a parameter to allow that.

Absolutely not. Data loss is always possible if you have CL=ONE. This means that this flag must always be true for anything to happen. It loses its meaning.

If you see value in informing the user that such situation has happened, we can do that. There are logs for that which I hate but are better than nothing, and Benny has this very nice idea of a system table where relevant system events can be logged.

But why starting overcomplicate things because of a data loss that is part of the model and more importantly: already happened?

> 
> Failing the streaming is an easy and obvious way to do so (inform the user).
> In order to keep things simple we should also scan all tables before we start streaming and report all tables where the data loss may have occurred.
> 
> In order to be able to reliably detect the data loss case we should also know the minimum CL used for writes for each table.

We can't know that. CL is per request. It's impossible to know for sure.

> The best would be to keep track of that automagically and store it in some system_distributed.xxx table on a per-table basis.

It's per request. Impossible.
> 
> If we have all this the use case described in this issue would also be covered.
> 
> With today's master if we loose two nodes and if they are replicas for the same token we won't be able to use replace procedure for any of them.
> 
> This is a regression and we need to address this better sooner than later.

I don't see this regression. Where is the regression ?

vladzcloudius:
> > @avikivity @slivne @asias @glommer
> > The problem is more generic: think of the case where you have a KS with RF=3, minimum write CL=2 (QUORUM), two nodes DOWN and these two nodes are replica for the same token (e.g. they are in different RACKs in the same DC).
> > This means that there is a possible data loss situation - similar to what's discussed here for a table with RF=1.
> 
> The model tolerates RF-1 failures. That's 2 for RF=3, and that's 0 for RF=1. The user makes a choice of what they can tolerate when the RF is chosen. Nothing new here.
> 
> > If the minimum write CL=1 then even with a single node DOWN we are in a data loss case.
> 
> Once more, nothing new here. There is nothing we can do in this situation and adding flags won't make anything better. It will just overcomplicate things.
> 
> As you noticed yourself, this actually depends on the consistency level. So now we need to take into account something we have no idea about - because it's per request - when we are replacing a node.
> 
> Since a request with CL=LOCAL_ONE is always possible, that means that there is always the possibility of data loss.
> 
> So for any replace to succeed this flag must always be set. What's the point of the flag, then ?
> 
> > I think that in each such case (including the one discussed above) we should inform the user about the problem and then let the user decide what to do:
> > 
> > * Start restoring the data from the backup.
> > * Force rebuild despite the data loss possibility: we should add a parameter to allow that.
> 
> Absolutely not. Data loss is always possible if you have CL=ONE. This means that this flag must always be true for anything to happen. It loses its meaning.
> 
> If you see value in informing the user that such situation has happened, we can do that. There are logs for that which I hate but are better than nothing, and Benny has this very nice idea of a system table where relevant system events can be logged.
> 
> But why starting overcomplicate things because of a data loss that is part of the model and more importantly: already happened?
> 
> > Failing the streaming is an easy and obvious way to do so (inform the user).
> > In order to keep things simple we should also scan all tables before we start streaming and report all tables where the data loss may have occurred.
> > In order to be able to reliably detect the data loss case we should also know the minimum CL used for writes for each table.
> 
> We can't know that. CL is per request. It's impossible to know for sure.
> 
> > The best would be to keep track of that automagically and store it in some system_distributed.xxx table on a per-table basis.
> 
> It's per request. Impossible.

It's possible - you need to keep only the smallest value.

However your point is that we don't need to care about the data loss at all and leave it totally to the user. 
But what to do if some replica is temporarily unreachable due to network partitioning? - Should we fail streaming or not? And how you can distinguish such a case from the case when 2 nodes are actually dead? I don't think we can/should decide this on our own but rather let the user decide what to do.

> 
> > If we have all this the use case described in this issue would also be covered.
> > With today's master if we loose two nodes and if they are replicas for the same token we won't be able to use replace procedure for any of them.
> > This is a regression and we need to address this better sooner than later.
> 
> I don't see this regression. Where is the regression ?

Without the repair-based streaming we were able to replace a node when 2 nodes were down if there is at least one replica that can stream data to the replacing node and today - we don't.

glommer:
> > > @avikivity @slivne @asias @glommer
> > > The problem is more generic: think of the case where you have a KS with RF=3, minimum write CL=2 (QUORUM), two nodes DOWN and these two nodes are replica for the same token (e.g. they are in different RACKs in the same DC).
> > > This means that there is a possible data loss situation - similar to what's discussed here for a table with RF=1.
> > 
> > 
> > The model tolerates RF-1 failures. That's 2 for RF=3, and that's 0 for RF=1. The user makes a choice of what they can tolerate when the RF is chosen. Nothing new here.
> > > If the minimum write CL=1 then even with a single node DOWN we are in a data loss case.
> > 
> > 
> > Once more, nothing new here. There is nothing we can do in this situation and adding flags won't make anything better. It will just overcomplicate things.
> > As you noticed yourself, this actually depends on the consistency level. So now we need to take into account something we have no idea about - because it's per request - when we are replacing a node.
> > Since a request with CL=LOCAL_ONE is always possible, that means that there is always the possibility of data loss.
> > So for any replace to succeed this flag must always be set. What's the point of the flag, then ?
> > > I think that in each such case (including the one discussed above) we should inform the user about the problem and then let the user decide what to do:
> > > 
> > > * Start restoring the data from the backup.
> > > * Force rebuild despite the data loss possibility: we should add a parameter to allow that.
> > 
> > 
> > Absolutely not. Data loss is always possible if you have CL=ONE. This means that this flag must always be true for anything to happen. It loses its meaning.
> > If you see value in informing the user that such situation has happened, we can do that. There are logs for that which I hate but are better than nothing, and Benny has this very nice idea of a system table where relevant system events can be logged.
> > But why starting overcomplicate things because of a data loss that is part of the model and more importantly: already happened?
> > > Failing the streaming is an easy and obvious way to do so (inform the user).
> > > In order to keep things simple we should also scan all tables before we start streaming and report all tables where the data loss may have occurred.
> > > In order to be able to reliably detect the data loss case we should also know the minimum CL used for writes for each table.
> > 
> > 
> > We can't know that. CL is per request. It's impossible to know for sure.
> > > The best would be to keep track of that automagically and store it in some system_distributed.xxx table on a per-table basis.
> > 
> > 
> > It's per request. Impossible.
> 
> It's possible - you need to keep only the smallest value.
> 
> However your point is that we don't need to care about the data loss at all and leave it totally to the user.

Not in general. Depending on the data loss.

Some are part of the model.
> But what to do if some replica is temporarily unreachable due to network partitioning? - Should we fail streaming or not?
We should fail
 
> And how you can distinguish such a case from the case when 2 nodes are actually dead? I don't think we can/should decide this on our own but rather let the user decide what to do.

I just explained to Asias over the phone what we should do, and it seems we reached agreement.

Here's an easy way to explain:
There is a replica set with X nodes.
The replacing node removes itself from the replica set.

The nodes that remain are the only ones you test for being reachable. 
If I am RF=1 and you remove yourself , you end up with the empty set.

> 
> > > If we have all this the use case described in this issue would also be covered.
> > > With today's master if we loose two nodes and if they are replicas for the same token we won't be able to use replace procedure for any of them.
> > > This is a regression and we need to address this better sooner than later.
> > 
> > 
> > I don't see this regression. Where is the regression ?
> 
> Without the repair-based streaming we were able to replace a node when 2 nodes were down if there is at least one replica that can stream data to the replacing node and today - we don't.

First, this is not my understanding. As far as I remember the replacing operation will calculate which token to receive data from based on ring positioning, and if the replica you calculated isn't alive you should fail.

But let's assume for a moment I am wrong: this is a different issue then the one we are talking about here. The case we are discussing is what to do if you lost (are replacing) as many replicas as you have RF. For example lost 1 node with RF=1. Since we can only replace one node at a time (for now), this is the only case that matters. But if we could replace 2 nodes at a time, it would be the same for RF=2, replacing two nodes. Or RF=3 replacing 3 nodes. Your data is gone. 

That's very different than having unreachable nodes that you are not replacing.

vladzcloudius:
> > > > @avikivity @slivne @asias @glommer
> > > > The problem is more generic: think of the case where you have a KS with RF=3, minimum write CL=2 (QUORUM), two nodes DOWN and these two nodes are replica for the same token (e.g. they are in different RACKs in the same DC).
> > > > This means that there is a possible data loss situation - similar to what's discussed here for a table with RF=1.
> > > 
> > > 
> > > The model tolerates RF-1 failures. That's 2 for RF=3, and that's 0 for RF=1. The user makes a choice of what they can tolerate when the RF is chosen. Nothing new here.
> > > > If the minimum write CL=1 then even with a single node DOWN we are in a data loss case.
> > > 
> > > 
> > > Once more, nothing new here. There is nothing we can do in this situation and adding flags won't make anything better. It will just overcomplicate things.
> > > As you noticed yourself, this actually depends on the consistency level. So now we need to take into account something we have no idea about - because it's per request - when we are replacing a node.
> > > Since a request with CL=LOCAL_ONE is always possible, that means that there is always the possibility of data loss.
> > > So for any replace to succeed this flag must always be set. What's the point of the flag, then ?
> > > > I think that in each such case (including the one discussed above) we should inform the user about the problem and then let the user decide what to do:
> > > > 
> > > > * Start restoring the data from the backup.
> > > > * Force rebuild despite the data loss possibility: we should add a parameter to allow that.
> > > 
> > > 
> > > Absolutely not. Data loss is always possible if you have CL=ONE. This means that this flag must always be true for anything to happen. It loses its meaning.
> > > If you see value in informing the user that such situation has happened, we can do that. There are logs for that which I hate but are better than nothing, and Benny has this very nice idea of a system table where relevant system events can be logged.
> > > But why starting overcomplicate things because of a data loss that is part of the model and more importantly: already happened?
> > > > Failing the streaming is an easy and obvious way to do so (inform the user).
> > > > In order to keep things simple we should also scan all tables before we start streaming and report all tables where the data loss may have occurred.
> > > > In order to be able to reliably detect the data loss case we should also know the minimum CL used for writes for each table.
> > > 
> > > 
> > > We can't know that. CL is per request. It's impossible to know for sure.
> > > > The best would be to keep track of that automagically and store it in some system_distributed.xxx table on a per-table basis.
> > > 
> > > 
> > > It's per request. Impossible.
> > 
> > 
> > It's possible - you need to keep only the smallest value.
> > However your point is that we don't need to care about the data loss at all and leave it totally to the user.
> 
> Not in general. Depending on the data loss.
> 
> Some are part of the model.
> 
> > But what to do if some replica is temporarily unreachable due to network partitioning? - Should we fail streaming or not?
> > We should fail

You won't be able to distinguish this case from the case when two nodes are dead. And in the later case users may decide that they want to rebuild the node because they have enough other replicas.

> 
> > And how you can distinguish such a case from the case when 2 nodes are actually dead? I don't think we can/should decide this on our own but rather let the user decide what to do.
> 
> I just explained to Asias over the phone what we should do, and it seems we reached agreement.
> 
> Here's an easy way to explain:
> There is a replica set with X nodes.
> The replacing node removes itself from the replica set.
> 
> The nodes that remain are the only ones you test for being reachable.
> If I am RF=1 and you remove yourself , you end up with the empty set.

See above - this proposal is not going to resolve the two-nodes-dead use case when the user DO wants for the replace procedure to stream from available nodes.

> 
> > > > If we have all this the use case described in this issue would also be covered.
> > > > With today's master if we loose two nodes and if they are replicas for the same token we won't be able to use replace procedure for any of them.
> > > > This is a regression and we need to address this better sooner than later.
> > > 
> > > 
> > > I don't see this regression. Where is the regression ?
> > 
> > 
> > Without the repair-based streaming we were able to replace a node when 2 nodes were down if there is at least one replica that can stream data to the replacing node and today - we don't.
> 
> First, this is not my understanding. As far as I remember the replacing operation will calculate which token to receive data from based on ring positioning, and if the replica you calculated isn't alive you should fail.

I rely of Asias here. According to him not alive replicas are going to be filtered out.

> 
> But let's assume for a moment I am wrong: this is a different issue then the one we are talking about here. The case we are discussing is what to do if you lost (are replacing) as many replicas as you have RF. For example lost 1 node with RF=1. Since we can only replace one node at a time (for now), this is the only case that matters. But if we could replace 2 nodes at a time, it would be the same for RF=2, replacing two nodes. Or RF=3 replacing 3 nodes. Your data is gone.

I agree, when we lost RF or more replicas then there is no much dilemma here. 
However if the number of lost replicas is less then RF - then there is.
Assuming we would allow user to choose what to do in the later case it would be rather confusing if we just let the replace to unconditionally continue for the former case.

> 
> That's very different than having unreachable nodes that you are not replacing.

asias:
Changes in 81f0260816410b0c2f69118878ca60fbdcb16776:
- Added patch "streaming: Use separate streaming reason for replace operation"
- Relax the check only for RF == 1 && replace operations

slivne:
@asias / @glommer / @vladzcloudius this is blocking 4.1 can we resolve this discussion

For 4.1: 
- we need to make this as good/correct as it currently works on 4.0 or older releases (even if we think there is a potential fault)
- we should not introduce regressions

If we have left overs items to discuss -  lets do it outside of this pull request on a seperate issue.

this is currently blocking 4.1 release

glommer:
@asias I thought we had agreed on how to proceed here:
- during replace RF=1 should just keep moving.
- During add / remove, it should not.

In general the algorithm is that the set of nodes that are being replaced should be removed from the expected replica set.

We can proceed with an empty expected replica set. We cannot proceed when the replica set is *not* empty but the nodes in it do not respond.

asias:
> @asias I thought we had agreed on how to proceed here:
> 
>     * during replace RF=1 should just keep moving.
> 
>     * During add / remove, it should not.

This is what the latest version of this patch does: https://github.com/scylladb/scylla/pull/6400#issuecomment-632438674

> 
> 
> In general the algorithm is that the set of nodes that are being replaced should be removed from the expected replica set.
> 
> We can proceed with an empty expected replica set. We cannot proceed when the replica set is _not_ empty but the nodes in it do not respond.

Please note, this is old stream based operation, so the replacing node only need to find one of the source replicas. Node being replaced is never being considered as source replicas (Because it is dead, that's why we want to replace it).  For a given vnode range, number of  possible source replica nodes X =  RF - Number of dead nodes belongs to the replicas sets. For example, with RF = 1. For a range that the dead node is the owner, X = 1 -1 = 0. For a range that the dead node is not the owner, X = 1 - 0 = 1.

asias:
> @asias / @glommer / @vladzcloudius this is blocking 4.1 can we resolve this discussion
> 
> For 4.1:
> 
>     * we need to make this as good/correct as it currently works on 4.0 or older releases (even if we think there is a potential fault)
> 
>     * we should not introduce regressions

The last version I sent https://github.com/scylladb/scylla/pull/6400#issuecomment-632438674, did exactly this. I am waiting for maintainers to review and merge.

> 
> If we have left overs items to discuss - lets do it outside of this pull request on a seperate issue.
> 
> this is currently blocking 4.1 release

glommer:
ok, thanks. Last comment from @avikivity asking for a change is just a dot, so I am just assuming he did that by mistake.

@avikivity can you take a look at this last version? This is a 4.1 blocker

PR #6408, opened 2020-05-09
dist/redhat/python3: drop dependency on pystache

syuu1228:
Same as dist/redhat, stop using mustache since pystache is no longer available
on Fedora 32.

see: https://github.com/scylladb/scylla/pull/6313


avikivity:
Queued as 4d957eeda

avikivity:
@syuu1228 please do debian as well. You can use `DEB_BUILD_OPTIONS="var1=value1 var2=value2"`, see https://www.debian.org/doc/debian-policy/ch-source.html

PR #6415, opened 2020-05-11
dist/debian: drop dependency on pystache

syuu1228:
Same as 9d91ac345ab64a203e5019144ec01c36b3aadc98, drop dependency on pystache
since it nolonger present in Fedora 32.

To implement it, simplified debian package build process.
It will be generate debian/ directory when building relocatable package,
we just need to run debuild using the package.

To generate debian/ directory this commit added debian_files_gen.py,
it construct whole directory including control and changelog files
from template files.
Since we need to stop pystache, these template files swiched to
string.Template class which is included python3 standard library.

see: https://github.com/scylladb/scylla/pull/6313

### DIFF
--- a/dist/debian/debian/rules
+++ b/dist/debian/debian/rules
@@ -1,13 +1,16 @@
 #!/usr/bin/make -f
 
+include /usr/share/dpkg/pkg-info.mk
+
 export PYBUILD_DISABLE=1
 jobs := $(shell echo $$DEB_BUILD_OPTIONS | sed -r "s/.*parallel=([0-9]+).*/-j\1/")
-{{#housekeeping}}
-install_arg := --housekeeping
-{{/housekeeping}}
-{{^housekeeping}}
-install_arg :=
-{{/housekeeping}}
+ifneq ($(findstring housekeeping, $(DEB_BUILD_OPTIONS)),)
+    install_arg := --housekeeping
+else
+    install_arg :=
+endif
+
+product := $(subst -server,,$(DEB_SOURCE))
 
 override_dh_auto_configure:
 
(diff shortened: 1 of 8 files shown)


avikivity:
Why can't we use DEB_BUILD_OPTIONS without a template?

avikivity:
I'd like to make .spec and rules more freestanding so they can be invoked without scripts.

syuu1228:
@avikivity debian/rules is now untemplatized and switched to use DEB_BUILD_OPTIONS so that is fine.
But debian/control and debian/changelog are not take any custom variables so still use template on this PR.
But there is a hack on debian packages, we actually able to modify debian directory contents on ```debian/rules clean```.

I able to do rename package name by following simple script on debian/rules:
```
clean:
    sed -e 's/scylla/newproduct/g' debian/control > debian/control.tmp
    mv debian/control.tmp debian/control
    sed -e 's/scylla/newproduct/g' debian/changelog > debian/changelog.tmp
    mv debian/changelog.tmp debian/changelog
    dh $@
```
I suppose we can make debian package script more simpler in this way, without template.

syuu1228:
@avikivity template-less version pushed, please review again.

dist/debian/debian/control
```diff
@@ -1,47 +1,47 @@
...
+Source: scylla-server
 Maintainer: Takuya ASADA <syuu@scylladb.com>
 Homepage: http://scylladb.com
 Section: database
 Priority: optional
 X-Python3-Version: >= 3.4
 Standards-Version: 3.9.5
 
-Package: {{product}}-conf
+Package: scylla-conf
```
avikivity, line 9: But we still need to make this work for enterprise. Can `control` not see DEB_BUILD_OPTIONS?

dist/debian/debian/rules
```diff
@@ -0,0 +1,91 @@
...
+ifneq ($(product),scylla)
+	sed -e 's/scylla-server/$(product)-server/g' \
+		-e 's/scylla-conf/$(product)-conf/g' \
+		-e 's/scylla-kernel-conf/$(product)-kernel-conf/g' \
+		-e 's/scylla-python3/$(product)-python3/g' \
+		-e 's/scylla-jmx/$(product)-jmx/g' \
+		-e 's/scylla-tools/$(product)-tools/g' \
+		-e 's/scylla-tools-core/$(product)-tools-core/g' \
+		-e 's/Package: scylla$$/Package: $(product)/g' debian/control > debian/control.tmp
+	mv debian/control.tmp debian/control
```
avikivity, line 30: Looks like I gave you bad advice. This is templating in sed. I thought we only needed to template `rules`, but I was wrong.

If we have to do this, then I prefer the first patch. Can you post a link to it so I can look at it again?

syuu1228:
@avikivity force-pushed the old version, please review again.

avikivity:
Queued as 9f6c9f07f5

PR #6427, opened 2020-05-11
Use time window compaction strategy for CDC Log table

haaawk:
CDC Log is a time series so it makes sense to use time window compaction strategy for it.
Our support for time series is limited so we make sure that we don't create more than 24 sstables.
If TTL is configured to 0, meaning data does not expire, we don't use time window compaction strategy.

This PR also sets gc_grace_seconds to 0 when TTL is not set to 0.

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -393,6 +393,9 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID> 
     schema_builder b(s.ks_name(), log_name(s.cf_name()));
     b.with_partitioner("com.scylladb.dht.CDCPartitioner");
     b.set_comment(sprint("CDC log for %s.%s", s.ks_name(), s.cf_name()));
+    if (s.cdc_options().ttl() > 0) {
+        b.set_gc_grace_seconds(0);
+    }
     b.with_column(log_meta_column_name_bytes("stream_id"), bytes_type, column_kind::partition_key);
     b.with_column(log_meta_column_name_bytes("time"), timeuuid_type, column_kind::clustering_key);
     b.with_column(log_meta_column_name_bytes("batch_seq_no"), int32_type, column_kind::clustering_key);
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -51,6 +51,7 @@
 #include "types/listlike_partial_deserializing_iterator.hh"
 #include "tracing/trace_state.hh"
 #include "stats.hh"
+#include "compaction_strategy.hh"
 
 namespace std {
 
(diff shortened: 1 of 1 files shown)


cdc/log.cc
```diff
@@ -389,12 +390,24 @@ bytes log_data_column_deleted_elements_name_bytes(const bytes& column_name) {
     return to_bytes(cdc_deleted_elements_column_prefix) + column_name;
 }
 
+// Makes sure that number of sstables won't grow too large
+static int calculate_window_size(int ttl) {
```
nyh, line 394: Please document what it takes and returns. What does it take (seconds?) and what does it return (minutes?)
Or better yet (I think): just inline this code where you use it, where the context of the output as "MINUTES" is obvious, and the input is obvious, so it doesn't need to be documented.

cdc/log.cc
```diff
@@ -389,12 +390,24 @@ bytes log_data_column_deleted_elements_name_bytes(const bytes& column_name) {
...
+    // What's the minimum window that won't create more than 24 sstables.
+    auto window_size = std::chrono::seconds(ttl / 24 + (ttl % 24 == 0 ? 0 : 1));
+    return std::chrono::ceil<std::chrono::minutes>(window_size).count();
```
nyh, line 397: This is a really elaborate way to divide by 60...
Couldn't this entire function be just written as:
    return ttl / 60 / 24
Or maybe if you are really bothered by the exact division or truncation or something,
    return ttl / 60 / 24 + 1

If the function was this simple, inlining it below will make even more sense.

cdc/log.cc
```diff
@@ -389,12 +390,24 @@ bytes log_data_column_deleted_elements_name_bytes(const bytes& column_name) {
     return to_bytes(cdc_deleted_elements_column_prefix) + column_name;
 }
 
+// Makes sure that number of sstables won't grow too large
+static int calculate_window_size(int ttl) {
+    // What's the minimum window that won't create more than 24 sstables.
+    auto window_size = std::chrono::seconds(ttl / 24 + (ttl % 24 == 0 ? 0 : 1));
```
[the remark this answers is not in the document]
haaawk: It rounds up if ttl does not divide by 24 with no remainder. For example if ttl is 50 then window size shouldn't be 2 but 3 because 50 / 2 > 24 but 50 / 3 <= 24.

Always adding +1 would be less efficient. For example if ttl is 48 then it would lead to window size of 3 while 2 would be perfectly good.

cdc/log.cc
haaawk: ok I will inline.

cdc/log.cc
haaawk: I think using the right types actually has a value. It documents the code.

haaawk:
I inlined the code and made it a bit more self documenting. Here's a diff of the changes since last version:

```
diff --git a/cdc/log.cc b/cdc/log.cc
index 52cf66137..84c82a754 100644
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -390,23 +390,27 @@ bytes log_data_column_deleted_elements_name_bytes(const bytes& column_name) {
     return to_bytes(cdc_deleted_elements_column_prefix) + column_name;
 }
 
-// Makes sure that number of sstables won't grow too large
-static int calculate_window_size(int ttl) {
-    // What's the minimum window that won't create more than 24 sstables.
-    auto window_size = std::chrono::seconds(ttl / 24 + (ttl % 24 == 0 ? 0 : 1));
-    return std::chrono::ceil<std::chrono::minutes>(window_size).count();
-}
-
 static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID> uuid) {
     schema_builder b(s.ks_name(), log_name(s.cf_name()));
     b.with_partitioner("com.scylladb.dht.CDCPartitioner");
     b.set_compaction_strategy(sstables::compaction_strategy_type::time_window);
     b.set_comment(sprint("CDC log for %s.%s", s.ks_name(), s.cf_name()));
-    if (s.cdc_options().ttl() > 0) {
+    auto ttl_seconds = s.cdc_options().ttl();
+    if (ttl_seconds > 0) {
         b.set_gc_grace_seconds(0);
+        auto ceil = [] (int dividend, int divisor) {
+            return dividend / divisor + (dividend % divisor == 0 ? 0 : 1);
+        };
+        auto seconds_to_minutes = [] (int seconds_value) {
+            using namespace std::chrono;
+            return std::chrono::ceil<minutes>(seconds(seconds_value)).count();
+        };
+        // What's the minimum window that won't create more than 24 sstables.
+        auto window_seconds = ceil(ttl_seconds, 24);
+        auto window_minutes = seconds_to_minutes(window_seconds);
         b.set_compaction_strategy_options({
                 {"compaction_window_unit", "MINUTES"},
-                {"compaction_window_size", std::to_string(calculate_window_size(s.cdc_options().ttl()))}
+                {"compaction_window_size", std::to_string(window_minutes)}
         });
     }
     b.with_column(log_meta_column_name_bytes("stream_id"), bytes_type, column_kind::partition_key);
```

haaawk:
Any more comments @nyh @avikivity @raphaelsc ?

haaawk:
This is already in master so closing...

Issue #3464, opened 2018-05-28
Avoid building result_messages


avikivity:
Currently, the signature of `cql_statement::execute` is

```c++
    virtual future<::shared_ptr<cql_transport::messages::result_message>>
        execute(service::storage_proxy& proxy, service::query_state& state, const query_options& options) = 0;
```

where `result_message` contains the result of the query, which is then transcribed by a visitor to the client protocol (native or thrift).

We can avoid building the `result_message` by instead passing the visitor to execute():

```c++
    virtual future<> execute(service::storage_proxy& proxy, service::query_state& state, const query_options& options, result_message_visitor visitor) = 0;
```

The new `result_message_visitor` would be have callbacks for starting and ending rows, for individual cells, or other metadata, and so it can create the response in-place instead of copying it from a separately built result_message.

pdziepak:
I am actually working on with that part of the code right now as a follow up to the IMR series to avoid linearisation of large buffers in the remaining parts of the pipeline. My original idea was just to simply use views to avoid copying and linearisation, but non-trivial selectors make this less simple and it doesn't avoid the cost of creating `cql3::result_set`.

Using visitors directly will allow much better optimisations for the common case. The biggest complication is that we would still need to build something resembling `cql3::result_set` if post-query ordering is required, but I guess that can remain suboptimal for the time being.

avikivity:
@haaawk @jul-stas this may be a difficult task, whoever set the n00b label on this didn't consider it carefully

haaawk:
LOL you've set it :D @avikivity

haaawk:
What difficulty do you have in mind? I thought it's just a change to pass a visitor/consumer that will create the right structure straight away instead of creating intermediate structure and then translating it to the final one.

avikivity:
> LOL you've set it :D @avikivity

Yes I know.

avikivity:
> What difficulty do you have in mind? I thought it's just a change to pass a visitor/consumer that will create the right structure straight away instead of creating intermediate structure and then translating it to the final one.

I don't think there's a specific difficulty, it's just a task involving a lot of complicated classes, and it bridges two complicated subsystems.

PR #6430, opened 2020-05-11
Avoid building `result_message`s

jul-stas:
`result_message`s are the intermediate products, built from the raw result of any query and later translated to CQL `response` (or Thrift response). For performance improvement, we might want to get rid of them and build `response`s from results _in statu nascendi_.

This patch does NOT provide the complete solution. `result_message`s are still needed and `statement::execute` overloads who use them cannot be simply eradicated. But this is something.

Refs #3464


cql3/query_result_consumer.hh
```diff
@@ -0,0 +1,44 @@
...
+#include <seastar/core/sstring.hh>
+
+#include "cql3/result_set.hh"
+#include "transport/event.hh"
+
+namespace cql3 {
+
+class query_result_consumer {
+public:
+    query_result_consumer() = default;
```
avikivity, line 37: Unneeded. A virtual destructor, however, is needed.

cql3/query_result_consumer.hh
```diff
@@ -0,0 +1,44 @@
...
+    query_result_consumer(query_result_consumer&) = delete;
```
avikivity, line 38: Better to make it a standard copy constructor to avoid confusion (const query_result_consumer&)

cql3/query_result_consumer.hh
```diff
@@ -0,0 +1,44 @@
...
+    virtual void move_to_shard(unsigned shard_id) = 0;
+    virtual void add_warning(const sstring& w) = 0;
+    virtual void set_keyspace(const sstring& keyspace) = 0;
+    virtual void set_result(cql3::result rs) = 0;
+    virtual void set_schema_change(shared_ptr<cql_transport::event::schema_change>& sc) = 0;
```
avikivity, line 43: There is a layering violation here in that cql3 now uses names from cql_transport. But it is minor, we can later define a schema_change event in cql3. We can fix it later.

Perhaps we should have a cql3-neutral namespace for defining result sets, one day.

test/lib/cql_test_env.cc
```diff
@@ -109,6 +110,51 @@ static future<> tst_init_ms_fd_gossiper(sharded<gms::feature_service>& features,
 }
 // END TODO
 
+class debug_fmt_consumer : public cql3::query_result_consumer {
```
avikivity, line 113: Let's call it cql_result_message_builder so what it does it more explicit. "debug" can mean many things, I expected it to print out the data when I saw the name.

cql3/query_result_consumer.hh
```diff
@@ -0,0 +1,44 @@
...
+};
+
```
avikivity, line 45: I need a better understanding of the motivation. Yes, we avoid creating a result_message, but we now have to pass a callback instead of getting a simple object. We can do that, if there is sufficient justification, but no justification was given.

avikivity: I think you remove an allocation, since result_message is returned as a shared_ptr, but the allocation can also be removed by defining result_message as

```c++
class result_message {
    std::vector<sstring> _warnings;
public:
    class visitor;
    class visitor_base;

    virtual ~result_message() {}

    virtual void accept(visitor&) const = 0;

    void add_warning(sstring w) {
        _warnings.push_back(std::move(w));
    }

    const std::vector<sstring>& warnings() const {
        return _warnings;
    }

    virtual std::optional<unsigned> move_to_shard() const {
        return std::nullopt;
    }
    //
    // Message types:
    //
    std::variant<
            void_message,
            set_keyspace,
            prepared,
            schema_change,
            rows,
            bounce_to_shard,
            > _payload;
};
```

Since you still construct the `rows` object in your patchset (I presume, didn't read it through yet), we end up with the same work.

avikivity: I'm pausing my review until we resolve this point.

haaawk: This is just the first step. Next is to get rid of rows. The intention was to split the change into smaller parts to reduce a risk of conflicts considering the size of the change. Would you like both parts to be in one PR, @avikivity?

denesb: Get rid of what rows @haaawk?

haaawk: The one Avi mentioned above ;). I believe he meant cql3::result_set::_rows.

cql3/query_result_consumer.hh
haaawk: done.

cql3/query_result_consumer.hh
haaawk: done.

test/lib/cql_test_env.cc
haaawk: done.

cql3/query_result_consumer.hh
haaawk: This is needed after all because copy constructor is removed and that means default constructor is not generated.

haaawk: To be fair, virtual destructor wasn't needed as well because consumers are created and destroyed using their concrete types and just passed down using reference to interface/abstract class. We added virtual destructor though because it's safer.

cql3/query_result_consumer.hh
avikivity: What's the plan? Before coding, I'd like to understand the direction.

Certainly getting rid of cql3::result_set::_rows is a worthy goal. It is even worth the inversion-of-control (passing a callback to the producer, instead of getting an object from the consumer).

But perhaps we can keep both, by returning a generator instead of a deque.

avikivity: And, for the future, communicate these intents in the cover letter. It's perfectly fine to split such work, but communicating it in advance will make it easier to review and result in less back-and-forth.

haaawk: I think @jul-stas was still preparing the PR. That's why he didn't add any reviewers. It could have been marked cleaner, I guess.

So a generator would be an object we `visit` after it's returned from the method?
I think it would be more complex because we would have to ensure the lifetime of all the parameters and intermediate results is extended as long as generator is not destroyed.

We're still working on the final plan. We will keep you posted.

avikivity: No, a generator is something you call operator() on to generate rows. In general callback syle can be converted to generators. Generators also work well with coroutines - instead of calling `callback(row)`, you call `co_yield row` and the caller calls `row = co_await gen()`.

If you have a callback-style function

```
void do_with_callback(callback c) {
    for (...) {
        callback(row)
     }
}
```

You move the state into a generator object and each call to generator::operator() advances the for loop by one row. Or if using coroutines, you keep the for loop and s/callback/co_yield /.

ptrsmrn:
@avikivity you want us to push this PR and eventually merge it? It doesn't seem to have a high priority.

avikivity:
If it improves performance, why not? But it's not high priority.

ptrsmrn:
>> @avikivity you want us to push this PR and eventually merge it? It doesn't seem to have a high priority.

> If it improves performance, why not? But it's not high priority.

@avikivity I thought it's because you requested my review (as in the attached pic), but now I think that CODEOWNERS (or something similar) could have caused that, as I see myeslf automatically added to other PRs as well.

image

avikivity:
I haven't dune anything like that and I'm annoyed at github for misattributing it.

ptrsmrn:
> I haven't dune anything like that and I'm annoyed at github for misattributing it.

@avikivity happened again https://github.com/scylladb/scylladb/pull/22007#event-16178352870

nyh:
> I haven't dune anything like that and I'm annoyed at github for misattributing it.

I think there is a weird bug here: This ancient PR is based on "next", which we usually don't do (we usually base PRs on master). When a PR is based on "next" and "next" is force-pushed (because of dequeue), it seems github comes to it and adds an ugly "... forced pushed next ..." message - there are about 60 of those above. I can also understand github "inviting" the code owners to come have a look. What I can't figure out is why Avi specifically is claimed to have done this invitation. Maybe the last commit in the newly pushed branch was Avi's?

Issue #6414, opened 2020-05-10
flat_mutation_reader: stream ended with unclosed partition, followed by coredump


bentsi:
*Installation details*
Scylla version (or git commit hash):  `4.0.0-0.20200505.d95aa77b62` with build-id c9cf14d33523aca01a1ba502a91eb9d400de9c34 
Cluster size: 6 (multidc, 3 nodes in each region)
OS (RHEL/CentOS/Ubuntu/AWS AMI): `ami-0ad095d521c760c3f ami-09ae7080d6e842625` [eu-west-1 us-east-1 ]

Possibly related to #6394 

RebuildStreamingErr nemesis on node alternator-3h-multidc-alternat-db-node-254e8b33-2 [52.214.126.130 | 10.0.37.49]

```
2020-05-10 02:41:44.879: (DisruptionEvent Severity.NORMAL): type=start name=RebuildStreamingErr node=Node alternator-3h-multidc-alternat-db-node-254e8b33-2 [52.214.126.130 | 10.0.37.49] (seed: False) duration=None
```

Stop Scylla server
```
2020-05-10 02:41:46,312 f:remote.py       l:306  c:sdcm.remote          p:DEBUG > RemoteCmdRunner [centos@52.214.126.130]: Running command "sudo systemctl stop scylla-server.service"...

2020-05-10T02:41:59+00:00  alternator-3h-multidc-alternat-db-node-254e8b33-2 !INFO    | scylla: [shard 0] init - Scylla version 4.0.0-0.20200505.d95aa77b62 shutdown complete.
```

Destroyed some sstables:
```
2020-05-10 02:42:00,131  p:INFO  > RemoteCmdRunner [centos@52.214.126.130]: Command "sudo systemctl stop scylla-server.service" finished with status 0
2020-05-10 02:42:00,227  p:DEBUG > sdcm.nemesis.ChaosMonkey: Selected files for destroy: /var/lib/scylla/data/keyspace1/standard1-01825240924811eabe6c000000000009/mc-400-*
2020-05-10 02:42:00,227  p:DEBUG > RemoteCmdRunner [centos@52.214.126.130]: Running command "sudo rm -f /var/lib/scylla/data/keyspace1/standard1-01825240924811eabe6c000000000009/mc-400-*"...
[...]

Issue #6394, opened 2020-05-07
flat_mutation_reader error and coredump after bootstrap was interrupted with node termination


aleksbykov:
May be similar to the issue https://github.com/scylladb/scylla/issues/6386

*Installation details*
Scylla version (or git commit hash):  4.0.rc3-0.20200501.eee4c00e29 with build-id 0fe095b80ee54c7f47a688aaac0c6bb835118ddc
Cluster size: 4 nodes
OS (RHEL/CentOS/Ubuntu/AWS AMI): ami-0352ad86d128c6419

job for longevity-1tb-7days was running with only TerminateAndReplace nemesis( decommission old node and add new node). Link to job is https://jenkins.scylladb.com/job/scylla-4.0/job/reproducers/job/longevity-1tb-7days/5/

4 times this nemesis had been run without error.
During 5th time, node-8 was decommissioned, and during bootstrap new node9, node-9 and node 7 got terminationspot from aws. These 2 nodes was powered off. rest 2 nodes ( node1 and node 2) reported next errors
node1 
```
2020-05-06T12:04:12+00:00  longevity-tls-1tb-7d-longevit-db-node-3e6cca86-1 !INFO    | scylla: [shard 0] compaction - Compacting [/var/lib/scylla/data/keyspace_zstd/standard1-9cf99e408e1a11eaac5f000000000003/mc-62538-big-Data.db:level=0, /var/lib/scylla/data/keyspace_zstd/standard1-9cf99e408e1a11eaac5f000000000003/mc-62524-big-Data.db:level=0, ]
2020-05-06T12:04:12+00:00  longevity-tls-1tb-7d-longevit-db-node-3e6cca86-1 !ERR     | scylla: [shard 5] flat_mutation_reader - [validator 0x60502e27fef0 for sstable writer /var/lib/scylla/data/keyspace_zstd/standard1-9cf99e408e1a11eaac5f000000000003/mc-68003-big-Data.db (keyspace_zstd.standard1 9cf99e40-8e1a-11ea-ac5f-000000000003)] Unexpected mutation fragment: previous partition end:{position: partition_end,null,0}, current partition end:{position: partition_end,null,0}, at:   0x331b31d#012  0x2e4c861#012  0x10d9973#012  0x10f8841#012  0x1235b05#012  0x1190e89#012  0x1191846#012  0x313b4fc
[...]

PR #6431, opened 2020-05-12
repair: Fix race between write_end_of_stream and apply_rows

asias:
Consider: n1, n2, n1 is the repair master, n2 is the repair follower.

=== Case 1 ===
1) n1 sends missing rows {r1, r2} to n2
2) n2 runs apply_rows_on_follower to apply rows, e.g., {r1, r2}, r1
   is written to sstable, r2 is not written yet, r1 belongs to
   partition 1, r2 belongs to partition 2. It yields after row r1 is
   written.
   data: partition_start, r1
3) n1 sends repair_row_level_stop to n2 because error has happened on n1
4) n2 calls wait_for_writer_done() which in turn calls write_end_of_stream()
   data: partition_start, r1, partition_end
5) Step 2 resumes to apply the rows.
   data: partition_start, r1, partition_end, partition_end, partition_start, r2

=== Case 2 ===
1) n1 sends missing rows {r1, r2} to n2
2) n2 runs apply_rows_on_follower to apply rows, e.g., {r1, r2}, r1
   is written to sstable, r2 is not written yet, r1 belongs to partition
   1, r2 belongs to partition 2. It yields after partition_start for r2
   is written but before _partition_opened is set to true.
   data: partition_start, r1, partition_end, partition_start
3) n1 sends repair_row_level_stop to n2 because error has happened on n1
4) n2 calls wait_for_writer_done() which in turn calls write_end_of_stream().
   Since _partition_opened[node_idx] is false, partition_end is skipped,
   end_of_stream is written.
   data: partition_start, r1, partition_end, partition_start, end_of_stream

This causes unbalanced partition_start and partition_end in the stream
written to sstables.

To fix, serialize the write_end_of_stream and apply_rows with a semaphore.

Fixes: #6394
Fixes: #6296
Fixes: #6414

### DIFF
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -450,6 +450,7 @@ class repair_writer {
     // written.
     std::vector<bool> _partition_opened;
     streaming::stream_reason _reason;
+    named_semaphore _sem{1, named_semaphore_exception_factory{"repair_writer"}};
 public:
     repair_writer(
             schema_ptr schema,
@@ -561,6 +562,7 @@ public:
 
     future<> write_end_of_stream(unsigned node_idx) {
         if (_mq[node_idx]) {
+          return with_semaphore(_sem, 1, [this, node_idx] {
             // Partition_end is never sent on wire, so we have to write one ourselves.
             return write_partition_end(node_idx).then([this, node_idx] () mutable {
                 // Empty mutation_fragment_opt means no more data, so the writer can seal the sstables.
@@ -571,6 +573,7 @@ public:
                         _schema->ks_name(), _schema->cf_name(), ep);
                 return make_exception_future<>(std::move(ep));
             });
+          });
         } else {
             return make_ready_future<>();
         }
@@ -592,6 +595,10 @@ public:
                     _schema->ks_name(), _schema->cf_name(), ep);
             return make_exception_future<>(std::move(ep));
         });
+    }
+
+    named_semaphore& sem() {
+        return _sem;
     }
 };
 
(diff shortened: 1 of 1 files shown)


asias:
On top of: https://github.com/scylladb/scylla/pull/6320. Will rebase once 6320 is merged.

repair/row_level.cc
```diff
@@ -1196,6 +1203,23 @@ class repair_meta {
...
+            return do_for_each(row_diff, [this, node_idx, update_buf] (repair_row& r) {
+                if (update_buf) {
+                    _working_row_buf_combined_hash.add(r.hash());
+                }
+                // The repair_row here is supposed to have
+                // mutation_fragment attached because we have stored it in
+                // to_repair_rows_list above where the repair_row is created.
+                mutation_fragment mf = std::move(r.get_mutation_fragment());
+                auto dk_with_hash = r.get_dk_with_hash();
+                return _repair_writer.do_write(node_idx, std::move(dk_with_hash), std::move(mf));
```
denesb, line 1218: Why not take the semaphore inside `repair_writer::do_write()`?

asias: Because it needs to protect the whole rows, do_write only handles a single row.

denesb:
I have a general comment with regards to `repair_writer`, it is only loosely related to this fix, so you don't have to address this in this patch.
When reading the repair_writer code, while looking for the possible cause of https://github.com/scylladb/scylla/issues/6394, I noticed that all methods are public, although only 1, or maybe 2 is called by external code (`repair_meta`). This makes it harder to reason about the code, as one cannot see at glance that a certain method can only be called from internal code.

asias:
> I have a general comment with regards to `repair_writer`, it is only loosely related to this fix, so you don't have to address this in this patch.
> When reading the repair_writer code, while looking for the possible cause of #6394, I noticed that all methods are public, although only 1, or maybe 2 is called by external code (`repair_meta`). This makes it harder to reason about the code, as one cannot see at glance that a certain method can only be called from internal code.

Yes, I agree. We definitely can move some of the methods to private. I will do after the fixes are merged.

repair/row_level.cc
```diff
@@ -450,6 +450,7 @@ class repair_writer {
     // written.
     std::vector<bool> _partition_opened;
     streaming::stream_reason _reason;
+    named_semaphore _sem{1, named_semaphore_exception_factory{"repair_writer"}};
```
denesb, line 453: Should this be per node? This will now serialize all writes to all node-specific queues.

asias: It does not matter because 1) on the repair master, the apply_rows runs one node after another 2) on the repair follower, there is only at most one peer node. That's why I am using a single sem for all the peer nodes.

slivne:
@avikivity we need this merged and backported

avikivity:
@asias the dependnecy got promoted.

repair/row_level.cc
```diff
@@ -561,6 +562,7 @@ class repair_writer {
 
     future<> write_end_of_stream(unsigned node_idx) {
         if (_mq[node_idx]) {
+          return with_semaphore(_sem, 1, [this, node_idx] {
```
avikivity, line 565: This looks drastic, it is done for every partition.

avikivity: Is there no simpler solution?

avikivity: Scratch that, it's only at the end.

asias: Yes, it is called at the end of the stream only.

asias:
Changes in de205e65328962f2f8e6300ff000b2e6c5ae98b7 (v2):
- Rebase to master
- Remove the dependency patch from this PR

avikivity:
Queued as b2c4d9fdbc

PR #6443, opened 2020-05-13
docs/redis: API reference documentation

penberg:
The Redis API in Scylla only supports a small subset of the Redis
commands. Let's document what we support so people have the right
expectations when they try it out.


nyh:
Looks good to me, seems to be missing the "exists" command in the list, but other than that appears to be accurate. But @syuu1228  @fastio can you please review too?
If we do add such a list, we'll need to then make an effort to keep it up-to-date when adding new features.

penberg:
@nyh Good catch, added `EXISTS` to the reference!

nyh:
Thanks. Committed as 96e35f841c6353e17d7190a7078f1f0dbbabd85c

PR #6463, opened 2020-05-15
Added avro library and a simple test

zirrock:
Adding avro library used in advanced scylla replication with kafka producer


haaawk:
This is a wrong repo. You should send it here https://github.com/haaawk/scylla/tree/cdc_kafka_integration

PR #6464, opened 2020-05-17
cdc: Tune expired sstables check frequency

haaawk:
CDC Log is a time series which uses time window compaction with some
time window. Data is TTLed with the same value. This means that sstable
won't become fully expired more often than once per time window
duration.

This patch sets expired_sstable_check_frequency_seconds compaction
strategy parameter to half of the time window. Default value of this
parameter is 10 minutes which in most cases won't be a good fit.
By default, we set TTL to 24h and time window to 1h. This means that
with a default value of the parameter we would be checking every 10
minutes but new expired sstable would appear only every 60 minutes.

The parameter is set to half of the time window duration because it's
the expected time we have to wait for sstable to become fully expired.
Half of the time we will wait longer and half of the time we will wait
shorter.

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -410,7 +410,12 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID> 
         auto window_minutes = seconds_to_minutes(window_seconds);
         b.set_compaction_strategy_options({
                 {"compaction_window_unit", "MINUTES"},
-                {"compaction_window_size", std::to_string(window_minutes)}
+                {"compaction_window_size", std::to_string(window_minutes)},
+                // A new SSTable will become fully expired every
+                // `window_seconds` seconds so we shouldn't check for expired
+                // sstables too often.
+                {"expired_sstable_check_frequency_seconds",
+                        std::to_string(std::max(1, window_seconds / 2))},
         });
     }
     b.with_column(log_meta_column_name_bytes("stream_id"), bytes_type, column_kind::partition_key);


avikivity:
@raphaelsc perhaps we can make the parameter redundant. We know from statistics when the sstable can be expired, so we can set a timer and expire it exactly on time.

avikivity:
(not affecting this patch)

haaawk:
> @raphaelsc perhaps we can make the parameter redundant. We know from statistics when the sstable can be expired, so we can set a timer and expire it exactly on time.

I think it might be a good idea.

haaawk:
> r an event that can only happen once every 60 minutes is silly, and as Avi said, we can calculate exactly when it will happen. But what's the harm of doing some arithmetic once every 10 minutes that caused you to submit this patch? It doesn't actually process the content of the table -

Let me reverse the question. Why not to avoid unnecessary work if all we have to do is setup an existing parameter? I agree this is not a game changer but still there is no harm in properly using what we already have implemented, is there?

nyh:
Pushed to next. commit cd33b9f4068e9adf56c41bb91de05cf26198f752

nyh:
> > r an event that can only happen once every 60 minutes is silly, and as Avi said, we can calculate exactly when it will happen. But what's the harm of doing some arithmetic once every 10 minutes that caused you to submit this patch? It doesn't actually process the content of the table -
> 
> Let me reverse the question. Why not to avoid unnecessary work if all we have to do is setup an existing parameter? I agree this is not a game changer but still there is no harm in properly using what we already have implemented, is there?

Sometimes it's better to inherit some silly default than to override it by a slightly-less-but-still-imperfect setting, because if in the future we change the default behavior to be smarter (perhaps as @avikivity suggested above), this override may remain and we forget to remove it. But never mind - this is fine. I committed your patch. We can reconsider this later.

haaawk:
Thanks @nyh. If we fix the behaviour then most certainly we should remove the parameter and then fix all the places that use it.

raphaelsc:
> @raphaelsc perhaps we can make the parameter redundant. We know from statistics when the sstable can be expired, so we can set a timer and expire it exactly on time.

Hi Avi. That is https://github.com/scylladb/scylla/issues/5743

raphaelsc:
> > > r an event that can only happen once every 60 minutes is silly, and as Avi said, we can calculate exactly when it will happen. But what's the harm of doing some arithmetic once every 10 minutes that caused you to submit this patch? It doesn't actually process the content of the table -
> > 
> > 
> > Let me reverse the question. Why not to avoid unnecessary work if all we have to do is setup an existing parameter? I agree this is not a game changer but still there is no harm in properly using what we already have implemented, is there?
> 
> Sometimes it's better to inherit some silly default than to override it by a slightly-less-but-still-imperfect setting, because if in the future we change the default behavior to be smarter (perhaps as @avikivity suggested above), this override may remain and we forget to remove it. But never mind - this is fine. I committed your patch. We can reconsider this later.

We live in an imperfect world, out of order happens, user overwrites data (not recommended at all for TWCS), and so TWCS may still have to fallback to periodic checks. That's because an overlapping sstable B can prevent a sstable A from being purged when its metadata tells us that all of its data are now expired. From there on, we rely on some periodic check to get rid of sstable A because its metadata is now useless, because we can't determine from it anymore when sstable A can possibly be purged.

We could have some hybrid approach. Set up an alarm clock for a sstable A based on metadata, if we were unable to purge sstable A when the alarm was triggered, reschedule it using the TWCS's interval parameter. I wonder if this is not overkill, couldn't we be affecting performance to some extent by having some many alarms set up for each TWCS sstable in the system?

PR #6471, opened 2020-05-18
dist: rename coredump .mount to var-lib-scylla-coredump.mount

### LINKED ISSUE #6402: incorrect name for coredump mount systemd unit file
**Installation details**

Scylla version (or git commit hash): 666.development-20200506.e813778bd1
OS (RHEL/CentOS/Ubuntu/AWS AMI): OEL7.6

**Description**

We have a [failure](https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-oel76/70/) of artifacts test because `scylla_coredump_setup` unable to load coredump mount unit file.

This happen on OEL only due to stricter systemd rules (I think) but anyway it's a requirement from systemd to have unit name matched with `Where=` statement.

```
$ systemd-analyze verify var-lib-systemd-coredump.mount
...
var-lib-systemd-coredump.mount's Where= setting doesn't match unit name. Refusing.
..

$ man systemd.mount
...
       Mount units must be named after the mount point directories they control. 
       Example: the mount point /home/lennart must be configured in a unit file 
       home-lennart.mount. For details about the escaping logic used to convert 
       a file system path to a unit name, see systemd.unit(5). Note that mount 
       units cannot be templated, nor is possible to add multiple names to a 
       mount unit by creating additional symlinks to it.
...
```

This bug introduced by fix for #6300 (#6301) and **we need to rename unit file to `var-lib-scylla-coredump.mount`**

See also https://unix.stackexchange.com/a/345518 for more detailed explanation.

syuu1228:
Since we change bind-mount opposite way, .mount file should be renamed too.

Fixes #6402

### DIFF
--- a/dist/common/scripts/scylla_coredump_setup
+++ b/dist/common/scripts/scylla_coredump_setup
@@ -73,11 +73,11 @@ Options=bind
 [Install]
 WantedBy=multi-user.target
 '''[1:-1]
-            with open('/etc/systemd/system/var-lib-systemd-coredump.mount', 'w') as f:
+            with open('/etc/systemd/system/var-lib-scylla-coredump.mount', 'w') as f:
                 f.write(dot_mount)
             makedirs('/var/lib/scylla/coredump')
             systemd_unit.reload()
-            systemd_unit('var-lib-systemd-coredump.mount').enable()
+            systemd_unit('var-lib-scylla-coredump.mount').enable()
         if os.path.exists('/usr/lib/sysctl.d/50-coredump.conf'):
             run('sysctl -p /usr/lib/sysctl.d/50-coredump.conf')
         else:


slivne:
@syuu1228 the referenced github issue seems wrong

syuu1228:
@slivne The issue no was typo, fixed now. Please review again

dist/common/scripts/scylla_coredump_setup
```diff
@@ -73,11 +73,11 @@ Options=bind
...
 '''[1:-1]
-            with open('/etc/systemd/system/var-lib-systemd-coredump.mount', 'w') as f:
+            with open('/etc/systemd/system/var-lib-scylla-coredump.mount', 'w') as f:
                 f.write(dot_mount)
             makedirs('/var/lib/scylla/coredump')
             systemd_unit.reload()
-            systemd_unit('var-lib-systemd-coredump.mount').enable()
+            systemd_unit('var-lib-scylla-coredump.mount').enable()
         if os.path.exists('/usr/lib/sysctl.d/50-coredump.conf'):
             run('sysctl -p /usr/lib/sysctl.d/50-coredump.conf')
```
avikivity, line 82: I'm confused, shouldn't we bind-mount /var/lib/scylla/coredump into /var/lib/systemd/coredump? So we get the filesystem with lots of space mounted into /var/lib/systemd/coredump?

Should we switch What and Where instead?

slivne:
@penberg / @syuu1228 this is needed to unblock 4.1 please comment

dist/common/scripts/scylla_coredump_setup
penberg: I am also confused, we just switched them around in e77dad3adf7051a489223bbd55ecbc461b377421. @syuu1228 @bentsi

penberg:
The underlying issue was fixed by commit 0b30df8f23b3377f435c57733b25f5227743e4ec, right @amoskong?

@syuu1228 we can close this pull request, no?

dist/common/scripts/scylla_coredump_setup
amoskong: https://github.com/scylladb/scylla/commit/e77dad3adf7051a489223bbd55ecbc461b377421 is a wrong fix.
I reverted it by https://github.com/scylladb/scylla/commit/9d9d54c804dd400d90edfdb84c8f5319d06b1633

amoskong:
> The underlying issue was fixed by commit [0b30df8](https://github.com/scylladb/scylla/commit/0b30df8f23b3377f435c57733b25f5227743e4ec), right @amoskong?

It's fixed by a revert (https://github.com/scylladb/scylla/commit/9d9d54c804dd400d90edfdb84c8f5319d06b1633)

> @syuu1228 we can close this pull request, no?

I think we can close this issue.

PR #6484, opened 2020-05-20
Allow our users to shoot themselves in their feet.

kbr-:
Allow a node to join without bootstrapping, even if it couldn't contact other nodes.

Print a BIG WARNING saying that you should never join nodes without
bootstrapping (by marking it as a seed or using auto_bootstrap=off).

Only the very first node should (must) be joined as a seed.

If you want to have more seeds, first join them using the only supported
way (i.e. bootstrap them), and only AFTER they have bootstrapped, change
their configuration to include them in the seed list.

Does not fix, but closes https://github.com/scylladb/scylla/issues/6005. Read the discussion: it's enlightening.
See https://github.com/scylladb/scylla-docs/issues/2647 for the correct procedure of joining a node.

Reverts https://github.com/scylladb/scylla/commit/7cb6ac33f5f54cae665cecd3a90428eb52d424e3.

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -626,6 +626,19 @@ void storage_service::join_token_ring(int delay) {
         }
     }
 
+
+    if (!is_auto_bootstrap()) {
+        slogger.warn("auto_bootstrap set to \"off\". This causes UNDEFINED BEHAVIOR. YOU MAY LOSE DATA.");
+    }
+
+    if (!db::system_keyspace::bootstrap_complete() && _gossiper.get_seeds().count(get_broadcast_address())) {
+        slogger.warn("Bootstrapping node marked as seed (present in the seed list)."
+                     " This can only be done for the very first node in a new cluster."
+                     " If this is not the first node, YOU MAY LOSE DATA."
+                     " Bootstrapping new nodes into an existing cluster as seeds"
+                     " causes UNDEFINED BEHAVIOR. DO NOT EVER do that.");
+    }
+
     slogger.debug("Setting tokens to {}", _bootstrap_tokens);
     // This node must know about its chosen tokens before other nodes do
     // since they may start sending writes to this node after it gossips status = NORMAL.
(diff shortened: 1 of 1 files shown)


service/storage_service.cc
```diff
@@ -654,6 +626,19 @@ void storage_service::join_token_ring(int delay) {
         }
     }
 
+
+    if (!is_auto_bootstrap()) {
+        slogger.warn("auto_bootstrap set to \"off\". This causes UNDEFINED BEHAVIOR. YOU WILL LOSE DATA.");
```
haaawk, line 631: Why this?

kbr-: For the same reason. auto_bootstrap=off does the same thing as marking the node as seed.

vladzcloudius: s/YOU WILL LOSE DATA/YOU MAY LOOSE DATA/

vladzcloudius:
I think that we should go even further and completely forbid joining as a seeder and put `auto_bootstrap` behind `experimental`.
I'll create a github issue for that.
In the meantime, the warning is a step into the right direction.

service/storage_service.cc
kbr-: I prefer "YOU WILL LOSE DATA", it's more dramatic. Scare them, because they should be scared.
But of course it's not completely true :D so I'll change it.

asias:
Hello @kbr-  I like the warning. Can we move the warning part into a separate patch? In case we want to backport them separately. Other than that, looks good to me.

kbr-:
1. I put the warnings into a separate commit so it can be easily cherry-picked for backporting.
2. Added info about reverting https://github.com/scylladb/scylla/commit/7cb6ac33f5f54cae665cecd3a90428eb52d424e3.

kbr-:
@avikivity or @nyh got acks from 3 reviewers, can you look at this?

PR #6491, opened 2020-05-21
CDC: Don't split collection tombstone away from base update

### LINKED ISSUE #6084: CDC: wrong value for set in cdc_log table after second insert
*Installation details*
scylla-release: 0.20200325.78f5afec30
scylla-version: 4.0.rc0
scylla-BuildID[sha1]: 6a210b1adac28793113c61e037cdd06040f0d217
Cluster size: 1

table configured with value column as set<text> and cdc postimage enabled. After inserting twice same row, postimage row contains empty value:
```
cqlsh> CREATE KEYSPACE ks WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1} ;
cqlsh> CREATE TABLE ks.cf (pkey int, ckey int, value list<int>, PRIMARY KEY (pkey, ckey)) WITH cdc={'enabled': true, 'postimage': true};
cqlsh> DROP TABLE ks.cf
   ... ;
cqlsh> CREATE TABLE ks.cf (pkey int, ckey int, value set<text>, PRIMARY KEY (pkey, ckey)) WITH cdc={'enabled': true, 'postimage': true};
cqlsh> INSERT INTO ks.cf (pkey, ckey, value) VALUES (1, 1, {"value1", "value2"});
SyntaxException: line 1:52 no viable alternative at input '{'
cqlsh> INSERT INTO ks.cf (pkey, ckey, value) VALUES (1, 1, \{"value1", "value2"\});
Invalid syntax at line 1, char 53
  INSERT INTO ks.cf (pkey, ckey, value) VALUES (1, 1, \{"value1", "value2"\});
                                                      ^
cqlsh> INSERT INTO ks.cf (pkey, ckey, value) VALUES (1, 1, {"value1", "value2"});
SyntaxException: line 1:52 no viable alternative at input '{'
cqlsh> INSERT INTO ks.cf (pkey, ckey, value) VALUES (1, 1, { 'value1', 'value2'});
cqlsh> SELECT * FROM ks.cf
cf                 cf_scylla_cdc_log  
cqlsh> SELECT * FROM ks.cf_scylla_cdc_log ;

 cdc$stream_id                      | cdc$time                             | cdc$batch_seq_no | cdc$deleted_elements_value | cdc$deleted_value | cdc$operation | cdc$ttl | ckey | pkey | value
------------------------------------+--------------------------------------+------------------+----------------------------+-------------------+---------------+---------+------+------+----------------------
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f1a-6f20-11ea-c953-f6fb8f2d34a5 |                0 |                       null |              True |             1 |    null |    1 |    1 |                 null
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f1a-6f20-11ea-c953-f6fb8f2d34a5 |                1 |                       null |              null |             9 |    null |    1 |    1 |                 null
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f24-6f20-11ea-94ca-5ce3c02b9709 |                0 |                       null |              null |             2 |    null |    1 |    1 |                 null
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f24-6f20-11ea-94ca-5ce3c02b9709 |                1 |                       null |              null |             9 |    null |    1 |    1 |                 null
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f24-6f20-11ea-94ca-5ce3c02b9709 |                2 |                       null |              null |             1 |    null |    1 |    1 | {'value1', 'value2'}
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f24-6f20-11ea-94ca-5ce3c02b9709 |                3 |                       null |              null |             9 |    null |    1 |    1 | {'value1', 'value2'}

(6 rows)
cqlsh> INSERT INTO ks.cf (pkey, ckey, value) VALUES (1, 1, { 'value1', 'value2'});
cqlsh> SELECT * FROM ks.cf_scylla_cdc_log ;

 cdc$stream_id                      | cdc$time                             | cdc$batch_seq_no | cdc$deleted_elements_value | cdc$deleted_value | cdc$operation | cdc$ttl | ckey | pkey | value
------------------------------------+--------------------------------------+------------------+----------------------------+-------------------+---------------+---------+------+------+----------------------
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f1a-6f20-11ea-c953-f6fb8f2d34a5 |                0 |                       null |              True |             1 |    null |    1 |    1 |                 null
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f1a-6f20-11ea-c953-f6fb8f2d34a5 |                1 |                       null |              null |             9 |    null |    1 |    1 |                 null
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f24-6f20-11ea-94ca-5ce3c02b9709 |                0 |                       null |              null |             2 |    null |    1 |    1 |                 null
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f24-6f20-11ea-94ca-5ce3c02b9709 |                1 |                       null |              null |             9 |    null |    1 |    1 |                 null
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f24-6f20-11ea-94ca-5ce3c02b9709 |                2 |                       null |              null |             1 |    null |    1 |    1 | {'value1', 'value2'}
 0x31be5747eefc68b185b1b5812701c5d1 | 36976f24-6f20-11ea-94ca-5ce3c02b9709 |                3 |                       null |              null |             9 |    null |    1 |    1 | {'value1', 'value2'}
 0x31be5747eefc68b185b1b5812701c5d1 | 46588b5a-6f20-11ea-4c32-73a08a236811 |                0 |                       null |              null |             0 |    null |    1 |    1 | {'value1', 'value2'}
 0x31be5747eefc68b185b1b5812701c5d1 | 46588b5a-6f20-11ea-4c32-73a08a236811 |                1 |                       null |              True |             1 |    null |    1 |    1 |                 null
 0x31be5747eefc68b185b1b5812701c5d1 | 46588b5a-6f20-11ea-4c32-73a08a236811 |                2 |                       null |              null |             9 |    null |    1 |    1 |                 null
 0x31be5747eefc68b185b1b5812701c5d1 | 46588b64-6f20-11ea-ead1-ca68aecd025e |                0 |                       null |              null |             0 |    null |    1 |    1 |                 null
 0x31be5747eefc68b185b1b5812701c5d1 | 46588b64-6f20-11ea-ead1-ca68aecd025e |                1 |                       null |              null |             2 |    null |    1 |    1 |                 null
>>>WRONG VALUE>>>>0x31be5747eefc68b185b1b5812701c5d1 | 46588b64-6f20-11ea-ead1-ca68aecd025e |                2 |                       null |              null |             9 |    null |    1 |    1 |       {'', 'value1'}<<<<<
 0x31be5747eefc68b185b1b5812701c5d1 | 46588b64-6f20-11ea-ead1-ca68aecd025e |                3 |                       null |              null |             0 |    null |    1 |    1 | {'value1', 'value2'}
 0x31be5747eefc68b185b1b5812701c5d1 | 46588b64-6f20-11ea-ead1-ca68aecd025e |                4 |                       null |              null |             1 |    null |    1 |    1 | {'value1', 'value2'}
 0x31be5747eefc68b185b1b5812701c5d1 | 46588b64-6f20-11ea-ead1-ca68aecd025e |                5 |                       null |              null |             9 |    null |    1 |    1 | {'value1', 'value2'}

(15 rows)
```

jul-stas:
This patch reduces the number of rows in CDC log when collections are inserted or updated.

Suppose we have a table `ks.t` with non-frozen collection column `X`.
The following statement:
`update ks.t using timestamp T set X = null where ...`
creates a tombstone for column `X` **with timestamp `T-1`**.
The following statement:
`update ks.t using timestamp T set X = {cell} where ...`
creates a tombstone for column `X` **with timestamp `T-1`** (same as above) and a cell for column `X` with timestamp `T`.

Because we don't want to split the second update into two log rows, then for any timestamp `T`, **we put collection tombstones with timestamp `T` into the `T+1` bucket** (they will show up in the log with timestamp `T+1`). In particular, if we have both a tombstone and a bunch of cells, such that the timestamps of these cells are the tombstone's timestamp + 1, they end up being put into a single log row (unless cells have TTLs, but that's a different story).
The user can then replay this log entry using an `update` statement with this timestamp.

Unfortunately, there is a small inconsistency. While:
`update ks.t using timestamp T set X = null where ...`
uses timestamp `T-1` for the tombstone,
`delete from ks.t using timestamp T where ...`
**uses timestamp `T`** :(
So if we're given a mutation with collection tombstone with timestamp `T`, what timestamp should we use for the `time` column? If we use `T`, it will be consistent with `delete`, but won't be consistent with `update`. If we use `T+1`, it will be consistent with `update`, but won't be consistent with `delete`. But **we have to choose something**. And the best thing to do is to use `T+1` to be consistent with `update`, because it causes less splitting.
We simply need to document the following for the user: when replaying the log, don't use `delete` statement, use `update` and then `cdc$time` provides you the correct timestamp. If you really want to use `delete` for some reason (you don't need to), you need to adjust the timestamp from `cdc$time` by +1.

Also, `cdc_test` had to be updated in places that relied on former splitting strategy. More tests will come in a separate PR.

Patch co-authored by @kbr- and @jul-stas.
Fixes #6084

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -547,6 +547,12 @@ api::timestamp_type find_timestamp(const schema& s, const mutation& m) {
                     [&] (collection_mutation_view_description mview) {
                 t = mview.tomb.timestamp;
                 if (t != api::missing_timestamp) {
+                    // A collection tombstone with timestamp T can be created with:
+                    // UPDATE ks.t USING TIMESTAMP T + 1 SET X = null WHERE ...
+                    // where X is a non-atomic column.
+                    // This is, among others, the reason why we show it in the CDC log
+                    // with cdc$time using timestamp T + 1 instead of T.
+                    t += 1;
                     return stop_iteration::yes;
                 }
 
@@ -742,17 +748,79 @@ private:
     const column_definition& _op_col;
     const column_definition& _ttl_col;
     ttl_opt _cdc_ttl_opt;
+
     /**
-     * #6070
-     * When mutation splitting was added, non-atomic column assignments were broken
-     * into two invocation of transform. This means the second (actual data assignment)
-     * does not know about the tombstone in first one -> postimage is created as if 
-     * we were _adding_ to the collection, not replacing it. 
+     * #6070, #6084
+     * Non-atomic column assignments which use a TTL are broken into two invocations
+     * of `transform`, such as in the following example:
+     * CREATE TABLE t (a int PRIMARY KEY, b map<int, int>) WITH cdc = {'enabled':true};
+     * UPDATE t USING TTL 5 SET b = {0:0} WHERE a = 0;
+     *
+     * The above UPDATE creates a tombstone and a (0, 0) cell; because tombstones don't have the notion
+     * of a TTL, we split the UPDATE into two separate changes (represented as two separate delta rows in the log,
+     * resulting in two invocations of `transform`): one change for the deletion with no TTL,
+     * and one change for adding cells with TTL = 5.
+     *
+     * In other words, we use the fact that
+     * UPDATE t USING TTL 5 SET b = {0:0} WHERE a = 0;
+     * is equivalent to
+     * BEGIN UNLOGGED BATCH
+     *  UPDATE t SET b = null WHERE a = 0;
+     *  UPDATE t USING TTL 5 SET b = b + {0:0} WHERE a = 0;
+     * APPLY BATCH;
+     * (the mutations are the same in both cases),
+     * and perform a separate `transform` call for each statement in the batch.
+     *
+     * An assignment also happens when an INSERT statement is used as follows:
+     * INSERT INTO t (a, b) VALUES (0, {0:0}) USING TTL 5;
      * 
-     * Not pretty, but to handle this we use the knowledge that we always get 
-     * invoked in timestamp order -> tombstone first, then assign.
-     * So we simply keep track of non-atomic columns deleted across calls 
-     * and filter out preimage data post this.
+     * This will be split into three separate changes (three invocations of `transform`):
+     * 1. One with TTL = 5 for the row marker (introduces by the INSERT), indicating that a row was inserted.
+     * 2. One without a TTL for the tombstone, indicating that the collection was cleared.
+     * 3. One with TTL = 5 for the addition of cell (0, 0), indicating that the collection
+     *    was extended by a new key/value.
+     *
+     * Why do we need three changes and not two, like in the UPDATE case?
+     * The tombstone needs to be a separate change because it doesn't have a TTL,
+     * so only the row marker change could potentially be merged with the cell change (1 and 3 above).
+     * However, we cannot do that: the row marker change is of INSERT type (cdc$operation == cdc::operation::insert),
+     * but there is no way to create a statement that
+     * - has a row marker,
+     * - adds cells to a collection,
+     * - but *doesn't* add a tombstone for this collection.
+     * INSERT statements that modify collections *always* add tombstones.
+     *
+     * Merging the row marker with the cell addition would result in such an impossible statement.
+     *
+     * Instead, we observe that
+     * INSERT INTO t (a, b) VALUES (0, {0:0}) USING TTL 5;
+     * is equivalent to
+     * BEGIN UNLOGGED BATCH
+     *  INSERT INTO t (a) VALUES (0) USING TTL 5;
+     *  UPDATE t SET b = null WHERE a = 0;
+     *  UPDATE t USING TTL 5 SET b = b + {0:0} WHERE a = 0;
+     * APPLY BATCH;
+     * and perform a separate `transform` call for each statement in the batch.
+     *
+     * Unfortunately, due to splitting, the cell addition call (b + b {0:0}) does not know about the tombstone.
+     * If it was performed independently from the tombstone call, it would create a wrong post-image:
+     * the post-image would look as if the previous cells still existed.
+     * For example, suppose that b was equal to {1:1} before the above statement was performed.
+     * Then the final post-image for b for above statement/batch would be {0:0, 1:1}, when instead it should be {0:0}.
+     *
+     * To handle this we use the fact that
+     * 1. changes without a TTL are treated as if TTL = 0,
+     * 2. `transform` is invoked in order of increasing TTLs,
+     * and we maintain state between `transform` invocations (`_non_atomic_column_deletes`).
+     *
+     * Thus, the tombstone call will happen *before* the cell addition call,
+     * so the cell addition call will know that there previously was a tombstone
+     * and create a correct post-image.
+     *
+     * Furthermore, `transform` calls for INSERT changes (i.e. with a row marker)
+     * happen before `transform` calls for UPDATE changes, so in the case of an INSERT
+     * which modifies a collection column as above, the row marker call will happen first;
+     * its post-image will still show {1:1} for the collection column. Good.
      */
     std::unordered_set<const column_definition*> _non_atomic_column_deletes;
 
--- a/cdc/split.cc
+++ b/cdc/split.cc
@@ -30,23 +30,16 @@ struct atomic_column_update {
     atomic_cell cell;
 };
 
-// see the comment inside `clustered_row_insert` for motivation for separating
-// nonatomic deletions from nonatomic updates
-struct nonatomic_column_deletion {
-    column_id id;
-    tombstone t;
-};
-
 struct nonatomic_column_update {
     column_id id;
+    tombstone t; // optional
     utils::chunked_vector<std::pair<bytes, atomic_cell>> cells;
 };
 
 struct static_row_update {
     gc_clock::duration ttl;
     std::vector<atomic_column_update> atomic_entries;
-    std::vector<nonatomic_column_deletion> nonatomic_deletions;
-    std::vector<nonatomic_column_update> nonatomic_updates;
+    std::vector<nonatomic_column_update> nonatomic_entries;
 };
 
 struct clustered_row_insert {
@@ -54,19 +47,14 @@ struct clustered_row_insert {
     clustering_key key;
     row_marker marker;
     std::vector<atomic_column_update> atomic_entries;
-    std::vector<nonatomic_column_deletion> nonatomic_deletions;
-    // INSERTs can't express updates of individual cells inside a non-atomic
-    // (without deleting the entire field first), so no `nonatomic_updates` field
-    // overwriting a nonatomic column inside an INSERT will be split into two changes:
-    // one with a nonatomic deletion, and one with a nonatomic update
+    std::vector<nonatomic_column_update> nonatomic_entries;
 };
 
 struct clustered_row_update {
     gc_clock::duration ttl;
     clustering_key key;
     std::vector<atomic_column_update> atomic_entries;
-    std::vector<nonatomic_column_deletion> nonatomic_deletions;
-    std::vector<nonatomic_column_update> nonatomic_updates;
+    std::vector<nonatomic_column_update> nonatomic_entries;
 };
 
 struct clustered_row_deletion {
@@ -95,8 +83,7 @@ using set_of_changes = std::map<api::timestamp_type, batch>;
 
 struct row_update {
     std::vector<atomic_column_update> atomic_entries;
-    std::vector<nonatomic_column_deletion> nonatomic_deletions;
-    std::vector<nonatomic_column_update> nonatomic_updates;
+    std::vector<nonatomic_column_update> nonatomic_entries;
 };
 
 static
@@ -122,7 +109,7 @@ extract_row_updates(const row& r, column_kind ckind, const schema& schema) {
                         v.timestamp(),
                         v.is_live_and_has_ttl() ? v.ttl() : gc_clock::duration(0)
                     );
-                auto& updates = result[timestamp_and_ttl].nonatomic_updates;
+                auto& updates = result[timestamp_and_ttl].nonatomic_entries;
                 if (updates.empty() || updates.back().id != id) {
                     updates.push_back({id, {}});
                 }
@@ -130,8 +117,12 @@ extract_row_updates(const row& r, column_kind ckind, const schema& schema) {
             }
 
             if (desc.tomb) {
-                auto timestamp_and_ttl = std::pair(desc.tomb.timestamp, gc_clock::duration(0));
-                result[timestamp_and_ttl].nonatomic_deletions.push_back({id, desc.tomb});
+                auto timestamp_and_ttl = std::pair(desc.tomb.timestamp + 1, gc_clock::duration(0));
+                auto& updates = result[timestamp_and_ttl].nonatomic_entries;
+                if (updates.empty() || updates.back().id != id) {
+                    updates.push_back({id, {}});
+                }
+                updates.back().t = std::move(desc.tomb);
             }
         });
     });
@@ -148,8 +139,7 @@ set_of_changes extract_changes(const mutation& base_mutation, const schema& base
         res[timestamp].static_updates.push_back({
                 ttl,
                 std::move(up.atomic_entries),
-                std::move(up.nonatomic_deletions),
-                std::move(up.nonatomic_updates)
+                std::move(up.nonatomic_entries)
             });
     }
 
@@ -173,6 +163,9 @@ set_of_changes extract_changes(const mutation& base_mutation, const schema& base
         };
 
         for (auto& [k, up]: cr_updates) {
+            // It is important that changes in the resulting `set_of_changes` are listed
+            // in increasing TTL order. The reason is explained in a comment in cdc/log.cc,
+            // search for "#6070".
             auto [timestamp, ttl] = k;
 
             if (is_insert(timestamp, ttl)) {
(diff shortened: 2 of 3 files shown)


cdc/log.cc
```diff
@@ -519,6 +519,8 @@ api::timestamp_type find_timestamp(const schema& s, const mutation& m) {
                     [&] (collection_mutation_view_description mview) {
                 t = mview.tomb.timestamp;
                 if (t != api::missing_timestamp) {
+                    t += 1; // TODO: test this behavior
```
haaawk, line 522: Could you please explain what this is for and why it's needed?

cdc/split.cc
```diff
@@ -30,22 +30,15 @@ struct atomic_column_update {
...
 struct nonatomic_column_update {
     column_id id;
+    tombstone t; // optional
     utils::chunked_vector<std::pair<bytes, atomic_cell>> cells;
 };
 
 struct static_row_update {
     gc_clock::duration ttl;
     std::vector<atomic_column_update> atomic_entries;
-    std::vector<nonatomic_column_deletion> nonatomic_deletions;
```
haaawk, line 48: We don't delete stuff from collections any more?

cdc/split.cc
```diff
@@ -54,18 +47,13 @@ struct clustered_row_insert {
...
-    // one with a nonatomic deletion, and one with a nonatomic update
+    std::vector<nonatomic_column_update> nonatomic_updates;
 };
 
 struct clustered_row_update {
     gc_clock::duration ttl;
     clustering_key key;
     std::vector<atomic_column_update> atomic_entries;
-    std::vector<nonatomic_column_deletion> nonatomic_deletions;
     std::vector<nonatomic_column_update> nonatomic_updates;
```
haaawk, line 57: Maybe rename this to nonatomic_entries?

cdc/split.cc
```diff
@@ -54,18 +47,13 @@ struct clustered_row_insert {
     clustering_key key;
     row_marker marker;
     std::vector<atomic_column_update> atomic_entries;
-    std::vector<nonatomic_column_deletion> nonatomic_deletions;
-    // INSERTs can't express updates of individual cells inside a non-atomic
-    // (without deleting the entire field first), so no `nonatomic_updates` field
-    // overwriting a nonatomic column inside an INSERT will be split into two changes:
-    // one with a nonatomic deletion, and one with a nonatomic update
+    std::vector<nonatomic_column_update> nonatomic_updates;
```
haaawk, line 50: Maybe call in nonatomic_entries?

cdc/split.cc
haaawk: NVM. I see that we keep updates now instead of deletions.

cdc/log.cc
elcallio: Yes please. Are you preserving the timestamp from the original collection mutation tombstone and trying to compensate for the adjustment there? If so, a more solid approach would be to use the timestamp or the rest of the collection mutation set, not adjusting by constant.

kbr-: There might be no cells in the collection mutation, just a tombstone.
This is the same adjustment that we're doing in the splitting logic:
```
            if (desc.tomb) {
                auto timestamp_and_ttl = std::pair(desc.tomb.timestamp + 1, gc_clock::duration(0));
                auto& updates = result[timestamp_and_ttl].nonatomic_updates;
                if (updates.empty() || updates.back().id != id) {
                    updates.push_back({id, {}});
                }
                updates.back().t = std::move(desc.tomb);
            }
```

Why are we doing this? I think it should be explained in detail in the PR cover letter (@jul-stas). For now let me explain it here.

Suppose we have a table `ks.t` with non-frozen collection column `X`.
The following statement:
```update ks.t using timestamp T set X = null where ...```
creates a tombstone for column `X` **with timestamp `T-1`**.
The following statement:
```update ks.t using timestamp T set X = X + {cell} where ...```
creates a tombstone for column `X` **with timestamp `T-1`** and a cell for column `X` with timestamp `T`.
Because we don't want to split the second update into two log rows, then for any timestamp `T`, **we put collection tombstones with timestamp `T` into the `T+1` bucket** (they will show up in the log with timestamp `T+1`). In particular, if we have both a tombstone and a bunch of cells, such that the timestamps of these cells are the tombstone's timestamp + 1, we put them together into a single log row (unless cells have TTLs, but that's a different story).
The user can then replay this log entry using an `update` statement with this timestamp.

Unfortunately, there is a small inconsistency. While:
```update ks.t using timestamp T set X = null where ...```
uses timestamp `T-1` for the tombstone,
```delete from ks.t using timestamp T where ...```
**uses timestamp `T`** :(
So if we're given a mutation with collection tombstone with timestamp `T`, what timestamp should we use for the `time` column? If we use `T`, it will be consistent with `delete`, but won't be consistent with `update`. If we use `T+1`, it will be consistent with `update`, but won't be consistent with `delete`. But **we have to choose something**. And the best thing to do is to use `T+1` to be consistent with `update`, because it causes less splitting.
We simply need to document the following for the user: when replaying the log, don't use `delete` statement, use `update` and then `cdc$time` provides you the correct timestamp. If you really want to use `delete` for some reason (you don't need to), you need to adjust the timestamp from `cdc$time` by +1.

Comment (kbr-) [reply cdc/log.cc:522]: > Overwriting or inserting a collection using timestamp T is a process with
> following steps:
> 
>  inserting a row marker (if applicable) with timestamp T;
>  writing a collection tombstone with timestamp T-1;
>  writing the new collection value with timestamp T.

@jul-stas not true for `DELETE`, as explained above.

File: cdc/split.cc (hunk 3 above)
Comment (jul-stas) [reply cdc/split.cc]: done

File: cdc/split.cc (hunk 4 above)
Comment (jul-stas) [reply cdc/split.cc]: done

**Comment by jul-stas** [comment]
I updated description from Kamil's essay.
@nyh - what you think?

**Comment by kbr-** [comment]
Without TTL:

BEFORE:
```
cqlsh> create table ks.t (a int primary key, b map<int, int>) with cdc = {'enabled':true};
cqlsh> insert into ks.t (a, b) values (0, {1:2});
cqlsh> select * from ks.t_scylla_cdc_log ;

 cdc$stream_id                      | cdc$time                             | cdc$batch_seq_no | a | b      | cdc$deleted_b | cdc$deleted_elements_b | cdc$operation | cdc$ttl
------------------------------------+--------------------------------------+------------------+---+--------+---------------+------------------------+---------------+---------
 0xce654506fb99e00aae2d9dabd627b256 | f1062b4c-a1a8-11ea-a918-d0320a5c7a45 |                0 | 0 |   null |          True |                   null |             1 |    null
 0xce654506fb99e00aae2d9dabd627b256 | f1062b56-a1a8-11ea-a63d-bab4aeb404be |                0 | 0 |   null |          null |                   null |             2 |    null
 0xce654506fb99e00aae2d9dabd627b256 | f1062b56-a1a8-11ea-a63d-bab4aeb404be |                1 | 0 | {1: 2} |          null |                   null |             1 |    null

(3 rows)
```

AFTER:
```
cqlsh> create table ks.t (a int primary key, b map<int, int>) with cdc = {'enabled':true};
cqlsh> insert into ks.t (a, b) values (0, {1:2});
cqlsh> select * from ks.t_scylla_cdc_log ;

 cdc$stream_id                      | cdc$time                             | cdc$batch_seq_no | a | b      | cdc$deleted_b | cdc$deleted_elements_b | cdc$operation | cdc$ttl
------------------------------------+--------------------------------------+------------------+---+--------+---------------+------------------------+---------------+---------
 0xcdc4a167196fb3256c60238e49b984bc | 5813adf6-a1a8-11ea-9a74-f35c034fffa6 |                0 | 0 | {1: 2} |          True |                   null |             2 |    null

(1 rows)
```

---

With TTL:

BEFORE:
```
cqlsh> create table ks.t (a int primary key, b map<int, int>) with cdc = {'enabled':true};
cqlsh> insert into ks.t (a, b) values (0, {1:2}) using ttl 5;
cqlsh> select * from ks.t_scylla_cdc_log ;

 cdc$stream_id                      | cdc$time                             | cdc$batch_seq_no | a | b      | cdc$deleted_b | cdc$deleted_elements_b | cdc$operation | cdc$ttl
------------------------------------+--------------------------------------+------------------+---+--------+---------------+------------------------+---------------+---------
 0xce654506fb99e00aae2d9dabd627b256 | 1834d3f8-a1a9-11ea-397c-a0ce0d32f9fb |                0 | 0 |   null |          True |                   null |             1 |    null
 0xce654506fb99e00aae2d9dabd627b256 | 1834d402-a1a9-11ea-48e1-00590ffcf6a3 |                0 | 0 |   null |          null |                   null |             2 |       5
 0xce654506fb99e00aae2d9dabd627b256 | 1834d402-a1a9-11ea-48e1-00590ffcf6a3 |                1 | 0 | {1: 2} |          null |                   null |             1 |       5

(3 rows)
```

AFTER:
```
cqlsh> create table ks.t (a int primary key, b map<int, int>) with cdc = {'enabled':true};
cqlsh> insert into ks.t (a, b) values (0, {1:2}) using ttl 5;
cqlsh> select * from ks.t_scylla_cdc_log ;

 cdc$stream_id                      | cdc$time                             | cdc$batch_seq_no | a | b      | cdc$deleted_b | cdc$deleted_elements_b | cdc$operation | cdc$ttl
------------------------------------+--------------------------------------+------------------+---+--------+---------------+------------------------+---------------+---------
 0xcdc4a167196fb3256c60238e49b984bc | 80c07e32-a1a8-11ea-c150-c671884341fb |                0 | 0 |   null |          null |                   null |             2 |       5
 0xcdc4a167196fb3256c60238e49b984bc | 80c07e32-a1a8-11ea-c150-c671884341fb |                1 | 0 |   null |          True |                   null |             1 |    null
 0xcdc4a167196fb3256c60238e49b984bc | 80c07e32-a1a8-11ea-c150-c671884341fb |                2 | 0 | {1: 2} |          null |                   null |             1 |       5

(3 rows)
```

(observe that entries now show up with the same cdc$time value even though they are split)

**Comment by haaawk** [comment]
@elcallio @kbr- @piodul @nyh Could you please review?

File: cdc/log.cc [hunk 5]
```diff
@@ -547,6 +547,12 @@ api::timestamp_type find_timestamp(const schema& s, const mutation& m) {
                     [&] (collection_mutation_view_description mview) {
                 t = mview.tomb.timestamp;
                 if (t != api::missing_timestamp) {
+                    // A collection tombstone with timestamp T can be created with:
+                    // UPDATE ks.t USING TIMESTAMP T + 1 SET X = null WHERE ...
+                    // where X is a non-atomic column.
+                    // This is, among others, the reason why we show it in the CDC log
+                    // with cdc$time using timestamp T + 1 instead of T.
+                    t += 1;
```
elcallio, line 555: Are you not doing this twice now? split.cc:120?

cdc/split.cc
```diff
@@ -122,16 +109,20 @@ extract_row_updates(const row& r, column_kind ckind, const schema& schema) {
...
                 if (updates.empty() || updates.back().id != id) {
                     updates.push_back({id, {}});
                 }
                 updates.back().cells.push_back({std::move(k), std::move(v)});
             }
 
             if (desc.tomb) {
-                auto timestamp_and_ttl = std::pair(desc.tomb.timestamp, gc_clock::duration(0));
-                result[timestamp_and_ttl].nonatomic_deletions.push_back({id, desc.tomb});
+                auto timestamp_and_ttl = std::pair(desc.tomb.timestamp + 1, gc_clock::duration(0));
```
elcallio, line 120: If not using the timestamp of the actual collection mutations, why not at least add some (debug?) code that verifies that the respective timestamps are T and T+1?

cdc/log.cc
kbr-: There are two independent code paths, one goes through `find_timestamp` (if we don't split), the other goes through `split`. Duplicate logic, unfortunately.

I wanted to have everything to go through the split logic, but there was an argument about efficiency or something.

cdc/split.cc
kbr-: There is no such assumption.
Inside a `collection_mutation` there can be a tombstone with timestamp `T` and a cell with timestamp e.g. `T+123`. The user can create any kind of weird mutations with `begin batch ... apply batch`.

nyh:
> The following statement:
> `update ks.t using timestamp T set X = X + {cell} where ...`
> creates a tombstone for column `X` **with timestamp `T-1`** (same as above) and a cell for column `X` with timestamp `T`.

@jul-stas I believe this example is wrong, but doesn't change the validity of the rest of the discussion because the example can be fixed. When you do "set X = X + .." as in your example, this is an **append** operation, and the whole point of the CRDT nature of Scylla's collections is that we do this with a single write of the new element. It does not write a tombstone for the old collection.

You probably meant to give "set X = ..." as an example, replacing the entire old contents of the collection with new contents. In that case, the code indeed uses `cql3::update_parameters::make_tombstone_just_before()` to set a tombstone for ts-1 in addition to the new value in timestamp ts. The reason why the tombstone has to be for ts-1 and not ts is that when a tombstone and data tie in the same ts, the tombstone wins (this is a rule enforced throughout Scylla and Cassandra) - so we cannot put the tombstone in the same ts, and need to use ts-1.

jul-stas:
> @jul-stas I believe this example is wrong
@nyh Yes, description updated!

nyh:
@jul-stas please clarify if this patch is still relevant - and does it still "Fixes #6084" despite @haaawk saying in issue 6084 saying that that issue should be closed?

nyh:
@jul-stas github review process makes it impossible to discuss the commit message and (more importantly) what it is saying you are doing, so I have to do it here.

My question is, when someone **replaces** (overwrites) a collection, is there a special flag in the CDC log saying that this is a complete replacement operation instead of just append to the existing collection? (@haaawk @kbr- I still hope one day you guys will document the log format in docs/cdcm.md or elsewhere I can always refer to).

The reason I'm asking this is that you can't just leave a tombstone at time t instead of t-1 - that would be wrong (because a tombstone overwrites modifications at the same timestamp). So if you put some special flag at time t, it needs to have a different name, calling it a "tombstone" would be too confusing. Do we have such a flag? Maybe I'll understand better what's going on after reading the full patch, but from the commit message I didn't understand what is the plan to mark at time t instead of the tombstone at time t-1.

nyh:
(sorry, it seems I accidentally closed this PR earlier, reopening).

nyh:
Committed to next (a1e235b1a4d53a7522fa43b24b766b0d1209d856)

haaawk:
> Ok, I struggled a lot with this patch, and believe I now understand that it is correct, and since three other reviewers even more familiar with this code approved it, I will commit it.
> HOWEVER, I like to ask again that you please write documentation on how to **USE** (i.e., read) the CDC log, which would have helped greatly here. We need to document that if you see a tombstone in a CDC log entry then you **must** use "UPDATE x = null where ..." and not "DELETE x where ..." - and document why (DELETE x will win over the data addition while UPDATE x = null will lose - but only in case of collections). There is no way that any user can guess that. Heck, I don't think any of you will remember this in a month :-)
> 
> Also, the fact that SET x = null puts a tombstone at time t-1 while DELETE x puts a tombstone at time t - and that this only (?) happens for collection columns is so counter-intuitive, we must have a test for it. If someone ever breaks this weird assumption by mistake, CDC (and possibly other stuff) will break in really surprising ways. We should have a test that confirms our understanding of these details - and with a big comment saying why this test is important and cannot be changed without breaking CDC.

Kamil has a task scheduled to refactor the log to make it more maintainable. He should get to it in a week or so.

kbr-:
> Kamil has a task scheduled to refactor the log to make it more maintainable. He should get to it in a week or so.

But Nadav was asking for documentation, not more/different code.
The refactor won't change the logic. The logic itself is complex and will stay complex, no matter how we write the code which implements the logic.
Documentation for collections and post-images is yet to be written:
https://github.com/scylladb/scylla-docs/issues/2696
https://github.com/scylladb/scylla-docs/issues/2695

cdc/log.cc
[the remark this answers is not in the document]
kbr-: Yes.

cdc/log.cc
```diff
@@ -742,17 +748,79 @@ class transformer final {
...
+     * resulting in two invocations of `transform`): one change for the deletion with no TTL,
+     * and one change for adding cells with TTL = 5.
+     *
+     * In other words, we use the fact that
+     * UPDATE t USING TTL 5 SET b = {0:0} WHERE a = 0;
+     * is equivalent to
+     * BEGIN UNLOGGED BATCH
+     *  UPDATE t SET b = null WHERE a = 0;
+     *  UPDATE t USING TTL 5 SET b = b + {0:0} WHERE a = 0;
+     * APPLY BATCH;
```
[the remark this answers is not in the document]
kbr-: This is exactly what happens.
`DELETE using timestamp T` creates a tombstone with timestamp `T`.
`SET = null using timestamp T` creates a tombstone with timestamp `T-1`

>  Do we have a test-case for this for both collections and non-collections? We should :-(

I agree. For collections, this is being indirectly tested by CDC tests, but I guess we can write a non-CDC-specific test to "ground" this behavior.

cdc/log.cc
```diff
@@ -742,17 +748,79 @@ class transformer final {
...
+     *
+     * Unfortunately, due to splitting, the cell addition call (b + b {0:0}) does not know about the tombstone.
+     * If it was performed independently from the tombstone call, it would create a wrong post-image:
+     * the post-image would look as if the previous cells still existed.
+     * For example, suppose that b was equal to {1:1} before the above statement was performed.
+     * Then the final post-image for b for above statement/batch would be {0:0, 1:1}, when instead it should be {0:0}.
+     *
+     * To handle this we use the fact that
+     * 1. changes without a TTL are treated as if TTL = 0,
+     * 2. `transform` is invoked in order of increasing TTLs,
```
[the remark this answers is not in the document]
kbr-: Yes.

PR #6497, opened 2020-05-25
reloc-pkg: move all files under project name directory

### LINKED ISSUE #6315: current relocatable packages are not optimized to unify single tarball

syuu1228:
*This is reopened version of  https://github.com/scylladb/scylla/pull/6327, since it dequeued*

To make unified relocatable package easily, we may want to merge tarballs to single tarball like this:
zcat *.tar.gz | gzip -c > scylla-unified.tar.xz
But it's not possible with current relocatable package format, since there are multiple files conflicts, install.sh, SCYLLA-*-FILE, dist/, README.md, etc..

To support this, we need to archive everything in the directory when building relocatable package.

This is modifying relocatable package format, we need to provide a way to
detect the format version.
To do this, we added a new file ".relocatable_package_version" on the top of the
archive, and set version number "2" to the file.

This also requires scylla-ccm change, here is a PR to do this:
https://github.com/scylladb/scylla-ccm/pull/236

Fixes #6315

### DIFF
--- a/scripts/create-relocatable-package-python3.py
+++ b/scripts/create-relocatable-package-python3.py
@@ -34,6 +34,21 @@ import tarfile
 from tempfile import mkstemp
 import magic
 
+
+RELOC_PREFIX='scylla-python3'
+def reloc_add(self, name, arcname=None, recursive=True, *, filter=None):
+    if arcname:
+        return self.add(name, arcname="{}/{}".format(RELOC_PREFIX, arcname))
+    else:
+        return self.add(name, arcname="{}/{}".format(RELOC_PREFIX, name))
+
+def reloc_addfile(self, tarinfo, fileobj=None):
+    tarinfo.name = "{}/{}".format(RELOC_PREFIX, tarinfo.name)
+    return self.addfile(tarinfo, fileobj)
+
+tarfile.TarFile.reloc_add = reloc_add
+tarfile.TarFile.reloc_addfile = reloc_addfile
+
 def should_copy(f):
     '''Given a file, returns whether or not we are interested in copying this file.
     We want the actual python interepreter, and the files in /lib(64) and /usr/lib(64)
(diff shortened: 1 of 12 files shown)


scripts/create-relocatable-package-python3.py
```diff
@@ -228,19 +243,24 @@ def generate_file_list(executables):
 
 file_list = generate_file_list(dependencies(packages))
 ar = tarfile.open(args.output, mode='w|gz')
+# relocatable package format version = 2
+with open('build/.relocatable_package_version', 'w') as f:
+    f.write('2\n')
+ar.add('build/.relocatable_package_version', arcname='.relocatable_package_version')
```
avikivity, line 249: You can add content directly, without creating a file. See TarFile.addfile() and io.BytesIO(). We already use it scylla-python3: 

```python
    ti = tarfile.TarInfo(name=os.path.join("bin", pybin))
    ti.size = len(thunk)
    ti.mode = 0o755
    ar.addfile(ti, fileobj=io.BytesIO(thunk))
```

avikivity:
Agree with the patch, waiting until issues with ccm are resolved.

avikivity:
We can simplify it by adding a root-level install.sh that calls scylla/install.sh. This way old ccm can still work with this package.

install.sh:
```sh
#!/bin/bash

exec "$(dirname "$0")/scylla/install.sh" "$@"
```

Then we don't need to coordinate with ccm.

If we have a script that merges packages, it can replace the top-level install.sh.

avikivity:
btw, let's harden install.sh so it doesn't need to be called from the same directory. That is /path/to/install.sh will work (and will work with relative directories input). But that's for another patch.

penberg:
@avikivity The CCM side of things is now merged https://github.com/scylladb/scylla-ccm/commit/a8fd146492faa9163f6590b159a0601d47ec627d.

@syuu1228 unfortunately there's some conflicts. Can you please rebase?

syuu1228:
@penberg Rebased with latest next.

PR #6500, opened 2020-05-25
cdc/storage_proxy: keep cdc_service alive in storage_proxy operations

kbr-:
storage_proxy is never deinitialized, so it may have still used cdc_service
after its destructor was called.

This fixes the problem by cdc_service inheriting from
async_sharded_service and storage_proxy calling shared_from_this on
the service whenever it uses it.

cdc_service inherits from async_sharded_service and not simply from
enable_shared_from_this, because there might be other services that
cdc_service depends on. Assuming that these services are
deinitialized after cdc_service (as they should), i.e. after stop() is
called on cdc_service, making cdc_service async_sharded_service will
keep their deinitialization code from being called until all references
to cdc_service disappear (async_sharded_service keeps stop() from
returning until this happens).

Some more improvements should be possible through some refactoring:
1. Make augment_mutation_call a free function, not a member of
   cdc_service: it doesn't need any state that cdc_service has.
   db_context can be passed down from storage_proxy when it calls the
   function.
2. Remove the storage_proxy -> cdc_service reference. storage_proxy
   only needs augment_mutation_call, which would not be a part of the
   service. This would also get rid of the proxy -> cdc -> proxy
   reference cycle that we have now, and would allow storage_proxy to be
   safely deinitialized after cdc_service.
3. Maybe we could even remove the cdc_service -> storage_proxy
   reference. Is it really needed?

### DIFF
--- a/cdc/log.hh
+++ b/cdc/log.hh
@@ -75,7 +75,7 @@ class metadata;
 /// CDC service will listen for schema changes and iff CDC is enabled/changed
 /// create/modify/delete corresponding log tables etc as part of the schema change. 
 ///
-class cdc_service {
+class cdc_service final : public async_sharded_service<cdc::cdc_service> {
     class impl;
     std::unique_ptr<impl> _impl;
 public:
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -277,6 +277,7 @@ cdc::cdc_service::cdc_service(db_context ctxt)
 }
 
 future<> cdc::cdc_service::stop() {
+    _impl->_ctxt._proxy.set_cdc_service(nullptr);
     return _impl->stop();
 }
 
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -1214,9 +1214,10 @@ future<> paxos_response_handler::learn_decision(lw_shared_ptr<paxos::proposal> d
         const auto base_tbl_id = update_mut.column_family_id();
         std::vector<mutation> update_mut_vec{std::move(update_mut)};
 
-        if (_proxy->get_cdc_service()->needs_cdc_augmentation(update_mut_vec)) {
-            f_cdc = _proxy->get_cdc_service()->augment_mutation_call(_timeout, std::move(update_mut_vec), tr_state, _cl_for_learn)
-                    .then([this, base_tbl_id] (std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>&& t) {
+        auto cdc = _proxy->get_cdc_service();
+        if (cdc && cdc->needs_cdc_augmentation(update_mut_vec)) {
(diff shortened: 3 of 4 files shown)


cdc/log.hh
```diff
@@ -75,7 +75,7 @@ class metadata;
 /// CDC service will listen for schema changes and iff CDC is enabled/changed
 /// create/modify/delete corresponding log tables etc as part of the schema change. 
 ///
-class cdc_service {
+class cdc_service final : public async_sharded_service<cdc::cdc_service> {
```
haaawk, line 78: Don't you have to make the destructor virtual now?

kbr-:
> I think the destructor should now be virtual

None of the other services have virtual destructors. Which makes sense: nothing destroys these services through a pointer to a class that it inherits from --- that would probably be a bug.

In case of `cdc_service`, someone would have to store a pointer to `async_sharded_service<cdc_service>`, but why the heck would it do that?

elcallio:
Destructor only needs to be virtual if any holder does not know the exact held type. Since cdc_service is never extended and passed as a pointer to base class (I assume), it is not needed.

kbr-:
@elcallio @nyh ping

elcallio:
Looks ok I guess, but should we not have a shutdown in cdc service that waits for pending log (etc) operations as well?

kbr-:
Should only be necessary if there are discarded futures created in service
methods which we don't do.

If there's a future returned from a service method to its caller, the
caller must keep the service alive until the future resolves.

In our case there is one user: storage_proxy.

śr., 27 maj 2020, 17:23 użytkownik Calle Wilund <notifications@github.com>
napisał:

elcallio:
Not true. The service is also registered as a migration listener.

kbr-:
Ok, but `stop` waits on the future returned by `_migration_notifier.unregister_listener(this)`, and then it's the responsibility of migration_notifier not to use the service after the future is resolved, because that means the service is not listening anymore (in particular, it could be destroyed at this point).

If migration_notifier still uses cdc_service after this future resolves,  that's a bug in migration_notifier.

kbr-:
And indeed there seems to be such a bug in migration_notifier...

kbr-:
I opened an issue:
https://github.com/scylladb/scylla/issues/6525

let's fix one thing at a time.

kbr-:
@elcallio what do you think?
I'm suggesting to leave the migration_notifier problem for another PR.

elcallio:
Can't say I 100% agree that it is migration_managers responsibility to guarantee that you are not currently in a callback, but sure, go ahead.

kbr-:
Ok, I guess we can discuss that on https://github.com/scylladb/scylla/issues/6525.
For now I'm going try to get this PR merged.

kbr-:
@nyh got 2 acks, can you look at this?

kbr-:
@nyh ping

nyh:
Committed to next: 013330199db03467a72db2e996e165ea8fd8c1d2

kbr-:
Thanks.

PR #6503, opened 2020-05-25
priority_manager: merge streaming_read and streaming_write classes to one class

avikivity:
Streaming is handled by just once group for CPU scheduling, so
separating it into read and write classes for I/O is artificial, and
inflates the resources we allow for streaming if both reads and writes
happen at the same time.

Merge both classes into one class ("streaming") and adjust callers. The
merged class has 200 shares, so it reduces streaming bandwidth if both
directions are active at the same time (which is rare; I think it only
happens in view building).


avikivity:
ping @nyh

avikivity:
ping

nyh:
It's my fault this patch decayed without being committed. I'll rebase it myself and commit.

nyh:
Rebased/ported to next, and committed as  de380918276883bf8ae62a0a0b112dc74ccd9725

PR #6504, opened 2020-05-25
CDC generation management refactors

kbr-:
Currently the management of CDC generations happens in storage_service,
which is a big ball of mud that does many unrelated things.

This PR introduces a new service specifically crafted for managing
CDC generations: creating new generations when joining a cluster or
performing a rolling upgrade, listening to gossiper for generation changes,
and passing generations between nodes.

storage_service will simply call two methods of the service:
1. before_join_token_ring, to create a new CDC generation when the node
   bootstraps,
2. after_join, to handle rolling upgrade, among other things.

These refactors also change (simplify) the startup procedure slightly.
Previously, a new CDC generation would be always created on the
auto_bootstrap=off/seed joining path. However, we would not be able to
ensure that it contains the tokens of all nodes. Therefore this part was
removed. Now it will only create a generation on rolling upgrade or if
its the very first node.

If the user joins a node as a seed or with auto_bootstrap=off
into a cluster, the old generation will be used, which means colocation
of streams will be lost. To restore it, the user will need to use a
nodetool command to manually create a new CDC generation (separate PR).

This PR depends on https://github.com/scylladb/scylla/pull/6500 and https://github.com/scylladb/scylla/pull/6484.

### DIFF
--- a/cdc/generation.cc
+++ b/cdc/generation.cc
@@ -34,8 +34,36 @@
 #include "gms/application_state.hh"
 #include "gms/inet_address.hh"
 #include "gms/gossiper.hh"
+#include "gms/feature_service.hh"
 
 #include "cdc/generation.hh"
+#include "cdc/generation_service.hh"
+
+class and_reducer {
+private:
+    bool _result = true;
+public:
+    future<> operator()(bool value) {
+        _result = value && _result;
+        return make_ready_future<>();
+    }
+    bool get() {
+        return _result;
+    }
+};
+
+class or_reducer {
+private:
+    bool _result = false;
+public:
+    future<> operator()(bool value) {
+        _result = value || _result;
+        return make_ready_future<>();
+    }
+    bool get() {
+        return _result;
+    }
+};
 
 extern logging::logger cdc_log;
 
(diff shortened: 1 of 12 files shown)


main.cc
```diff
@@ -867,12 +867,6 @@ int main(int ac, char** av) {
 
             distributed_loader::ensure_system_table_directories(db).get();
 
-            static sharded<cdc::cdc_service> cdc;
```
haaawk, line 870: @elcallio was there any particular reason you put the initialization of cdc service in this place or is it fine to move it?

elcallio:
No reason beyond prerequisite services/objects.

cdc/generation_service.hh
```diff
@@ -0,0 +1,136 @@
...
+     * such as learning a generation or updating the streams description table. */
+    gate _async_gate;
+
+public:
+    // See comment in main.cc for the reason of taking sharded<db::sys_dist_ks>& instead of simply db::sys_dist_ks&.
+    generation_service(gms::gossiper&, sharded<db::system_distributed_keyspace>&,
+            abort_source&, const db::config&, const locator::token_metadata&);
+
+    future<> stop();
+    ~generation_service();
```
haaawk, line 104: Shouldn't this be virtual?

haaawk:
> No reason beyond prerequisite services/objects.

Thanks.

cdc/generation.cc
```diff
@@ -328,4 +356,290 @@ void update_streams_description(
...
+        return;
+    }
+
+    learn_generation(*ts);
+}
+
+void generation_service::on_join(gms::inet_address ep, gms::endpoint_state ep_state) {
+    assert_shard_zero(__PRETTY_FUNCTION__);
+
+    for (const auto& e : ep_state.get_application_state_map()) {
```
haaawk, line 540: Do we have to loop over all the application states? Can't we just get CDC_STREAMS_TIMESTAMP?

cdc/generation.cc
```diff
@@ -328,4 +356,290 @@ void update_streams_description(
...
+}
+
+void generation_service::before_join_token_ring(
+        const std::unordered_set<dht::token>& bootstrap_tokens,
+        const gms::feature_service& feature_svc,
+        bool in_single_node_test) {
+    assert_shard_zero(__PRETTY_FUNCTION__);
+    assert(!db::system_keyspace::bootstrap_complete());
+
+    if (!_db_cfg.check_experimental(db::experimental_features_t::CDC)) {
```
haaawk, line 420: Isn't this already handled by features mechanism?

service/storage_proxy.cc
```diff
@@ -1214,9 +1214,10 @@ future<> paxos_response_handler::learn_decision(lw_shared_ptr<paxos::proposal> d
...
 
-        if (_proxy->get_cdc_service()->needs_cdc_augmentation(update_mut_vec)) {
-            f_cdc = _proxy->get_cdc_service()->augment_mutation_call(_timeout, std::move(update_mut_vec), tr_state, _cl_for_learn)
-                    .then([this, base_tbl_id] (std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>&& t) {
+        auto cdc = _proxy->get_cdc_service();
+        if (cdc && cdc->needs_cdc_augmentation(update_mut_vec)) {
+            f_cdc = cdc->augment_mutation_call(_timeout, std::move(update_mut_vec), tr_state, _cl_for_learn)
+                    .then([this, base_tbl_id, cdc = cdc->shared_from_this()] (std::tuple<std::vector<mutation>, lw_shared_ptr<cdc::operation_result_tracker>>&& t) {
                 auto mutations = std::move(std::get<0>(t));
                 auto tracker = std::move(std::get<1>(t));
```
piodul, line 1222: You will also have to add a shared pointer to cdc service inside `cdc::operation_result_tracker` structure. The tracker updates cdc stats after cdc mutations complete, so it needs to keep the service alive.

kbr-: cdc_stats are actually in storage_proxy.

kbr-:
1. remove unnecessary loop in `generation_service::on_join`
2. rebased on next. I need yet to run tests though: waiting for Fedora to upgrade.

cdc/generation.cc
kbr-: fixd

cdc/generation_service.hh
kbr-: Only gossiper holds a pointer to a base class of `generation_service` (it holds a pointer to the subscriber interface). But we `unregister_` this pointer in `generation_service::stop()`. So the pointer which destroys us is of type `generation_service*`.

cdc/generation.cc
kbr-: I don't understand the question.
We need to check if CDC is enabled on this node before creating a new CDC generation.
If it's disabled, we don't create.

We could instead check if the whole cluster supports CDC but that's not necessary. The first node which enables CDC on rolling upgrade can already safely create a CDC generation: other nodes will just ignore it until they upgrade.

haaawk: Makes sense.

cdc/generation_service.hh
```diff
@@ -28,8 +28,23 @@
 #include "metadata.hh"
 #include "generation.hh"
 
+namespace seastar {
+    class abort_source;
```
haaawk, line 32: Wrong indent.

cdc/generation_service.hh
```diff
@@ -28,8 +28,23 @@
...
+}
+
 namespace gms {
     class feature_service;
+    class inet_address;
+    class gossiper;
+}
+
+namespace db {
+    class config;
```
haaawk, line 42: Wrong indent.

cdc/generation_service.hh
```diff
@@ -28,8 +28,23 @@
...
+    class system_distributed_keyspace;
+}
+
+namespace locator {
+    class token_metadata;
```
haaawk, line 47: Wrong indent.

cdc/generation_service.hh
```diff
@@ -28,8 +28,23 @@
 #include "metadata.hh"
 #include "generation.hh"
 
+namespace seastar {
+    class abort_source;
+}
+
 namespace gms {
     class feature_service;
+    class inet_address;
```
haaawk, line 37: Wrong indent.

cdc/generation_service.hh
```diff
@@ -0,0 +1,136 @@
...
+
+#include <seastar/core/semaphore.hh>
+
+#include "gms/i_endpoint_state_change_subscriber.hh"
+
+#include "metadata.hh"
+#include "generation.hh"
+
+namespace gms {
+    class feature_service;
```
haaawk, line 36: Wrong indent.

cdc/log.cc
```diff
@@ -266,8 +267,8 @@ class cdc::cdc_service::impl : service::migration_listener::empty_listener {
     }
 };
 
-cdc::cdc_service::cdc_service(service::storage_proxy& proxy)
-    : cdc_service(db_context::builder(proxy).build())
+cdc::cdc_service::cdc_service(service::storage_proxy& proxy, cdc::generation_service& gen_svc)
```
haaawk, line 270: I think it would be better not to couple cdc_service with generation_service. You can just take cdc::metadata& and use generation_service::get_cdc_metadata() by the caller of cdc_service constructor.

main.cc
```diff
@@ -989,9 +991,26 @@ int main(int ac, char** av) {
...
+                    std::ref(stop_signal.as_sharded_abort_source()), std::ref(*cfg), std::ref(token_metadata)).get();
+            auto stop_cdc_gen_service = defer_verbose_shutdown("CDC Generation Management service", [] {
+                cdc_gen_service.stop().get();
+            });
+
+            supervisor::notify("starting CDC Log service");
             static sharded<cdc::cdc_service> cdc;
-            cdc.start(std::ref(proxy)).get();
-            auto stop_cdc_service = defer_verbose_shutdown("cdc", [] {
+            // FIXME: find a way to pass only metadata; cdc log doesn't need the entire cdc gen service
```
haaawk, line 1011: I see you already considered my comment above :)

service/storage_proxy.cc
piodul: Right, my bad.

cdc/log.cc
kbr-: Yeah.
Unfortunately, I'm unsure how to do that elegantly.

It would probably be best if `service` was a [functor](https://en.wikipedia.org/wiki/Functor_(functional_programming)) and we had a function
```
template <typename T1, typename T2>
service<T2> fmap(function<T1, T2> f, service<T1>)
```
which we could use for `f = generation_service::get_cdc_metadata` to obtain `service<cdc::metadata>`, which wouldn't be a "real" service but could be passed to `cdc.start(...)`.

kbr-: I was also thinking of using `sharded<T>::start_single`, but it doesn't allow to start multiple instances one by one:
```
sharded<Service>::start_single(Args&&... args) {
    assert(_instances.empty());
    _instances.resize(1);
```

kbr-:
Rebased and fixed the most important thing in the world: indentation.

slivne:
@haaawk do we need this merged ?

haaawk:
It's not crucial but @kbr- did some improvements so probably worth getting this in eventually. Not a blocker though.

kbr-:
Superseded by https://github.com/scylladb/scylla/pull/8101, closing.

cdc/generation_service.hh
elcallio: Adding virtual does not hurt.

kbr-: @elcallio this PR is no longer relevant, it is superseded by https://github.com/scylladb/scylla/pull/8101

Issue #6501, opened 2020-05-25
scylla_raid_setup: failed to get Linux distribution info


enaydanov:
This is Scylla's bug tracker, to be used for reporting bugs only.
If you have a question about Scylla, and not a bug, please ask it in
our mailing-list at scylladb-dev@googlegroups.com or in our slack channel.

- [x] I have read the disclaimer above, and I am reporting a suspected malfunction in Scylla.

**Installation details**

Scylla version (or git commit hash): 666.development-0.20200524.076c8317c

**Description**

We're using deprecated function in `dist/common/scripts/scylla_util.py`: `platform.dist()`

https://github.com/scylladb/scylla/blob/67a07e9cbc498bbac21a9bb545993232f5a164cb/dist/common/scripts/scylla_util.py#L429

https://github.com/scylladb/scylla/blob/67a07e9cbc498bbac21a9bb545993232f5a164cb/dist/common/scripts/scylla_util.py#L433

And we've updated to Python 3.8.2 recently which removes this function: https://docs.python.org/3.7/library/platform.html#platform.dist

In result, `scylla_raid_setup` script failed:

```
Creating XFS volume for scylla using 1 disk(s): /dev/nvme0n1
RAID setup failed.

Stderr:

WARN  2020-05-25 01:16:28,646 [shard 0] iotune - Available space on filesystem at /var/tmp/mnt: 124 MB: is less than recommended: 10 GB
INFO  2020-05-25 01:16:28,646 [shard 0] iotune - /var/tmp/mnt passed sanity checks
Created symlink /etc/systemd/system/multi-user.target.wants/scylla-server.service → /lib/systemd/system/scylla-server.service.
Traceback (most recent call last):
  File "/opt/scylladb/scripts/libexec/scylla_raid_setup", line 98, in <module>
    if dist_name() == 'Ubuntu' and dist_ver() == '14.04':
  File "/opt/scylladb/scripts/scylla_util.py", line 437, in dist_name
    return platform.dist()[0]
AttributeError: module 'platform' has no attribute 'dist'
```

fruch:
@avikivity I think that the update to of dbuild to Fedora 32, is causing this:
https://github.com/scylladb/scylla/commit/a61b3f2d78e0ae30426ad188ee51e96813301b05

I find it weird that we deliver the relocatable python stright from the Fedora distro, I thought we are building it on our own, like what was written in the blog about it...

Any we should think of a way to test/lint all our python code with specific version of python we want to use.

[...]

PR #6509, opened 2020-05-25
scylla_util.py: replace platform.dist() with distro package

fruch:
since dbuild was updated to fedora-32, hence to python3.8
`platform.dist()` is deprecated, and need to be replaced

Fixes: #6501

### DIFF
--- a/reloc/python3/build_reloc.sh
+++ b/reloc/python3/build_reloc.sh
@@ -32,5 +32,5 @@ PYVER=$(python3 -V | cut -d' ' -f2)
 echo "$PYVER" > build/python3/SCYLLA-VERSION-FILE
 ln -fv build/SCYLLA-RELEASE-FILE build/python3/SCYLLA-RELEASE-FILE
 
-PACKAGES="python3-pyyaml python3-urwid python3-pyparsing python3-requests python3-pyudev python3-setuptools python3-psutil"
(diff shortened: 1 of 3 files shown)


reloc/python3/build_reloc.sh
```diff
@@ -32,5 +32,5 @@ PYVER=$(python3 -V | cut -d' ' -f2)
 echo "$PYVER" > build/python3/SCYLLA-VERSION-FILE
 ln -fv build/SCYLLA-RELEASE-FILE build/python3/SCYLLA-RELEASE-FILE
 
-PACKAGES="python3-pyyaml python3-urwid python3-pyparsing python3-requests python3-pyudev python3-setuptools python3-psutil"
+PACKAGES="python3-pyyaml python3-urwid python3-pyparsing python3-requests python3-pyudev python3-setuptools python3-psutil python3-distro"
```
avikivity, line 35: Need to add to install-dependency.sh too

fruch: I've check the image, and it's already installed, but I'll add it to make it explicit

fruch: added it in be8dd05

slivne:
@avikivity ping

avikivity:
Queued as eabcb31503eb9896abeeaefe49cd548012dca001

PR #6510, opened 2020-05-25
Adding pylint excecution to test.py

fruch:
* Adiing pylint run to cover all script in `dist/common/scripts/`
  since those are not executed as part of next at all.
  adding this linting step to help catch breakage caused by changes
  in the disto used.
  in later stage we could cover more of the python files in scylla repo.

* Applying chages recommand by pylint to the files in `dist/common/scripts/`


dist/common/scripts/node_exporter_install
```diff
@@ -20,15 +20,13 @@
...
 import tarfile
-from scylla_util import *
 import argparse
 
-VERSION='0.17.0'
-INSTALL_DIR=scylladir()+'/Prometheus/node_exporter'
+from scylla_util import *
+
+VERSION = '0.17.0'
+INSTALL_DIR = scylladir()+'/Prometheus/node_exporter'
```
avikivity, line 30: Is this related?

dist/common/scripts/node_exporter_install
```diff
@@ -65,7 +63,7 @@ if __name__ == '__main__':
             node_exporter_p.unlink()
         os.symlink('{install_dir}/node_exporter-{version}.linux-amd64/node_exporter'.format(install_dir=INSTALL_DIR, version=VERSION), str(node_exporter_p))
         if is_systemd():
-            node_exporter = systemd_unit('node-exporter.service')
+            node_exporter = SystemdUnit('node-exporter.service')
```
avikivity, line 66: And this?

dist/common/scripts/scylla-housekeeping
```diff
@@ -86,9 +92,9 @@ def version_compare(a, b):
     return parse_version(a) < parse_version(b)
 
 
-def create_uuid_file(fl):
-    with open(args.uuid_file, 'w') as myfile:
-        myfile.write(str(uuid.uuid1()) + "\n")
+def create_uuid_file(uuid_filename):
+    with open(uuid_filename, 'w') as uuid_file:
+        uuid_file.write(str(uuid.uuid1()) + "\n")
```
avikivity, line 98: I'm really confused, I have no idea what this patch does.

dist/common/scripts/node_exporter_install
fruch: It's adding spaces to conform to PEP8
And remove already imported packages, since the star import from scylla_utils already import those.

dist/common/scripts/node_exporter_install
fruch: This is a class and they should be named in PascalCase, again PEP8

dist/common/scripts/scylla-housekeeping
fruch: In this case args from the global scope was used instead the parameter of the file name passed to this function

dist/common/scripts/node_exporter_install
avikivity: The subject is "Adding pylint execution"

This isn't adding pylint execution.

dist/common/scripts/node_exporter_install
avikivity: Ok, I think you're applying pylint recommendations for PEP8, not adding pylint execution.

nyh:
> add pylint run to all script in `dist/common/scripts/`
> since those are not executed as part of next at all.
> 
> adding this linting step to help catch breakage caused by changes
> in the disto used.

I'm afraid I don't understand the goal of this. I'm not familiar with pylint, but its website suggests it checks for coding style and a bunch of bad programming practices. How does this catch "breakage caused by changes in the distro"?
 
Additionally, the commit message doesn't mention whether this just adds running pylint, and everything passes - or did you need to make changes to existing code to have it pass on pylint. If it's the latter, what kind of changes did you need to do, and do you consider these changes really improvements, or annoyances?

dist/common/scripts/node_exporter_install
fruch: hmm, I see no point of adding a test that would just fail... wouldn't it ?

fruch:
> > add pylint run to all script in `dist/common/scripts/`
> > since those are not executed as part of next at all.
> > adding this linting step to help catch breakage caused by changes
> > in the disto used.
> 
> I'm afraid I don't understand the goal of this. I'm not familiar with pylint, but its website suggests it checks for coding style and a bunch of bad programming practices. How does this catch "breakage caused by changes in the distro"?
>
pylint is doing static analysis of the code, but part of the static analysis it's running python
so it can figure out if some import are broken, or if variable are not used, of not assigned to

> Additionally, the commit message doesn't mention whether this just adds running pylint, and everything passes - or did you need to make changes to existing code to have it pass on pylint. If it's the latter, what kind of changes did you need to do, and do you consider these changes really improvements, or annoyances?

It's the latter, it didn't magically passed :)

Lots of them are a bit like compiler warnings, and yes I've considered, and in the pylintrc file I've disable a few:
```
        line-too-long,
        missing-module-docstring,
        missing-function-docstring,
        missing-class-docstring,
        fixme,
        wildcard-import,
        unused-wildcard-import,
        duplicate-code,
        invalid-name,
```

In the end cause of python dynamic nature,  and those scripts are not being run as part of the unittest and not even part of next. we should have something to guard them a bit.
we are using pylint in SCT and in siren-tests for almost a year, and we are very happy with it.

we can later build some unittests for some of those scripts, but this is a first stage.

in later stage we could cover more of the python scripts in the source tree, like in `test/` (alternator-test and redis-test)

I know @ultrabug trying doing something similar, but he tried to introduce `black` for automatic handling of styling, and it was never merged.

nyh:
Thanks @fruch. Can you give an example of an actual problem (not a coding style choice) that pylint found in Scylla?
I'm a always worried about the "looking for the coin under the lamp" syndrome - it's easy to have automated coding-style checks so let's do it. I think automated coding style checks add more annoyance than quality to a project (it's an annoyance because we'll need to revert commits where the patch author forgot to check the coding style). I'm just trying to understand if this pylint adds **more** than just coding-style checks. It seems you are saying the answer is yes, but I'd love to see an example (when the long string of calls finish, I'll take a look at your actual patches :-)).

ultrabug:
@nyh I know we obviously do not share the same point of view and experience on this topic in both our professional and OSS lives but since I'm given the opportunity I'll just add something non technical without the will to start a bikeshedding thread here.

Even if you're not 100% convinced about the purely technical benefits of those tools please consider that they undoubtedly add sanity to your developers and contributors. Maybe you don't feed strongly about this on C++ but on a such versatile and permissive language as Python this is a game changer imho.

I think this is also good (and harmless) education for any team member or contributor. Not to mention that it also eases code reviews and fosters normalizing your code base over time.

Whether you want those tools to be a hard blocker is up to you btw.
You can choose what rule to apply to your project even tho most of the projects I know have CI that checks for this before merging. So if you're not comfortable with it for Scylla, so be it, but that doesn't mean you should block such initiatives in the first place.

Once again, I'm not willing to open a never ending argument and I think you're absolutely right to be cautious about what gets in. I thought appropriate to give another point of view from another experience.

Thanks for reading and considering them!

dist/common/scripts/node_exporter_install
avikivity: Please word the changelog message in a way that isn't confusing.

You wrote "Adding pylint execution to test.py"

I expected a change to test.py that executes pylint.

Either the wording is incorrect, or the change is incorrect (or both). 

My expectation is that the wording matches the change.

kostja:
* Israel Fruchter <notifications@github.com> [20/05/26 01:18]:

I'm happy the multiple suite concept is paying off.


-- 
https://scylladb.com

fruch:
> Thanks @fruch. Can you give an example of an actual problem (not a coding style choice) that pylint found in Scylla?
> I'm always worried about the "looking for the coin under the lamp" syndrome - it's easy to have automated coding-style checks so let's do it. I think automated coding style checks add more annoyance than quality to a project (it's an annoyance because we'll need to revert commits where the patch author forgot to check the coding style). I'm just trying to understand if this pylint adds **more** than just coding-style checks. It seems you are saying the answer is yes, but I'd love to see an example (when the long string of calls finish, I'll take a look at your actual patches :-)).

in the case we were talking about:
```
dist/common/scripts/scylla_util.py:461:11: E1101: Module 'platform' has no 'dist' member (no-member)
```

this kind of error would have failed for us on next, and not propagate into the next stages.

pylint is covering much more then style guidelines, and is highly configurable, you can choose which rules you don't care about, and can ignore them.

dist/common/scripts/node_exporter_install
fruch: I've change the description a bit to explain this change is also applying the pylint recommendation. (otherwise it would be failing)

fruch:
@nyh ping

nyh:
@fruch found this open PR. Please either rebase it, or close it, it's not usable as-is.
I'd also like to hear the developers of these dist/common/* (mainly @syuu1228 ) what they think about this testing and its specific results in this patch. Even if the grand idea of python linting is the best idea since slice bread, someone needs to review the specific way in which you used it in this patch.

fruch:
@penberg @syuu1228, would you guys want to help me with getting this one back on track ?

I think you guys are the most prominent contributors to those part of packaging, and I think we could benefit for catching those issues early

penberg:
@fruch I don't mind adding `pylint` to the build process, but `test.py` is not the right place for it. Something like `ninja pylint` would be fine, and then we can fix the issues incrementally over time.

fruch:
> @fruch I don't mind adding `pylint` to the build process, but `test.py` is not the right place for it. Something like `ninja pylint` would be fine, and then we can fix the issues incrementally over time.

the idea was to hook into the current way tests are being run, and not deem this kind of change in changes in scylla-pkg
and to make this mandatory.

you want to take it step by step, o.k., but without that we are seeing issue with those python files, much much later after this code is already merged into master.

penberg:
@fruch I don't want tests to fail because of `pylint` complaints because I think it's too big of a hammer. Also, people who change the Python scripts are less likely to to even run `test.py` because they're testing the `scylla` executable functionality, not the packages.

The way to get this moving forward is (1) make this `pylint` functionality available to people and Jenkins (as warnings), (2) gradually fix up the issues, and (3) then consider making the check more strict.

nyh:
I already stated my view on this patch above, I don't need to repeat it.

I just want to ask that whatever you do in this patch for the dist/ directory, *please* *please* do not try later to impose the same style requirements also for the test directories, especially test/alternator/*.py. There is significantly more Python code in those tests than in dist/ (currently, 7 times more!), and we want to encourage developers to write more test code quickly and with as little friction as possible. The last thing I'd like to see is a patch being refused - or worse - dequeued (with huge maintainer overhead) - just because it uses a single quote instead of a double quote or vice versa. Yes, I agree that broken test code is a real problem - we wasted a lot of time in the past because of a test that wasn't comprehensive enough and missed real bugs, and tests which were buggy themselves and indicated non-existent bugs. But something which never ever happened (as far as I know) is that a Python style faux-pas broke a test or wasted our time.

Creating stylistic friction for test authors is not a victim-less crime: If you don't believe me check out https://github.com/scylladb/scylla-dtest/pull/1698 where a perfectly valid and useful test was never committed because it failed the style police and the author (me) just moved on.

nyh:
By the way, check this post out: https://pythonspeed.com/articles/pylint/
It suggests that pylint is actually great, and can help you catch real bugs, but the problem is that it also outputs a gazillion warnings on what it thinks is bad style. @fruch if you can get pylint to run a way that only (or *most*) catches real bugs, not style issues, I'll gladly support running it on all our code including tests and even making it mandatory. The post suggests you can configure pylint this way.

fruch:
> I already stated my view on this patch above, I don't need to repeat it.
> 
> I just want to ask that whatever you do in this patch for the dist/ directory, _please_ _please_ do not try later to impose the same style requirements also for the test directories, especially test/alternator/*.py. There is significantly more Python code in those tests than in dist/ (currently, 7 times more!), and we want to encourage developers to write more test code quickly and with as little friction as possible. The last thing I'd like to see is a patch being refused - or worse - dequeued (with huge maintainer overhead) - just because it uses a single quote instead of a double quote or vice versa. Yes, I agree that broken test code is a real problem - we wasted a lot of time in the past because of a test that wasn't comprehensive enough and missed real bugs, and tests which were buggy themselves and indicated non-existent bugs. But something which never ever happened (as far as I know) is that a Python style faux-pas broke a test or wasted our time.
> 
> Creating stylistic friction for test authors is not a victim-less crime: If you don't believe me check out [scylladb/scylla-dtest#1698](https://github.com/scylladb/scylla-dtest/pull/1698) where a perfectly valid and useful test was never committed because it failed the style police and the author (me) just moved on.

I wasn't imposing it on the unittest code.
a for the dtest issue, it's the similar to this issue, as the submitter eventfully going give up.

fruch:
> By the way, check this post out: https://pythonspeed.com/articles/pylint/
> It suggests that pylint is actually great, and can help you catch real bugs, but the problem is that it also outputs a gazillion warnings on what it thinks is bad style. @fruch if you can get pylint to run a way that only (or _most_) catches real bugs, not style issues, I'll gladly support running it on all our code including tests and even making it mandatory. The post suggests you can configure pylint this way.

I was using a carefully select configuration we are using in other repos for more then a year now.
you want to sit down and review each rule together, I'll gladly do that.

fruch:
> @fruch I don't want tests to fail because of `pylint` complaints because I think it's too big of a hammer. Also, people who change the Python scripts are less likely to to even run `test.py` because they're testing the `scylla` executable functionality, not the packages.
> 
> The way to get this moving forward is (1) make this `pylint` functionality available to people and Jenkins (as warnings), (2) gradually fix up the issues, and (3) then consider making the check more strict.

I don't agree it's a big hammer, 
and since I'm not the one going to track those warnings, it's completely your call.

PR #6526, opened 2020-05-27
scylla_current_repo: support different $PRODUCT

fruch:
Support point to the correct download url for
diffrent scylla products

### DIFF
--- a/scripts/scylla_current_repo
+++ b/scripts/scylla_current_repo
@@ -3,7 +3,7 @@
 VERSION=$(./SCYLLA-VERSION-GEN)
 SCYLLA_VERSION=$(cat build/SCYLLA-VERSION-FILE)
 SCYLLA_RELEASE=$(cat build/SCYLLA-RELEASE-FILE)
-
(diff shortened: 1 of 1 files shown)


yaronkaikov:
@fruch Please change the base branch to `next` instead of master

penberg:
Did the old AMI image call this script? I would prefer we fix this in scylla-machine-image.git and just eliminate this script from scylla.git altogether.

fruch:
> Did the old AMI image call this script? I would prefer we fix this in scylla-machine-image.git and just eliminate this script from scylla.git altogether.

Yes it was called:
https://github.com/scylladb/scylla/blob/master/dist/ami/build_ami.sh#L70

penberg:
@fruch Ok, thanks! I queued this as e5da79c211586ba49621858f3b1ac966153b6071, but please open a ticket on scylla-machine-image.git to eliminate this dependency in the future.

fruch:
What do you think about just moving this in to scylla-machine-image ? or maybe scylla-pkg, sine that is the repo owns the downloads directory structure ?

PR #6530, opened 2020-05-28
service/storage_service.cc: Make effective_ownership preemptable

### LINKED ISSUE #6380: REST API storage_service/ownership/<keyspace> (get_effective_ownership) can cause stalls
copy pasting analysis @avikivity did 

```c++

future<std::map<gms::inet_address, float>> storage_service::effective_ownership(sstring keyspace_name) {
    return run_with_no_api_lock([keyspace_name] (storage_service& ss) mutable {
        if (keyspace_name != "") {
            //find throws no such keyspace if it is missing
            const keyspace& ks = ss._db.local().find_keyspace(keyspace_name);
            // This is ugly, but it follows origin
            auto&& rs = ks.get_replication_strategy();  // clang complains about typeid(ks.get_replication_strategy());
            if (typeid(rs) == typeid(locator::local_strategy)) {
                throw std::runtime_error("Ownership values for keyspaces with LocalStrategy are meaningless");
            }
        } else {
            auto non_system_keyspaces = ss._db.local().get_non_system_keyspaces();

            //system_traces is a non-system keyspace however it needs to be counted as one for this process
            size_t special_table_count = 0;
            if (std::find(non_system_keyspaces.begin(), non_system_keyspaces.end(), "system_traces") !=
                    non_system_keyspaces.end()) {
                special_table_count += 1;
            }
            if (non_system_keyspaces.size() > special_table_count) {
                throw std::runtime_error("Non-system keyspaces don't have the same replication settings, effective ownership information is meaningless");
            }
            keyspace_name = "system_traces";
        }
        auto token_ownership = dht::global_partitioner().describe_ownership(ss._token_metadata.sorted_tokens());

        std::map<gms::inet_address, float> final_ownership;

        // calculate ownership per dc
        for (auto endpoints : ss._token_metadata.get_topology().get_datacenter_endpoints()) {
            // calculate the ownership with replication and add the endpoint to the final ownership map
            for (const gms::inet_address& endpoint : endpoints.second) {
                float ownership = 0.0f;
                for (range<token> r : ss.get_ranges_for_endpoint(keyspace_name, endpoint)) {
                    // get_ranges_for_endpoint will unwrap the first range.
                    // With t0 t1 t2 t3, the first range (t3,t0] will be splitted
                    // as (min,t0] and (t3,max]. Skippping the range (t3,max]
                    // we will get the correct ownership number as if the first
                    // range were not splitted.
                    if (!r.end()) {
                        continue;
                    }
                    auto end_token = r.end()->value();
                    if (token_ownership.find(end_token) != token_ownership.end()) {
                        ownership += token_ownership[end_token];
                    }
                }
                final_ownership[endpoint] = ownership;
            }
        }
        return final_ownership;
    });
}
```

The whole thing runs in one task without preemption.

Each iteration calls

```c++
dht::token_range_vector
abstract_replication_strategy::get_ranges(inet_address ep) const {
    dht::token_range_vector ret;
    auto prev_tok = _token_metadata.sorted_tokens().back();
    for (auto tok : _token_metadata.sorted_tokens()) {
        for (inet_address a : calculate_natural_endpoints(tok, _token_metadata)) {
            if (a == ep) {
                insert_token_range_to_sorted_container_while_unwrapping(prev_tok, tok, ret);
                break;
            }
        }
        prev_tok = tok;
    }
    return ret;
}
```

which has this comment:

```c++
    // get_ranges() returns the list of ranges held by the given endpoint.
    // The list is sorted, and its elements are non overlapping and non wrap-around.
    // It the analogue of Origin's getAddressRanges().get(endpoint).
    // This function is not efficient, and not meant for the fast path.
    dht::token_range_vector get_ranges(inet_address ep) const;
```

So maybe just some old-stye optimization is needed, plus add preemption. This shouldn't take two seconds.

amnonh:
A lot is going on when calculating effective ownership.
For each node in the cluster, we need to go over all the ranges belong
to that node and see if that node is the owner or not.

This patch uses futurized loops with do_for_each so it would preempt if
needed.

The patch replaces the current for-loops with do_for_each and do_with
but does not replace the logic.

Fixes #6380

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -1807,33 +1807,57 @@ future<std::map<gms::inet_address, float>> storage_service::effective_ownership(
             }
             keyspace_name = "system_traces";
         }
-        auto token_ownership = dht::token::describe_ownership(ss._token_metadata.sorted_tokens());
 
-        std::map<gms::inet_address, float> final_ownership;
+        // The following loops seems computationally heavy, but it's not as bad.
+        // The upper two simply iterate over all the endpoints by iterating over all the
+        // DC and all the instances in each DC.
+        //
+        // The call for get_range_for_endpoint is done once per endpoint
+        return do_with(dht::token::describe_ownership(ss._token_metadata.sorted_tokens()),
+                ss._token_metadata.get_topology().get_datacenter_endpoints(),
+                std::map<gms::inet_address, float>(),
+                [&ss, keyspace_name](const std::map<token, float>& token_ownership, std::unordered_map<sstring,
+                        std::unordered_set<gms::inet_address>>& datacenter_endpoints,
+                        std::map<gms::inet_address, float>& final_ownership) {
+            return do_for_each(datacenter_endpoints, [&ss, &keyspace_name, &final_ownership, &token_ownership](std::pair<sstring,std::unordered_set<inet_address>>&& endpoints) mutable {
+                return do_with(std::unordered_set<inet_address>(endpoints.second), [&ss, &keyspace_name, &final_ownership, &token_ownership](const std::unordered_set<inet_address>& endpoints_map) mutable {
+                    return do_for_each(endpoints_map, [&ss, &keyspace_name, &final_ownership, &token_ownership](const gms::inet_address& endpoint) mutable {
+                        // calculate the ownership with replication and add the endpoint to the final ownership map
+                        try {
+                            return do_with(float(0.0f), dht::token_range_vector(ss.get_ranges_for_endpoint(keyspace_name, endpoint)),
+                                    [&final_ownership, &token_ownership, &endpoint](float& ownership, const dht::token_range_vector& ranges) mutable {
+                                ownership = 0.0f;
+                                return do_for_each(ranges, [&token_ownership,&ownership](const dht::token_range& r) mutable {
+                                    // get_ranges_for_endpoint will unwrap the first range.
+                                    // With t0 t1 t2 t3, the first range (t3,t0] will be splitted
+                                    // as (min,t0] and (t3,max]. Skippping the range (t3,max]
+                                    // we will get the correct ownership number as if the first
+                                    // range were not splitted.
+                                    if (!r.end()) {
+                                        return make_ready_future<>();
+                                    }
+                                    auto end_token = r.end()->value();
+                                    auto loc = token_ownership.find(end_token);
+                                    if (loc != token_ownership.end()) {
+                                        ownership += loc->second;
+                                    }
+                                    return make_ready_future<>();
+                                }).then([&final_ownership, &ownership, &endpoint]() mutable {
+                                    final_ownership[endpoint] = ownership;
+                                });
+                            });
+                        }  catch (no_such_keyspace&) {
+                            // In case ss.get_ranges_for_endpoint(keyspace_name, endpoint) is not found, just mark it as zero and continue
+                            final_ownership[endpoint] = 0;
+                            return make_ready_future<>();
+                        }
 
-        // calculate ownership per dc
-        for (auto endpoints : ss._token_metadata.get_topology().get_datacenter_endpoints()) {
-            // calculate the ownership with replication and add the endpoint to the final ownership map
-            for (const gms::inet_address& endpoint : endpoints.second) {
-                float ownership = 0.0f;
-                for (range<token> r : ss.get_ranges_for_endpoint(keyspace_name, endpoint)) {
-                    // get_ranges_for_endpoint will unwrap the first range.
-                    // With t0 t1 t2 t3, the first range (t3,t0] will be splitted
-                    // as (min,t0] and (t3,max]. Skippping the range (t3,max]
-                    // we will get the correct ownership number as if the first
-                    // range were not splitted.
-                    if (!r.end()) {
-                        continue;
-                    }
-                    auto end_token = r.end()->value();
-                    if (token_ownership.find(end_token) != token_ownership.end()) {
-                        ownership += token_ownership[end_token];
-                    }
-                }
-                final_ownership[endpoint] = ownership;
-            }
-        }
-        return final_ownership;
+                    });
+                });
+            }).then([&final_ownership] {
+                return final_ownership;
+            });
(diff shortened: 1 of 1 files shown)


service/storage_service.cc
```diff
@@ -1876,33 +1876,56 @@ future<std::map<gms::inet_address, float>> storage_service::effective_ownership(
...
-                    if (token_ownership.find(end_token) != token_ownership.end()) {
-                        ownership += token_ownership[end_token];
-                    }
-                }
-                final_ownership[endpoint] = ownership;
-            }
-        }
-        return final_ownership;
+
+        auto final_ownership = make_lw_shared<std::map<gms::inet_address, float>>();
```
avikivity, line 1880: Why not put final_ownership in the  do_with? you can return it from within the do_with().

service/storage_service.cc
```diff
@@ -1876,33 +1876,56 @@ future<std::map<gms::inet_address, float>> storage_service::effective_ownership(
...
+        // The following loops seems computationally heavy, but it's not as bad.
+        // The upper two simply iterate over all the endpoints by iterating over all the
+        // DC and all the instances in each DC.
+        //
+        // The call for get_range_for_endpoint is done once per endpoint
+        return do_with(std::map<token, float>(dht::token::describe_ownership(ss._token_metadata.sorted_tokens())),
```
avikivity, line 1886: Why the extra constructor call? describe_ownership already returns a map<token, float>.

service/storage_service.cc
amnonh: I had a poor experience with returning values own by do_with in the past, I can try

service/storage_service.cc
amnonh: I'll add a move

service/storage_service.cc
```diff
@@ -1876,33 +1876,56 @@ future<std::map<gms::inet_address, float>> storage_service::effective_ownership(
...
+                [&ss, keyspace_name, final_ownership](const std::map<token, float>& token_ownership, std::unordered_map<sstring,
+                        std::unordered_set<gms::inet_address>>& datacenter_endpoints) {
+            return do_for_each(datacenter_endpoints, [&ss, &keyspace_name, final_ownership, &token_ownership, &datacenter_endpoints](std::pair<sstring,std::unordered_set<inet_address>>&& endpoints) {
+                return do_with(std::unordered_set<inet_address>(endpoints.second), [&ss, &keyspace_name, final_ownership, &token_ownership, &endpoints](const std::unordered_set<inet_address>& endpoints_map){
+                    return do_for_each(endpoints_map, [&ss, &keyspace_name, final_ownership, &token_ownership, &endpoints](const gms::inet_address& endpoint) {
+                        // calculate the ownership with replication and add the endpoint to the final ownership map
+                        try {
+                            return do_with(float(0.0f), dht::token_range_vector(ss.get_ranges_for_endpoint(keyspace_name, endpoint)),
```
avikivity, line 1895: get_ranges_for_endpoint() is also a nested loop!

service/storage_service.cc
```diff
@@ -1876,33 +1876,56 @@ future<std::map<gms::inet_address, float>> storage_service::effective_ownership(
...
+                                        ownership += loc->second;
+                                    }
+                                    return make_ready_future<>();
+                                }).then([final_ownership = final_ownership, &ownership, &endpoint] {
+                                    (*final_ownership)[endpoint] = ownership;
+                                });
+                            });
+                        }  catch (...) {
+                            // In case ss.get_ranges_for_endpoint(keyspace_name, endpoint) is not found, just mark it as zero and continue
+                            (*final_ownership)[endpoint] = 0;
```
avikivity, line 1919: Where did this come from? Just because you had an exception, doesn't mean it's okay to return zero!

service/storage_service.cc
avikivity: Why?

amnonh: To remove the constructor call

service/storage_service.cc
amnonh: it is, but not a too bad one, assuming that we are preempting between each call, I belive it should be ok

service/storage_service.cc
amnonh: get_ranges_for_endpoint throws if there is no keyspace,endpoint match.
Which means that the specific endpoint has zero ownership.
Note that it only catches an exception that happens in get_ranges_for_endpoint.
I can catch only that specific exception to be on the safe side.

service/storage_service.cc
avikivity: It's enough to remove it. No need for an extra move.

service/storage_service.cc
avikivity: `...` is not a specific exception. Any bad_alloc (or some other exception) here will be caught too.

amnonh: right, that's why I wrote that I'll change it to a specific exception so that only the case that there is no keyspace for the endpoint will be caught.

service/storage_service.cc
amnonh: if you are ok with not explicitly stating the type, great

avikivity: I don't understand.

amnonh: I meant that type will be deduce and will not be stated explicitly.

avikivity: The type is explicit in the argument list of the lambda:

```c++
        [&ss, keyspace_name, final_ownership](const std::map<token, float>& token_ownership, 
```

service/storage_service.cc
avikivity: Ah, I somehow missed "can"

amnonh:
@avikivity I have forced-push a new version with the following changes:

* final_ownership is now in the do_with.
* removed unneeded explicit constructors from do_with.
* changed the catch to an explicit exception.

`get_ranges_for_endpoint` remains a loop, but it is inside `do_for_each` so the code will be preempted between calls if needed.

slivne:
@asias please review as well

PR #6534, opened 2020-05-28
dist/debian/python3: drop dependency on pystache

syuu1228:
Same as 287d6e5, we need to drop pystache from package build script
since Fedora 32 dropped it.


syuu1228:
287d6e5 wasn't contain dist/debian/python3 changes, so sending it in another PR.

Note: this will conflict with #6497, we need to rebase after merged it

syuu1228:
Rebased with latest next, since it conflicted with 536ab4e.

PR #6535, opened 2020-05-29
install.sh: support calling install.sh from other directory

syuu1228:
On .deb package with new relocatable package format, all files moved to
under scylla/ directory.
So we need to call ./scylla/install.sh on debian/rules, but it does not work
correctly, since install.sh does not support calling from other directory.

To support this, we need to changedir to scylla top directory before
copying files.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -51,7 +51,7 @@ upgrade=false
 while [ $# -gt 0 ]; do
     case "$1" in
         "--root")
-            root="$2"
+            root="$(realpath "$2")"
             shift 2
             ;;
         "--prefix")
@@ -166,6 +166,9 @@ installconfig() {
         install "-m$perm" "$src" -Dt "$dest"
     fi
 }
+
+# change directory to the package's root directory
+cd "$(dirname "$0")"
 
 if [ -z "$prefix" ]; then
     if $nonroot; then
--- a/dist/debian/debian/rules
+++ b/dist/debian/debian/rules
@@ -20,7 +20,7 @@ override_dh_auto_clean:
 
 override_dh_auto_install:
 	dh_auto_install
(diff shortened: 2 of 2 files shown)


install.sh
```diff
@@ -22,6 +22,9 @@
 
 set -e
 
+# change directory to the package's root directory
+cd $(dirname $0)
```
avikivity, line 26: Please use proper quoting. Both `$0` and the outer expression need to be quoted: `"$(dirname "$0")"`.

avikivity: If --root is given as a relative directory, it will be interpreted according to the directory of install.sh, not $PWD. So we first need to parse the parameters, convert --root to an absolute path, and only then cd.

syuu1228: @avikivity Sorry I do not understand why --root is related with this, isn't it --root specifies target directory, not source directory?

syuu1228: @avikivity Ah okay I understood, will send new version

penberg: @syuu1228 I think you forgot to the part about converting `--root` to an absolute address. IOW, you need something like:

```
diff --git a/install.sh b/install.sh
index d2df63a50f..1cd76a88d1 100755
--- a/install.sh
+++ b/install.sh
@@ -51,7 +51,7 @@ upgrade=false
 while [ $# -gt 0 ]; do
     case "$1" in
         "--root")
-            root="$2"
+            root="$(realpath "$2")"
             shift 2
             ;;
         "--prefix")
```

And you have to do that *before* changing the current directory.

syuu1228:
@avikivity @penberg Fixed quoting and convert --root to absolute path, please review again.

avikivity:
Queued as 5da8784494

penberg:
This is merged to master.

Issue #6478, opened 2020-05-18
Stream ended with unclosed partition and coredump


juliayakovlev:
*Installation details*
Scylla version (or git commit hash): 4.0.0-0.20200505.d95aa77b62
Cluster size: 5 nodes (i3en.3xlarge) 
OS (RHEL/CentOS/Ubuntu/AWS AMI): ami-0ad095d521c760c3f

Test: longevity-large-partition-200k_pks-4days.yaml ([jenkins job](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/yulia/job/Staging-unclosed-partition/34))

Nemesis: DecommissionStreamingErr
```
< t:2020-05-18 13:23:20,848 f:sct_events.py   l:887  c:sdcm.sct_events      p:INFO  > 2020-05-18 13:23:20.836: (DisruptionEvent Severity.NORMAL): type=start name=DecommissionStreamingErr node=Node longevity-large-partitions-200k-pks-db-node-3243f275-2 [54.194.234.7 | 10.0.51.224] (seed: False) duration=None
```

Decommission starts on the node 2 (54.194.234.7):
```
< t:2020-05-18 13:23:22,856 f:cluster.py      l:1409 c:sdcm.cluster         p:DEBUG > -05-18T13:23:21+00:00  longevity-large-partitions-200k-pks-db-node-3243f275-2 !INFO    | scylla: [shard 0] storage_service - DECOMMISSIONING: starts
```

By test, decommission was interrupted:
```
< t:2020-05-18 13:29:44,289 f:nemesis.py      l:1796 c:sdcm.nemesis         p:INFO  > sdcm.nemesis.ChaosMonkey: Interrupt the task (decommission) to trigger streaming error
< t:2020-05-18 13:29:44,836 f:nemesis.py      l:1771 c:sdcm.nemesis         p:DEBUG > sdcm.nemesis.ChaosMonkey: decommission is stopped
```

Node was hardly rebooted:
```
< t:2020-05-18 13:29:44,317 f:cluster.py      l:1072 c:sdcm.cluster_aws     p:DEBUG > Node longevity-large-partitions-200k-pks-db-node-3243f275-2 [54.194.234.7 | 10.0.51.224] (seed: False): Hardly rebooting node
```

Node is up:
```
[...]

PR #6536, opened 2020-05-29
streaming: Do not send end of stream in case of error

asias:
Current sender sends stream_mutation_fragments_cmd::end_of_stream to
receiver when an error is receiver from peer node. To be safe, send
stream_mutation_fragments_cmd::error instead of
stream_mutation_fragments_cmd::end_of_stream to prevent end_of_stream to
be written into the sstable when a partition is not closed yet.

In addition, check if partition_start and partition_end emitted from
reader are paried when the reader is done. If not, fail the stream
session and send stream_mutation_fragments_cmd::error instead of
stream_mutation_fragments_cmd::end_of_stream to isolate the problematic
sstables on the sender node.

Refs: #6478

### DIFF
--- a/streaming/stream_transfer_task.cc
+++ b/streaming/stream_transfer_task.cc
@@ -44,6 +44,7 @@
 #include "streaming/stream_reason.hh"
 #include "streaming/stream_mutation_fragments_cmd.hh"
 #include "mutation_reader.hh"
+#include "flat_mutation_reader.hh"
 #include "frozen_mutation.hh"
 #include "mutation.hh"
 #include "message/messaging_service.hh"
@@ -203,15 +204,27 @@ future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
         }();
 
         auto sink_op = [sink, si, got_error_from_peer] () mutable -> future<> {
-            return do_with(std::move(sink), [si, got_error_from_peer] (rpc::sink<frozen_mutation_fragment, stream_mutation_fragments_cmd>& sink) {
-                return repeat([&sink, si, got_error_from_peer] () mutable {
-                    return si->reader(db::no_timeout).then([&sink, si, s = si->reader.schema(), got_error_from_peer] (mutation_fragment_opt mf) mutable {
-                        if (mf && !(*got_error_from_peer)) {
+            mutation_fragment_stream_validator validator(*(si->reader.schema()));
+            return do_with(std::move(sink), std::move(validator), [si, got_error_from_peer] (rpc::sink<frozen_mutation_fragment, stream_mutation_fragments_cmd>& sink, mutation_fragment_stream_validator& validator) {
+                return repeat([&sink, &validator, si, got_error_from_peer] () mutable {
+                    return si->reader(db::no_timeout).then([&sink, &validator, si, s = si->reader.schema(), got_error_from_peer] (mutation_fragment_opt mf) mutable {
+                        if (*got_error_from_peer) {
+                            return make_exception_future<stop_iteration>(std::runtime_error("Got status error code from peer"));
+                        }
+                        if (mf) {
+                            if (!validator(mf->mutation_fragment_kind())) {
+                                return make_exception_future<stop_iteration>(std::runtime_error(format("Stream reader mutation_fragment validator failed, previous={}, current={}",
+                                        validator.previous_mutation_fragment_kind(), mf->mutation_fragment_kind())));
+                            }
                             frozen_mutation_fragment fmf = freeze(*s, *mf);
                             auto size = fmf.representation().size();
                             streaming::get_local_stream_manager().update_progress(si->plan_id, si->id.addr, streaming::progress_info::direction::OUT, size);
                             return sink(fmf, stream_mutation_fragments_cmd::mutation_fragment_data).then([] { return stop_iteration::no; });
                         } else {
+                            if (!validator.on_end_of_stream()) {
+                                return make_exception_future<stop_iteration>(std::runtime_error(format("Stream reader mutation_fragment validator failed on end_of_stream, previous={}, current=end_of_stream",
+                                        validator.previous_mutation_fragment_kind())));
+                            }
                             return make_ready_future<stop_iteration>(stop_iteration::yes);
                         }
                     });


slivne:
@asias 
- who should review this ?
- can we try and simulate this case and see that we add an incorrect state ?

slivne:
@asias ping ... can we try and simulate this ?

asias:
> @asias
> 
>     * who should review this ?
> 
>     * can we try and simulate this case and see that we add an incorrect state ?

The receiver side can already detect such end of stream with unpaired partition start and partition end. That's why we see "stream ended with unclosed partition and coredump #6478".

This patch detects and isolates such error on the sender side.

streaming/stream_transfer_task.cc
```diff
@@ -203,15 +203,27 @@ future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
         }();
 
         auto sink_op = [sink, si, got_error_from_peer] () mutable -> future<> {
-            return do_with(std::move(sink), [si, got_error_from_peer] (rpc::sink<frozen_mutation_fragment, stream_mutation_fragments_cmd>& sink) {
-                return repeat([&sink, si, got_error_from_peer] () mutable {
-                    return si->reader(db::no_timeout).then([&sink, si, s = si->reader.schema(), got_error_from_peer] (mutation_fragment_opt mf) mutable {
-                        if (mf && !(*got_error_from_peer)) {
+            bool partition_opened = false;
+            return do_with(std::move(sink), partition_opened, [si, got_error_from_peer] (rpc::sink<frozen_mutation_fragment, stream_mutation_fragments_cmd>& sink, bool& partition_opened) {
```
denesb, line 207: You can just replace `partition_opened` here with `false` and then you don't have to declare it above.

streaming/stream_transfer_task.cc
```diff
@@ -203,15 +203,27 @@ future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
...
+                    return si->reader(db::no_timeout).then([&sink, &partition_opened, si, s = si->reader.schema(), got_error_from_peer] (mutation_fragment_opt mf) mutable {
+                        if (*got_error_from_peer) {
+                            return make_exception_future<stop_iteration>(std::runtime_error("Got status error code from peer"));
+                        }
+                        if (mf) {
+                            if (mf->is_partition_start()) {
+                                partition_opened = true;
+                            } else if (mf->is_end_of_partition()) {
+                                partition_opened = false;
+                            }
```
denesb, line 218: It would be simpler to replace `partition_opened` with `last_mutation_fragment_kind` and just do:
```C++
last_mutation_fragment_kind = mf->mutation_fragment_kind();
```
It would avoid two branches and thus would be less error-prone.

streaming/stream_transfer_task.cc
```diff
@@ -203,15 +203,27 @@ future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
...
                             frozen_mutation_fragment fmf = freeze(*s, *mf);
                             auto size = fmf.representation().size();
                             streaming::get_local_stream_manager().update_progress(si->plan_id, si->id.addr, streaming::progress_info::direction::OUT, size);
                             return sink(fmf, stream_mutation_fragments_cmd::mutation_fragment_data).then([] { return stop_iteration::no; });
                         } else {
+                            if (partition_opened) {
```
denesb, line 224: ... and replace this with:
```c++
if (last_mutation_fragment_kind != mutation_fragment::kind::partition_end) {
```

denesb:
If I understood right, this check is on the sender side. While I don't think it is unnecessary, I think we better place this check it on the receiver side too, as just checking on the sender side doesn't cover any potential problems in-transit.

denesb:
Ok, I just saw @asias's comment above, the receiver already has this check.

streaming/stream_transfer_task.cc
asias: Such check would be wrong if the reader emits zero partition, no?

streaming/stream_transfer_task.cc
asias: Yes we can, but I intentionally declare the variable to give the false more context.

streaming/stream_transfer_task.cc
denesb: Not if you initialize `last_mutation_fragment_kind` to `mutation_fragment::kind::partition_end`.

asias:
> Ok, I just saw @asias's comment above, the receiver already has this check.

Yes, the checker on the receiver side is the one you wrote on the write path. We do not have the checker at streaming level on receiver.

streaming/stream_transfer_task.cc
denesb: Ok, fair enough, I will not complain about more context.

streaming/stream_transfer_task.cc
asias: Yes, this should work. Sounds a bit weird to me though because we do not get partition_end from the reader.

streaming/stream_transfer_task.cc
asias: I am thinking about if we should add more checks like below:

```
if (mf->is_partition_start()) {
   // Already opened?
   if (partition_opened) {
          throw
   }
    partition_opened = true;
} else if (mf->is_end_of_partition()) {
   // Already closed ?
    if (!partition_opened) {
          throw
    }
    partition_opened = false;
}
                                        

```

denesb: This is starting to look like the `mutation_fragment_stream_validator` open-coded :).

streaming/stream_transfer_task.cc
denesb: It is weird on first read, but if you think about it, if you haven't read anything yet, you are positioned before a partition start, and `partition_end` is the only position which expresses this.

streaming/stream_transfer_task.cc
asias: Is it possible to use mutation_fragment_stream_validator here?

denesb: Yes. You would have to use the lower level validator, which is the one actually called `mutation_fragment_stream_validator` (the higher level one is called `mutation_fragment_stream_validating_filter`).
You can use it like this:
```c++
mutation_fragment_stream_validator validator(schema);
//...
// validates that the fragment kind montonicity, e.g. that after partition end the next fragment is partition start and not anything else.
if (!validator(mf->mutation_fragment_kind())) {
    throw;
}

// ... finally
if (!validator.on_end_of_stream()) {
    throw;
}
```
Note that the low level validator just returns `bool` from its methods and you have to do the error handling.

denesb: This should be pretty cheap as it only validates the fragment kind validity, i.e. it will only compares enum values no keys are compared.

asias: Thanks. This looks good. If it is very cheap, I am thinking about if we should add the check inside  flat mutation reader.

denesb: That would be quite redundant. We often build stack of readers half-a-dozen readers deep. This validation is cheap but not free, it would add up to be visible.

asias:
Changes in 103f42cd49f6ba36907da62caf1b6d17d5981cfe (v2):
- Use mutation_fragment_stream_validator to valid

streaming/stream_transfer_task.cc
asias: Make sense.

asias: I am using the mutation_fragment_stream_validator in v2. Thanks!

streaming/stream_transfer_task.cc
```diff
@@ -203,15 +204,27 @@ future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
...
+                                        validator.previous_mutation_fragment_kind(), mf->mutation_fragment_kind())));
+                            }
                             frozen_mutation_fragment fmf = freeze(*s, *mf);
                             auto size = fmf.representation().size();
                             streaming::get_local_stream_manager().update_progress(si->plan_id, si->id.addr, streaming::progress_info::direction::OUT, size);
                             return sink(fmf, stream_mutation_fragments_cmd::mutation_fragment_data).then([] { return stop_iteration::no; });
                         } else {
+                            if (!validator.on_end_of_stream()) {
+                                return make_exception_future<stop_iteration>(std::runtime_error(format("Stream reader mutation_fragment validator failed on end_of_stream, previous={}, current={}",
+                                        validator.previous_mutation_fragment_kind(), mf->mutation_fragment_kind())));
```
denesb, line 226: If we are in this branch `mf` is not engaged.

asias: Are you talking about the case reader emits zero fragment?

This branch means we are in the end of stream. 

We have:

```
 953 mutation_fragment_stream_validator::mutation_fragment_stream_validator(const schema& s)
 954     : _schema(s)
 955     , _prev_kind(mutation_fragment::kind::partition_end)
```

and 

```
bool mutation_fragment_stream_validator::on_end_of_stream() {
    return _prev_kind == mutation_fragment::kind::partition_end;
}
```

so, this is ok even if reader returns zero mf.

denesb: No, I'm talking about this: `mf->mutation_fragment_kind()`, passed to `format()`.

asias: Right! I will fix it.

asias:
Changes in v3 (0b55c6eb11c62010865affec5c8275788d63cf45):
- Fix log in end of stream branch

streaming/stream_transfer_task.cc
```diff
@@ -203,15 +204,27 @@ future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
...
         auto sink_op = [sink, si, got_error_from_peer] () mutable -> future<> {
-            return do_with(std::move(sink), [si, got_error_from_peer] (rpc::sink<frozen_mutation_fragment, stream_mutation_fragments_cmd>& sink) {
-                return repeat([&sink, si, got_error_from_peer] () mutable {
-                    return si->reader(db::no_timeout).then([&sink, si, s = si->reader.schema(), got_error_from_peer] (mutation_fragment_opt mf) mutable {
-                        if (mf && !(*got_error_from_peer)) {
+            mutation_fragment_stream_validator validator(*(si->reader.schema()));
+            return do_with(std::move(sink), std::move(validator), [si, got_error_from_peer] (rpc::sink<frozen_mutation_fragment, stream_mutation_fragments_cmd>& sink, mutation_fragment_stream_validator& validator) {
+                return repeat([&sink, &validator, si, got_error_from_peer] () mutable {
+                    return si->reader(db::no_timeout).then([&sink, &validator, si, s = si->reader.schema(), got_error_from_peer] (mutation_fragment_opt mf) mutable {
+                        if (*got_error_from_peer) {
```
avikivity, line 211: This is a poor API that requires you to check for an error, and if you forget, it corrupts data. Can we replace it with something that throws an exception?

streaming/stream_transfer_task.cc
```diff
@@ -203,15 +204,27 @@ future<> send_mutation_fragments(lw_shared_ptr<send_info> si) {
         }();
 
         auto sink_op = [sink, si, got_error_from_peer] () mutable -> future<> {
-            return do_with(std::move(sink), [si, got_error_from_peer] (rpc::sink<frozen_mutation_fragment, stream_mutation_fragments_cmd>& sink) {
-                return repeat([&sink, si, got_error_from_peer] () mutable {
-                    return si->reader(db::no_timeout).then([&sink, si, s = si->reader.schema(), got_error_from_peer] (mutation_fragment_opt mf) mutable {
-                        if (mf && !(*got_error_from_peer)) {
+            mutation_fragment_stream_validator validator(*(si->reader.schema()));
```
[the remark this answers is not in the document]
asias: The check is pretty cheap because it only checks the kind of the fragment.

streaming/stream_transfer_task.cc
asias: This is the limitation of the rpc. Assume n1 is the sender and n2 is the receiver. The stream (s1) n1 to n2 is used to send data. We need another stream (s2) to send the error from n2 to n1. 
The loop for s1 and s2 works inside different fibers. How do we use exception  in this case to signal the error between the fibers?

slivne:
@avikivity ping ...

streaming/stream_transfer_task.cc
avikivity: Agree it's an rpc bug. I think we'll need to communicate errors on the main rpc connection, not via the streams. This will need an updated API. 

@gleb-cloudius

PR #6544, opened 2020-06-01
fix(scylla_prepare): missing platform import

fruch:
as part of eabcb31503eb9896abeeaefe49cd548012dca001 `import platform`
was removed from scylla_utils.py

seem like we missed it's usage in scylla_prepare script

### DIFF
--- a/dist/common/scripts/scylla_prepare
+++ b/dist/common/scripts/scylla_prepare
@@ -23,6 +23,8 @@
 import os
 import sys
 import glob
+import platform
+
 from scylla_util import *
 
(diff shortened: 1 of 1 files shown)


nyh:
Thanks. Committed to next (cd96202dcb4e435dcbb7fe0e0a0c11edd35632b9).
Does this need backporting (i.e., did the change which broke this also go into 4.1?)

fruch:
@nyh I think the fedora-32 dbulid wasn't backported, since artifacts test are passing on 4.1

PR #6546, opened 2020-06-01
Debian version number fix

syuu1228:
Now we generate dist/changelog on relocatable package generation time,
we cannot run '.rc' fixup on .deb package building time, need to do it
in debian_files_gen.py.

Also, we uses '_'  in version number for some test version packages,
which does not supported in .deb packaging system, need to replaced
with '-'.


penberg:
@syuu1228 @avikivity Is this different from the commits that Avi had to zap from `next`?

syuu1228:
@penberg I think only scylla-jmx PR is broken, since it mistakenly calling f.read() twice: https://github.com/scylladb/scylla-jmx/pull/108/files#diff-8536101365d8ba6f66e422b5838aa09eR44

syuu1228:
rebased with latest next

PR #6547, opened 2020-06-01
reloc: don't build anything in build_reloc.sh

### LINKED ISSUE #6477: build_reloc.sh should not run configure.py
Even if you already built Scylla yourself, `reloc/build_reloc.sh` still runs `configure.py` with slightly different parameters and causes a recompile of everything. This is not just frustrating for purists, it is also genuinely annoying because the new build can be very slow - even slower than usual (e.g., if the user usually builds using distcc). 

@penberg has been working to make sure that the various options which `build_reloc.sh` sets with `configure.py` all move to be become defaults or things that ninja itself can do instead of through `configure.py`. But apparently we're not done with that so I opened this issue. The end result should be that `build_reloc.sh` should NOT run `configure.py`. It probably shouldn't run `ninja` either, but running `ninja` again after the user already did is harmless.

syuu1228:
Remove configure.py and ninja from build_reloc.sh, let it done by user.

Fixes #6477

We require to run following commit after the commit:
```
./install-dependencies.sh
./configure.py --mode release
ninja-build
./reloc/build_reloc.sh
```

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1412,8 +1412,6 @@ with open(buildfile_tmp, 'w') as f:
         rule copy
             command = cp $in $out
             description = COPY $out
-        rule package
-            command = scripts/create-relocatable-package.py --mode $mode $out
         ''').format(**globals()))
     for mode in build_modes:
         modeval = modes[mode]
@@ -1654,9 +1652,6 @@ with open(buildfile_tmp, 'w') as f:
         f.write(textwrap.dedent('''\
             build build/{mode}/iotune: copy build/{mode}/seastar/apps/iotune/iotune
             ''').format(**locals()))
-        f.write('build build/{mode}/scylla-package.tar.gz: package build/{mode}/scylla build/{mode}/iotune build/SCYLLA-RELEASE-FILE build/SCYLLA-VERSION-FILE build/debian/debian | always\n'.format(**locals()))
-        f.write('  pool = submodule_pool\n')
-        f.write('  mode = {mode}\n'.format(**locals()))
         f.write('rule libdeflate.{mode}\n'.format(**locals()))
         f.write('  command = make -C libdeflate BUILD_DIR=../build/{mode}/libdeflate/ CFLAGS="{libdeflate_cflags}" CC={args.cc} ../build/{mode}/libdeflate//libdeflate.a\n'.format(**locals()))
         f.write('build build/{mode}/libdeflate/libdeflate.a: libdeflate.{mode}\n'.format(**locals()))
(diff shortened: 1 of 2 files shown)


configure.py
```diff
@@ -1654,9 +1652,6 @@ def configure_zstd(build_dir, mode):
         f.write(textwrap.dedent('''\
             build build/{mode}/iotune: copy build/{mode}/seastar/apps/iotune/iotune
             ''').format(**locals()))
-        f.write('build build/{mode}/scylla-package.tar.gz: package build/{mode}/scylla build/{mode}/iotune build/SCYLLA-RELEASE-FILE build/SCYLLA-VERSION-FILE build/debian/debian | always\n'.format(**locals()))
-        f.write('  pool = submodule_pool\n')
-        f.write('  mode = {mode}\n'.format(**locals()))
```
avikivity, line 1659: Why are you removing this? It's still nice to be able to create a package using a ninja target.

In fact we may want to make rpm/deb be a ninja target.

reloc/build_reloc.sh
```diff
@@ -2,101 +2,49 @@
...
-            shift 2
-            ;;
-        "--c-compiler")
-            CCOMPILER=$2
-            shift 2
-            ;;
-        "--nodeps")
-            shift 1
-            ;;
         *)
```
avikivity, line 21: Agree with this.

Is scylla-pkg ready to accept this?

reloc/build_reloc.sh
```diff
@@ -2,101 +2,49 @@
...
     exit 1
 fi
+
+if [ ! -e build/SCYLLA-RELEASE-FILE ] || [ ! -e build/SCYLLA-VERSION-FILE ]; then
+    ./SCYLLA-VERSION-GEN
+fi
+
+python3 -m compileall ./dist/common/scripts/ ./seastar/scripts/perftune.py ./tools/scyllatop
+./dist/debian/debian_files_gen.py
+./scripts/create-relocatable-package.py --mode $MODE build/$MODE/scylla-package.tar.gz
```
avikivity, line 50: I think we should keep the call to create-relocatable-package in ninja. We might even move rpm/dev there, so to build rpm/deb you can write `./configure.py; ninja release-rpm release-deb`.

configure.py
```diff
@@ -1697,12 +1692,6 @@ def configure_zstd(build_dir, mode):
         ''').format(modes_list=' '.join(default_modes), **globals()))
     f.write(textwrap.dedent('''\
         build always: phony
-        rule scylla_version_gen
-            command = ./SCYLLA-VERSION-GEN
-        build build/SCYLLA-RELEASE-FILE build/SCYLLA-VERSION-FILE: scylla_version_gen
-        rule debian_files_gen
-            command = ./dist/debian/debian_files_gen.py
-        build build/debian/debian: debian_files_gen | always
```
nyh, line 1705: How is removing this stuff related to fixing isue 6477?

reloc/build_reloc.sh
```diff
@@ -2,101 +2,49 @@
...
-fi
-
-NINJA=$(which ninja-build) &&:
-if [ -z "$NINJA" ]; then
-    NINJA=$(which ninja) &&:
-fi
-if [ -z "$NINJA" ]; then
-    echo "ninja not found."
+if [ ! -e build/$MODE/scylla ] || [ ! -e build/$MODE/iotune ]; then
+    echo "run ./configure.py && ninja-build before running build_reloc.sh"
```
nyh, line 33: Nitpick: the name "ninja-build" is an archaic redhat-ism. On modern Fedora, and everywhere else, it is called just "ninja", not "ninja-build".

PR #6549, opened 2020-06-01
reloc: simplified .rpm build process

syuu1228:
We don't really need to have two build_rpm.sh, merge it to reloc.


reloc/build_rpm.sh
```diff
@@ -49,11 +45,36 @@ if [ ! -e $RELOC_PKG ]; then
     exit 1
 fi
 RELOC_PKG=$(readlink -f $RELOC_PKG)
-if [[ ! $OPTS =~ --reloc-pkg ]]; then
-    OPTS="$OPTS --reloc-pkg $RELOC_PKG"
-fi
 mkdir -p $BUILDDIR/scylla-package
 tar -C $BUILDDIR/scylla-package -xpf $RELOC_PKG SCYLLA-RELOCATABLE-FILE SCYLLA-RELEASE-FILE SCYLLA-VERSION-FILE SCYLLA-PRODUCT-FILE dist/redhat
```
avikivity, line 49: Not related to this patch: we should move to a directory with a random name, so if we move package builds to ninja, we can build debug and release packages in parallel, without the two builds interfering.

install-dependencies.sh
```diff
@@ -90,6 +90,7 @@ fedora_packages=(
     glibc-langpack-en
     lld
     xxhash-devel
+    rpm-build
```
avikivity, line 93: Not strictly necessary since it's in tools/toolchain/Dockerfile (but okay to have it here).

avikivity:
Dequeued, fails to build.

PR #6550, opened 2020-06-01
reloc: simplified .deb build process

syuu1228:
We don't really need to have two build_deb.sh, merge it to reloc.


avikivity:
Queued as e9871956ca

avikivity:
Dequeued, doesn't build.

PR #6551, opened 2020-06-01
api: Command to check and repair cdc streams

### LINKED ISSUE #6498: Create a nodetool command to check CDC streams and regenerate them if needed.
This will be a way to check that CDC streams are correct and repair them in case something went wrong.

jul-stas:
The command regenerates streams when:
- generations corresponding to a gossiped timestamp cannot be
fetched from `system_distributed` table,
- or when generation token ranges do not align with token metadata.

In such case the streams are regenerated and new timestamp is
gossiped around. The returned JSON is always empty, regardless of
whether streams needed regeneration or not.

Fixes #6498
Accompanied by: scylladb/scylla-jmx#109, scylladb/scylla-tools-java#172

### DIFF
--- a/api/api-doc/storage_service.json
+++ b/api/api-doc/storage_service.json
@@ -512,6 +512,21 @@
          ]
       },
       {
+         "path":"/storage_service/cdc_streams_check_and_repair",
+         "operations":[
+            {
+               "method":"POST",
+               "summary":"Checks that CDC streams reflect current cluster topology and regenerates them if not.",
+               "type":"void",
+               "nickname":"cdc_streams_check_and_repair",
+               "produces":[
+                  "application/json"
+               ],
+               "parameters":[]
+            }
+         ]
+      },
+      {
          "path":"/storage_service/snapshots",
          "operations":[
             {
--- a/api/storage_service.cc
+++ b/api/storage_service.cc
@@ -42,6 +42,8 @@
 #include "database.hh"
 #include "db/extensions.hh"
 
+extern logging::logger apilog;
+
 namespace api {
 
 namespace ss = httpd::storage_service_json;
@@ -230,6 +232,11 @@ void set_storage_service(http_context& ctx, routes& r) {
         auto keyspace = validate_keyspace(ctx, req.param);
         return container_to_vec(service::get_local_storage_service().get_natural_endpoints(keyspace, req.get_query_param("cf"),
                 req.get_query_param("key")));
+    });
+
+    ss::cdc_streams_check_and_repair.set(r, [&ctx] (std::unique_ptr<request> req) {
+        // TODO(JS): regenerate streams
+        return make_ready_future<json::json_return_type>(json_void());
     });
 
     ss::force_keyspace_compaction.set(r, [&ctx](std::unique_ptr<request> req) {
--- a/api/storage_service.cc
+++ b/api/storage_service.cc
@@ -42,8 +42,6 @@
 #include "database.hh"
 #include "db/extensions.hh"
 
-extern logging::logger apilog;
-
 namespace api {
 
 namespace ss = httpd::storage_service_json;
@@ -235,8 +233,9 @@ void set_storage_service(http_context& ctx, routes& r) {
     });
 
     ss::cdc_streams_check_and_repair.set(r, [&ctx] (std::unique_ptr<request> req) {
-        // TODO(JS): regenerate streams
-        return make_ready_future<json::json_return_type>(json_void());
+        return service::get_local_storage_service().check_and_repair_cdc_streams().then([] {
+            return make_ready_future<json::json_return_type>(json_void());
+        });
     });
 
     ss::force_keyspace_compaction.set(r, [&ctx](std::unique_ptr<request> req) {
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -830,6 +830,100 @@ void storage_service::scan_cdc_generations() {
     }
 }
 
+future<> storage_service::check_and_repair_cdc_streams() {
+    return async([this] { 
+        auto latest = _cdc_streams_ts;
+        const auto& endpoint_states = _gossiper.get_endpoint_states();
+        for (const auto& [addr, state] : endpoint_states) {
+            if (!_gossiper.is_normal(addr))  {
+                throw std::runtime_error(format("All nodes must be in NORMAL state while performing check_and_repair_cdc_streams"
+                        " ({} is in state {})", addr, _gossiper.get_gossip_status(state)));
+            }
+
+            const auto ts = cdc::get_streams_timestamp_for(addr, _gossiper);
+            if (!latest || (ts && *ts > *latest)) {
+                latest = ts;
+            }
+        }
+
+        bool should_regenerate = false;
+        std::optional<cdc::topology_description> gen;
+
+        static const auto timeout_msg = "Timeout while fetching CDC topology description";
+        static const auto topology_read_error_note = "Note: this is likely caused by"
+                " node(s) being down or unreachable. It is recommended to check the network and"
+                " restart/remove the failed node(s), then retry checkAndRepairCdcStreams command";
+        static const auto exception_translating_msg = "Translating the exception to `request_execution_exception`";
+        try {
+            gen = _sys_dist_ks.local().read_cdc_topology_description(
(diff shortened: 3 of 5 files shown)


service/storage_service.cc
```diff
@@ -830,6 +830,68 @@ void storage_service::scan_cdc_generations() {
...
+
+            const auto ts = cdc::get_streams_timestamp_for(addr, _gossiper);
+            if (!latest || (ts && *ts > *latest)) {
+                latest = ts;
+            }
+        }
+
+        bool should_regenerate = false;
+        const auto gen = _sys_dist_ks.local().read_cdc_topology_description(
+                *latest, { _token_metadata.count_normal_token_owners() }).get0();
```
kbr-, line 851: One thing I would add there.
`read_cdc_topology_description` might  throw and error, for example when:
- replicas are unreachable due to a network partition: timeout, unavailable exception
- the table is corrupted: stuff like "read expected size mismatch", see https://github.com/scylladb/scylla/issues/6537

There is an expected class of errors (timeout etc.) and there is an unexpected class of errors that occur due to bugs (table corruption etc.).

In the expected class we should return the error to the user and finish the procedure.
So, for example, if replicas are unreachable due to a network partition, so `read_cdc_topology_description` throws a timeout, we should inform the user and return.
Although that's debatable. Maybe it would be safer to continue and create a new generation nevertheless.

In the unexpected class we should definitely create a new generation. For example, if the latest generation entry is corrupted in the table, CDC won't work unless we create a new gen.

So, the `read_cdc_topology_description` call should be wrapped in a `try...catch`, we should catch unexpected errors such as `read_simple_exactly - size mismatch`, report them using `cdc_log.error(...)`, set `should_regenerate = true` and continue.

haaawk:
@jul-stas Could you address @kbr- comments?

jul-stas:
@kbr- What would you say about such an approach: https://gist.github.com/jul-stas/6723d5e4a961b2fc7d6114759d23d261 ?

(I created a gist to avoid unnecessary messing with force-pushes. That's mostly because I'm unsure where to draw the line between "expected exception" and "unexpected exception".)

kbr-:
@jul-stas looks good!

haaawk:
> @jul-stas looks good!

You mean the approach not the current PR? @kbr-

kbr-:
The approach

jul-stas:
@kbr-  Added gist to the PR.

kbr-:
I need to perform some manual tests and then I'll approve this PR.

service/storage_service.cc
```diff
@@ -830,6 +830,84 @@ void storage_service::scan_cdc_generations() {
...
+                    break;
+                }
+            }
+        }
+
+        if (!should_regenerate) {
+            if (latest != _cdc_streams_ts) {
+                do_handle_cdc_generation(*latest);
+            }
+            cdc_log.info("CDC generation does not need repair");
```
kbr-, line 891: Optional: I would add here the generation's timestamp.
`cdc_log.info("CDC generation {} does not need repair", latest)`.

service/storage_service.cc
```diff
@@ -830,6 +830,84 @@ void storage_service::scan_cdc_generations() {
...
+
+        bool should_regenerate = false;
+        std::optional<cdc::topology_description> gen;
+        try {
+            gen = _sys_dist_ks.local().read_cdc_topology_description(
+                    *latest, { _token_metadata.count_normal_token_owners() }).get0();
+        } catch (exceptions::request_timeout_exception& e) {
+            throw; // Stop and inform the API caller about timeout
+        } catch (exceptions::unavailable_exception& e) {
+            throw; // Stop and inform the API caller about unavailability
```
kbr-, line 857: I checked how it works in practice:
```
nodetool: Scylla API server HTTP POST to URL '/storage_service/cdc_streams_check_and_repair' failed: exceptions::unavailable_exception (Cannot achieve consistency level for cl QUORUM. Requires 2, alive 1)
See 'nodetool help' or 'nodetool help <command>'.
```

I'm afraid it could leave the user very confused. What happened? Why didn't it work? I didn't do any QUORUM requests, why is it mentioning QUORUM?

We should wrap the error, adding a comment for the user, with some guidance what to do next.

For example, we could tell him that the request failed because either some nodes are down or unreachable due to a networking issue, and that he should fix the network/restart the node and retry the command. If a node is unrecoverable, the user should first remove the node from the cluster and then retry the command.

The specific internal error should also be printed, but it's information for field engineers/developers more than for the user himself.

(this review comment applies to the timeout cases too).

service/storage_service.cc
```diff
@@ -830,6 +830,84 @@ void storage_service::scan_cdc_generations() {
...
+            cdc_log.info("CDC generation does not need repair");
+            return;
+        }
+        const auto new_streams_ts = cdc::make_new_cdc_generation(db().local().get_config(),
+                {}, _token_metadata, _gossiper,
+                _sys_dist_ks.local(), get_ring_delay(), false /* for_testing */);
+        // Need to artificially update our STATUS so other nodes handle the timestamp change
+        auto status = _gossiper.get_application_state_ptr(get_broadcast_address(), application_state::STATUS);
+        if (!status) {
+            return;
```
kbr-, line 900: `!status` should never happen, but we can be defensive and say something in the log (using `slogger` too, because it wouldn't be a CDC specific problem), e.g.:
```
slogger.error("Our STATUS is missing");
cdc_log.error("Aborting CDC generation repair due to missing STATUS");
return;
```

service/storage_service.cc
```diff
@@ -830,6 +830,84 @@ void storage_service::scan_cdc_generations() {
...
+                latest, db_clock::now());
+            should_regenerate = true;
+        } else {
+            std::unordered_set<dht::token> gen_ends;
+            for (const auto& entry : gen->entries()) {
+                gen_ends.insert(entry.token_range_end);
+            }
+            for (const auto& metadata_token : _token_metadata.sorted_tokens()) {
+                if (!gen_ends.count(metadata_token)) {
+                    should_regenerate = true;
```
kbr-, line 881: Optional: we could inform in the log what caused the regeneration. E.g.:
`cdc_log.warn("CDC generation {} missing token {}. Regenerating.", latest, metadata_token)`.

api/storage_service.cc
```diff
@@ -42,6 +42,8 @@
 #include "database.hh"
 #include "db/extensions.hh"
 
+extern logging::logger apilog;
```
kbr-, line 45: It looks like we don't use this anywhere.

service/storage_service.cc
jul-stas: done

service/storage_service.cc
jul-stas: done

service/storage_service.cc
jul-stas: done

api/storage_service.cc
jul-stas: right, done

service/storage_service.cc
jul-stas: Please have a look now. I throw `request_execution_exception` at the user, which arrives at nodetool with something like this:
```
nodetool: Scylla API server HTTP POST to URL '/storage_service/cdc_streams_check_and_repair' failed: exceptions::request_execution_exception (Timeout while fetching CDC topology description. Note: this is likely caused by node(s) being down or unreachable. It is recommended to check the network and restart/remove the failed node(s), then retry checkAndRepairCdcStreams command.)
```
...while the `cdc_log` gets such a message:
```
ERROR 2020-06-06 16:47:39,468 [shard 0] cdc - Timeout while fetching CDC topology description: "Operation timed out for ks.cf - received only 666 responses from 777 CL=TWO.". Translating the exception to `request_execution_exception`.
```

jul-stas:
@nyh would you be so kind to have a look? Apparently, other reviewers have no more remarks.

slivne:
@nyh ping

api/api-doc/storage_service.json
```diff
@@ -511,6 +511,21 @@
             }
          ]
       },
+      {
+         "path":"/storage_service/cdc_streams_check_and_repair",
```
nyh, line 515: nitpick (but you can ignore): maybe the word "repair" is a problem because it already means something else in Scylla?

PR #6562, opened 2020-06-02
scylla_coredump_setup: fix coredump directory mount

amoskong:
Currently in coredump setup, we enabled a systemd mount to mount default coredump directory to /var/lib/scylla/coredump, but we didn't start it. So the coredump will still be saved to default coredump directory before a system reboot, it might touch enospc problem.

One patch started the systemd mount during coredump setup, and make the mount effect.
Another patch improved the error message of systemd unit, it's confused when the unit config is invalid.

### DIFF
--- a/dist/common/scripts/scylla_coredump_setup
+++ b/dist/common/scripts/scylla_coredump_setup
@@ -78,6 +78,7 @@ WantedBy=multi-user.target
             makedirs('/var/lib/scylla/coredump')
             systemd_unit.reload()
             systemd_unit('var-lib-systemd-coredump.mount').enable()
+            systemd_unit('var-lib-systemd-coredump.mount').start()
         if os.path.exists('/usr/lib/sysctl.d/50-coredump.conf'):
             run('sysctl -p /usr/lib/sysctl.d/50-coredump.conf')
         else:
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -618,7 +618,7 @@ class systemd_unit:
         try:
             run('systemctl {} cat {}'.format(self.ctlparam, unit), silent=True)
         except subprocess.CalledProcessError:
-            raise SystemdException('unit {} not found'.format(unit))
+            raise SystemdException('unit {} is not found or invalid'.format(unit))
         self._unit = unit
 
     def start(self):


amoskong:
Quote from https://www.freedesktop.org/software/systemd/man/systemd.mount.html


```
What=
Takes an absolute path of a device node, file or other resource to mount. See mount(8) for details. If this refers to a device node, a dependency on the respective device unit is automatically created. (See systemd.device(5) for more information.) This option is mandatory. Note that the usual specifier expansion is applied to this setting, literal percent characters should hence be written as "%%". If this mount is a bind mount and the specified path does not exist yet it is created as directory.

Where=
Takes an absolute path of a file or directory for the mount point; in particular, the destination cannot be a symbolic link. If the mount point does not exist at the time of mounting, it is created as directory. This string must be reflected in the unit filename. (See above.) This option is mandatory.
```

So the mount point is  '/var/lib/systemd/coredump'
'/var/lib/scylla/coredump' is the file to mount. because /var/lib/scylla had mounted a second big storage, which has enough space for Huge coredumps.

----

Bentsi or other touched problem with old scylla-master AMI, a coredump occurred but not successfully saved to disk for enospc.
The directory /var/lib/systemd/coredump wasn't mounted to /var/lib/scylla/coredump.
They WRONGLY thought the wrong mount was caused by the config problem, so he posted a fix.

Actually scylla-ami-setup / coredump wasn't executed on that AMI, err: unit scylla-ami-setup.service not found
Because 'scylla-ami-setup.service' config file doesn't exist or is invalid.

Details of my testing: https://github.com/scylladb/scylla/issues/6300#issuecomment-637324507

So we need to revert Bentsi's patch, it changed the right config to wrong.

amoskong:
Append a patch to this PR, which fixed (var-lib-systemd-coredump.mount isn't started after setup #6566)

penberg:
@amoskong I already did the revert myself:

https://github.com/scylladb/scylla/commit/9d9d54c804dd400d90edfdb84c8f5319d06b1633

Can you rebase this on top of next, and fix up the pull request description?

amoskong:
> @amoskong I already did the revert myself:
> 
> [9d9d54c](https://github.com/scylladb/scylla/commit/9d9d54c804dd400d90edfdb84c8f5319d06b1633)
> 
> Can you rebase this on top of next, and fix up the pull request description?

Rebased to latest next, and updated the PR description.
Thanks!

Issue #6436, opened 2020-05-12
replace_node_no_hibernate_state_test failed: The process is dead, returncode=1


bhalevy:
Seen since https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/471/testReport/replace_address_test/TestReplaceAddress/replace_node_no_hibernate_state_test/
Scylla version e44b2826abba0d1c6ae91b7d40f0af112a24517d
```
Traceback (most recent call last):
  File "/usr/lib64/python3.7/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib64/python3.7/unittest/case.py", line 645, in run
    testMethod()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/replace_address_test.py", line 404, in replace_node_no_hibernate_state_test
    node4.start(replace_address=node2_address, wait_for_binary_proto=True, wait_other_notice=True)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/scylla_node.py", line 541, in start
    ext_env)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/scylla_node.py", line 281, in _start_scylla
    self.wait_for_binary_interface(from_mark=self.mark, process=self._process_scylla, timeout=420)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/node.py", line 494, in wait_for_binary_interface
    self.watch_log_for("Starting listening for CQL clients", **kwargs)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/node.py", line 460, in watch_log_for
    raise RuntimeError("The process is dead, returncode={}".format(process.returncode))
RuntimeError: The process is dead, returncode=1
```

https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/471/artifact/logs-release.2/1588473761239_replace_address_test.TestReplaceAddress.replace_node_no_hibernate_state_test/node4.log
```
ERROR 2020-05-03 02:42:41,124 [shard 0] init - Startup failed: std::runtime_error (Cannot replace token 3444658265712363823 which does not exist!)
```

asias:
[...]

PR #6574, opened 2020-06-03
gossip: Do not send shutdown message when a node is in unknown status

asias:
When a replacing node is in early boot up and is not in HIBERNATE sate
yet, if the node is killed by a user, the node will wrongly send a
shutdown message to other nodes. This is because UNKNOWN is not in
SILENT_SHUTDOWN_STATES, so in gossiper::do_stop_gossiping, the node will
send shutdown message. Other nodes in the cluster will call
storage_service::handle_state_normal for this node, since NORMAL and
SHUTDOWN status share the same status handler. As a result, other nodes
will incorrectly think the node is part of the cluster and the replace
operation is finished.

Such problem was seen in replace_node_no_hibernate_state_test dtest:

   n1, n2 are in the cluster
   n2 is dead
   n3 is started to replace n2, but n3 is killed in the middle
   n3 announces SHUTDOWN status wrongly
   n1 runs storage_service::handle_state_normal for n3
   n1 get tokens for n3 which is empty, because n3 hasn't gossip tokens yet
   n1 skips update normal tokens for n3,  but think n3 has replaced n2
   n4 starts to replace n2
   n4 checks the tokens for n2 in storage_service::join_token_ring (Cannot
      replace token {} which does not exist!) or
      storage_service::prepare_replacement_info (Cannot replace_address {}
      because it doesn't exist in gossip)

To fix, we add UNKNOWN into SILENT_SHUTDOWN_STATES and avoid sending
shutdown message.

Tests: replace_address_test.py:TestReplaceAddress.replace_node_no_hibernate_state_test
Fixes: #6436

### DIFF
--- a/gms/gossiper.hh
+++ b/gms/gossiper.hh
@@ -175,6 +175,7 @@ public:
         versioned_value::STATUS_LEFT,
         versioned_value::HIBERNATE,
         versioned_value::STATUS_BOOTSTRAPPING,
+        versioned_value::STATUS_UNKNOWN,
     };
     static constexpr std::chrono::milliseconds INTERVAL{1000};
     static constexpr std::chrono::hours A_VERY_LONG_TIME{24 * 3};


bhalevy:
makes sense to me.
@tgrabiec please review/merge too

bhalevy:
@asias can you please change the target branch to `next`?

asias:
> @asias can you please change the target branch to `next`?

done

bhalevy:
Can be closed with dddde335123169f8f81d299b38bae64c41c1adcd

asias:
@avikivity Can you please backport to 4.1? Amos saw the same issue on 4.1 test here: https://github.com/scylladb/scylla/issues/6648.

PR #6586, opened 2020-06-04
Alternator: should output empty map, not empty JSON

### LINKED ISSUE #6568: Alternator: should output empty map, not empty JSON

fruch:
Fixes #6568

### DIFF
--- a/alternator/executor.cc
+++ b/alternator/executor.cc
@@ -1274,9 +1274,9 @@ std::optional<shard_id> rmw_operation::shard_for_execute(bool needs_read_before_
 // optionally return Attributes if requested via the ReturnValues option.
 static future<executor::request_return_type> rmw_operation_return(rjson::value&& attributes) {
     // As an optimization, in the simple and common case that nothing is to be
-    // returned, quickly return an empty result:
+    // returned, quickly return an empty json object as string:
     if (attributes.IsNull()) {
-        return make_ready_future<executor::request_return_type>(json_string(""));
+        return make_ready_future<executor::request_return_type>(json_string("{}"));
     }
     rjson::value ret = rjson::empty_object();
     if (!attributes.IsNull()) {


fruch:
@nyh, I've tried adding a test for it, but seem like boto3 client is handling it nice.
the code that was having the issue wasn't using boto3, but botocore directly...

so I gave up on adding a test

nyh:
@psarna wrote a test for it, `test/alternator/test_manual_requests.py`, `test_put_item_return_type`.
As you noted, boto3 can't test for this, it silently accepts the empty string.
So his test uses a manual request and tries to parse it with json.loads - and fails because as you correctly noted, the empty string cannot be parsed as a JSON object.

nyh:
Note that @psarna also wrote his own patch for this (I didn't see yours...). 8fc3ca855eeb12a173ccf3bede499ce0fd71bff7. Committed yesterday.

fruch:
Yeah I wasn't looking at next, and seem like his change is better anyhow

fruch:
I've practiced reading a bit of C code... Need to shave the rust once in a while

nyh:
Thanks. Your patch seems good. And Rust is a completely different language :-)

Issue #6555, opened 2020-06-01
Data can get resurrected if flush and truncate run on the same time


haaawk:

haaawk:
@elcallio @gleb-cloudius could you please comment?

slivne:
@elcallio ping ?

elcallio:
Call me old fashioned, but I think that if one says there is a race, one should explain said race instead of insisting on a treasure hunt. 
Truncate flushes or clears memtables, picking up the RP at this point, then removes all sstables before or at said RP. Data can absolutely come in after this, but is there an actual race you have observed?

haaawk:
Yes, there is a race w observed. This issue wasn't created after eyeballing the code but after running tests and observing this behavior. The test was as follows:
1. Have two nodes
2. Run write workload on them
3. Stop workload
4. Run truncate

This occasionally led to resurrected data but if we include nodetool flush between 3 and 4 then the resurrection never happened.

slivne:
One theory is that hints were sent - so if we had a hints backlog and in parallel did a truncate (and hints were being replayed) then it may cause this.

whats the interaction between truncate and hints

haaawk:
It's not about the hints @slivne. If the memtable is being flashed and truncate appears in the middle of that flash, the flash will continue after the truncate and data will be stored to disk despite truncate. At least, that's my understanding.

elcallio:
Truncate only looks at whatever memtables says is its flush position, i.e. highest RP value. Then it prunes sstables. 
I can only assume this problem is because `cf.flush()/cf.clear()` does not wait for initiated/pending flush operations to finish. 
The code for flushing etc has been optimized and changed so many times without considering poor truncate...
I honestly have no idea how to ensure waiting for flushes in the code as it looks today.

kostja:
@gleb should truncate go through raft log?

raphaelsc:
> Truncate only looks at whatever memtables says is its flush position, i.e. highest RP value. Then it prunes sstables. I can only assume this problem is because `cf.flush()/cf.clear()` does not wait for initiated/pending flush operations to finish. The code for flushing etc has been optimized and changed so many times without considering poor truncate... I honestly have no idea how to ensure waiting for flushes in the code as it looks today.

the initiated/pending flush operation part was addressed by 3ad0f156b939de5ee507e122b03d3b24e2f4a98e

gleb-cloudius:
> @gleb should truncate go through raft log?

Why?

[...]

PR #6593, opened 2020-06-04
Two-phase truncate

### LINKED ISSUE #6096: Data may be resurrected aftre truncate by hinted handoff
Installation details*
Scylla version (or git commit hash): 3159ad4484b26f424d9a4657c58f678d76db70d4

HH does not check a table's truncation time when sending or applying a hint which may cause data to be resurrected.

piodul:
Currently, TRUNCATE is a single-phase operation: the coordinator checks if all nodes in the cluster are alive, and sends a command to each of them, telling them to get rid of their local data for the subject table (and its materialized views). Truncate commands may arrive at different points in time on each node, therefore they may race with replica writes, and a write operation might leave data that has lower consistency than requested after truncation finishes.

Hinted handoff writes suffer from a similar problem. First, hints older than the truncation time are not prevented from being sent out. While this is easily fixed (also in this PR), the local truncation time might differ on each node, and a node which performs truncation later might send a hint to another node which has already performed the truncation.

To mitigate the problems described above, the old, single phase of TRUNCATE (now called ACCEPT) is preceded by another phase, PREPARE. PREPARE makes sure that nothing is being written to the subject table by waiting for the writes that have already started, and by rejecting new writes. As for hints, ongoing hints are waited on, and more hints to that table won't be attempted to be sent.

After all nodes acknowledge the PREPARE phase, the ACCEPT phase starts. All nodes are instructed to remove their local data, and unblock writes/hints for that table.

In case something goes wrong and the ACCEPT phase won't be started on some of the nodes, writes and hints will become unblocked after the truncation timeout.

If somebody wishes not to block writes during truncation, there is an enable_write_blocking_during_truncate configuration option that allows to disable this feature. TRUNCATE will still use two phases, and will block hint traffic for the duration of truncation.

Fixes #6096

### DIFF
--- a/db/config.cc
+++ b/db/config.cc
@@ -728,6 +728,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
         "Set to true if the cluster was initially installed from 3.1.0. If it was upgraded from an earlier version,"
         " or installed from a later version, leave this set to false. This adjusts the communication protocol to"
         " work around a bug in Scylla 3.1.0")
+    , enable_write_blocking_during_truncate(this, "enable_write_blocking_during_truncate", liveness::LiveUpdate, value_status::Used, true, "Enables waiting for ongoing writes before performing TRUNCATE operation, and rejecting new writes for the duration of the TRUNCATE operation")
     , enable_user_defined_functions(this, "enable_user_defined_functions", value_status::Used, false,  "Enable user defined functions. You must also set experimental-features=udf")
     , user_defined_function_time_limit_ms(this, "user_defined_function_time_limit_ms", value_status::Used, 10, "The time limit for each UDF invocation")
     , user_defined_function_allocation_limit_bytes(this, "user_defined_function_allocation_limit_bytes", value_status::Used, 1024*1024, "How much memory each UDF invocation can allocate")
--- a/db/config.hh
+++ b/db/config.hh
@@ -305,6 +305,7 @@ public:
     named_value<uint32_t> max_clustering_key_restrictions_per_query;
     named_value<uint64_t> max_memory_for_unlimited_query;
     named_value<bool> enable_3_1_0_compatibility_mode;
+    named_value<bool> enable_write_blocking_during_truncate;
     named_value<bool> enable_user_defined_functions;
     named_value<unsigned> user_defined_function_time_limit_ms;
     named_value<unsigned> user_defined_function_allocation_limit_bytes;
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -187,6 +187,13 @@ bool manager::should_ignore_hints_during_replay_for_table(schema_ptr s) const {
     return _tables_to_ignore_hints_for.count(base_id);
 }
 
+future<> manager::flush_current_hints() {
+    manager_logger.debug("Flushing hints on all endpoint managers");
+    return parallel_for_each(_ep_managers, [] (auto& ep) {
+        return ep.second.flush_current_hints();
+    });
+}
+
 future<> manager::await_in_flight_hints() {
     return parallel_for_each(_ep_managers, [] (auto& ep) {
         return ep.second.await_in_flight_hints();
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -172,6 +172,21 @@ bool manager::is_generating_hints_for_schema_allowed(schema_ptr s) const {
     return _tables_blocked_from_generating_hints.count(base_id);
 }
 
+void manager::start_ignoring_hints_during_replay_for_table_and_its_views(utils::UUID cf_id) {
+    manager_logger.debug("Start ignoring hints for table {} and all its views", cf_id);
+    _tables_to_ignore_hints_for.insert(cf_id);
+}
+
+void manager::stop_ignoring_hints_during_replay_for_table_and_its_views(utils::UUID cf_id) {
+    manager_logger.debug("Start ignoring hints for table {} and all its views", cf_id);
+    _tables_to_ignore_hints_for.erase(cf_id);
+}
+
+bool manager::should_ignore_hints_during_replay_for_table(schema_ptr s) const {
+    const auto base_id = s->is_view() ? s->view_info()->base_id() : s->id();
+    return _tables_to_ignore_hints_for.count(base_id);
+}
+
 bool manager::end_point_hints_manager::store_hint(schema_ptr s, lw_shared_ptr<const frozen_mutation> fm, tracing::trace_state_ptr tr_state) noexcept {
     try {
         // Future is waited on indirectly in `stop()` (via `_store_gate`).
@@ -723,6 +738,11 @@ future<> manager::end_point_hints_manager::sender::send_one_hint(lw_shared_ptr<s
                 // Files are aggregated for at most manager::hints_timer_period therefore the oldest hint there is
                 // (last_modification - manager::hints_timer_period) old.
                 if (gc_clock::now().time_since_epoch() - secs_since_file_mod > gc_grace_sec - manager::hints_flush_period) {
+                    ctx_ptr->rps_set.erase(rp);
+                    return make_ready_future<>();
+                }
+
+                if (_shard_manager.should_ignore_hints_during_replay_for_table(m.s)) {
                     ctx_ptr->rps_set.erase(rp);
                     return make_ready_future<>();
                 }
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -776,6 +776,14 @@ future<> manager::end_point_hints_manager::sender::send_one_hint(lw_shared_ptr<s
                     return make_ready_future<>();
                 }
 
+                // Don't send hints older than the truncation record of the table
+                auto& cf = _db.find_column_family(m.s);
+                const auto secs_since_truncation = cf.get_truncation_record().time_since_epoch();
+                if (secs_since_file_mod < secs_since_truncation) {
+                    ctx_ptr->rps_set.erase(rp);
+                    return make_ready_future<>();
+                }
+
                 return this->send_one_mutation(std::move(m)).then([this, rp, ctx_ptr] {
                     ctx_ptr->rps_set.erase(rp);
                     ++this->shard_stats().sent;
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -200,6 +200,21 @@ future<> manager::await_in_flight_hints() {
     });
 }
 
+future<> manager::block_hints_for_table_and_its_views(utils::UUID cf_id) {
+    forbid_generating_hints_for_table_and_its_views(cf_id);
+    start_ignoring_hints_during_replay_for_table_and_its_views(cf_id);
+
+    return when_all(
+        flush_current_hints(),
+        await_in_flight_hints()
+    ).discard_result();
+}
+
+void manager::unblock_hints_for_table_and_its_views(utils::UUID cf_id) {
+    allow_generating_hints_for_table_and_its_views(cf_id);
+    stop_ignoring_hints_during_replay_for_table_and_its_views(cf_id);
+}
+
 bool manager::end_point_hints_manager::store_hint(schema_ptr s, lw_shared_ptr<const frozen_mutation> fm, tracing::trace_state_ptr tr_state) noexcept {
     try {
         // Future is waited on indirectly in `stop()` (via `_store_gate`).
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -38,6 +38,7 @@
 #include "service/priority_manager.hh"
 #include "database.hh"
 #include "service_permit.hh"
+#include "view_info.hh"
 
 using namespace std::literals::chrono_literals;
 
@@ -154,6 +155,21 @@ void manager::forbid_hints_for_eps_with_pending_hints() {
             ep_man.allow_hints();
         }
     });
+}
+
+void manager::forbid_generating_hints_for_table_and_its_views(utils::UUID cf_id) {
+    manager_logger.debug("Stop generating hints for table {} and all its views", cf_id);
+    _tables_blocked_from_generating_hints.insert(cf_id);
+}
+
+void manager::allow_generating_hints_for_table_and_its_views(utils::UUID cf_id) {
+    manager_logger.debug("Resume generating hints for table {} and all its views", cf_id);
+    _tables_blocked_from_generating_hints.erase(cf_id);
+}
+
+bool manager::is_generating_hints_for_schema_allowed(schema_ptr s) const {
+    const auto base_id = s->is_view() ? s->view_info()->base_id() : s->id();
+    return _tables_blocked_from_generating_hints.count(base_id);
 }
 
 bool manager::end_point_hints_manager::store_hint(schema_ptr s, lw_shared_ptr<const frozen_mutation> fm, tracing::trace_state_ptr tr_state) noexcept {
@@ -290,7 +306,7 @@ inline bool manager::have_ep_manager(ep_key_type ep) const noexcept {
 }
 
 bool manager::store_hint(ep_key_type ep, schema_ptr s, lw_shared_ptr<const frozen_mutation> fm, tracing::trace_state_ptr tr_state) noexcept {
-    if (stopping() || !started() || !can_hint_for(ep)) {
+    if (stopping() || !started() || !can_hint_for(ep, s)) {
         manager_logger.trace("Can't store a hint to {}", ep);
         ++_stats.dropped;
         return false;
@@ -482,13 +498,17 @@ bool manager::too_many_in_flight_hints_for(ep_key_type ep) const noexcept {
     return _stats.size_of_hints_in_progress > max_size_of_hints_in_progress && !utils::fb_utilities::is_me(ep) && hints_in_progress_for(ep) > 0 && local_gossiper().get_endpoint_downtime(ep) <= _max_hint_window_us;
 }
 
-bool manager::can_hint_for(ep_key_type ep) const noexcept {
+bool manager::can_hint_for(ep_key_type ep, schema_ptr s) const noexcept {
     if (utils::fb_utilities::is_me(ep)) {
         return false;
     }
 
     auto it = find_ep_manager(ep);
     if (it != ep_managers_end() && (it->second.stopping() || !it->second.can_hint())) {
+        return false;
+    }
+
+    if (is_generating_hints_for_schema_allowed(s)) {
         return false;
     }
 
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -187,6 +187,12 @@ bool manager::should_ignore_hints_during_replay_for_table(schema_ptr s) const {
     return _tables_to_ignore_hints_for.count(base_id);
 }
 
+future<> manager::await_in_flight_hints() {
+    return parallel_for_each(_ep_managers, [] (auto& ep) {
+        return ep.second.await_in_flight_hints();
+    });
+}
+
 bool manager::end_point_hints_manager::store_hint(schema_ptr s, lw_shared_ptr<const frozen_mutation> fm, tracing::trace_state_ptr tr_state) noexcept {
     try {
         // Future is waited on indirectly in `stop()` (via `_store_gate`).
@@ -717,6 +723,7 @@ future<> manager::end_point_hints_manager::sender::send_one_mutation(frozen_muta
 }
 
 future<> manager::end_point_hints_manager::sender::send_one_hint(lw_shared_ptr<send_one_file_ctx> ctx_ptr, fragmented_temporary_buffer buf, db::replay_position rp, gc_clock::duration secs_since_file_mod, const sstring& fname) {
+    auto op = _hint_send_barrier.start();
     return _resource_manager.get_send_units_for(buf.size_bytes()).then([this, secs_since_file_mod, &fname, buf = std::move(buf), rp, ctx_ptr] (auto units) mutable {
         // Future is waited on indirectly in `send_one_file()` (via `ctx_ptr->file_send_gate`).
         (void)with_gate(ctx_ptr->file_send_gate, [this, secs_since_file_mod, &fname, buf = std::move(buf), rp, ctx_ptr] () mutable {
@@ -772,7 +779,7 @@ future<> manager::end_point_hints_manager::sender::send_one_hint(lw_shared_ptr<s
     }).handle_exception([this, ctx_ptr] (auto eptr) {
         manager_logger.trace("send_one_file(): Hmmm. Something bad had happend: {}", eptr);
         ctx_ptr->state.set(send_state::segment_replay_failed);
-    });
+    }).finally([op = std::move(op)] {});
 }
 
 // runs in a seastar::async context
--- a/db/hints/manager.hh
+++ b/db/hints/manager.hh
@@ -472,6 +472,7 @@ private:
     std::unordered_set<ep_key_type> _eps_with_pending_hints;
     seastar::named_semaphore _drain_lock = {1, named_semaphore_exception_factory{"drain lock"}};
     std::unordered_set<utils::UUID> _tables_blocked_from_generating_hints;
+    std::unordered_set<utils::UUID> _tables_to_ignore_hints_for;
 
 public:
     manager(sstring hints_directory, std::vector<sstring> hinted_dcs, int64_t max_hint_window_ms, resource_manager&res_manager, distributed<database>& db);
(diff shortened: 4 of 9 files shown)


gleb-cloudius:
Before reviewing more in-depth can you explain what is the goal of the patch? Does it suppose to ensure that after successful truncation no old data is present on the cluster? If yes I do not think it achieve its goals. Mostly because tracking writes in flight is much harder that the patch assumes. Waiting for all write handler to disappear does not guaranty that since the write itself may outlive it.

tzach:
Will this also fix #6555 ?

haaawk:
> Will this also fix #6555 ?

No. That's something else discovered while working on this one.

piodul:
@gleb-cloudius

I developed this patch after discussion in https://github.com/scylladb/scylla/pull/6593 . It was created in response to the problem of hints resurrecting data after their destination table was truncated.

Originally, the fix was only supposed to stop hint traffic for the duration of truncation. With @vladzcloudius's suggestion, I expanded it to block regular writes as well.

About tracking of writes... I am aware that a write operation may time out and there is be no way for the coordinator to know if and when the mutation was accepted on replicas. This patch aims to improve the current situation by preventing non-timed out writes from resurrecting data.

Are there other cases that I missed? Do you have an idea how to improve this patch so that all cases are covered?

gleb-cloudius:
> @gleb-cloudius
> 
> I developed this patch after discussion in #6593 . It was created in response to the problem of hints resurrecting data after their destination table was truncated.

#6593 is this one.

> 
> Originally, the fix was only supposed to stop hint traffic for the duration of truncation. With @vladzcloudius's suggestion, I expanded it to block regular writes as well.
> 
> About tracking of writes... I am aware that a write operation may time out and there is be no way for the coordinator to know if and when the mutation was accepted on replicas. This patch aims to improve the current situation by preventing non-timed out writes from resurrecting data.

Truncate is not supposed to run along with writes and if it does I would not call what happens "resurrection of data" since it is not defined what happens when both run simultaneously. Hints is another matter since it may resurrect a writes that was done long before truncate.

So I question the need for all the complexity in the patch especially as it does not fix the problem, but only somewhat improves it.  And the patch is not even complex enough. It misses a feature negotiations for new verbs support.

> 
> Are there other cases that I missed? Do you have an idea how to improve this patch so that all cases are covered?

It is very hard to ensure that no write will appear somewhere in the future if they stuck in some queue. The most problematic case is if it stuck in a networking queue in which case we cannot even timeout it properly. Theoretically since truncate and mutate verbs use same connection a truncate should not bypass mutate. but in practice they can be reordered.

gleb-cloudius:
This also does not cover LWT writes at all.

piodul:
@gleb-cloudius

> #6593 is this one.

Sorry, this is the correct link: https://github.com/scylladb/scylla/issues/6096

> Truncate is not supposed to run along with writes and if it does I would not call what happens "resurrection of data" since it is not defined what happens when both run simultaneously. Hints is another matter since it may resurrect a writes that was done long before truncate.
> 
> So I question the need for all the complexity in the patch especially as it does not fix the problem, but only somewhat improves it. And the patch is not even complex enough. It misses a feature negotiations for new verbs support.

Ok, I see that stopping regular mutation traffic might be a futile effort due to the nature of truncate and unreliable nature of the network.

Something needs to be done about the hints, though. As hints suffer from the same problems, but cannot be stopped manually, we need to decide on the level of brokenness a fix for issue https://github.com/scylladb/scylla/issues/6096 will have. This might be achieved either by:

- Using this patch, but restricting it to stop and wait for in-progress hints only,
- Switching to something simpler.

In case of this patch, the missing feature negotiation can be added, of course.

@vladzcloudius

Could you comment on this?

avikivity:
Isn't rejecting all writes causing an even bigger problem? The applicaiton may not expect that, and start failing.

Suppose I have a time-series application that is logging information, and I decide to start the data from scratch. Why should writes start failing?

Incidentally, time-series workloads don't suffer from resurrection.

vladzcloudius:
> @gleb-cloudius
> 
> > #6593 is this one.
> 
> Sorry, this is the correct link: #6096
> 
> > Truncate is not supposed to run along with writes and if it does I would not call what happens "resurrection of data" since it is not defined what happens when both run simultaneously. Hints is another matter since it may resurrect a writes that was done long before truncate.
> > So I question the need for all the complexity in the patch especially as it does not fix the problem, but only somewhat improves it. And the patch is not even complex enough. It misses a feature negotiations for new verbs support.
> 
> Ok, I see that stopping regular mutation traffic might be a futile effort due to the nature of truncate and unreliable nature of the network.
> 
> Something needs to be done about the hints, though. As hints suffer from the same problems, but cannot be stopped manually, we need to decide on the level of brokenness a fix for issue #6096 will have. This might be achieved either by:
> 
>     * Using this patch, but restricting it to stop and wait for in-progress hints only,
> 
>     * Switching to something simpler.
> 
> 
> In case of this patch, the missing feature negotiation can be added, of course.
> 
> @vladzcloudius
> 
> Could you comment on this?

@piodul Hi.
I agree with Gleb that we'd rather not attempt to fix the problem with writes, at least not right now.

Like I mentioned on #6096 - the simplest way to fix resurrection issues related to the TRUNCATE would be to change the implementation of TRUNCATE to be a range tombstone.

However I assume that you've discussed it with @slivne and @avikivity and it was decided not to go that path.

denesb:
Sorry if this has been suggested before but why not add a REST command or something that drops all hints on all nodes for a table? As truncate is an admin operation, the same admin could run the drop-hints commands. Or maybe it could be run as part of the truncate.

avikivity:
It should be part of the truncate, why burden the admin with these details?

vladzcloudius:
> It should be part of the truncate, why burden the admin with these details?

@avikivity @denesb The algorithm above attempts to implement exactly that (depending on how you define the time frame of the TRUNCATE operation). However this is not as trivial as dropping a file since hints are currently aggregated per destination therefore there may be hints to different tables in the same hints file.

Therefore if you want to make sure that all currently existing hints to a particular table are dropped you should either scan all hints files are clean them up or you should change the way we write hints files and then there WILL be as many commitlog writers as there are tables on each shard.

What about my "implement a TRUNCATE as a range tombstone" idea? Any comments, @avikivity ?

avikivity:
I see, so best not to touch hints. However, when sending and receiving hints, we can compare them against the truncate timestamp (which is basicallly a tombstone for the entire table).

denesb:
On Thu, 2020-06-11 at 08:07 -0700, vladzcloudius wrote:
> > It should be part of the truncate, why burden the admin with these
> > details?
> 
> @avikivity @denesb The algorithm above attempts to implement exactly
> that (depending on how you define the time frame of the TRUNCATE
> operation). However this is not as trivial as dropping a file since
> hints are currently aggregated per destination therefore there may be
> hints to different tables in the same hints file.
> 
> Therefore if you want to make sure that all currently existing hints
> to a particular table are dropped you should either scan all hints
> files are clean them up or you should change the way we write hints
> files and then there WILL be as many commitlog writers as there are
> tables on each shard.
> 
> What about my "implement a TRUNCATE as a range tombstone" idea? Any
> comments, @avikivity ?

A cross-partition range-tombstone?
Nonononononononononoooooooooooooooooo.

The existing range tombstones are a major source of pain, they are the
most problematic atom we have and you want to introduce a new super
version of it?

vladzcloudius:
> On Thu, 2020-06-11 at 08:07 -0700, vladzcloudius wrote: > It should be part of the truncate, why burden the admin with these > details? @avikivity @denesb The algorithm above attempts to implement exactly that (depending on how you define the time frame of the TRUNCATE operation). However this is not as trivial as dropping a file since hints are currently aggregated per destination therefore there may be hints to different tables in the same hints file. Therefore if you want to make sure that all currently existing hints to a particular table are dropped you should either scan all hints files are clean them up or you should change the way we write hints files and then there WILL be as many commitlog writers as there are tables on each shard. What about my "implement a TRUNCATE as a range tombstone" idea? Any comments, @avikivity ?
> A cross-partition range-tombstone? Nonononononononononoooooooooooooooooo. The existing range tombstones are a major source of pain, they are the most problematic atom we have and you want to introduce a new super version of it?
> […](#)
> — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

@avikivity I had no idea that we have issues with cross partitions tombstones. As far as I know we do support range tombstones in production. Are there open github issues about that?

I think it makes more sense to fix our multiple-partitions tombstones issues which we need to fix anyway instead of trying to fix a dead beef which our TRUNCATE implementation is.

vladzcloudius:
@denesb After taking a closer look I see that a cross-partitions range tombstone would be a completely new thing for us (or Cassandra) and this is a rather fat project with a far reaching implications. 

However if (!!!) we want to proper fix this TRUNCATE nonsense I don't see a better option.
However the question is do we really want to. Probably as broken as it is - it's good enough.

In this case there is little sense to do anything more than what is really required and this is covered in my post [here](https://github.com/scylladb/scylla/issues/6096#issuecomment-606704311).

vladzcloudius:
> I see, so best not to touch hints. However, when sending and receiving hints, we can compare them against the truncate timestamp (which is basicallly a tombstone for the entire table).

@avikivity Right and this is exactly what was proposed [here](https://github.com/scylladb/scylla/issues/6096#issuecomment-606704311) and AFAIU what @piodul wanted to implement here.
The only difference from the original idea here is that he extended it to regular writes too which, as @gleb-cloudius mentioned and I agree with him, is an overkill.

piodul:
> I see, so best not to touch hints. However, when sending and receiving hints, we can compare them against the truncate timestamp (which is basicallly a tombstone for the entire table).

@avikivity We do not reject mutations with timestamps older than truncation time. It is legal now to truncate a table and then fill it with mutations with timestamps from the past, e.g. a year ago. Therefore, it does not work as a tombstone.

Two concepts need to be separated here: mutation timestamp, and hint creation time. We do not want to send or accept hints that were created before the truncation procedure, irrelevant of their mutation timestamps.

I'm not sure what would be the correct way to approach filtering hints on the receiver side (this feature is missing in the current state of the patch). We can extend HINT_MUTATION verb to include the hint creation timestamp and filter them based on the truncate timestamp on the receiver side.

This will also require all nodes to agree on the same truncation timestamp. If we assume that clocks on each node are synchronized relatively well, then I guess we can gather current time from all node and assume that the truncation time is the largest value of them all. It might discard some new hints, but it won't accept any pre-truncate hints. Is the assumption about clocks reasonable?

gleb-cloudius:
On Fri, Jun 12, 2020 at 05:51:56AM -0700, piodul wrote:
> > I see, so best not to touch hints. However, when sending and receiving hints, we can compare them against the truncate timestamp (which is basicallly a tombstone for the entire table).
> 
> @avikivity We do not reject mutations with timestamps older than truncation time. It is legal now to truncate a table and then fill it with mutations with timestamps from the past, e.g. a year ago. Therefore, it does not work as a tombstone.
> 
That's a out bug that does not exists in Cassandra and the one that we
actually need to fix.

> Two concepts need to be separated here: mutation timestamp, and hint creation time. We do not want to send or accept hints that were created before the truncation procedure, irrelevant of their mutation timestamps.
> 
> I'm not sure what would be the correct way to approach filtering hints on the receiver side (this feature is missing in the current state of the patch). We can extend HINT_MUTATION verb to include the hint creation timestamp and filter them based on the truncate timestamp on the receiver side.
> 
Filer them on a sender only. This is what Cassandra does.
 
> This will also require all nodes to agree on the same truncation timestamp. If we assume that clocks on each node are synchronized relatively well, then I guess we can gather current time from all node and assume that the truncation time is the largest value of them all. It might discard some new hints, but it won't accept any pre-truncate hints. Is the assumption about clocks reasonable?
> 
Truncate should not be done in parallel with writes, so there should not
be hint generated while it is running. So ultimately this is not
_that_ important to synchronise the timestamp. Stop traffic, wait 5
seconds, truncate should guaranty that there will be no resurrection
with reasonable certainty.

The only real way to fix truncate is to invent a special kind of
tombstone and persist it in sstables. But this is hard. The point of the
fix should be to bring us on par with what Cassandra is doing and this
should be a super simple fix.
 
--
			Gleb.

avikivity:
Maybe we can filter the hints with the hint creation timestamp (we record that?) on the sender side. It isn't accurate, but truncate running in parallel with writes isn't accurate either.

So we end up with the following:
 - if the user doesn't write during truncate, for a period of time that is more than the clock skew (e.g. wait a few seconds), then truncate won't let any older mutations pass
 - if truncate runs in parallel with writes, a few may not get truncated

vladzcloudius:
> Maybe we can filter the hints with the hint creation timestamp (we record that?) on the sender side. 

@avikivity 
We don't record the hint generation time.
Remember the gc_grace discussion in the context of the initial HH series? 
v1 had changes I added to `commitlog` serializer in order to be able to record the exact hint time stamp. These changes had more lines than the rest of the series so we dropped this and instead I flush hints files every 10 seconds therefore every hint in each hints file can be at most <file write time> + 10s old.

We use this approach while handling the gc_grace already.

> It isn't accurate, but truncate running in parallel with writes isn't accurate either.
> 
> So we end up with the following:
> 
>     * if the user doesn't write during truncate, for a period of time that is more than the clock skew (e.g. wait a few seconds), then truncate won't let any older mutations pass
> 
>     * if truncate runs in parallel with writes, a few may not get truncated

piodul:
@avikivity @gleb-cloudius @vladzcloudius

It seems that there is an agreement to solve this by filtering on the sender side. Like @vladzcloudius explained, although we don't store hint creation time, we approximate it by modification time of the file from which the hint was read.

There is a possibility that if an operator truncates quickly after stopping writes, some unflushed hints created before truncation might get flushed after truncation happens. The new file will have a later timestamp than the truncation timestamp.

The above can be solved either by force-flushing hints during truncation (so that the new file will have a timestamp before truncation), or requiring an operator to wait at least 10 seconds before truncating.

There is also an issue of hints that are being sent while truncation is in progress. Should we do something about them? Truncating for the second time should probably fix this in most cases.

vladzcloudius:
> @avikivity @gleb-cloudius @vladzcloudius
> 
> It seems that there is an agreement to solve this by filtering on the sender side. Like @vladzcloudius explained, although we don't store hint creation time, we approximate it by modification time of the file from which the hint was read.
> 
> There is a possibility that if an operator truncates quickly after stopping writes, some unflushed hints created before truncation might get flushed after truncation happens. The new file will have a later timestamp than the truncation timestamp.

@piodul 
Please, see how I address exactly the same issue for gc: you don't compare the file's timestamp to the truncate time - you compare it to the `<truncate time> + 10s`. This way you are sure that any hint that was written since the truncation time is not sent out. 

In this case (for safeness) we may want to even add some extra padding (make it 2*10s) in order to take the collateral delays related to writing files to the disk and other stalls in the system.

This way you may drop some hints you didn't have to - but that's the price of the approximation.

> 
> The above can be solved either by force-flushing hints during truncation (so that the new file will have a timestamp before truncation), or requiring an operator to wait at least 10 seconds before truncating.

I think my proposal is simpler than the first option and the second option unacceptable.

> 
> There is also an issue of hints that are being sent while truncation is in progress. Should we do something about them? Truncating for the second time should probably fix this in most cases.

You should stop all sending, yes.

avikivity:
By "flush every 10 seconds" you mean that we close the file and create a new one every 10 seconds? That seems excessive.

vladzcloudius:
> By "flush every 10 seconds" you mean that we close the file and create a new one every 10 seconds? That seems excessive.

Yes, this is what we do.

avikivity:
This is terrible, why not keep using the same file?

Could be the cause of #6409

vladzcloudius:
> This is terrible, why not keep using the same file?
> 
> Could be the cause of #6409

It's beyond HH layer - it simply calls whatever Calle back then told me to do in order to flush the current content (if any) to the disk. See here: https://github.com/scylladb/scylla/blob/master/db/hints/manager.cc#L349

Optimizations like above should be implemented on the commitlog level - HH should simply call the correct API.

Back then there wasn't anything better than ^^

piodul:
> > @avikivity We do not reject mutations with timestamps older than truncation time. It is legal now to truncate a table and then fill it with mutations with timestamps from the past, e.g. a year ago. Therefore, it does not work as a tombstone.

> That's a out bug that does not exists in Cassandra and the one that we
actually need to fix.

@gleb-cloudius I'm not sure if we understood each other. I meant that the time when you truncate and timestamp of the mutation are separate concepts. It doesn't matter when you perform a truncation, it does not affect the timestamps. TRUNCATE brings the table to a state as if it was just created.

To verify what you wrote, I built a recent version of Cassandra (4.0 alpha1), and did the following:

```cql
cqlsh> create table ks.tbl (pk int primary key, c int);
cqlsh> insert into ks.tbl (pk, c) values (1, 2) using timestamp 2;
cqlsh> truncate table ks.tbl;
cqlsh> insert into ks.tbl (pk, c) values (1, 1) using timestamp 1;
cqlsh> select * from ks.tbl;

 pk | c
----+---
  1 | 1

(1 rows)
```

Truncating does not prevent you from writing data with low timestamps, both in Scylla and Cassandra.

> Filer them on a sender only. This is what Cassandra does.

Actually, Cassandra filters them on the receiver side as well. HintVerbHandler, [here](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/hints/HintVerbHandler.java#L97), calls `Hint.applyFuture`. You can see the filtering by looking at [the definition](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/hints/Hint.java#L102).

I think it is necessary to filter on the receiver for correctness. If a sender truncates earlier than receiver, the receiver might get some pre-truncation hints. Receiver must know if they need to drop them or not.

We may either augment HINT_MUTATION by including the approximate creation timestamp of the hint, or just drop hints to the truncated table for a short period of time (this is an idea that you proposed in the discussion under the issue that this PR was supposed to fix).

> @piodul
> Please, see how I address exactly the same issue for gc: you don't compare the file's timestamp to the truncate time - you compare it to the <truncate time> + 10s. This way you are sure that any hint that was written since the truncation time is not sent out.

@vladzcloudius  Right, that makes sense. Better to be on the safe side and lose some hints generated soon after truncation instead of resurrecting data. We already require all nodes to be available during truncation, so the cluster should be in a relatively good shape and not many hints will be lost that way.

gleb-cloudius:
On Fri, Jun 19, 2020 at 08:39:31AM -0700, piodul wrote:
> > > @avikivity We do not reject mutations with timestamps older than truncation time. It is legal now to truncate a table and then fill it with mutations with timestamps from the past, e.g. a year ago. Therefore, it does not work as a tombstone.
> 
> > That's a out bug that does not exists in Cassandra and the one that we
> actually need to fix.
> 
> @gleb-cloudius I'm not sure if we understood each other. I meant that the time when you truncate and timestamp of the mutation are separate concepts. It doesn't matter when you perform a truncation, it does not affect the timestamps. TRUNCATE brings the table to a state as if it was just created.
> 
> To verify what you wrote, I built a recent version of Cassandra (4.0 alpha1), and did the following:
> 
> ```cql
> cqlsh> create table ks.tbl (pk int primary key, c int);
> cqlsh> insert into ks.tbl (pk, c) values (1, 2) using timestamp 2;
> cqlsh> truncate table ks.tbl;
> cqlsh> insert into ks.tbl (pk, c) values (1, 1) using timestamp 1;
> cqlsh> select * from ks.tbl;
> 
>  pk | c
> ----+---
>   1 | 1
> 
> (1 rows)
> ```
> 
> Truncating does not prevent you from writing data with low timestamps, both in Scylla and Cassandra.
> 
Why should it? Creating new table does not prevent you from writing data
with timestamps earlier than table creation.

> > Filer them on a sender only. This is what Cassandra does.
> 
> Actually, Cassandra filters them on the receiver side as well. HintVerbHandler, [here](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/hints/HintVerbHandler.java#L97), calls `Hint.applyFuture`. You can see the filtering by looking at [the definition](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/hints/Hint.java#L102).
> 
> I think it is necessary to filter on the receiver for correctness. If a sender truncates earlier than receiver, the receiver might get some pre-truncation hints. Receiver must know if they need to drop them or not.
> 
> We may either augment HINT_MUTATION by including the approximate creation timestamp of the hint, or just drop hints to the truncated table for a short period of time (this is an idea that you proposed in the discussion under the issue that this PR was supposed to fix).
> 
We should amend our HINT_MUTATION as well than.

--
			Gleb.

vladzcloudius:
> > > @avikivity We do not reject mutations with timestamps older than truncation time. It is legal now to truncate a table and then fill it with mutations with timestamps from the past, e.g. a year ago. Therefore, it does not work as a tombstone.
> 
> > That's a out bug that does not exists in Cassandra and the one that we
> > actually need to fix.
> 
> @gleb-cloudius I'm not sure if we understood each other. I meant that the time when you truncate and timestamp of the mutation are separate concepts. It doesn't matter when you perform a truncation, it does not affect the timestamps. TRUNCATE brings the table to a state as if it was just created.
> 
> To verify what you wrote, I built a recent version of Cassandra (4.0 alpha1), and did the following:
> 
> ```sql
> cqlsh> create table ks.tbl (pk int primary key, c int);
> cqlsh> insert into ks.tbl (pk, c) values (1, 2) using timestamp 2;
> cqlsh> truncate table ks.tbl;
> cqlsh> insert into ks.tbl (pk, c) values (1, 1) using timestamp 1;
> cqlsh> select * from ks.tbl;
> 
>  pk | c
> ----+---
>   1 | 1
> 
> (1 rows)
> ```
> 
> Truncating does not prevent you from writing data with low timestamps, both in Scylla and Cassandra.
> 
> > Filer them on a sender only. This is what Cassandra does.
> 
> Actually, Cassandra filters them on the receiver side as well. HintVerbHandler, [here](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/hints/HintVerbHandler.java#L97), calls `Hint.applyFuture`. You can see the filtering by looking at [the definition](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/hints/Hint.java#L102).
> 
> I think it is necessary to filter on the receiver for correctness. If a sender truncates earlier than receiver, the receiver might get some pre-truncation hints. Receiver must know if they need to drop them or not.

@piodul Applying hints mutations should undergo exactly same constraints as regular mutations and regular mutations that have been issued prior to the last truncation should be dropped. 

For hints that would be corresponding hints file write time minus 10s.
We may append this timestamp to the HINT_MUTATION payload.

piodul:
@gleb-cloudius @vladzcloudius Thanks. I think that all details are clear to me now. I'm closing this PR right now, and will send another one soon with the solution discussed here.

avikivity:
> > This is terrible, why not keep using the same file?
> > Could be the cause of #6409
> 
> It's beyond HH layer - it simply calls whatever Calle told me to do back then in order to flush the current content (if any) to the disk. See here: https://github.com/scylladb/scylla/blob/master/db/hints/manager.cc#L349
> 
> Optimizations like above should be implemented on the commitlog level - HH should simply call the correct API.
> 
> Back then there wasn't anything better than ^^.
> 
> Why is it so terrible by the way? We are talking about a single file per shard that is being re-opened every 10 seconds...

We have a segment manager per file, no?

File operations are heavyweight, and files have a lot of overhead in terms of metadata.

vladzcloudius:
> > > This is terrible, why not keep using the same file?
> > > Could be the cause of #6409
> > 
> > 
> > It's beyond HH layer - it simply calls whatever Calle told me to do back then in order to flush the current content (if any) to the disk. See here: https://github.com/scylladb/scylla/blob/master/db/hints/manager.cc#L349
> > Optimizations like above should be implemented on the commitlog level - HH should simply call the correct API.
> > Back then there wasn't anything better than ^^.
> > Why is it so terrible by the way? We are talking about a single file per shard that is being re-opened every 10 seconds...
> 
> We have a segment manager per file, no?
> 
> File operations are heavyweight, and files have a lot of overhead in terms of metadata.

AFAIR a segment manager is per-commitlog object instance. It manages multiple commitlog files (segments). 

In the HH case - there will be one commitlog object per destination per shard on each node where objects are lazily created by the HH manager.

However every time we shut a commitlog object  down it destroys the segment manager.
I'm not sure how heavyweight those file operations are since files are re-created from scratch - the only thing that matters at that point is the next segment index which requires scanning all existing files and checking their name. This should not be that heavyweight since its just an operation on top of the file system and doesn't require reading any metadata from the file.

I agree with you that it would be much better if we could trigger a "flush" to a commitlog object without having to shut it down.
However back then the code in question has been written there was no such a possibility.

PR #6605, opened 2020-06-07
scylla_prepare: verify data partition mounted

### LINKED ISSUE #6012: Scylla should refuse starting if its FS is missing
*Installation details*
Scylla version (or git commit hash): 3.3.rc1-0.20200223.756574d0946 (but relevant since 2.3).
Cluster size: ANY
OS (RHEL/CentOS/Ubuntu/AWS AMI): AWS AMI

In case scylla starts when its FS is missing (/var/lib/scylla), the service is able to start but produces many errors like:
"ERR     | scylla: [shard 0] storage_proxy - Exception when communicating with 10.0.181.73: std::runtime_error (Can't find a column family with UUID 3cc8e850-57dd-11ea-9e81-00000000000c)"
And/or:
"Failed to load schema version"
"No space left on device"

This is reproducing every time longevity runs "RestartAndRepair" nemesis which is stoping i3 instance, causing it to lose its ephemeral disks and not able to mount the FS.
(The nemesis then later take care of it, fixing the FS mount and running a repair on the node).

syuu1228:
Seems like RequiresMountsFor does not refuse service startup when the mount
point removed from /etc/fstab, run manual check in scylla_prepare script.

Fixes #6012

### DIFF
--- a/dist/common/scripts/scylla_prepare
+++ b/dist/common/scripts/scylla_prepare
@@ -37,8 +37,23 @@ def verify_cpu():
                     print('Scylla requires the sse4.2 and clmul instruction sets, check your processor and hypervisor')
                     sys.exit(1)
 
+#XXX: RequiresMountFor does not refuse service startup when mountpoint
+#     commented out on /etc/fstab, run manual check
+MOUNTS_CONF='/etc/systemd/system/scylla-server.service.d/mounts.conf'
+def verify_mount():
+    if os.path.exists(MOUNTS_CONF):
+        with open(MOUNTS_CONF) as f:
+            for l in f:
+                m = re.search(r'RequiresMountsFor=(.*)\n', l)
+                if m:
+                    mount_at = m.group(1)
+        if not os.path.ismount(mount_at):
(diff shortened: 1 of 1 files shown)


dist/common/scripts/scylla_prepare
```diff
@@ -37,8 +37,23 @@ def verify_cpu():
...
+                m = re.search(r'RequiresMountsFor=(.*)\n', l)
+                if m:
+                    mount_at = m.group(1)
+        if not os.path.ismount(mount_at):
+            print('data partition {} not mounted'.format(mount_at))
+            sys.exit(1)
+
 if __name__ == '__main__':
     verify_cpu()
+    verify_mount()
```
avikivity, line 57: I don't think we should handle this case. If someone removes the mountpoint, then it means they don't want it mounted. Maybe they want the data on the root partition.

btw, we should be using systemd .mount files, not /etc/fstab, but thats a separate issue.

I'll comment on the issue too.

avikivity: In fact the issue is closed already.

PR #6610, opened 2020-06-08
scylla_setup: add swapfile setup

### LINKED ISSUE #6539: Scylla_setup does warns about swap, but not help to set it correctly
Scylla version (or git commit hash): 4.0

Installing Scylla ion AWS i3 add the following warning:
**No swap is configured, it is highly recommended to setup swap on Scylla node.**

It is not clear what one should do next, if at all.
A better  UX will either:
- guide the user on fixing this 
- Add this as a step in scylla_setup (better)

syuu1228:
Adding swapfile setup on scylla_setup.

Fixes #6539

### DIFF
--- a/dist/common/scripts/scylla_swap_setup
+++ b/dist/common/scripts/scylla_swap_setup
@@ -1,0 +1,80 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019 ScyllaDB
+#
+
+#
+# This file is part of Scylla.
+#
+# Scylla is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Scylla is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import sys
+import argparse
+from pathlib import Path
+from scylla_util import *
+
+if __name__ == '__main__':
+    if os.getuid() > 0:
+        print('Requires root permission.')
+        sys.exit(1)
+    parser = argparse.ArgumentParser(description='Configure swap for Scylla.')
+    parser.add_argument('--swap-directory',
+                        help='specify swapfile directory', default='/')
+    args = parser.parse_args()
+
+    if swap_exists():
+        print('swap already configured, exiting setup')
+        sys.exit(1)
+
+    memtotal = get_memtotal_gb()
+    # Scylla document says 'swap size should be set to either total_mem/3 or
+    # 16GB - lower of the two', so we need to compare 16g vs memtotal/3 and
+    # choose lower one
+    # see: https://docs.scylladb.com/faq/#do-i-need-to-configure-swap-on-a-scylla-node
+    swapsize = 16 if 16 < int(memtotal / 3) else int(memtotal / 3)
+    diskfree = get_disk_free_gb(args.swap_directory)
+    if swapsize > diskfree:
+        print('swap directory {} does not have enough disk space. {}GB space required.'.format(args.swap_directory, swapsize))
+        sys.exit(1)
+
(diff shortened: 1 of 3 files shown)


slivne:
@noamha I know we had issue in the cloud with swap and setting it up - I can't recall all the incarnations

Can you please review this as well - at the end we would like the AMI to include SWAP as well.

Thanks

dist/common/scripts/scylla_swap_setup
```diff
@@ -0,0 +1,58 @@
...
+    parser.add_argument('--swap-directory',
+                        help='specify swapfile directory', required=True)
+    args = parser.parse_args()
+
+    if swap_exists():
+        print('swap already configured, exiting setup')
+        sys.exit(1)
+
+    memtotal = get_memtotal_gb()
+    swapsize = 16 if 16 < memtotal / 3 else memtotal / 3
```
avikivity, line 43: This can be a huge amount. Let's clamp it between 16 and 50.

avikivity: Better also convert it to an integer.

dist/common/scripts/scylla_swap_setup
```diff
@@ -0,0 +1,58 @@
...
+    swapfile = Path(args.swap_directory) / 'swapfile'
+    if swapfile.exists():
+        print('swapfile {} already exists'.format(swapfile))
+        sys.exit(1)
+    run('dd if=/dev/zero of={} bs=1GB count={}'.format(swapfile, swapsize))
+    swapfile.chmod(0o600)
+    run('mkswap -f {}'.format(swapfile))
+    run('swapon -v {}'.format(swapfile))
+    with open('/etc/fstab', 'a') as f:
+        f.write('{} swap swap defaults 0 0\n'.format(swapfile))
```
avikivity, line 58: https://www.freedesktop.org/software/systemd/man/systemd.swap.html

dist/common/scripts/scylla_swap_setup
syuu1228: The logic is
```
if 16g <  (memtotal / 3)
    return 16g
else
    return (memtotal / 3)
```
It shouldn't be bigger than 16GB.
referenced: https://docs.scylladb.com/faq/#do-i-need-to-configure-swap-on-a-scylla-node

But agree with it need to to convert an integer, will fix.

syuu1228:
@avikivity v2 pushed, please review again:
Changes from v1:
- Convert memtotal / 3 to int
- use fallocate instead of dd
- use systemd swap unit instead of fstab
- add default value of --swap-directory (default: /)

slivne:
@noamha / @rjeczalik / @AdamNuclear  - I need the info how you create swap and any history of why 

We want to get swap creation in the ami so you guys can stop creating it.

dist/common/scripts/scylla_swap_setup
```diff
@@ -0,0 +1,80 @@
...
+        sys.exit(1)
+    run('fallocate -v -l {}GB {}'.format(swapsize, swapfile))
+    swapfile.chmod(0o600)
+    run('mkswap -f {}'.format(swapfile))
+    swapunit_bn = out('/usr/bin/systemd-escape -p --suffix=swap {}'.format(swapfile))
+    swapunit = Path('/etc/systemd/system/{}'.format(swapunit_bn))
+    if swapunit.exists():
+        print('swap unit {} already exists'.format(swapunit))
+        sys.exit(1)
+    unit_data = '''
```
avikivity, line 65: You can use an f-string here: "unit_data = f```"

dist/common/scripts/scylla_swap_setup
```diff
@@ -0,0 +1,80 @@
...
+[Unit]
+Description=swapfile
+
+[Swap]
+What={}
```
avikivity, line 70: And then you can write `What={swapfile}` here, instead of the .format() call below.

avikivity: This can be a follow-up.

AdamNuclear:
@slivne I don't know history reasons but we are creating swap via this https://github.com/scylladb/siren/blob/master/command/cluster/create-swap.go

noamha:
> @noamha / @rjeczalik / @AdamNuclear - I need the info how you create swap and any history of why
> 
> We want to get swap creation in the ami so you guys can stop creating it.

@slivne The issues we had with swap (at least int he current kernel version):
- Creating it on the root disk as a file caused corruption
- Creating it on a separate swap requires a reboot or else it created corruption on the data partition

PR #6612, opened 2020-06-08
build: default enable dpdk in release mode

avikivity:
To reduce special cases for the build bots, default dpdk to enabled
in release mode, keeping it disabled for debug and dev.

To allow release modes without dpdk to be build, the --enable-dpdk
switch is converted to a tri-state. When disabled, dpdk is disabled
across all modes. Similarly when enabled the effect is global. When
unspecified, dpdk is enabled for release mode only.


avikivity:
@penberg ping

penberg:
Let's remove "--enable-dpdk" from `build_reloc.sh` too?

avikivity:
> Let's remove "--enable-dpdk" from `build_reloc.sh` too?

Yes, will update.

avikivity:
Updated: remove --enable-dpdk from build_reloc.sh.

PR #6619, opened 2020-06-09
scylla_bootparam_setup: support Amazon Linux 2

syuu1228:
CentOS7 uses GRUB_CMDLINE_LINUX on /etc/default/grub, but Amazon Linux 2 only
has GRUB_CMDLINE_LINUX_DEFAULT, we need to support both.


syuu1228:
This is for Amazon Linux 2 AMI: https://github.com/scylladb/scylla-machine-image/pull/44

Issue #6580, opened 2020-06-03
log2utils.hh: pow2_rank_constexpr seems bogus


amnonh:
While working on something unrelated, I've notice the implementation of `pow2_rank_constexpr`

according to its description: // Returns largest N such that 2^N <= v.
But it's off by one.

For example v = 2.
It would return 1 + 1 = 2, instead of 1.
Note that `pow2_rank` does work correctly.

PR #6623, opened 2020-06-10
Adding Optimized pseudo floating point estimated histogram

### LINKED ISSUE #4746: Scylla metrics: read timeouts of 5s are reported in 8s buffer - needs to be fixed
copy pasting from @gleb-cloudius 

> It looks like 8s is not a real value but an artifact of how histograms
> are aggregated for Prometheus. The number of buckets is reduced from 90
> to 20 and all 5s timeouts are aggregated into 8s bucket (there is ~4s and
> ~8s buckets in the Prometheus).

### LINKED ISSUE #5815: estimated_historgram: switch to pseudo floating point for bucket calculation
Based on @avikivity comment:
https://github.com/scylladb/scylla/pull/4828/files#r374765085

The current implementation holds an array of bucket limit and uses binary search to find the right bucket when adding a number.
Instead, use a pseudo gloating point representation, this will reduce the insertion cost and reduce memory consumption by half by eliminating the bucket limits array

amnonh:
This series Adds a pseudo-floating-point histogram implementation.
The histogram is used for time_estimated_histogram a histogram for latency tracking and then used in storage_proxy as a more efficient with a higher resolution histogram.

Follow up series would use the new histogram in other places in the system and will add an implementation that supports lower values.
Fixes #5815
Fixes #4746

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -326,6 +326,7 @@ scylla_tests = set([
     'test/boost/linearizing_input_stream_test',
     'test/boost/loading_cache_test',
     'test/boost/log_heap_test',
+    'test/boost/estimated_histogram_test',
     'test/boost/logalloc_test',
     'test/boost/managed_vector_test',
     'test/boost/map_difference_test',
@@ -1003,6 +1004,7 @@ deps['test/boost/UUID_test'] = ['utils/UUID_gen.cc', 'test/boost/UUID_test.cc', 
 deps['test/boost/murmur_hash_test'] = ['bytes.cc', 'utils/murmur_hash.cc', 'test/boost/murmur_hash_test.cc']
 deps['test/boost/allocation_strategy_test'] = ['test/boost/allocation_strategy_test.cc', 'utils/logalloc.cc', 'utils/dynamic_bitset.cc']
 deps['test/boost/log_heap_test'] = ['test/boost/log_heap_test.cc']
+deps['test/boost/estimated_histogram_test'] = ['test/boost/estimated_histogram_test.cc']
 deps['test/boost/anchorless_list_test'] = ['test/boost/anchorless_list_test.cc']
 deps['test/perf/perf_fast_forward'] += ['release.cc']
 deps['test/perf/perf_simple_query'] += ['release.cc']
--- a/utils/estimated_histogram.hh
+++ b/utils/estimated_histogram.hh
@@ -47,12 +47,301 @@
 #include <chrono>
 #include <seastar/core/metrics_types.hh>
 #include <seastar/core/print.hh>
-
-#include <seastar/core/print.hh>
-
 #include "seastarx.hh"
+#include <seastar/core/bitops.hh>
+#include <limits>
+#include <array>
 
 namespace utils {
+
+/**
+ * This is a pseudo floating point implementation of an estimated histogram.
+ * When entering a value:
+ * All values lower than the MIN will be included in the first bucket.
+ * All values higher than MAX will be included the last bucket that serves as the
+ * infinity bucket.
+ *
+ * buckets are distributed as pseudo floating point:
+ * The range [MIN, MAX) is split into log2 ranges.
+ *     ranges = log2(max/min)
+ * Each of that ranges is split according to the number of buckets:
+ *     resolution = (NUM_BUCKETS - 1)/ranges
+ *
+ * For example, if the MIN value is 128, the MAX is 1024 and the number of buckets is 13:
+ *
+ * Anything below 128 will be in the bucket 0, anything above 1024 will be in bucket 13.
+ *
+ * the range [128, 1024) will be split into log2(1024/128) = 3:
+ * 128, 256, 512, 1024
+ *
+ * Each range is split into 12/3 = 4.
+ * 128            | 256            | 512            | 1024
+ * 128 160 192 224| 256 320 384 448| 512 640 768 896|
+ *
+ *
+ * Calculating the bucket limit of bucket i:
+ * The range: 2^(i/4)* min
+ * The sub range: i%4 * range/4
+ *
+ * How to find a bucket index for a value.
+ * The bucket index consist of two part:
+ * higher bits are based on log2(value/min)
+ *
+ * lower bits are based on the high 2 MSB (ignoring the leading 1).
+
+ * for example: 330 (101001010)
+ * higher bits: log2(330/128) = 1
+ * low bit MSB: 330 = 01 (the lower two bits out of the upper 3)
+ * So the index: 101 = 5
+ *
+ *
+ * About the min/max and number of buckets.
+ * ========================================
+ *
+ * For MIN and MAX choose numbers that are a power of 2.
+ * The number of buckets will determine the resolution that should also be a power of 2
+ * So the total number of bucket should be log2(MAX/MIN) * Resolution + 1
+ *
+ * Limitation: You must set the MIN value to be higher then the resolution.
+ * For example, for a 2 bits resolution MIN should be 2^2 = 4 or higher.
+ *
+ */
+template<uint64_t Min, uint64_t Max, size_t NumBuckets>
+requires (Min > 0 && Min < Max)
+class approx_exponential_histogram {
+    std::array<uint64_t, NumBuckets> _buckets;
+public:
+
+    static constexpr unsigned RANGES = log2floor(Max/Min);
+    static constexpr unsigned RESOLUTION = (NumBuckets - 1)/RANGES;
+    static constexpr unsigned RESOLUTION_BITS = log2floor(RESOLUTION);
+    static constexpr unsigned BASESHIFT = (Min == 0) ? 0 : log2floor(Min);
+    static constexpr uint64_t LOWER_BITS_MASK = (1 << RESOLUTION_BITS) - 1;
+    static_assert(BASESHIFT >= RESOLUTION);
+
+    approx_exponential_histogram() {
+        clear();
+    }
+
+    /*!
+     * \brief Returns the bucket lower limit given the bucket id.
+     * The first and last bucket will always return the MIN and MAX respectively.
+     *
+     */
+    uint64_t get_bucket_lower_limit(uint16_t bucket_id) const {
+        if (bucket_id == 0) {
+            return Min;
+        }
+        if (bucket_id == NumBuckets - 1) {
+            return Max;
+        }
+        int16_t range_id = (bucket_id >> RESOLUTION_BITS);
+        return (1 << (range_id + BASESHIFT)) +  ((bucket_id & LOWER_BITS_MASK) << (range_id + BASESHIFT - RESOLUTION_BITS));
+    }
+
+    /*!
+     * \brief Returns the bucket upper limit given the bucket id.
+     * The last bucket will return MAX.
+     *
+     */
+    uint64_t get_bucket_upper_limit(uint16_t bucket_id) const {
+        if (bucket_id == NumBuckets - 1) {
+            return Max;
+        }
+        return get_bucket_lower_limit(bucket_id + 1);
+    }
+
+    /*!
+     * \brief Find the bucket index for a given value
+     * The position of a value that is lower or equal to Min will always be 0.
+     * The position of a value is that is higher or equal to MAX will always be NUM_BUCKETS - 1.
+     */
+    uint16_t find_bucket_pos(uint64_t val) const {
+        if (val >= Max) {
+            return NumBuckets - 1;
+        }
+        if (val <= Min) {
+            return 0;
+        }
+        uint16_t range = log2floor(val);
+        val >>= range - RESOLUTION_BITS; // leave the top most N+1 bits where N is the resolution.
+        return ((range - BASESHIFT) << RESOLUTION_BITS) + (val & LOWER_BITS_MASK);
+    }
+
+    /*!
+     * \brief returns a cumulative histogram.
+     *
+     * The metrics cumulative histogram uses upper bounds.
+     * The histogram.count serves as an infinite upper bound bucket
+     */
+
+    /*!
+     * \brief clear the current values.
+     */
+    void clear() {
+        std::fill(_buckets.begin(), _buckets.end(), 0);
+    }
+
+    /*!
+     * \brief Add an item to the histogram
+     * Increments the count of the bucket closest to n, rounding DOWN.
+     */
+    void add(uint64_t n) {
+        _buckets.at(find_bucket_pos(n))++;
+    }
+
+    /*!
+     * \brief returns the smallest value that could have been added to this histogram
+     */
+    uint64_t min() const {
+        for (size_t i = 0; i < NumBuckets; i ++) {
+            if (_buckets[i] > 0) {
+                return get_bucket_lower_limit(i);
+            }
+        }
+        return 0;
+    }
+
+    /*!
+     * \brief returns the largest value that could have been added to this histogram.  If the histogram
+     * overflowed, returns UINT64_MAX.
+     */
+    uint64_t max() const {
+        if (_buckets[NumBuckets - 1] > 0) {
+            return UINT64_MAX;
+        }
+        for (int i = NumBuckets - 1; i >= 0; i--) {
+            if (_buckets[i] > 0) {
+                return get_bucket_upper_limit(i);
+            }
+        }
+        return 0;
+    }
+
+    /*!
+     * \brief merge a histogram to the current one.
+     */
+    approx_exponential_histogram& merge(const approx_exponential_histogram& b) {
+        for (size_t i = 0; i < NumBuckets; i++) {
+            _buckets[i] += b.get(i);
+        }
+        return *this;
+    }
+
+    template<uint64_t A, uint64_t B, size_t C>
+    friend approx_exponential_histogram<A, B, C> merge(approx_exponential_histogram<A, B, C> a, const approx_exponential_histogram<A, B, C>& b);
+
+    /*
+     * \brief returns the count in the given bucket
+     */
+    uint64_t get(size_t bucket) const {
+        return _buckets[bucket];
+    }
+
+    /*!
+     * \brief get a histogram quantile
+     *
+     * returns the estimated value at given quantile
+     */
+    uint64_t quantile(double quantile) const {
+        if (quantile < 0 || quantile > 1.0) {
+            throw std::runtime_error("Invalid quantile value " + std::to_string(quantile) + ". Value should be between 0 and 1");
+        }
+        auto c = count();
+
+        if (!c) {
+            return 0; // no data
+        }
+
+        auto pcount = uint64_t(std::floor(c * quantile));
+        uint64_t elements = 0;
+        for (size_t i = 0; i < NumBuckets - 2; i++) {
+            if (_buckets[i]) {
+                elements += _buckets[i];
+                if (elements >= pcount) {
+                    return get_bucket_lower_limit(i);
+                }
+            }
+        }
+        return Max; // overflowed value is in the requested quantile
+    }
+
+    /*!
+     * \brief returns the mean histogram value (average of bucket offsets, weighted by count)
+     */
+    uint64_t mean() const {
+        double elements = 0;
+        uint64_t sum = 0;
+        for (size_t i = 0; i < NumBuckets - 1; i++) {
+            elements += _buckets[i];
+            sum += _buckets[i] * get_bucket_lower_limit(i);
+        }
+        return (sum + elements - 1) / elements;
+    }
+
+    /*!
+     * \brief returns the number of buckets;
+     */
+    size_t size() const {
+        return NumBuckets;
+    }
+
+    /*!
+     * \brief returns the total number of values inserted
+     */
+    uint64_t count() const {
+        uint64_t sum = 0L;
+        for (size_t i = 0; i < NumBuckets; i++) {
+            sum += _buckets[i];
+        }
+        return sum;
+    }
+
+    /*!
+     * \brief multiple all the buckets content in the histogram by a constant
+     */
+    approx_exponential_histogram& operator*=(double v) {
+        for (size_t i = 0; i < NumBuckets; i++) {
+            _buckets[i] *= v;
+        }
+        return *this;
+    }
+};
+
+template<uint64_t Min, uint64_t Max, size_t NumBuckets>
+inline approx_exponential_histogram<Min, Max, NumBuckets> base_estimated_histogram_merge(approx_exponential_histogram<Min, Max, NumBuckets> a, const approx_exponential_histogram<Min, Max, NumBuckets>& b) {
+    return a.merge(b);
+}
+
+/*!
+ * \brief estimated histogram for duration values
+ * time_estimated_histogram is used for common task timing.
+ * It covers the range of 0.5ms to 33s with a 2 bits granularity.
+ *
+ * 512us, 640us, 768us, 896us, 1024us, 1280us, 1536us, 1792us, 2048us, 2560us...
+ */
+class time_estimated_histogram : public approx_exponential_histogram<512, 33554432, 65> {
+public:
+    using clock = std::chrono::steady_clock;
+    using duration = clock::duration;
+    using approx_exponential_histogram<512, 33554432, 65>::add;
+    time_estimated_histogram& merge(const time_estimated_histogram& b) {
+        approx_exponential_histogram<512, 33554432, 65>::merge(b);
+        return *this;
+    }
+
+    void add_micro(uint64_t n) {
+        add(n);
+    }
+
+    void add(const duration& latency) {
+        add_micro(std::chrono::duration_cast<std::chrono::microseconds>(latency).count());
+    }
+};
+
+inline time_estimated_histogram time_estimated_histogram_merge(time_estimated_histogram a, const time_estimated_histogram& b) {
+    return a.merge(b);
+}
 
 struct estimated_histogram {
     using clock = std::chrono::steady_clock;
--- a/utils/histogram_metrics_helper.hh
+++ b/utils/histogram_metrics_helper.hh
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2020 ScyllaDB
+ *
+ * Modified by ScyllaDB
+ */
+
+/*
+ * This file is part of Scylla.
+ *
+ * Scylla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Scylla is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <seastar/core/metrics_types.hh>
+#include "seastarx.hh"
+#include "estimated_histogram.hh"
+
+
+template<uint64_t Min, uint64_t Max, size_t NumBuckets>
+seastar::metrics::histogram to_metrics_histogram(const utils::approx_exponential_histogram<Min, Max, NumBuckets>& hist) {
+    seastar::metrics::histogram res;
+    res.buckets.resize(hist.size() - 1);
+    uint64_t cummulative_count = 0;
+    res.sample_sum = 0;
+
+    for (size_t i = 0; i < NumBuckets - 1; i++) {
+        auto& v = res.buckets[i];
+        v.upper_bound = hist.get_bucket_lower_limit(i + 1);
+        cummulative_count += hist.get(i);
+        v.count = cummulative_count;
+        res.sample_sum += hist.get(i) * v.upper_bound;
+    }
+    // The count serves as the infinite bucket
+    res.sample_count = cummulative_count + hist.get(NumBuckets - 1);
+    res.sample_sum += hist.get(NumBuckets - 1) * hist.get_bucket_lower_limit(NumBuckets - 1);
+    return res;
+}
--- a/api/storage_proxy.cc
+++ b/api/storage_proxy.cc
@@ -116,6 +116,23 @@ static future<json::json_return_type>  sum_timed_rate_as_long(distributed<proxy>
     });
 }
 
+utils_json::estimated_histogram time_to_json_histogram(const utils::time_estimated_histogram& val) {
+    utils_json::estimated_histogram res;
+    for (size_t i = 0; i < val.size(); i++) {
+        res.buckets.push(val.get(i));
+        res.bucket_offsets.push(val.get_bucket_lower_limit(i));
+    }
+    return res;
+}
+
+static future<json::json_return_type>  sum_estimated_histogram(http_context& ctx, utils::time_estimated_histogram service::storage_proxy_stats::stats::*f) {
+
+    return two_dimensional_map_reduce(ctx.sp, f, utils::time_estimated_histogram_merge,
+            utils::time_estimated_histogram()).then([](const utils::time_estimated_histogram& val) {
+        return make_ready_future<json::json_return_type>(time_to_json_histogram(val));
+    });
+}
+
 static future<json::json_return_type>  sum_estimated_histogram(http_context& ctx, utils::estimated_histogram service::storage_proxy_stats::stats::*f) {
 
     return two_dimensional_map_reduce(ctx.sp, f, utils::estimated_histogram_merge,
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -89,6 +89,7 @@
 #include "db/consistency_level_validations.hh"
 #include "cdc/log.hh"
 #include "cdc/stats.hh"
+#include "utils/histogram_metrics_helper.hh"
 
 namespace bi = boost::intrusive;
 
@@ -1452,7 +1453,7 @@ void storage_proxy_stats::write_stats::register_stats() {
     _metrics.add_group(COORDINATOR_STATS_CATEGORY, {
             sm::make_histogram("write_latency", sm::description("The general write latency histogram"),
                     {storage_proxy_stats::current_scheduling_group_label()},
-                    [this]{return estimated_write.get_histogram(16, 20);}),
+                    [this]{return to_metrics_histogram(estimated_write);}),
 
             sm::make_queue_length("foreground_writes", [this] { return writes - background_writes; },
                            sm::description("number of currently pending foreground write requests"),
@@ -1528,7 +1529,7 @@ void storage_proxy_stats::stats::register_stats() {
     _metrics.add_group(COORDINATOR_STATS_CATEGORY, {
         sm::make_histogram("read_latency", sm::description("The general read latency histogram"),
                 {storage_proxy_stats::current_scheduling_group_label()},
-                [this]{ return estimated_read.get_histogram(16, 20);}),
+                [this]{ return to_metrics_histogram(estimated_read);}),
 
         sm::make_queue_length("foreground_reads", foreground_reads,
                 sm::description("number of currently pending foreground read requests"),
@@ -1580,11 +1581,11 @@ void storage_proxy_stats::stats::register_stats() {
 
         sm::make_histogram("cas_read_latency", sm::description("Transactional read latency histogram"),
                 {storage_proxy_stats::current_scheduling_group_label()},
-                [this]{ return estimated_cas_read.get_histogram(16, 20);}),
+                [this]{ return to_metrics_histogram(estimated_cas_read);}),
 
         sm::make_histogram("cas_write_latency", sm::description("Transactional write latency histogram"),
                 {storage_proxy_stats::current_scheduling_group_label()},
-                [this]{return estimated_cas_write.get_histogram(16, 20);}),
+                [this]{return to_metrics_histogram(estimated_cas_write);}),
 
         sm::make_total_operations("cas_write_timeouts", cas_write_timeouts._count,
                        sm::description("number of transactional write request failed due to a timeout"),
@@ -2078,7 +2079,7 @@ future<> storage_proxy::mutate_end(future<> mutate_result, utils::latency_counte
     assert(mutate_result.available());
     stats.write.mark(lc.stop().latency());
     if (lc.is_start()) {
-        stats.estimated_write.add(lc.latency(), stats.write.hist.count);
+        stats.estimated_write.add(lc.latency());
     }
     try {
         mutate_result.get();
@@ -4262,7 +4263,7 @@ storage_proxy::do_query(schema_ptr s,
                         std::move(query_options)).finally([lc, p] () mutable {
                     p->get_stats().read.mark(lc.stop().latency());
                     if (lc.is_start()) {
-                        p->get_stats().estimated_read.add(lc.latency(), p->get_stats().read.hist.count);
+                        p->get_stats().estimated_read.add(lc.latency());
                     }
                 });
             } catch (const no_such_column_family&) {
@@ -4277,7 +4278,7 @@ storage_proxy::do_query(schema_ptr s,
                 std::move(query_options)).finally([lc, p] () mutable {
             p->get_stats().range.mark(lc.stop().latency());
             if (lc.is_start()) {
-                p->get_stats().estimated_range.add(lc.latency(), p->get_stats().range.hist.count);
+                p->get_stats().estimated_range.add(lc.latency());
             }
         });
     }
@@ -4502,8 +4503,8 @@ future<bool> storage_proxy::cas(schema_ptr schema, shared_ptr<cas_request> reque
             get_stats().cas_foreground--;
             write ? get_stats().cas_write.mark(lc.stop().latency()) : get_stats().cas_read.mark(lc.stop().latency());
             if (lc.is_start()) {
-                write ? get_stats().estimated_cas_write.add(lc.latency(), get_stats().cas_write.hist.count) :
-                        get_stats().estimated_cas_read.add(lc.latency(), get_stats().cas_read.hist.count);
+                write ? get_stats().estimated_cas_write.add(lc.latency()) :
+                        get_stats().estimated_cas_read.add(lc.latency());
             }
             if (contentions > 0) {
                 write ? get_stats().cas_write_contention.add(contentions) : get_stats().cas_read_contention.add(contentions);
--- a/service/storage_proxy_stats.hh
+++ b/service/storage_proxy_stats.hh
@@ -90,13 +90,13 @@ struct write_stats {
     utils::timed_rate_moving_average write_timeouts;
 
     utils::timed_rate_moving_average_and_histogram write;
-    utils::estimated_histogram estimated_write;
+    utils::time_estimated_histogram estimated_write;
 
     utils::timed_rate_moving_average cas_write_unavailables;
     utils::timed_rate_moving_average cas_write_timeouts;
 
     utils::timed_rate_moving_average_and_histogram cas_write;
-    utils::estimated_histogram estimated_cas_write;
+    utils::time_estimated_histogram estimated_cas_write;
 
     utils::estimated_histogram cas_write_contention;
 
@@ -169,11 +169,11 @@ struct stats : public write_stats {
 
     utils::timed_rate_moving_average_and_histogram read;
     utils::timed_rate_moving_average_and_histogram range;
-    utils::estimated_histogram estimated_read;
-    utils::estimated_histogram estimated_range;
+    utils::time_estimated_histogram estimated_read;
+    utils::time_estimated_histogram estimated_range;
 
     utils::timed_rate_moving_average_and_histogram cas_read;
-    utils::estimated_histogram estimated_cas_read;
+    utils::time_estimated_histogram estimated_cas_read;
 
     uint64_t reads = 0;
     uint64_t foreground_reads = 0; // client still waits for the read
--- a/test/boost/estimated_histogram_test.cc
+++ b/test/boost/estimated_histogram_test.cc
@@ -1,0 +1,124 @@
+/*
+ * Copyright (C) 2020 ScyllaDB
+ */
+
+/*
+ * This file is part of Scylla.
+ *
+ * Scylla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Scylla is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#define BOOST_TEST_MODULE core
+
+#include <boost/test/unit_test.hpp>
+#include <iostream>
+#include "utils/estimated_histogram.hh"
+#include "utils/histogram_metrics_helper.hh"
+
+template<uint64_t MIN, uint64_t MAX, std::vector<uint64_t>::size_type NUM_BUCKETS>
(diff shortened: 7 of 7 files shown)


avikivity:
Please add numbers to the performance measurement (not just how much faster it was, but also how many nanosecond per operation).

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+
+    void clear() {
+        std::fill(buckets.begin(), buckets.end(), 0);
+    }
+
+    /*!
+     * \brief Add an item to the histogram
+     * Increments the count of the bucket closest to n, rounding DOWN.
+     */
+    histogram_type& add(uint64_t n) {
```
avikivity, line 212: It's better not to return *this, it just muddies the interface and is usually not needed.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+        assert(BASESHIFT >= RESOLUTION);
+        buckets.resize(NUM_BUCKETS, 0);
+    }
+
+    /*!
+     * \brief Returns the bucket limit given the bucket id.
+     * The first and last bucket will always return the MIN and MAX respectively.
+     *
+     */
+    uint64_t get_bucket_limit(uint16_t bucket_id) const {
```
avikivity, line 144: Please make it explicit in the name whether this is the upper bound or lower bound.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+    uint64_t get(size_t bucket) const {
+        return buckets[bucket];
+    }
+
+    /*!
+     * \brief get a histogram percentile
+     *
+     * returns the estimated value at given percentile
+     */
+    int64_t percentile(double perc) const {
```
avikivity, line 270: Maybe we should return both limits here, we don't know if the user wants an upper or lower bound.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+        assert(perc >= 0 && perc <= 1.0);
```
avikivity, line 271: Better call it quantile if it's in the 0-1 range (and not 0-100).

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+
+        auto c = count();
+
+        if (!c) {
+            return 0; // no data
+        }
+
+        auto pcount = int64_t(std::floor(c * perc));
+        int64_t elements = 0;
+        for (size_t i = 0; i < last_bucket; i++) {
```
avikivity, line 282: Aren't you missing the last bucket here? last_bucket is already NUM_BUCKETS - 1.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+                }
+            }
+        }
+        return MAX; // overflowed value is in the requested percentile
+    }
+
+    /*!
+     * \brief return the mean histogram value (average of bucket offsets, weighted by count)
+     */
+    uint64_t mean() const {
```
avikivity, line 296: Again maybe return both bounds.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+        for (size_t i = 0; i < NUM_BUCKETS; i++) {
+            buckets[i] *= v;
+        }
+        return *this;
+    }
+
+    template<uint64_t A, uint64_t B, std::vector<uint64_t>::size_type C>
+    friend std::ostream& operator<<(std::ostream& out, const pseudo_floatingpoint_histogram<A, B, C>& h);
+
+    sstring name_of_range(size_t index) const {
```
avikivity, line 332: This shoud be out-of-line.

configure.py
```diff
@@ -326,6 +326,7 @@ def find_headers(repodir, excluded_dirs):
     'test/boost/linearizing_input_stream_test',
     'test/boost/loading_cache_test',
     'test/boost/log_heap_test',
+    'test/boost/estimated_histogram_test',
     'test/boost/logalloc_test',
```
avikivity, line 330: You can merge this with the previous patch.

amnonh:
@avikivity I thought about it, but then realized that it's a test done on my local machine, does it says anything with absolute numbers?
I'm replying here but I'll also add it to the release.
I've run both comparisons again.

In debug mode 45,850 vs 504,908 microseconds.
In release mode 4,224 vs 13,525 microseconds.

I'll make some more tests and update the patch notes

avikivity:
Every machine is local somewhere.

Please normalize to time per operation. 45850 and 12525 don't mean anything to me.

amnonh:
Sure, I'll add it to the patch

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+#include <limits>
 
 namespace utils {
 
+
+// Returns largest N such that 2^N <= v.
+// Undefined for v == 0.
+inline size_t pow2_rank(uint64_t v) {
+    return std::numeric_limits<uint64_t>::digits - 1 - count_leading_zeros(v);
+}
```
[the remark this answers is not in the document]
amnonh: It has two issues, wrong type, two it's wrong (#6580) but I didn't want to change that in case someone is relying on that wrongness.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+
+
+// Returns largest N such that 2^N <= v.
+// Undefined for v == 0.
+inline constexpr size_t pow2_rank_constexpr(uint64_t v) {
+    return v <= 1 ? 0 : 1 + pow2_rank_constexpr(v >> 1);
+}
```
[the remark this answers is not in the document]
amnonh: I will check

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+
+template<uint64_t MIN, uint64_t MAX, std::vector<uint64_t>::size_type NUM_BUCKETS>
+struct pseudo_floatingpoint_histogram {
```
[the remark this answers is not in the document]
amnonh: I Used template parameters to help the compiler optimize, technically I can even damp the vector and replace it with C array because of that (it brought only an additional 10% improvement so I didn't)

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+     */
+    using histogram_type = pseudo_floatingpoint_histogram<MIN, MAX, NUM_BUCKETS>;
+
+    static constexpr uint8_t RANGES = pow2_rank_constexpr(MAX/MIN);
+    static constexpr uint8_t RESOLUTION = (NUM_BUCKETS - 1)/RANGES;
+    static constexpr uint8_t RESOLUTION_BITS = pow2_rank_constexpr(RESOLUTION);
+    static constexpr uint8_t BASESHIFT = (MIN == 0) ? 0 : pow2_rank_constexpr(MIN);
+    static constexpr uint64_t LOWER_BITS_MASK = (1 << RESOLUTION_BITS) - 1;
+
+    std::vector<uint64_t> buckets;
```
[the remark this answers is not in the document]
amnonh: The additional performance gain was not significant but maybe I'll change to that.

I don't think there will be a large variety of histograms, and I expect that people will not use the template variance explicitly but one of a limited set of derived classes.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+     * The position of a value is that is higher or equal to MAX will always be NUM_BUCKETS - 1.
+     */
+    uint16_t find_bucket_pos(uint64_t val) const {
+        if (val >= MAX) {
+            return NUM_BUCKETS - 1;
+        }
+        if (val <= MIN) {
+            return 0;
+        }
+        uint16_t range_id = pow2_rank(val);
```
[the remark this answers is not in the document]
amnonh: ok

utils/estimated_histogram.hh
amnonh: ok

utils/estimated_histogram.hh
amnonh: I'll check again

utils/estimated_histogram.hh
amnonh: The mean, and percentile are for internal usage and I wanted to keep the current API so it will be easy to migrate.

utils/estimated_histogram.hh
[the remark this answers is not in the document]
avikivity: What's wrong with the type?

Please post a patch to fix it. If someone relies on it, they're broken too.

utils/estimated_histogram.hh
amnonh: ok

configure.py
amnonh: ok

utils/estimated_histogram.hh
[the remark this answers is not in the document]
amnonh: in log_heap.hh it's size_t and I need uint64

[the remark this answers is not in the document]
avikivity: Ah, I was looking at log2ceil, which is a template.

utils/estimated_histogram.hh
amnonh: I've checked again looking at current implementation in estimated_histogram, it's on purpose so a no found value is found the result will be Max.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+        return ((range_id - BASESHIFT) << RESOLUTION_BITS) + (val & LOWER_BITS_MASK);
+    }
+
+    /*!
+     * \brief returns a cumulative histogram.
+     *
+     * The metrics cumulative histogram uses upper bounds.
+     * The histogram.count serves as an infinite upper bound bucket
+     */
+    seastar::metrics::histogram get_histogram() const {
```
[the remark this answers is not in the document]
amnonh: I see your point, I'll make it out of line.
A free function has it's pros, but in practice, it would make anyone that uses a histogram (typically for reporting it) life harder, because they would now need an extra step in returning a result when registering a histogram metrics.
This is how registering a histogram looks like:
```
sm::make_histogram("write_latency", sm::description("The general write latency histogram"),
                    {storage_proxy_stats::current_scheduling_group_label()},
                    [this]{return estimated_write.get_histogram();}),
```

If it's going to be an external function, it means that whoever is using histogram will need to include the file that holds that function and call it.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,376 @@
...
+        // The count serves as the infinite bucket
+        res.sample_count = cummulative_count + buckets[NUM_BUCKETS - 1];
+        res.sample_sum += buckets[NUM_BUCKETS - 1] * get_bucket_limit(NUM_BUCKETS - 1);
+        return res;
+    }
+
+    /*!
+     * \brief return the histogram buckets
+     */
+    const std::vector<uint64_t>& get_buckets() const {
```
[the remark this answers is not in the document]
amnonh: For compatibilities purposes with the existing estimated_histogram implementation

utils/estimated_histogram.hh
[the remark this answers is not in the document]
avikivity: It's not too bad, an additional include, and it makes the new code standalone.

utils/estimated_histogram.hh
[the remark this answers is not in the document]
avikivity: Please see if you can avoid it. Users should access the data via APIs, not by grabbing the internal data structure.

amnonh:
@avikivity I've just forced push V2
I kept it as a template under the assumption that pseudo_floatingpoint_histogram will not be used directly.

Main changes: 
* Switch from struct to class.
* Change template variable to CamelCase.
* Switch from vector to array for the buckets.
* Remove `pow2_rank_constexpr`.
* Explicitly state lower/upper bucket limits.
* Move the conversion function to an external file.
* The `add` methods now return void.
* Remove direct access to the buckets.
* Move some methods outside of the class definitions.
* Replace parameter names were it was not clear.

avikivity:
What about de-templating it? I'd like to reduce compile time for this supporting component.

amnonh:
I think it will be a shame to de-template it, all the constexpr that are used for calculation will either be calculated in run time or calculated and store without letting the compiler optimize them out.
If you insist, I'll change it

amnonh:
I've tested what is the effect of template vs class.
I insert 2^20 values and calculate the average insert.
results are in nanosecond
Using template: 1.8829317092895508
Using class (the same initial calculations are done in the constructor)  4.8328943252563477
Using the old implementation: 9.2892847061157227

avikivity:
Are you sure it's 1.8829317092895508? I would expect 1.8829317092895509, if not 1.882931709289551.

avikivity:
You are right that the template parameter for the scaling is critical. Once that's a template parameter, it doesn't matter if the others are.

utils/estimated_histogram.hh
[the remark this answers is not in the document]
avikivity: So what about log2ceil? I don't want yet another implementation.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
 
 namespace utils {
 
+// Returns largest N such that 2^N <= v.
+// Undefined for v == 0.
+inline constexpr size_t pow2_rank(uint64_t v) {
+    return std::numeric_limits<uint64_t>::digits - 1 - count_leading_zeros(v);
+}
+
+template<uint64_t Min, uint64_t Max, size_t NumBuckets>
```
avikivity, line 62: Please add documentation.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+class pseudo_floatingpoint_histogram {
+    std::array<uint64_t, NumBuckets> buckets;
```
avikivity, line 64: This needs a #include

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+public:
+    /**
+     * This is a pseudo floating point implementation of an estimated histogram.
+     * When entering a value:
+     * All values lower than the MIN will be included in the first bucket.
+     * All values higher than MAX will be included the last bucket that serves as the
+     * infinity bucket.
+     *
+     * buckets are distributed as pseudo floating point:
+     * The range [MIN, MAX) is split into log2 ranges.
```
avikivity, line 74: This documentation should be moved to the class header, not be here in the middle of the class.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+     *
+     * For MIN and MAX choose numbers that are a power of 2.
+     * The number of buckets will determine the resolution that should also be a power of 2
+     * So the total number of bucket should be log2(MAX/MIN) * Resolution + 1
+     *
+     * Limitation: You must set the MIN value to be higher then the resolution.
+     * For example, for a 2 bits resolution MIN should be 2^2 = 4 or higher.
+     *
+     */
+    using histogram_type = pseudo_floatingpoint_histogram<Min, Max, NumBuckets>;
```
avikivity, line 118: Please use just `pseudo_floatingpoint_historgram` to reduce indirection.

Maybe a nicer name is approx_exponential_histogram, no one needs to care about the implementation details.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+
+    static constexpr uint8_t RANGES = pow2_rank(Max/Min);
+    static constexpr uint8_t RESOLUTION = (NumBuckets - 1)/RANGES;
+    static constexpr uint8_t RESOLUTION_BITS = pow2_rank(RESOLUTION);
+    static constexpr uint8_t BASESHIFT = (Min == 0) ? 0 : pow2_rank(Min);
```
avikivity, line 123: Max/Min will fail if Min==0, better add a static_assert in front (or requires).

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+    static constexpr uint64_t LOWER_BITS_MASK = (1 << RESOLUTION_BITS) - 1;
+
+    pseudo_floatingpoint_histogram() {
+        assert(BASESHIFT >= RESOLUTION);
```
avikivity, line 127: static_assert

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+     * \brief returns the largest value that could have been added to this histogram.  If the histogram
+     * overflowed, returns UINT64_MAX.
+     */
+    uint64_t max() const {
+        if (buckets[NumBuckets - 1] > 0) {
+            return UINT64_MAX;
+        }
+        for (int i = NumBuckets - 1; i >= 0; i--) {
+            if (buckets[i] > 0) {
+                return get_bucket_lower_limit(i);
```
avikivity, line 220: upper?

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+    uint64_t get(size_t bucket) const {
+        return buckets[bucket];
+    }
+
+    /*!
+     * \brief get a histogram percentile
+     *
+     * returns the estimated value at given percentile
+     */
+    uint64_t percentile(double quantile) const {
```
avikivity, line 251: s/percentile/quantile/

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+        assert(quantile >= 0 && quantile <= 1.0);
```
avikivity, line 252: dangerous, better to throw than to kill the server

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+    uint64_t mean() const {
+        auto lastBucket = NumBuckets - 1;
+        uint64_t elements = 0;
+        uint64_t sum = 0;
+        for (size_t i = 0; i < lastBucket; i++) {
+            long bCount = buckets[i];
+            elements += bCount;
+            sum += bCount * get_bucket_lower_limit(i);
+        }
+        return ((double) (sum + elements -1)/ elements);
```
avikivity, line 284: Use C++ style conversions, not C casts. Equal spacing aroung `/`.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+            }
+        }
+        return Max; // overflowed value is in the requested percentile
+    }
+
+    /*!
+     * \brief returns the mean histogram value (average of bucket offsets, weighted by count)
+     */
+    uint64_t mean() const {
+        auto lastBucket = NumBuckets - 1;
```
avikivity, line 276: our coding convention is last_bucket.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+    s += format("{:d}", get_bucket_lower_limit(index) + 1);
+    s += "..";
+    if (index == buckets.size()) {
+        s += "Inf";
+    } else {
+        s += format("{:d}", get_bucket_upper_limit(index));
+    }
+    s += "]";
+    return s;
+}
```
avikivity, line 339: This should call into a function in .cc that accepts the template parameters as normal parameters. This avoids having to include fmt.h everywhere (it's already included everywhere but let's not add another reason).

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+
+template<uint64_t A, uint64_t B, std::vector<uint64_t>::size_type C>
+inline std::ostream& operator<<(std::ostream& out, const pseudo_floatingpoint_histogram<A, B, C>& h) {
+    // only print overflow if there is any
+    size_t name_count;
```
avikivity, line 344: This too should be in .cc, with the template parameters passed as normal parameters. You can pass _buckets.data() to avoid having to pass the template.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,359 @@
...
+        // sort-of-hack to not print empty ranges at the start that are only used to demarcate the
+        // first populated range. for code clarity we don't omit this record from the maxNameLength
+        // calculation, and accept the unnecessary whitespace prefixes that will occasionally occur
+        if (i == 0 && count == 0) {
+            continue;
+        }
+        out << format(formatstr.c_str(), names[i], count);
+    }
+    return out;
+}
```
avikivity, line 371: Let's move this to a new file.

utils/estimated_histogram.hh
avikivity: and coding style: _buckets

utils/histogram_metrics_helper.hh
```diff
@@ -0,0 +1,49 @@
...
+#pragma once
+
+#include <seastar/core/metrics_types.hh>
+#include "seastarx.hh"
+#include "estimated_histogram.hh"
+
+
+template<uint64_t Min, uint64_t Max, std::vector<uint64_t>::size_type NumBuckets>
+seastar::metrics::histogram to_metrics_histogram(const utils::pseudo_floatingpoint_histogram<Min, Max, NumBuckets>& hist) {
+    seastar::metrics::histogram res;
```
avikivity, line 33: This can also work by passing the template parameters as non-template parameters to an out-of-line funciton.

avikivity:
> Are you sure it's 1.8829317092895508? I would expect 1.8829317092895509, if not 1.882931709289551.

If you insert random values the old implementation will be much worse

utils/estimated_histogram.hh
amnonh: sure

utils/estimated_histogram.hh
amnonh: ok

utils/estimated_histogram.hh
amnonh: you're right, moving from struct to class

utils/estimated_histogram.hh
amnonh: ok

utils/estimated_histogram.hh
amnonh: ok

amnonh:
@avikivity I've forced push V3.
I hope I didn't miss any of your comments.
Major changes:
* Rename to approx_exponential_histogram
* Move the doc to the top of the class
* Change the validations to `static_assert` and `require`
* Change a function assert to throw
* Remove the redirection operator method from the implementation, it is currently not needed.
* Fix variables name according to the naming convention.
* Use log2floor
* Add test for min, max and mean
* rename all the percentile to quantile

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+                return get_bucket_upper_limit(i);
+            }
+        }
+        return 0;
+    }
+
+    /*!
+     * \brief merge a histogram to the current one.
+     */
+    approx_exponential_histogram<Min, Max, NumBuckets>& merge(const approx_exponential_histogram<Min, Max, NumBuckets>& b) {
```
avikivity, line 224: You can just write approx_exponential_histogram (without the specializtion).

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+        for (size_t i = 0; i < NumBuckets; i++) {
+            sum += _buckets[i];
+        }
+        return sum;
+    }
+
+    /*!
+     * \brief multiple all the buckets content in the histogram by a constant
+     */
+    approx_exponential_histogram<Min, Max, NumBuckets>& operator*=(double v) {
```
avikivity, line 303: here too

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+        for (size_t i = 0; i < NumBuckets; i++) {
+            _buckets[i] *= v;
+        }
+        return *this;
+    }
+};
+
+template<uint64_t Min, uint64_t Max, std::vector<uint64_t>::size_type NumBuckets>
```
avikivity, line 311: And simplify the size_type to size_t

utils/histogram_metrics_helper.hh
avikivity: What about this?

amnonh:
@avikivity I didn't understand what you were referring to with 'what about this'

amnonh:
@avikivity I have forced push V4.

There is one comment about the helper function in utils/histogram_metrics_helper.hh that I didn't understand.

Main changes:
* Remove the template specification from the method declarations.
* Switch from size_type to size_t

avikivity:
> @avikivity I didn't understand what you were referring to with 'what about this'

I guess github dropped the context due to force push. The question was about moving to_metrics_histogram out-of-line to .cc. I guess we can do it later.

utils/histogram_metrics_helper.hh
avikivity: You can still see the context if you go to the individual commit.

amnonh:
@avikivity just to see we're on the same page. Following the comment of V1, I've separate the helper function that translates the approx_exponential_histogram  to the metrics histogram to its own file.

To move it's implementation to a source file it needs to pass the internal vector data, this is no longer available following a different comment in V1.

I can return that option (getting a const reference to the array and use its internal pointer to pass to an implementation in the  source file)

I can move the to_metrics_histogram back to approx_exponential_histogram as a member or friend function (but then we'll have the metrics dependency issue you wanted to avoid).

Is there some other option you prefer?

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+
+    /*!
+     * \brief returns the mean histogram value (average of bucket offsets, weighted by count)
+     */
+    uint64_t mean() const {
+        double elements = 0;
+        uint64_t sum = 0;
+        for (size_t i = 0; i < NumBuckets - 1; i++) {
+            elements += _buckets[i];
+            sum += _buckets[i] * get_bucket_lower_limit(i);
```
nyh, line 277: sum can easily overflow uint64_t, no?

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+        }
+        return (sum + elements - 1) / elements;
```
nyh, line 279: I don't understand the "+ elements - 1" here. What is it supposed to do - it's like adding +1 to get_bucket_lower_limit() but why?

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+        for (size_t i = 0; i < NumBuckets; i++) {
+            sum += _buckets[i];
+        }
+        return sum;
+    }
+
+    /*!
+     * \brief multiple all the buckets content in the histogram by a constant
+     */
+    approx_exponential_histogram& operator*=(double v) {
```
nyh, line 303: That's a really odd operation. Why is it useful? If it's useful, I would gave it a name instead of overloading an operator (you don't overload any other operator...).

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+inline approx_exponential_histogram<Min, Max, NumBuckets> base_estimated_histogram_merge(approx_exponential_histogram<Min, Max, NumBuckets> a, const approx_exponential_histogram<Min, Max, NumBuckets>& b) {
+    return a.merge(b);
+}
+
+/*!
+ * \brief estimated histogram for duration values
+ * time_estimated_histogram is used for common task timing.
+ * It covers the range of 0.5ms to 33s with a 2 bits granularity.
+ *
+ * 512us, 640us, 768us, 896us, 1024us, 1280us, 1536us, 1792us, 2048us, 2560us...
```
nyh, line 321: Would be nice to see the *end*, how big are the last buckets.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+};
+
+template<uint64_t Min, uint64_t Max, size_t NumBuckets>
+inline approx_exponential_histogram<Min, Max, NumBuckets> base_estimated_histogram_merge(approx_exponential_histogram<Min, Max, NumBuckets> a, const approx_exponential_histogram<Min, Max, NumBuckets>& b) {
+    return a.merge(b);
+}
+
+/*!
+ * \brief estimated histogram for duration values
+ * time_estimated_histogram is used for common task timing.
```
nyh, line 318: I'm not sure "common task" is the right phrase here. Maybe "short request"? These parameters are useful mainly for measuring requests - which we *hope* take around one millisecond time, and at worst a few seconds and never 30 seconds. Many other "tasks" don't really have the same assumptions...

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+ * It covers the range of 0.5ms to 33s with a 2 bits granularity.
+ *
+ * 512us, 640us, 768us, 896us, 1024us, 1280us, 1536us, 1792us, 2048us, 2560us...
+ */
+class time_estimated_histogram : public approx_exponential_histogram<512, 33554432, 65> {
+public:
+    using clock = std::chrono::steady_clock;
+    using duration = clock::duration;
+    using approx_exponential_histogram<512, 33554432, 65>::add;
```
nyh, line 327: Why do you need to do this "using", isn't it inherited by default?
By the way, below you provide different add_micro() and add() functions which I supposed you want users to use. I think it would have been better to hide the add(uint64_t) function, instead of exposing it. In add_micro() you could call it directly as approx_exponential_histogram::add().

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
-
 #include "seastarx.hh"
+#include <seastar/core/bitops.hh>
+#include <limits>
+#include <array>
 
 namespace utils {
 
+/**
+ * This is a pseudo floating point implementation of an estimated histogram.
```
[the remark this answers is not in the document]
amnonh: The implementation is similar to pseudo-floating-point implementation.

Giving a range: power of 2 growing sub-ranges (ie. exponential with a coefficient of 2)
Each of those sub-ranges is split linearly.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+ * When entering a value:
+ * All values lower than the MIN will be included in the first bucket.
+ * All values higher than MAX will be included the last bucket that serves as the
+ * infinity bucket.
+ *
+ * buckets are distributed as pseudo floating point:
+ * The range [MIN, MAX) is split into log2 ranges.
```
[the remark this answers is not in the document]
amnonh: i.e. `n` ranges where n equals log2(Max/Min)

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+ *     ranges = log2(max/min)
+ * Each of that ranges is split according to the number of buckets:
+ *     resolution = (NUM_BUCKETS - 1)/ranges
+ *
+ * For example, if the MIN value is 128, the MAX is 1024 and the number of buckets is 13:
+ *
+ * Anything below 128 will be in the bucket 0, anything above 1024 will be in bucket 13.
+ *
+ * the range [128, 1024) will be split into log2(1024/128) = 3:
```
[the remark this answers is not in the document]
amnonh: ok

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+ * 128, 256, 512, 1024
+ *
+ * Each range is split into 12/3 = 4.
```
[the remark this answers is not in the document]
amnonh: The original discussion started here:
https://github.com/scylladb/scylla/pull/4828
The idea is to get high performance with flexible coefficient support

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+ * 128            | 256            | 512            | 1024
+ * 128 160 192 224| 256 320 384 448| 512 640 768 896|
+ *
+ *
+ * Calculating the bucket limit of bucket i:
+ * The range: 2^(i/4)* min
```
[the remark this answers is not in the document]
amnonh: I'll try to think of a better word, Avi didn't like range id

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+ * All values higher than MAX will be included the last bucket that serves as the
+ * infinity bucket.
+ *
+ * buckets are distributed as pseudo floating point:
+ * The range [MIN, MAX) is split into log2 ranges.
+ *     ranges = log2(max/min)
+ * Each of that ranges is split according to the number of buckets:
+ *     resolution = (NUM_BUCKETS - 1)/ranges
+ *
+ * For example, if the MIN value is 128, the MAX is 1024 and the number of buckets is 13:
```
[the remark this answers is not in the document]
amnonh: it would still work but some of the buckets will be empty.
It's documented, but maybe I should make it striker.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+ * low bit MSB: 330 = 01 (the lower two bits out of the upper 3)
+ * So the index: 101 = 5
+ *
+ *
+ * About the min/max and number of buckets.
+ * ========================================
+ *
+ * For MIN and MAX choose numbers that are a power of 2.
+ * The number of buckets will determine the resolution that should also be a power of 2
+ * So the total number of bucket should be log2(MAX/MIN) * Resolution + 1
```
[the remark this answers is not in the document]
amnonh: Ok, I'll look for a different name

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+ *
+ * Limitation: You must set the MIN value to be higher then the resolution.
+ * For example, for a 2 bits resolution MIN should be 2^2 = 4 or higher.
+ *
+ */
+template<uint64_t Min, uint64_t Max, size_t NumBuckets>
+requires (Min > 0 && Min < Max)
```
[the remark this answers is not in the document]
amnonh: Ok, I'll force those requirements which would make some of the code simpler

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+ *
+ * About the min/max and number of buckets.
+ * ========================================
+ *
+ * For MIN and MAX choose numbers that are a power of 2.
+ * The number of buckets will determine the resolution that should also be a power of 2
+ * So the total number of bucket should be log2(MAX/MIN) * Resolution + 1
+ *
+ * Limitation: You must set the MIN value to be higher then the resolution.
+ * For example, for a 2 bits resolution MIN should be 2^2 = 4 or higher.
```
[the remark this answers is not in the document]
amnonh: yes, I'll force the requirement so it would make it clearer

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+class approx_exponential_histogram {
+    std::array<uint64_t, NumBuckets> _buckets;
+public:
+
+    static constexpr unsigned RANGES = log2floor(Max/Min);
+    static constexpr unsigned RESOLUTION = (NumBuckets - 1)/RANGES;
+    static constexpr unsigned RESOLUTION_BITS = log2floor(RESOLUTION);
+    static constexpr unsigned BASESHIFT = (Min == 0) ? 0 : log2floor(Min);
+    static constexpr uint64_t LOWER_BITS_MASK = (1 << RESOLUTION_BITS) - 1;
+    static_assert(BASESHIFT >= RESOLUTION);
```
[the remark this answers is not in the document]
amnonh: ok

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+     */
+    uint64_t get_bucket_lower_limit(uint16_t bucket_id) const {
+        if (bucket_id == 0) {
+            return Min;
+        }
+        if (bucket_id == NumBuckets - 1) {
+            return Max;
+        }
+        int16_t range_id = (bucket_id >> RESOLUTION_BITS);
+        return (1 << (range_id + BASESHIFT)) +  ((bucket_id & LOWER_BITS_MASK) << (range_id + BASESHIFT - RESOLUTION_BITS));
```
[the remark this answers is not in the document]
amnonh: it was because I didn't relly on Min being a power of two, once I'll enforce that, I can do the change

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+    }
+
+    /*!
+     * \brief Returns the bucket upper limit given the bucket id.
+     * The last bucket will return MAX.
```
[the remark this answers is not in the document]
amnonh: There is no special lowest bucket for below Min bucket.
The assumption is that you start with a value that is low enough.

I see your point about the infinite bucket, I'll change its value accordingly.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+        }
+        return get_bucket_lower_limit(bucket_id + 1);
+    }
+
+    /*!
+     * \brief Find the bucket index for a given value
+     * The position of a value that is lower or equal to Min will always be 0.
+     * The position of a value is that is higher or equal to MAX will always be NUM_BUCKETS - 1.
+     */
+    uint16_t find_bucket_pos(uint64_t val) const {
```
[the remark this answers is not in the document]
amnonh: I took the names and some of the implementation from the original estimated_histogram (that goes all the way back to Cassandra).

It's a good idea, and I'll change it

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+    uint64_t get_bucket_upper_limit(uint16_t bucket_id) const {
+        if (bucket_id == NumBuckets - 1) {
+            return Max;
+        }
+        return get_bucket_lower_limit(bucket_id + 1);
+    }
+
+    /*!
+     * \brief Find the bucket index for a given value
+     * The position of a value that is lower or equal to Min will always be 0.
```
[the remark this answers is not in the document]
amnonh: I'll document it better

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+    /*!
+     * \brief clear the current values.
+     */
+    void clear() {
+        std::fill(_buckets.begin(), _buckets.end(), 0);
+    }
+
+    /*!
+     * \brief Add an item to the histogram
+     * Increments the count of the bucket closest to n, rounding DOWN.
```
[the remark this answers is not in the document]
amnonh: ok, I'll make it clearer

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+     */
+    void add(uint64_t n) {
+        _buckets.at(find_bucket_pos(n))++;
+    }
+
+    /*!
+     * \brief returns the smallest value that could have been added to this histogram
```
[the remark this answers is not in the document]
amnonh: ok, I'll try to make it clearer

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+            if (_buckets[i] > 0) {
+                return get_bucket_lower_limit(i);
+            }
+        }
+        return 0;
+    }
+
+    /*!
+     * \brief returns the largest value that could have been added to this histogram.  If the histogram
+     * overflowed, returns UINT64_MAX.
```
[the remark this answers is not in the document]
amnonh: ok

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+    /*!
+     * \brief returns the smallest value that could have been added to this histogram
+     */
+    uint64_t min() const {
+        for (size_t i = 0; i < NumBuckets; i ++) {
+            if (_buckets[i] > 0) {
+                return get_bucket_lower_limit(i);
+            }
+        }
+        return 0;
```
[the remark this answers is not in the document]
amnonh: ok

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+     * \brief returns the count in the given bucket
+     */
+    uint64_t get(size_t bucket) const {
+        return _buckets[bucket];
+    }
+
+    /*!
+     * \brief get a histogram quantile
+     *
+     * returns the estimated value at given quantile
```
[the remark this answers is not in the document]
amnonh: ok

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+     */
+    uint64_t quantile(double quantile) const {
```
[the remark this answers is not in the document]
amnonh: ok

utils/estimated_histogram.hh
amnonh: theoretically, I can just switch to double all the way

utils/estimated_histogram.hh
amnonh: I took the original implementation. I have a vague memory about (s + n-1)/n from statistic but it's been ages.
I did try to look why it was so and decided to keep that implementation.

utils/estimated_histogram.hh
amnonh: I think it was added when it was used when combining histograms but you want to look at time window

utils/estimated_histogram.hh
amnonh: ok

utils/estimated_histogram.hh
amnonh: ok

utils/estimated_histogram.hh
amnonh: ok

amnonh: note that because of C++ overload, redefinition of add(duration) would hide add(uint64)

nyh: Ok, I'm not sure why it hides, but I guess you checked it... In that case, even better that it hides, no? I thought the point of adding add_micro(...) was that the add(uint64_t) is too unclear, so it's better not to have it at all.
By the way, I would go even further, and hide *all* the functions of the base class (i.e., don't inherit public at all). Because all the mean(), quantile(), etc., could be overridden to return times instead of numbers.

utils/estimated_histogram.hh
nyh: I'm curious what it was supposed to mean (pun intended) :-)
As far as I can tell, adding "n" is just like adding 1 to each element, which is kind of silly assuming that the entire bucket is considered equivalent, so for each bucket its minimum and minimum+1 are considered the same anyway. But where does "n-1" come from? As I said, I'm curious. It looked to me completely arbitrary.

utils/estimated_histogram.hh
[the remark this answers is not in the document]
nyh: Oh! I think I understood now why you called it "pseudo floating point": It has the exponent part, and inside each "range" defined by the exponent, you have the mantissa part - which addresses the individual points linearly. 
I think that this is a very interesting analogy. But I think it would not be intuitively clear to readers (it wasn't to me), so it shouldn't be part of the commit message and the first line of the comment. It would be nice, indeed, to have it in a comment or the commit message, but not so prominently.
Or so I think.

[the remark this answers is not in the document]
amnonh: @nyh Thanks for all the other comments as well, it help me see all the places that it was not clear for an external reader.
The next version will be submitted soon where I address it

utils/estimated_histogram.hh
amnonh: I'm not sure, so I just dropped it in the next version

utils/estimated_histogram.hh
amnonh: I believe they are useful especially for things like the API and the metrics layer.
I did follow your comment and hide the add(unit64) one

avikivity:
Note I already queued the last version so improvements will have to come as a follow-up.

amnonh:
@avikivity one of the important comment by @nyh is that the relation between the Min,Max and Number of bucket is unclear.

As soon as the current patch will be merge, I'll send a followup patch
that changes the histogram to use
Min,Max and Precision (and infer the number of buckets from it)

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+    uint16_t find_bucket_pos(uint64_t val) const {
+        if (val >= Max) {
+            return NumBuckets - 1;
+        }
+        if (val <= Min) {
+            return 0;
+        }
+        uint16_t range = log2floor(val);
+        val >>= range - RESOLUTION_BITS; // leave the top most N+1 bits where N is the resolution.
+        return ((range - BASESHIFT) << RESOLUTION_BITS) + (val & LOWER_BITS_MASK);
```
bhalevy, line 168: @amnonh shouldn't `(val & LOWER_BITS_MASK)` be taken before val is shifted right above?
(Ref https://github.com/scylladb/scylla/issues/3806#issuecomment-645803974)

utils/estimated_histogram.hh
bhalevy: @amnonh 
Also, a running (and accurate) _sum can be maintained internally on each add()

utils/estimated_histogram.hh
bhalevy: It looks like rounding up the mean value to the ceil(sum / elements), isn't it?
If that's the intention then rounding by adding elements/2 makes more sense.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+    }
+
+    /*!
+     * \brief returns the total number of values inserted
+     */
+    uint64_t count() const {
+        uint64_t sum = 0L;
+        for (size_t i = 0; i < NumBuckets; i++) {
+            sum += _buckets[i];
+        }
```
bhalevy, line 296: like `_sum` I proposed above, `add()` can also maintain a running `_count`.

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+    approx_exponential_histogram& operator*=(double v) {
+        for (size_t i = 0; i < NumBuckets; i++) {
+            _buckets[i] *= v;
+        }
+        return *this;
+    }
+};
+
+template<uint64_t Min, uint64_t Max, size_t NumBuckets>
+inline approx_exponential_histogram<Min, Max, NumBuckets> base_estimated_histogram_merge(approx_exponential_histogram<Min, Max, NumBuckets> a, const approx_exponential_histogram<Min, Max, NumBuckets>& b) {
```
bhalevy, line 312: @amnonh I don't it used anywhere

utils/estimated_histogram.hh
```diff
@@ -47,13 +47,302 @@
...
+    void add_micro(uint64_t n) {
+        add(n);
+    }
+
+    void add(const duration& latency) {
+        add_micro(std::chrono::duration_cast<std::chrono::microseconds>(latency).count());
+    }
+};
+
+inline time_estimated_histogram time_estimated_histogram_merge(time_estimated_histogram a, const time_estimated_histogram& b) {
```
bhalevy, line 342: @amnonh  I think it should return a `time_estimated_histogram&` rather than a copy of the object, since it's
being called like:
```
                s->result = s->reduce(std::move(s->result), std::move(f.get0()));
```
by `seastar::map_reduce`

So return an object rather than a reference causes 2 extra copies.

utils/estimated_histogram.hh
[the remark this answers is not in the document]
bhalevy: @amnonh also, I think we should protect against overflow of `range_id` by capping `NumBuckets`.

Also, since we return uint64_t, shouldn't we calculate `(uint64_t(1) << (range_id + BASESHIFT))`?
(the rest of the expression should be in `uint64_t` implicitly because of `& LOWER_BITS_MASK`, but we can make it more explicit so it will be more rebost to changes.

utils/histogram_metrics_helper.hh
```diff
@@ -0,0 +1,49 @@
...
+    res.buckets.resize(hist.size() - 1);
+    uint64_t cummulative_count = 0;
+    res.sample_sum = 0;
+
+    for (size_t i = 0; i < NumBuckets - 1; i++) {
+        auto& v = res.buckets[i];
+        v.upper_bound = hist.get_bucket_lower_limit(i + 1);
+        cummulative_count += hist.get(i);
+        v.count = cummulative_count;
+        res.sample_sum += hist.get(i) * v.upper_bound;
```
bhalevy, line 43: I think this should be multiplied by the lower bound (or by the (geometric) average  of the lower and upper bounds) of this bucket since the last upper bound is added outside the loop.

utils/estimated_histogram.hh
bhalevy: I #ifdef'ed it out to see where it's used:
```
table.cc:2411:43: error: no match for ‘operator*=’ (operand types are ‘utils::estimated_histogram’ and ‘double’)
 2411 |         _stats.estimated_coordinator_read *= 0.9; // decay values a little to give new data points more weight
```

If we end up implementing the running count and running sum, we'll need to scale them here respectively,
or recalculate them using the existing methods.

PR #6630, opened 2020-06-10
build: remove zstd submodule

avikivity:
Now that Fedora provides the zstd static library, we can remove the
submodule.

The frozen toolchain is regenerated to include the new package.


avikivity:
@penberg ping

PR #6632, opened 2020-06-10
add out of box configs for GCP VMs with nvmes

### LINKED ISSUE #6631: support gcp disks tuning out of box
Scylla should automatically set the speeds of nvme raids for GCP VMs,
similar as it does for AWS

tarzanek:
not recommended setups will still run iotune
fixes #6631

### DIFF
--- a/dist/common/scripts/scylla_io_setup
+++ b/dist/common/scripts/scylla_io_setup
@@ -30,6 +30,60 @@ import logging
 import sys
 import scylla_blocktune as blocktune
 
+def run_iotune():
+            if "SCYLLA_CONF" in os.environ:
+                conf_dir = os.environ["SCYLLA_CONF"]
+            else:
+                conf_dir = etcdir() + "/scylla"
+            cfg = yaml.safe_load(open(os.path.join(conf_dir, "scylla.yaml")))
+            default_path = cfg.get('workdir') or datadir()
+            if not "data_file_directories" in cfg:
+                cfg["data_file_directories"] = [os.path.join(default_path, 'data')]
+            data_dirs = cfg["data_file_directories"]
+
+            for t in [ "commitlog", "hints", "view_hints", "saved_caches" ]:
+                key = "%s_directory" % t
+                if key in cfg:
+                    data_dirs += [ cfg[key] ]
+                elif os.path.isdir(os.path.join(default_path, t)):
+                    data_dirs += [ os.path.join(default_path, t) ]
+
+            iotune_args = []
+            for data_dir in data_dirs:
+                if os.path.exists(data_dir) == False:
+                    logging.error("%s was not found. Please check the configuration and run scylla_io_setup again.\n", data_dir)
+                    sys.exit(1)
+                if os.path.isdir(data_dir) == False:
+                    logging.error("%s is not a directory. Please check the configuration and run scylla_io_setup again.\n", data_dir)
+                    sys.exit(1)
+                st = os.statvfs(data_dir)
+                avail = st.f_bavail * st.f_frsize
+                rec = 10000000000
+                if avail < rec:
+                    logging.error("Filesystem at %s has only %d bytes available; that is less than the recommended 10 GB. Please free up space and run scylla_io_setup again.\n", data_dir, avail)
+                    sys.exit(1)
+                blocktune.tune_fs(data_dir, '2')
+                iotune_args += [ "--evaluation-directory", data_dir ]
+
+            if cpudata.cpuset():
+                iotune_args += [ "--cpuset", ",".join(map(str, cpudata.cpuset())) ]
+            elif cpudata.smp():
+                iotune_args += [ "--smp", str(cpudata.smp()) ]
+
+            try:
+                subprocess.check_call([bindir() + "/iotune",
+                                       "--format", "envfile",
+                                       "--options-file", etcdir() + "/scylla.d/io.conf",
+                                       "--properties-file", etcdir() + "/scylla.d/io_properties.yaml"] + iotune_args)
+            except Exception as e:
+                logging.error(e)
+                logging.error("%s did not pass validation tests, it may not be on XFS and/or has limited disk space.\n"
+                              "This is a non-supported setup, and performance is expected to be very bad.\n"
+                              "For better performance, placing your data on XFS-formatted directories is required.\n"
+                              "To override this error, enable developer mode as follow:\n"
+                              "sudo %s/scylla_dev_mode_setup --developer-mode 1", data_dirs, scriptsdir())
+                sys.exit(1)
+
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='IO Setup script for Scylla.')
     parser.add_argument('--ami', dest='ami', action='store_true',
@@ -88,56 +142,6 @@ if __name__ == "__main__":
             yaml.dump({ "disks": [ disk_properties ] }, properties_file,  default_flow_style=False)
             ioconf = open(etcdir() + "/scylla.d/io.conf", "w")
             ioconf.write("SEASTAR_IO=\"--io-properties-file={}\"\n".format(properties_file.name))
+
         else:
-            if "SCYLLA_CONF" in os.environ:
-                conf_dir = os.environ["SCYLLA_CONF"]
-            else:
-                conf_dir = etcdir() + "/scylla"
-            cfg = yaml.safe_load(open(os.path.join(conf_dir, "scylla.yaml")))
-            default_path = cfg.get('workdir') or datadir()
-            if not "data_file_directories" in cfg:
-                cfg["data_file_directories"] = [os.path.join(default_path, 'data')]
-            data_dirs = cfg["data_file_directories"]
-
-            for t in [ "commitlog", "hints", "view_hints", "saved_caches" ]:
-                key = "%s_directory" % t
-                if key in cfg:
-                    data_dirs += [ cfg[key] ]
-                elif os.path.isdir(os.path.join(default_path, t)):
-                    data_dirs += [ os.path.join(default_path, t) ]
-
-            iotune_args = []
-            for data_dir in data_dirs:
-                if os.path.exists(data_dir) == False:
-                    logging.error("%s was not found. Please check the configuration and run scylla_io_setup again.\n", data_dir)
-                    sys.exit(1)
-                if os.path.isdir(data_dir) == False:
-                    logging.error("%s is not a directory. Please check the configuration and run scylla_io_setup again.\n", data_dir)
-                    sys.exit(1)
-                st = os.statvfs(data_dir)
-                avail = st.f_bavail * st.f_frsize
-                rec = 10000000000
-                if avail < rec:
-                    logging.error("Filesystem at %s has only %d bytes available; that is less than the recommended 10 GB. Please free up space and run scylla_io_setup again.\n", data_dir, avail)
-                    sys.exit(1)
-                blocktune.tune_fs(data_dir, '2')
-                iotune_args += [ "--evaluation-directory", data_dir ]
-
-            if cpudata.cpuset():
-                iotune_args += [ "--cpuset", ",".join(map(str, cpudata.cpuset())) ]
-            elif cpudata.smp():
-                iotune_args += [ "--smp", str(cpudata.smp()) ]
-
-            try:
-                subprocess.check_call([bindir() + "/iotune",
-                                       "--format", "envfile",
-                                       "--options-file", etcdir() + "/scylla.d/io.conf",
-                                       "--properties-file", etcdir() + "/scylla.d/io_properties.yaml"] + iotune_args)
-            except Exception:
-                logging.error("%s did not pass validation tests, it may not be on XFS and/or has limited disk space.\n"
-                              "This is a non-supported setup, and performance is expected to be very bad.\n"
-                              "For better performance, placing your data on XFS-formatted directories is required.\n"
-                              "To override this error, enable developer mode as follow:\n"
-                              "sudo %s/scylla_dev_mode_setup --developer-mode 1", data_dirs, scriptsdir())
-                sys.exit(1)
-
+            run_iotune()
--- a/dist/common/scripts/scylla_io_setup
+++ b/dist/common/scripts/scylla_io_setup
@@ -142,6 +142,51 @@ if __name__ == "__main__":
             yaml.dump({ "disks": [ disk_properties ] }, properties_file,  default_flow_style=False)
             ioconf = open(etcdir() + "/scylla.d/io.conf", "w")
             ioconf.write("SEASTAR_IO=\"--io-properties-file={}\"\n".format(properties_file.name))
+        elif gcp_instance().is_gce_instance():
+            idata = gcp_instance()
 
+            if idata.is_recommended_instance():
+                disk_properties = {}
+                disk_properties["mountpoint"] = datadir()
+                nr_disks = idata.nvmeDiskCount
+                # below is based on https://cloud.google.com/compute/docs/disks/local-ssd#performance
+                # and https://cloud.google.com/compute/docs/disks/local-ssd#nvme
+                # note that scylla iotune might measure more, this is GCP recommended
+                mbs=1024*1024
+                if nr_disks >= 1 & nr_disks < 4:
+                    disk_properties["read_iops"] = 170000 * nr_disks
+                    disk_properties["read_bandwidth"] = 660 * mbs * nr_disks
+                    disk_properties["write_iops"] = 90000 * nr_disks
+                    disk_properties["write_bandwidth"] = 350 * mbs * nr_disks
+                elif nr_disks >= 4 & nr_disks <= 8:
+                    disk_properties["read_iops"] = 680000
+                    disk_properties["read_bandwidth"] = 2650 * mbs
+                    disk_properties["write_iops"] = 360000
+                    disk_properties["write_bandwidth"] = 1400 * mbs
+                elif nr_disks == "16":
+                    disk_properties["read_iops"] = 1600000
+                    disk_properties["read_bandwidth"] = 4521251328
+                    #below is google, above is our measured
+                    #disk_properties["read_bandwidth"] = 6240 * mbs
+                    disk_properties["write_iops"] = 800000
+                    disk_properties["write_bandwidth"] = 2759452672
+                    #below is google, above is our measured
+                    #disk_properties["write_bandwidth"] = 3120 * mbs
+                elif nr_disks == "24":
+                    disk_properties["read_iops"] = 2400000
+                    disk_properties["read_bandwidth"] = 5921532416
+                    #below is google, above is our measured
+                    #disk_properties["read_bandwidth"] = 9360 * mbs
+                    disk_properties["write_iops"] = 1200000
+                    disk_properties["write_bandwidth"] = 4663037952
+                    #below is google, above is our measured
+                    #disk_properties["write_bandwidth"] = 4680 * mbs
+                properties_file = open(etcdir() + "/scylla.d/io_properties.yaml", "w")
+                yaml.dump({"disks": [disk_properties]}, properties_file,  default_flow_style=False)
+                ioconf = open(etcdir() + "/scylla.d/io.conf", "w")
+                ioconf.write("SEASTAR_IO=\"--io-properties-file={}\"\n".format(properties_file.name))
+            else:
+                logging.error('This is not a recommended Google Cloud instance setup for auto tuning, running manual iotune.')
+                run_iotune()
         else:
             run_iotune()
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -106,6 +106,209 @@ def curl(url, byte=False, headers={}):
             if (retries >= max_retries):
                 raise
 
+class gcp_instance:
+    """Describe several aspects of the current GCP instance"""
+
+    EPHEMERAL = "ephemeral"
+    ROOT = "root"
+
+    def __init__(self):
+        self.__type = None
+        self.__cpu = None
+        self.__memoryGB = None
+        self.__nvmeDiskCount = None
+        self.__firstNvmeSize = None
+        self.__osDisks = None
+
+    @staticmethod
+    def is_gce_instance():
+        """Check if it's GCE instance via DNS lookup to metadata server."""
+        import socket
+        try:
+            addrlist = socket.getaddrinfo('metadata.google.internal', 80)
+        except socket.gaierror:
+            return False
+        for res in addrlist:
+            af, socktype, proto, canonname, sa = res
+            if af == socket.AF_INET:
+                addr, port = sa
+                if addr == "169.254.169.254":
+                    return True
+        return False
+
+    def __instance_metadata(self, path):
+        """query GCP metadata server, recursively!"""
+        return curl("http://169.254.169.254/computeMetadata/v1/instance/" + path+"?recursive=true", headers={"Metadata-Flavor": "Google"})
+        #169.254.169.254 is metadata.google.internal
+
+    def _non_root_nvmes(self):
+        """get list of nvme disks from os, filter away if one of them is root"""
+        nvme_re = re.compile(r"nvme\d+n\d+$")
+
+        root_dev_candidates = [x for x in psutil.disk_partitions() if x.mountpoint == "/"]
+        if len(root_dev_candidates) != 1:
+            raise Exception("found more than one disk mounted at root ".format(root_dev_candidates))
+
+        root_dev = root_dev_candidates[0].device
+        # if root_dev.startswith("/dev/mapper"):
+        #     raise Exception("mapper used for root, not checking if nvme is used ".format(root_dev))
+
+        nvmes_present = list(filter(nvme_re.match, os.listdir("/dev")))
+        return {self.ROOT: [root_dev], self.EPHEMERAL: [x for x in nvmes_present if not root_dev.startswith(os.path.join("/dev/", x))]}
+
+    @property
+    def os_disks(self):
+        """populate disks from /dev/ and root mountpoint"""
+        if self.__osDisks is None:
+            __osDisks = {}
+            nvmes_present = self._non_root_nvmes()
+            for k, v in nvmes_present.items():
+                __osDisks[k] = v
+            self.__osDisks = __osDisks
+        return self.__osDisks
+
+    def getEphemeralOsDisks(self):
+        """return just transient disks"""
+        return self.os_disks[self.EPHEMERAL]
+
+    @staticmethod
+    def isNVME(gcpdiskobj):
+        """check if disk from GCP metadata is a NVME disk"""
+        if gcpdiskobj["interface"]=="NVME":
+            return True
+        return False
+
+    def __get_nvme_disks_from_metadata(self):
+        """get list of nvme disks from metadata server"""
+        import json
+        try:
+            disksREST=self.__instance_metadata("disks")
+            disksobj=json.loads(disksREST)
+            nvmedisks=list(filter(self.isNVME, disksobj))
+        except Exception as e:
+            print ("Problem when parsing disks from metadata:")
+            print (e)
+            nvmedisks={}
+        return nvmedisks
+
+    @property
+    def nvmeDiskCount(self):
+        """get # of nvme disks available for scylla raid"""
+        if self.__nvmeDiskCount is None:
+            try:
+                ephemeral_disks = self.getEphemeralOsDisks()
+                count_os_disks=len(ephemeral_disks)
+            except Exception as e:
+                print ("Problem when parsing disks from OS:")
+                print (e)
+                count_os_disks=0
+            nvme_metadata_disks = self.__get_nvme_disks_from_metadata()
+            count_metadata_nvme_disks=len(nvme_metadata_disks)
+            self.__nvmeDiskCount = count_os_disks if count_os_disks<count_metadata_nvme_disks else count_metadata_nvme_disks
+        return self.__nvmeDiskCount
+
+    @property
+    def instancetype(self):
+        """return the type of this instance, e.g. n2-standard-2"""
+        if self.__type is None:
+            self.__type = self.__instance_metadata("machine-type").split("/")[-1]
+        return self.__type
+
+    @property
+    def cpu(self):
+        """return the # of cpus of this instance"""
+        if self.__cpu is None:
+            self.__cpu = psutil.cpu_count()
+        return self.__cpu
+
+    @property
+    def memoryGB(self):
+        """return the size of memory in GB of this instance"""
+        if self.__memoryGB is None:
+            self.__memoryGB = psutil.virtual_memory().total/1024/1024/1024
+        return self.__memoryGB
+
+    def instance_size(self):
+        """Returns the size of the instance we are running in. i.e.: 2"""
+        return self.instancetype.split("-")[2]
+
+    def instance_class(self):
+        """Returns the class of the instance we are running in. i.e.: n2"""
+        return self.instancetype.split("-")[0]
+
+    def instance_purpose(self):
+        """Returns the purpose of the instance we are running in. i.e.: standard"""
+        return self.instancetype.split("-")[1]
+
+    m1supported="m1-megamem-96" #this is the only exception of supported m1 as per https://cloud.google.com/compute/docs/machine-types#m1_machine_types
+
+    def is_unsupported_instance(self):
+        """Returns if this instance type belongs to unsupported ones for nvmes"""
+        if self.instancetype == self.m1supported:
+            return False
+        if self.instance_class() in ['e2', 'f1', 'g1', 'm2', 'm1']:
+            return True
+        return False
+
+    def is_supported_instance(self):
+        """Returns if this instance type belongs to supported ones for nvmes"""
+        if self.instancetype == self.m1supported:
+            return True
+        if self.instance_class() in ['n1', 'n2', 'n2d' ,'c2']:
+            return True
+        return False
+
+    def is_recommended_instance_size(self):
+        """if this instance has at least 2 cpus, it has a recommended size"""
+        if int(self.instance_size()) > 1:
+            return True
+        return False
+
+    @staticmethod
+    def get_file_size_by_seek(filename):
+        "Get the file size by seeking at end"
+        fd= os.open(filename, os.O_RDONLY)
+        try:
+            return os.lseek(fd, 0, os.SEEK_END)
+        finally:
+            os.close(fd)
+
+    # note that GCP has 3TB physical devices actually, which they break into smaller 375GB disks and share the same mem with multiple machines
+    # this is a reference value, disk size shouldn't be lower than that
+    GCP_NVME_DISK_SIZE_2020=375
+
+    @property
+    def firstNvmeSize(self):
+        """return the size of first non root NVME disk in GB"""
+        if self.__firstNvmeSize is None:
+            ephemeral_disks = self.getEphemeralOsDisks()
+            firstDisk = ephemeral_disks[0]
+            firstDiskSize = self.get_file_size_by_seek(os.path.join("/dev/", firstDisk))
+            firstDiskSizeGB = firstDiskSize/1024/1024/1024
+            if firstDiskSizeGB >= self.GCP_NVME_DISK_SIZE_2020:
+                self.__firstNvmeSize = firstDiskSizeGB
+            else:
+                raise Exception("First nvme is smaller than lowest expected size. ".format(firstDisk))
+        return self.__firstNvmeSize
+
+    def is_recommended_instance(self):
+        if self.is_recommended_instance_size() and not self.is_unsupported_instance() and self.is_supported_instance():
+            # at least 1:2GB cpu:ram ratio , GCP is at 1:4, so this should be fine
+            if self.cpu/self.memoryGB < 0.5:
+              # 30:1 Disk/RAM ratio must be kept at least(AWS), we relax this a little bit
+              # on GCP we are OK with 50:1 , n1-standard-2 can cope with 1 disk, not more
+              diskCount = self.nvmeDiskCount
+              # to reach max performance for > 16 disks we mandate 32 or more vcpus
+              # https://cloud.google.com/compute/docs/disks/local-ssd#performance
+              if diskCount >= 16 and self.cpu < 32:
+                  return False
+              diskSize= self.firstNvmeSize
+              if diskCount < 1:
+                  return False
+              disktoramratio = (diskCount*diskSize)/self.memoryGB
+              if (disktoramratio <= 50) and (disktoramratio > 0):
+                  return True
+        return False
 
 class aws_instance:
     """Describe several aspects of the current AWS instance"""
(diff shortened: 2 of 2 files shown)


tarzanek:
OK, removed disk size hardcode
we now tune all possible variations out of box depending on disk count
I think the only problem now is if there would be a non nvme ssd disk, how would it check it, so I need to retest this negative test case, if we want to have this code enabled for all GCP machines with nvmes

dist/common/scripts/scylla_io_setup
```diff
@@ -30,6 +30,60 @@ import logging
 import sys
 import scylla_blocktune as blocktune
 
+def run_iotune():
```
vladzcloudius, line 33: When you send patches to scylla-dev make sure to put this change in a separate patch.

dist/common/scripts/scylla_util.py
```diff
@@ -106,6 +110,137 @@ def curl(url, byte=False):
...
+    def __disk_name(self, dev):
+        name = re.compile(r"(?:/dev/)?(?P<devname>[a-zA-Z]+)\d*")
+        return name.search(dev).group("devname")
+
+    def __device_exists(self, dev):
+        if dev[0:4] != "/dev":
+            dev = "/dev/%s" % dev
+        return os.path.exists(dev)
+
+    def get_file_size(self, filename):
```
vladzcloudius, line 155: That's a little unorthodox way of finding a file size? Did you consider `fstat` instead? 
See here: https://docs.python.org/3/library/os.html#os.stat_result
And see an example of the usage in perftune.py: __learn_directory()

dist/common/scripts/scylla_util.py
```diff
@@ -106,6 +110,137 @@ def curl(url, byte=False):
...
+            return True
+        return False
+
+    def is_supported_instance_type(self):
+        if self.instance_type() in ['standard', 'highmem', 'ultramem', 'megamem']:
+            return True
+        return False
+
+    def is_supported_instance(self):
+        if self.is_supported_instance_class() and self.is_supported_instance_type() and self.is_supported_instance_size() and self._populated:
```
vladzcloudius, line 228: Why do we want to limit the support to only these instances?
I think we should support any instance.

dist/common/scripts/scylla_util.py
tarzanek: doesn't work, I rewrote it to use sysfs instead

tarzanek:
let me squash the diffs and make just 2 diffs, the refactoring to function and the actual new checks and recommended disk iotune params

vladzcloudius:
> OK, removed disk size hardcode
> we now tune all possible variations out of box depending on disk count
> I think the only problem now is if there would be a non nvme ssd disk, how would it check it, so I need to retest this negative test case, if we want to have this code enabled for all GCP machines with nvmes
> that said the regexp should match it and skip those disks, but I didn't try yet
> 
> if the configuration won't match recommended machine + nvmes, it will run iotune manually

We need to make sure that we have NVMe cards at hand.
For instance on my laptop I can see this:
```
sudo fdisk -l | grep -i nvme | grep -i "disk model"
Disk model: PC401 NVMe SK hynix 512GB 
```

vladzcloudius:
On GCP instance I see this:
```
$ sudo lspci -vvv
.....
00:04.0 Non-Volatile memory controller: Google, Inc. Device 001f (rev 01) (prog-if 02 [NVM Express])
        Physical Slot: 4
        Control: I/O+ Mem+ BusMaster+ SpecCycle- MemWINV+ VGASnoop- ParErr- Stepping- SERR- FastB2B- DisINTx+
        Status: Cap+ 66MHz- UDF- FastB2B- ParErr- DEVSEL=fast >TAbort- <TAbort- <MAbort- >SERR- <PERR- INTx-
        Latency: 0
        Interrupt: pin A routed to IRQ 11
        NUMA node: 0
        Region 0: Memory at c0000000 (64-bit, non-prefetchable) [size=16K]
        Region 2: I/O ports at c0a0 [size=8]
        Region 4: Memory at c0100000 (32-bit, non-prefetchable) [size=512]
        Capabilities: [80] MSI-X: Enable+ Count=16 Masked-
                Vector table: BAR=4 offset=00000008
                PBA: BAR=4 offset=00000000
        Kernel driver in use: nvme
        Kernel modules: nvme

```

dist/common/scripts/scylla_util.py
```diff
@@ -106,6 +110,135 @@ def curl(url, byte=False):
...
+    def __device_exists(self, dev):
+        if dev[0:4] != "/dev":
+            dev = "/dev/%s" % dev
+        return os.path.exists(dev)
+
+    def get_blockdev_size(self, blockdevname):
+      "Get the file size by checking sysfs"
+      blocks = int(open('/sys/block/{blockdevname}/size'.format(**locals())).read())
+      block_size_bytes = int(open('/sys/block/{blockdevname}/queue/physical_block_size'.format(**locals())).read())
+      return blocks * block_size_bytes
```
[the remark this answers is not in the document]
vladzcloudius: The most straight forward is to query the Google metadata server:

```
[vladz@vladz-centos7-nvme-check ~]$ curl "http://metadata.google.internal/computeMetadata/v1/instance/disks/1/" -H "Metadata-Flavor: Google"
device-name
index
interface
mode
type
[vladz@vladz-centos7-nvme-check ~]$ curl "http://metadata.google.internal/computeMetadata/v1/instance/disks/1/device-name" -H "Metadata-Flavor: Google"
local-ssd-0[vladz@vladz-centos7-nvme-check ~]$ curl "http://metadata.google.internal/computeMetadata/v1/instance/disks/1/interface" -H "Metadata-Flavor: Google"
NVME[vladz@vladz-centos7-nvme-check ~]$ curl "http://metadata.google.internal/computeMetadata/v1/instance/disks/1/type" -H "Metadata-Flavor: Google"
LOCAL-SSD[vladz@vladz-centos7-nvme-check ~]$ 
```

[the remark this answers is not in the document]
vladzcloudius:
```
[vladz@vladz-centos7-nvme-check ~]$ curl "http://metadata.google.internal/computeMetadata/v1/instance/disks/" -H "Metadata-Flavor: Google"
0/
1/
2/
```

dist/common/scripts/scylla_util.py
```diff
@@ -106,6 +110,135 @@ def curl(url, byte=False):
...
+            return True
+        return False
+
+    def is_supported_instance_type(self):
+        if self.instance_type() in ['standard', 'highmem', 'ultramem', 'megamem']:
+            return True
+        return False
+
+    def is_supported_instance(self):
+        if self.is_supported_instance_size() and self._populated:
```
vladzcloudius, line 226: And what if it's not populated?

tarzanek: then iotune will be run, since we lack either cpu, or memory or disk info

tarzanek:
> 
> 
> On GCP instance I see this:
> 
> ```
> $ sudo lspci -vvv
> .....
> 00:04.0 Non-Volatile memory controller: Google, Inc. Device 001f (rev 01) (prog-if 02 [NVM Express])
>         Physical Slot: 4
>         Control: I/O+ Mem+ BusMaster+ SpecCycle- MemWINV+ VGASnoop- ParErr- Stepping- SERR- FastB2B- DisINTx+
>         Status: Cap+ 66MHz- UDF- FastB2B- ParErr- DEVSEL=fast >TAbort- <TAbort- <MAbort- >SERR- <PERR- INTx-
>         Latency: 0
>         Interrupt: pin A routed to IRQ 11
>         NUMA node: 0
>         Region 0: Memory at c0000000 (64-bit, non-prefetchable) [size=16K]
>         Region 2: I/O ports at c0a0 [size=8]
>         Region 4: Memory at c0100000 (32-bit, non-prefetchable) [size=512]
>         Capabilities: [80] MSI-X: Enable+ Count=16 Masked-
>                 Vector table: BAR=4 offset=00000008
>                 PBA: BAR=4 offset=00000000
>         Kernel driver in use: nvme
>         Kernel modules: nvme
> ```
> 
> And of course:
> 
> ```
> [vladz@vladz-centos7-nvme-check ~]$ lsblk 
> NAME    MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
> sda       8:0    0   128G  0 disk 
> ├─sda1    8:1    0   200M  0 part /boot/efi
> └─sda2    8:2    0 127.8G  0 part /
> nvme0n1 259:0    0   375G  0 disk 
> nvme0n2 259:1    0   375G  0 disk 
> ```
> 
> However you may want to double check that SSD devices don't get the `nvm...`-like name.
> I think they should not.

yes, I tested this, they don't get the nvme prefix

dist/common/scripts/scylla_util.py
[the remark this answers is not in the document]
tarzanek: ok, so this is a gotcha
I tested only with physical nvmes and it seems google partitions the 3TB nvme which backs up small disks and kernel is provided with parent info ... d'oh! (I'd even say this is googles virtualization bug of presenting the disk to VM)
so I need some better way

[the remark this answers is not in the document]
tarzanek: I wanted to avoid parsing too much data from metadata server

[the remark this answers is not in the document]
vladzcloudius: I understand and I agree that this is their bug, however we can't wait till they fix it.
IMO the best way to move forward is to work with their metadata server output.

dist/common/scripts/scylla_io_setup
```diff
@@ -88,56 +142,42 @@ if __name__ == "__main__":
...
-            except Exception:
-                logging.error("%s did not pass validation tests, it may not be on XFS and/or has limited disk space.\n"
-                              "This is a non-supported setup, and performance is expected to be very bad.\n"
-                              "For better performance, placing your data on XFS-formatted directories is required.\n"
-                              "To override this error, enable developer mode as follow:\n"
-                              "sudo %s/scylla_dev_mode_setup --developer-mode 1", data_dirs, scriptsdir())
-                sys.exit(1)
+        elif gcp_instance().is_gce_instance() and (is_debian_variant() or is_redhat_variant()):
+            idata = gcp_instance()
+            idata.populate()  # lazy populate, avoid curl calls in case it's not gce_instance
```
vladzcloudius, line 147: I don't think what populate() does is what's called "lazy".
It would be lazy if you wouldn't read info until you needed it (try to read it) which doesn't seem to be the case.

dist/common/scripts/scylla_util.py
vladzcloudius: By looking at the code I think that it that case the corresponding code would throw. Otherwise you'd always get to setting that variable to `True`. 

BTW, it's very unexpected to access the private variable (that's what '_' usually denotes) directly. You can either drop the '_' or introduce accessor (see examples in perftune.py): may be either a '@property' or a plain method (function).

dist/common/scripts/scylla_io_setup
```diff
@@ -88,56 +142,42 @@ if __name__ == "__main__":
...
-                                       "--options-file", etcdir() + "/scylla.d/io.conf",
-                                       "--properties-file", etcdir() + "/scylla.d/io_properties.yaml"] + iotune_args)
-            except Exception:
-                logging.error("%s did not pass validation tests, it may not be on XFS and/or has limited disk space.\n"
-                              "This is a non-supported setup, and performance is expected to be very bad.\n"
-                              "For better performance, placing your data on XFS-formatted directories is required.\n"
-                              "To override this error, enable developer mode as follow:\n"
-                              "sudo %s/scylla_dev_mode_setup --developer-mode 1", data_dirs, scriptsdir())
-                sys.exit(1)
+        elif gcp_instance().is_gce_instance() and (is_debian_variant() or is_redhat_variant()):
```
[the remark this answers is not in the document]
tarzanek: this is to rule out a combination of e.g. gentoo + gcp

obviously, the question is whether you want deeper limiting (e.g. smallest kernel version supported, or os release ? )
do I get your comment properly?

dist/common/scripts/scylla_util.py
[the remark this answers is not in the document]
tarzanek: metadata doesn't contain the disk size,
previously I used the seek method, so that would be a portable check (I think there is no spinning on nvmes, I'd expect it'd jump directly till the end)
resp. on our talk we agreed that I will revert back to previous hardcode of the disk size to 375GB

dist/common/scripts/scylla_util.py
tarzanek: fixed

dist/common/scripts/scylla_io_setup
tarzanek: fixed

dist/common/scripts/scylla_util.py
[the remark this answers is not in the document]
tarzanek: as agreed I will take the number of disks from metadata
then fetch the root device if it is nvme, if not, then I'd keep all the disks for the raid (as our raid setup does that), 
if yes, then I'd subtract this from the metadata count

tarzanek:
current patch lacks the metadata disks parsing for the count of disks, still testing it

dist/common/scripts/scylla_io_setup
tarzanek: fixed

dist/common/scripts/scylla_util.py
tarzanek: fixed

tarzanek:
alternatives to hardcoded disk size is either a "seek" hack:
```

File: dist/common/scripts/scylla_util.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (tarzanek) [reply dist/common/scripts/scylla_util.py]: resolved with metadata check

File: dist/common/scripts/scylla_io_setup (hunk 7 above)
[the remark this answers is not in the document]
Comment (tarzanek) [reply dist/common/scripts/scylla_io_setup:145]: Notes from talk:
- remove debian OSes to be the same with AWS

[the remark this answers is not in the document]
Comment (tarzanek) [reply dist/common/scripts/scylla_io_setup:145]: actually to be the same as AWS, we shouldn't check debian or redhat, just go with every Linux flavour

**Comment by tarzanek** [comment]
@avikivity @penberg could you have a look at this if possible?
can it be merged?

**Comment by bhalevy** [comment]
@xemul please review

File: dist/common/scripts/scylla_io_setup (hunk 7 above)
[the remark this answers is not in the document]
Comment (avikivity) [reply dist/common/scripts/scylla_io_setup:145]: yes

File: dist/common/scripts/scylla_io_setup [hunk 8]
```diff
@@ -142,6 +142,39 @@ if __name__ == "__main__":
...
 
+            if idata.is_recommended_instance():
+                disk_properties = {}
+                disk_properties["mountpoint"] = datadir()
+                nr_disks = idata.nvmeDiskCount
+                if nr_disks >= 1 & nr_disks < 4:
+                    disk_properties["read_iops"] = 179861 * nr_disks
+                    disk_properties["read_bandwidth"] = 737304512 * nr_disks
+                    disk_properties["write_iops"] = 99963 * nr_disks
+                    disk_properties["write_bandwidth"] = 409628288 * nr_disks
```
avikivity, line 156: How constant are these values? Please run iotune on ~10 instances in a few regions to see that we don't get surprises.

dist/common/scripts/scylla_io_setup
```diff
@@ -142,6 +142,39 @@ if __name__ == "__main__":
...
+                elif nr_disks >= 4 & nr_disks <= 8:
+                    disk_properties["read_iops"] = 721719
+                    disk_properties["read_bandwidth"] = 2951769344
```
avikivity, line 159: How come the rates for 4 disks are the same as the rates for 8 disks?

dist/common/scripts/scylla_util.py
```diff
@@ -87,12 +87,16 @@ def datadir():
 def scyllabindir():
     return str(scyllabindir_p())
 
-def curl(url, byte=False):
+# @param headers semicolon ; separated values of k:v
```
avikivity, line 90: Please change to a dict and convert in the function.

dist/common/scripts/scylla_util.py
```diff
@@ -105,6 +109,141 @@ def curl(url, byte=False):
...
+        self.__nvmeDiskCount = None
+
+    @staticmethod
+    def is_gce_instance():
+      """Check if it's GCE instance via DNS lookup to metadata server."""
+      import socket
+      try:
+        addrlist = socket.getaddrinfo('metadata.google.internal', 80)
+      except socket.gaierror:
+        return False
```
avikivity, line 131: Please use standard Python indents of 4 spaces.

dist/common/scripts/scylla_util.py
```diff
@@ -105,6 +109,141 @@ def curl(url, byte=False):
...
+      for res in addrlist:
+        af, socktype, proto, canonname, sa = res
+        if (af == socket.AF_INET):
+          addr, port = sa
+          if (addr == "169.254.169.254"):
```
avikivity, line 136: No need for parentheses in Python if statements.

dist/common/scripts/scylla_util.py
```diff
@@ -105,6 +109,141 @@ def curl(url, byte=False):
...
+      return False
+
+    def __instance_metadata(self, path):
+        """query GCP metadata server, recursively!"""
+        return curl("http://169.254.169.254/computeMetadata/v1/instance/" + path+"?recursive=true", headers="Metadata-Flavor: Google")
+        #169.254.169.254 is metadata.google.internal
+
+    def _non_root_nvmes(self):
+        """get list of nvme disks from os, filter away if one of them is root"""
+        nvme_re = re.compile(r"nvme\d+n\d+$")
```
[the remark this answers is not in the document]
avikivity: Can we differentiate between local and remote disks?

dist/common/scripts/scylla_util.py
```diff
@@ -105,6 +109,141 @@ def curl(url, byte=False):
...
+        root_dev_candidates = [x for x in psutil.disk_partitions() if x.mountpoint == "/"]
+        if len(root_dev_candidates) != 1:
+            raise Exception("found more than one disk mounted at root'".format(root_dev_candidates))
+
+        root_dev = root_dev_candidates[0].device
+        if root_dev.startswith("/dev/mapper"):
+            raise Exception("mapper used for root, not checking if nvme is used".format(root_dev))
+
+        nvmes_present = list(filter(nvme_re.match, os.listdir("/dev")))
+        return {"root": [root_dev], "ephemeral": [x for x in nvmes_present if not root_dev.startswith(os.path.join("/dev/", x))]}
```
avikivity, line 158: How can you be sure it is ephemeral? In AWS there can be nvme EBS drives.

dist/common/scripts/scylla_util.py
```diff
@@ -105,6 +109,141 @@ def curl(url, byte=False):
...
+
+    def is_recommended_instance_size(self):
+        """if this instance has at least 2 cpus, it has a recommended size"""
+        if int(self.instance_size()) > 1:
+            return True
+        return False
+
+    # TODO below is in GB and is hardcoded, so this needs to be changed once the disks in GCP change
+    # note that they have 3TB physical devices actually, which they break into smaller 375GB disks and share the same mem with multiple machines
+    GCP_NVME_DISK_SIZE_2020=375
```
avikivity, line 232: This needs to be fixed.

dist/common/scripts/scylla_util.py
```diff
@@ -105,6 +109,141 @@ def curl(url, byte=False):
...
+
+    def is_recommended_instance(self):
+        if self.is_recommended_instance_size():
+            # at least 1:2GB cpu:ram ratio , GCP is at 1:4, so this should be fine
+            if self.cpu/self.memoryGB < 0.5:
+              # 30:1 Disk/RAM ratio must be kept at least(AWS), we relax this a little bit
+              # on GCP we are OK with 50:1 , n1-standard-2 can cope with 1 disk, not more
+              diskcount = self.nvmeDiskCount
```
avikivity, line 240: We're okay with 100:1

dist/common/scripts/scylla_io_setup
tarzanek: 1 controller is used, they have 3TB nvme cards, which they partition in 8x375GB
they limit the speed for lower than 4 cards (they share same nvme card between smaller instances)
for more than 4 you get the full card without sharing

dist/common/scripts/scylla_util.py
tarzanek: @avikivity  can you help me deciding what is better than this hardcode?

alternatives to hardcoded disk size is either a "seek" hack(which Vlad didn't like that much):
```
def get_file_size(filename):
    "Get the file size by seeking at end"
    fd= os.open(filename, os.O_RDONLY)
    try:
        return os.lseek(fd, 0, os.SEEK_END)
    finally:
        os.close(fd)
```

or a call to fdisk -l and parse from Disk info

or https://pypi.org/project/blkinfo/ (which would add dependency to code from pip)

dist/common/scripts/scylla_io_setup
tarzanek: as per test docs https://docs.google.com/spreadsheets/d/1lMsGTfX0HQURH1L-A8czQXHhwLARQHBjXUDNWoEwrl4/edit#gid=2122362307
this was run on each instances, test repeated 5x, smallest number picked
However the variation between retests was around 7% for some of test repeats, which I think is partially due to Pavels fix for seastar (https://github.com/scylladb/seastar/commit/83919516c98deb78de4e6d909bb4468426c7836d (iotune: Report disk IOPS instead of kernel IOPS) ) was missing when the tests were done, it makes sense for me to do a quick retest of these numbers

tarzanek: I only tested in us-east1-d region, I can try EU regions

dist/common/scripts/scylla_util.py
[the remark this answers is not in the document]
tarzanek: we are getting count of local nvmes from metadata and that is the source of truth - this code is mainly a check if nvme is used as root disk (fwiw, we do the same for AWS)
so source of truth for local nvme count are metadata (see below for the call to metadata parsing - __get_nvme_disks_from_metadata which is used in nvmeDiskCount, unless this list is smaller(which means they used local nvme for root disk))

dist/common/scripts/scylla_util.py
tarzanek: this is legacy code similar to AWS, I can use other name for the list
the idea was to make some calls between AWS and GCP generic and share them between the classes, but I guess it's not a good idea

avikivity: Making the code generic is a good idea. But what about my question?

dist/common/scripts/scylla_util.py
avikivity: How about the stat() function?

Adding more dependencies should be avoided (better to add a dependency than to open-code, but in this case, stat() is fine).

tarzanek: stat didn't work unfortunately for block devices

dist/common/scripts/scylla_util.py
tarzanek: you cannot be sure, that's why the count from metadata is the primary source
that said in GCP you cannot have nvme attached disks at the moment
+ my tests that tested other SSD drives attached showed their naming is standard and doesn't use /dev/nvme*
so GCP uses /dev/nvme only for local attached ssd disks

dist/common/scripts/scylla_util.py
tarzanek: I even tried sysfs ( disk_size= /sys/block/nvme0n1/queue/physical_block_size * /sys/block/nvme0n1/size ) and sysfs is buggy on GCP since it shows full size for the whole nvme card (so 3TB) )

avikivity: ok. I think seek() is fine then.

dist/common/scripts/scylla_util.py
tarzanek: the idea was to stay similar to AWS, should we relax it even more? (I am voting for no ;-)  stay at 50:1  )

avikivity: I don't understand.

tarzanek: so I am voting for staying at current ratio
don't relax it more, I can change the comment and say scylla can handle more

dist/common/scripts/scylla_io_setup
avikivity: Is this documented somewhere? please add a link.

Is there a risk that we'll lose some of the bandwidth if another VM on the same host uses the remaining disks?

dist/common/scripts/scylla_util.py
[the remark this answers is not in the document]
avikivity: I didn't understand the bottom line, can we differentiate between local and remote disks?

dist/common/scripts/scylla_util.py
avikivity: "at the moment" may change in the next moment. But our images will remain.

So I'm trying to make it robust against changes.

dist/common/scripts/scylla_util.py
avikivity: The 100:1 ratio is for Scylla, not the cloud.



AWS I3en have 70:1 ratio or so.

dist/common/scripts/scylla_io_setup
tarzanek: documented on https://cloud.google.com/compute/docs/disks/#localssds (I will add this to code as doc pointer), resp. in detail on https://cloud.google.com/compute/docs/disks/local-ssd#performance
+ if you try to add a local nvme in https://console.cloud.google.com/compute/instancesAdd?project=upload-179716 it will show you changes to performance for 1-4 and then 6-8 will stay same as per above doc (the testing confirms this behaviour)

Documentation doesn't show what happens if more VMs share the same card, above docs suggest however what I mentioned before (that >4 disks == you get the whole controller for yourself).

avikivity: It's really strange. from the 16 and 24 numbers, I expect 100k read IOPS per disk. But in the 1-4 range, you get 170k read IOPS per disk. in the 4-8 range, it stops increasing.

tarzanek: I'd say it's totally in line with table https://cloud.google.com/compute/docs/disks/local-ssd#nvme
my empirical results are even more optimistic than their table.

avikivity: It's better to use their table then. There could be variance between machines.

dist/common/scripts/scylla_util.py
tarzanek: so you're right Avi,
and this is sample metadata:
"disks":[{"deviceName":"scylla-disk-test2","index":0,"interface":"SCSI","mode":"READ_WRITE","type":"PERSISTENT-SSD"},{"deviceName":"local-ssd-0","index":1,"interface":"NVME","mode":"READ_WRITE","type":"LOCAL-SSD"},{"deviceName":"local-ssd-1","index":2,"interface":"NVME","mode":"READ_WRITE","type":"LOCAL-SSD"}]

none of above is visible in OS though (since the data visible to OS is just partitioned 3TB card)
(so mapping it to OS is N/A)

I have a bad feeling we won't be able to be robust here unless Google fixes their device exposure to VM and add a way to map drives in some sane way
from above I really could just see the count of disks (which however is enough for us to set the limit)
what can break here is that we could pick wrong disk for size check and in worst case the system won't be set directly, but iotune will run for it (fwiw this sounds robust enough for me)

and we can of course add a smarter check once Google really adds support for persistent nvme disks
- fwiw AWS also doesn't solve this problem - so how about we address this once this is fully clear and address it for all clouds we support?
(I will file a tracking issue for it)

avikivity: Let me understand - you assigned 8 375GB disks, and it shows up as 2 disks in the metadata?

Let's at least check the instance type. So if a newer instance comes along with different disk setup, we don't assume anything about it.

tarzanek: no no, 2 disks in metadata show as 2 nvme disks in OS, but it's not possible to create the mapping between metadata and what info OS provides (e.g. local-ssd-1 from metadata is not to be found anywhere in OS and similar for other metadata about local nvmes)

I can add back the instance check, it was dropped due to ratio checks are more generic ... but then I think it's not needed, the code will be safe enough (and same approach is used in AWS without any error reports)

avikivity: It is not safe. The numbers in the table are for those instance types and not others. 

If tomorrow GCP adds an instance type with different local disks, we don't want to use those numbers.

tarzanek: ok, so then it makes sense to check if we have 32 vcpus and then use the numbers for 16 or 24 disks
otherwise run iotune
and also respect https://cloud.google.com/compute/docs/disks#local_ssds_and_machine_types and restrict based on them
should we also use the flat numbers directly from the table instead of our measurements?

tarzanek: also if we want to be diligent, we shouldn't even proceed with machines that don't support local ssds as per https://cloud.google.com/compute/docs/machine-types#memory-optimized_machine_type_family

avikivity: I don't understand why we need to count vcpus. Can't we determine the instance type from the metadata?

We can support machines without local disks, but need to run iotune (on the persistent disk). Unless we can determine the persistent disk's performance.

avikivity: We can use the table's numbers directly, if our kernel is able to reach them.

tarzanek: per tests with iotune without Pavels fixes, we can reach more iops now, but I have no problem to lower it to what Google says it supports (note that I won't do other geos testing or retest with new iotune, since in scope of this it doesn't make sense - I am taking Google numbers as granted)

yes, instance type is fetched from metadata, I just dropped the checks, but I will add them back as per above docs

and yes, if machine won't fall into the blessed category iotune will be run(this is already the behaviour)

dist/common/scripts/scylla_util.py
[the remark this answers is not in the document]
tarzanek: we can differentiate only based on seeing "nvme" since right now no other disk in GCP is creating nvme devices in /dev/
(and as pointed in other comments, I didn't find a way to do mapping between metadata about local disks and OS presented devices)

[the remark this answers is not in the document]
avikivity: Each time you use the word "right now" you can expect me to explain again that google can add new instance types and our integration has to work with them.

dist/common/scripts/scylla_io_setup
tarzanek: adjusted per their table, note that for 16 and 24 disks our results were LOWER than theirs (on recommended VM sizes (see https://docs.google.com/spreadsheets/d/1lMsGTfX0HQURH1L-A8czQXHhwLARQHBjXUDNWoEwrl4/edit#gid=2122362307 ), so I used our results instead
I will retest though with Pavels fixes in seastar iotune, for now keeping what was measured for 16 and 24 disk bandwidths

dist/common/scripts/scylla_util.py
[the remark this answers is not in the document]
tarzanek: fixed, I restricted the auto tuning code to only kick in for supported instances and anything else will trigger a new run of iotune

tarzanek:
@avikivity thank you for all the reviews, sir!
The code is testing now for instances that allow NVMEs properly and we can detect them, other instances are ignored and iotune runs as it does for unknown setups
the numbers are adjusted as per Google docs ( minus the ones where our tests showed something else performance wise)
Do I need to fix something else?
if not I will do a quick round of final tests and from my side it will look ready for merge

dist/common/scripts/scylla_io_setup
```diff
@@ -142,6 +142,39 @@ if __name__ == "__main__":
...
+                    disk_properties["read_iops"] = 1682373
+                    disk_properties["read_bandwidth"] = 4521251328
+                    disk_properties["write_iops"] = 910985
+                    disk_properties["write_bandwidth"] = 2759452672
+                elif nr_disks == "24":
+                    disk_properties["read_iops"] = 2502141
+                    disk_properties["read_bandwidth"] = 5921532416
+                    disk_properties["write_iops"] = 1290502
+                    disk_properties["write_bandwidth"] = 4663037952
+                properties_file = open(etcdir() + "/scylla.d/io_properties.yaml", "w")
```
avikivity, line 184: Is it possible to have nr_disks = 19?

In any case we need to detect it (with an extra else clause), not just fall through and write a partial file.

dist/common/scripts/scylla_util.py
avikivity: I didn't see this fixed.

dist/common/scripts/scylla_util.py
avikivity: Or this.

dist/common/scripts/scylla_io_setup
```diff
@@ -149,6 +149,7 @@ if __name__ == "__main__":
                 disk_properties = {}
                 disk_properties["mountpoint"] = datadir()
                 nr_disks = idata.nvmeDiskCount
+                # below is based on https://cloud.google.com/compute/docs/disks/local-ssd#performance
                 if nr_disks >= 1 & nr_disks < 4:
```
avikivity, line 156: Please fold fixes into the patches that introduce the problems. I'm not doing this folding in my mind.

dist/common/scripts/scylla_io_setup
tarzanek: not possible, GCP blocks it

dist/common/scripts/scylla_io_setup
tarzanek: so there are patches based on your comments, e.g.
https://github.com/scylladb/scylla/pull/6632/commits/3f5a8b0a6ed1b4e4e843d4c33c454b1867ffaec6
(see small selector on left top side of patch view, you can select which patches you want to see there, even create sets of them to review what changed since you last reviewed)
I hope this is what you meant, or you want me to squash the patches?
(or split current patches to smaller ones?)

avikivity: Each patch should be about a single topic.

Changes that fix review comments should be folded into the patches that had the problems in the first place.

dist/common/scripts/scylla_util.py
tarzanek: fixed in https://github.com/scylladb/scylla/pull/6632/commits/97c250e4d19b11d7ca1ed2937cc64cdcebda4441

tarzanek:
@avikivity forgot to ping you here
I changed the commits so they only change their theme + any of them should be standalone
I validated and tested the changes on sample instances (positive and negative test cases)
is it ok like this?

dist/common/scripts/scylla_util.py
```diff
@@ -120,19 +120,19 @@ def __init__(self):
...
     @staticmethod
     def is_gce_instance():
-      """Check if it's GCE instance via DNS lookup to metadata server."""
-      import socket
-      try:
-        addrlist = socket.getaddrinfo('metadata.google.internal', 80)
-      except socket.gaierror:
+        """Check if it's GCE instance via DNS lookup to metadata server."""
+        import socket
+        try:
```
avikivity, line 127: This commit should be folded into the commit that introduced the problem. We try not to introduce problems and fix them in the same series.

tarzanek: would it make then sense to squash everything into original 

commit b93284b00197526e894d9f83ce8ff9e3d05f8175
Author: Lubos Kosco <tarzanek@gmail.com>
Date:   Mon Jun 22 16:04:22 2020 +0200

    add out of box configs for GCP VMs with nvmes

    not recommended setups will still run iotune
    fixes #6631

?

because that commit basically says it all :-)
I am kinda confused now with want we want to achieve with the commits and their messages.

I understand we want:
1) have a clear message in the commit about what it does
2) be able to review fixes to original commit (do we care?)
3) have a sane commits for backports

hence the suggestion to fold everything into original commit (and have just the refactor iotune commit and adding the supported/recommended configs)

tarzanek: of course the message would be:
```
scylla_io_setup: add out of box configs for GCP VMs with nvmes

not recommended setups will still run iotune
fixes #6631
```
with the subsystem prefix (if prefix isn't free form, some LOV should define it)

tarzanek: I'll have a look at other messages used in repo

avikivity: > would it make then sense to squash everything into original
> 
> commit [b93284b](https://github.com/scylladb/scylla/commit/b93284b00197526e894d9f83ce8ff9e3d05f8175)
> Author: Lubos Kosco [tarzanek@gmail.com](mailto:tarzanek@gmail.com)
> Date: Mon Jun 22 16:04:22 2020 +0200
> 
> ```
> add out of box configs for GCP VMs with nvmes
> 
> not recommended setups will still run iotune
> fixes #6631
> ```
> 
> ?
> 
> because that commit basically says it all :-)
> I am kinda confused now with want we want to achieve with the commits and their messages.
> 
> I understand we want:
> 
> 1. have a clear message in the commit about what it does
> 2. be able to review fixes to original commit (do we care?)
> 3. have a sane commits for backports
> 
> hence the suggestion to fold everything into original commit (and have just the refactor iotune commit and adding the supported/recommended configs)

A huge patch, especially with a two-line description, is very hard to review.

If you move something out and then make a small change, it is much easier to review as two patches. For the first patch that moves things, I make a cursory check to ensure that's all (or, if I have prior good experience with the contributor, I just skip it). The changes in the second patch then stand out from the noise of the code movement and can be reviewed in isolation.

tarzanek:
@avikivity would this work?

dist/common/scripts/scylla_util.py
tarzanek: I guess my conflict is mainly the continuity of review process
since you saw the original patch
and the new commits were mainly the resolved noise from first patch (which all made sense and I did fix them hopefully to satisfy all reqs)

but I guess I should more than the noise worry about first patch (and as you said fold fixes into first patch, which in this case it was all of them)

tarzanek:
thank you @avikivity !
I will try to improve my commits, I think based on your comments and Vlads explanations I finally understand where you come from
(and at the same time GitHub won't be able to satisfy the demands in its current form, so you just moved my thinking over to the mailing list patches side of the review process (though coloring of diffs helps a lot :-D can I get it in a mail client? ) ... 
it seems the biggest prob is the new version of patches and their history rewrite - which github would be only be able to do with opening and closing PRs for every new patchset version - which sounds like lots of branches to me, or maybe there is another way)

avikivity:
@dekimir has a trick where he lists the incremental patches in a comment, _and_ squashes them into their parent commits. So this both makes re-review for someone already in the loop, and makes it possible for a new reviewer to see the new proposed patches.

dekimir:
I create a private branch off the current PR, then I implement each feedback item in it as a separate commit.  Then I reset the PR's branch to this new branch and perform `rebase -i` there.  After that I force-push the PR branch and add a PR comment that's literally a `git log --oneline` of the private branch with individual commits.  GitHub recognizes commit hashes (even though they are from my own repo) in that comment and makes them into links.

vladzcloudius:
@dekimir Could you, please, send a link to the PR you mentioned?

dekimir:
[Here](https://github.com/scylladb/scylla/pull/5763#issuecomment-614240146) is a good example of the comment Avi described.

PR #6633, opened 2020-06-10
database, streaming, messaging: drop streaming memtables

avikivity:
Before Scylla 3.0, we used to send streaming mutations using
individual RPC requests and flush them together using dedicated
streaming memtables. This mechanism is no longer in use and all
versions that use it have long reached end-of-life.

Remove this code.

### DIFF
--- a/api/api-doc/messaging_service.json
+++ b/api/api-doc/messaging_service.json
@@ -249,7 +249,7 @@
                  "MIGRATION_REQUEST",
                  "PREPARE_MESSAGE",
                  "PREPARE_DONE_MESSAGE",
-                 "STREAM_MUTATION",
+                 "UNUSED__STREAM_MUTATION",
                  "STREAM_MUTATION_DONE",
                  "COMPLETE_MESSAGE",
                  "REPAIR_CHECKSUM_RANGE",
--- a/database.cc
+++ b/database.cc
@@ -1612,20 +1612,6 @@ future<> database::apply_hint(schema_ptr s, const frozen_mutation& m, tracing::t
     });
 }
 
-future<> database::apply_streaming_mutation(schema_ptr s, utils::UUID plan_id, const frozen_mutation& m, bool fragmented) {
-    if (!s->is_synced()) {
-        throw std::runtime_error(format("attempted to mutate using not synced schema of {}.{}, version={}",
-                                 s->ks_name(), s->cf_name(), s->version()));
-    }
-    return with_scheduling_group(_dbcfg.streaming_scheduling_group, [this, s = std::move(s), &m, fragmented, plan_id] () mutable {
-        return _streaming_dirty_memory_manager.region_group().run_when_memory_available([this, &m, plan_id, fragmented, s = std::move(s)] {
-            auto uuid = m.column_family_id();
-            auto& cf = find_column_family(uuid);
-            cf.apply_streaming_mutation(s, plan_id, std::move(m), fragmented);
-        }, db::no_timeout);
-    });
-}
-
 keyspace::config
 database::make_keyspace_config(const keyspace_metadata& ksm) {
     keyspace::config cfg;
--- a/streaming/stream_transfer_task.cc
+++ b/streaming/stream_transfer_task.cc
@@ -130,47 +130,6 @@ struct send_info {
     }
 };
 
-future<stop_iteration> do_send_mutations(lw_shared_ptr<send_info> si, frozen_mutation fm, bool fragmented) {
-    return get_local_stream_manager().mutation_send_limiter().wait().then([si, fragmented, fm = std::move(fm)] () mutable {
-        sslog.debug("[Stream #{}] SEND STREAM_MUTATION to {}, cf_id={}", si->plan_id, si->id, si->cf_id);
-        auto fm_size = fm.representation().size();
-        // Do it in the background.
-        (void)netw::get_local_messaging_service().send_stream_mutation(si->id, si->plan_id, std::move(fm), si->dst_cpu_id, fragmented, si->reason).then([si, fm_size] {
-            sslog.debug("[Stream #{}] GOT STREAM_MUTATION Reply from {}", si->plan_id, si->id.addr);
-            get_local_stream_manager().update_progress(si->plan_id, si->id.addr, progress_info::direction::OUT, fm_size);
-            si->mutations_done.signal();
-        }).handle_exception([si] (auto ep) {
-            // There might be larger number of STREAM_MUTATION inflight.
-            // Log one error per column_family per range
-            if (!si->error_logged) {
-                si->error_logged = true;
-                sslog.warn("[Stream #{}] stream_transfer_task: Fail to send STREAM_MUTATION to {}: {}", si->plan_id, si->id, ep);
-            }
-            si->mutations_done.broken();
-        }).finally([] {
-            get_local_stream_manager().mutation_send_limiter().signal();
-        });
-        return stop_iteration::no;
-    });
-}
-
-future<> send_mutations(lw_shared_ptr<send_info> si) {
-    size_t fragment_size = default_frozen_fragment_size;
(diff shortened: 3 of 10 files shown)


denesb:
Finally!

api/api-doc/messaging_service.json
```diff
@@ -249,7 +249,7 @@
                  "MIGRATION_REQUEST",
                  "PREPARE_MESSAGE",
                  "PREPARE_DONE_MESSAGE",
-                 "STREAM_MUTATION",
+                 "UNUSED_STREAM_MUTATION",
```
denesb, line 252: This reads like a legit verb name, maybe separate `UNUSED` with two `_`, like: `UNUSED__STREAM_MTUATION`.

streaming/stream_transfer_task.cc
```diff
@@ -270,7 +229,7 @@ future<> stream_transfer_task::execute() {
             if (si->db.features().cluster_supports_stream_with_rpc_stream()) {
                 return send_mutation_fragments(std::move(si));
             } else {
-                return send_mutations(std::move(si));
+                throw std::runtime_error("cluster does not STREAM_WITH_RPC_STREAM feature");
```
denesb, line 232: I failed to parse this message.

table.cc
```diff
@@ -1820,63 +1695,9 @@ future<> table::flush() {
 // be indiscriminately touching the cache during repair. We will just have to
 // invalidate the entries that are relevant to things we already have in the cache.
 future<> table::flush_streaming_mutations(utils::UUID plan_id, dht::partition_range_vector ranges) {
```
denesb, line 1697: Why not remove altogether? Who needs this?

slivne:
In general I think that removing features and adding features should be done in a similar manner - behind a feature bit.

api/api-doc/messaging_service.json
avikivity: Ok.

streaming/stream_transfer_task.cc
avikivity: Will fix.

table.cc
avikivity: It is called from the new streaming code too (and ends up just advancing the phaser).

avikivity:
> In general I think that removing features and adding features should be done in a similar manner - behind a feature bit.

How can we add a feature bit for something we remove?

In any case this will only fail if someone tries to add a 4.2 node to a 2.3 cluster.

avikivity:
Updated:
 - use UNUSED__ prefix (two underscores instead of one) to make it clear that UNUSED isn't part of the verb name
 - improved exception message when someone tries to call the missing code

table.cc
denesb: I see. Maybe we should rename it to something else, flush is pretty meaningless without the streaming memtables. Can be done in a follow-up.

denesb: (and rename the `_streaming_flush_phaser` too).

slivne:
On Thu, Jun 11, 2020 at 1:14 PM Avi Kivity <notifications@github.com> wrote:

> In general I think that removing features and adding features should be
> done in a similar manner - behind a feature bit.
>
> How can we add a feature bit for something we remove?
>
It will require 2 releases
- First release
   - feature bit marking message will be deprecated, allowing to have the
feature (rpc verb) used till all nodes mark it as deprecated from that
point its not supported.
- Next release
   - actual removal

>
>
In any case this will only fail if someone tries to add a 4.2 node to a 2.3

glommer:
awesome!  I was planning to do it myself, just waiting for a good time. Will review now

streaming/stream_transfer_task.cc
```diff
@@ -270,7 +229,7 @@ future<> stream_transfer_task::execute() {
             if (si->db.features().cluster_supports_stream_with_rpc_stream()) {
                 return send_mutation_fragments(std::move(si));
             } else {
-                return send_mutations(std::move(si));
+                throw std::runtime_error("cluster does not support STREAM_WITH_RPC_STREAM feature");
```
glommer, line 232: Can we try to be a bit friendlier to the one odd user who decided to try their luck migration Scylla 2.0 to Scylla 4.2 ?
The message should maybe explain why the failure, that they should use an intermediate version, and preferably which version.

avikivity:
> On Thu, Jun 11, 2020 at 1:14 PM Avi Kivity ***@***.***> wrote: In general I think that removing features and adding features should be done in a similar manner - behind a feature bit. How can we add a feature bit for something we remove?
> It will require 2 releases - First release - feature bit marking message will be deprecated, allowing to have the feature (rpc verb) used till all nodes mark it as deprecated from that point its not supported. - Next release - actual removal

Let's say we do this and 4.2 marks this as deprecated. In 4.3 we remove it. What happens if someone tries to add a 4.3 node to a 2.3 cluster?

The feature bit solves nothing.

streaming/stream_transfer_task.cc
avikivity: I can expand the message, though I fear a user trying to add a 4.2 node to a  2.3 cluster isn't very good at reading messages.

database.cc
```diff
@@ -1608,20 +1608,6 @@ future<> database::apply_hint(schema_ptr s, const frozen_mutation& m, tracing::t
     });
 }
 
-future<> database::apply_streaming_mutation(schema_ptr s, utils::UUID plan_id, const frozen_mutation& m, bool fragmented) {
```
glommer, line 1611: @avikivity I don't see you deleting the _streaming_dirty_memory_manager construction statement in the database constructor, or its .hh counterparts.

the streaming_concurrency_sem I guess is used by new repair too ?

There are also metrics in database.cc that reference the virtual dirty for streaming, but I guess those will naturally disappear once you remove the .hh declaration

denesb: The semaphore is used by new streaming, including row-level repair.

streaming/stream_transfer_task.cc
glommer: Let's act like God and show infinite patience by giving the user one more change.
And true to character, if they don't listen to this last advice we're justified in flooding their home and killing everyone inside

database.cc
avikivity: I planned to attack _streaming_dirty_memory_manager in a follow up, since it's more involved - we'll have to redistribute its 10%.

streaming_concurrency_sem  is still used.

glommer: awesome.

slivne:
On Thu, Jun 11, 2020 at 2:34 PM Avi Kivity <notifications@github.com> wrote:

> On Thu, Jun 11, 2020 at 1:14 PM Avi Kivity *@*.***> wrote: In general I
> think that removing features and adding features should be done in a
> similar manner - behind a feature bit. How can we add a feature bit for
> something we remove?
> It will require 2 releases - First release - feature bit marking message
> will be deprecated, allowing to have the feature (rpc verb) used till all
> nodes mark it as deprecated from that point its not supported. - Next
> release - actual removal
>
> Let's say we do this and 4.2 marks this as deprecated. In 4.3 we remove
> it. What happens if someone tries to add a 4.3 node to a 2.3 cluster?
>

It does solve it (the feature bit is never removed - we only add new ones)

We do a check of feature bits when a new node connects to the cluster. A
node of 4.3 can't join a cluster that doesn't have the minimum set of
feature bits.

The feature bit solves nothing.

avikivity:
I don't think we have a way to prevent a node from joining a cluster if the cluster doesn't have enough bits (because we don't know who's joining who). It's only the other way around.

avikivity:
ping

avikivity:
@scylladb/scylla-maint ping

tgrabiec:
@avikivity This branch has conflicts, please rebase

avikivity:
Rebased and retested, repushed.

avikivity:
Committed as e5be3352cf7557f7509e698b1915b009c9d86ed5

PR #6634, opened 2020-06-11
range: rename range template family to interval

avikivity:
nonwrapping_range<T> and related templates represent mathematical
intervals, and are different from C++ ranges. This causes confusion,
especially when C++ ranges and the range templates are used together.

As the first step to disentable this, introduce a new interval.hh
header with the contents of the old range.hh header, renaming as
follows:

  range_bound  -> interval_bound
  nonwrapping_range -> nonwrapping_interval
  wrapping_range -> wrapping_interval
  Range -> Interval (concepts)

The range alias, which previously aliased wrapping_range, did
not get renamed - instead the interval alias now aliases
nonwrapping_interval, which is the natural interval type. I plan
to follow up making interval the template, and nonwrapping_interval
the alias (or perhaps even remove it).

To avoid churn, a new range.hh header is provided with the old names
as aliases (range, nonwrapping_range, wrapping_range, range_bound,
and Range) with the same meaning as their former selves.

Tests: unit (dev)


avikivity:
Note to reviewers: this is easier to review using git rather than github, as git understands the rename+modify while github shows it as a new file.

denesb:
I'm not a big fan of this rename. The term "range" has a long and extensive legacy in the C* ecosystem, I don't associate to a or clustering-range partition-range from the word "interval".

> nonwrapping_range and related templates represent mathematical
intervals, and are different from C++ ranges. This causes confusion,
especially when C++ ranges and the range templates are used together.

Is this confusing to the compiler too or mainly to developers?

avikivity:
> I'm not a big fan of this rename. The term "range" has a long and extensive legacy in the C* ecosystem, I don't associate to a clustering or partition-range from the word "interval".

It doesn't rename partition_range or clustering_range, just wrapping_range and nonwrapping_range. I guess we can keep partition_range and clustering_range.

> 
> > nonwrapping_range and related templates represent mathematical
> > intervals, and are different from C++ ranges. This causes confusion,
> > especially when C++ ranges and the range templates are used together.
> 
> Is this confusing to the compiler too or mainly to developers?

Developers.

interval.hh
```diff
@@ -0,0 +1,728 @@
+/*
+ * Copyright (C) 2015 ScyllaDB
```
denesb, line 2: 2020

range.hh
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015 ScyllaDB
+ * Copyright (C) 2020 ScyllaDB
```
denesb, line 2: Updated the wrong file? Or did you mean to update it. If the latter I'm curious as to why, I didn't see copyright year updated in existing source files before.

interval.hh
dekimir: This is an old file (renamed).

range.hh
dekimir: This is a new file, with backward-compatible name and content.

avikivity:
> > I'm not a big fan of this rename. The term "range" has a long and extensive legacy in the C* ecosystem, I don't associate to a clustering or partition-range from the word "interval".
> 
> It doesn't rename partition_range or clustering_range, just wrapping_range and nonwrapping_range. I guess we can keep partition_range and clustering_range.


Maybe we should rename them too. We do have ranges of partition_range or clustering_range, and seeing the same word used with two different meanings is confusing. This will get worse as we use more of std::ranges.

PR #6638, opened 2020-06-14
build: fix --enable-dpdk/--disable-dpdk configure switch

avikivity:
5ceb20c43992e2cb5292e7e7ad929c6798dfc560 switched --enable-dpdk
to a tristate switch, but forgot that add_tristate() prepends
--enable and --disable itself; so now the switch looks like
--enable-enable-dpdk and --disable-enable-dpdk.

Fix by removing the "enable-" prefix.


nyh:
Committed to next - 493d16e800929957b29d6ef626b55c6c84717c6e

PR #6644, opened 2020-06-15
dbuild: Add a flag to run dbuild using podman

yaronkaikov:
Following https://github.com/scylladb/scylla/pull/5333, we want to beable to run dbuild using podman or docker by setting enviorment variable named: DBUILD_TOOL

dbuild will check if podman is available to use and if not it will look for docker. 
when we  want to explicitly set the tool to one or the other (for example Jenkins pipeline), DBUILD_TOOL need to be set before

### DIFF
--- a/tools/toolchain/dbuild
+++ b/tools/toolchain/dbuild
@@ -4,6 +4,14 @@ if [[ -f ~/.config/scylladb/dbuild ]]; then
     . ~/.config/scylladb/dbuild
 fi
 
+if which podman >/dev/null 2>&1 ; then
+  tool=${DBUILD_TOOL-podman}
+elif which docker >/dev/null 2>&1 ; then
+  tool=${DBUILD_TOOL-docker}
+else
+  die "Please make sure you install either podman or docker on this machine to run dbuild"
+fi
+
 here="$(realpath $(dirname "$0"))"
 toplevel="$(realpath "$here/../..")"
 group_args=()
@@ -73,8 +81,6 @@ EOF
     exit 1
 }
 
-docker --version >/dev/null 2>&1 || die "Please install Docker on this machine to run dbuild"
-
 if [[ $# -eq 0 ]]; then
     interactive=y
     docker_args=(-it)
@@ -90,7 +96,7 @@ elif [[ "$1" = -* ]]; then
                 if [[ -z "$image" ]]; then
                     die "Expected docker image identifier to follow the --image option"
                 fi
-                if ! docker image inspect "$image" >/dev/null && ! docker image pull "$image"; then
+                if ! $tool image inspect "$image" >/dev/null && ! $tool image pull "$image"; then
                     die
                 fi
                 continue
@@ -127,7 +133,7 @@ MAVEN_LOCAL_REPO="$HOME/.m2"
 
 mkdir -p "$MAVEN_LOCAL_REPO"
 
-is_podman="$(docker --help | grep -o podman)"
+is_podman="$($tool --help | grep -o podman)"
 
 docker_common_args=()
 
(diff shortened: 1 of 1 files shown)


tools/toolchain/dbuild
```diff
@@ -90,11 +91,19 @@ elif [[ "$1" = -* ]]; then
                 if [[ -z "$image" ]]; then
                     die "Expected docker image identifier to follow the --image option"
                 fi
-                if ! docker image inspect "$image" >/dev/null && ! docker image pull "$image"; then
+                if ! $tool image inspect "$image" >/dev/null && ! $tool image pull "$image"; then
                     die
                 fi
                 continue
                 ;;
+            --tool)
```
penberg, line 99: Let's have @avikivity's feedback on the name of this command line option, but otherwise looks good to me.

avikivity:
I think it may be more useful to use an environment variable for this; since you don't want to specify it for every single invocation of dbuild.

yaronkaikov:
@avikivity there is no need to pass `--tool` for every dbuild. default will use `docker` for now. this will allow us to migrate our scylla-pkg code from `docker` to `podman` . once we move completely we can change the default to `podman` or even start using `podman-docker` package

penberg:
@yaronkaikov If you want to use `podman`, then you _do_ have to pass the flag all the time. And it's not like it's terribly hard to fix either in this patch:

```python3
tool = os.getenv('DBUILD_TOOL', default='docker')
```

yaronkaikov:
@avikivity ^^

tools/toolchain/dbuild
nyh: "tool" is a very generic name. Why don't we call it $docker, --docker, since that's what it is?

tools/toolchain/dbuild
```diff
@@ -90,11 +91,19 @@ elif [[ "$1" = -* ]]; then
...
+                tool="$2"
+                shift 2
+                if [[ -z "$tool" ]]; then
+                  die "Expected tool name to use, available options: docker or podman"
```
nyh, line 103: You say that the available options are just docker or podman, but this is not what the if tests. If someone does "dbuild --tool --xyz" this code will happy set tool="-xyz" :-)

nyh:
> I think it may be more useful to use an environment variable for this; since you don't want to specify it for every single invocation of dbuild.

I'm at a loss why you'd want to specify this option at all. We can check automatically which of the two exists on the system, and if both do and we believe podman (?) is better, we should use that. Why would a user care which is used?

tools/toolchain/dbuild
yaronkaikov: @nyh Removed this part, and the `--tool` flag

yaronkaikov:
> > I think it may be more useful to use an environment variable for this; since you don't want to specify it for every single invocation of dbuild.
> 
> I'm at a loss why you'd want to specify this option at all. We can check automatically which of the two exists on the system, and if both do and we believe podman (?) is better, we should use that. Why would a user care which is used?

Good idea, did it like that.

tools/toolchain/dbuild
yaronkaikov: maybe i will change it to `dbuild_tool` ?

penberg:
There are two orthogonal issues to decide on: (1) what is the default value and (2) how can it be configured.

I don't mind (1) to automagically decide between `podman` and `docker`, but we anyway need (2) because there are cases (for example, Jenkins pipelines) where you want to explicitly set the tool to one or the other. Allowing the script to pick one by themselves only leads to weird bugs where it picks the wrong tool by accident.

yaronkaikov:
@penberg  in Jenkins pipeline we can set the env before running those commands (as we do today)  . so we can still control which tool we are using and yet, not extra flag is required.

nyh:
> I don't mind (1) to automagically decide between `podman` and `docker`, but we anyway need (2) because there are cases (for example, Jenkins pipelines) where you want to explicitly set the tool to one or the other. Allowing the script to pick one by themselves only leads to weird bugs where it picks the wrong tool by accident.

Maybe I'm missing something, because I don't understand this line of reasoning. Why does Jenkins in particular need to control whether podman or docker is run? If one of the two doesn't work, don't install it on the Jenkins machine! If both work, then the script can pick one in a consistent manner (e.g., always pick "podman" if both podman and docker exist). I don't understand what sort of "weird bugs" you're trying to avoid.

penberg:
Queued as ac7237f99116698cd8ff75444b1908f7b4517b6b

PR #6651, opened 2020-06-15
reloc: support unified relocatable package

### LINKED ISSUE #6626: Unified offline installer script
We are going to generate a single, unified tarball for the offline installer. Let's also implement a unified installer script, which calls the different component (core, jmx, tools) installers to simplify offline installation process for the user.

syuu1228:
This introduce unified relocatable package, a single tarball to install
all Scylla packages.

Fixes #6626
See scylladb/scylla-pkg#1218

### DIFF
--- a/unified/build_unified.sh
+++ b/unified/build_unified.sh
@@ -1,0 +1,69 @@
+#!/bin/bash -e
+#
+# Copyright (C) 2020 ScyllaDB
+#
+
+#
+# This file is part of Scylla.
+#
+# Scylla is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Scylla is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+print_usage() {
+    echo "build_unified.sh --mode <mode>"
+    echo "  --mode specify mode (default: release)"
+    echo "  --unified-pkg specify package path (default: build/release/scylla-unified-package.tar.gz)"
+    exit 1
+}
+
+MODE="release"
+UNIFIED_PKG="build/release/scylla-unified-package.tar.gz"
+while [ $# -gt 0 ]; do
+    case "$1" in
+        "--mode")
+            MODE="$2"
+            shift 2
+            ;;
+        "--unified-pkg")
+            UNIFIED_PKG="$2"
+            shift 2
+            ;;
+        *)
+            print_usage
+            ;;
+    esac
+done
+
+UNIFIED_PKG="$(realpath -s $UNIFIED_PKG)"
+PKGS="build/$MODE/scylla-package.tar.gz build/$MODE/scylla-python3-package.tar.gz tools/jmx/build/scylla-jmx-package.tar.gz tools/java/build/scylla-tools-package.tar.gz"
+
+rm -rf build/"$MODE"/unified/
+mkdir -p build/"$MODE"/unified/
+for pkg in $PKGS; do
+    if [ ! -e "$pkg" ]; then
+        echo "$pkg not found."
+        echo "please build relocatable package before building unified package."
+        exit 1
+    fi
+    pkg="$(readlink -f $pkg)"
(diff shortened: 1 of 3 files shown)


syuu1228:
Note about build_unified.sh:
I tried "zcat *.tar.gz | gzip -c > scylla-unified.tar.gz" @penberg mentioned at https://github.com/scylladb/scylla-pkg/issues/1218 in the beginning, but combined tarball does not able to extract without --ignore-zeros option, that seems easy to mistake when user extracting it (if user extract the tarball without --ignore-zeros, they will get only first part of the archive)
So I suppose it's better to extract everything and archive them again.

Referenced: https://linux.die.net/man/1/tar

unified/install.sh
```diff
@@ -0,0 +1,104 @@
...
+
+if $housekeeping; then
+    arg_housekeeping=--housekeeping
+fi
+if $nonroot; then
+    arg_nonroot=--nonroot
+fi
+
+(cd scylla; ./install.sh --root $root --prefix $prefix --python3 $python3 $arg_housekeeping $arg_nonroot --sysconfdir $sysconfdir)
+
```
avikivity, line 99: Please canonicalize paths, because "cd scylla" makes relative paths wrong.

Use correct shell quoting.

If there is an error in ./install.sh, does it stop the rest?

avikivity:
Please add ninja rules so it is automatically built along with the other packages.

unified/install.sh
syuu1228: @avikivity 

> Please canonicalize paths, because "cd scylla" makes relative paths wrong.

Is it something like `cd "$(pwd)/scylla"` instead of `cd scylla`?
 
> If there is an error in ./install.sh, does it stop the rest?

It should, since `set -e`

unified/build_unified.sh
```diff
@@ -0,0 +1,40 @@
...
+    exit 1
+fi
+
+rm -rf build/unified
+mkdir -p build/unified
+for pkg in "$@"; do
+    tar -C build/unified -xpf "$pkg"
+done
+for pkg in scylla scylla-python3 scylla-jmx scylla-tools; do
+    if [ ! -d build/unified/$pkg ]; then
```
avikivity, line 34: What about separate modes?

We'll want to make this part of ninja, so we can write ninja build/dev/scylla-unified-package.tar.gz, so different modes can't share the same temporary directory.

unified/install.sh
```diff
@@ -0,0 +1,104 @@
...
+fi
+
+if $housekeeping; then
+    arg_housekeeping=--housekeeping
+fi
+if $nonroot; then
+    arg_nonroot=--nonroot
+fi
+
+(cd scylla; ./install.sh --root "$root" --prefix "$prefix" --python3 "$python3" "$arg_housekeeping" "$arg_nonroot" --sysconfdir "$sysconfdir")
```
avikivity, line 98: "$arg_housekeeping" translates to an empty argument instead of nothing.

I think it's nicer to use arrays:

```sh
args=()

if $housekeeping; then
    args+=(--housekeeping)
fi

if $nonroot; then
    args+=(--nonroot)
fi
```

syuu1228:
@avikivity rebased with latest next (after moving jmx/tools submodules to tools/) and also add support ninja-build, please review again

unified/install.sh
syuu1228: @avikivity but we only need --housekeeping for scylla main package, but --nonroot needed for all package.

avikivity: Does it work when $arg_housekeeping or $arg_nonroot is empty? I think it generates an empty parameter that will confuse install.sh.

unified/install.sh
avikivity: > @avikivity
> 
> > Please canonicalize paths, because "cd scylla" makes relative paths wrong.
> 
> Is it something like `cd "$(pwd)/scylla"` instead of `cd scylla`?
> 

No, the parameters like --root. If someone says --root ../foo we should convert it to a full path (readlink -f or similar).

penberg:
@syuu1228 Please rebase the pull request to at least remove the following commit, which is already in `master`:

https://github.com/scylladb/scylla/pull/6651/commits/9aca91cd91074b0d9668c15a77d7815498ccdd4e

Also, is the Python submodule movement supposed to be in this pull request?

syuu1228:
@penberg there were unrelated commits on the branch, rebased.
@avikivity canonicalize paths and args handling issue on install.sh fixed, please review again.

unified/build_unified.sh
syuu1228: @avikivity the issue on this PR has been fixed, but I think relocatable / .rpm / .deb package build directories also shared between different modes, I suppose we need to fix them too since we support to build it on ninja, correct?

PR #6660, opened 2020-06-17
scylla_swap_setup: handle <1GB environment

### LINKED ISSUE #6659: scylla_swap_setup: swapsize become 0 when memory size less than 1GB
swapsize become 0GB when memory size less than 1GB, and causes the error:
```
subprocess.CalledProcessError: Command '['fallocate', '-v', '-l', '0GB', '/swapfile']' returned non-zero exit status 1
```

Scylla does not work such low memory environment anyway, we probably should raise error with better message.

syuu1228:
Show better error message and exit with non-zero status when memory size <1GB.

Fixes #6659


syuu1228:
@avikivity ping

PR #6664, opened 2020-06-17
cdc: Fix segfault when selecting stream ID by a too short key

### LINKED ISSUE #6570: Scylla crashes when CDC Log is queried with a wrong stream_id
Executing:
`select * from tb_scylla_cdc_log where "cdc$stream_id" = 0x00;`

causes: 

```
scylla: cdc/generation.cc:84: int64_t cdc::bytes_to_int64(bytes_view, size_t): Assertion `b.size() >= offset + sizeof(int64_t)' failed.
```

jul-stas:
When a token is calculated for stream_id, we check that the key is exactly 16 bytes long. If it's not - `minimum_token` is returned and client receives empty result.

This used to be the expected behavior for empty keys; now it's extended to keys of any incorrect length.

Fixes #6570

### DIFF
--- a/cdc/cdc_partitioner.cc
+++ b/cdc/cdc_partitioner.cc
@@ -40,7 +40,8 @@ static dht::token to_token(int64_t value) {
 }
 
 static dht::token to_token(bytes_view key) {
-    if (key.empty()) {
+    // Key should be 16 B long, of which first 8 B are used for token calculation
+    if (key.size() != 2*sizeof(int64_t)) {
         return dht::minimum_token();
     }
     return to_token(stream_id::token_from_bytes(key));
--- a/test/cql/cdc_too_short_stream_id_test.cql
+++ b/test/cql/cdc_too_short_stream_id_test.cql
@@ -1,0 +1,7 @@
+create table tb (pk int primary key) with cdc = {'enabled': true};
+insert into tb (pk) VALUES (0);
+
+-- Key of length != 128 b should return empty result set (issue #6570)
+select * from tb_scylla_cdc_log where "cdc$stream_id" = 0x00;
+
+select * from tb_scylla_cdc_log where "cdc$stream_id" = 0x;
--- a/test/cql/cdc_too_short_stream_id_test.result
+++ b/test/cql/cdc_too_short_stream_id_test.result
@@ -1,0 +1,19 @@
+create table tb (pk int primary key) with cdc = {'enabled': true};
+{
+	"status" : "ok"
+}
+insert into tb (pk) VALUES (0);
+{
+	"status" : "ok"
+}
+
+-- Key of length != 128 b should return empty result set (issue #6570)
+select * from tb_scylla_cdc_log where "cdc$stream_id" = 0x00;
+{
+	"rows" : null
+}
+
+select * from tb_scylla_cdc_log where "cdc$stream_id" = 0x;
+{
+	"rows" : null
+}


cdc/cdc_partitioner.cc
```diff
@@ -40,7 +40,7 @@ static dht::token to_token(int64_t value) {
 }
 
 static dht::token to_token(bytes_view key) {
-    if (key.empty()) {
+    if (key.size() < sizeof(int64_t)) {
```
haaawk, line 43: I think it should be key.size() != sizeof(int64_t)

jul-stas: @haaawk  If I do `key.size() != sizeof(int64_t)` here, then I won't be able to query by `stream_id` at all. That's because the key should be, in fact, **16** bytes long, even tho only first 8 bytes are needed for token calculation.

So I made this function have "minimum expectations" - it does not assert the absolute correctness of the key; it just performs the necessary check to to its job.

haaawk: ok then != 2 * sizeof(int64_t)

jul-stas: I can agree with that.
Done.

test/cql/cdc_too_short_stream_id_test.cql
```diff
@@ -0,0 +1,6 @@
+create table tb (pk int primary key) with cdc = {'enabled': true};
+
+-- Key of length != 128 b should return empty result set (issue #6570)
+select * from ks.tb_scylla_cdc_log where "cdc$stream_id" = 0x00;
+
+select * from ks.tb_scylla_cdc_log where "cdc$stream_id" = 0x;
```
piodul, line 6: `tb_scylla_cdc_log` is empty, so it is not surprising that the result is empty... While I am aware that the fix is against a segfault occuring here, I think that we should probably make the cdc log non-empty so that this test guards against bugs which could cause a non-empty result set to be returned in this case.

haaawk: Good point!

jul-stas:
> Looks good to me. Should I wait for an improved test before committing?

Thanks, I already improved the test.

test/cql/cdc_too_short_stream_id_test.cql
jul-stas: done!

nyh:
Committed to next - 8628ede009b339205e26b680c4e95c601fa38a6a

test/cql/cdc_too_short_stream_id_test.cql
```diff
@@ -0,0 +1,7 @@
+create table tb (pk int primary key) with cdc = {'enabled': true};
+insert into tb (pk) VALUES (0);
+
+-- Key of length != 128 b should return empty result set (issue #6570)
+select * from tb_scylla_cdc_log where "cdc$stream_id" = 0x00;
+
+select * from tb_scylla_cdc_log where "cdc$stream_id" = 0x;
```
nyh, line 7: @kostja this is an interesting example of the limitations of the pure-CQL tests without any programming language.
hey insert one item into the table, and will get one entry in the CDC log. But they don't know which key this entry will have - it's a random key. Even if we can seed it the same and get the same on every run, it will be very annoying if we change this random function in the future and random tests start to fail.
So all this test can test is the empty select case (where a bad key doesn't match anything). They can't check a successful case because 1. they don't know what "where" to use, and 2. if they knew, the result will contain the random key and can change.

PR #6668, opened 2020-06-18
dbuild: Add an option to run dbuild using podman

yaronkaikov:
Following https://github.com/scylladb/scylla/pull/5333, we want to be able to run dbuild using podman or docker by setting environment variable
named: DBUILD_TOOL

DBUILD_TOOL will use docker by default unless we explicitly set the tool podman

Fixes: https://github.com/scylladb/scylla/pull/6644

### DIFF
--- a/tools/toolchain/dbuild
+++ b/tools/toolchain/dbuild
@@ -4,6 +4,14 @@ if [[ -f ~/.config/scylladb/dbuild ]]; then
     . ~/.config/scylladb/dbuild
 fi
 
+if which docker >/dev/null 2>&1 ; then
+  tool=${DBUILD_TOOL-docker}
+elif which podman >/dev/null 2>&1 ; then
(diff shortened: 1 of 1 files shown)


yaronkaikov:
@nyh The only different is that default is still docker. So now we will not break anything, and can migrate to use podman without breaking the all process

nyh:
I don't know what the default should be, I think *either* would have been fine (if a user installed both, it means both should be working).
 So I'm ok with this patch. Again, because I'm not familiar with the Jenkins problems you're trying to solve, I'll let @penberg be the final judge and give him the honor to commit the patch.

penberg:
@nyh Yes, so my preference was to make sure `docker` is the default if both of them are installed.

penberg:
Queued as e9d5852b0c57efde7ceb242be2aeb50343bf2779

PR #6670, opened 2020-06-18
cdc: Handling of timeout/unavailable exceptions in streams fetching

### LINKED ISSUE #6557: cdc: report unexpected exceptions using "error" level in generation management code
When a node sees in gossip that a new CDC generation has appeared, it will try to retrieve the generation data from distributed system tables (`system_distributed.cdc_topology_description`).

The attempt may fail due to e.g. a network partition. It would then report something like UnavailableException or Timeout in the log using the WARN level, and schedule a background process to retry the retrieval. The idea is that eventually, after the partition resolves, the node will retrieve the data.

The relevant functions are: `storage_service::handle_cdc_generation` and `storage_service::async_handle_cdc_generation`.

This behavior might hide bugs, unfortunately. For example, if due to a bug the `cdc_topology_description` table is corrupted, no node will ever be able to retrieve the data, and the exception could be something like `read_simple_exactly - size mismatch`. See, e.g., https://github.com/scylladb/scylla/issues/6537.

For this class of unexpected errors, we should:
- report the error using the ERROR level (instead of WARN), so that these problems will get noticed in tests and tests will fail,
- NOT schedule a retry, but finish the retrieval procedure.

jul-stas:
Retrying the operation of fetching generations makes sense only when the underlying problem is "temporary". In this patch only the `timeout` and `unavailable` exceptions trigger retrying, while all the other ones (presumably: unrecoverable ones) abort the operation and get logged on ERROR level.

Fixes #6557

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -682,6 +682,31 @@ bool storage_service::do_handle_cdc_generation(db_clock::time_point ts) {
     }).get0();
 }
 
+namespace {
+class cdc_generation_handling_nonfatal_exception : public std::runtime_error {
+    using std::runtime_error::runtime_error;
+};
+
+constexpr char could_not_retrieve_msg_template[]
+        = "Could not retrieve CDC streams with timestamp {} upon gossip event. Reason: \"{}\". Action: {}.";
+} // anon. namespace
+
+bool storage_service::do_handle_cdc_generation_intercept_nonfatal_errors(db_clock::time_point ts) {
+    try {
+        return do_handle_cdc_generation(ts);
+    } catch (exceptions::request_timeout_exception& e) {
+        throw cdc_generation_handling_nonfatal_exception(e.what());
+    } catch (exceptions::unavailable_exception& e) {
+        throw cdc_generation_handling_nonfatal_exception(e.what());
+    } catch (...) {
+        const auto ep = std::current_exception();
+        if (is_timeout_exception(ep)) {
+            throw cdc_generation_handling_nonfatal_exception(format("{}", ep));
+        }
+        throw;
+    }
+}
+
 class ander {
 private:
     bool _result = true;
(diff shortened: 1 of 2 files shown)


service/storage_service.cc
```diff
@@ -682,6 +682,31 @@ bool storage_service::do_handle_cdc_generation(db_clock::time_point ts) {
     }).get0();
 }
 
+namespace {
+class cdc_generation_handling_nonfatal_exception : public std::runtime_error {
+    using std::runtime_error::runtime_error;
+};
+
+constexpr char could_not_retrieve_msg_template[]
+        = "Could not retrieve CDC streams with timestamp {} upon gossip event. Reason: \"{}\". Action: {}.";
```
haaawk, line 691: Wouldn't string_view be better? I'm not an expert but my impression was that it's a recommended way to use string literals.

jul-stas: Tried googling for arguments pro/con and found nothing conclusive. I can switch to the `""sv` literal for the sole sake of writing "modern C++" :)

haaawk: I meant sth like this https://github.com/scylladb/scylla/blob/9d9d54c804dd400d90edfdb84c8f5319d06b1633/gms/feature_service.cc#L30

jul-stas: Tried it - apparently our `fmt::format` is not ready for `string_view`:
```
service/storage_service.cc:779:23: note:   cannot convert ‘service::{anonymous}::could_not_retrieve_msg_template’ (type ‘const string_view’ {aka ‘const std::basic_string_view<char>’}) to type ‘const char*’
  779 |         cdc_log.error(could_not_retrieve_msg_template, ts, std::current_exception(), "not retrying");
```

haaawk: Ok. Let's leave that then.

kbr-:
No, wait, I just noticed something.

service/storage_service.cc
```diff
@@ -745,12 +771,13 @@ void storage_service::handle_cdc_generation(std::optional<db_clock::time_point>
...
     bool using_this_gen = false;
     try {
-        using_this_gen = do_handle_cdc_generation(*ts);
-    } catch(...) {
-        cdc_log.warn("Could not retrieve CDC streams with timestamp {} on gossip event: {}."
-                " Will retry in the background.", *ts, std::current_exception());
+        using_this_gen = do_handle_cdc_generation_intercept_nonfatal_errors(*ts);
+    } catch (cdc_generation_handling_nonfatal_exception& e) {
+        cdc_log.warn(could_not_retrieve_msg_template, ts, e.what(), "retrying in the background");
         async_handle_cdc_generation(*ts);
```
kbr-, line 777: Nit: no need to remove this `return`, but it doesn't matter I guess.

jul-stas: done

jul-stas:
Thanks for the reviews, I'm calling @nyh now.

nyh:
Committed to next: 6b4b00275776c02d6fd1d9e87dcccd31d09d9e32

PR #6671, opened 2020-06-18
thrift: switch csharp backend to netstd

avikivity:
The thrift compiler (since 0.13 at least) complains that
the csharp target is deprecated and recommend replacing it
with netstd. Since we don't use either, humor it.

I suspect that this warning caused some spurious rebuilds,
but have not proven it.


avikivity:
@scylladb/scylla-maint ping

avikivity:
@tgrabiec please check your settings, the commit shows as if noreply@github.com merged it.

PR #6677, opened 2020-06-19
scylla_setup: follow hugepages package name change on Ubuntu 20.04LTS

### LINKED ISSUE #6673: scylla_setup: "apt-get install -y hugepages" fails on Ubuntu 20.04LTS
Seems like "hugepages" package does not exists on Ubuntu 20.04LTS.
See: https://packages.ubuntu.com/search?suite=focal&searchon=names&keywords=hugepages
Related: #6343
```
ScyllaDB setup finished.
scylla_setup accepts command line arguments as well! For easily provisioning in a similar environment than this, type:

    scylla_setup --no-raid-setup

Also, to avoid the time-consuming I/O tuning you can add --no-io-setup and copy the contents of /etc/scylla.d/io*
Only do that if you are moving the files into machines with the exact same hardware
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package hugepages
  File "/opt/scylladb/scripts/libexec/scylla_setup", line 472, in <module>
    run('apt-get install -y hugepages')
  File "/opt/scylladb/scripts/scylla_util.py", line 342, in run
    return subprocess.check_call(cmd, shell=shell, stdout=stdout, stderr=stderr, env=scylla_env)
  File "/opt/scylladb/python3/lib64/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['apt-get', 'install', '-y', 'hugepages']' returned non-zero exit status 100.

```

syuu1228:
hugepages package now renamed to libhugetlbfs-bin, we need to follow
the change.

Fixes #6673

### DIFF
--- a/dist/common/scripts/scylla_setup
+++ b/dist/common/scripts/scylla_setup
@@ -27,6 +27,7 @@ import glob
 import shutil
 import io
 import stat
+import distro
 from scylla_util import *
 
 interactive = False
@@ -468,5 +469,10 @@ if __name__ == '__main__':
             print('Please restart your machine before using ScyllaDB, as you have disabled')
             print(' SELinux.')
 
-        if dist_name() == 'Ubuntu':
-            run('apt-get install -y hugepages')
+        if distro.id() == 'ubuntu':
+            # Ubuntu version is 20.04 or later
+            if int(distro.major_version()) >= 20:
+                hugepkg = 'libhugetlbfs-bin'
+            else:
+                hugepkg = 'hugepages'
+            run(f'apt-get install -y {hugepkg}')


dist/common/scripts/scylla_setup
```diff
@@ -469,4 +469,8 @@ if __name__ == '__main__':
             print(' SELinux.')
 
         if dist_name() == 'Ubuntu':
-            run('apt-get install -y hugepages')
+            if ubuntu_version() == '20.04':
```
avikivity, line 472: Let's check for later release too, so we don't hit the same problem in April 2022.

syuu1228: @avikivity supported later version.

syuu1228:
@avikivity ping, the version check for later version of Ubuntu pushed, please review agiain

PR #6678, opened 2020-06-19
reloc: simplified .deb build process

syuu1228:
This is v2 version of #6550, rebased with latest next & tested on dbuild.

----

We don't really need to have two build_deb.sh, merge it to reloc.


avikivity:
Queued as 76332020bf

PR #6680, opened 2020-06-19
reloc: simplified .rpm build process

syuu1228:
This is v2 version of #6549, rebased with latest next & tested on dbuild.

----

We don't really need to have two build_rpm.sh, merge it to reloc.


avikivity:
Queued as 7c9122fb2cbd4995b63d114704dc1070c0970cd3

PR #6701, opened 2020-06-23
scylla_util.py: switch to subprocess.run()

syuu1228:
When we started to porting bash script to python script, we are not able to use
subprocess.run() since EPEL only provides python 3.4, but now we have
relocatable python, so we can switch to it.

### DIFF
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -338,21 +338,13 @@ def run(cmd, shell=False, silent=False, exception=True):
     stderr = subprocess.DEVNULL if silent else None
     if not shell:
         cmd = shlex.split(cmd)
-    if exception:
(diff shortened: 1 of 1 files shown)


syuu1228:
Note: scylla_util.py's run() / out() was implemented because python 3.4 does not have subprocess.run(), so we might able to drop it and switch to subprocess.run() now.
But it requires various parameters to run unlike our run() / out(), we at least requires shell, env options for every command.
(It also generating Array for command line when shell=False mode, but it can avoid when we switch to shell=True)
Like this:
```
run('scylla_setup')
res = out('ls')
```
becomes
```
from subprocess import run

run('scylla_setup', shell=True, env=scylla_env)
res = run('ls', capture_output=True, shell=True, env=scylla_env, encoding='utf-8').stdout.strip()
```
I suppose it become too complex from previous one, so I guess it's better to keep run() / out().

syuu1228:
@avikivity ping

PR #6704, opened 2020-06-23
scylla_util.py: stop using /etc/os-release, use distro

### LINKED ISSUE #6691: scylla_util.py: duplication on detecting distribution
Seems like we mistakenly made two different way to detect distribution, 
 1. is_redhat_variant() / is_debian_variant() uses /etc/os-release, also we have a variable 'os_release' to get ID, ID_LIKE, VERSION, VERSION_ID etc.
 2. dist_name() / dist_ver() were previously used platform.dist(), now switched to distro package.

We can keep these functions, but no need to keep using both interface, os-release and distro.
I suppose distro is better choice, since it provides better abstraction.

syuu1228:
Currently we we mistakenly made two different way to detect distribution,
directly reading /etc/os-release and use distro package.

distro package provides well abstracted APIs and still have full access to
os-release informations, we should switch to it.

Fixes #6691

### DIFF
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -355,30 +355,28 @@ def out(cmd, shell=False, exception=True, timeout=None):
         return p.communicate()[0].strip().decode('utf-8')
 
 
-def parse_os_release_line(line):
-    id, data = line.split('=', 1)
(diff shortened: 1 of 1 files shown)


syuu1228:
@avikivity ping

PR #6705, opened 2020-06-24
scylla_ntp_setup: switch to distro package

syuu1228:
Use distro API to simplify distribution detection.


syuu1228:
@avikivity ping

PR #6706, opened 2020-06-24
scylla_raid_setup: switch to systemd mount unit

syuu1228:
Since we already use systemd unit file for coredump bind mount and swapfile,
we should move to systemd mount unit for data partition as well.

### DIFF
--- a/dist/common/scripts/scylla_raid_setup
+++ b/dist/common/scripts/scylla_raid_setup
@@ -37,8 +37,8 @@ if __name__ == '__main__':
                         help='specify disks for RAID')
     parser.add_argument('--raiddev', default='/dev/md0',
                         help='MD device name for RAID')
-    parser.add_argument('--update-fstab', action='store_true', default=False,
-                        help='update /etc/fstab for RAID')
+    parser.add_argument('--enable-on-nextboot', '--update-fstab', action='store_true', default=False,
+                        help='mount RAID on next boot')
     parser.add_argument('--root', default='/var/lib/scylla',
                         help='specify the root of the tree')
     parser.add_argument('--volume-role', default='all',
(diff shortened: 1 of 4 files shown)


syuu1228:
This deletes --update-fstab since we don't rely on fstab anymore, we need update a script on scylla-machine-image after merging this.
We call it here: https://github.com/scylladb/scylla-machine-image/blob/master/aws/scylla_create_devices#L86

syuu1228:
@avikivity ping

dist/common/scripts/scylla_raid_setup
```diff
@@ -37,14 +37,14 @@ if __name__ == '__main__':
...
-    parser.add_argument('--update-fstab', action='store_true', default=False,
-                        help='update /etc/fstab for RAID')
     parser.add_argument('--root', default='/var/lib/scylla',
                         help='specify the root of the tree')
     parser.add_argument('--volume-role', default='all',
                         help='specify how will this device be used (data, commitlog, or all)')
     parser.add_argument('--force-raid', action='store_true', default=False,
                         help='force constructing RAID when only one disk is specified')
+    parser.add_argument('--disable-on-nextboot', action='store_true', default=False,
+                        help='do not mount RAID on next boot')
```
avikivity, line 48: Let's make it `--enable-on-nextboot` (default true) to avoid double negatives. And make `--update-fstab` an alias to avoid breaking scripts.

dist/common/scripts/scylla_raid_setup
```diff
@@ -131,7 +131,47 @@ if __name__ == '__main__':
...
+        print('mount unit {} already exists'.format(mntunit))
+        sys.exit(1)
+
+    res = out('blkid {}'.format(fsdev))
+    match = re.search(r'^/dev/\S+: (UUID="\S+")', res.strip())
+    uuid = match.group(1)
+    unit_data = f'''
+[Unit]
+Description=Scylla data directory
+Conflicts=umount.target
```
avikivity, line 146: The manual says:

       ·   All mount units acquire automatic Before= and Conflicts= on umount.target in order to be stopped during shutdown.

So this looks unneeded.

dist/common/scripts/scylla_raid_setup
```diff
@@ -131,7 +131,47 @@ if __name__ == '__main__':
...
     makedirs(mount_at)
-    run('mount -t xfs -o noatime {raid} "{mount_at}"'.format(raid=fsdev, mount_at=mount_at))
+    mntunit_bn = out('systemd-escape -p --suffix=mount {}'.format(mount_at))
+    mntunit = Path('/etc/systemd/system/{}'.format(mntunit_bn))
+    if mntunit.exists():
+        print('mount unit {} already exists'.format(mntunit))
+        sys.exit(1)
+
+    res = out('blkid {}'.format(fsdev))
+    match = re.search(r'^/dev/\S+: (UUID="\S+")', res.strip())
```
avikivity, line 141: You can use `blkid  --match-tag PTUUID -o value` to get blkid to print just the uuid.

syuu1228:
@avikivity pushed three changes based on your suggestion, please review again.

dist/common/scripts/scylla_raid_setup
```diff
@@ -37,8 +37,8 @@ if __name__ == '__main__':
                         help='specify disks for RAID')
     parser.add_argument('--raiddev', default='/dev/md0',
                         help='MD device name for RAID')
-    parser.add_argument('--update-fstab', action='store_true', default=False,
-                        help='update /etc/fstab for RAID')
+    parser.add_argument('--enable-on-nextboot', action='store_true', default=False,
+                        help='mount RAID on next boot')
     parser.add_argument('--root', default='/var/lib/scylla',
```
avikivity, line 42: Please keep --update-fstab for compatibility. add_argument supports ['--enable-on-nextboot', '--update-fstab'].

avikivity: Actually without the brackets.

syuu1228:
@avikivity Restored --update-fstab, please review again.

PR #6710, opened 2020-06-25
scylla_coredump_setup: detect missing coredump file

syuu1228:
Print error message and exit with non-zero status by following condition:
    - coredumpctl says the coredump file is inaccessible
    - failed to detect coredump file path from 'coredumpctl info <pid>'
    - deleting coredump file failed because the file is missing

Fixes #6654


syuu1228:
@avikivity ping

PR #6735, opened 2020-06-30
dist/docker/redhat/Dockerfile: update 4.2 params

yaronkaikov:
Set SCYLLA_REPO and VERSION values for scylla-4.2


avikivity:
Let's get the parameters from SCYLLA-VERSION-GEN instead of this separate file.

PR #6740, opened 2020-07-01
repair: Handle dropped table in repair_range

asias:
In commit 12d929a5ae1f3166ec56e54239f40e6137f6d2c2 (repair: Add table_id
to row_level_repair), a call to find_column_family() was added in
repair_range. In case of the table is dropped, it will fail the
repair_range which in turn fails the bootstrap operation.

Tests: update_cluster_layout_tests.py:TestUpdateClusterLayout.simple_add_new_node_while_schema_changes_test
Fixes: #5942

### DIFF
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -1050,7 +1050,12 @@ static future<> repair_range(repair_info& ri, const dht::token_range& range) {
       }
       return ::service::get_local_migration_manager().sync_schema(ri.db.local(), neighbors).then([&neighbors, &ri, range, id] {
         return do_for_each(ri.table_ids.begin(), ri.table_ids.end(), [&ri, &neighbors, range] (utils::UUID table_id) {
-            auto cf = ri.db.local().find_column_family(table_id).schema()->cf_name();
+            sstring cf;
+            try {
+                cf = ri.db.local().find_column_family(table_id).schema()->cf_name();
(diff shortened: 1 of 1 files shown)


repair/repair.cc
```diff
@@ -1050,7 +1050,12 @@ static future<> repair_range(repair_info& ri, const dht::token_range& range) {
       }
       return ::service::get_local_migration_manager().sync_schema(ri.db.local(), neighbors).then([&neighbors, &ri, range, id] {
         return do_for_each(ri.table_ids.begin(), ri.table_ids.end(), [&ri, &neighbors, range] (utils::UUID table_id) {
-            auto cf = ri.db.local().find_column_family(table_id).schema()->cf_name();
+            sstring cf;
+            try {
+                cf = ri.db.local().find_column_family(table_id).schema()->cf_name();
```
avikivity, line 1055: `cf` is not used, why keep it?

asias: It is used in logging. A table name is more readable than a uuid. It avoids looking up it from a table_id again and again.

avikivity: Sorry, I thought the patch introduced cf and did not use it, but it existed before.

asias: Ah, I see ;-)

Issue #6598, opened 2020-06-05
CDC: within a batch, clearing collection in one clustering row can cause wrong pre/postimage in another row


piodul:
Consider this example:

```cql
cqlsh:ks> create table tbl (pk int, ck int, c list<int>, primary key(pk, ck)) with cdc = {'enabled': true, 'preimage': true, 'postimage': true};
cqlsh:ks> update tbl set c = [1, 2, 3] where pk = 0 and ck = 1;
cqlsh:ks> begin batch update tbl set c = [] where pk = 0 and ck = 0; update tbl set c = c + [4] where pk = 0 and ck = 1; apply batch;
cqlsh:ks> select * from tbl_scylla_cdc_log;

 cdc$stream_id                      | cdc$time                             | cdc$batch_seq_no | c                                                                                                                           | cdc$deleted_c | cdc$deleted_elements_c | cdc$operation | cdc$ttl | ck | pk
------------------------------------+--------------------------------------+------------------+-----------------------------------------------------------------------------------------------------------------------------+---------------+------------------------+---------------+---------+----+----
 0xcbe95a3fdaa42af8fe59a473cd5a34cd | 861a68d2-a72c-11ea-1f8b-dde731a4bd55 |                0 |                                                                                                                        null |          True |                   null |             1 |    null |  1 |  0
 0xcbe95a3fdaa42af8fe59a473cd5a34cd | 861a68d2-a72c-11ea-1f8b-dde731a4bd55 |                1 |                                                                                                                        null |          null |                   null |             9 |    null |  1 |  0
 0xcbe95a3fdaa42af8fe59a473cd5a34cd | 861a68dc-a72c-11ea-a6ec-8ea5ab2512b7 |                0 | {861a6da0-a72c-11ea-8224-000000000000: 1, 861a6da1-a72c-11ea-8224-000000000000: 2, 861a6da2-a72c-11ea-8224-000000000000: 3} |          null |                   null |             1 |    null |  1 |  0
 0xcbe95a3fdaa42af8fe59a473cd5a34cd | 861a68dc-a72c-11ea-a6ec-8ea5ab2512b7 |                1 | {861a6da0-a72c-11ea-8224-000000000000: 1, 861a6da1-a72c-11ea-8224-000000000000: 2, 861a6da2-a72c-11ea-8224-000000000000: 3} |          null |                   null |             9 |    null |  1 |  0
[...]

PR #6741, opened 2020-07-01
cdc: better pre/postimages for complicated batches

### LINKED ISSUE #6597: CDC: collection updates which are split have incorrect pre/postimages
Consider the following example:

```cql
cqlsh:ks> create table tbl (pk int PRIMARY KEY, c list<int>) with cdc = {'enabled': true, 'preimage': true, 'postimage': true};
cqlsh:ks> begin unlogged batch update tbl using ttl 10000 set c = c + [1] where pk = 0; update tbl using ttl 20000 set c = c + [2] where pk = 0; apply batch;
cqlsh:ks> select * from tbl_scylla_cdc_log;

 cdc$stream_id                      | cdc$time                             | cdc$batch_seq_no | c                                         | cdc$deleted_c | cdc$deleted_elements_c | cdc$operation | cdc$ttl | pk
------------------------------------+--------------------------------------+------------------+-------------------------------------------+---------------+------------------------+---------------+---------+----
 0xcf452e9350d6490d58c082071c5c3082 | d0428b0a-a728-11ea-5760-719318441668 |                0 | {d042a360-a728-11ea-a8e8-000000000000: 1} |          null |                   null |             1 |   10000 |  0
 0xcf452e9350d6490d58c082071c5c3082 | d0428b0a-a728-11ea-5760-719318441668 |                1 | {d042a360-a728-11ea-a8e8-000000000000: 1} |          null |                   null |             9 |    null |  0
 0xcf452e9350d6490d58c082071c5c3082 | d0428b0a-a728-11ea-5760-719318441668 |                2 | {d042a361-a728-11ea-a8e8-000000000000: 2} |          null |                   null |             1 |   20000 |  0
 0xcf452e9350d6490d58c082071c5c3082 | d0428b0a-a728-11ea-5760-719318441668 |                3 | {d042a361-a728-11ea-a8e8-000000000000: 2} |          null |                   null |             9 |    null |  0

(4 rows)
cqlsh:ks> select * from tbl;

 pk | c
----+--------
  0 | [1, 2]

(1 rows)
```

The batch contains two UPDATEs, each of them appending a cell to the same row. Appended cells have different TTLs, therefore we create separate log entries for them.

The first two rows of the `tbl_scylla_cdc_log` correspond to cell 1 with TTL 10000. The table was empty before the first update, so there is no preimage. There is postimage, with correct value.

However, information about cell 2 with TTL 20000 is wrong - there is no preimage, and postimage is wrong (it contains only cell 2, but not 1).

piodul:
This PR changes the algorithm used to generate preimages and postimages
in CDC log. While its behavior is the same for non-batch operations
(with one exception described later), it generates pre/postimages that
are organized more nicely, and account for multiple updates to the same
row in one CQL batch.

Fixes #6597, #6598

Tests:
	- unit(dev), for each consecutive commit
	- unit(debug), for the last commit

----

__Previous method__

The previous method worked on a per delta row basis. First, the base
table is queried for the current state of the rows being modified in
the processed mutation (this is called the "preimage query"). Then,
for each delta row (representing a modification of a row):

- If preimage is enabled and the row was already present in the table,
  a corresponding preimage row is inserted before the delta row.
  The preimage row contains data taken directly from the preimage
  query result. Only columns that are modified by the delta are
  included in the preimage.
- If postimage is enabled, then a postimage row is inserted after the
  delta row. The postimage row contains data which was a result of
  taking row data directly from the preimage query result and applying
  the change the corresponding delta row represented. All columns
  of the row are included in the postimage.

The above works well for simple cases such like singular CQL INSERT,
UPDATE, DELETE, or simple CQL BATCH-es. An example:

```cql
cqlsh:ks> BEGIN UNLOGGED BATCH
			INSERT INTO tbl (pk, ck, v) VALUES (0, 1, 111); 
			INSERT INTO tbl (pk, ck, v) VALUES (0, 2, 222);
			APPLY BATCH;
cqlsh:ks> SELECT "cdc$batch_seq_no", "cdc$operation", "cdc$ttl",
			pk, ck, v from ks.tbl_scylla_cdc_log ;

 cdc$batch_seq_no | cdc$operation | cdc$ttl | pk | ck | v
------------------+---------------+---------+----+----+-----
...snip...
                0 |             0 |    null |  0 |  1 | 100
                1 |             2 |    null |  0 |  1 | 111
                2 |             9 |    null |  0 |  1 | 111
                3 |             0 |    null |  0 |  2 | 200
                4 |             2 |    null |  0 |  2 | 222
                5 |             9 |    null |  0 |  2 | 222
```

Preimage rows are represented by cdc operation 0, and postimage by 9.
Please note that all rows presented above share the same value of
cdc$time column, which was not shown here for brevity.

__Problems with previous approach__

This simple algorithm has some conceptual and implementational problems
which arise when processing more complicated CQL BATCH-es. Consider
the following example:

```cql
cqlsh:ks> BEGIN UNLOGGED BATCH
			INSERT INTO tbl (pk, ck, v1) VALUES (0, 0, 1) USING TTL 1000;
			INSERT INTO tbl (pk, ck, v2) VALUES (0, 0, 2) USING TTL 2000;
			APPLY BATCH;
cqlsh:ks> SELECT "cdc$batch_seq_no", "cdc$operation", "cdc$ttl",
			pk, ck, v1, v2 FROM tbl_scylla_cdc_log;

 cdc$batch_seq_no | cdc$operation | cdc$ttl | pk | ck | v1   | v2
------------------+---------------+---------+----+----+------+------
...snip...
                0 |             0 |    null |  0 |  0 | null |    0
                1 |             2 |    2000 |  0 |  0 | null |    2
                2 |             9 |    null |  0 |  0 |    0 |    2
                3 |             0 |    null |  0 |  0 |    0 | null
                4 |             1 |    1000 |  0 |  0 |    1 | null
                5 |             9 |    null |  0 |  0 |    1 |    0
```

A single cdc group (corresponding to rows sharing the same cdc$time)
might have more than one delta that modify the same row. For example,
this happens when modifying two columns of the same row with
different TTLs - due to our choice of CDC log schema, we must
represent such change with two delta rows.

It does not make sense to present a postimage after the first delta
and preimage before the second - both deltas are applied
simultaneously by the same CQL BATCH, so the middle "image" is purely
imaginary and does not appear at any point in the table.

Moreover, in this example, the last postimage is wrong - v1 is updated,
but v2 is not. None of the postimages presented above represent the
final state of the row.

__New algorithm__

The new algorithm works now on per cdc group basis, not delta row.
When starting processing a CQL BATCH:

- Load preimage query results into a data structure representing
  current state of the affected rows.

For each cdc group:

- For each row modified within the group, a preimage is produced,
  regardless if the row was present in the table. The preimage
  is calculated based on the current state. Only include columns
  that are modified for this row within the group.
- For each delta, produce a delta row and update the current state
  accordingly.
- Produce postimages in the same way as preimages - but include all
  columns for each row in the postimage.

The new algorithm produces postimage correctly when multiple deltas
affect one, because the state of the row is updated on the fly.

This algorithm moves preimage and postimage rows to the beginning and
the end of the cdc group, accordingly. This solves the problem of
imaginary preimages and postimages appearing inside a cdc group.

Unfortunately, it is possible for one CQL BATCH to contain changes that
use multiple timestamps. This will result in one CQL BATCH creating
multiple cdc groups, with different cdc$time. As it is impossible, with
our choice of schema, to tell that those cdc groups were created from
one CQL BATCH, instead we pretend as if those groups were separate CQL
operations. By tracking the state of the affected rows, we make sure
that preimage in later groups will reflect changes introduces in
previous groups.

One more thing - this algorithm should have the same results for
singular CQL operations and simple CQL BATCH-es, with one exception.
Previously, preimage not produced if a row was not present in the
table. Now, the preimage row will appear unconditionally - it will have
nulls in place of column values.


piodul:
I apologize for the number of commits in this patch - I needed to do some refactors before introducing necessary changes to the algorithm. I've split it up so that most commits should be very simple to understand and review.

cdc/log.cc
```diff
@@ -834,6 +834,8 @@ class transformer final {
      */
     std::unordered_set<const column_definition*> _non_atomic_column_deletes;
 
+    stats::part_type_set _touched_parts;
```
haaawk, line 838: I'm not sure why this change is needed. Maybe I will understand it while reading the following patches but just reading this patch it does not seem like a good idea. The previous approach was cleaner in my opinion.

cdc/log.cc
```diff
@@ -836,16 +836,19 @@ class transformer final {
 
     stats::part_type_set _touched_parts;
 
-    clustering_key set_pk_columns(const partition_key& pk, api::timestamp_type ts, bytes decomposed_tuuid, int batch_no, mutation& m) const {
+    api::timestamp_type _ts;
```
haaawk, line 846: I think it's a step into a good direction but I think it would be better to introduce a new object that wraps mutation and stores _ts and _tuuid and use it inside transform. 

I think splitting the API into two function calls is not a good idea.
It is also not a good idea, in my opinion, to store state in the transformer that lives only for a length of a single function.
It's better to encapsulate _ts, _tuuid, mutation and logic around them in an object and use it inside transform.

cdc/log.cc
```diff
@@ -838,6 +838,7 @@ class transformer final {
 
     api::timestamp_type _ts;
     bytes _tuuid;
+    lw_shared_ptr<cql3::untyped_result_set> _preimage_select_result;
```
haaawk: I'm not sure whether this is an improvement. Maybe reading the following patches will clear it up for me.

cdc/log.cc
```diff
@@ -1284,15 +1295,17 @@ class transformer final {
...
                     }
-                    set_operation(log_ck, cdc_op, res);
+                    set_operation(log_ck, cdc_op);
                 }
             }
         }
+    }
 
-        return res;
+    std::vector<mutation> take_generated_log_mutations() {
```
haaawk, line 1295: Maybe take_generated_log_mutations() && { to make it clear that the function destroys the object?

cdc/log.cc
haaawk: Now I see that transformer willl need to have state now and it makes sense to store preimage in it.

cdc/log.cc
haaawk: Now in later patch, I see that the two functions are necessary.

cdc/log.cc
```diff
@@ -1320,6 +1291,28 @@ class transformer final : public change_processor {
...
+        return std::nullopt;
+    }
+
+    cell_map* get_current_row_state(const clustering_key* ck) {
+        if (!ck) {
+            // static row case
+            return &_static_row_state;
+        } else {
+            auto it = _clustering_row_states.find(*ck);
+            if (it != _clustering_row_states.end()) {
```
haaawk, line 1318: Suggestion: maybe `return it == _clustering_row_states.end() ? nullptr : &it->second;`

cdc/log.cc
```diff
@@ -1320,6 +1291,28 @@ class transformer final : public change_processor {
...
+        if (state) {
+            if (auto it = state->find(&cdef); it != state->end()) {
+                return it->second;
+            }
+        }
+        return std::nullopt;
+    }
+
+    cell_map* get_current_row_state(const clustering_key* ck) {
+        if (!ck) {
```
haaawk, line 1313: Wouldn't it make sense to split those two cases apart into separated methods?
I have an impression that get_current_row_state with null and get_current_row_state with not null are called in different places so it will be not only more efficient but also more clear to have get_current_static_row_state() and get_current_ck_row_state(const clustering_key* ck).

cdc/log.cc
```diff
@@ -1282,19 +1257,15 @@ class transformer final : public change_processor {
                     if (r.row().deleted_at()) {
                         _touched_parts.set<stats::part_type::ROW_DELETE>();
                         cdc_op = operation::row_delete;
-                        if (pirow && pikey) {
+                        if (pikey) {
                             for (const column_definition& column: _schema->regular_columns()) {
-                                if (!pirow->has(column.name_as_text())) {
```
haaawk: Why is it ok to drop this? Won't line 1263 dereference unset optional (*value) when a column did not exist before?

cdc/log.cc
```diff
@@ -1458,8 +1451,53 @@ class transformer final : public change_processor {
...
+                }
+                ck_parts.emplace_back(*v);
+            }
+            auto ck = clustering_key::from_exploded(std::move(ck_parts));
+
+            // Collect regular rows
+            cell_map cells;
+            for (auto& c : _schema->regular_columns()) {
+                if (auto maybe_cell_view = get_preimage_col_value(c, &row)) {
+                    cells[&c] = *maybe_cell_view;
```
haaawk, line 1501: This concerns me because now we copy the values of cells more than before. This can have a big negative impact on performance.  What's worse it will affect not only batches but also workloads with no batches. Is there a way to optimize this? Maybe we could lazily compute get_col_from_row_state instead of precomputing it.

cdc/log.cc
```diff
@@ -962,6 +962,14 @@ class transformer final : public change_processor {
         _tuuid = timeuuid_type->decompose(generate_timeuuid(ts));
     }
 
+    void produce_preimage(const clustering_key* ck, const one_kind_column_set& columns_to_include) override {
+        // TODO
```
haaawk: I think this patch should be embedded in the following patch that actually makes use of those methods.

cdc/split.cc
```diff
@@ -70,13 +70,61 @@ struct partition_deletion {
...
+        process_change_type(clustered_updates);
+
+        return ret;
+    }
+
+    cdc::one_kind_column_set get_affected_static_columns(const schema& s) const {
+        cdc::one_kind_column_set ret{s.static_columns_count()};
+
+        for (const auto& change : static_updates) {
+            for (const auto& entry : change.atomic_entries) {
```
haaawk, line 118: These two inner for loops are very similar and there are another two for loops like that in process_change_type. Maybe we could extract a helper function to avoid code duplication?

cdc/log.cc
piodul: It's not OK, it's a bug :) thanks for spotting it. I'll fix it, even though this logic here is changed in later commits.

Preimage row may not exist for this ck, so `get_col_from_row_state` will return nullopt for each regular column, and we could hit this bug.

Otherwise we should get all regular columns of the deleted row, because we request them explicitly (see [here](https://github.com/piodul/scylla/blob/cdc-pre-postimage-persistence/cdc/log.cc#L1415))...

...unless there is a possibility of a schema changing in-flight and us racing with columns being added to the schema? In this case, we should also additionally check that `value` is set.

piodul:
v2:
- Marked `take_generated_log_mutations` with `&&`
- Changed if-else into a ternary operator
- Changed `get_current_row_state` into `get_current_clustering_row_state`
- Moved the same pattern used in `get_affected_clustered_columns_per_row` and `get_affected_static_columns` into a separate function
- Merged commit adding `produce_preimage` and `produce_postimage` methods with the commit which adds logic to split.cc that calls those methods (the commit is entitled `cdc: add interface for producing pre/postimages`)
- Fixed a bug that appears in one of the middle commits - requesting preimage when non-existent row is deleted would read uninitialized memory
- Prevented rows from being deleted from state when updating the state is not enabled
- Corrected a bug in unit test related to ordering of serialized timeuuids

Previous version can be found on branch [`cdc-pre-postimage-persistence-v1`](https://github.com/piodul/scylla/commits/cdc-pre-postimage-persistence-v1).

diff v1 v2:

```diff
diff --git a/cdc/log.cc b/cdc/log.cc
index 68dfc0c70..bbeace29f 100644
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -996,10 +996,15 @@ class transformer final : public change_processor {
 
         const auto kind = ck ? column_kind::regular_column : column_kind::static_column;
 
-        auto row_state = get_current_row_state(ck);
-        if (!row_state) {
-            // We have no data for this row, we can stop here
-            return;
+        cell_map* row_state;
+        if (ck) {
+            row_state = get_current_clustering_row_state(*ck);
+            if (!row_state) {
+                // We have no data for this row, we can stop here
+                return;
+            }
+        } else {
+            row_state = &_static_row_state;
         }
 
         auto process_cell = [&, this] (const column_definition& cdef) {
@@ -1239,10 +1244,9 @@ class transformer final : public change_processor {
 
             if (!p.static_row().empty()) {
                 _touched_parts.set<stats::part_type::STATIC_ROW>();
-                auto* row_state = get_current_row_state(nullptr);
 
                 auto log_ck = set_pk_columns();
-                auto ttl = process_cells(p.static_row().get(), column_kind::static_column, log_ck, nullptr, row_state);
+                auto ttl = process_cells(p.static_row().get(), column_kind::static_column, log_ck, nullptr, &_static_row_state);
 
                 set_operation(log_ck, operation::update);
 
@@ -1252,7 +1256,7 @@ class transformer final : public change_processor {
             } else {
                 _touched_parts.set_if<stats::part_type::CLUSTERING_ROW>(!p.clustered_rows().empty());
                 for (const rows_entry& r : p.clustered_rows()) {
-                    auto* row_state = get_current_row_state(&r.key());
+                    auto* row_state = get_current_clustering_row_state(r.key());
                     auto ck_value = r.key().explode(*_schema);
                     auto log_ck = set_pk_columns();
 
@@ -1269,7 +1273,7 @@ class transformer final : public change_processor {
                         _touched_parts.set<stats::part_type::ROW_DELETE>();
                         cdc_op = operation::row_delete;
 
-                        if (row_state) {
+                        if (_enable_updating_state && row_state) {
                             // Erase so that postimage will be empty
                             _clustering_row_states.erase(r.key());
                         }
@@ -1292,7 +1296,7 @@ class transformer final : public change_processor {
         }
     }
 
-    std::vector<mutation> take_generated_log_mutations() {
+    std::vector<mutation> take_generated_log_mutations() && {
         return std::move(_result_mutations);
     }
 
@@ -1309,17 +1313,9 @@ class transformer final : public change_processor {
         return std::nullopt;
     }
 
-    cell_map* get_current_row_state(const clustering_key* ck) {
-        if (!ck) {
-            // static row case
-            return &_static_row_state;
-        } else {
-            auto it = _clustering_row_states.find(*ck);
-            if (it != _clustering_row_states.end()) {
-                return &it->second;
-            }
-            return nullptr;
-        }
+    cell_map* get_current_clustering_row_state(const clustering_key& ck) {
+        auto it = _clustering_row_states.find(ck);
+        return it == _clustering_row_states.end() ? nullptr : &it->second;
     }
 
     bytes_opt get_preimage_col_value(const column_definition& cdef, const cql3::untyped_result_set_row *pirow) {
@@ -1598,7 +1594,7 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
                     tracing::trace(tr_state, "CDC: No need to split {}", m.decorated_key());
                     process_changes_without_splitting(m, trans, preimage, postimage);
                 }
-                auto log_mut = trans.take_generated_log_mutations();
+                auto log_mut = std::move(trans).take_generated_log_mutations();
                 const int generated_count = log_mut.size();
                 mutations.insert(mutations.end(), std::make_move_iterator(log_mut.begin()), std::make_move_iterator(log_mut.end()));
 
diff --git a/cdc/split.cc b/cdc/split.cc
index bfbad503f..a209a6a1f 100644
--- a/cdc/split.cc
+++ b/cdc/split.cc
@@ -25,6 +25,8 @@
 #include "split.hh"
 #include "log.hh"
 
+#include <type_traits>
+
 struct atomic_column_update {
     column_id id;
     atomic_cell cell;
@@ -72,6 +74,22 @@ struct partition_deletion {
 
 using clustered_column_set = std::map<clustering_key, cdc::one_kind_column_set, clustering_key::less_compare>;
 
+template<typename Container>
+concept EntryContainer = requires(Container& container) {
+    { container.atomic_entries } -> std::same_as<std::vector<atomic_column_update>&>;
+    { container.nonatomic_entries } -> std::same_as<std::vector<nonatomic_column_update>&>;
+};
+
+template<EntryContainer Container>
+static void add_columns_affected_by_entries(cdc::one_kind_column_set& cset, const Container& cont) {
+    for (const auto& entry : cont.atomic_entries) {
+        cset.set(entry.id);
+    }
+    for (const auto& entry : cont.nonatomic_entries) {
+        cset.set(entry.id);
+    }
+}
+
 struct batch {
     std::vector<static_row_update> static_updates;
     std::vector<clustered_row_insert> clustered_inserts;
@@ -96,12 +114,7 @@ struct batch {
             for (const auto& change : changes) {
                 auto& cset = ret[change.key];
                 cset.resize(s.regular_columns_count());
-                for (const auto& entry : change.atomic_entries) {
-                    cset.set(entry.id);
-                }
-                for (const auto& entry : change.nonatomic_entries) {
-                    cset.set(entry.id);
-                }
+                add_columns_affected_by_entries(cset, change);
             }
         };
 
@@ -113,16 +126,9 @@ struct batch {
 
     cdc::one_kind_column_set get_affected_static_columns(const schema& s) const {
         cdc::one_kind_column_set ret{s.static_columns_count()};
-
         for (const auto& change : static_updates) {
-            for (const auto& entry : change.atomic_entries) {
-                ret.set(entry.id);
-            }
-            for (const auto& entry : change.nonatomic_entries) {
-                ret.set(entry.id);
-            }
+            add_columns_affected_by_entries(ret, change);
         }
-
         return ret;
     }
 };
diff --git a/test/boost/cdc_test.cc b/test/boost/cdc_test.cc
index 3d23b9e26..c75cc6b74 100644
--- a/test/boost/cdc_test.cc
+++ b/test/boost/cdc_test.cc
@@ -1663,7 +1663,7 @@ static void test_pre_post_image(cql_test_env& e, const std::vector<image_persist
         sort_by_time(*rows, results);
 
         // Indexed by serialized timeuuid of the group
-        std::map<bytes, std::vector<std::vector<bytes_opt>>> groups;
+        std::map<bytes, std::vector<std::vector<bytes_opt>>, serialized_compare> groups(timeuuid_type->as_less_comparator());
         const auto time_index = column_index(*rows, cdc::log_meta_column_name("time"));
         for (const auto& row : results) {
             const auto time = *row[time_index];
@@ -1721,9 +1721,12 @@ static void test_pre_post_image(cql_test_env& e, const std::vector<image_persist
             // Filter preimage and postimage
             // TODO: Assert that all preimages are at the beginning,
             // and that postimages are at the end
+            const auto& actual_group_id = actual_it->first;
             const auto& actual_results = (actual_it++)->second;
             const auto& expected_set = *expected_it++;
 
+            BOOST_TEST_MESSAGE(format("Checking group {}", actual_group_id));
+
             const auto actual_preimage = filter_by_operation(*rows, actual_results, cdc::operation::pre_image);
             if (preimage) {
                 BOOST_TEST_MESSAGE("Checking preimage");
```

cdc/split.cc
piodul: Done.

cdc/log.cc
piodul: Done

cdc/log.cc
piodul: Done.

cdc/log.cc
piodul: I agree. Done.

cdc/log.cc
piodul: Done

cdc/log.cc
piodul: I agree that in current shape this PR adds more copying. I'll try to do something about it.

I'm thinking of writing a consumer object that we could put in place of the one we pass [here](https://github.com/piodul/scylla/blob/cdc-pre-postimage-persistence/cdc/log.cc#L1440). This way we would avoid constructing an `untyped_result_set` object, and instead populate our state directly.

If that doesn't work, I'll try to make initialization lazy.

haaawk: Let's do this as a follow up. This change is complex enough.

cdc/log.cc
```diff
@@ -939,24 +949,25 @@ class transformer final {
...
         _tuuid = std::move(tuuid);
     }
 
     // TODO: is pre-image data based on query enough. We only have actual column data. Do we need
     // more details like tombstones/ttl? Probably not but keep in mind.
-    mutation transform(const mutation& m, int& batch_no) {
-        auto stream_id = _ctx._cdc_metadata.get_stream(_ts, m.token());
-        mutation res(_log_schema, stream_id.to_partition_key(*_log_schema));
+    void transform(const mutation& m, int& batch_no) {
+        mutation& res = current_mutation();
```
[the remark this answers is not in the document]
piodul: Sure it is. In original code, `mutation res` was created at the beginning of `transform` and returned at the end of the function. Now, we take a reference to an existing mutation.

In both versions, `transform` appends new log rows to the `res` mutation.

cdc/log.cc
```diff
@@ -1458,8 +1444,53 @@ class transformer final : public change_processor {
       }
     }
 
-    void set_preimage_select_result(lw_shared_ptr<cql3::untyped_result_set> rs) {
-        _preimage_select_result = std::move(rs);
+    // Note: this assumes that the results are from one partition only
+    void load_preimage_results_into_state(lw_shared_ptr<cql3::untyped_result_set> preimage_set, bool static_only) {
```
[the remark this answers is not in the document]
piodul: In the future, I'd like to change the way load preimage results to the state, see [my other comment](https://github.com/scylladb/scylla/pull/6741#discussion_r449011530). It will have to be a method of the `transformer` in order to be able to access `_clustering_row_states` and `_static_row_state` fields.

cdc/split.hh
```diff
@@ -31,8 +31,17 @@ class mutation;
 
 namespace cdc {
 
+class change_processor {
+public:
+    virtual ~change_processor() {};
+
+    virtual void begin_timestamp(api::timestamp_type ts, bytes tuuid) = 0;
+
+    virtual void process_change(const mutation& m, int& batch_no) = 0;
```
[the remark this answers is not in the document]
elcallio: Agree on docs. But like the approach.

cdc/log.cc
```diff
@@ -847,10 +848,10 @@ class transformer final : public change_processor {
         return _result_mutations.back();
     }
 
-    clustering_key set_pk_columns(int batch_no) {
+    clustering_key set_pk_columns() {
```
elcallio, line 854: Hey, you did it. Neat!

cdc/log.cc
```diff
@@ -1232,15 +1243,15 @@ class transformer final {
                     }
 
                     if (preimage) {
-                        pikey = set_pk_columns(m.key(), batch_no++, res);
-                        set_operation(*pikey, operation::pre_image, res);
+                        pikey = set_pk_columns(batch_no++);
+                        set_operation(*pikey, operation::pre_image);
                     }
```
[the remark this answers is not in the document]
elcallio: Oh, you did! Nice.

cdc/log.cc
```diff
@@ -838,6 +838,10 @@ class transformer final : public change_processor {
 
     stats::part_type_set _touched_parts;
 
+    using cell_map = std::unordered_map<const column_definition*, bytes_opt>;
```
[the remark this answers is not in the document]
elcallio: They only change on schema change, and then the schema is replaced. So as long as schema is held alive.

cdc/log.cc
```diff
@@ -1204,18 +1206,18 @@ class transformer final : public change_processor {
...
-                        prev = std::nullopt;
                     }
 
                     if (value) {
                         res.set_cell(log_ck, *dst, atomic_cell::make_live(*dst->type, _ts, *value, _cdc_ttl_opt));
                     }
 
+                    if (_enable_updating_state) {
+                    // don't merge with pre-image iff column delete
+                    bytes_opt prev = is_column_delete ? std::nullopt : get_col_from_row_state(row_state, cdef);
```
elcallio: indent?

elcallio: aha, second patch.

piodul:
v3:

- Added assertions in places requested by Kamil
- Merged `take_generated_log_mutations` and `get_touched_parts` into `finish` method
- Made destructor of `change_processor` non-virtual and private
- Documented methods of `change_processor`
- Moved `find_timestamp` to `split.cc`
- Renamed `set_pk_columns` to `allocate_new_log_row`

Previous version can be found on branch [cdc-pre-postimage-persistence-v2](https://github.com/piodul/scylla/tree/cdc-pre-postimage-persistence-v2).

diff v2 v3:

```diff
diff --git a/cdc/log.cc b/cdc/log.cc
index bbeace29f..c58371ec7 100644
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -511,106 +511,6 @@ db_context db_context::builder::build() {
     };
 }
 
-/* Find some timestamp inside the given mutation.
- *
- * If this mutation was created using a single insert/update/delete statement, then it will have a single,
- * well-defined timestamp (even if this timestamp occurs multiple times, e.g. in a cell and row_marker).
- *
- * This function shouldn't be used for mutations that have multiple different timestamps: the function
- * would only find one of them. When dealing with such mutations, the caller should first split the mutation
- * into multiple ones, each with a single timestamp.
- */
-// TODO: We need to
-// - in the code that calls `augument_mutation_call`, or inside `augument_mutation_call`,
-//   split each mutation to a set of mutations, each with a single timestamp.
-// - optionally: here, throw error if multiple timestamps are encountered (may degrade performance).
-// external linkage for testing
-api::timestamp_type find_timestamp(const mutation& m) {
-    auto& p = m.partition();
-    const auto& s = *m.schema();
-    api::timestamp_type t = api::missing_timestamp;
-
-    t = p.partition_tombstone().timestamp;
-    if (t != api::missing_timestamp) {
-        return t;
-    }
-
-    for (auto& rt: p.row_tombstones()) {
-        t = rt.tomb.timestamp;
-        if (t != api::missing_timestamp) {
-            return t;
-        }
-    }
-
-    auto walk_row = [&t, &s] (const row& r, column_kind ckind) {
-        r.for_each_cell_until([&t, &s, ckind] (column_id id, const atomic_cell_or_collection& cell) {
-            auto& cdef = s.column_at(ckind, id);
-
-            if (cdef.is_atomic()) {
-                t = cell.as_atomic_cell(cdef).timestamp();
-                if (t != api::missing_timestamp) {
-                    return stop_iteration::yes;
-                }
-                return stop_iteration::no;
-            }
-
-            return cell.as_collection_mutation().with_deserialized(*cdef.type,
-                    [&] (collection_mutation_view_description mview) {
-                t = mview.tomb.timestamp;
-                if (t != api::missing_timestamp) {
-                    // A collection tombstone with timestamp T can be created with:
-                    // UPDATE ks.t USING TIMESTAMP T + 1 SET X = null WHERE ...
-                    // where X is a non-atomic column.
-                    // This is, among others, the reason why we show it in the CDC log
-                    // with cdc$time using timestamp T + 1 instead of T.
-                    t += 1;
-                    return stop_iteration::yes;
-                }
-
-                for (auto& kv : mview.cells) {
-                    t = kv.second.timestamp();
-                    if (t != api::missing_timestamp) {
-                        return stop_iteration::yes;
-                    }
-                }
-
-                return stop_iteration::no;
-            });
-        });
-    };
-
-    walk_row(p.static_row().get(), column_kind::static_column);
-    if (t != api::missing_timestamp) {
-        return t;
-    }
-
-    for (const rows_entry& cr : p.clustered_rows()) {
-        const deletable_row& r = cr.row();
-
-        t = r.deleted_at().regular().timestamp;
-        if (t != api::missing_timestamp) {
-            return t;
-        }
-
-        t = r.deleted_at().shadowable().tomb().timestamp;
-        if (t != api::missing_timestamp) {
-            return t;
-        }
-
-        t = r.created_at();
-        if (t != api::missing_timestamp) {
-            return t;
-        }
-
-        walk_row(r.cells(), column_kind::regular_column);
-        if (t != api::missing_timestamp) {
-            return t;
-        }
-    }
-
-    throw std::runtime_error("cdc: could not find timestamp of mutation");
-}
-
 // iterators for collection merge
 template<typename T>
 class collection_iterator : public std::iterator<std::input_iterator_tag, const T> {
@@ -848,10 +748,12 @@ class transformer final : public change_processor {
     bool _enable_updating_state = false;
 
     mutation& current_mutation() {
+        assert(!_result_mutations.empty());
         return _result_mutations.back();
     }
 
-    clustering_key set_pk_columns() {
+    clustering_key allocate_new_log_row() {
+        assert(_batch_no != -1);
         auto& m = current_mutation();
         const auto log_ck = clustering_key::from_exploded(
                 *m.schema(), { _tuuid, int32_type->decompose(_batch_no++) });
@@ -976,7 +878,7 @@ class transformer final : public change_processor {
 
         auto& res = current_mutation();
 
-        auto image_ck = set_pk_columns();
+        auto image_ck = allocate_new_log_row();
         set_operation(image_ck, op);
 
         if (ck) {
@@ -1036,7 +938,7 @@ class transformer final : public change_processor {
         if (p.partition_tombstone()) {
             // Partition deletion
             _touched_parts.set<stats::part_type::PARTITION_DELETE>();
-            auto log_ck = set_pk_columns();
+            auto log_ck = allocate_new_log_row();
             set_operation(log_ck, operation::partition_delete);
         } else if (!p.row_tombstones().empty()) {
             // range deletion
@@ -1059,7 +961,7 @@ class transformer final : public change_processor {
                     }
                 };
                 {
-                    auto log_ck = set_pk_columns();
+                    auto log_ck = allocate_new_log_row();
                     set_bound(log_ck, rt.start);
                     const auto start_operation = rt.start_kind == bound_kind::incl_start
                             ? operation::range_delete_start_inclusive
@@ -1067,7 +969,7 @@ class transformer final : public change_processor {
                     set_operation(log_ck, start_operation);
                 }
                 {
-                    auto log_ck = set_pk_columns();
+                    auto log_ck = allocate_new_log_row();
                     set_bound(log_ck, rt.end);
                     const auto end_operation = rt.end_kind == bound_kind::incl_end
                             ? operation::range_delete_end_inclusive
@@ -1245,7 +1147,7 @@ class transformer final : public change_processor {
             if (!p.static_row().empty()) {
                 _touched_parts.set<stats::part_type::STATIC_ROW>();
 
-                auto log_ck = set_pk_columns();
+                auto log_ck = allocate_new_log_row();
                 auto ttl = process_cells(p.static_row().get(), column_kind::static_column, log_ck, nullptr, &_static_row_state);
 
                 set_operation(log_ck, operation::update);
@@ -1258,7 +1160,7 @@ class transformer final : public change_processor {
                 for (const rows_entry& r : p.clustered_rows()) {
                     auto* row_state = get_current_clustering_row_state(r.key());
                     auto ck_value = r.key().explode(*_schema);
-                    auto log_ck = set_pk_columns();
+                    auto log_ck = allocate_new_log_row();
 
                     size_t pos = 0;
                     for (const auto& column : _schema->clustering_key_columns()) {
@@ -1296,12 +1198,10 @@ class transformer final : public change_processor {
         }
     }
 
-    std::vector<mutation> take_generated_log_mutations() && {
-        return std::move(_result_mutations);
-    }
-
-    stats::part_type_set get_touched_parts() const {
-        return _touched_parts;
+    // Takes and returns generated cdc log mutations and associated statistics about parts touched during transformer's lifetime.
+    // The `transformer` object on which this method was called on should not be used anymore.
+    std::tuple<std::vector<mutation>, stats::part_type_set> finish() && {
+        return std::make_pair<std::vector<mutation>, stats::part_type_set>(std::move(_result_mutations), std::move(_touched_parts));
     }
 
     bytes_opt get_col_from_row_state(const cell_map* state, const column_definition& cdef) const {
@@ -1594,13 +1494,13 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
                     tracing::trace(tr_state, "CDC: No need to split {}", m.decorated_key());
                     process_changes_without_splitting(m, trans, preimage, postimage);
                 }
-                auto log_mut = std::move(trans).take_generated_log_mutations();
+                auto [log_mut, touched_parts] = std::move(trans).finish();
                 const int generated_count = log_mut.size();
                 mutations.insert(mutations.end(), std::make_move_iterator(log_mut.begin()), std::make_move_iterator(log_mut.end()));
 
                 // `m` might be invalidated at this point because of the push_back to the vector
                 tracing::trace(tr_state, "CDC: Generated {} log mutations from {}", generated_count, mutations[idx].decorated_key());
-                details.touched_parts.add(trans.get_touched_parts());
+                details.touched_parts.add(touched_parts);
             });
         }).then([this, tr_state, &details](std::vector<mutation> mutations) {
             tracing::trace(tr_state, "CDC: Finished generating all log mutations");
diff --git a/cdc/log.hh b/cdc/log.hh
index cf8c3a697..dd2c19871 100644
--- a/cdc/log.hh
+++ b/cdc/log.hh
@@ -142,6 +142,5 @@ seastar::sstring log_data_column_deleted_elements_name(std::string_view column_n
 bytes log_data_column_deleted_elements_name_bytes(const bytes& column_name);
 
 utils::UUID generate_timeuuid(api::timestamp_type t);
-api::timestamp_type find_timestamp(const mutation& m);
 
 } // namespace cdc
diff --git a/cdc/split.cc b/cdc/split.cc
index a209a6a1f..5a53934da 100644
--- a/cdc/split.cc
+++ b/cdc/split.cc
@@ -319,6 +319,106 @@ set_of_changes extract_changes(const mutation& base_mutation) {
 
 namespace cdc {
 
+/* Find some timestamp inside the given mutation.
+ *
+ * If this mutation was created using a single insert/update/delete statement, then it will have a single,
+ * well-defined timestamp (even if this timestamp occurs multiple times, e.g. in a cell and row_marker).
+ *
+ * This function shouldn't be used for mutations that have multiple different timestamps: the function
+ * would only find one of them. When dealing with such mutations, the caller should first split the mutation
+ * into multiple ones, each with a single timestamp.
+ */
+// TODO: We need to
+// - in the code that calls `augument_mutation_call`, or inside `augument_mutation_call`,
+//   split each mutation to a set of mutations, each with a single timestamp.
+// - optionally: here, throw error if multiple timestamps are encountered (may degrade performance).
+// external linkage for testing
+api::timestamp_type find_timestamp(const mutation& m) {
+    auto& p = m.partition();
+    const auto& s = *m.schema();
+    api::timestamp_type t = api::missing_timestamp;
+
+    t = p.partition_tombstone().timestamp;
+    if (t != api::missing_timestamp) {
+        return t;
+    }
+
+    for (auto& rt: p.row_tombstones()) {
+        t = rt.tomb.timestamp;
+        if (t != api::missing_timestamp) {
+            return t;
+        }
+    }
+
+    auto walk_row = [&t, &s] (const row& r, column_kind ckind) {
+        r.for_each_cell_until([&t, &s, ckind] (column_id id, const atomic_cell_or_collection& cell) {
+            auto& cdef = s.column_at(ckind, id);
+
+            if (cdef.is_atomic()) {
+                t = cell.as_atomic_cell(cdef).timestamp();
+                if (t != api::missing_timestamp) {
+                    return stop_iteration::yes;
+                }
+                return stop_iteration::no;
+            }
+
+            return cell.as_collection_mutation().with_deserialized(*cdef.type,
+                    [&] (collection_mutation_view_description mview) {
+                t = mview.tomb.timestamp;
+                if (t != api::missing_timestamp) {
+                    // A collection tombstone with timestamp T can be created with:
+                    // UPDATE ks.t USING TIMESTAMP T + 1 SET X = null WHERE ...
+                    // where X is a non-atomic column.
+                    // This is, among others, the reason why we show it in the CDC log
+                    // with cdc$time using timestamp T + 1 instead of T.
+                    t += 1;
+                    return stop_iteration::yes;
+                }
+
+                for (auto& kv : mview.cells) {
+                    t = kv.second.timestamp();
+                    if (t != api::missing_timestamp) {
+                        return stop_iteration::yes;
+                    }
+                }
+
+                return stop_iteration::no;
+            });
+        });
+    };
+
+    walk_row(p.static_row().get(), column_kind::static_column);
+    if (t != api::missing_timestamp) {
+        return t;
+    }
+
+    for (const rows_entry& cr : p.clustered_rows()) {
+        const deletable_row& r = cr.row();
+
+        t = r.deleted_at().regular().timestamp;
+        if (t != api::missing_timestamp) {
+            return t;
+        }
+
+        t = r.deleted_at().shadowable().tomb().timestamp;
+        if (t != api::missing_timestamp) {
+            return t;
+        }
+
+        t = r.created_at();
+        if (t != api::missing_timestamp) {
+            return t;
+        }
+
+        walk_row(r.cells(), column_kind::regular_column);
+        if (t != api::missing_timestamp) {
+            return t;
+        }
+    }
+
+    throw std::runtime_error("cdc: could not find timestamp of mutation");
+}
+
 bool should_split(const mutation& base_mutation) {
     const auto& base_schema = *base_mutation.schema();
     auto& p = base_mutation.partition();
diff --git a/cdc/split.hh b/cdc/split.hh
index 611dadeb7..7f329471d 100644
--- a/cdc/split.hh
+++ b/cdc/split.hh
@@ -32,18 +32,40 @@ class mutation;
 
 namespace cdc {
 
+// Represents a set of column ids of one kind (partition key, clustering key, regular row or static row).
+// There already exists a column_set type, but it keeps ordinal_column_ids, not column_ids (ordinal column ids
+// are unique across whole table, while kind-specific ids are unique only within one column kind).
+// To avoid converting back and forth between ordinal and kind-specific ids, one_kind_column_set is used instead.
 using one_kind_column_set = boost::dynamic_bitset<uint64_t>;
 
+// An object that processes changes from a single, big mutation.
+// It is intended to be used with process_changes_xxx_splitting. Those functions define the order and layout in which
+// changes should appear in CDC log, and change_processor is responsible for producing CDC log rows from changes given
+// by those two functions.
 class change_processor {
+protected:
+    ~change_processor() {};
 public:
-    virtual ~change_processor() {};
-
+    // Tells the processor that changes that follow from now on will be of given timestamp.
+    // begin_timestamp can be called only once for a given timestamp and change_processor object.
+    //   ts - timestamp of mutation parts
     virtual void begin_timestamp(api::timestamp_type ts, bool is_last) = 0;
 
-    // if ck is null, pre/postimage for static row is requested
+    // Tells the processor to produce a preimage for a given clustering/static row.
+    //   ck - clustering key of the row for which to produce a preimage; if nullptr, static row preimage is requested
+    //   columns_to_include - include information about the current state of those columns only, leave others as null
     virtual void produce_preimage(const clustering_key* ck, const one_kind_column_set& columns_to_include) = 0;
+
+    // Tells the processor to produce a postimage for a given clustering/static row.
+    // Contrary to preimage, this requires data from all columns to be present.
+    //   ck - clustering key of the row for which to produce a postimage; if nullptr, static row postimage is requested
     virtual void produce_postimage(const clustering_key* ck) = 0;
 
+    // Processes a smaller mutation which is a subset of the big mutation.
+    // The mutation provided to process_change should be simple enough for it to be possible to convert it
+    // into CDC log rows - for example, it cannot represent a write to two columns of the same row, where
+    // both columns have different timestamp or TTL set.
+    //   m - the small mutation to be converted into CDC log rows.
     virtual void process_change(const mutation& m) = 0;
 };
 
```

cdc/log.cc
```diff
@@ -836,16 +836,19 @@ class transformer final {
 
     stats::part_type_set _touched_parts;
 
-    clustering_key set_pk_columns(const partition_key& pk, api::timestamp_type ts, bytes decomposed_tuuid, int batch_no, mutation& m) const {
+    api::timestamp_type _ts;
+    bytes _tuuid;
```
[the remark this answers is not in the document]
piodul: Done

cdc/log.cc
```diff
@@ -836,14 +837,20 @@ class transformer final {
...
     stats::part_type_set _touched_parts;
 
+    std::vector<mutation> _result_mutations;
     api::timestamp_type _ts;
     bytes _tuuid;
     lw_shared_ptr<cql3::untyped_result_set> _preimage_select_result;
 
-    clustering_key set_pk_columns(const partition_key& pk, int batch_no, mutation& m) {
+    mutation& current_mutation() {
+        return _result_mutations.back();
```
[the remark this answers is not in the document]
piodul: Done

cdc/log.cc
```diff
@@ -1521,27 +1534,25 @@ cdc::cdc_service::impl::augment_mutation_call(lowres_clock::time_point timeout,
...
-                    generated_count = 1;
+                    trans.transform(m, batch_no);
                 }
+                auto log_mut = std::move(trans).take_generated_log_mutations();
+                const int generated_count = log_mut.size();
+                mutations.insert(mutations.end(), std::make_move_iterator(log_mut.begin()), std::make_move_iterator(log_mut.end()));
+
                 // `m` might be invalidated at this point because of the push_back to the vector
                 tracing::trace(tr_state, "CDC: Generated {} log mutations from {}", generated_count, mutations[idx].decorated_key());
                 details.touched_parts.add(trans.get_touched_parts());
```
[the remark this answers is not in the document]
piodul: Good idea - done.

cdc/split.cc
```diff
@@ -470,27 +471,35 @@ void for_each_change(const mutation& base_mutation,
...
             m.partition().apply(btch.partition_deletions->t);
-            f(std::move(m), change_ts, tuuid, batch_no);
+            processor.process_change(m, batch_no);
         }
     }
 }
 
+void process_changes_without_splitting(const mutation& base_mutation, change_processor& processor) {
+    int batch_no = 0;
+    auto ts = find_timestamp(base_mutation);
```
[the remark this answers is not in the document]
piodul: Done

cdc/split.hh
```diff
@@ -31,8 +31,17 @@ class mutation;
 
 namespace cdc {
 
+class change_processor {
+public:
+    virtual ~change_processor() {};
```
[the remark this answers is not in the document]
piodul: Done

cdc/split.hh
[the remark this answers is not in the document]
piodul: Added comments with explanation of each method (apart from destructor) and each parameter.

cdc/log.cc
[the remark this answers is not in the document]
piodul: Done

cdc/log.cc
```diff
@@ -847,10 +848,10 @@ class transformer final : public change_processor {
         return _result_mutations.back();
     }
 
-    clustering_key set_pk_columns(int batch_no) {
+    clustering_key set_pk_columns() {
         auto& m = current_mutation();
         const auto log_ck = clustering_key::from_exploded(
```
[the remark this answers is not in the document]
piodul: Done

cdc/split.hh
```diff
@@ -31,17 +32,25 @@ class mutation;
 
 namespace cdc {
 
+using one_kind_column_set = boost::dynamic_bitset<uint64_t>;
```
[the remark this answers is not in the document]
piodul: Done

cdc/log.cc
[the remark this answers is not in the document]
kbr-: Where are the comments? :(

Btw., it would also be nice to comment what `_enable_updating_state` is.

cdc/split.hh
```diff
@@ -31,8 +32,47 @@ class mutation;
...
+// changes should appear in CDC log, and change_processor is responsible for producing CDC log rows from changes given
+// by those two functions.
+class change_processor {
+protected:
+    ~change_processor() {};
+public:
+    // Tells the processor that changes that follow from now on will be of given timestamp.
+    // begin_timestamp can be called only once for a given timestamp and change_processor object.
+    //   ts - timestamp of mutation parts
+    virtual void begin_timestamp(api::timestamp_type ts, bool is_last) = 0;
```
kbr-, line 52: This method must be called in increasing timestamp order. IMO it should be documented.

(that's why I don't like keeping state: there are too many moving parts, and you must always remember not to do something at the wrong moment)

cdc/log.cc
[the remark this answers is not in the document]
piodul: Sorry, I thought you were asking for comments about `change_processor::begin_timestamp` arguments... I'll add more comments in the places you requested.

cdc/split.hh
```diff
@@ -31,8 +32,47 @@ class mutation;
...
+
+    // Tells the processor to produce a preimage for a given clustering/static row.
+    //   ck - clustering key of the row for which to produce a preimage; if nullptr, static row preimage is requested
+    //   columns_to_include - include information about the current state of those columns only, leave others as null
+    virtual void produce_preimage(const clustering_key* ck, const one_kind_column_set& columns_to_include) = 0;
```
kbr-, line 57: Should also document when is is safe to call this function. Same for functions below.

cdc/split.hh
```diff
@@ -31,8 +32,47 @@ class mutation;
...
+
+    // Tells the processor to produce a postimage for a given clustering/static row.
+    // Contrary to preimage, this requires data from all columns to be present.
+    //   ck - clustering key of the row for which to produce a postimage; if nullptr, static row postimage is requested
+    virtual void produce_postimage(const clustering_key* ck) = 0;
+
+    // Processes a smaller mutation which is a subset of the big mutation.
+    // The mutation provided to process_change should be simple enough for it to be possible to convert it
+    // into CDC log rows - for example, it cannot represent a write to two columns of the same row, where
+    // both columns have different timestamp or TTL set.
```
kbr-, line 67: Would it be possible to specify precisely what "simple enough" means?
E.g.: single timestamp, single TTL, no tombstones and live cells at the same time?

cdc/log.cc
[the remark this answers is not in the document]
piodul: It should be easy to switch, because `column_definition` already keeps `column_id` as a field.

cdc/split.hh
piodul: I think that, ironically, the definition of `simple enough` would be complicated and I don't have a constructive definition.
The most exact definition I could come up with is `simple enough for it to be possible to convert it into CDC log rows`.
If you could propose a better definition, then I'm happy to add it as a comment.

piodul:
v4: more comments on `transformer` fields, and better descriptions in `change_processor`.

Previous version can be found on branch [cdc-pre-postimage-persistence-v3](https://github.com/piodul/scylla/tree/cdc-pre-postimage-persistence-v3).

diff v3 v4:

```diff
diff --git a/cdc/log.cc b/cdc/log.cc
index c58371ec7..b1f0fd57a 100644
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -742,9 +742,14 @@ class transformer final : public change_processor {
     cell_map _static_row_state;
 
     std::vector<mutation> _result_mutations;
+
+    // Keeps the next cdc$batch_seq_no value
     int _batch_no = -1;
+    // The timestamp of changes being currently processed
     api::timestamp_type _ts;
+    // The cdc$time value of changes being currently processed
     bytes _tuuid;
+    // When enabled, process_change will update _clustering_row_states and _static_row_state
     bool _enable_updating_state = false;
 
     mutation& current_mutation() {
diff --git a/cdc/split.hh b/cdc/split.hh
index 7f329471d..6fc9619ce 100644
--- a/cdc/split.hh
+++ b/cdc/split.hh
@@ -42,13 +42,23 @@ using one_kind_column_set = boost::dynamic_bitset<uint64_t>;
 // It is intended to be used with process_changes_xxx_splitting. Those functions define the order and layout in which
 // changes should appear in CDC log, and change_processor is responsible for producing CDC log rows from changes given
 // by those two functions.
+//
+// The flow of calling its methods should go as follows:
+//   -> begin_timestamp #1
+//     -> produce_preimage (one call for each preimage row to be generated)
+//     -> process_change (one call for each part generated by the splitting function)
+//     -> produce_postimage (one call for each postimage row to be generated)
+//   -> begin_timestamp #2
+//   ...
 class change_processor {
 protected:
     ~change_processor() {};
 public:
     // Tells the processor that changes that follow from now on will be of given timestamp.
+    // This method must be called in increasing timestamp order.
     // begin_timestamp can be called only once for a given timestamp and change_processor object.
     //   ts - timestamp of mutation parts
+    //   is_last - determines if this will be the last timestamp to be processed by this change_processor instance.
     virtual void begin_timestamp(api::timestamp_type ts, bool is_last) = 0;
 
     // Tells the processor to produce a preimage for a given clustering/static row.
```

piodul:
Please do not merge yet, I found a small bug in this PR while fixing CDC dtests. I'll provide a patch after I verify that no more bugs are detected by dtests.

piodul:
v5: Fixed a bug where a column delete would not be reflected in the postimage. Also, rebased to newest master.

Previous version can be found on branch [cdc-pre-postimage-persistence-v4](https://github.com/piodul/scylla/tree/cdc-pre-postimage-persistence-v4).

```diff
diff --git a/cdc/log.cc b/cdc/log.cc
index b1f0fd57a..e5411f319 100644
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -1133,16 +1133,14 @@ class transformer final : public change_processor {
                                 return merge(type, prev, value, deleted_elements);
                             });
                         }
-                        if (next) {
-                            if (!row_state) {
-                                // static row always has a valid state, so this must be
-                                // a clustering row missing
-                                assert(base_ck);
-                                auto [it, inserted] = _clustering_row_states.try_emplace(*base_ck);
-                                row_state = &it->second;
-                            }
-                            (*row_state)[&cdef] = std::move(next);
+                        if (!row_state) {
+                            // static row always has a valid state, so this must be
+                            // a clustering row missing
+                            assert(base_ck);
+                            auto [it, inserted] = _clustering_row_states.try_emplace(*base_ck);
+                            row_state = &it->second;
                         }
+                        (*row_state)[&cdef] = std::move(next);
                     }
                 });
```

piodul:
Updated dtests, PR is [here](https://github.com/scylladb/scylla-dtest/pull/1589). I ran both of the modified test files on my machine, with scylla built in dev mode, and all tests passed.

nyh:
Committed to next: 9042161ba3971de8bfd9067d9c610291a872aa13

PR #6742, opened 2020-07-01
filtering: Do not compare NULLs in <, <=, >, >=

### LINKED ISSUE #6295: Filtering captures uninitialized/deleted values of certain types
On fresh master, e.g. 7304a795e584f2697f37d0ca78e8428127c8c3d1

Unset values pass some predicates when filtering:
```

 pk | i1 | i2
----+----+------
  0 |  0 | null

(1 rows)
```
Sadly, deleted values behave in the same way:
```
cqlsh> INSERT INTO ks.tbl_int2 (pk, i1, i2) VALUES (1, 1, 1);
cqlsh> DELETE i2 from ks.tbl_int2 WHERE pk = 1;

 pk | i1 | i2
----+----+------
  1 |  1 | null
  0 |  0 | null

(2 rows)
```
I checked that it happens for all integer types, but other types (like floats or UUIDs) have similar semantics of `compare_visitor` and might be affected as well. Possible solution would be to exclude NULLs from filtering on higher levels.

jul-stas:
When `ALLOW FILTERING` was applied, NULLs used to give nonsensical results of GT, LT, GEQ and LEQ comparison. This patch excludes NULLs from comparison by assuring that NULL is never a member of any range.

The EQ and IN operators did not need any changes.

Fixes #6295

### DIFF
--- a/cql3/restrictions/statement_restrictions.cc
+++ b/cql3/restrictions/statement_restrictions.cc
@@ -708,6 +708,9 @@ bool single_column_restriction::slice::is_satisfied_by(const schema& schema,
 }
 
 bool single_column_restriction::slice::is_satisfied_by(bytes_view data, const query_options& options) const {
+    if (data.empty()) {
+        return false; // NULL is not contained in any range
+    }
     return to_range(_slice, options, _column_def.name_as_text()).contains(
             data, _column_def.type->underlying_type()->as_tri_comparator());
 }
--- a/test/boost/filtering_test.cc
+++ b/test/boost/filtering_test.cc
@@ -1176,3 +1176,155 @@ SEASTAR_TEST_CASE(test_filtering) {
 
     });
 }
+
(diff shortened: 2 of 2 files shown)


cql3/restrictions/statement_restrictions.cc
```diff
@@ -708,6 +708,9 @@ bool single_column_restriction::slice::is_satisfied_by(const schema& schema,
 }
 
 bool single_column_restriction::slice::is_satisfied_by(bytes_view data, const query_options& options) const {
+    if (data.empty()) {
+        return false; // NULL is not contained in any range
+    }
```
avikivity, line 713: NULLs are signified by a bytes_view_opt being disengaged. An empty string is not null.

jul-stas: Ah, I see. So it's much harder, I'm closing this one.

PR #6743, opened 2020-07-02
repair: Add synchronous API to query repair status

### LINKED ISSUE #6445: API: Add long polling to StorageServiceRepairAsyncByKeyspaceGet
This is needed to make repairs faster as measured in walltime.

Scylla Manager asks Scylla to repair a given token range and then waits for Scylla to end using `/storage_service/repair_async/{keyspace}` endpoint.

The API where we need to poll for repair job results each Xms introduces a significant delay in the whole process as seen by user.

Let's say we have
- 9 nodes 
- 5k tables
- 4 shards each node
- 16 ranges per shard
- RF=3

If we poll each 50ms, then the mean wasted time per operation is 25ms.

That makes 25 minutes of walltime spent on waiting that could be avoided `(9 * 256 * 5000) / (4 * 16) * 25 / 1000 / 3 / 60`. The wasted CPU time is 3 times as much. It explodes with nr. of tables and if you want to go slower it degradates as well.

Repairing 1 token range at a time results in over a DAY (1600 minutes) of waiting that could be avoided `(9 * 256 * 5000) * 25 / 1000 / 60 / 3`.

To work around it we can:
1. Increase poll time or spin waiting for result - expensive?
1. Repair the whole keyspace - in that case progress UX and resume speed suffer - slow updates and need to redo killed repairs

Suggested API changes:

Extend `/storage_service/repair_async/{keyspace}` endpoint with `int wait` parameter that describes how many seconds to wait for repair result.

asias:
This new api blocks until the repair job is either finished or failed.

E.g., curl -X GET http://127.0.0.1:10000/storage_service/repair_status/?id=123

The current asynchronous api returns immediately even if the repair is in progress.

E.g., curl -X GET http://127.0.0.1:10000/storage_service/repair_async/ks?id=123

User can use the new synchronous API to avoid keep sending the query to
poll if the repair job is finished.

Fixes #6445

### DIFF
--- a/api/api-doc/storage_service.json
+++ b/api/api-doc/storage_service.json
@@ -834,6 +834,43 @@
          ]
       },
       {
+         "path":"/storage_service/repair_status/",
+         "operations":[
+            {
+               "method":"GET",
+               "summary":"Query the repair status and return when the repair is finished or timeout",
+               "type":"string",
+               "enum":[
+                  "RUNNING",
+                  "SUCCESSFUL",
+                  "FAILED"
+               ],
+               "nickname":"repair_await_completion",
+               "produces":[
+                  "application/json"
+               ],
+               "parameters":[
+                  {
+                     "name":"id",
+                     "description":"The repair ID to check for status",
+                     "required":true,
+                     "allowMultiple":false,
+                     "type": "long",
+                     "paramType":"query"
+                  },
+                  {
+                     "name":"timeout",
+                     "description":"Seconds to wait before the query returns even if the repair is not finished. The value -1 or not providing this parameter means no timeout",
+                     "required":false,
+                     "allowMultiple":false,
+                     "type": "long",
+                     "paramType":"query"
+                  }
+               ]
+            }
+         ]
+      },
+      {
          "path":"/storage_service/repair_async/{keyspace}",
          "operations":[
             {
--- a/api/storage_service.cc
+++ b/api/storage_service.cc
@@ -403,6 +403,39 @@ void set_storage_service(http_context& ctx, routes& r) {
         });
     });
 
+    ss::repair_await_completion.set(r, [&ctx](std::unique_ptr<request> req) {
+        int id;
+        using clock = std::chrono::steady_clock;
+        clock::time_point expire;
+        try {
+            id = boost::lexical_cast<int>(req->get_query_param("id"));
+            // If timeout is not provided, it means no timeout.
+            sstring s = req->get_query_param("timeout");
+            int64_t timeout = s.empty() ? int64_t(-1) : boost::lexical_cast<int64_t>(s);
+            if (timeout < 0 && timeout != -1) {
+                return make_exception_future<json::json_return_type>(
+                        httpd::bad_param_exception("timeout can only be -1 (means no timeout) or non negative integer"));
+            }
+            if (timeout < 0) {
+                expire = clock::time_point::max();
+            } else {
+                expire = clock::now() + std::chrono::seconds(timeout);
+            }
+        } catch (std::exception& e) {
+            return make_exception_future<json::json_return_type>(httpd::bad_param_exception(e.what()));
+        }
+        return repair_await_completion(ctx.db, id, expire)
+                .then_wrapped([] (future<repair_status>&& fut) {
+            ss::ns_repair_async_status::return_type_wrapper res;
+            try {
+                res = fut.get0();
+            } catch (std::exception& e) {
+                return make_exception_future<json::json_return_type>(httpd::server_error_exception(e.what()));
+            }
+            return make_ready_future<json::json_return_type>(json::json_return_type(res));
+        });
+    });
+
     ss::force_terminate_all_repair_sessions.set(r, [](std::unique_ptr<request> req) {
         return repair_abort_all(service::get_local_storage_service().db()).then([] {
             return make_ready_future<json::json_return_type>(json_void());
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -266,6 +266,7 @@ void tracker::done(int id, bool succeeded) {
     } else {
         _status[id] = repair_status::FAILED;
     }
+    _done_cond.broadcast();
 }
 repair_status tracker::get(int id) {
     if (id >= _next_repair_command) {
@@ -277,6 +278,30 @@ repair_status tracker::get(int id) {
     } else {
         return it->second;
     }
+}
+
+future<repair_status> tracker::repair_await_completion(int id, std::chrono::steady_clock::time_point timeout) {
+    return seastar::with_gate(_gate, [this, id, timeout] {
+        if (id >= _next_repair_command) {
+            return make_exception_future<repair_status>(std::runtime_error(format("unknown repair id {:d}", id)));
+        }
+        return repeat_until_value([this, id, timeout] {
+            auto it = _status.find(id);
+            if (it == _status.end()) {
+                return make_ready_future<std::optional<repair_status>>(repair_status::SUCCESSFUL);
+            } else {
+                if (it->second == repair_status::FAILED) {
+                    return make_ready_future<std::optional<repair_status>>(repair_status::FAILED);
+                } else {
+                    return _done_cond.wait(timeout).then([] {
+                        return make_ready_future<std::optional<repair_status>>(std::nullopt);
+                    }).handle_exception_type([] (condition_variable_timed_out&) {
+                        return make_ready_future<std::optional<repair_status>>(repair_status::RUNNING);
+                    });
+                }
+            }
+        });
+    });
 }
 
 int tracker::next_repair_command() {
@@ -1542,6 +1567,12 @@ future<std::vector<int>> get_active_repairs(seastar::sharded<database>& db) {
 future<repair_status> repair_get_status(seastar::sharded<database>& db, int id) {
     return db.invoke_on(0, [id] (database& localdb) {
         return repair_tracker().get(id);
+    });
+}
+
+future<repair_status> repair_await_completion(seastar::sharded<database>& db, int id, std::chrono::steady_clock::time_point timeout) {
+    return db.invoke_on(0, [id, timeout] (database& localdb) {
+        return repair_tracker().repair_await_completion(id, timeout);
     });
 }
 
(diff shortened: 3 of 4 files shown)


asias:
@mmatczuk FYI. I tries this with scyllarepair.py and it was trivial to switch to the new API. We can easily add timeout parameter if it is really needed by scylla manager.

api/api-doc/storage_service.json
```diff
@@ -833,6 +833,35 @@
...
       },
+      {
+         "path":"/storage_service/repair_status/",
+         "operations":[
+            {
+               "method":"GET",
+               "summary":"Query the repair status and return when the repair is finished",
+               "type":"string",
+               "enum":[
+                  "RUNNING",
```
avikivity, line 844: How can it ever return "RUNNING"?

However, I suggest to add a timeout parameter (and if it times out, return RUNNING). This is so that the caller can detect a node failure (but failing if it did not get a response after timeout+slack)

api/api-doc/storage_service.json
```diff
@@ -833,6 +833,35 @@
...
+               "produces":[
+                  "application/json"
+               ],
+               "parameters":[
+                  {
+                     "name":"id",
+                     "description":"The repair ID to check for status",
+                     "required":true,
+                     "allowMultiple":false,
+                     "type": "long",
```
avikivity, line 858: We should have made repair IDs UUIDs, not lonm but too late now.

What is returned if the repair is not recognized?  That can happen if the repair was completed before the request was sent.

api/storage_service.cc
```diff
@@ -403,6 +403,19 @@ void set_storage_service(http_context& ctx, routes& r) {
         });
     });
 
+    ss::repair_async_status_until_finished.set(r, [&ctx](std::unique_ptr<request> req) {
+        return repair_get_status_until_finished(ctx.db, boost::lexical_cast<int>(req->get_query_param("id")))
```
avikivity, line 407: int? Not a larger type?

api/api-doc/storage_service.json
avikivity: Maybe we should keep completed repairs for a while.

api/storage_service.cc
```diff
@@ -403,6 +403,19 @@ void set_storage_service(http_context& ctx, routes& r) {
...
+                .then_wrapped([] (future<repair_status>&& fut) {
+            ss::ns_repair_async_status::return_type_wrapper res;
+            try {
+                res = fut.get0();
+            } catch(std::runtime_error& e) {
```
avikivity, line 412: Space after `catch`.

avikivity: Why just runtime_error?

repair/repair.cc
```diff
@@ -266,6 +266,7 @@ void tracker::done(int id, bool succeeded) {
     } else {
         _status[id] = repair_status::FAILED;
     }
+    _done_cond.broadcast();
```
avikivity, line 269: Will this get triggered on shutdown?

Please test shutdown with this call running, and a repair running, in debug mode, to check for use-after-free.

avikivity: (and to make sure we return to the caller with a reasonable response).

api/api-doc/storage_service.json
asias: Because I have adding "timeout" parameter in mind: https://github.com/scylladb/scylla/pull/6743#issuecomment-652745063

api/storage_service.cc
asias: It is historic which follows C*. We currently only return an int as the repair ID anyway.

repair/repair.cc
asias: Yes. When shutdown happens, repair jobs will be aborted and tracker::done() will be called.

The caller will get a response saying the repair failed.

api/storage_service.cc
asias: It is a copied from the old ss::repair_async_status.set function. I will fix it.

api/api-doc/storage_service.json
asias: Yes, I though about using UUID for repair id as well. My idea is to keep the old int based id for the old api caller. Use uuid based id with new api. 

A repair id completed successfully will be removed. Only failed repairs are stored. So, a repair id is not recognized, it will be SUCCESSFUL. Currently, we have a very simple check to see if the id is a valid one by checking if the id is beyond the next_repair_id.

asias:
Changes in 8ce5981a217dc4999bff2612b4cd38535d0e3957:

- Add timeout_ms parameter
- Fix space around catch
- Catch more exception, not just runtime_error

asias:
@avikivity Ping. The timeout support is added. Can you please take a look?

api/api-doc/storage_service.json
avikivity: It may be successful, or non-existent.

api/api-doc/storage_service.json
```diff
@@ -833,6 +833,43 @@
...
+            {
+               "method":"GET",
+               "summary":"Query the repair status and return when the repair is finished or timeout",
+               "type":"string",
+               "enum":[
+                  "RUNNING",
+                  "SUCCESSFUL",
+                  "FAILED"
+               ],
+               "nickname":"repair_async_status_until_finished",
```
avikivity, line 848: why "async"? it's a synchronous API.

let's call it repair_await_completion.

api/storage_service.cc
```diff
@@ -403,6 +403,39 @@ void set_storage_service(http_context& ctx, routes& r) {
...
+            }
+            if (timeout_ms < 0) {
+                expire = clock::time_point::max();
+            } else {
+                expire = clock::now() + std::chrono::milliseconds(timeout_ms);
+            }
+        } catch (std::exception& e) {
+            return make_exception_future<json::json_return_type>(httpd::bad_param_exception(e.what()));
+        }
+        return repair_get_status_until_finished(ctx.db, id, expire)
```
avikivity, line 427: again, await_completion is more descriptive.

repair/repair.cc
```diff
@@ -279,6 +280,28 @@ repair_status tracker::get(int id) {
     }
 }
 
+future<repair_status> tracker::get_status_until_finished(int id, std::chrono::steady_clock::time_point timeout) {
```
avikivity, line 283: It's a mistake that we have to use steady_clock_type, but it's a seastar bug. We'll fix it later.

repair/repair.cc
```diff
@@ -279,6 +280,28 @@ repair_status tracker::get(int id) {
...
+    if (id >= _next_repair_command) {
+        return make_exception_future<repair_status>(std::runtime_error(format("unknown repair id {:d}", id)));
+    }
+    return repeat_until_value([this, id, timeout] {
+        auto it = _status.find(id);
+        if (it == _status.end()) {
+            return make_ready_future<std::optional<repair_status>>(repair_status::SUCCESSFUL);
```
avikivity, line 290: I don't like it much, but ok.

avikivity: If a node restarts, then a failed repair can become successful, no? We have to make repair results persistent.

api/storage_service.cc
```diff
@@ -403,6 +403,39 @@ void set_storage_service(http_context& ctx, routes& r) {
...
+                .then_wrapped([] (future<repair_status>&& fut) {
+            ss::ns_repair_async_status::return_type_wrapper res;
+            try {
+                res = fut.get0();
+            } catch (std::exception& e) {
+                return make_exception_future<json::json_return_type>(httpd::server_error_exception(e.what()));
+            }
+            return make_ready_future<json::json_return_type>(json::json_return_type(res));
+        });
+    });
```
avikivity, line 437: Suppose the API service is shut down. Will we have a use-after-free? Will it block shutdown?

repair/repair.cc
```diff
@@ -279,6 +280,28 @@ repair_status tracker::get(int id) {
...
+            } else {
+                return _done_cond.wait(timeout).then([] {
+                    return make_ready_future<std::optional<repair_status>>(std::nullopt);
+                }).handle_exception_type([] (condition_variable_timed_out&) {
+                    return make_ready_future<std::optional<repair_status>>(repair_status::RUNNING);
+                });
+            }
+        }
+    });
+}
```
avikivity, line 303: Again need to think about shutdown while this call is running. We should broadcast on the condition variable during shutdown and wait for all calls to return (using async_sharded_service).

api/api-doc/storage_service.json
asias: OK. I like this name.

repair/repair.cc
asias: I will think about it.

api/api-doc/storage_service.json
```diff
@@ -833,6 +833,43 @@
...
+                  {
+                     "name":"id",
+                     "description":"The repair ID to check for status",
+                     "required":true,
+                     "allowMultiple":false,
+                     "type": "long",
+                     "paramType":"query"
+                  },
+                  {
+                     "name":"timeout_ms",
```
mmatczuk, line 862: Setting ms timeouts on HTTP API is not very common, this could be easily seconds since if end state is reached it would return early anyway.

asias: Thanks for the feedback. I will make it seconds.

repair/repair.cc
asias: No. When a node restarts,  the next repair id is 1. Status queries of exiting repair repair id (which >=1) will fail with "unknown repair id " because repair knows no repair job is started yet.  

After we transit to uuid repair ids, we can store them in a system table.

api/storage_service.cc
asias: The repair service is shutodwn before http api service. 
When repair service shutdown(repair_shutdown()), it calls tracker::shutdown() which waits for _gate::close() and tracker::repair_await_completion runs under _gate. 
So after repair shutdowns, no pending api calls anymore.

repair/repair.cc
asias: See above.

asias:
Changes in f98349d21c93fe33fdeb577e5048aafb1a226082:
- Change tiemout to seconds from milliseconds
- Rename to repair_await_completion
- Use _gate to protect repair_await_completion

repair/repair.cc
avikivity: Ah. It's still vulnerable to to collisions, but that's the best we can do for now.

avikivity:
Looks good. Will wait for @mmatczuk to review the API.

asias:
> Looks good. Will wait for @mmatczuk to review the API.

@mmatczuk Can you please take a look? I hope we can use it in scylla manager soon ;-)

mmatczuk:
LGTM

Issue #6744, opened 2020-07-02
init - Startup failed: std::runtime_error (bootstrap_with_repair: keyspace=system_distributed, range=(9183073555191895134, 9196226903124807343], no existing node in local dc)


asias:
Consider a cluster with two nodes:
    
    - n1 (dc1)
    - n2 (dc2)
    
    A third node is bootstrapped:
    
    - n3 (dc2)
    
The n3 fails to bootstrap as follows:
    
[shard 0] init - Startup failed: std::runtime_error (bootstrap_with_repair: keyspace=system_distributed, range=(9183073555191895134, 9196226903124807343], no existing node in local dc)

slivne:
@asias is this a regression

what happens with user tables ?

which versions are effected

asias:
> @asias is this a regression
> 
> what happens with user tables ?
> 
> which versions are effected

Yes, it is a regression. It does matter it is a system or user table. Bootstrap a node with table using SimpleStrategy has this issue in case multiple DC are configured. Versions with repair based ops are affected.

asias:
Patch posted: https://github.com/scylladb/scylla/pull/6745

avikivity:
Backported to 4.2 (only version with repair based node ops enabled).

PR #6745, opened 2020-07-02
repair: Relax node selection in bootstrap when nodes are less than RF

asias:
Consider a cluster with two nodes:

 - n1 (dc1)
 - n2 (dc2)

A third node is bootstrapped:

 - n3 (dc2)

The n3 fails to bootstrap as follows:

 [shard 0] init - Startup failed: std::runtime_error
 (bootstrap_with_repair: keyspace=system_distributed,
 range=(9183073555191895134, 9196226903124807343], no existing node in
 local dc)

The system_distributed keyspace is using SimpleStrategy with RF 3. For
the keyspace that does not use NetworkTopologyStrategy, we should not
require the source node to be in the same DC.

Fixes: #6744
Backports: 4.0, 4.1, 4.2

### DIFF
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -1633,6 +1633,7 @@ future<> bootstrap_with_repair(seastar::sharded<database>& db, locator::token_me
             auto& ks = db.local().find_keyspace(keyspace_name);
             auto& strat = ks.get_replication_strategy();
             dht::token_range_vector desired_ranges = strat.get_pending_address_ranges(tm, tokens, myip);
+            bool find_node_in_local_dc_only = strat.get_type() == locator::replication_strategy_type::network_topology;
 
             //Active ranges
             auto metadata_clone = tm.clone_only_token_map();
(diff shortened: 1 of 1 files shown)


bhalevy:
@tgrabiec please review

avikivity:
Looks okay to me.

asias:
ping.

avikivity:
@asias I already noted I'm okay with it.

asias:
> @asias I already noted I'm okay with it.

Thanks. Avi! You are fast ;-)

bhalevy:
ping @tgrabiec can you please review and merge this pr?

avikivity:
> > @asias I already noted I'm okay with it.
> 
> Thanks. Avi! You are fast ;-)

You're confusing me with someone else maybe.

asias:
Merged in 38d964352da7b950abdadb5fc903e7105b3ca238.

asias:
> Relying on replication strategy type is sad, it breaks the abstraction.

For the type check: 

strat.get_type() == locator::replication_strategy_type::network_topology

Perhaps, we can hide it with abstract_replication_strategy::is_dc_aware().

Another usage is harder: 

```
1732                         auto get_rf_in_local_dc = [&] () {
1733                             size_t rf_in_local_dc = strat.get_replication_factor();
1734                             if (strat.get_type() == locator::replication_strategy_type::network_topology) {
1735                                 auto nts = dynamic_cast<locator::network_topology_strategy*>(&strat);
1736                                 if (!nts) {
1737                                     throw std::runtime_error(format("bootstrap_with_repair: keyspace={}, range={}, failed to cast to network_topology_strategy",
1738                                             keyspace_name, desired_range));
1739                                 }
1740                                 rf_in_local_dc = nts->get_replication_factor(local_dc);
1741                             }
1742                             return rf_in_local_dc;
1743                         };
```

The get_replication_factor(sstring dc) version is available for the network strategy only.

avikivity:
The abstraction is very poor in the first place.

Maybe a better way would be to make simple strategy erase the DC/rack property of a node, so it thinks that is running in a single DC and rack. All the other logic would remain the same between them (and could be moved out of the virtual functions).

See things like count_per_dc_endpoints and assure_sufficient_live_nodes_each_quorum. /cc @gleb-cloudius

PR #6746, opened 2020-07-02
token_metadata: Fix incorrect log in update_normal_tokens

asias:
Currently, when update_normal_tokens is called, a warning logged.

   Token X changing ownership from A to B

It is not correct to log so because we can call update_normal_tokens
against a temporary token_metadata object during topology calculation.

Refs: 6437


asias:
ping.

PR #6748, opened 2020-07-02
scylla-housekeeping: wrap urllib.request with try ... except

dkropachev:
We could get "cannot serialize '_io.BufferedReader' object" when request get 404 error from the server
Now you will get legit error message in the case.

Fixes - https://github.com/scylladb/scylla/issues/6690

### DIFF
--- a/dist/common/scripts/scylla-housekeeping
+++ b/dist/common/scripts/scylla-housekeeping
@@ -61,7 +61,15 @@ def sh_command(*args):
     return out
 
 def get_url(path):
-    return urllib.request.urlopen(path).read().decode('utf-8')
+    # If server returns any error, like 403, or 500 urllib.request throws exception, which is not serializable.
+    # When multiprocessing routines fail to serialize it, it throws ambiguous serialization exception
+    #   from get_json_from_url.
+    # In order to see legit error we catch it from the inside of process, covert to string and
+    #   pass it as part of return value
+    try:
+        return 0, urllib.request.urlopen(path).read().decode('utf-8')
+    except Exception as exc:
+        return 1, str(exc)
 
 def get_json_from_url(path):
     pool = mp.Pool(processes=1)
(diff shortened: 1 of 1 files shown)


avikivity:
Please add the "Fixes" annotation to the commit.

dkropachev:
> Please add the "Fixes" annotation to the commit.

Done.

dist/common/scripts/scylla-housekeeping
```diff
@@ -61,7 +61,10 @@ def sh_command(*args):
     return out
 
 def get_url(path):
-    return urllib.request.urlopen(path).read().decode('utf-8')
+    try:
+        return 0, urllib.request.urlopen(path).read().decode('utf-8')
+    except Exception as exc:
+        return 1, str(exc)
```
avikivity, line 67: Why are you returning the exception in such a strange way? I understand what you're doing, but not why you're using such a non-conventional way to do it.

dist/common/scripts/scylla-housekeeping
```diff
@@ -71,13 +74,16 @@ def get_json_from_url(path):
...
     try:
-        retval = result.get(timeout=5)
+        status, retval = result.get(timeout=5)
     except mp.TimeoutError as err:
         pool.terminate()
         pool.join()
         raise
+    if status == 1:
+        raise RuntimeError(f'Failed to get "{path}" due to the following error: {retval}')
     return json.loads(retval)
```
avikivity, line 84: Why not put an except clause here instead?

dist/common/scripts/scylla-housekeeping
fruch: the whole issue that this code it running in a different process and the exception in some cases isn't serializable
hence why it's converted to string, and pass back with some kind of status code

dist/common/scripts/scylla-housekeeping
fruch: as mentioned above, the whole reason for this fix, is an Exception which isn't serializable, hence except clause would be much use in that case.

dkropachev: @avikivity , in addition to what israel (@fruch) said, catchig exception there, you won't get legit error, instead of it you will get serialization error.

dist/common/scripts/scylla-housekeeping
avikivity: Please add a comment to explain it.

dkropachev: Done

Issue #6776, opened 2020-07-06
hinted handoff: memory leak caused by improper commitlog usage from hints manager


piodul:
HEAD: 36b6ee7b11b86aab26e5f69f6cde4ba089f43983 (happens on master as of writing this)

Refs: #6409

Currently, hints manager uses a `commitlog` object for hints storage. In order to manually trigger a flush and repopulate the list of hint segments written to disk, this object is recreated roughly every 10 seconds, like this:

https://github.com/scylladb/scylla/blob/de82b3efaec3a9f285e5e676d32d88c163cee262/db/hints/manager.cc#L354-L363

Unfortunately. the commitlog is not properly cleaned up - `shutdown()` is called, but `release()` is not called afterwards. Because of that, the [`segment_manager::_segments`](https://github.com/scylladb/scylla/blob/de82b3efaec3a9f285e5e676d32d88c163cee262/db/commitlog/commitlog.cc#L369) field is not cleared and references to some `segment`s might still be there. `segment`s hold a [shared pointer backreference to the owning `segment_manager`](https://github.com/scylladb/scylla/blob/de82b3efaec3a9f285e5e676d32d88c163cee262/db/commitlog/commitlog.cc#L439), which causes a reference cycle to occur, and prevents both `segment_manager` and `segment`s from being destroyed when `commitlog` is destroyed.

Found based on investigation in #6409 and by adding logging information about the global count of `segment_manager`s and `segment`s.

avikivity:
Backported to 4.0, 4.1, 4.2.

vladzcloudius:
> Backported to 4.0, 4.1, 4.2.

@avikivity 
Please, consider backporting to 3.0 and to the current Enterprise builds (2019.1.x).

avikivity:
3.0 is at end of life.

Don't discuss enterprise releases on the open-source tracker.

PR #6777, opened 2020-07-06
hinted handoff: fix commitlog memory leak

piodul:
When commitlog is recreated in hints manager, only shutdown() method is
called, but not release(). Because of that, some internal commitlog
objects (`segment_manager` and `segment`s) may be left pointing to each
other through shared_ptr reference cycles, which may result in memory
leak when the parent commitlog object is destroyed.

This PR prevents memory leaks that may happen this way by calling
release() after shutdown() from the hints manager.

Fixes: #6409, #6776

### DIFF
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -224,7 +224,9 @@ future<> manager::end_point_hints_manager::stop(drain should_drain) noexcept {
         with_lock(file_update_mutex(), [this] {
             if (_hints_store_anchor) {
                 hints_store_ptr tmp = std::exchange(_hints_store_anchor, nullptr);
-                return tmp->shutdown().finally([tmp] {});
+                return tmp->shutdown().finally([tmp] {
+                    return tmp->release();
+                }).finally([tmp] {});
             }
             return make_ready_future<>();
         }).handle_exception([&eptr] (auto e) { eptr = std::move(e); }).get();
@@ -352,7 +354,9 @@ future<> manager::end_point_hints_manager::flush_current_hints() noexcept {
         return futurize_invoke([this] {
             return with_lock(file_update_mutex(), [this]() -> future<> {
                 return get_or_load().then([] (hints_store_ptr cptr) {
-                    return cptr->shutdown();
+                    return cptr->shutdown().finally([cptr] {
+                        return cptr->release();
+                    }).finally([cptr] {});
                 }).then([this] {
                     // Un-hold the commitlog object. Since we are under the exclusive _file_update_mutex lock there are no
                     // other hints_store_ptr copies and this would destroy the commitlog shared value.
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -328,6 +328,10 @@ future<db::commitlog> manager::end_point_hints_manager::add_store() noexcept {
             // HH doesn't utilize the flow that benefits from reusing segments.
             // Therefore let's simply disable it to avoid any possible confusion.
             cfg.reuse_segments = false;
+            // HH leaves segments on disk after commitlog shutdown, and later reads
+            // them when commitlog is re-created. This is expected to happen regularly
+            // during standard HH workload, so no need to print a warning about it.
+            cfg.warn_about_segments_left_on_disk_after_shutdown = false;
 
             return commitlog::create_commitlog(std::move(cfg)).then([this] (commitlog l) {
                 // add_store() is triggered every time hint files are forcefully flushed to I/O (every hints_flush_period).
--- a/db/commitlog/commitlog.cc
+++ b/db/commitlog/commitlog.cc
@@ -521,7 +521,7 @@ public:
             _segment_manager->totals.total_size_on_disk -= size_on_disk();
             _segment_manager->totals.total_size -= (size_on_disk() + _buffer.size_bytes());
             _segment_manager->add_file_to_delete(_file_name, _desc);
-        } else {
+        } else if (_segment_manager->cfg.warn_about_segments_left_on_disk_after_shutdown) {
             clogger.warn("Segment {} is dirty and is left on disk.", *this);
         }
     }
--- a/db/commitlog/commitlog.hh
+++ b/db/commitlog/commitlog.hh
@@ -137,6 +137,7 @@ public:
 
         bool reuse_segments = true;
         bool use_o_dsync = false;
+        bool warn_about_segments_left_on_disk_after_shutdown = true;
 
         const db::extensions * extensions = nullptr;
     };


db/hints/manager.cc
```diff
@@ -224,7 +224,9 @@ future<> manager::end_point_hints_manager::stop(drain should_drain) noexcept {
         with_lock(file_update_mutex(), [this] {
             if (_hints_store_anchor) {
                 hints_store_ptr tmp = std::exchange(_hints_store_anchor, nullptr);
-                return tmp->shutdown().finally([tmp] {});
+                return tmp->shutdown().then([tmp] {
+                    return tmp->release();
```
vladzcloudius, line 228: AFAIU `release()` should not be needed if `shutdown()` completed successfully.
It IS needed otherwise however. Therefore it has to be in the `finally` clause - not `then`.

db/hints/manager.cc
```diff
@@ -352,7 +354,9 @@ future<> manager::end_point_hints_manager::flush_current_hints() noexcept {
         return futurize_invoke([this] {
             return with_lock(file_update_mutex(), [this]() -> future<> {
                 return get_or_load().then([] (hints_store_ptr cptr) {
-                    return cptr->shutdown();
+                    return cptr->shutdown().then([cptr] {
+                        return cptr->release();
```
vladzcloudius, line 362: The same here.

piodul:
First and foremost, a disclaimer: what I explain here is purely based on what I learned by reading commitlog code. I might have missed some things hidden between the lines, so @elcallio please correct if I write something wrong here.

----

@vladzcloudius

> Please, note that release() is not needed if shutdown() succeeds.
> That's why we don't see the memory leak in question all the time.

Unfortunately, that's not true, and the point of this patch is to fix this. What you wrote is what one would expect from the comments, but there is more nuance to the situation. What `release()` does that `shutdown()` does not is clearing the `_segments` field of `segment_manager`. Because those segments keep shared_ptr backreferences to the `segment_manager`, this causes both types of objects not to be cleared because of the shared_ptr cycle.

> Therefore I think the proper way to fix this was not to patch HH but rather patch commitlog::shutdown() and make it call commitlog::release() when there was any exception/error inside it.

> This is of course unless there is a VERY good reason not to (which I don't see at the moment).

> It's a very confusing API when you need to call "object::stop()" and then "object::proper_stop()".
> The caller should not care about the internal mechanics of the "object::stop()" - it should know that if "object::stop()" completed - the object can be safely destroyed.

You have a point about the API. I made this patch as it looks like now because it was very simple to do, and it mimics the way how commitlog destruction is handled in database.cc ([link for context](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/database.cc#L1771-L1779)).

Now that I look at the [original patch](https://groups.google.com/g/scylladb-dev/c/4oTlwctxZv0/m/QxxxZRVcDAAJ) that added the `release()` invocation in database.cc, there was an argument against doing only `shutdown()` instead of using `shutdown()` and then `release()`, but I don't think it applies now.

@elcallio, is that correct? Would it be OK to add `_segments.clear()` inside `segment_manager::shutdown`? If that's OK, I'm all for fixing `shutdown()` instead of calling `release()` later.

----

> We know for a fact that this message is not printed all time when HH is working and that HH code handles rp_handles the way that should prevent these messages (see manager::end_point_hints_manager::store_hint()).

Regarding the warning message, it does not currently appear because of the memory leak I described. The warning is issued in `segment`'s destructor, and the memory leak prevents them from running.

> Namely we DO "unmark" the position in the file.
> If you see these messages something is definitely not right.

Let me clarify what I meant about unmarking. Actually, I think I got confused and thought that rp_handle used its `replay_position _rp` field to remove an element of some map, but it doesn't, so let me rephrase what I meant without using the "marked position" terminology.

Each commitlog segment contains a map that maps table IDs into numbers (_cf_dirty). Such number is incremented when a mutation to that table is stored ([here](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/db/commitlog/commitlog.cc#L891)), and can be decremented on rp_handle destruction ([here](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/db/commitlog/commitlog.cc#L2243-L2247)). rp_handle also has a release() method that allows to lift responsibility from the rp_handle object over decrementing the counter.

A commitlog segment is dirty if there are some non-zero numbers in _cf_dirty.

In case of hints, release() is called ([here](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/db/hints/manager.cc#L172)) but the count is not released manually later. I thought this is by design because it leaves "dirty" commitlog segments, which - contrary to "clean" segments - are not automatically deleted from disk after being filled up and flushed. We need this so that those segments are later available to be replayed when the commitlog is re-created.

The warning about the dirty segments was not printed because of the memory leak. Now that the memory leak is fixed, we need to disable the warning for hints commitlog because dirty segments are something normal in this case.

----

(Addressing a fragment of your [comment](https://github.com/scylladb/scylla/issues/6409#issuecomment-654455101))

> Note that we intentionally (as of now) allow some segments of the current commitlog object to be alive after we shut it down because they may still be written while we do that and we don't want to block on that (see how store_hint()) is implemented.

I don't see or understand this mechanism. Could you clarify what you mean by "alive" here? Wouldn't it be wrong if some writes or flushes are happening to any commitlog segment after the corresponding `commitlog::shutdown()` returns?

All operations on the hints commitlog object, excluding replaying commitlog segments, are protected with `file_update_mutex()`, including both `flush_current_hints` and `store_hint`. What I'd expect from the implementation (and, if I understand correctly, this is what happens now), `commitlog::shutdown()` stops admitting new writes and waits until all segments are flushed.

vladzcloudius:
> First and foremost, a disclaimer: what I explain here is purely based on what I learned by reading commitlog code. I might have missed some things hidden between the lines, so @elcallio please correct if I write something wrong here.
> 
> @vladzcloudius
> 
> > Please, note that release() is not needed if shutdown() succeeds.
> > That's why we don't see the memory leak in question all the time.
> 
> Unfortunately, that's not true, and the point of this patch is to fix this. What you wrote is what one would expect from the comments, but there is more nuance to the situation. What `release()` does that `shutdown()` does not is clearing the `_segments` field of `segment_manager`. Because those segments keep shared_ptr backreferences to the `segment_manager`, this causes both types of objects not to be cleared because of the shared_ptr cycle.

Unless I'm missing something after the successful call for `shutdown` the _segments should be empty: commitlog::shutdown() -> segments_manager::shutdown() -> discard_unused_segments() 

However if I DO miss something it seem to me that the correct way would be to call `commitlog::release()` in the commitlog destructor (like was suggested by Calle on that patch thread that you mentioned below but for some reason was never done).

> 
> > Therefore I think the proper way to fix this was not to patch HH but rather patch commitlog::shutdown() and make it call commitlog::release() when there was any exception/error inside it.
> 
> > This is of course unless there is a VERY good reason not to (which I don't see at the moment).
> 
> > It's a very confusing API when you need to call "object::stop()" and then "object::proper_stop()".
> > The caller should not care about the internal mechanics of the "object::stop()" - it should know that if "object::stop()" completed - the object can be safely destroyed.
> 
> You have a point about the API. I made this patch as it looks like now because it was very simple to do, and it mimics the way how commitlog destruction is handled in database.cc ([link for context](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/database.cc#L1771-L1779)).
> 
> Now that I look at the [original patch](https://groups.google.com/g/scylladb-dev/c/4oTlwctxZv0/m/QxxxZRVcDAAJ) that added the `release()` invocation in database.cc, there was an argument against doing only `shutdown()` instead of using `shutdown()` and then `release()`, but I don't think it applies now.

That ^^ thread  ;)

> 
> @elcallio, is that correct? Would it be OK to add `_segments.clear()` inside `segment_manager::shutdown`? If that's OK, I'm all for fixing `shutdown()` instead of calling `release()` later.
> 
> > We know for a fact that this message is not printed all time when HH is working and that HH code handles rp_handles the way that should prevent these messages (see manager::end_point_hints_manager::store_hint()).
> 
> Regarding the warning message, it does not currently appear because of the memory leak I described. The warning is issued in `segment`'s destructor, and the memory leak prevents them from running.

I'm still curious how this may happen in a regular case. Why segments_manager::discard_unused_segments() doesn't clear _segments?

> 
> > Namely we DO "unmark" the position in the file.
> > If you see these messages something is definitely not right.
> 
> Let me clarify what I meant about unmarking. Actually, I think I got confused and thought that rp_handle used its `replay_position _rp` field to remove an element of some map, but it doesn't, so let me rephrase what I meant without using the "marked position" terminology.
> 
> Each commitlog segment contains a map that maps table IDs into numbers (_cf_dirty). Such number is incremented when a mutation to that table is stored ([here](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/db/commitlog/commitlog.cc#L891)), and can be decremented on rp_handle destruction ([here](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/db/commitlog/commitlog.cc#L2243-L2247)). rp_handle also has a release() method that allows to lift responsibility from the rp_handle object over decrementing the counter.
> 
> A commitlog segment is dirty if there are some non-zero numbers in _cf_dirty.
> 
> In case of hints, release() is called ([here](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/db/hints/manager.cc#L172)) but the count is not released manually later. I thought this is by design because it leaves "dirty" commitlog segments, which - contrary to "clean" segments - are not automatically deleted from disk after being filled up and flushed. We need this so that those segments are later available to be replayed when the commitlog is re-created.
> 
> The warning about the dirty segments was not printed because of the memory leak. Now that the memory leak is fixed, we need to disable the warning for hints commitlog because dirty segments are something normal in this case.

I see your point now. We should be seeing this warning all the time indeed.
We need to understand why `shutdown()` doesn't cleanup _segments to begin with.
And then make sure that `release()` is always called from either `shutdown()` or from the commitlog's destructor.

> 
> (Addressing a fragment of your [comment](https://github.com/scylladb/scylla/issues/6409#issuecomment-654455101))
> 
> > Note that we intentionally (as of now) allow some segments of the current commitlog object to be alive after we shut it down because they may still be written while we do that and we don't want to block on that (see how store_hint()) is implemented.
> 
> I don't see or understand this mechanism. Could you clarify what you mean by "alive" here? Wouldn't it be wrong if some writes or flushes are happening to any commitlog segment after the corresponding `commitlog::shutdown()` returns?

I think it would. Like I wrote - I'd expect that after I call a shutdown() and the returned future resolves I should be able to safely destroy the object. However we see that many of supposedly obvious assumptions about the commitlog API don't hold. ;)
At the end of the day it's up to the commitlog implementation. There are different ways to implement things. What's important is that everybody understands the API. 
Documenting it usually helps but the more intuitive the API - the better.

I still think that the intent of the author was that calling the shutdown() should be enough unless it fails for any reason. 

I do see that commitlog::release() is being called unconditionally in the database::stop() flow  but I believe it's meant to handle bad case I mentioned above.

However according to you it seems that there is bug lurking in the commitlog code which somehow prevents shutdown() from clearing the _segments array.

> 
> All operations on the hints commitlog object, excluding replaying commitlog segments, are protected with `file_update_mutex()`, including both `flush_current_hints` and `store_hint`. What I'd expect from the implementation (and, if I understand correctly, this is what happens now), `commitlog::shutdown()` stops admitting new writes and waits until all segments are flushed.

You are right. It's actually already synchronized. I got confused (it was too long ago).
This means that it's already been taken care of and all what's left is to understand why commitlog::shutdown() doesn't do what it's supposed to do. ;)

vladzcloudius:
> > First and foremost, a disclaimer: what I explain here is purely based on what I learned by reading commitlog code. I might have missed some things hidden between the lines, so @elcallio please correct if I write something wrong here.
> > @vladzcloudius
> > > Please, note that release() is not needed if shutdown() succeeds.
> > > That's why we don't see the memory leak in question all the time.
> > 
> > 
> > Unfortunately, that's not true, and the point of this patch is to fix this. What you wrote is what one would expect from the comments, but there is more nuance to the situation. What `release()` does that `shutdown()` does not is clearing the `_segments` field of `segment_manager`. Because those segments keep shared_ptr backreferences to the `segment_manager`, this causes both types of objects not to be cleared because of the shared_ptr cycle.
> 
> @piodul
> 
> Unless I'm missing something after the successful call for `shutdown` the _segments should be empty: commitlog::shutdown() -> segments_manager::shutdown() -> discard_unused_segments()
> 
> However if I DO miss something it seem to me that the correct way would be to call `commitlog::release()` in the commitlog destructor (like was suggested by Calle on that patch thread that you mentioned below but for some reason was never done).
> 
> > > Therefore I think the proper way to fix this was not to patch HH but rather patch commitlog::shutdown() and make it call commitlog::release() when there was any exception/error inside it.
> > 
> > 
> > > This is of course unless there is a VERY good reason not to (which I don't see at the moment).
> > 
> > 
> > > It's a very confusing API when you need to call "object::stop()" and then "object::proper_stop()".
> > > The caller should not care about the internal mechanics of the "object::stop()" - it should know that if "object::stop()" completed - the object can be safely destroyed.
> > 
> > 
> > You have a point about the API. I made this patch as it looks like now because it was very simple to do, and it mimics the way how commitlog destruction is handled in database.cc ([link for context](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/database.cc#L1771-L1779)).
> > Now that I look at the [original patch](https://groups.google.com/g/scylladb-dev/c/4oTlwctxZv0/m/QxxxZRVcDAAJ) that added the `release()` invocation in database.cc, there was an argument against doing only `shutdown()` instead of using `shutdown()` and then `release()`, but I don't think it applies now.
> 
> That ^^ thread ;)
> 
> > @elcallio, is that correct? Would it be OK to add `_segments.clear()` inside `segment_manager::shutdown`? If that's OK, I'm all for fixing `shutdown()` instead of calling `release()` later.
> > > We know for a fact that this message is not printed all time when HH is working and that HH code handles rp_handles the way that should prevent these messages (see manager::end_point_hints_manager::store_hint()).
> > 
> > 
> > Regarding the warning message, it does not currently appear because of the memory leak I described. The warning is issued in `segment`'s destructor, and the memory leak prevents them from running.
> 
> I'm still curious how this may happen in a regular case. Why segments_manager::discard_unused_segments() doesn't clear _segments?
> 
> > > Namely we DO "unmark" the position in the file.
> > > If you see these messages something is definitely not right.
> > 
> > 
> > Let me clarify what I meant about unmarking. Actually, I think I got confused and thought that rp_handle used its `replay_position _rp` field to remove an element of some map, but it doesn't, so let me rephrase what I meant without using the "marked position" terminology.
> > Each commitlog segment contains a map that maps table IDs into numbers (_cf_dirty). Such number is incremented when a mutation to that table is stored ([here](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/db/commitlog/commitlog.cc#L891)), and can be decremented on rp_handle destruction ([here](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/db/commitlog/commitlog.cc#L2243-L2247)). rp_handle also has a release() method that allows to lift responsibility from the rp_handle object over decrementing the counter.
> > A commitlog segment is dirty if there are some non-zero numbers in _cf_dirty.
> > In case of hints, release() is called ([here](https://github.com/scylladb/scylla/blob/2d63acdd6aaf00c4c8bf979ad332776e532d2f56/db/hints/manager.cc#L172)) but the count is not released manually later. I thought this is by design because it leaves "dirty" commitlog segments, which - contrary to "clean" segments - are not automatically deleted from disk after being filled up and flushed. We need this so that those segments are later available to be replayed when the commitlog is re-created.
> > The warning about the dirty segments was not printed because of the memory leak. Now that the memory leak is fixed, we need to disable the warning for hints commitlog because dirty segments are something normal in this case.
> 
> I see your point now. We should be seeing this warning all the time indeed.
> We need to understand why `shutdown()` doesn't cleanup _segments to begin with.
> And then make sure that `release()` is always called from either `shutdown()` or from the commitlog's destructor.
> 
> > (Addressing a fragment of your [comment](https://github.com/scylladb/scylla/issues/6409#issuecomment-654455101))
> > > Note that we intentionally (as of now) allow some segments of the current commitlog object to be alive after we shut it down because they may still be written while we do that and we don't want to block on that (see how store_hint()) is implemented.
> > 
> > 
> > I don't see or understand this mechanism. Could you clarify what you mean by "alive" here? Wouldn't it be wrong if some writes or flushes are happening to any commitlog segment after the corresponding `commitlog::shutdown()` returns?
> 
> I think it would. Like I wrote - I'd expect that after I call a shutdown() and the returned future resolves I should be able to safely destroy the object. However we see that many of supposedly obvious assumptions about the commitlog API don't hold. ;)
> At the end of the day it's up to the commitlog implementation. There are different ways to implement things. What's important is that everybody understands the API.
> Documenting it usually helps but the more intuitive the API - the better.
> 
> I still think that the intent of the author was that calling the shutdown() should be enough unless it fails for any reason.
> 
> I do see that commitlog::release() is being called unconditionally in the database::stop() flow but I believe it's meant to handle bad case I mentioned above.
> 
> However according to you it seems that there is bug lurking in the commitlog code which somehow prevents shutdown() from clearing the _segments array.
> 
> > All operations on the hints commitlog object, excluding replaying commitlog segments, are protected with `file_update_mutex()`, including both `flush_current_hints` and `store_hint`. What I'd expect from the implementation (and, if I understand correctly, this is what happens now), `commitlog::shutdown()` stops admitting new writes and waits until all segments are flushed.
> 
> You are right. It's actually already synchronized. I got confused (it was too long ago).
> This means that it's already been taken care of and all what's left is to understand why commitlog::shutdown() doesn't do what it's supposed to do. ;)

@piodul 
I think I figured why `discard_unused_segments()` didn't clean _segments - I misunderstood the remove_if() semantics.

vladzcloudius:
@piodul getting back to the original series. 
Unless you decide to patch commitlog - please, address my initial comments about using `finally` instead of `then`.

Except for that - looks good.

piodul:
@vladzcloudius I'm glad everything is clear!

I agree with your remarks and I have applied your suggestions. I also reworded the description of the second commit so that it is more correct.

I rebased the PR, too.

piodul:
@elcallio could you take a look?

I'd also like to ask you about the `shutdown()` method - should we change it so that it clears the `_segments` field? I agree with @vladzcloudius that it would make the API nicer (see this [comment](https://github.com/scylladb/scylla/pull/6777#pullrequestreview-443353684)).

(If so, I'll change this patch again so that `release()` is called inside `handle_exception`, not `finally`)

vladzcloudius:
> @elcallio could you take a look?
> 
> I'd also like to ask you about the `shutdown()` method - should we change it so that it clears the `_segments` field? I agree with @vladzcloudius that it would make the API nicer (see this [comment](https://github.com/scylladb/scylla/pull/6777#pullrequestreview-443353684)).
> 
> (If so, I'll change this patch again so that `release()` is called inside `handle_exception`, not `finally`)

@piodul 
If @elcallio Agrees then you should 
1) Make shutdown() noexcept (rework it if needed in order to achieve it).
2) Call `release()` in the .finally() as the last thing in the `shutdown()`.

The `shutdown()` should be self-sufficient and not require any external extra calls if/when it fails.
That would be a proper API.

piodul:
I just noticed that "Backport candidate" label was not added automatically, so I added it manually.

The bug that this patch fixes was present at least since 3.2.4 (as reported in the issue), most likely even longer.

vladzcloudius:
> I just noticed that "Backport candidate" label was not added automatically, so I added it manually.
> 
> The bug that this patch fixes was present at least since 3.2.4 (as reported in the issue), most likely even longer.

This bug was present in scylla since HH has been introduced.

avikivity:
@piodul backport candidate labels are added on the issue, not pull request.

PR #6781, opened 2020-07-06
Shard aware: Support of source port based load balancing

fruch:
new configuration `cpu_sharding_algorithm_name` that support one of the two:

* `biased-token-round-robin` (default) -
  seastar `load_balancing_algorithm::connection_distribution` would be used
  for CQL connections

* `biased-token-source-port` -
  seastar `load_balancing_algorithm::port` would be used for CQL connections

example in scylla-driver (python): scylladb/python-driver#54

### DIFF
--- a/db/config.cc
+++ b/db/config.cc
@@ -714,6 +714,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
     , enable_dangerous_direct_import_of_cassandra_counters(this, "enable_dangerous_direct_import_of_cassandra_counters", value_status::Used, false, "Only turn this option on if you want to import tables from Cassandra containing counters, and you are SURE that no counters in that table were created in a version earlier than Cassandra 2.1."
         " It is not enough to have ever since upgraded to newer versions of Cassandra. If you EVER used a version earlier than 2.1 in the cluster where these SSTables come from, DO NOT TURN ON THIS OPTION! You will corrupt your data. You have been warned.")
     , enable_shard_aware_drivers(this, "enable_shard_aware_drivers", value_status::Used, true, "Enable native transport drivers to use connection-per-shard for better performance")
+    , cpu_sharding_algorithm_name(this, "cpu_sharding_algorithm_name", value_status::Used, "biased-token-round-robin", "Select the type of connection-per-shard algorithm")
     , enable_ipv6_dns_lookup(this, "enable_ipv6_dns_lookup", value_status::Used, false, "Use IPv6 address resolution")
     , abort_on_internal_error(this, "abort_on_internal_error", liveness::LiveUpdate, value_status::Used, false, "Abort the server instead of throwing exception when internal invariants are violated")
     , max_partition_key_restrictions_per_query(this, "max_partition_key_restrictions_per_query", liveness::LiveUpdate, value_status::Used, 100,
--- a/transport/server.cc
+++ b/transport/server.cc
@@ -224,6 +224,9 @@ cql_server::listen(socket_address addr, std::shared_ptr<seastar::tls::credential
     }
     return f.then([this, addr, keepalive](shared_ptr<seastar::tls::server_credentials> creds) {
         listen_options lo;
+        if (_config.cpu_sharding_algorithm_name == "biased-token-source-port"){
+            lo.lba = server_socket::load_balancing_algorithm::port;
+        }
         lo.reuse_address = true;
         server_socket ss;
         try {
@@ -1226,7 +1229,7 @@ std::unique_ptr<cql_server::response> cql_server::connection::make_supported(int
     if (_server._config.allow_shard_aware_drivers) {
         opts.insert({"SCYLLA_SHARD", format("{:d}", this_shard_id())});
         opts.insert({"SCYLLA_NR_SHARDS", format("{:d}", smp::count)});
-        opts.insert({"SCYLLA_SHARDING_ALGORITHM", dht::cpu_sharding_algorithm_name()});
+        opts.insert({"SCYLLA_SHARDING_ALGORITHM",  _server._config.cpu_sharding_algorithm_name});
         opts.insert({"SCYLLA_SHARDING_IGNORE_MSB", format("{:d}", _server._config.sharding_ignore_msb)});
         opts.insert({"SCYLLA_PARTITIONER", _server._config.partitioner_name});
     }
--- a/db/config.hh
+++ b/db/config.hh
@@ -299,6 +299,7 @@ public:
     named_value<bool> enable_sstables_mc_format;
     named_value<bool> enable_dangerous_direct_import_of_cassandra_counters;
     named_value<bool> enable_shard_aware_drivers;
+    named_value<sstring> cpu_sharding_algorithm_name;
     named_value<bool> enable_ipv6_dns_lookup;
     named_value<bool> abort_on_internal_error;
     named_value<uint32_t> max_partition_key_restrictions_per_query;
--- a/dht/token-sharding.hh
+++ b/dht/token-sharding.hh
@@ -25,10 +25,6 @@
 
 namespace dht {
 
-inline sstring cpu_sharding_algorithm_name() {
-    return "biased-token-round-robin";
-}
-
 std::vector<uint64_t> init_zero_based_shard_start(unsigned shards, unsigned sharding_ignore_msb_bits);
 
 unsigned shard_of(unsigned shard_count, unsigned sharding_ignore_msb_bits, const token& t);
--- a/transport/controller.cc
+++ b/transport/controller.cc
@@ -71,6 +71,7 @@ future<> controller::do_start_server() {
         cql_server_config.max_request_size = service::get_local_storage_service()._service_memory_total;
         cql_server_config.get_service_memory_limiter_semaphore = [ss = std::ref(service::get_storage_service())] () -> semaphore& { return ss.get().local()._service_memory_limiter; };
         cql_server_config.allow_shard_aware_drivers = cfg.enable_shard_aware_drivers();
+        cql_server_config.cpu_sharding_algorithm_name = cfg.cpu_sharding_algorithm_name();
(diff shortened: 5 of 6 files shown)


transport/server.cc
```diff
@@ -224,6 +224,9 @@ cql_server::listen(socket_address addr, std::shared_ptr<seastar::tls::credential
     }
     return f.then([this, addr, keepalive](shared_ptr<seastar::tls::server_credentials> creds) {
         listen_options lo;
+        if (_config.cpu_sharding_algorithm_name == "biased-token-source-port"){
```
avikivity, line 227: Space before {

db/config.cc
```diff
@@ -714,6 +714,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
     , enable_dangerous_direct_import_of_cassandra_counters(this, "enable_dangerous_direct_import_of_cassandra_counters", value_status::Used, false, "Only turn this option on if you want to import tables from Cassandra containing counters, and you are SURE that no counters in that table were created in a version earlier than Cassandra 2.1."
         " It is not enough to have ever since upgraded to newer versions of Cassandra. If you EVER used a version earlier than 2.1 in the cluster where these SSTables come from, DO NOT TURN ON THIS OPTION! You will corrupt your data. You have been warned.")
     , enable_shard_aware_drivers(this, "enable_shard_aware_drivers", value_status::Used, true, "Enable native transport drivers to use connection-per-shard for better performance")
+    , cpu_sharding_algorithm_name(this, "cpu_sharding_algorithm_name", value_status::Used, "biased-token-round-robin", "Select the type of connection-per-shard algorithm")
```
avikivity, line 717: Need to explain the options, and to validate them. If the user selects xyz, everything will break.

avikivity:
It's less-than-perfect that this is a configuration item. It's crucial for performance, but users won't know about it.

If we do make it an option, it should be live updateable (not easy) to avoid the need to restart the server.

haaawk:
Will this fix https://github.com/scylladb/scylla/issues/5239?

transport/server.cc
haaawk, line 227: Is it necessary to compare strings on every new connection? Is there any reason not to create a bool variable computed once and check it instead?

haaawk: Or maybe use enum instead of streams?

fruch: I'll look into how convert it internally to an enum

haaawk: Even better just have a variable in cql_server of type load_palancing_algorithm, set it once, and then just use it to initialize lo.lba.

db/config.cc
fruch: I'll try look on other ones that are verified as part of config.cc

fruch:
> It's less-than-perfect that this is a configuration item. It's crucial for performance, but users won't know about it.
> 
> If we do make it an option, it should be live updateable (not easy) to avoid the need to restart the server.

Any idea how can automatically do that ? without breaking older drivers that doesn't support this option ? figuring it by the client, but we can have multiple clients.

As for the updateble, I assume it should effect new connection from that point. and not close ongoing connections, right ?

haaawk:
> > It's less-than-perfect that this is a configuration item. It's crucial for performance, but users won't know about it.
> > If we do make it an option, it should be live updateable (not easy) to avoid the need to restart the server.
> 
> Any idea how can automatically do that ? without breaking older drivers that doesn't support this option ? figuring it by the client, but we can have multiple clients.
> 
> As for the updateble, I assume it should effect new connection from that point. and not close ongoing connections, right ?

We could introduce our own protocol version and then on the server switch to load balancing based on ports after we receive the first connection which uses this new protocol version. The problem is that old and new clients can't be mixed (work on the same cluster).

fruch:
> > > It's less-than-perfect that this is a configuration item. It's crucial for performance, but users won't know about it.
> > > If we do make it an option, it should be live updateable (not easy) to avoid the need to restart the server.
> > 
> > 
> > Any idea how can automatically do that ? without breaking older drivers that doesn't support this option ? figuring it by the client, but we can have multiple clients.
> > As for the updateble, I assume it should effect new connection from that point. and not close ongoing connections, right ?
> 
> We could introduce our own protocol version and then on the server switch to load balancing based on ports after we receive the first connection which uses this new protocol version. The problem is that old and new clients can't be mixed (work on the same cluster).

Also with this setup in configuration, supporting mixed clients is broken in same way.
But at least you can control and revert on the server, to something supported by current drivers

slivne:
I think we need to find a way this will work with both type of clients, we can't merge this as is - clients of "older" mode will keep on attempting to open connections and fail and fail.

We should also support both modes in parallel - think of a deployment that attempts to be upgraded from one mode to the other. they can't stop all their traffic they want todo a rolling restart and have traffic ongoing. 

An "ugly" solution would be to return back a new port scylla will be listening to for clients that want to use source port for sharding. Maybe there is a better idea than that.

fruch:
> I think we need to find a way this will work with both type of clients, we can't merge this as is - clients of "older" mode will keep on attempting to open connections and fail and fail.
> 
this is exactly the reason why to put it under a configuration flag, and it won't be the default
also they won't fail, but it would take them long to get to the correct shard, 
regular drivers won't be affected, but the distribution sine it won't be round-robin.

> We should also support both modes in parallel - think of a deployment that attempts to be upgraded from one mode to the other. they can't stop all their traffic they want todo a rolling restart and have traffic ongoing.
> 
If we'll do this as a live config, it can enabled in runtime after all new client are up, and it would start working from the next connection that would be opened

> An "ugly" solution would be to return back a new port scylla will be listening to for clients that want to use source port for sharding. Maybe there is a better idea than that.

it's ugly, and client code might need to change. I don't think we'll want to change port magically, if the user ask for specific ones.
other ideas, would need some kind of handshake, which gets us to square one (if we have this kind of handshake, we could select the shard with it, couldn't we ?)

slivne:
On Wed, Jul 8, 2020 at 6:43 PM Israel Fruchter <notifications@github.com>
wrote:

> I think we need to find a way this will work with both type of clients, we
> can't merge this as is - clients of "older" mode will keep on attempting to
> open connections and fail and fail.
>
> this is exactly the reason why to put it under a configuration flag, and
> it won't be the default
> also they won't fail, but it would take them long to get to the correct
> shard,
> regular drivers won't be affected, but the distribution sine it won't be
> round-robin.
>
they may fail connecting for a longer period - they will keep on attempting
to open connections thinking they are round robin which they aren't
multiple applications connecting at the same time will probably make this
worse.

We should also support both modes in parallel - think of a deployment that
> attempts to be upgraded from one mode to the other. they can't stop all
> their traffic they want todo a rolling restart and have traffic ongoing.
>
> If we'll do this as a live config, it can enabled in runtime after all new
> client are up, and it would start working from the next connection that
> would be opened
>
Very large customers have multiple used languages - python / go / C++ /
java ... etc. this will require all the drivers to be moved to a version
that supports this - not only the ones the user really needs it for. It
will also require us to support it on all drivers as well.

This assumes that a customer will be happy to redeploy all its applications
with the new driver ...

Very large customers have a lot of applications. They don't redeploy them
all happily; it can take days (and then it takes days to find the ones they
forgot), and days to find the ones they deployed with the incorrect
version.

> An "ugly" solution would be to return back a new port scylla will be
> listening to for clients that want to use source port for sharding. Maybe
> there is a better idea than that.
>
> it's ugly, and client code might need to change. I don't think we'll want
> to change port magically, if the user ask for specific ones.
>
it shouldn't be random it should be something customer is aware of (like
secured cql port) - they have to open ports in firewall / security groups
etc.

ultrabug:
> This assumes that a customer will be happy to redeploy all its applications with the new driver ...

As a customer, if I bothered to switch from Cassandra to Scylla in the first place I guess I'd be willing to squeeze most of it even if migrating driver is required.

But I agree that it doesn't hold true for everyone and that it could take time and even that some of the applications might never migrate too of course.

> it shouldn't be random it should be something customer is aware of (like secured cql port) - they have to open ports in firewall / security groups etc.

AFAIK scylla already offers two options to allow clients select how they want to connect:
- native_transport_port: 9042
- native_transport_port_ssl: 9142

Why couldn't we extend on this and propose two new options which has this sharding algo enabled like :
- shard_aware_transport_port: 9242
- shard_aware_transport_port_ssl: 9342

(with the same logic, if ssl enabled and only `shard_aware_transport_port` is set, enforce ssl on it too)

Wouln't that sound reasonable?

haaawk:
I'm wondering whether we could keep old logic on 9042 and open another port which would have the new logic. This way clients would be able to decide which behaviour they want.

avikivity:
Having another listener is reasonable.

ultrabug:
@avikivity cool; if that's a green light forward, can this be scheduled to be done in a reasonable time frame in your opinion?

It would be a great benefit for all drivers and Scylla performance and efficiency!

ultrabug:
Friendly ping to @tzach who I guess is also essential to getting this scheduled/done :pray:

jul-stas:
My 3 cents: some kernels seem to assign only even port numbers ([this](https://unix.stackexchange.com/questions/379533/debian-stretch-source-tcp-port-is-always-even) looks related). So a non-shard-aware client will, by default, miss the odd shards with `load_balancing_algorithm::port`.

Experimental result from Fedora 32:
```
cqlsh> select * from system.clients;

 address   | port  | client_type | ...
 127.0.0.1 | 39328 |         cql | ...
 127.0.0.1 | 39330 |         cql | ...
 127.0.0.1 | 39332 |         cql | ...
 127.0.0.1 | 39334 |         cql | ...
 127.0.0.1 | 39336 |         cql | ...
 127.0.0.1 | 39352 |         cql | ...
 127.0.0.1 | 39354 |         cql | ...
 
 (7 rows)
```

fruch:
But in the clone we will bind to other ports, regardless of the kernel.

I don't think any kernel would block that option...

slivne:
@fruch I think a different approach was merged - please close if this is irrelevant

fruch:
@slivne you are correct... even that we didn't yet implemented it on python-driver...

PR #6787, opened 2020-07-07
Repair improve logging with uuid for repair job and more accurate error messages

### LINKED ISSUE #5612: repair: log recoverable errors as warnings rather than info messages
Scylla version 12bc965f713336792ee5d32609b3885ff09c25aa
As seen in https://jenkins.scylladb.com/view/master/job/scylla-master/job/next/1507/artifact/logs-release.2/1579358513144_counter_tests.TestCountersOnMultipleNodes.counter_consistency_node_repair_test/node3.log
repair may hit recoverable errors like:
```
WARN  2020-01-18 14:41:28,152 [shard 0] repair - Repair 769 out of 769 ranges, id=2, shard=0, keyspace=test, table={cf}, range=(9206711021811862052, +inf), peers={127.0.12.2, 127.0.12.1}, live_peers={127.0.12.2}, status=partial
```

but eventually it prints the summary as INFO:
```
INFO  2020-01-18 14:41:28,153 [shard 0] repair - repair id 2 on shard 0 failed - 769 ranges failed
INFO  2020-01-18 14:41:28,153 [shard 0] repair - repair id 2 on shard 0 failed: std::runtime_error (repair 2 on shard 0 failed to do checksum for 769 sub ranges)
INFO  2020-01-18 14:41:28,153 [shard 0] repair - repair id 2 failed: std::runtime_error ({shard 0: std::runtime_error (repair 2 on shard 0 failed to do checksum for 769 sub ranges), shard 1: std::runtime_error (repair 2 on shard 1 failed to do checksum for 769 sub ranges)})
```

This is misleading as repair failed, but since we tolerate the failure which is recoverable, we need to log this as a warning rather than info message.

### LINKED ISSUE #6785: repair: inaccurate log from check_failed_ranges
When a repair fails we see log like:

format("repair {:d} on shard {:d} failed to do checksum for {:d} sub ranges", id, shard, nr_failed_ranges)

This is inaccurate because the error for the failure can be other reasons than checksum failure.

### LINKED ISSUE #6786: Add uuid to a repair job
Currently, repair uses an integer to identify a repair job. The repair
id starts from 1 since node restart. As a result, different repair jobs
will have same id across restart.

To make the id more unique across restart, we can use an uuid in
addition to the integer id. We can not drop the use of the integer id
completely since the http api and nodetool use it.

asias:
This series contains patches to improve repair logging.

1) repair: Add uuid to a repair job

Currently, repair uses an integer to identify a repair job. The repair
id starts from 1 since node restart. As a result, different repair jobs
will have same id across restart.

To make the id more unique across restart, we can use an uuid in
addition to the integer id. We can not drop the use of the integer id
completely since the http api and nodetool use it.

Fixes #6786

2) repair: Fix inaccurate in check_failed_ranges

The reason for the failure can be other reasons than failure of
checksum.

Fixes #6785

3)  repair: Use warn level for logs with recoverable failures

Those logs are not fatal and recoverable. We should make them warn level
instead of info level.

Fixes #5612

### DIFF
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -759,7 +759,7 @@ void repair_info::check_failed_ranges() {
         id, shard, ranges.size(), _sub_ranges_nr, _stats.get_stats());
     if (nr_failed_ranges) {
         rlogger.warn("repair id {} on shard {} failed - {} ranges failed", id, shard, nr_failed_ranges);
-        throw std::runtime_error(format("repair {:d} on shard {:d} failed to do checksum for {:d} sub ranges", id, shard, nr_failed_ranges));
+        throw std::runtime_error(format("repair id {:d} on shard {:d} failed to repair {:d} sub ranges", id, shard, nr_failed_ranges));
     } else {
         if (dropped_tables.size()) {
             rlogger.warn("repair id {} on shard {} completed successfully, keyspace={}, ignoring dropped tables={}", id, shard, keyspace, dropped_tables);
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -256,15 +256,15 @@ tracker::~tracker() {
     _the_tracker = nullptr;
 }
 
-void tracker::start(int id) {
-    _status[id] = repair_status::RUNNING;
+void tracker::start(repair_uniq_id id) {
+    _status[id.id] = repair_status::RUNNING;
 }
 
-void tracker::done(int id, bool succeeded) {
+void tracker::done(repair_uniq_id id, bool succeeded) {
     if (succeeded) {
-        _status.erase(id);
+        _status.erase(id.id);
     } else {
-        _status[id] = repair_status::FAILED;
+        _status[id.id] = repair_status::FAILED;
     }
 }
 repair_status tracker::get(int id) {
@@ -279,8 +279,8 @@ repair_status tracker::get(int id) {
     }
 }
 
-int tracker::next_repair_command() {
-    return _next_repair_command++;
+repair_uniq_id tracker::next_repair_command() {
+    return repair_uniq_id{_next_repair_command++, utils::make_random_uuid()};
 }
 
 future<> tracker::shutdown() {
@@ -345,7 +345,7 @@ named_semaphore& tracker::range_parallelism_semaphore() {
     return _range_parallelism_semaphores[this_shard_id()];
 }
 
-future<> tracker::run(int id, std::function<void ()> func) {
+future<> tracker::run(repair_uniq_id id, std::function<void ()> func) {
     return seastar::with_gate(_gate, [this, id, func =std::move(func)] {
         start(id);
         return seastar::async([func = std::move(func)] { func(); }).then([this, id] {
@@ -690,7 +690,7 @@ repair_info::repair_info(seastar::sharded<database>& db_,
     const sstring& keyspace_,
     const dht::token_range_vector& ranges_,
     std::vector<utils::UUID> table_ids_,
-    int id_,
+    repair_uniq_id id_,
     const std::vector<sstring>& data_centers_,
     const std::vector<sstring>& hosts_,
     streaming::stream_reason reason_)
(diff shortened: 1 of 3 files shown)


repair/repair.hh
```diff
@@ -46,6 +46,14 @@ public:
     repair_stopped_exception() : repair_exception("Repair stopped") { }
 };
 
+struct repair_uniq_id {
+    int id;
+    utils::UUID uuid;
+    friend std::ostream& operator<<(std::ostream& os, const repair_uniq_id& x) {
```
avikivity, line 52: Please move this to .cc.

repair/repair.cc
```diff
@@ -711,8 +711,8 @@ repair_info::repair_info(seastar::sharded<database>& db_,
 future<> repair_info::do_streaming() {
     size_t ranges_in = 0;
     size_t ranges_out = 0;
-    _sp_in = make_lw_shared<streaming::stream_plan>(format("repair-in-id-{:d}-shard-{:d}-index-{:d}", id, shard, sp_index), streaming::stream_reason::repair);
-    _sp_out = make_lw_shared<streaming::stream_plan>(format("repair-out-id-{:d}-shard-{:d}-index-{:d}", id, shard, sp_index), streaming::stream_reason::repair);
+    _sp_in = make_lw_shared<streaming::stream_plan>(format("repair-in-id-{:d}-shard-{:d}-index-{:d}", id.id, shard, sp_index), streaming::stream_reason::repair);
+    _sp_out = make_lw_shared<streaming::stream_plan>(format("repair-out-id-{:d}-shard-{:d}-index-{:d}", id.id, shard, sp_index), streaming::stream_reason::repair);
```
avikivity, line 715: Why not print id instead of id.id? So we start using the uuids.

Later, we can add the uuid to the repair start API (client selects uuid, not server, so if the server crashes before returning we still know the uuid).

repair/repair.cc
```diff
@@ -256,15 +256,15 @@ tracker::~tracker() {
...
 }
 
-void tracker::done(int id, bool succeeded) {
+void tracker::done(repair_uniq_id id, bool succeeded) {
     if (succeeded) {
-        _status.erase(id);
+        _status.erase(id.id);
     } else {
-        _status[id] = repair_status::FAILED;
+        _status[id.id] = repair_status::FAILED;
```
avikivity, line 267: Let's call it id.sequence_number to make it clear.

repair/repair.hh
avikivity: You can add `std::strong_ordering operator<=>(repair_uniq_id) const = default;` and get automatic compares.

avikivity:
I queued the first two patches.

repair/repair.cc
asias: Here, it is used by the partition level repair only. I want to keep it as it now for compatibility reasons.

repair/repair.cc
asias: But, being a sequence number is just an implementation detail. It is still an id used to identify a repair job.

asias:
> I queued the first two patches.

OK. Thanks. 

Btw, this 3rd patch have conflicts with the "repair: Add synchronous API to query repair status #6743" series. I will send new version once they are in master.

repair/repair.hh
asias: OK.

repair/repair.cc
avikivity: ok.

repair/repair.cc
avikivity: yes, but the uuid is also an id. So we have id, id.id, and id.uuid. It is very unfriendly to someone who isn't aware of the history and doesn't know which of the three choices to use.

Better add a comment to repair_uniq_id too.

asias: I added comments to repair_uniq_id.

repair/repair.hh
asias: Although I love spaceship a lot, here I wanted a print operator not less operator.

asias:
I am closing this one in favour of a new PR containing the last patch alone: repair: Add uuid to a repair job.

PR #6791, opened 2020-07-07
node_exporter_install: stop service before force installing

### LINKED ISSUE #6782: node_exporter_install --force failing
*Installation details*
Scylla version (or git commit hash): 4.1.1-0.20200703.1d9bbbc9577
Cluster size:

Platform (physical/VM/cloud instance type/docker): t3.large

```
[centos@ip-10-1-56-33 ~]$ sudo node_exporter_install --force
node_exporter already installed, reinstalling
  File "/opt/scylladb/scripts/libexec/node_exporter_install", line 62, in <module>
    tf.extractall(INSTALL_DIR)
  File "/opt/scylladb/python3/lib64/python3.7/tarfile.py", line 2000, in extractall
    numeric_owner=numeric_owner)
  File "/opt/scylladb/python3/lib64/python3.7/tarfile.py", line 2042, in extract
    numeric_owner=numeric_owner)
  File "/opt/scylladb/python3/lib64/python3.7/tarfile.py", line 2112, in _extract_member
    self.makefile(tarinfo, targetpath)
  File "/opt/scylladb/python3/lib64/python3.7/tarfile.py", line 2153, in makefile
    with bltn_open(targetpath, "wb") as target:
OSError: [Errno 26] Text file busy: '/opt/scylladb/Prometheus/node_exporter/node_exporter-0.17.0.linux-amd64/node_exporter'
```

The script should be able to deal with node_exporter already running and holding the file open

syuu1228:
Stop node-exporter.service before re-install it, to avoid 'Text file busy' error.

Fixes #6782

### DIFF
--- a/dist/common/scripts/node_exporter_install
+++ b/dist/common/scripts/node_exporter_install
@@ -42,6 +42,11 @@ if __name__ == '__main__':
     if node_exporter_p.exists() or (bindir_p() / 'prometheus-node_exporter').exists():
         if force:
             print('node_exporter already installed, reinstalling')
+            try:
+                node_exporter = systemd_unit('node-exporter.service')
+                node_exporter.stop()
+            except:
+                pass
         else:
             print('node_exporter already installed, you can use `--force` to force reinstallation')
             sys.exit(1)


avikivity:
More standard is to create a file with a temporary name, then rename it into the original name.

dist/common/scripts/node_exporter_install
```diff
@@ -42,6 +42,11 @@ if __name__ == '__main__':
     if node_exporter_p.exists() or (bindir_p() / 'prometheus-node_exporter').exists():
         if force:
             print('node_exporter already installed, reinstalling')
+            try:
+                node_exporter = systemd_unit('node-exporter.service')
+                node_exporter.stop()
+            except:
+                pass
```
dyasny, line 49: If the service fails to stop, the error will not be caught

avikivity: Indeed a better method is to rename the file into the new position, and issue a restart.

PR #6815, opened 2020-07-09
dist: fix debian generated files for non-default PRODUCT setting

avikivity:
There are a bunch of renames that are done if PRODUCT is not the
default, but the Python code for them is incorrect. Path.glob()
is not a static method, and Path does not support .endswith().

Fix by constructing a Path object, and later casting to str.


penberg:
Queued as https://github.com/scylladb/scylla/commit/06341d2528b161be3031c820975e4403ba2379c4

PR #5763, opened 2020-02-10
Rewrite CQL3 restriction representation

dekimir:
This is the first stage of replacing the existing restrictions code with a new representation. It adds a new class `expression` to replace the existing class `restriction`. Lots of the old code is deleted, though not all -- that will come in subsequent stages.

Tests: unit (dev, debug restrictions_test), dtest (next-gating)

PR #6818, opened 2020-07-11
Follow-up to CQL restrictions rewrite

dekimir:
This takes care of some follow-up requests from the #5763 review.

Use concepts to explicitly state some interface constraints, and rename `find_if` to `find_atom`.

Tests: unit (dev)

### DIFF
--- a/cql3/restrictions/statement_restrictions.cc
+++ b/cql3/restrictions/statement_restrictions.cc
@@ -26,6 +26,7 @@
 #include <boost/range/adaptors.hpp>
 #include <boost/range/algorithm.hpp>
 #include <functional>
+#include <ranges>
 #include <stdexcept>
 
 #include "query-result-reader.hh"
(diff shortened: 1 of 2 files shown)


cql3/restrictions/statement_restrictions.cc
```diff
@@ -1095,6 +1096,7 @@ bytes_opt get_kth(size_t k, const query_options& options, const ::shared_ptr<ter
 
 template<typename Range>
 value_list to_sorted_vector(const Range& r, const serialized_compare& comparator) {
+    BOOST_CONCEPT_ASSERT((boost::ForwardRangeConcept<Range>));
```
avikivity, line 1099: Why are you using boost concepts? Isn't this std::ranges::forward_range?

dekimir: Because I'm passing boost ranges, which apparently don't satisfy the std requirement.  When I `require std::ranges::forward_range`, I get this error at a call site:
```
/usr/include/c++/10/concepts:67:28: note:   ‘std::forward_iterator_tag’ is not a base of ‘boost::iterators::detail::iterator_category_with_traversal<std::input_iterator_tag, boost::iterators::bidirectional_traversal_tag>’
```

avikivity: So how about using std::ranges instead of boost? I want to reduce use of new boost libraries, and I want our concepts to be real concepts, not boost macros.

dekimir:
Feedback addressed, PTAL.  Changelog:

1. Rebase on master (which already received the first two commits)
2. d66f57eda Fix tombstone-range check for TRUE

cql3/restrictions/statement_restrictions.cc
dekimir: > So how about using std::ranges instead of boost?

That would be inconsistent with the other dozen uses within the same file.  I can look into replacing them all in a follow-up, of course.

avikivity: I don't want to add boost concepts. I'd like to restrict the amount of vocabulary a developer needs to learn in order to understand the code base, not to expand it.

The requirement to have consistency within the same source file makes it harder to make progress.

dekimir: Switched to using std ranges and concept.

dekimir:
Feedback addressed, PTAL.  Changelog:
cfa027953 Use std::ranges concept instead of Boost concept

cql3/statements/delete_statement.cc
```diff
@@ -100,7 +100,7 @@ const column_definition* single_column(const binary_operator& oper) {
 /// See #6493.
 bool bounded_ck(const expression& expr) {
```
avikivity, line 101: Poorly named considering its documentation.

Issue #6252, opened 2020-04-20
Nodetool Repair failing on keyspace with std::runtime_error (row_diff.size() != set_diff.size())


jtnagashima:
This is Scylla's bug tracker, to be used for reporting bugs only.
If you have a question about Scylla, and not a bug, please ask it in
our mailing-list at scylladb-dev@googlegroups.com or in our slack channel.

- [x] I have read the disclaimer above, and I am reporting a suspected malfunction in Scylla.

**Installation details**
Scylla version (or git commit hash): `scylla-3.3.0-0.20200319.7ae2cdf46c7.x86_64`
Cluster size: 5
OS (RHEL/CentOS/Ubuntu/AWS AMI): ScyllaDB 2.3.1 (ami-0402378ae7131b7a0)
Single Datacenter 

**Issue**
We recently ran through an exercise of upgrading a 5 node cluster from Scylla 2.3.6 to 3.3.0. During the upgrade process, we encountered one other issue https://github.com/scylladb/scylla/issues/6205, but it was an otherwise seamless upgrade process.  Disclaimer: during the debug process for the issue, we did end up dropping a node - node 4, and re-adding it to the cluster as a fresh node as a replacement for itself using the `replace_address_first_boot` parameter.

After the nodes were upgraded, we ran `nodetool repair` on the nodes, beginning with node-4 and ran into the following error:
`[shard 1] repair - Got error in row level repair: std::runtime_error (row_diff.size() != set_diff.size())`

**Scylla.yaml**
```
cluster_name: dev
num_tokens: 256
data_file_directories:
    - /var/lib/scylla/data
commitlog_directory: /var/lib/scylla/commitlog
commitlog_sync: periodic
commitlog_sync_period_in_ms: 10000
commitlog_segment_size_in_mb: 32
seed_provider:
    - class_name: org.apache.cassandra.locator.SimpleSeedProvider
      parameters:
          - seeds: "x.x.x.1"
listen_address: x.x.x.4
native_transport_port: 9042
read_request_timeout_in_ms: 5000
write_request_timeout_in_ms: 2000
endpoint_snitch: Ec2Snitch
rpc_address: 0.0.0.0
rpc_port: 9160
api_port: 10000
api_address: 127.0.0.1
batch_size_warn_threshold_in_kb: 5
batch_size_fail_threshold_in_kb: 50
broadcast_rpc_address: x.x.x.4
partitioner: org.apache.cassandra.dht.Murmur3Partitioner
[...]

PR #6823, opened 2020-07-13
repair: Relax size check of get_row_diff and set_diff

asias:
In case a row hash conflict, a hash in set_diff will get more than one
row from get_row_diff.

For example,

Node1 (Repair master):
row1  -> hash1
row2  -> hash2
row3  -> hash3
row3' -> hash3

Node2 (Repair follower):
row1  -> hash1
row2  -> hash2

We will have set_diff = {hash3} between node1 and node2, while
get_row_diff({hash3}) will return two rows: row3 and row3'. And the
error below was observed:

   repair - Got error in row level repair: std::runtime_error
   (row_diff.size() != set_diff.size())

In this case, node1 should send both row3 and row3' to peer node
instead of fail the whole repair. Because node2 does not have row3 or
row3', otherwise node1 won't send row with hash3 to node1 in the first
place.

Refs: #6252

### DIFF
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -1731,7 +1731,8 @@ public:
             auto sz = set_diff.size();
             return get_row_diff(std::move(set_diff), needs_all_rows).then([this, remote_node, sz] (std::list<repair_row> row_diff) {
                 if (row_diff.size() != sz) {
-                    throw std::runtime_error("row_diff.size() != set_diff.size()");
+                    rlogger.warn("Hash conflict detected, keyspace={}, table={}, range={}, row_diff.size={}, set_diff.size={}. It is recommended to compact the table and rerun repair for the range.",
+                            _schema->ks_name(), _schema->cf_name(), _range, row_diff.size(), sz);
                 }
                 return do_with(std::move(row_diff), [this, remote_node] (std::list<repair_row>& row_diff) {
                     return get_repair_rows_size(row_diff).then([this, remote_node, &row_diff] (size_t row_bytes) mutable {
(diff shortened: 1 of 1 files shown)


repair/row_level.cc
```diff
@@ -1731,7 +1731,8 @@ class repair_meta {
             auto sz = set_diff.size();
             return get_row_diff(std::move(set_diff), needs_all_rows).then([this, remote_node, sz] (std::list<repair_row> row_diff) {
                 if (row_diff.size() != sz) {
-                    throw std::runtime_error("row_diff.size() != set_diff.size()");
+                    rlogger.warn("Hash conflict detected, keyspace={}, table={}, range={}, row_diff.size={}, set_diff.size={}",
+                            _schema->ks_name(), _schema->cf_name(), _range, row_diff.size(), sz);
```
avikivity, line 1735: What is the user supposed to do if they see the warning? I suggest to drop it to debug, or maybe to info but with an explanation that no action is needed.

avikivity:
Nice bug. Why refs and not fixes?

asias:
> Nice bug. Why refs and not fixes?

Because it does not resolve the the hash conflict issue completely.

repair/row_level.cc
asias: The use can rerun the repair. Since each repair job will use a random hash seed, hopefully next time the hash conflict is gone.  

I am thinking about detecting conflicts and fallback to sync all the rows inside the row buffer which is 32MB in size, or fallback to a hasher with more bits.

asias: In this particular case, the user needs no action because node2 is perfectly repaired (got the missing row3 and row3').  In other cases,  like:

```
    Node1:
    row1  -> hash1
    row2  -> hash2
    row3  -> hash3
   
    Node2:
    row1  -> hash1
    row2  -> hash2  
    row3' -> hash3
```
n1 will not ask n2 to send row with hash3. So row3 and row3' will not be repaired in this run.

avikivity: Well, need to explain more in the message. Some of our users are not experts in row-level repair.

asias: OK. I added more in the message in the new version.

asias:
Changes in 4a84f42226bbc7225675c81c51fa060d703d995b:
- Add recommendation for users what to do when the hash conflict happens.

bhalevy:
@avikivity can you please backport to all 4.x branches?

avikivity:
It is already backported. Please communicate backports on the issue, not patch.

bhalevy:
@avikivity the reason I requested backport on the pull request rather than the issue in this case is that it Refs the issue rather than Fixes it, so the issue is not marked as Backport candidate

PR #6826, opened 2020-07-13
cdc: Retry generation fetching after `read_failure_exception`

### LINKED ISSUE #6804: cdc: better classify errors when fetching CDC generations
This PR: https://github.com/scylladb/scylla/pull/6670
changed the behavior of CDC generation fetching code to stop retrying if an "exotic" error, such as a deserialization error, appears. These errors indicate probable bugs.

However, it doesn't classify all errors properly. For example, the following error message taken from an order version of Scylla:
```
Could not retrieve CDC streams with timestamp 2020/06/18 20:14:37: exceptions::read_failure_exception (Operation failed for system_distributed.cdc_generations - received 1 responses and 2 failures from 2 CL=QUORUM.). Will retry again.
```

would now be classified as an "exotic" error, even though it could happen due to a simple network partition (when the querying node fails to receive responses from a subset of nodes).
This needs to be improved.

jul-stas:
While fetching CDC generations, various exceptions can occur. They are divided into "fatal" and "nonfatal", where "fatal" ones prevent retrying of the fetch operation.

This patch makes `read_failure_exception` "non-fatal", because such error may appear during restart. In general this type of error can mean a few different things (e.g. an error code in a response from replica, but also a broken connection) so retrying seems reasonable.

Fixes #6804

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -698,6 +698,8 @@ bool storage_service::do_handle_cdc_generation_intercept_nonfatal_errors(db_cloc
         throw cdc_generation_handling_nonfatal_exception(e.what());
     } catch (exceptions::unavailable_exception& e) {
         throw cdc_generation_handling_nonfatal_exception(e.what());
+    } catch (exceptions::read_failure_exception& e) {
+        throw cdc_generation_handling_nonfatal_exception(e.what());
     } catch (...) {
         const auto ep = std::current_exception();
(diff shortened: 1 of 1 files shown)


haaawk:
Could you please have a look @nyh ?

nyh:
Committed to next: d1dec3fcd7537136ae1068fe349f26a21f5ebecc

PR #6829, opened 2020-07-13
Switch Estimated histogram to time_estimated_histogram latency

amnonh:
This series is a continuation of the series that introduced the time_estimated_histogram.

time_estimated_histogram is more efficient and uses a higher reporting resolution.

See  #5815 and  #4746

### DIFF
--- a/db/view/row_locking.cc
+++ b/db/view/row_locking.cc
@@ -80,7 +80,7 @@ row_locker::lock_pk(const dht::decorated_key& pk, bool exclusive, db::timeout_cl
     // even in the case of rehashing.
     return f.then([this, pk = &i->first, exclusive, &single_lock_stats, waiting_latency = std::move(waiting_latency)] () mutable {
         waiting_latency.stop();
-        single_lock_stats.estimated_waiting_for_lock.add(waiting_latency.latency(), single_lock_stats.operations_currently_waiting_for_lock);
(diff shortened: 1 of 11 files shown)


db/view/row_locking.cc
```diff
@@ -80,7 +80,7 @@ row_locker::lock_pk(const dht::decorated_key& pk, bool exclusive, db::timeout_cl
     // even in the case of rehashing.
     return f.then([this, pk = &i->first, exclusive, &single_lock_stats, waiting_latency = std::move(waiting_latency)] () mutable {
         waiting_latency.stop();
-        single_lock_stats.estimated_waiting_for_lock.add(waiting_latency.latency(), single_lock_stats.operations_currently_waiting_for_lock);
+        single_lock_stats.estimated_waiting_for_lock.add(waiting_latency.latency());
```
avikivity, line 83: What is lost by not passing that parameter?

amnonh:
> Is there any compatiblity impact for JMX/REST?

No, from the API it returns the same histogram.

db/view/row_locking.cc
amnonh: It's no longer supported. Originally, the idea was to sample the data, so when adding a value to the histogram it was added with the new total number of values (you can see that the code here just adds n+1)

Because all the values are added to the histogram there is no need for all those manipulations

PR #6834, opened 2020-07-14
cql3/restrictions: exclude NULLs from comparison in filtering v2

### LINKED ISSUE #6295: Filtering captures uninitialized/deleted values of certain types

jul-stas:
NULLs used to give false positives in GT, LT, GEQ and LEQ ops performed upon `ALLOW FILTERING`. That was a consequence of not distinguishing NULL from an empty buffer.

This patch excludes NULLs on high level, preventing them from entering LHS of any comparison, i.e. it assumes that any binary operation should return `false` whenever the LHS operand is NULL (note: at the moment filters with RHS NULL, such as `...WHERE x=NULL ALLOW FILTERING`, return empty sets anyway).

Fixes #6295

### DIFF
--- a/cql3/restrictions/statement_restrictions.cc
+++ b/cql3/restrictions/statement_restrictions.cc
@@ -807,7 +807,7 @@ bool limits(const binary_operator& opr, const column_value_eval_bag& bag) {
     } else if (columns.size() == 1) {
         auto lhs = get_value(columns[0], bag);
         if (!lhs) {
-            lhs = bytes(); // Compatible with old code, which feeds null to type comparators.
+            return false; // NULL never compares to anything (#6295)
         }
         const auto tup = get_tuple(*opr.rhs, bag.options);
         auto rhs = (tup && tup->size() == 1) ? tup->get_elements()[0] // Assume an external query WHERE (ck1)>(123).
--- a/test/boost/restrictions_test.cc
+++ b/test/boost/restrictions_test.cc
@@ -232,7 +232,7 @@ SEASTAR_THREAD_TEST_CASE(regular_col_slice) {
         require_rows(e, "select * from t where q<11 and q>11 allow filtering", {});
         require_rows(e, "select q from t where q<=12 and r>=21 allow filtering", {{I(11), I(21)}, {I(12), I(22)}});
         cquery_nofail(e, "insert into t(p) values (4)");
-        require_rows(e, "select q from t where q<12 allow filtering", {{std::nullopt}, {I(10)}, {I(11)}});
+        require_rows(e, "select q from t where q<12 allow filtering", {{I(10)}, {I(11)}});
         require_rows(e, "select q from t where q>10 allow filtering", {{I(11)}, {I(12)}, {I(13)}});
         require_rows(e, "select q from t where q<12 and q>10 allow filtering", {{I(11)}});
     }).get();
--- a/test/boost/restrictions_test.cc
+++ b/test/boost/restrictions_test.cc
@@ -260,7 +260,47 @@ SEASTAR_THREAD_TEST_CASE(regular_col_neq) {
 }
 #endif // 0
 
-SEASTAR_THREAD_TEST_CASE(null) {
+namespace {
+/// The `op` argument can be one of: "<", "<=", ">", ">=", "=". Remove float args to use "!=".
+future<> test_lhs_null_with_operator(sstring op) {
+    return do_with_cql_env_thread([op = std::move(op)] (cql_test_env& e) {
+        cquery_nofail(e, "CREATE TABLE cf (pk int primary key, flt float,"
+                " dbl double, dec decimal, b boolean, ti tinyint, si smallint," 
+                " i int, bi bigint, vi varint, ip inet, u uuid, tu timeuuid,"
+                " t time, d date, a ascii, txt text);");
+        cquery_nofail(e, "INSERT INTO cf (pk) VALUES (0);");
+        
+        // Now we have all types of NULLs
+        require_rows(e, format("SELECT * FROM cf WHERE flt {} 99.9 ALLOW FILTERING;", op), {});
+        require_rows(e, format("SELECT * FROM cf WHERE dbl {} 99.9 ALLOW FILTERING;", op), {});
+        require_rows(e, format("SELECT * FROM cf WHERE dec {} 99.9 ALLOW FILTERING;", op), {});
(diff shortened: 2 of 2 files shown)


cql3/restrictions/statement_restrictions.cc
```diff
@@ -807,7 +807,7 @@ bool limits(const binary_operator& opr, const column_value_eval_bag& bag) {
     } else if (columns.size() == 1) {
         auto lhs = get_value(columns[0], bag);
         if (!lhs) {
-            lhs = bytes(); // Compatible with old code, which feeds null to type comparators.
```
haaawk, line 810: What is this old code that empty value was ensuring compatibility with?

jul-stas: I presume it was "the code before Dejan's overhaul", which passes NULLs as empty buffers, being exactly the source of this bug.

test/boost/filtering_test.cc
```diff
@@ -1176,3 +1176,155 @@ SEASTAR_TEST_CASE(test_filtering) {
 
     });
 }
+
+SEASTAR_TEST_CASE(test_filtering_gt_not_capturing_nulls) {
```
nyh, line 1180: I would have liked to have a comment explaining in words (not C++ code) what this code is checking (maybe also referencing the issue), and give some justification why the tested behavor is really correct: Is it because of common sense? Because Cassandra behaves this way? Because SQL behaves this way? Or what?

test/boost/filtering_test.cc
```diff
@@ -1176,3 +1176,155 @@ SEASTAR_TEST_CASE(test_filtering) {
...
+            assert_that(msg).is_rows().is_empty();
+        }
+        {
+            auto msg = e.execute_cql("SELECT * FROM cf WHERE txt > '0' ALLOW FILTERING;").get0();
+            assert_that(msg).is_rows().is_empty();
+        }
+    });
+}
+
+SEASTAR_TEST_CASE(test_filtering_lt_not_capturing_nulls) {
```
nyh, line 1256: I think it would be better to just have *one* test, which will run a loop on several operators. There are two benefits to my idea:
1. Less code duplication, and less chance of accidental difference between the duplicated versions.
2. Easy to check additional operators - what about testing <=, >=, =, != ?

cql3/restrictions/statement_restrictions.cc
dekimir: That's correct.

test/boost/filtering_test.cc
```diff
@@ -1176,3 +1176,155 @@ SEASTAR_TEST_CASE(test_filtering) {
 
     });
 }
+
+SEASTAR_TEST_CASE(test_filtering_gt_not_capturing_nulls) {
+    return do_with_cql_env_thread([] (cql_test_env& e) {
+        e.execute_cql("CREATE TABLE cf (pk int primary key, flt float,"
```
dekimir, line 1182: Please use cquery_nofail.

test/boost/filtering_test.cc
```diff
@@ -1176,3 +1176,155 @@ SEASTAR_TEST_CASE(test_filtering) {
...
+                " dbl double, dec decimal, b boolean, ti tinyint, si smallint," 
+                " i int, bi bigint, vi varint, ip inet, u uuid, tu timeuuid,"
+                " t time, d date, a ascii, txt text);").get();
+        e.execute_cql("INSERT INTO cf (pk) VALUES (0);").get();
+        
+        // Now we have all types of NULLs
+        {
+            auto msg = e.execute_cql("SELECT * FROM cf WHERE flt > -999.9 ALLOW FILTERING;").get0();
+            assert_that(msg).is_rows().is_empty();
```
dekimir, line 1191: Please use require_rows.

test/boost/filtering_test.cc
dekimir: Yes, there's way too much duplication here.  Also, I'm not sure we need to cover every single operator/type combination -- that's just bloating the runtime without much additional confidence.

test/boost/filtering_test.cc
dekimir: This really is a test of `null` on the LHS.  We already have a test named null (line 263) that tests `null` on the RHS.  It would be nice if both tests were renamed to emphasize this.

jul-stas: - I added explanation
- I renamed both tests to `null_{l,r}hs` and put them into the same file

test/boost/filtering_test.cc
jul-stas: I reduced duplication, but kept only the "<" and ">" cases. Would you like to see any other operator? (honestly, I don't know)

test/boost/filtering_test.cc
jul-stas: done

test/boost/filtering_test.cc
jul-stas: done

test/boost/filtering_test.cc
dekimir: I think `>` and `<` are enough to catch any bugs that creep in.  (I also think that trying every single type is superfluous and stale-bait, but that's not enough to hold back this PR.)

jul-stas:
@nyh ping

test/boost/filtering_test.cc
nyh: It depends how much time each case takes. If each case takes a millisecond, I don't see any reason not to do it. But I don't insist. I'll accept this as-is.

nyh:
Committed to next  09a71ccd84ce76d01aba5bd74a35b2733f7372e2

PR #6835, opened 2020-07-14
build: install jmx and tools-java submodule dependencies

avikivity:
Let each submodule be responsible for its own dependencies, and
call the submodule's dependency installation script.

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -28,6 +28,8 @@ else
 fi
 
 bash seastar/install-dependencies.sh
+bash tools/jmx/install-dependencies.sh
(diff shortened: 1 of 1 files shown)


syuu1228:
This looks good, but after this change we also need to do following things:
- Drop -jmx / -tools dependencies from main repo's install-dependencies.sh
- Update & cleanup install-dependencies.sh in -jmx / -tools, currently these are unmaintained

avikivity:
@penberg ping

penberg:
Queued as b4b9deadf3ee4f581dc34b6f4bf2f79d2f2f06b9

PR #6836, opened 2020-07-14
build: remove pystache from install-dependencies

avikivity:
As of d6165bc1c348f0c38231a6e6ab85b259b0c4143a we do not
depend on pystache, so don't install it.


avikivity:
@penberg ping

penberg:
Queued as 7fbe50a4e4d170e7d84a9a95830ee19f6450c752

Issue #6733, opened 2020-06-30
Repair should expose it's progress


amnonh:
The repair has traces for its progress, it should expose metrics as well

slivne:
Repairs run via manager provide progress. 

this is a duplicate of https://github.com/scylladb/scylla/issues/1244

avikivity:
Not a bug, not backporting.

avikivity:
Looks like the bug was reused, but the previous backport decision holds.

PR #6842, opened 2020-07-15
repair: Add progress metrics for node ops

### LINKED ISSUE #1244: repair: need a method to monitor progress
nodetool netstats does provide some information that repair is progressing but it is not providing any info on how much is left to be completed

asias:
This series adds progress metrics for the node operations. Metrics for bootstrap and rebuild progress are added as a starter. I will add more for the remaining operations after getting feedback.

With this the Scylla Monitor and Scylla Manager can know the progress of the bootstrap and other node operations. E.g.,
```
    scylla_node_ops_bootstrap_nr_ranges_finished{shard="0",type="derive"} 50
    scylla_node_ops_bootstrap_nr_ranges_total{shard="0",type="derive"} 1040
```

Fixes #1244, #6733

### DIFF
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -60,12 +60,16 @@ public:
                 [this] { return bootstrap_finished_percentage(); }),
             sm::make_gauge("rebuild_finished_percentage", sm::description("Number of finished percentage for rebuild operation on this shard."),
                 [this] { return rebuild_finished_percentage(); }),
+            sm::make_gauge("repair_finished_percentage", sm::description("Number of finished percentage for repair operation on this shard."),
+                [this] { return repair_finished_percentage(); }),
         });
     }
     uint64_t bootstrap_total_ranges{0};
     uint64_t bootstrap_finished_ranges{0};
     uint64_t rebuild_total_ranges{0};
     uint64_t rebuild_finished_ranges{0};
+    uint64_t repair_total_ranges_sum{0};
+    uint64_t repair_finished_ranges_sum{0};
 private:
     seastar::metrics::metric_groups _metrics;
     float bootstrap_finished_percentage() {
@@ -74,6 +78,7 @@ private:
     float rebuild_finished_percentage() {
         return rebuild_total_ranges == 0 ? 1 : float(rebuild_finished_ranges) / float(rebuild_total_ranges);
     }
+    float repair_finished_percentage();
 };
 
 static thread_local node_ops_metrics _node_ops_metrics;
@@ -271,6 +276,13 @@ tracker& repair_tracker() {
     }
 }
 
+float node_ops_metrics::repair_finished_percentage() {
+    if (_the_tracker) {
+        return repair_tracker().report_progress(streaming::stream_reason::repair);
+    }
+    return 1;
+}
+
 tracker::tracker(size_t nr_shards, size_t max_repair_memory)
     : _shutdown(false)
     , _repairs(nr_shards) {
@@ -396,6 +408,19 @@ void tracker::abort_all_repairs() {
         ri->abort();
     }
     rlogger.info0("Aborted {} repair job(s)", count);
+}
+
+float tracker::report_progress(streaming::stream_reason reason) {
+    uint64_t nr_ranges_finished = 0;
+    uint64_t nr_ranges_total = 0;
+    for (auto& x : _repairs[this_shard_id()]) {
+        auto& ri = x.second;
+        if (ri->reason == reason) {
+            nr_ranges_total += ri->nr_ranges_total;
+            nr_ranges_finished += ri->nr_ranges_finished;
+        }
+    }
+    return nr_ranges_total == 0 ? 1 : float(nr_ranges_finished) / float(nr_ranges_total);
 }
 
 named_semaphore& tracker::range_parallelism_semaphore() {
@@ -762,6 +787,7 @@ repair_info::repair_info(seastar::sharded<database>& db_,
     , data_centers(data_centers_)
     , hosts(hosts_)
     , reason(reason_)
+    , nr_ranges_total(ranges.size())
     , _row_level_repair(db.local().features().cluster_supports_row_level_repair()) {
 }
 
@@ -1423,6 +1449,9 @@ static future<> do_repair_ranges(lw_shared_ptr<repair_info> ri) {
                         _node_ops_metrics.bootstrap_finished_ranges++;
                     } else if (ri->reason == streaming::stream_reason::rebuild) {
                         _node_ops_metrics.rebuild_finished_ranges++;
+                    } else if (ri->reason == streaming::stream_reason::repair) {
+                        _node_ops_metrics.repair_finished_ranges_sum++;
+                        ri->nr_ranges_finished++;
                     }
                 });
             });
@@ -1452,6 +1481,9 @@ static future<> do_repair_ranges(lw_shared_ptr<repair_info> ri) {
                     _node_ops_metrics.bootstrap_finished_ranges++;
                 } else if (ri->reason == streaming::stream_reason::rebuild) {
                     _node_ops_metrics.rebuild_finished_ranges++;
+                } else if (ri->reason == streaming::stream_reason::repair) {
+                    _node_ops_metrics.repair_finished_ranges_sum++;
+                    ri->nr_ranges_finished++;
                 }
             });
         }).then([ri] {
@@ -1562,6 +1594,11 @@ static int do_repair_start(seastar::sharded<database>& db, sstring keyspace,
         rlogger.info("repair id {} completed successfully: no tables to repair", id);
         return id.id;
     }
+
+    auto nr_ranges_total = ranges.size();
+    db.invoke_on_all([nr_ranges_total] (database&) {
+        _node_ops_metrics.repair_total_ranges_sum += nr_ranges_total;
+    }).get();
 
     // Do it in the background.
     (void)repair_tracker().run(id, [&db, id, keyspace = std::move(keyspace),
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -47,8 +47,29 @@
 
 #include <seastar/core/gate.hh>
 #include <seastar/util/defer.hh>
+#include <seastar/core/metrics_registration.hh>
 
 logging::logger rlogger("repair");
+
+class node_ops_metrics {
+public:
+    node_ops_metrics() {
+        namespace sm = seastar::metrics;
+        _metrics.add_group("node_maintenance_operations", {
+            sm::make_gauge("bootstrap_finished_percentage", sm::description("Number of finished percentage for bootstrap operation on this shard."),
+                [this] { return bootstrap_finished_percentage(); }),
+        });
+    }
+    uint64_t bootstrap_total_ranges{0};
+    uint64_t bootstrap_finished_ranges{0};
+private:
+    seastar::metrics::metric_groups _metrics;
+    float bootstrap_finished_percentage() {
+        return bootstrap_total_ranges == 0 ? 1 : float(bootstrap_finished_ranges) / float(bootstrap_total_ranges);
+    }
+};
+
+static thread_local node_ops_metrics _node_ops_metrics;
 
 template <typename T1, typename T2>
 inline
@@ -1390,7 +1411,11 @@ static future<> do_repair_ranges(lw_shared_ptr<repair_info> ri) {
                 ri->ranges_index++;
                 rlogger.info("Repair {} out of {} ranges, id={}, shard={}, keyspace={}, table={}, range={}",
                     ri->ranges_index, ri->ranges.size(), ri->id, ri->shard, ri->keyspace, ri->table_names(), range);
-                return repair_range(*ri, range);
+                return repair_range(*ri, range).then([ri] {
+                    if (ri->reason == streaming::stream_reason::bootstrap) {
+                        _node_ops_metrics.bootstrap_finished_ranges++;
+                    }
+                });
             });
         });
     } else {
@@ -1413,6 +1438,10 @@ static future<> do_repair_ranges(lw_shared_ptr<repair_info> ri) {
                         return make_ready_future<stop_iteration>(stop_iteration::yes);
                     }
                 });
+            }).then([ri] {
+                if (ri->reason == streaming::stream_reason::bootstrap) {
+                    _node_ops_metrics.bootstrap_finished_ranges++;
+                }
             });
         }).then([ri] {
             // Do streaming for the remaining ranges we do not stream in
@@ -1662,9 +1691,25 @@ future<> bootstrap_with_repair(seastar::sharded<database>& db, locator::token_me
     using inet_address = gms::inet_address;
     return seastar::async([&db, tm = std::move(tm), tokens = std::move(bootstrap_tokens)] () mutable {
         auto keyspaces = db.local().get_non_system_keyspaces();
-        rlogger.info("bootstrap_with_repair: started with keyspaces={}", keyspaces);
         auto myip = utils::fb_utilities::get_broadcast_address();
         auto reason = streaming::stream_reason::bootstrap;
+        // Calculate number of ranges to sync data
+        size_t nr_ranges_total = 0;
+        for (auto& keyspace_name : keyspaces) {
+            if (!db.local().has_keyspace(keyspace_name)) {
+                continue;
+            }
+            auto& ks = db.local().find_keyspace(keyspace_name);
+            auto& strat = ks.get_replication_strategy();
+            dht::token_range_vector desired_ranges = strat.get_pending_address_ranges(tm, tokens, myip);
+            seastar::thread::maybe_yield();
+            nr_ranges_total += desired_ranges.size();
+        }
+        db.invoke_on_all([nr_ranges_total] (database&) {
+            _node_ops_metrics.bootstrap_finished_ranges = 0;
+            _node_ops_metrics.bootstrap_total_ranges = nr_ranges_total;
+        }).get();
+        rlogger.info("bootstrap_with_repair: started with keyspaces={}, nr_ranges_total={}", keyspaces, nr_ranges_total);
         for (auto& keyspace_name : keyspaces) {
             if (!db.local().has_keyspace(keyspace_name)) {
                 rlogger.info("bootstrap_with_repair: keyspace={} does not exist any more, ignoring it", keyspace_name);
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -58,14 +58,21 @@ public:
         _metrics.add_group("node_maintenance_operations", {
             sm::make_gauge("bootstrap_finished_percentage", sm::description("Number of finished percentage for bootstrap operation on this shard."),
                 [this] { return bootstrap_finished_percentage(); }),
+            sm::make_gauge("rebuild_finished_percentage", sm::description("Number of finished percentage for rebuild operation on this shard."),
+                [this] { return rebuild_finished_percentage(); }),
         });
     }
     uint64_t bootstrap_total_ranges{0};
     uint64_t bootstrap_finished_ranges{0};
+    uint64_t rebuild_total_ranges{0};
+    uint64_t rebuild_finished_ranges{0};
 private:
     seastar::metrics::metric_groups _metrics;
     float bootstrap_finished_percentage() {
         return bootstrap_total_ranges == 0 ? 1 : float(bootstrap_finished_ranges) / float(bootstrap_total_ranges);
+    }
+    float rebuild_finished_percentage() {
+        return rebuild_total_ranges == 0 ? 1 : float(rebuild_finished_ranges) / float(rebuild_total_ranges);
     }
 };
 
@@ -1414,6 +1421,8 @@ static future<> do_repair_ranges(lw_shared_ptr<repair_info> ri) {
                 return repair_range(*ri, range).then([ri] {
                     if (ri->reason == streaming::stream_reason::bootstrap) {
                         _node_ops_metrics.bootstrap_finished_ranges++;
+                    } else if (ri->reason == streaming::stream_reason::rebuild) {
+                        _node_ops_metrics.rebuild_finished_ranges++;
                     }
                 });
             });
@@ -1441,6 +1450,8 @@ static future<> do_repair_ranges(lw_shared_ptr<repair_info> ri) {
             }).then([ri] {
                 if (ri->reason == streaming::stream_reason::bootstrap) {
                     _node_ops_metrics.bootstrap_finished_ranges++;
+                } else if (ri->reason == streaming::stream_reason::rebuild) {
+                    _node_ops_metrics.rebuild_finished_ranges++;
                 }
             });
         }).then([ri] {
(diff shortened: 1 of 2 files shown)


repair/repair.cc
```diff
@@ -47,9 +47,33 @@
...
+    uint64_t bootstrap_nr_ranges_finished{0};
+    uint64_t rebuild_nr_ranges_total{0};
+    uint64_t rebuild_nr_ranges_finished{0};
+    node_ops_metrics() {
+        namespace sm = seastar::metrics;
+        _metrics.add_group("node_ops", {
+            sm::make_derive("bootstrap_nr_ranges_total", bootstrap_nr_ranges_total,
+                            sm::description("Number of total ranges for bootstrap operation on this shard.")),
+            sm::make_derive("bootstrap_nr_ranges_finished", bootstrap_nr_ranges_finished,
+                            sm::description("Number of finished ranges for bootstrap operation on this shard.")),
```
amnonh, line 66: I don't understand the difference between total and finish.
the nr in the middle is not clear.

A good metrics name's suffix should be the metrics unit, total typically used for a counter (or derived)

repair/repair.cc
```diff
@@ -47,9 +47,33 @@
...
 
+struct node_ops_metrics {
+    seastar::metrics::metric_groups _metrics;
+    uint64_t bootstrap_nr_ranges_total{0};
+    uint64_t bootstrap_nr_ranges_finished{0};
+    uint64_t rebuild_nr_ranges_total{0};
+    uint64_t rebuild_nr_ranges_finished{0};
+    node_ops_metrics() {
+        namespace sm = seastar::metrics;
+        _metrics.add_group("node_ops", {
```
amnonh, line 62: I don't think that node_ops is not a good group name.

repair/repair.cc
```diff
@@ -1409,6 +1439,12 @@ static future<> do_repair_ranges(lw_shared_ptr<repair_info> ri) {
                         return make_ready_future<stop_iteration>(stop_iteration::yes);
                     }
                 });
+            }).then([ri] {
+                if (ri->reason == streaming::stream_reason::bootstrap) {
+                    _node_ops_metrics.bootstrap_nr_ranges_finished++;
+                } else if (ri->reason == streaming::stream_reason::rebuild) {
+                    _node_ops_metrics.rebuild_nr_ranges_finished++;
+                }
```
amnonh, line 1447: The code indicated that maybe bootstrap and rebuild should be a label.

Does it make sense to sum them both together?

repair/repair.cc
asias: There is no unit. The metric is the number of ranges. For example,  

    scylla_node_ops_bootstrap_nr_ranges_finished{shard="0",type="derive"} 50
    scylla_node_ops_bootstrap_nr_ranges_total{shard="0",type="derive"} 1040

"total" means we have 1040 ranges to work on.
"finished "means 50 out of 1040 ranges have been finished.

repair/repair.cc
asias: How about node_operations? Those are for operations like bootstrap, rebuild, replace and so on.

repair/repair.cc
asias: We only allow one operation at a time. It does not make sense to sum different operations.

asias: We have limited number of operations, so we do not have tons of "labels". It is much easier not to use label.

repair/repair.cc
amnonh: so finish is going from 0 to `total` if the value is going up and down it should be a gauge.
Is total a static value, or will it changed over time?

asias: Yes, finished goes from 0 to total. Use sm::make_gauge?

total is a static value per operation, but total can change for different operations, e.g., rebuild after topology change, the number of ranges a node owns have changed. Looks like I need to use sm::make_gauge for total too?

amnonh: yes, both need to be gauge

repair/repair.cc
amnonh: node_operations is better, though still not great, for me it sounds like the total number of operations.
@tzach do you have any input?
the `_nr_` is also not clear

repair/repair.cc
amnonh: I will not insist on labels, but it does sound like you are counting the same things (i.e. ranges).
In what perspective using labels make it harder?
There is a single place in the code that you defined them so it's an additional one of code to define them.
And when using them it's pretty similar if you are looking for a metric or a metric with a label.

repair/repair.cc
asias: Or node_maintenance_operations? They are all under maintenance group.

repair/repair.cc
asias: Ok. I will make them gauge.

repair/repair.cc
asias: How about: 

scylla_node_maintenance_operations_bootstrap_finished_ranges
scylla_node_maintenance_operations_bootstrap_total_ranges

?

repair/repair.cc
asias: I did not find a document / code comments how to use it. The example I found in cql3/query_processor.cc looks very complicated.

asias:
Changes in 2b861d3772d77a0f66b69587d814c3aa1087139f:
- Rename group name from node_ops to node_maintenance_operations
- Rename metrics name to rebuild_finished_ranges and rebuild_total_ranges
- Rename metrics name to bootstrap_finished_ranges and bootstrap_total_ranges

repair/repair.cc
amnonh: I think it's better

asias: Good. I sent out v2 below.

repair/repair.cc
amnonh: you can look at `./alternator/stats.cc` though their macro use is a bit confusing.

you define a label somewhere in source file like:
```
seastar::metrics::label op("op");
```

and then `op("rebuild")` will generate a new label that  you can add to a metrics either as a parameter:
```
sm::make_gauge("bootstrap_nr_ranges_total", 
                            sm::description("Number of total ranges for bootstrap operation on this shard."),
{op("rebuild")},
bootstrap_nr_ranges_total)
```

Or using the parentheses operator:
```
sm::make_gauge("bootstrap_nr_ranges_total", 
                            sm::description("Number of total ranges for bootstrap operation on this shard."),
bootstrap_nr_ranges_total)(op("rebuild"))
```

repair/repair.cc
amnonh: one more thing I've noticed the parameter references or function should be the last parameter, we support the older order, but it's nicer that way

repair/repair.cc
asias: OK. Thanks for the tip.

repair/repair.cc
asias: OK. I will change the order.

asias:
Changes in 55e6bfa17ded0112279c5c57774279f973de2245:
- Make sm::description come first in sm::make_gauge.

asias:
@avikivity Can you please take a look too?

repair/repair.cc
```diff
@@ -47,9 +47,27 @@
...
+    uint64_t bootstrap_total_ranges{0};
+    uint64_t bootstrap_finished_ranges{0};
+    node_ops_metrics() {
+        namespace sm = seastar::metrics;
+        _metrics.add_group("node_maintenance_operations", {
+            sm::make_gauge("bootstrap_total_ranges", sm::description("Number of total ranges for bootstrap operation on this shard."),
+                            bootstrap_total_ranges),
+            sm::make_gauge("bootstrap_finished_ranges", sm::description("Number of finished ranges for bootstrap operation on this shard."),
+                            bootstrap_finished_ranges),
+        });
```
avikivity, line 71: So, each time we initiate a bootstrap, total_ranges will jump, and finished_ranges will climb until it reaches it? And is a boostrap is aborted both will jump down?

repair/repair.cc
```diff
@@ -1386,7 +1404,11 @@ static future<> do_repair_ranges(lw_shared_ptr<repair_info> ri) {
                 ri->ranges_index++;
                 rlogger.info("Repair {} out of {} ranges, id={}, shard={}, keyspace={}, table={}, range={}",
                     ri->ranges_index, ri->ranges.size(), ri->id, ri->shard, ri->keyspace, ri->table_names(), range);
-                return repair_range(*ri, range);
+                return repair_range(*ri, range).then([ri] {
+                    if (ri->reason == streaming::stream_reason::bootstrap) {
+                        _node_ops_metrics.bootstrap_finished_ranges++;
+                    }
+                });
```
avikivity, line 1419: We'll need to undo on failure, in order to preserve consistency with total.

repair/repair.cc
```diff
@@ -1409,6 +1431,10 @@ static future<> do_repair_ranges(lw_shared_ptr<repair_info> ri) {
                         return make_ready_future<stop_iteration>(stop_iteration::yes);
                     }
                 });
+            }).then([ri] {
+                if (ri->reason == streaming::stream_reason::bootstrap) {
+                    _node_ops_metrics.bootstrap_finished_ranges++;
+                }
```
avikivity, line 1447: Maybe you can do `++_node_ops_metrics[reason].finished_ranges` to monitor all types.

repair/repair.cc
```diff
@@ -1658,9 +1684,25 @@ future<> bootstrap_with_repair(seastar::sharded<database>& db, locator::token_me
...
+            auto& ks = db.local().find_keyspace(keyspace_name);
+            auto& strat = ks.get_replication_strategy();
+            dht::token_range_vector desired_ranges = strat.get_pending_address_ranges(tm, tokens, myip);
+            seastar::thread::maybe_yield();
+            nr_ranges_total += desired_ranges.size();
+        }
+        db.invoke_on_all([nr_ranges_total] (database&) {
+            _node_ops_metrics.bootstrap_finished_ranges = 0;
+            _node_ops_metrics.bootstrap_total_ranges = nr_ranges_total;
+        }).get();
```
avikivity, line 1714: I think we should let those keep incrementing. So total_ranges += nr_ranges_total. This way you can look at the history of the metric and see where repairs are initiated and how they complete.

Another thing to consider is the unit. Do we want it to be ranges? Perhaps we should define "1" as the entire ring, so total is incremented by (sum of vnodes / total ring size) and finished is incremented by (range size in tokens / total ring size). (total ring size = 2^64). This normalizes the work to units that are independent of implementation details like ranges.

repair/repair.cc
asias: > So, each time we initiate a bootstrap, total_ranges will jump, and finished_ranges will climb until it reaches it?

Yes. 

>  And is a boostrap is aborted both will jump down?

No. bootstrap_total_ranges never jump down. If the bootstrap is aborted, the only thing we can do is restart the node. It is better not to jump down so the monitor can know the ranges it is supposed to be. Dropping to zero will consume the monitor even more.

repair/repair.cc
asias: On failure, the finished is not increased at all. We have nothing to undo.

repair/repair.cc
asias: Normal repair is more complicated than the node operations. User can fire more repair jobs in parallel. I want to limit those to node operations for now. If we end up we can simplify with the  _node_ops_metrics[reason], we can easily factor this out. OK?

repair/repair.cc
asias: > I think we should let those keep incrementing. So total_ranges += nr_ranges_total. This way you can look at the history of the metric and see where repairs are initiated and how they complete.

Keep increasing the metrics has the problem of one operation mess up with previous ones. For example, first rebuild needed to finish 100 ranges, but it finished only 50 due to errors. The second rebuild needed to finish 200 and finished 200. The always incrementing finished and total number will be wrong.  Well we can increase finished number for the first rebuild when it fails, but we can not tell if it is really finished successfully or not, plus we have to maintain very carefully otherwise the number will be wrong for all following operations. 

To track the history, we can add another metrics. To track progress, I think it is better to "gauge" per operation.

asias: > Another thing to consider is the unit. Do we want it to be ranges? Perhaps we should define "1" as the entire ring, so total is incremented by (sum of vnodes / total ring size) and finished is incremented by (range size in tokens / total ring size). (total ring size = 2^64). This normalizes the work to units that are independent of implementation details like ranges.

The progress = finished / total. The total ring size cancels with each other in the equation. I think number of "ranges" is much easier for user to reason about than the sum of length of of ranges divided by the total ring length:  (len(r1) + len (r2) + ...  len(rn))  / 2^64. 

We have plenty of ranges information in the log too, e.g., in streaming and repair "x out of y  ranges".

At some point, if we ever want to drop the concept of ranges, we can revise those metrics.

repair/repair.cc
avikivity: I thought we can resume bootstrap. But let's consider regular repair. How should total_ranges behave?

repair/repair.cc
avikivity: Whenever finished != total, it means that something is in progress. So we have to do something.

repair/repair.cc
avikivity: Ok, but I do want to make sure the metrics are similar and work for both repair and other node operations.

repair/repair.cc
avikivity: If we make it a gauge then that's fine - we don't need finished/total, just progress.

But let's count tokens, not ranges. So it's tokens_finished/tokens_total. If we have multiple repairs in parallel, it's sum(tokens_finished in each active repair) / sum(tokens_total in each active repair).

asias: Counting tokens do not give any more information or make the result more accurate. For example, we have 3 ranges to work on.  [10,20], [30,40], [50,70].  At he the point the first two ranges are finished:

Th progress counting tokens = 20 - 10 + 40 - 30 /   20 - 10 + 40 - 30 + 70 - 50  =  20 / 40 = 50%.
The progress counting ranges = 2 / 4 = 50%.

Why do we even bother to calculate the length of the ranges (counting by tokens)?


With multiple repair jobs, we can still use sum(ranges finished in each active repair) / sum (ranges total in each active repair).

repair/repair.cc
asias: I agree that it is better to make progress metrics for repair and the rest of the operations look similar.

repair/repair.cc
asias: Two metrics: finished and total can not express the information "failed".  User can know an operation is failed by other means, e.g., api. If the operation is failed, the user  know it is partially done.

repair/repair.cc
asias: We can sum all the finished_ragnes / total_ranges of individual repair.  We have to remove finished_ranges and total_ranges once a individual repair job is finished though.

avikivity: Another option is to use a label with the repair job. That might cause a large number of labels when manager is creating small repair jobs, which isn't good.

repair/repair.cc
avikivity: Ranges are meaningless for the user, because their size is unclear.

If we divide ranges, they cancel out, but it only works if the ranges are of the same size. In that case it's the same as dividing tokens: completed_ranges/total_tanges = (completed_ranges*tokens_per_range)/(total_ranges*tokens_per_range) = completed_tokens/total_tokens.

asias: I think ranges do have meaning. It presents how many fragments of the token ranges the operations have to work on. The progress is a rough estimation anyway. Even if we use count len(range.end() - range.start()), it is still a rough estimation because of the actual data distribution. 

If we agree we only provide the progress metric (as percentage), it makes the argument to count the len(range.end() - range.start()) or number of ranges even less important, because only the percentage is exposed to user. 

Btw, we do not have the "length"  in range class. 

```
510     const optional<bound>& start() const {
511         return _interval.start();
512     }
513     const optional<bound>& end() const {
514         return _interval.end();
515     }
```

asias: You are right about the same size thing. Another example [10,20], [30,40], [50,50000] shows it clearly.

repair/repair.cc
avikivity: We will have to set the metrics to something. We can drop them both to zero after the operation completes (one way or the other).

repair/repair.cc
avikivity: We don't have length in intervals because intervals might not be numberic (and might be open). You have to extract the bounds and compute them yourself.

Alternatively we can add an optional member function with constraints, see https://godbolt.org/z/1xKc89.

asias: Considering the little gain of computing "length" v.s. number of ranges. I still think it is better and good enough to use ranges_finished.size() / ranges_total.size(). It is simpler: it is O(1), no need to calculate "length" of all ranges. No dependence on the interval computing. 

What do you think?

repair/repair.cc
asias: Exactly.

repair/repair.cc
asias: Setting to zero after the operation makes it is hard to use the metrics. It is hard to tell if the operation has finished or hasn't started yet.

avikivity: Well, we can't leave it at a random place. The graphs well look terrible.

repair/repair.cc
avikivity: Depends on the variation between range sizes. If there is high variation, we will get poor progress tracking.

If there is low variation, it is good enough.

repair/repair.cc
asias: It is not a random place. It is a place where either it stopped progressing due to error or it is 100% in the case of successful operation.

repair/repair.cc
asias: The real progress depends on the data distribution over those ranges as well. This data distribution can have very big variation too. Considering this factor, why not keep it simple. 


Btw, Avi, do you insist we should only expose the progress percentage through the metrics or it is fine to expose both finished and total?

repair/repair.cc
avikivity: Those graphs that are stuck in some place just look bad (this example looks especially bad because I drew it)

Drawing

It looks like a bug.

asias: LOL. 

You wanted this (dropping to zero).

asias: Dropping to zero has the issue the dashboard hasn't sampled the 100% progress while the progress dropped to zero already. We have false unfinished progress in the graph.

avikivity: Well, you're a better drawer than me. You're right about not sampling 100%.

Maybe @amnonh can add annotations about repair start and repair end/fail to the graph.

asias:
Changes in e53e68cc2e7d5f7c7e63fab046ee654062408488:
- Expose only the finished percentage instead of number of ranges
- Add metrics for regular repair which is similar to regular ops as well.

repair/repair.cc
avikivity: > The real progress depends on the data distribution over those ranges as well. This data distribution can have very big variation too. Considering this factor, why not keep it simple.

Usually a non-tiny token range has good distibution. If it doesn't, we can't do anything about it. If it does, let's be as accurate as we can be.

> 
> Btw, Avi, do you insist we should only expose the progress percentage through the metrics or it is fine to expose both finished and total?

I dislike percentages and prefer exposing more direct metrics, and to do the percentage computation on the client side. So I prefer finished/total, provided we can get reasonable values for them when repair is not running.

avikivity:
> Changes in [e53e68c](https://github.com/scylladb/scylla/commit/e53e68cc2e7d5f7c7e63fab046ee654062408488):
> 
> * Expose only the finished percentage instead of number of ranges
> * Add metrics for regular repair which is similar to regular ops as well.

Sorry, I missed your last comment. I do not mean to force an opinion by not responding, it's just oversight.

repair/repair.cc
asias: Do you insist on we must compute "length" of a range than size of the range (ranges.size())? I can give you another example. Say we have 2 keyspaces, one has finished 9 out of 10 ranges, another also has finished 1 out of 10 ranges. But one has 1GB of data while the other has 1TB of data. We will get progress =  9 + 1 / 10 + 10 = 50%. Apparently, it hasn't finished 50% of data yet. We will get similar 50% progress even if we compute "lengh" of a range.

asias: > I dislike percentages and prefer exposing more direct metrics, and to do the percentage computation on the client side. So I prefer finished/total, provided we can get reasonable values for them when repair is not running.

For regular repair, in the latest version, I am combining all ongoing repair jobs. When a repair job is finished it will be removed from the calculation. With percentage, it  gives "1" when there is no repair running. With finish/total, we will return 0 and 0.

asias:
> > Changes in [e53e68c](https://github.com/scylladb/scylla/commit/e53e68cc2e7d5f7c7e63fab046ee654062408488):
> > 
> > * Expose only the finished percentage instead of number of ranges
> > * Add metrics for regular repair which is similar to regular ops as well.
> 
> Sorry, I missed your last comment. I do not mean to force an opinion by not responding, it's just oversight.

OK. I will change it back to finished / total then.

repair/repair.cc
asias: For node operations, when it is finished, we keep the finish/total value, so the client can sample it correctly. 

Exposing percentage has the advantage of hiding the slight differences between regular and node operations and unit of ranges.

avikivity: But the graph looks bad. Bad looking graphs are not the most important motivation, but if we can find a good way to expose the information and have a nice graph, we should do it.

avikivity:
> > > Changes in [e53e68c](https://github.com/scylladb/scylla/commit/e53e68cc2e7d5f7c7e63fab046ee654062408488):
> > > 
> > > * Expose only the finished percentage instead of number of ranges
> > > * Add metrics for regular repair which is similar to regular ops as well.
> > 
> > 
> > Sorry, I missed your last comment. I do not mean to force an opinion by not responding, it's just oversight.
> 
> OK. I will change it back to finished / total then.

I'm not saying I like finished/total either (with the proposed way of letting them keep their value after repair is finished). We need to find a good way.

asias:
> > > > Changes in [e53e68c](https://github.com/scylladb/scylla/commit/e53e68cc2e7d5f7c7e63fab046ee654062408488):
> > > > 
> > > > * Expose only the finished percentage instead of number of ranges
> > > > * Add metrics for regular repair which is similar to regular ops as well.
> > > 
> > > 
> > > Sorry, I missed your last comment. I do not mean to force an opinion by not responding, it's just oversight.
> > 
> > 
> > OK. I will change it back to finished / total then.
> 
> I'm not saying I like finished/total either (with the proposed way of letting them keep their value after repair is finished). We need to find a good way.

Avi, IMHO, the solution I presented in patch  https://github.com/scylladb/scylla/pull/6842#issuecomment-666046812 is the best one so far I can think of. It unifies both the regular repair and repair based ops. When finished the progress is 100%, so it solves the "client sample" issue. User does not need to interpret what finished/total actually means. It also handles multiple repair jobs correctly.

avikivity:
> 
> Avi, IMHO, the solution I presented in patch [#6842 (comment)](https://github.com/scylladb/scylla/pull/6842#issuecomment-666046812) is the best one so far I can think of. It unifies both the regular repair and repair based ops. When finished the progress is 100%, so it solves the "client sample" issue. User does not need to interpret what finished/total actually means. It also handles multiple repair jobs correctly.

Ok. I'm not totally happy with it but we can go with that.

asias:
> > Avi, IMHO, the solution I presented in patch [#6842 (comment)](https://github.com/scylladb/scylla/pull/6842#issuecomment-666046812) is the best one so far I can think of. It unifies both the regular repair and repair based ops. When finished the progress is 100%, so it solves the "client sample" issue. User does not need to interpret what finished/total actually means. It also handles multiple repair jobs correctly.
> 
> Ok. I'm not totally happy with it but we can go with that.

OK. Can you please merge the latest version:

Changes in version b9de82e6218f2b2c61b19eab739def372feba898
- Rebased to master branch

PR #6843, opened 2020-07-15
repair: Add uuid to a repair job

### LINKED ISSUE #6786: Add uuid to a repair job

asias:
Currently, repair uses an integer to identify a repair job. The repair
id starts from 1 since node restart. As a result, different repair jobs
will have same id across restart.

To make the id more unique across restart, we can use an uuid in
addition to the integer id. We can not drop the use of the integer id
completely since the http api and nodetool use it.

Fixes #6786

### DIFF
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -87,6 +87,10 @@ static std::vector<sstring> list_column_families(const database& db, const sstri
     return ret;
 }
 
+std::ostream& operator<<(std::ostream& os, const repair_uniq_id& x) {
+    return os << format("[id={}, uuid={}]", x.id, x.uuid);
+}
+
 // Must run inside a seastar thread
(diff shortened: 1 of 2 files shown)


avikivity:
I remember I had comments before, but I don't see them, and there is no changelog from the previous version.

asias:
> I remember I had comments before, but I don't see them, and there is no changelog from the previous version.

Sorry about that. I mentioned the change in the previous PR here: https://github.com/scylladb/scylla/pull/6787#discussion_r454731405

Anyway, changes in cdb8e3d05941701e44f0b0fb8b563f096f738c61:
- Add comments to explain id and uuid in repair_uniq_id 
- Rebase to master.

Issue #6852, opened 2020-07-16
Test failure: cql/cdc_allow_dropping_manually_created_log_test


piodul:
Regression was introduced in commit 61f52da9b1c53da6dad25cc90f22a98df56e752c

The `cql/cdc_allow_dropping_manually_created_log_test` fails with the following message:

```
Test cql/cdc_allow_dropping_manually_created_log_test (dev) failed: test output does not match expected result
--- test/cql/cdc_allow_dropping_manually_created_log_test.result	Tue Apr 14 16:51:04 2020
+++ test/cql/cdc_allow_dropping_manually_created_log_test.reject	Thu Jul 16 10:35:13 2020
@@ -1,9 +1,10 @@
 create table ks.test_scylla_cdc_log (pk int primary key);
 {
 	"status" : "ok"
 }
 -- Should be possible to drop it
 drop table ks.test_scylla_cdc_log;
 {
-	"status" : "ok"
+	"message" : "std::_Nested_exception<no_such_column_family> (Can't find a column family test in keyspace ks)",
+	"status" : "error"
 }

```

haaawk:
@nyh it might make sense to revert the alternator streams change and wait for @elcallio to fix it.

piodul:
@haaawk I'm not sure it will be necessary - I identified the problematic place in code and will submit a patch in a second.

nyh:
No need to backport - this fixes a regression introduced by Alternator Streams, which we never backported to any release.

PR #6854, opened 2020-07-16
cdc: fix a corner case inside get_base_table

piodul:
It is legal for a user to create a table with name that has a
_scylla_cdc_log suffix. In such case, the table won't be treated as a
cdc log table, and does not require a corresponding base table to exist.

During refactoring done as a part of initial implemetation of of
Alternator streams (#6694), `is_log_for_some_table` started throwing
when trying to check a name like `X_scylla_cdc_log` when there was no
table with name `X`. Previously, it just returned false.

The exception originates inside `get_base_table`, which tries to return
the base table schema, not checking for its existence - which may throw.
It makes more sense for this function to return nullptr in such case (it
already does when provided log table name does not have the cdc log
suffix), so this patch adds an explicit check and returns nullptr when
necessary.

A similar oversight happened before (see #5987), so this patch also adds
a comment which explains why existence of `X_scylla_cdc_log` does not
imply existence of `X`.

Fixes: #6852
Refs: #5724, #5987

Tests: unit(dev)

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -371,7 +371,16 @@ schema_ptr get_base_table(const database& db, sstring_view ks_name,std::string_v
     if (!is_log_name(table_name)) {
         return nullptr;
     }
-    return db.find_schema(sstring(ks_name), base_name(table_name));
+    // Note: It is legal for a user to directly create a table with name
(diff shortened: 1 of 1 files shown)


nyh:
Thanks. Committed to next: eb991827d797f1d6168143527120bc52db3aa228

PR #6857, opened 2020-07-16
configure.py: $mode-test targets depend on scylla

### LINKED ISSUE #6855: Ninja target `dev-test` should rebuild target `scylla`
One of the tests that `dev-test` runs is [alternator/run](../blob/master/test/alternator/run), which stands up a Scylla server to query.  This means that `dev-test` should rebuild `build/dev/scylla`, but it doesn't.

dekimir:
The targets {dev|debug|release}-test run all unit tests, including
alternator/run.  But this test requires the Scylla executable, which
wasn't among the dependencies.  Fix it by adding build/$mode/scylla to
the dependency list.

Fixes #6855.

Tests: `ninja dev-test` after removing build/dev/scylla

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1608,7 +1608,7 @@ with open(buildfile_tmp, 'w') as f:
         )
 
         f.write(
-            'build {mode}-test: test.{mode} {test_executables} $builddir/{mode}/test/tools/cql_repl\n'.format(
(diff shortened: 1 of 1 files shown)


penberg:
Queued as 2c4a41f017d7eaaedb0792df5186fca31191b099

PR #6859, opened 2020-07-16
schema: take into account features when converting a table creation to schema_mutations

eliransin:
When upgrading from a version that lacks some schema features,
during the transition, when we have a mixed cluster. Schema digests
are calculated without taking into account the mixed cluster supported
features. Every node calculate the digest as if the whole cluster supports
its supported features.
Scylla already has a mechanism of redaction to the lowest common
denominator, but it haven't been used in this context.

This commit is using the redaction mechanism when calculating the digest on
the newly added table so it will match the supported features of the
whole cluster.

Tests: Manual upgrading - upgraded to a version with an additional
feature and additional schema column and validated that the digest
of the tables schema is identical on every node on the mixed cluster.

### DIFF
--- a/db/schema_tables.cc
+++ b/db/schema_tables.cc
@@ -779,7 +779,9 @@ read_schema_partition_for_table(distributed<service::storage_proxy>& proxy, sche
             .with_range(std::move(clustering_range))
             .build();
     auto cmd = make_lw_shared<query::read_command>(schema->id(), schema->version(), std::move(slice), query::max_rows);
-    return query_partition_mutation(proxy.local(), std::move(schema), std::move(cmd), std::move(keyspace_key));
+    return query_partition_mutation(proxy.local(), std::move(schema), std::move(cmd), std::move(keyspace_key)).then([&proxy] (mutation mut) {
+         return redact_columns_for_missing_features(std::move(mut), proxy.local().get_db().local().features().cluster_schema_features());
+    });
 }
 
 future<mutation>


eliransin:
@nyh FYI.

avikivity:
Looks reasonable. @tgrabiec please take a look as well.

db/schema_tables.cc
```diff
@@ -779,7 +779,9 @@ read_schema_partition_for_table(distributed<service::storage_proxy>& proxy, sche
             .with_range(std::move(clustering_range))
             .build();
     auto cmd = make_lw_shared<query::read_command>(schema->id(), schema->version(), std::move(slice), query::max_rows);
-    return query_partition_mutation(proxy.local(), std::move(schema), std::move(cmd), std::move(keyspace_key));
+    return query_partition_mutation(proxy.local(), std::move(schema), std::move(cmd), std::move(keyspace_key)).then([&proxy] (mutation mut) {
+         return redact_columns_for_missing_features(std::move(mut),proxy.local().get_db().local().features().cluster_schema_features());
```
tgrabiec, line 783: missing space after `,`

eliransin: fixed

eliransin:
@avikivity fixed coding style.

haaawk:
I may be wrong but I don't think this PR is correct. @avikivity @tgrabiec please correct me if I'm wrong but my understanding was that in a mixed cluster new nodes are using the new schema and just redacting it for inter-node communication. This change redacts schema when it's loaded from disk at startup. How is the schema ever becoming a new schema now?

haaawk:
Why do we need to redact "in this context" and what is "this context" mentioned in the commit message?

PR #6870, opened 2020-07-17
auth: Drop needless role-manager check

dekimir:
The service constructor included a check ensuring that only
standard_role_manager can be used with password_authenticator. But
after 00f7bc6, password_authenticator does not depend on any action of
standard_role_manager. All queries to meta::roles_table in
password_authenticator seem self-contained: the table is created at
the start if missing, and salted_hash is CRUDed independently of any
other columns bar the primary key role_col_name.

NOTE: a nonstandard role manager may not delete a role's row in
meta::roles_table when that role is dropped. This will result in
successful authentication for that non-existing role. But the clients
call check_user_can_login() after such authentication, which in turn
calls role_manager::exists(role). Any correctly implemented role
manager will then return false, and authentication_exception will be
thrown. Therefore, no dependencies exist on the role-manager
behaviour, other than it being self-consistent.

Tests: unit (dev)

### DIFF
--- a/auth/service.cc
+++ b/auth/service.cc
@@ -31,9 +31,7 @@
 #include "auth/allow_all_authenticator.hh"
 #include "auth/allow_all_authorizer.hh"
 #include "auth/common.hh"
-#include "auth/password_authenticator.hh"
 #include "auth/role_or_anonymous.hh"
-#include "auth/standard_role_manager.hh"
 #include "cql3/query_processor.hh"
 #include "cql3/untyped_result_set.hh"
 #include "db/consistency_level_type.hh"
@@ -125,18 +123,7 @@ service::service(
             , _authorizer(std::move(z))
             , _authenticator(std::move(a))
             , _role_manager(std::move(r))
-            , _migration_listener(std::make_unique<auth_migration_listener>(*_authorizer)) {
-    // The password authenticator requires that the `standard_role_manager` is running so that the roles metadata table
-    // it manages is created and updated. This cross-module dependency is rather gross, but we have to maintain it for
-    // the sake of compatibility with Apache Cassandra and its choice of auth. schema.
-    if ((_authenticator->qualified_java_name() == password_authenticator_name)
-            && (_role_manager->qualified_java_name() != standard_role_manager_name())) {
-        throw incompatible_module_combination(
-                format("The {} authenticator must be loaded alongside the {} role-manager.",
-                        password_authenticator_name,
-                        standard_role_manager_name()));
-    }
-}
+            , _migration_listener(std::make_unique<auth_migration_listener>(*_authorizer)) {}
 
 service::service(
         permissions_cache_config c,


auth/service.cc
```diff
@@ -125,18 +123,7 @@ service::service(
             , _authorizer(std::move(z))
             , _authenticator(std::move(a))
             , _role_manager(std::move(r))
-            , _migration_listener(std::make_unique<auth_migration_listener>(*_authorizer)) {
-    // The password authenticator requires that the `standard_role_manager` is running so that the roles metadata table
-    // it manages is created and updated. This cross-module dependency is rather gross, but we have to maintain it for
-    // the sake of compatibility with Apache Cassandra and its choice of auth. schema.
-    if ((_authenticator->qualified_java_name() == password_authenticator_name())
```
avikivity, line 132: Isn't the comment correct?

```c++
future<> password_authenticator::alter(std::string_view role_name, const authentication_options& options) const {
    if (!options.password) {
        return make_ready_future<>();
    }

    static const sstring query = format("UPDATE {} SET {} = ? WHERE {} = ?",
            meta::roles_table::qualified_name(),
            SALTED_HASH,
            meta::roles_table::role_col_name);

    return _qp.execute_internal(
            query,
            consistency_for_user(role_name),
            internal_distributed_timeout_config(),
            {passwords::hash(*options.password, rng_for_salt), sstring(role_name)}).discard_result();
}
```

It looks like a true depdendency.

dekimir: The comment is not correct, as the dependency is on _tables_, not on a particular role manager.  Any role manager that creates these tables should work fine.

Furthermore, the code reading scylla.yaml will abort on any role-manager name except standard_role_manager, so the `then` part of this `if` will never execute.

avikivity: > The comment is not correct, as the dependency is on _tables_, not on a particular role manager. Any role manager that creates these tables should work fine.

So if you have a role_manager that uses the same tables, add it to the list. As long as it's possible to have a role_manager that doesn't use these tables, the dependency exists.

> 
> Furthermore, the code reading scylla.yaml will abort on any role-manager name except standard_role_manager, so the `then` part of this `if` will never execute.

I guess this was inherited from Cassandra, which supports user-provided role managers. Still, I think the check is correct (if inelegant). If role_manager is an interface, but not all role managers can be used with password_authenticator, we need some kind of check.

We can make it less ugly by checking role_manager::uses_roles_table(). Then a role_manager can declare if it is compatible with password_authenticator or not.

dekimir: Actually, after 00f7bc6 I don't see any dependences in password_authenticator on any action of standard_role_manager.  All queries to meta::roles_table in password_authenticator seem self-contained to me: the table is created at the start if missing, and salted_hash is CRUDed independently of any other columns bar the primary key role_col_name.

(It is true that `DROP ROLE` with a nonstandard role manager may not delete the role's row in meta::roles_table, resulting in successful authentication for that role.  But the clients call check_user_can_login() after such authentication, which in turn calls role_manager::exists(role).  A correctly implemented role manager will then return false, and authentication_exception will be thrown.)

Therefore, I posit that no dependencies exist on the role-manager behaviour, other than it being self-consistent.

avikivity: Makes sense.

avikivity: Please update the commit changelog with the details, for posterity.

dekimir: > Please update the commit changelog with the details, for posterity.

Done.  Also updated the top PR comment.

dekimir:
Also rebased on current master. @avikivity PTAL

PR #6872, opened 2020-07-19
build: don't package tools/java and tools/jmx in relocatable pacakge

avikivity:
tools/java and tools/jmx have their own relocatable packages (and rpm/deb),
so they should not be part of the main relocatable package.

Enforce this by enabling the filter parameter in reloc_add, and passing
a filter that excludes tools/java and tools/jmx.


penberg:
Queued as 19da4a5b8ff5a0ebdfa22bfe9b63b16f0267c120.

PR #6873, opened 2020-07-19
dist: redhat: reduce log spam from unpacking sources when building rpm

avikivity:
rpmbuild defaults to logging the name of every file it unpacks from
the archive.

Make it quiet with the %setup -q flag.


penberg:
Queued as 98a22e572a38a2f4c9447be948698b8e16743465

PR #6881, opened 2020-07-20
alternator::streams: Do not allow enabling streams when CDC is off

### LINKED ISSUE #6866: alternator streams API should return a clear error if CDC is not enabled
Currently, CDC is an experimental feature and needs to be explicitly enabled. The Alternator Streams API implementation currently just blindly uses CDC without checking if it is enabled or not, resulting in some obscure errors if it is not.
For example, in `test_streams.py::test_describe_stream` we see that the DescribeStream call succeeds, but is missing a the necessary attributes (e.g., StreamLabel) in its response. Some other tests appear to succeed despite CDC not being enabled...

What we should have seen is the `DescribeStream` call - and the three other Streams API calls - should immediately return an error if running without CDC enabled. The error message - which the user may see - should explain clearly that the Streams API feature is experimental, and to enable it you must enable the experimental CDC feature.

CC @elcallio

elcallio:
Fixes #6866

If we try to create/alter an Alternator table to include streams,
we must check that the cluster does in fact support CDC
(experimental still). If not, throw a hopefully somewhat descriptive
error.
(Normal CQL table create goes through a similar check in cql_prop_defs)

Note: no other operations are prohibited. The cluster could have had CDC
enabled before, so streams could exist to list and even read.
Any tables loaded from schema tables should be reposnsible for their
own validation.

### DIFF
--- a/alternator/executor.hh
+++ b/alternator/executor.hh
@@ -141,7 +141,7 @@ public:
 
 
 
-    static void add_stream_options(const rjson::value& stream_spec, schema_builder&);
+    void add_stream_options(const rjson::value& stream_spec, schema_builder&);
 };
 
 }
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -36,6 +36,8 @@
 #include "cql3/result_set.hh"
 #include "cql3/type_json.hh"
 #include "schema_builder.hh"
+#include "gms/feature.hh"
+#include "gms/feature_service.hh"
 
 #include "executor.hh"
 #include "tags_extension.hh"
@@ -862,6 +864,12 @@ void executor::add_stream_options(const rjson::value& stream_specification, sche
     }
 
     if (stream_enabled->GetBool()) {
+        auto& db = _proxy.get_db().local();
+
+        if (!db.features().cluster_supports_cdc()) {
+            throw api_error::validation("StreamSpecification: streams (CDC) feature not enabled in cluster.");
+        }
+
         cdc::options opts;
         opts.enabled(true);
         auto type = rjson::get_opt<stream_view_type>(stream_specification, "StreamViewType").value_or(stream_view_type::KEYS_ONLY);


elcallio:
I'll just wait and rebase.

elcallio:
@nyh ping. rebased.

alternator/executor.hh
```diff
@@ -141,7 +141,7 @@ public:
 
 
 
-    static void add_stream_options(const rjson::value& stream_spec, schema_builder&);
+    void add_stream_options(const rjson::value& stream_spec, schema_builder&);
```
haaawk, line 144: Shouldn't this be private?

alternator/streams.cc
```diff
@@ -862,6 +864,12 @@ void executor::add_stream_options(const rjson::value& stream_specification, sche
     }
 
     if (stream_enabled->GetBool()) {
+        auto& db = _proxy.get_db().local();
+
+        if (!db.features().cluster_supports_cdc()) {
+            throw api_error::validation("StreamSpecification: streams (CDC) feature not enabled in cluster.");
```
haaawk, line 870: Super Nit: s/in cluster/in the cluster/

elcallio: Actually, no. The use of "the" is "spoken"/"natural" langague. Error reporting should (imho) use the above unspecific form.

haaawk: I'm not sure I follow. So you're saying that "the" is not used in "formal" or "written" language?

Anyways, if you feel like the current version is ok then I don't have any strong feelings about it.

elcallio: No, I am saying error reporting uses mostly unqualified terms. Like "file unreadable". not "the file is unreadable".

haaawk: Ok.

elcallio:
ping?

slivne:
@nyh ping

nyh:
Rebased (conflicted with 05851578d4f12627734bbed30193795bc17ae53a) and committed to next: a978e043c3152999807fa4e4e8dab82dc50cf985

PR #6887, opened 2020-07-21
docker: Make I/O configuration setup configurable

### LINKED ISSUE #6587: docker: Running scylla_io_setup does not work and is not needed in some cases
*Installation details*
Scylla version (or git commit hash): 4.1 and 4.0

In [scyllasetup.py](https://github.com/scylladb/scylla/blob/branch-4.1/dist/docker/redhat/scyllasetup.py#L62) we now run scylla_io_setup, one needs to note that it requires running with privileges and some setup to make it work. I suggest to disable that if developer mode is enabled.

penberg:
This adds a '--io-setup N' command line option, which users can pass to
specify whether they want to run the "scylla_io_setup" script or not.
This is useful if users want to specify I/O settings themselves in
environments such as Kubernetes, where running "iotune" is problematic.

While at it, add the same option to "scylla_setup" to keep the interface
between that script and Docker consistent.

Fixes #6587

### DIFF
--- a/dist/common/scripts/scylla_setup
+++ b/dist/common/scripts/scylla_setup
@@ -70,7 +70,7 @@ def print_non_interactive_suggestion_message(args):
         outlines[-1] += x
 
     print(' \\\n'.join(outlines))
-    if args.no_io_setup == False:
+    if args.io_setup:
         print('\nAlso, to avoid the time-consuming I/O tuning you can add --no-io-setup and copy the contents of /etc/scylla.d/io*')
         print('Only do that if you are moving the files into machines with the exact same hardware')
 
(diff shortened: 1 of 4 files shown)


avikivity:
Please use positive parameters, e.g. --io-setup 1 / --io-setup 0.

slivne:
Avi its parallel to scylla_setup ...

I think its better to use the same flags in scylla_setup and in
docker command line

avikivity:
If we make a mistake, we have to repeat it?

slivne:
If we want users to be able to use it - lets make their life easier ...

We can add this to scylla_setup as well as an additional option - if this
is your view.

We should not have different flags for docker and scylla_setup - users will
struggle moving from one to the other.

avikivity:
We can add --io-setup to setup, if we want.

penberg:
@avikivity Changed the option to `--io-setup` and updated `scylla_setup` to support that too.

Issue #6864, opened 2020-07-16
alternator streams write fails when attempted too early


nyh:
The test `test_streams.py::test_get_records`, when run alone with a command line such as
```
test/alternator/run --runxfail test_streams.py::test_get_records
```
seems to always fail. The relevant error lines:
```
>           table.put_item(Item={'p': p, 'c': c, 'a1': val, 'a2': val2})
test_streams.py:318: 
E           botocore.exceptions.ClientError: An error occurred (Internal Server Error) when calling the PutItem operation (reached max retries: 3): Internal server error: std::runtime_error (cdc::metadata::get_stream: could not find any CDC stream (current time: 2020/07/16 17:49:53). Are we in the middle of a cluster upgrade?)
```

Because of these failures I marked this test "xfail".

However, if the entire Alternator test suite is run (just `test/alternator/run` without parameters), and `test_streams.py` comes near the end, this test always passes! (i.e., it is listed as an XPASS in the complete run).

I suspect that this is some sort of CDC initialization timing issue - perhaps the test cannot succeed if run before CDC finished to initialize itself. If this is the case, we will need to add to the relevant Alternator Streams API requests the code to wait until CDC initialization is complete, instead of failing with bizarre exceptions.

CC @elcallio

elcallio:
It is most definitely init issues. @kbr - we apparently can both create a table with CDC log enabled and start writing to it, before the CDC streams set has been created. In this case we get the above exception when trying to get streams in `log.cc:begin_timestamp`. Should not something somewhere in this path wait for the cluster to have CDC finished enabled? I cannot see anything to really wait on? Am I missing something?

elcallio:
Verified that we do in fact have a generation, but it is not valid for the timestamp in the write.

haaawk:
Cool. @kbr- Could you please implement one of the solutions we discussed on the call? Either a check that the generation contains only streams for a single node or changing delay from 2*ring delay + epsilon to 2 * ring delay.

elcallio:
I think we should also consider letting writes to CDC enabled tables silently fail adding CDC log, but succeed the main write, if streams are not available yet. This would be neatly consistent with dynamos concept of streams being in a "enabling" status for a while until they actually collect data.

haaawk:
I guess we could special case this just for dynamo streams

elcallio:
[...]

PR #6901, opened 2020-07-22
alternator::streams: Report streams as not ready until CDC stream id:…

elcallio:
…s are available

Refs #6864

When booting a clean scylla, CDC stream ID:s will not be availble until
a n*ring delay time period has passed. Before this, writing to a CDC
enabled table will fail hard.
For alternator (and its tests), we can report the stream(s) for tables as not yet
available (ENABLING) until such time as id:s are
computed.

### DIFF
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -36,6 +36,7 @@
 #include "cql3/result_set.hh"
 #include "cql3/type_json.hh"
 #include "schema_builder.hh"
+#include "service/storage_service.hh"
 
 #include "executor.hh"
 #include "tags_extension.hh"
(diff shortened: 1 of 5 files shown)


alternator/streams.cc
```diff
@@ -390,7 +391,18 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
 
     auto& opts = bs->cdc_options();
 
-    rjson::set(stream_desc, "StreamStatus", rjson::from_string(opts.enabled() ? "ENABLED"s : "DISABLED"s));
+    auto status = "DISABLED";
+
+    if (opts.enabled()) {
+        auto& metadata = service::get_local_storage_service().get_cdc_metadata();
```
kbr-, line 397: Globals!
Avi once told me in review that we're in war against globals, so we should not be adding more global variable references. We should be removing them.
I'm passing this info on.

elcallio:
Once storage service is reachable from a non-global location I'll be happy to remove the use of the global. But it is not.

kbr-:
Could this `executor` thingie store a reference to `storage_service`?
It does store `_proxy`.

elcallio:
Yes, but it would have to initiate it from the same global. Because there literally is no owner/holder of `storage_service`. So you would just trade one ref of global for another. Imho not much of an improvement.

kbr-:
We would pass the responsibility of obtaining `storage_service&` to main.cc, which is top-level. Eventually, we'd like main.cc to create and initialize the service, and then appropriately destroy it after it makes sure that nobody is using it. But a prerequisite is that nobody accesses `storage_service` through a global variable. So why not make future refactoring easier?

elcallio:
Fair enough. More members means more fun. New version with `storage_service` ref as member and the global ref restricted to main.

elcallio:
@nyh, ping?

slivne:
@nyh ping

nyh:
Thanks, committed to next 05851578d4f12627734bbed30193795bc17ae53a

PR #6902, opened 2020-07-22
cdc: if ring_delay == 0, don't add delay to newly created generation

kbr-:
If ring_delay == 0, something fishy is going on, e.g. single-node tests
are being performed. In this case we want the CDC generation to start
operating immediately. There is no need to wait until it propagates to
the cluster.

You should not use ring_delay == 0 in production.

Fixes https://github.com/scylladb/scylla/issues/6864.

### DIFF
--- a/cdc/generation.cc
+++ b/cdc/generation.cc
@@ -256,12 +256,13 @@ db_clock::time_point make_new_cdc_generation(
         db::system_distributed_keyspace& sys_dist_ks,
         std::chrono::milliseconds ring_delay,
         bool for_testing) {
+    using namespace std::chrono;
     auto gen = topology_description_generator(cfg, bootstrap_tokens, tm, g).generate();
 
     // Begin the race.
     auto ts = db_clock::now() + (
-            for_testing ? std::chrono::milliseconds(0) : (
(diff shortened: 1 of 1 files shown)


nyh:
The problem I have with this solution is that the usage of `--ring-delay-ms 0`  in `test/alternator/run` was supposed to be an optimization, not affect correctness. With this patch, removing the `--ring-delay-ms 0` will not just make it slower (as expected), but also cause one test to potentially fail.

But that being said, I'm not against this patch and am ok with it being committed. It is correct that on a single-node setup no delay is needed at all - not 2*ringdelay and not 5s, so in that sense the patch is correct. It will just not solve the same problem for actual distributed tests, which will need to artificially wait for this delay (and as you said, you actually did this in dtest).

kbr-:
> With this patch, removing the --ring-delay-ms 0 will not just make it slower (as expected), but also cause one test to potentially fail.

Ok, so in the general case the test should query the `cdc_streams` table and wait for the first generation to start, like dtests do.
The optimization will then just reduce this wait to `0`.

elcallio:
The tests actually already does this, however, alternator does not answer correctly without #6901. Once that is in, the above is not correctness, just optimization (less waiting).

nyh:
Thanks. Committed to next: test/alternator/run --runxfail test_streams.py::test_get_record

PR #6910, opened 2020-07-23
Increase row limits

wmitros:
This patch enables selecting more than 2^32 rows from a table. The change becomes active after upgrading whole cluster - until then old limits are used.
    
Tested reading 4.5*10^9 rows from a virtual table, manually upgrading a cluster with ccm and performing cql SELECT queries during the upgrade, ran unit tests in dev mode and cql and paging dtests.


cql3/statements/select_statement.cc
```diff
@@ -262,6 +264,9 @@ uint64_t select_statement::do_get_limit(const query_options& options, ::shared_p
         if (l <= 0) {
             throw exceptions::invalid_request_exception("LIMIT must be strictly positive");
         }
+        if (!supports_large_paging_state && l > std::numeric_limits<int32_t>::max()) {
+            throw exceptions::invalid_request_exception("LIMIT value out of range");
```
haaawk, line 268: A more informative error message would be better. Maybe sth like "Cluster does not support 64-bit LIMIT value: <actual value>"

alternator/executor.cc
```diff
@@ -3190,7 +3190,7 @@ future<executor::request_return_type> executor::query(client_state& client_state
                 "Consistent reads are not allowed on global indexes (GSI)"));
     }
     rjson::value* limit_json = rjson::find(request, "Limit");
-    uint32_t limit = limit_json ? limit_json->GetUint64() : query::max_partitions;
+    uint64_t limit = limit_json ? limit_json->GetUint64() : query::partition_max_rows;
```
haaawk, line 3193: Is this intentional to use different variable here?

cql3/result_generator.hh
```diff
@@ -114,7 +114,7 @@ private:
             }
         }
 
-        uint32_t rows_read() const { return _total_row_count; }
+        uint64_t rows_read() const { return _total_row_count; }
```
haaawk, line 117: Nit: Just an idea. Maybe it would make sense to define the return type of this function as auto to prevent the need of changing it in the future?

cql3/selection/selection.hh
```diff
@@ -279,7 +279,7 @@ public:
         }
         void reset(const partition_key* = nullptr) {
         }
-        uint32_t get_rows_dropped() const {
+        uint64_t get_rows_dropped() const {
```
haaawk, line 282: Nit: auto maybe?

cql3/selection/selection.hh
```diff
@@ -290,25 +290,25 @@ public:
...
                 schema_ptr schema,
-                uint32_t per_partition_limit,
+                uint64_t per_partition_limit,
                 std::optional<partition_key> last_pkey = {},
-                uint32_t rows_fetched_for_last_partition = 0);
+                uint64_t rows_fetched_for_last_partition = 0);
         bool operator()(const selection& selection, const std::vector<bytes>& pk, const std::vector<bytes>& ck, const query::result_row_view& static_row, const query::result_row_view* row) const;
         void reset(const partition_key* key = nullptr);
-        uint32_t get_rows_dropped() const {
+        uint64_t get_rows_dropped() const {
```
haaawk, line 311: Nit: maybe auto?

cql3/statements/cas_request.cc
```diff
@@ -101,7 +101,7 @@ lw_shared_ptr<query::read_command> cas_request::read_command() const {
         ranges.reserve(op.ranges.size());
         std::copy(op.ranges.begin(), op.ranges.end(), std::back_inserter(ranges));
     }
-    uint32_t max_rows = query::max_rows;
+    uint64_t max_rows = query::partition_max_rows;
```
haaawk, line 104: Is it intentionally changed from max_rows to partition_max_rows?

cql3/statements/select_statement.cc
```diff
@@ -304,7 +304,7 @@ select_statement::do_execute(service::storage_proxy& proxy,
 
     validate_for_read(cl);
 
-    int32_t limit = get_limit(options);
+    int64_t limit = get_limit(options);
```
haaawk: get_limit returns uint64_t - would it make sense to define limit as unsigned as well?

cql3/statements/select_statement.cc
```diff
@@ -511,15 +511,15 @@ indexed_table_select_statement::do_execute_base_query(
...
+        base_query_state(uint64_t row_limit, service::query_ranges_to_vnodes_generator&& ranges_to_vnodes_)
                 : merger(row_limit, query::max_partitions)
                 , ranges_to_vnodes(std::move(ranges_to_vnodes_))
                 {}
         base_query_state(base_query_state&&) = default;
         base_query_state(const base_query_state&) = delete;
     };
 
-    base_query_state query_state{cmd->row_limit * queried_ranges_count, std::move(ranges_to_vnodes)};
+    base_query_state query_state{cmd->get_row_limit() * queried_ranges_count, std::move(ranges_to_vnodes)};
```
haaawk, line 528: What's the difference between row_limit and get_row_limit()?

database.cc
```diff
@@ -1335,7 +1335,7 @@ future<mutation> database::do_apply_counter_update(column_family& cf, const froz
                           regular_columns.end());
 
     auto slice = query::partition_slice(std::move(cr_ranges), std::move(static_columns),
-        std::move(regular_columns), { }, { }, cql_serialization_format::internal(), query::max_rows);
+        std::move(regular_columns), { }, { }, cql_serialization_format::internal());
```
haaawk, line 1338: We can remove max_rows because of the default value of the parameter?

cql3/statements/select_statement.cc
haaawk: Ok. I see that it just merges low and high bits.

cql3/statements/select_statement.cc
```diff
@@ -1590,7 +1590,7 @@ bool select_statement::contains_alias(const column_identifier& name) const {
 lw_shared_ptr<column_specification> select_statement::limit_receiver(bool per_partition) {
     sstring name = per_partition ? "[per_partition_limit]" : "[limit]";
     return make_lw_shared<column_specification>(keyspace(), column_family(), ::make_shared<column_identifier>(name, true),
-        int32_type);
+        long_type);
```
avikivity, line 1599: I think this breaks during mixed-version operation. The client may send a value for one of the types and the server may have a prepared statement for the other type. The sizes don't match, and validation will break.

avikivity: In any case, I don't see a query such as `SELECT * FROM tab LIMIT 5000000000` happening. People want everything or just a few.

idl/paging_state.idl.hh
```diff
@@ -11,11 +11,13 @@ namespace pager {
...
     std::optional<clustering_key> get_clustering_key();
-    uint32_t get_remaining();
+    uint32_t get_remaining_low_bits();
     utils::UUID get_query_uuid() [[version 2.2]] = utils::UUID();
     std::unordered_map<dht::token_range, std::vector<utils::UUID>> get_last_replicas() [[version 2.2]] = std::unordered_map<dht::token_range, std::vector<utils::UUID>>();
     std::optional<db::read_repair_decision> get_query_read_repair_decision() [[version 2.3]] = std::nullopt;
-    uint32_t get_rows_fetched_for_last_partition() [[version 3.1]] = 0;
+    uint32_t get_rows_fetched_for_last_partition_low_bits() [[version 3.1]] = 0;
+    std::optional<uint32_t> get_remaining_high_bits() [[version 4.2]];
+    std::optional<uint32_t> get_rows_fetched_for_last_partition_high_bits() [[version 4.2]];
```
avikivity, line 20: Good attention to detail - you can't s/32/64/ here.

But, 4.2 is already branched, so needs to be 4.3.

avikivity:
I think we can drop the "LIMIT 64-bit" thing. It's unrealistic and just adds risk because you change the type of the prepared statement. 

The pager change is the important part.

wmitros:
I've added changes from review, should I rebase now or add a fixup commit and rebase when all changes are accepted?

alternator/executor.cc
wmitros: Actually, it looks like both variables have misleading names - the limit is later used as size of a page to be fetched by the query pager

cql3/selection/selection.hh
wmitros: Will this work here? Or do you mean to change return 0 to return 0ull too?

cql3/statements/cas_request.cc
wmitros: Yes, the variable is later used in set_partition_row_limit so I wanted to use the max_rows per partition.(although both hold the same value)

avikivity:
Better to rebase now.

database.cc
wmitros: It was the reason but on second thought there is no reason to change it here

cql3/selection/selection.hh
haaawk: I guess it's ok to just leave it as is for now.

idl/result.idl.hh
```diff
@@ -30,8 +30,9 @@ class result {
     std::optional<query::result_digest> digest();
     api::timestamp_type last_modified() [ [version 1.2] ] = api::missing_timestamp;
     query::short_read is_short_read() [[version 1.6]] = query::short_read::no;
-    std::optional<uint32_t> row_count() [[version 2.1]];
+    std::optional<uint32_t> row_count_low_bits() [[version 2.1]];
     std::optional<uint32_t> partition_count() [[version 2.1]];
+    std::optional<uint32_t> row_count_high_bits() [[version 4.2]];
```
haaawk, line 35: I think @avikivity requested change to 4.3, no?

idl/reconcilable_result.idl.hh
```diff
@@ -20,12 +20,14 @@
...
     frozen_mutation mut();
+    std::optional<uint32_t> row_count_high_bits() [[version 4.2]];
 };
 
 class reconcilable_result {
-    uint32_t row_count();
+    uint32_t row_count_low_bits();
     utils::chunked_vector<partition> partitions();
     query::short_read is_short_read() [[version 1.6]] = query::short_read::no;
+    std::optional<uint32_t> row_count_high_bits() [[version 4.2]];
```
haaawk, line 32: I think @avikivity requested change to 4.3, no?

idl/read_command.idl.hh
```diff
@@ -37,19 +37,21 @@ class partition_slice {
...
     utils::UUID schema_version;
     query::partition_slice slice;
-    uint32_t row_limit;
+    uint32_t row_limit_low_bits;
     std::chrono::time_point<gc_clock, gc_clock::duration> timestamp;
     std::optional<tracing::trace_info> trace_info [[version 1.3]];
     uint32_t partition_limit [[version 1.3]] = std::numeric_limits<uint32_t>::max();
     utils::UUID query_uuid [[version 2.2]] = utils::UUID();
     query::is_first_page is_first_page [[version 2.2]] = query::is_first_page::no;
+    std::optional<uint32_t> row_limit_high_bits [[version 4.2]];
```
haaawk, line 54: I think @avikivity requested change to 4.3, no?

idl/read_command.idl.hh
```diff
@@ -37,19 +37,21 @@ class partition_slice {
     query::partition_slice::option_set options;
     std::unique_ptr<query::specific_ranges> get_specific_ranges();
     cql_serialization_format cql_format();
-    uint32_t partition_row_limit() [[version 1.3]] = std::numeric_limits<uint32_t>::max();
+    uint32_t partition_row_limit_low_bits() [[version 1.3]] = std::numeric_limits<uint32_t>::max();
+    std::optional<uint32_t> partition_row_limit_high_bits() [[version 4.2]] = std::numeric_limits<uint32_t>::max();
```
haaawk, line 41: I think @avikivity requested change to 4.3, no?

idl/reconcilable_result.idl.hh
wmitros: yes of course, for some reason I missed it

cql3/statements/select_statement.cc
wmitros: If we're abandoning manually setting the 64-bit limit, there is nothing to throw here anymore

haaawk: Correct

wmitros:
After dropping the option of directly setting LIMIT to something 64-bit the feature flag from the second commit makes no difference in practice - we needed it to change the limits over 2^32-1 to 2^32-1 if the whole cluster wasn't upgraded yet. Now if the limit is set to >2^32-1 it means that both integers representing upper and lower 32 bits are set to 2^32-1. On the other hand the feature would still be necessary if the new max was something else. Does that make us want to remove the feature flag?

haaawk:
> After dropping the option of directly setting LIMIT to something 64-bit the feature flag from the second commit makes no difference in practice - we needed it to change the limits over 2^32-1 to 2^32-1(so that the bottom bits were maximal) if the whole cluster wasn't upgraded yet. Now if the limit is set to >2^32-1 it means that both integers representing upper and lower 32 bits are set to 2^32-1. On the other hand the feature would still be necessary if the new max was something else. Does that make us want to remove the feature flag?

I have hard time understanding why the flag is no needed anymore. Aren't get_limit and get_per_partition_limit using it?

wmitros:
They are using it right now to set the default limit to max uint32 instead of setting it to max uint64 while there are old nodes present - and those nodes ignore the top 32 bits anyway. The main use of the flag was to recognize when to throw if we read a limit over max uint32 - now we can't read higher limits at all.

haaawk:
What about a case when select has no limit? Then we want to set the limit to 2^32 for cluster during upgrade and 2^64 for the fully upgraded cluster, no?

haaawk:
oh ok. I see that setting limit to 2^64 works for old cluster too.

wmitros:
That's what I meant, it's not that the flag is wrong but it has no effect for our values

haaawk:
You're probably right. It seems we could remove the flag now.

wmitros:
I've added some tests to the patch. In the "manual" test checking selecting lots of rows I've used a table with a virtual reader, which I got thanks to @piodul. He based it on the virtual reader from size_estimate table and I tried to move it outside the system keyspace to be only used in testing. Either of the two cases included in it can take up to 5h to execute, so they shouldn't be ran too frequently. I've also added some automatic tests that check if whole 64-bit values in paging_state are being used, by querying the db with a modified remaining rows value whose lower 32 bits are filled with zeroes but positive if including all 64 bits.
This patch doesn't change the limits of numbers of partitions, but it fixes some places where e.g. the max row count was used as a partition count limit. It does change the limit of rows from a single partition however.

cql3/result_generator.hh
```diff
@@ -114,7 +114,7 @@ private:
             }
         }
 
-        uint32_t rows_read() const { return _total_row_count; }
+        auto rows_read() const { return _total_row_count; }
```
avikivity, line 117: This is a bad idea, as it moves the interface of the function into its implementation. The implementation is trivial, but still it's moving the code in a JavaScript-like direction.

We should be able to terminate questions about types in the function declaration.

idl/paging_state.idl.hh
```diff
@@ -11,11 +11,13 @@ namespace pager {
...
     std::optional<clustering_key> get_clustering_key();
-    uint32_t get_remaining();
+    uint32_t get_remaining_low_bits();
     utils::UUID get_query_uuid() [[version 2.2]] = utils::UUID();
     std::unordered_map<dht::token_range, std::vector<utils::UUID>> get_last_replicas() [[version 2.2]] = std::unordered_map<dht::token_range, std::vector<utils::UUID>>();
     std::optional<db::read_repair_decision> get_query_read_repair_decision() [[version 2.3]] = std::nullopt;
-    uint32_t get_rows_fetched_for_last_partition() [[version 3.1]] = 0;
+    uint32_t get_rows_fetched_for_last_partition_low_bits() [[version 3.1]] = 0;
+    std::optional<uint32_t> get_remaining_high_bits() [[version 4.3]];
+    std::optional<uint32_t> get_rows_fetched_for_last_partition_high_bits() [[version 4.3]];
```
avikivity, line 20: I wonder if we need the optional here. Won't those two fields default to zero?

query-request.hh
```diff
@@ -282,16 +305,29 @@ public:
...
     { }
 
+
+    uint64_t get_row_limit() const {
+        return (static_cast<uint64_t>(row_limit_high_bits.value_or(0)) << 32) | row_limit_low_bits;
+    }
+    void set_row_limit(uint32_t new_row_limit) {
+        row_limit_low_bits = new_row_limit;
+        row_limit_high_bits = 0;
+    }
```
avikivity, line 326: Is this overload needed? the other overload will work just fine.

idl/paging_state.idl.hh
wmitros: After reading the idl compiler it looks like you're right. For some reason I thought that the optionals were needed to generate the default initialization if the field can't be deserialized, while the deciding factor is actually a presence of [[version]] attribute.

avikivity: I am sorry you had to read the idl compiler source.

query-request.hh
wmitros: I think I wanted to avoid excessive casts, removed in the next push

cql3/result_generator.hh
wmitros: Ok, fixed.

test/boost/large_paging_state_test.cc
```diff
@@ -0,0 +1,125 @@
...
+
+        auto qo = std::make_unique<cql3::query_options>(db::consistency_level::LOCAL_ONE, infinite_timeout_config, std::vector<cql3::raw_value>{},
+                    cql3::query_options::specific_options{5, nullptr, {}, api::new_timestamp()});
+        auto msg = e.execute_cql("select * from test;", std::move(qo)).get0();
+        auto paging_state = extract_paging_state(msg);
+        uint64_t rows_fetched = count_rows_fetched(msg);
+        BOOST_REQUIRE_EQUAL(paging_state->get_remaining() + rows_fetched, query::max_rows);
+
+        uint64_t test_remaining = paging_state->get_remaining() & ((uint64_t(1) << 32) - 1) << 32;
+        assert(static_cast<uint32_t>(test_remaining) == 0);
```
nyh, line 69: I don't understand the point of this assert, you just created test_remaining one line up, so obviously it will succeed if static_cast<> just does &MAX_UINT32. Why do you want to check that?

test/boost/large_paging_state_test.cc
```diff
@@ -0,0 +1,125 @@
...
+        BOOST_REQUIRE(test_remaining > std::numeric_limits<uint32_t>::max());
+        paging_state->set_remaining(test_remaining);
```
nyh, line 71: I don't understand what this set_remaining() is supposed to do... Instead of remaining() being 2^64, you changed it to 2^64 - 2^32, what does this achieve or test? A comment here, or on the top of the test, saying what you are trying to test, would have been nice.

test/lib/enormous_table.hh
```diff
@@ -0,0 +1,67 @@
+#pragma once
+
+#include "mutation_reader.hh"
+
+namespace enormous_table {
+
+extern logging::logger elogger;
+
+class enormous_table_reader final : public flat_mutation_reader::impl {
```
nyh, line 9: You put this code in lib/, do you think there's a chance it will ever be reused in other tests? If not, probably we don't need it in lib. But if you do, please add a comment explaining what this class does - that it reads into a virtual table with 4.5 billion rows.

test/boost/large_paging_state_test.cc
wmitros: It was more of a check for a bug in the calculations above, I can delete it considering it passed at least once

nyh: I guess there is no harm done, but it did take me a minute to understand that this is an "assert", not a "BOOST_REQUIRE", so maybe it's better to remove it :-)

test/boost/large_paging_state_test.cc
wmitros: While executing a paged select statement, if the bottom bits of the remaining rows number hold a high enough value the returned rows number will be subtracted from just them and the top bits will remain unchanged, so we can't tell if they're even being read. In this test, the bottom bits are set to 0 so that the top bits must be used when deciding how many rows to return. We check if the rows are returned and if the top bits are properly modified in the response. 
We're modifying the paging state returned with the response to "select * from test" instead of simply adding a limit to the query because we've decided that the LIMIT in queries should be restricted to 32 bits (changing that would cause problems with queries prepared before the upgrade, and we don't expect queries like "SELECT * FROM tab LIMIT 5000000000" anyway, rather simply "SELECT * FROM tab"). We're using the 64-bit value only if we're selecting all rows from a table.
I'll add a similar comment to the test.

test/lib/enormous_table.hh
wmitros: We probably won't use it any more. I'll include it within the test and add a comment

nyh:
Committed to next: 936cf4cce0b7e50706c73a48ee7fb01758ff9ac6

PR #6914, opened 2020-07-24
cdc: New delta modes: `off`, `keys`, `fulll`

### LINKED ISSUE #6838: CDC: need option to turn off delta values
It running CDC with either pre- or postimage, there might be situations where we simply don't care about the delta info.
One very poignant example is alternator streams, where one pre and/or post data is needed.
Conversely, it might be useful to have an option "keys only" that records delta, but only PK/CK data.

Suggest adding additional parameters to the CDC ext settings struct, that (default full delta) has "mode" flag for delta (off, keys, full).

jul-stas:
The goal is to have finer control over CDC "delta" rows, i.e.:
- disable them totally (mode `off`);
- record only base PK+CK columns (mode `keys`);
- make them behave as usual (mode `full`, default).

The editing of log rows is performed at the stage of `finish`ing CDC mutation.

Fixes #6838

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -292,6 +292,21 @@ future<> cdc::cdc_service::stop() {
 
 cdc::cdc_service::~cdc_service() = default;
 
+namespace {
+const sstring delta_mode_string_off  = "off";
+const sstring delta_mode_string_keys = "keys";
+const sstring delta_mode_string_full = "full";
+
+sstring to_string(cdc::delta_mode dm) {
+    switch (dm) {
+        case cdc::delta_mode::off  : return delta_mode_string_off;
+        case cdc::delta_mode::keys : return delta_mode_string_keys;
+        case cdc::delta_mode::full : return delta_mode_string_full;
+    }
+    throw std::logic_error("Impossible value of cdc::delta_mode");
+}
+} // anon. namespace
+
 cdc::options::options(const std::map<sstring, sstring>& map) {
     if (map.find("enabled") == std::end(map)) {
         return;
@@ -304,6 +319,14 @@ cdc::options::options(const std::map<sstring, sstring>& map) {
             _preimage = p.second == "true";
         } else if (p.first == "postimage") {
             _postimage = p.second == "true";
+        } else if (p.first == "delta") {
+            if (p.second == delta_mode_string_keys) {
+                _delta_mode = delta_mode::keys;
+            } else if (p.second == delta_mode_string_off) {
+                _delta_mode = delta_mode::off;
+            } else if (p.second != delta_mode_string_full) {
+                throw exceptions::configuration_exception("Invalid value for CDC option \"delta\": " + p.second);
+            }
         } else if (p.first == "ttl") {
             _ttl = std::stoi(p.second);
             if (_ttl < 0) {
@@ -312,6 +335,11 @@ cdc::options::options(const std::map<sstring, sstring>& map) {
         } else {
             throw exceptions::configuration_exception("Invalid CDC option: " + p.first);
         }
+    }
+
+    if (_enabled && !_preimage && !_postimage && _delta_mode == delta_mode::off) {
+        throw exceptions::configuration_exception("Invalid combination of CDC options: neither of"
+                " {preimage, postimage, delta} is enabled");
     }
 }
 
@@ -323,6 +351,7 @@ std::map<sstring, sstring> cdc::options::to_map() const {
         { "enabled", _enabled ? "true" : "false" },
         { "preimage", _preimage ? "true" : "false" },
         { "postimage", _postimage ? "true" : "false" },
+        { "delta", to_string(_delta_mode) },
         { "ttl", std::to_string(_ttl) },
     };
 }
@@ -332,7 +361,8 @@ sstring cdc::options::to_sstring() const {
 }
 
 bool cdc::options::operator==(const options& o) const {
-    return _enabled == o._enabled && _preimage == o._preimage && _postimage == o._postimage && _ttl == o._ttl;
+    return _enabled == o._enabled && _preimage == o._preimage && _postimage == o._postimage && _ttl == o._ttl
+            && _delta_mode == o._delta_mode;
 }
 bool cdc::options::operator!=(const options& o) const {
     return !(*this == o);
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -841,6 +841,56 @@ private:
         m.set_cell(ck, _ttl_col, atomic_cell::make_live(*_ttl_col.type, _ts, _ttl_col.type->decompose(ttl.count()), _cdc_ttl_opt));
     }
 
+    // Remove non-key columns or entire delta rows, according to `delta` setting in `cdc` options.
+    void adjust_or_delete_deltas() {
+        if (_schema->cdc_options().get_delta_mode() == cdc::delta_mode::full) {
+            return;
+        }
+
+        static const auto preimg_op_bytes  = _op_col.type->decompose(operation_native_type(operation::pre_image));
+        static const auto postimg_op_bytes = _op_col.type->decompose(operation_native_type(operation::post_image));
+        for (auto& m : _result_mutations) {
+            auto& clustered_rows = m.partition().clustered_rows();
+            int deleted_rows_cnt = 0;
+            for (auto it = clustered_rows.begin(); it != clustered_rows.end(); /* no increment */) {
+                const auto& op_cell = it->row().cells().cell_at(_op_col.id).as_atomic_cell(_op_col);
+                const auto op_val = op_cell.value().linearize();
+                if (op_val != preimg_op_bytes && op_val != postimg_op_bytes) {
+                    if (_schema->cdc_options().get_delta_mode() == cdc::delta_mode::off) {
+                        it = m.partition().clustered_rows().erase_and_dispose(it, current_deleter<rows_entry>());
+                        ++deleted_rows_cnt;
+                        continue;
+                    }
+
+                    // The case of `get_delta_mode() == delta_mode::keys`:
+                    it->row().cells().remove_if([this, log_s = m.schema()] (column_id id, atomic_cell_or_collection& acoc) {
+                        const auto& log_cdef = log_s->column_at(column_kind::regular_column, id);
+                        // We can surely remove "cdc$*" columns.
+                        if (is_cdc_metacolumn_name(log_cdef.name_as_text())) {
+                            return true;
+                        }
+                        const auto* base_cdef = _schema->get_column_definition(log_cdef.name());
+                        // Remove columns from delta that correspond to non-PK/CK columns in the base table.
+                        return base_cdef != nullptr
+                                && (base_cdef->kind != column_kind::partition_key && base_cdef->kind != column_kind::clustering_key);
+                    });
+                }
+
+                // Deletion of deltas might leave gaps in `batch_seq_no` - let's fix them.
+                if (deleted_rows_cnt > 0) {
+                    const auto* batch_seq_no_cdef = m.schema()->get_column_definition(log_meta_column_name_bytes("batch_seq_no"));
+                    assert(batch_seq_no_cdef != nullptr);
+                    auto exploded_ck = it->key().explode();
+                    const size_t batch_seq_no_idx = batch_seq_no_cdef->component_index();
+                    const auto old_batch_seq_no = value_cast<int32_t>(int32_type->deserialize(exploded_ck[batch_seq_no_idx]));
+                    exploded_ck[batch_seq_no_idx] = int32_type->decompose(old_batch_seq_no - deleted_rows_cnt);
+                    it->key() = clustering_key::from_exploded(std::move(exploded_ck));
+                }
+                ++it;
+            }
+        }
+    }
+
 public:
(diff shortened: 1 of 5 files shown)


cdc/log.cc
```diff
@@ -292,6 +292,21 @@ future<> cdc::cdc_service::stop() {
...
+const sstring delta_mode_string_off  = "off";
+const sstring delta_mode_string_keys = "keys";
+const sstring delta_mode_string_full = "full";
+
+sstring to_string(cdc::delta_mode dm) {
+    switch (dm) {
+        case cdc::delta_mode::off  : return delta_mode_string_off;
+        case cdc::delta_mode::keys : return delta_mode_string_keys;
+        case cdc::delta_mode::full : return delta_mode_string_full;
+        default : throw std::runtime_error("Impossible value of cdc::delta_mode");
```
haaawk, line 305: No need for default. Compiler should complain about switch not covering all possible enum values.

cdc/log.cc
```diff
@@ -1258,6 +1258,40 @@ class transformer final : public change_processor {
...
+                for (auto it = clustered_rows.begin(); it != clustered_rows.end(); /* no increment */) {
+                    const auto& op_cell = it->row().cells().cell_at(_op_col.id).as_atomic_cell(_op_col);
+                    const auto op_val = op_cell.value().linearize();
+                    if (op_val != preimg_op_bytes && op_val != postimg_op_bytes) {
+                        if (_schema->cdc_options().get_delta_mode() == cdc::delta_mode::off) {
+                            it = m.partition().clustered_rows().erase_and_dispose(it, current_deleter<rows_entry>());
+                            continue;
+                        }
+
+                        if (_schema->cdc_options().get_delta_mode() == cdc::delta_mode::keys) {
```
haaawk, line 1277: Do we have to check this condition here? Wouldn't an else be enough?

haaawk:
> I'm skeptical about the approach where extra work is done to produce less data;

I like the fact that it's so well separated but you've got the point that it could have been more efficient. What would be the alternative? Placing ifs in existing code. I think the best way to proceed is to do this like in your RFC for now and later when @kbr- finishes his refactoring, we will see if we can painlessly integrate the no delta option into the main code.

> In off mode batch_seq_no is monotonic, but non-continuous;

That's something we should fix. Let's just remove every following row and reinsert it with batch_seq_no - 1  (or -X).

> I'm unsure about the behavior of keys mode when things are deleted;

You don't remove cdc$<name>_deleted so I think it will be possible to figure out what was deleted. I think it's ok as it is now.

> I could not determine whether current_deleter is correct;

I think it's alright.

haaawk:
> > I'm skeptical about the approach where extra work is done to produce less data;
> 
> I like the fact that it's so well separated but you've got the point that it could have been more efficient. What would be the alternative? Placing ifs in existing code. I think the best way to proceed is to do this like in your RFC for now and later when @kbr- finishes his refactoring, we will see if we can painlessly integrate the no delta option into the main code.
> 
> > In off mode batch_seq_no is monotonic, but non-continuous;
> 
> That's something we should fix. Let's just remove every following row and reinsert it with batch_seq_no - 1 (or -X).

Actually we don't have to reinsert the rows. We can just replace the batch_seq_no. This won't change the order of elements so it should be ok since they are stored in a tree.

> 
> > I'm unsure about the behavior of keys mode when things are deleted;
> 
> You don't remove cdc$_deleted so I think it will be possible to figure out what was deleted. I think it's ok as it is now.
> 
> > I could not determine whether current_deleter is correct;
> 
> I think it's alright.

cdc/log.cc
jul-stas: Maybe my compiler is stupid. Without `default`:
```
distcc[638630] ERROR: compile cdc/log.cc on localhost failed
cdc/log.cc: In function ‘seastar::sstring {anonymous}::to_string(cdc::delta_mode)’:
cdc/log.cc:307:1: error: control reaches end of non-void function [-Werror=return-type]
```
I can move `throw` outside of `switch` - just checked that my gcc allows this.

cdc/log.cc
jul-stas: We don't have to - I did it for future-proofness (in case we add more delta_mode levels).

haaawk: I don't think this is much more future-proof than just else. If you were throwing when no case is  taken then it would be a different story.

cdc/log.cc
haaawk: Yes. You have to throw unconditionally at the end.

nyh:
>     * I'm skeptical about the approach where extra work is done to produce less data;

I don't know about the extra work (I still need to review the code), but this extra "work" should amount to an if(), while it results in writing less data and thus **less work** work for writing this fewer data to disk, less work when reading the extra information (if you don't want to use it anyway), and probably most importantly - less disk space usage.  

>     * I'm unsure about the behavior of `keys` mode when things are deleted;

@elcallio the way I usually work here is to write a test first: Write a test in (Alternator's pytest) which deletes a cell/row/partition, and sees what appears in the streams. Alternator will need to do exactly the same - either because CDC will also do the same, or because we will "hide" what CDC decided to do with an Alternator-specific fix. But at least such a test will let us know what DynamoDB thought makes sense.  

> Refs #6838

Wouldn't that be "Fixes" instead of Refs?

cdc/log.cc
```diff
@@ -1258,6 +1258,52 @@ class transformer final : public change_processor {
     // Takes and returns generated cdc log mutations and associated statistics about parts touched during transformer's lifetime.
     // The `transformer` object on which this method was called on should not be used anymore.
     std::tuple<std::vector<mutation>, stats::part_type_set> finish() && {
+        static const auto preimg_op_bytes  = _op_col.type->decompose(operation_native_type(operation::pre_image));
+        static const auto postimg_op_bytes = _op_col.type->decompose(operation_native_type(operation::post_image));
+
+        // Adjust or remove deltas according to `delta_mode` in `cdc_options`.
```
nyh, line 1264: I'm afraid I missed the motivation of this approach... Why add the deltas first as rows, and then remove them? Couldn't we just *not* add them?

cdc/log.cc
```diff
@@ -1258,6 +1258,52 @@ class transformer final : public change_processor {
...
+                            it = m.partition().clustered_rows().erase_and_dispose(it, current_deleter<rows_entry>());
+                            ++deleted_rows_cnt;
+                            continue;
+                        }
+
+                        // The case of `get_delta_mode() == delta_mode::keys`:
+                        it->row().cells().remove_if([this, log_s = m.schema()] (column_id id, atomic_cell_or_collection& acoc) {
+                            const auto log_cdef = log_s->column_at(column_kind::regular_column, id);
+                            if (is_cdc_metacolumn_name(log_cdef.name_as_text())) {
+                                return false; // Keep "cdc$*" columns in the delta row.
```
nyh, line 1283: Why is this? If someone asked just for the key columns, what info could the cdc$ columns contain?

cdc/log.cc
```diff
@@ -1258,6 +1258,52 @@ class transformer final : public change_processor {
...
+                            }
+                            const auto* base_cdef = _schema->get_column_definition(log_cdef.name());
+                            // Remove columns from delta that correspond to non-PK/CK columns in the base table.
+                            return base_cdef != nullptr
+                                    && (base_cdef->kind != column_kind::partition_key && base_cdef->kind != column_kind::clustering_key);
```
nyh, line 1288: Nitpick: Could we have a static_column here? If the answer is no (I'm not sure...), then it's a tiny bit faster to check if kind != regular column, than to check kind==partition_key || kind==clustering_key.

cdc/log.cc
haaawk: I asked @jul-stas  to make it this way. @kbr- is in the middle of refactoring of CDC code so it's better to keep this "no delta" option separated from the rest of the code. Later when @kbr- is finished we will integrate it back into the main code path.

cdc/log.cc
```diff
@@ -304,6 +319,14 @@ cdc::options::options(const std::map<sstring, sstring>& map) {
             _preimage = p.second == "true";
         } else if (p.first == "postimage") {
             _postimage = p.second == "true";
+        } else if (p.first == "delta_mode") {
```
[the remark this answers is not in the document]
kbr-: The other options are binary, on/off, this has three options. I also prefer "delta_mode".

cdc/log.cc
```diff
@@ -1258,6 +1258,52 @@ class transformer final : public change_processor {
     // Takes and returns generated cdc log mutations and associated statistics about parts touched during transformer's lifetime.
     // The `transformer` object on which this method was called on should not be used anymore.
     std::tuple<std::vector<mutation>, stats::part_type_set> finish() && {
+        static const auto preimg_op_bytes  = _op_col.type->decompose(operation_native_type(operation::pre_image));
+        static const auto postimg_op_bytes = _op_col.type->decompose(operation_native_type(operation::post_image));
+
+        // Adjust or remove deltas according to `delta_mode` in `cdc_options`.
+        if (_schema->cdc_options().get_delta_mode() != cdc::delta_mode::full) {
```
kbr-, line 1265: I think it would be better if you place the adjust_or_remove in a separate function.
This way you could get rid of some of this right-drift using early returns instead of blocks under `if`s.

kbr-:
@haaawk 
>> In off mode batch_seq_no is monotonic, but non-continuous;

> That's something we should fix.

I think it doesn't matter.

> Let's just remove every following row and reinsert it with batch_seq_no - 1 (or -X).

I think we should leave it as it is, since it's a non-problem.

cdc/log.cc
kbr-: We could

cdc/log.cc
haaawk: The refactoring is happening now. It's not in the future.

haaawk:
> As stated above, I would prefer a solution that does not create mutation entries and then remove them. If this is not feasible, perhaps a motivation why?

The motivation is to not blend this non-critical case into the already complicated and full of other cases main part of the code. We're in process of untangling the main code and separating various concepts like pre/post image, deltas, building the result, and so on. This means there are two options:

1. Implement this non-critical case in a nonintrusive way like this and later integrate it into the refactored code
2. Wait for the refactoring and only after that implement this case

I know the approach is not ideal from the performance perspective but it's very easy to maintain. Considering the fact that even this implementation will for sure have a positive impact on Alternator streams (less network and disk IO), I think it's ok for this code to be a bit less performant.

elcallio:
If you are confident that a better solution is coming soon, I guess I am fine with it as-is.

cdc/log.cc
```diff
@@ -292,6 +292,21 @@ future<> cdc::cdc_service::stop() {
...
+const sstring delta_mode_string_keys = "keys";
+const sstring delta_mode_string_full = "full";
+
+sstring to_string(cdc::delta_mode dm) {
+    switch (dm) {
+        case cdc::delta_mode::off  : return delta_mode_string_off;
+        case cdc::delta_mode::keys : return delta_mode_string_keys;
+        case cdc::delta_mode::full : return delta_mode_string_full;
+    }
+    throw std::runtime_error("Impossible value of cdc::delta_mode");
```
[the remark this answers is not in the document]
jul-stas: done

cdc/log.cc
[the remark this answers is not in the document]
jul-stas: Nadav and Calle need this feature, so let's let them decide :) 
Done.

cdc/log.cc
jul-stas: Done.

cdc/log.cc
jul-stas: done

elcallio:
code looks ok to me. But there should maybe be a unit test to verify we get the expected results?

jul-stas:
A simple test is added. Kind reviewers, please have a look.

elcallio:
lgtm

slivne:
@nyh ping - (calle and piotrj approved)

test/cql/cdc_delta_modes_test.cql
```diff
@@ -0,0 +1,16 @@
+create table tb1 (pk int, ck int, i1 int, PRIMARY KEY (pk, ck)) with cdc = {'enabled': true, 'preimage': false, 'postimage': false, 'delta': 'off'};
+
+create table tb2 (pk int, ck int, i1 int, PRIMARY KEY (pk, ck)) with cdc = {'enabled': true, 'preimage': true, 'postimage': true, 'delta': 'off'};
+-- Should add 2 rows (preimage + postimage)
+insert into tb2 (pk, ck, i1) VALUES (1, 11, 111) USING TTL 1111;
```
nyh, line 5: nitpick (I'll commit this anyway, just asking): Why this "USING TTL 1111"? How is it relevant for this test?

test/cql/cdc_delta_modes_test.cql
```diff
@@ -0,0 +1,16 @@
...
+select "cdc$batch_seq_no", "cdc$operation", pk, ck, i1 from tb2_scylla_cdc_log where pk = 1 and ck = 11 allow filtering;
```
nyh, line 6: Just curious, why not "SELECT *"? Is there a specific column which this list of columns try to avoid?

Issue #5634, opened 2020-01-23
Allow changing the hinted handoff configuration in the run time


vladzcloudius:
HEAD: 6909d4db42d3e7aa78035c44c038f102108e1ca3

## Description

Allow changing the node's hinted handoff configuration in the run time, e.g. via REST API.

Today changing HH state requires a node restart which is a serious pain for big clusters.

There isn't any technical difficulty to change HH state in the run time: we simply change the HH configuration and let all outstanding hints to be stored.

Hints sending doesn't need to be touched at all - after all outstanding hints are sent out it will dry out by itself...

slivne:
we support changing config at runtime - we have some samples in the code lets align (I think there are also nodetool specific commands for this

PR #6916, opened 2020-07-24
Allow changing hinted handoff configuration in runtime

piodul:
This PR allows changing the hinted_handoff_enabled option in runtime, either by modifying and reloading YAML configuration, or through HTTP API.

This PR also introduces an important change in semantics of hinted_handoff_enabled:
- Previously, hinted_handoff_enabled controlled whether _both writing and sending_ hints is allowed at all, or to particular DCs,
- Now, hinted_handoff_enabled only controls whether _writing hints_ is enabled. Sending hints from disk is now always enabled.

Fixes: #5634
Tests:
- unit(dev) for each commit of the PR
- unit(debug) for the last commit of the PR


piodul:
Associated dtest: https://github.com/scylladb/scylla-dtest/pull/1615

vladzcloudius:
@piodul I'll take a look at this on Monday - too tired today :(

db/hints/host_filter.cc
```diff
@@ -0,0 +1,119 @@
...
+host_filter host_filter::parse_from_config_string(sstring opt) {
+    if (boost::iequals(opt, "false") || opt == "0") {
+        return host_filter(disabled_for_all_tag());
+    } else if (boost::iequals(opt, "true") || opt == "1") {
+        return host_filter(enabled_for_all_tag());
+    }
+
+    return parse_from_dc_list(std::move(opt));
+}
+
```
avikivity, line 65: From cassandra.yaml:

```yaml
# May either be "true" or "false" to enable globally
hinted_handoff_enabled: true

# When hinted_handoff_enabled is true, a black list of data centers that will not
# perform hinted handoff
# hinted_handoff_disabled_datacenters:
#    - DC1
#    - DC2
```

So let's follow that.

As a general remark, giving multiple parse options to a single input is usually bad, because it creates ambiguity (I like naming my datacenters `true` and `false`, because I'm that kind of person).

db/hints/manager.cc
```diff
@@ -527,9 +527,7 @@ bool manager::can_hint_for(ep_key_type ep) const noexcept {
 bool manager::check_dc_for(ep_key_type ep) const noexcept {
     try {
         // If target's DC is not a "hintable" DCs - don't hint.
-        // If there is an end point manager then DC has already been checked and found to be ok.
-        return _hinted_dcs.empty() || have_ep_manager(ep) ||
-               _hinted_dcs.find(_local_snitch_ptr->get_datacenter(ep)) != _hinted_dcs.end();
+        return _host_filter.can_hint_for(_local_snitch_ptr, ep);
```
avikivity, line 536: We seriously need a node_ptr and datacenter_ptr thing so we don't have to keep looking up hash tables (general remark, not a requirement here).

db/hints/host_filter.cc
avikivity: Edit: I see it already exists. Why did we do that?

service/storage_proxy.cc
```diff
@@ -1765,6 +1765,8 @@ inline std::ostream& operator<<(std::ostream& os, const hint_wrapper& h) {
 
 using namespace std::literals::chrono_literals;
 
+static thread_local const utils::updateable_value_source<db::hints::host_filter> allow_all_hint_filter{db::hints::host_filter(db::hints::host_filter::enabled_for_all_tag())};
```
avikivity, line 1768: Can't we store it outside a global? in storage_proxy maybe?

api/storage_proxy.cc
```diff
@@ -201,29 +202,38 @@ void set_storage_proxy(http_context& ctx, routes& r) {
...
 
     sp::set_hinted_handoff_enabled.set(r, [](std::unique_ptr<request> req)  {
-        //TBD
-        unimplemented();
         auto enable = req->get_query_param("enable");
-        return make_ready_future<json::json_return_type>(json_void());
+        auto filter = (enable == "true")
+                ? db::hints::host_filter(db::hints::host_filter::enabled_for_all_tag {})
+                : db::hints::host_filter(db::hints::host_filter::disabled_for_all_tag {});
+        return service::get_storage_proxy().invoke_on_all([filter] (service::storage_proxy& sp) {
```
avikivity, line 214: Let's here at least separate the enable/disable flag from the list.

Perhaps we should also implement the configuration with a separate item (but keep compatibility with ourselves).

service/storage_proxy.hh
```diff
@@ -584,6 +584,9 @@ public:
     void allow_replaying_hints() noexcept;
     future<> drain_on_shutdown();
 
+    void set_hints_host_filter(const db::hints::host_filter& new_filter);
+    db::hints::host_filter get_hints_host_filter() const;
+
     const stats& get_stats() const {
```
avikivity, line 590: There is a problem with two configuration sources. Suppose you set via REST and then SIGHUP for some unrelated change. Which should take precedence? I think REST will now not be overridden by SIGHUP, but it's really an implementation accident.

We should route REST via config, not have two separate paths talking to the controlled object.

I think this is a preexisting problem.

avikivity:
Looks good overall.

db/hints/host_filter.cc
vladzcloudius: @avikivity For compatibility reasons. I assume your initial post had the same motivation exactly.
I didn't like it either but C* did that so we had to follow.

db/hints/host_filter.cc
```diff
@@ -0,0 +1,119 @@
...
+        : _enabled_kind(allowed_dcs.empty() ? enabled_kind::disabled_for_all : enabled_kind::enabled_selectively)
+        , _dcs(std::move(allowed_dcs)) {
+}
+
+bool host_filter::can_hint_for(locator::snitch_ptr& snitch, gms::inet_address ep) const {
+    switch (_enabled_kind) {
+    case enabled_kind::enabled_for_all:
+        return true;
+    case enabled_kind::enabled_selectively:
+        return _dcs.count(snitch->get_datacenter(ep));
```
vladzcloudius, line 48: Please, cache a reference to the snitch_ptr in the `host_filter`. 
The best would be to cache a reference to the object itself but it would be a lot riskier and open a possibility for nasty bugs.

db/hints/host_filter.cc
```diff
@@ -0,0 +1,119 @@
...
+    f = host_filter::parse_from_config_string(std::move(tmp));
+    return is;
+}
+
+sstring host_filter::to_configuration_string() const {
+    switch (_enabled_kind) {
+    case enabled_kind::enabled_for_all: return "true";
+    case enabled_kind::enabled_selectively: return ::join(",", _dcs);
+    case enabled_kind::disabled_for_all: return "false";
+    default: abort();
```
vladzcloudius, line 94: Please, don't use `abort()` - throw a corresponding exception instead.

db/hints/host_filter.cc
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+
+std::string_view host_filter::enabled_kind_to_string(host_filter::enabled_kind ek) {
+    switch (ek) {
+    case host_filter::enabled_kind::enabled_for_all: return "enabled_for_all";
+    case host_filter::enabled_kind::enabled_selectively: return "enabled_selectively";
+    case host_filter::enabled_kind::disabled_for_all: return "disabled_for_all";
+    default: abort();
```
vladzcloudius, line 104: The same here

db/hints/host_filter.cc
```diff
@@ -0,0 +1,119 @@
...
+std::istream& operator>>(std::istream& is, host_filter& f) {
+    sstring tmp;
+    is >> tmp;
+    f = host_filter::parse_from_config_string(std::move(tmp));
+    return is;
+}
+
+sstring host_filter::to_configuration_string() const {
+    switch (_enabled_kind) {
+    case enabled_kind::enabled_for_all: return "true";
```
vladzcloudius, line 91: I personally don't like this style.
I think `return` deserves to be in a separate line.

I see that we have both kinds in the code base however there are many more of the kind that I like:

```
$ for f in `find . -name "*.cc"`; do grep -w case $f | grep -w -v return; done | wc -l
2566
$ for f in `find . -name "*.cc"`; do grep -w case $f | grep -w return; done | wc -l
422
$ for f in `find . -name "*.hh"`; do grep -w case $f | grep -w -v return; done | wc -l
817
$ for f in `find . -name "*.hh"`; do grep -w case $f | grep -w return; done | wc -l
113
```

db/hints/host_filter.hh
```diff
@@ -0,0 +1,103 @@
...
+    enabled_kind _enabled_kind;
+    std::unordered_set<sstring> _dcs;
+
+    static std::string_view enabled_kind_to_string(host_filter::enabled_kind ek);
+
+public:
+    struct enabled_for_all_tag {};
+    struct disabled_for_all_tag {};
+
+    // Creates a filter that does not allow any hints (default)
```
vladzcloudius, line 56: I think it makes more sense to have the same default as in the application, which is `enabled to all hosts`.

db/hints/host_filter.hh
```diff
@@ -0,0 +1,103 @@
...
+    // Creates a filter that allows sending hints to specified DCs.
+    explicit host_filter(std::unordered_set<sstring> allowed_dcs);
+
+    // Parses hint filtering configuration from the hinted_handoff_enabled option.
+    static host_filter parse_from_config_string(sstring opt);
+
+    // Parses hint filtering configuration from a list of DCs.
+    static host_filter parse_from_dc_list(sstring opt);
+
+    bool can_hint_for(locator::snitch_ptr& snitch, gms::inet_address ep) const;
```
vladzcloudius, line 71: Please, simplify the interface - get rid of `locator::snitch_ptr& snitch`

db/hints/host_filter.cc
```diff
@@ -0,0 +1,119 @@
...
+    case host_filter::enabled_kind::disabled_for_all: return "disabled_for_all";
+    default: abort();
+    }
+}
+
+std::ostream& operator<<(std::ostream& os, const host_filter& f) {
+    os << "host_filter{enabled_kind="
+        << host_filter::enabled_kind_to_string(f._enabled_kind)
+        << ", dcs={"
+        << ::join(",", f._dcs)
```
vladzcloudius, line 112: This would create a weird output for `enabled/disabled  for all`: "enabled_kind=enabled/disabled_for_all, dcs={}".
Printing an empty dcs list is rather confusing.

db/hints/host_filter.hh
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    sstring to_configuration_string() const;
+
+    friend std::ostream& operator<<(std::ostream& os, const host_filter& f);
+};
+
+std::istream& operator>>(std::istream& is, host_filter& f);
+
+class hints_configuration_parse_error : public std::runtime_error {
```
vladzcloudius, line 97: You seem to even create a special error class for invalid format - you should use it instead of `abort()`.

db/config.cc
```diff
@@ -567,7 +588,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
         "Time interval in milliseconds to reset all node scores, which allows a bad node to recover.")
     , dynamic_snitch_update_interval_in_ms(this, "dynamic_snitch_update_interval_in_ms", value_status::Unused, 100,
         "The time interval for how often the snitch calculates node scores. Because score calculation is CPU intensive, be careful when reducing this interval.")
-    , hinted_handoff_enabled(this, "hinted_handoff_enabled", value_status::Used, "true",
+    , hinted_handoff_enabled(this, "hinted_handoff_enabled", value_status::Used, db::hints::host_filter(db::hints::host_filter::enabled_for_all_tag()),
```
vladzcloudius: Use `db::config::hinted_handoff_enabled_type` instead of `db::hints::host_filter`

db/hints/manager.cc
vladzcloudius: @avikivity what is `node_ptr`?
You can't cache datacenter_ptr here because ep is a parameter here.
The whole idea behind these lines is, first, to look up the corresponding DC in the snitch, and then lookup the corresponding DC in the hints configuration (hints_filter).

vladzcloudius: @piodul Dropping a `have_ep_manager(ep)` check is not a good idea.
This is meant to eliminate the need for those hash tables lookups every time except for the first time we see this `ep` IF this `ep` can be hinted.

One thing that can be optimized here however is that you can also cache negative results in a local cache (`unordered_set<inet_address>) of `inet_address` (which is effectively an int or a bunch of ints for IPv6) thereby optimizing the two string cache lookup into a single `int` cache lookup which should be a lot more efficient.

service/storage_proxy.cc
```diff
@@ -1788,11 +1788,11 @@ storage_proxy::storage_proxy(distributed<database>& db, storage_proxy::config cf
                        sm::description("number of currently throttled write requests")),
     });
 
-    if (cfg.hinted_handoff_enabled) {
+    if (!cfg.hinted_handoff_enabled.is_disabled_for_all()) {
         const db::config& dbcfg = _db.local().get_config();
         supervisor::notify("creating hints manager");
-        slogger.trace("hinted DCs: {}", *cfg.hinted_handoff_enabled);
-        _hints_manager.emplace(dbcfg.hints_directory(), *cfg.hinted_handoff_enabled, dbcfg.max_hint_window_in_ms(), _hints_resource_manager, _db);
+        slogger.trace("hinted DCs: {}", cfg.hinted_handoff_enabled);
```
vladzcloudius: Please, adjust this log message.
The resulting message is going to be SUPER confusing, e.g. for all hints enabled it's going to be: "hinted DCs: enabled_kind=enabled, dcs={}".

service/storage_proxy.hh
```diff
@@ -158,7 +159,7 @@ class storage_proxy : public seastar::async_sharded_service<storage_proxy>, publ
 public:
     using clock_type = lowres_clock;
     struct config {
-        std::optional<std::vector<sstring>> hinted_handoff_enabled = {};
+        db::hints::host_filter hinted_handoff_enabled = {};
```
vladzcloudius: Don't use `db::hints::host_filter` directly - use `db::config::hinted_handoff_enabled_type` instead.

db/hints/manager.hh
```diff
@@ -468,7 +481,7 @@ private:
     seastar::named_semaphore _drain_lock = {1, named_semaphore_exception_factory{"drain lock"}};
 
 public:
-    manager(sstring hints_directory, host_filter filter, int64_t max_hint_window_ms, resource_manager&res_manager, distributed<database>& db);
+    manager(sstring hints_directory, const utils::updateable_value<host_filter>& filter, int64_t max_hint_window_ms, resource_manager&res_manager, distributed<database>& db);
```
vladzcloudius, line 484: See my previous comment: please, get rid of 'const ref'.

service/storage_proxy.cc
vladzcloudius: The name is supper confusing too. 
See my comment at the end of this patch.

service/storage_proxy.hh
```diff
@@ -159,7 +159,6 @@ class storage_proxy : public seastar::async_sharded_service<storage_proxy>, publ
 public:
     using clock_type = lowres_clock;
     struct config {
-        db::hints::host_filter hinted_handoff_enabled = {};
```
vladzcloudius: Instead of removing this you should have made it `utils::updateable_value_source`.

db/hints/manager.cc
```diff
@@ -50,9 +50,10 @@ const std::string manager::FILENAME_PREFIX("HintsLog" + commitlog::descriptor::S
 const std::chrono::seconds manager::hint_file_write_timeout = std::chrono::seconds(2);
 const std::chrono::seconds manager::hints_flush_period = std::chrono::seconds(10);
 
-manager::manager(sstring hints_directory, host_filter filter, int64_t max_hint_window_ms, resource_manager& res_manager, distributed<database>& db)
+manager::manager(sstring hints_directory, const utils::updateable_value<host_filter>& filter, int64_t max_hint_window_ms, resource_manager& res_manager, distributed<database>& db)
     : _hints_dir(fs::path(hints_directory) / format("{:d}", this_shard_id()))
-    , _host_filter(std::move(filter))
+    , _host_filter(filter())
```
vladzcloudius, line 55: Unless I miss something given the current implementation this is going to return a "disable all" filter for view hints manager instantiation (goes through the default constructor).

If I'm right then there is a question if you tested that this series works for both regular and view hints?

service/storage_proxy.cc
vladzcloudius: Note that your current series code is going to disable hints for view hints.

db/hints/manager.cc
```diff
@@ -1006,5 +1012,19 @@ void manager::update_backlog(size_t backlog, size_t max_backlog) {
     }
 }
 
+void manager::on_host_filter_update(const host_filter& filter) {
+    manager_logger.debug("updating host filtering configuration: {}", filter);
+
+    _host_filter = filter;
+
+    for (auto& [ep, mgr] : _ep_managers) {
+        if (filter.can_hint_for(_local_snitch_ptr, ep)) {
```
vladzcloudius, line 1029: This is not enough - this will leave manages to "blocked" DCs alive and they will continue sending hints. 
You should prevent them from doing that and you should also stop the corresponding sender thread from waking up every second too.

The cleanest would be to shut them down in the context of this call.

This in particular means that using the updatable_vlaue in its current state is not a good choice because it doesn't allow propagating the state of callbacks to the caller (e.g. `updateable_value_source::set(T) could have returned `future` that would resolve only when all callbacks' returned futures resolve to success and would rollback if any of them fails).

This would create a clean user interface where the caller would know that the change has been actually applied when the corresponding future resolves, for instance when triggered from the REST API. It would also allow letting the user know that the state change has failed.

No matter what however you should still keep the `hints::manager` clean.
Change in hints enable/disable should not leave senders alive and the corresponding call should denote the completion of the operation including a rollback if a state transition has failed.

If in the current state of the `liveness::LiveUpdate` framework it's impossible (as it seems to me) then we should probably not allow changing `hints_enabled` value in the runtime for now and wait till it becomes possible.

Given the state of things today (if I'm not missing something big here) the best you can do it to attempt shutting corresponding endpoint_managers in the background.
However this will be a very confusing interface because there will be no way to determine when the change has actually took place (when new hints are both stopped been generated and old hints being sent to corresponding destinations).

Having said all that I think it makes more sense to start with improving the `liveness::LiveUpdate` framework in order to address what's described above and then get back to this patch and do it properly.

main.cc
```diff
@@ -558,7 +558,6 @@ int main(int ac, char** av) {
             sstring api_address = cfg->api_address() != "" ? cfg->api_address() : rpc_address;
             sstring broadcast_address = cfg->broadcast_address();
             sstring broadcast_rpc_address = cfg->broadcast_rpc_address();
-            const auto hinted_handoff_enabled = cfg->hinted_handoff_enabled();
```
vladzcloudius: The change in the behaviour is unacceptable and it relates to my comment for the previous patch in this series.

There is not justification whatsoever in such a change and there is no technical difficulty to preserve the current behaviour.

We should be able to disable any hints activity: both generation and sending; effective immediately as soon as hints were disabled - especially via scylla.yaml.

api/storage_proxy.cc
```diff
@@ -201,29 +202,38 @@ void set_storage_proxy(http_context& ctx, routes& r) {
...
+            sp.set_hints_host_filter(filter);
```
vladzcloudius, line 215: Changing a configuration is an asynchronous operation therefore `sp.set_hints_host_filter(filter)` should return a future which should be what the corresponding manager::on_host_filter_update(...)` returns (as I described before).

The api/storage_proxy.cc code implies that - exactly as it should be.

db/hints/host_filter.cc
piodul: @avikivity @vladzcloudius Format of hinted handoff options went through a change in Cassandra. The snippet from `cassandra.yaml` has the new format, while we still implement the old format. See `scylla.yaml` in our repository:

```yaml
# See https://docs.scylladb.com/architecture/anti-entropy/hinted-handoff
# May either be "true" or "false" to enable globally, or contain a list
# of data centers to enable per-datacenter.
# hinted_handoff_enabled: DC1,DC2
# hinted_handoff_enabled: true
```

Perhaps we should consider switching to the new format, but I'd rather not do it in this patch. I sent an email with more details, so that more people can discuss it.

avikivity: ok. I agree changing the format shouldn't be done here.

piodul:
v2: addressed some of the review comments related to style.
Previous (rebased) version can be found [here](https://github.com/piodul/scylla/tree/allow-changing-hinted-handoff-configuration-in-runtime-v1-rebased)

diff v1 v2:
```diff
diff --git a/db/config.cc b/db/config.cc
index 135de941a..899a546e9 100644
--- a/db/config.cc
+++ b/db/config.cc
@@ -592,7 +592,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
         "Time interval in milliseconds to reset all node scores, which allows a bad node to recover.")
     , dynamic_snitch_update_interval_in_ms(this, "dynamic_snitch_update_interval_in_ms", value_status::Unused, 100,
         "The time interval for how often the snitch calculates node scores. Because score calculation is CPU intensive, be careful when reducing this interval.")
-    , hinted_handoff_enabled(this, "hinted_handoff_enabled", liveness::LiveUpdate, value_status::Used, db::hints::host_filter(db::hints::host_filter::enabled_for_all_tag()),
+    , hinted_handoff_enabled(this, "hinted_handoff_enabled", liveness::LiveUpdate, value_status::Used, db::config::hinted_handoff_enabled_type(db::config::hinted_handoff_enabled_type::enabled_for_all_tag()),
         "Enable or disable hinted handoff. To enable per data center, add data center list. For example: hinted_handoff_enabled: DC1,DC2. A hint indicates that the write needs to be replayed to an unavailable node. "
         "Related information: About hinted handoff writes")
     , hinted_handoff_throttle_in_kb(this, "hinted_handoff_throttle_in_kb", value_status::Unused, 1024,
diff --git a/db/hints/host_filter.cc b/db/hints/host_filter.cc
index 6cc9f6d56..595958b41 100644
--- a/db/hints/host_filter.cc
+++ b/db/hints/host_filter.cc
@@ -45,11 +45,11 @@ bool host_filter::can_hint_for(locator::snitch_ptr& snitch, gms::inet_address ep
     case enabled_kind::enabled_for_all:
         return true;
     case enabled_kind::enabled_selectively:
-        return _dcs.count(snitch->get_datacenter(ep));
+        return _dcs.contains(snitch->get_datacenter(ep));
     case enabled_kind::disabled_for_all:
         return false;
     default:
-        abort();
+        throw std::logic_error("Uncovered variant of enabled_kind");
     }
 }
 
@@ -88,29 +88,38 @@ std::istream& operator>>(std::istream& is, host_filter& f) {
 
 sstring host_filter::to_configuration_string() const {
     switch (_enabled_kind) {
-    case enabled_kind::enabled_for_all: return "true";
-    case enabled_kind::enabled_selectively: return ::join(",", _dcs);
-    case enabled_kind::disabled_for_all: return "false";
-    default: abort();
+    case enabled_kind::enabled_for_all:
+        return "true";
+    case enabled_kind::enabled_selectively:
+        return ::join(",", _dcs);
+    case enabled_kind::disabled_for_all:
+        return "false";
+    default:
+        throw std::logic_error("Uncovered variant of enabled_kind");
     }
 }
 
 
 std::string_view host_filter::enabled_kind_to_string(host_filter::enabled_kind ek) {
     switch (ek) {
-    case host_filter::enabled_kind::enabled_for_all: return "enabled_for_all";
-    case host_filter::enabled_kind::enabled_selectively: return "enabled_selectively";
-    case host_filter::enabled_kind::disabled_for_all: return "disabled_for_all";
-    default: abort();
+    case host_filter::enabled_kind::enabled_for_all:
+        return "enabled_for_all";
+    case host_filter::enabled_kind::enabled_selectively:
+        return "enabled_selectively";
+    case host_filter::enabled_kind::disabled_for_all:
+        return "disabled_for_all";
+    default:
+        throw std::logic_error("Uncovered variant of enabled_kind");
     }
 }
 
 std::ostream& operator<<(std::ostream& os, const host_filter& f) {
     os << "host_filter{enabled_kind="
-        << host_filter::enabled_kind_to_string(f._enabled_kind)
-        << ", dcs={"
-        << ::join(",", f._dcs)
-        << "}";
+        << host_filter::enabled_kind_to_string(f._enabled_kind);
+    if (f._enabled_kind == host_filter::enabled_kind::enabled_selectively) {
+        os << ", dcs={" << ::join(",", f._dcs);
+    }
+    os << "}";
     return os;
 }
 
diff --git a/db/hints/host_filter.hh b/db/hints/host_filter.hh
index 1c4b6c9bd..2618d1d5a 100644
--- a/db/hints/host_filter.hh
+++ b/db/hints/host_filter.hh
@@ -53,11 +53,11 @@ class host_filter final {
     struct enabled_for_all_tag {};
     struct disabled_for_all_tag {};
 
-    // Creates a filter that does not allow any hints (default)
-    host_filter(disabled_for_all_tag tag = {});
+    // Creates a filter that allows hints to all endpoints (default)
+    host_filter(enabled_for_all_tag tag = {});
 
-    // Creates a filter that allows hints to all endpoints.
-    host_filter(enabled_for_all_tag);
+    // Creates a filter that does not allow any hints.
+    host_filter(disabled_for_all_tag);
 
     // Creates a filter that allows sending hints to specified DCs.
     explicit host_filter(std::unordered_set<sstring> allowed_dcs);
diff --git a/db/hints/manager.cc b/db/hints/manager.cc
index 63e42a9ef..4fd617346 100644
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -50,8 +50,12 @@ const std::string manager::FILENAME_PREFIX("HintsLog" + commitlog::descriptor::S
 const std::chrono::seconds manager::hint_file_write_timeout = std::chrono::seconds(2);
 const std::chrono::seconds manager::hints_flush_period = std::chrono::seconds(10);
 
+static fs::path get_hint_directory_for_current_shard(sstring hints_directory) {
+    return fs::path(hints_directory) / format("{:d}", this_shard_id());
+}
+
 manager::manager(sstring hints_directory, const utils::updateable_value<host_filter>& filter, int64_t max_hint_window_ms, resource_manager& res_manager, distributed<database>& db)
-    : _hints_dir(fs::path(hints_directory) / format("{:d}", this_shard_id()))
+    : _hints_dir(get_hint_directory_for_current_shard(std::move(hints_directory)))
     , _host_filter(filter())
     , _host_filter_change_observer(filter.observe(std::bind(&manager::on_host_filter_update, this, std::placeholders::_1)))
     , _local_snitch_ptr(locator::i_endpoint_snitch::get_local_snitch_ptr())
@@ -60,6 +64,16 @@ manager::manager(sstring hints_directory, const utils::updateable_value<host_fil
     , _resource_manager(res_manager)
 {}
 
+manager::manager(sstring hints_directory, host_filter filter, int64_t max_hint_window_ms, resource_manager& res_manager, distributed<database>& db)
+    : _hints_dir(get_hint_directory_for_current_shard(std::move(hints_directory)))
+    , _host_filter(filter)
+    , _host_filter_change_observer(nullptr, {})
+    , _local_snitch_ptr(locator::i_endpoint_snitch::get_local_snitch_ptr())
+    , _max_hint_window_us(max_hint_window_ms * 1000)
+    , _local_db(db.local())
+    , _resource_manager(res_manager)
+{}
+
 manager::~manager() {
     assert(_ep_managers.empty());
 }
diff --git a/db/hints/manager.hh b/db/hints/manager.hh
index 98954da14..24431e710 100644
--- a/db/hints/manager.hh
+++ b/db/hints/manager.hh
@@ -481,7 +481,8 @@ class manager : public service::endpoint_lifecycle_subscriber {
     seastar::named_semaphore _drain_lock = {1, named_semaphore_exception_factory{"drain lock"}};
 
 public:
-    manager(sstring hints_directory, const utils::updateable_value<host_filter>& filter, int64_t max_hint_window_ms, resource_manager&res_manager, distributed<database>& db);
+    explicit manager(sstring hints_directory, const utils::updateable_value<host_filter>& filter, int64_t max_hint_window_ms, resource_manager&res_manager, distributed<database>& db);
+    explicit manager(sstring hints_directory, host_filter filter, int64_t max_hint_window_ms, resource_manager&res_manager, distributed<database>& db);
     virtual ~manager();
     manager(manager&&) = delete;
     manager& operator=(manager&&) = delete;
diff --git a/service/storage_proxy.cc b/service/storage_proxy.cc
index e5dd88c33..8b46dfc8f 100644
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -1779,8 +1779,6 @@ inline std::ostream& operator<<(std::ostream& os, const hint_wrapper& h) {
 
 using namespace std::literals::chrono_literals;
 
-static thread_local const utils::updateable_value_source<db::hints::host_filter> allow_all_hint_filter{db::hints::host_filter(db::hints::host_filter::enabled_for_all_tag())};
-
 storage_proxy::~storage_proxy() {}
 storage_proxy::storage_proxy(distributed<database>& db, storage_proxy::config cfg, db::view::node_update_backlog& max_view_update_backlog,
         scheduling_group_key stats_key, gms::feature_service& feat, locator::token_metadata& tm)
@@ -1792,7 +1790,7 @@ storage_proxy::storage_proxy(distributed<database>& db, storage_proxy::config cf
     , _next_response_id(std::chrono::system_clock::now().time_since_epoch()/1ms)
     , _hints_resource_manager(cfg.available_memory / 10)
     , _hints_manager(_db.local().get_config().hints_directory(), _db.local().get_config().hinted_handoff_enabled, _db.local().get_config().max_hint_window_in_ms(), _hints_resource_manager, _db)
-    , _hints_for_views_manager(_db.local().get_config().view_hints_directory(), utils::updateable_value<db::hints::host_filter>(allow_all_hint_filter), _db.local().get_config().max_hint_window_in_ms(), _hints_resource_manager, _db)
+    , _hints_for_views_manager(_db.local().get_config().view_hints_directory(), db::hints::host_filter(db::hints::host_filter::enabled_for_all_tag()), _db.local().get_config().max_hint_window_in_ms(), _hints_resource_manager, _db)
     , _stats_key(stats_key)
     , _features(feat)
     , _background_write_throttle_threahsold(cfg.available_memory / 10)
```

db/hints/host_filter.cc
piodul: `host_filter`  will be moved/copied across shards. I'm not sure if it will be safe to cache a reference to the snitch_ptr, even if we initialize it with `locator::i_endpoint_snitch::get_local_snitch_ptr()` in (move|copy) (constructor|assignment operator).

db/hints/host_filter.cc
piodul: Ok, changed.

db/hints/host_filter.cc
piodul: Changed to an exception throw.

It's a pity that C++ does not have pattern matching that checks for exhaustiveness automatically...

db/hints/host_filter.cc
piodul: Done.

db/hints/host_filter.cc
piodul: I changed it so that the DC list is printed only if `_enabled_kind == host_filter::enabled_kind::enabled_selectively`

db/hints/host_filter.hh
piodul: I agree. Changed.

db/hints/host_filter.hh
piodul: This class is intended to represent a parse error. An abort in `host_filter.cc` will only happen in case of a bug in our logic, e.g. if somebody adds a variant to `host_filter` and forgets to cover it in a switch which was supposed to exhaust all variants.

I feel that `std::logic_error` might be a more suitable replacement for aborts.

db/config.cc
piodul: Done

service/storage_proxy.cc
piodul: Log message adjusted.

As for view hints, could you explain why this change would disable them? `_hints_for_views_manager` is initialized with a `host_filter` that accepts all DCs, so it should be enabled.

service/storage_proxy.hh
piodul: Done

service/storage_proxy.cc
piodul: Done - I made a separate constructor overload for `manager` that takes a non-updateable `host_filter` value.

service/storage_proxy.hh
piodul: In the case you described, SIGHUP won't cause observers to be notified, because the value that was re-read from config compares equal with the value that was parsed at startup (or at the previous reload). And yes, it is clearly an accident.

I'm not sure which way should have precedence, or if there should be precedence at all. I know that, for example, CLI arguments have precedence over configuration, so that if you specify `--hinted-handoff-enabled` in CLI, SIGHUP-reloading from config won't work for it.

An idea: maybe we should have a generic API for live-updating configuration parameters through REST? It would override CLI and YAML configuration when set, but could be cleared so that the CLI or config parameter is available again.

> We should route REST via config, not have two separate paths talking to the controlled object.

Could you clarify what you meant? Does it mean that REST should modify the `config` object directly so that SIGHUP in the scenario above would cause the host_filter to be updated?

service/storage_proxy.cc
vladzcloudius: I thought it was enabled with a default constructed hints_filter which originally was a "disable_all". If I misread it then I apologize.

db/hints/host_filter.cc
vladzcloudius: You can make it safe by implementing corresponding copy/move constructors.

service/storage_proxy.hh
avikivity: Too much time has passed, so I don't know if that's what I meant. It some seem reasonable to consolidate REST config changes into db::config. There is still the problem that there can be confusion between static (yaml) and dynamic (REST) configuration, but at least db::config/named_value can keep track of it (new config_source enum).

db/hints/host_filter.cc
avikivity: If you drop the default and one of the cases, does the compiler not complain?

Seems like it does: https://godbolt.org/z/9MT8Wj

db/hints/host_filter.cc
piodul: It looks like we can't do that, as `config` (which holds `host_filter` as a member) seems to be created earlier than the snitch instance. In boost tests, I got a lot of assertions similar to this:
```
row_cache_alloc_stress_test: /home/piodul/code/scylla-3/seastar/include/seastar/core/sharded.hh:715: Service& seastar::sharded<T>::local() [with Service = locator::snitch_ptr]: Assertion `local_is_initialized()' failed.
```

db/hints/host_filter.cc
piodul: Indeed it does, thanks for the tip! I'll move the throw outside the `default` clause.

db/hints/host_filter.cc
vladzcloudius: I guess it's easier to leave this `snitch` argument then.
Alternatives are worse.

avikivity:
@piodul I get github notifications for this, but they don't point to a comment, so I have no idea if I need to respond to something.

slivne:
@piodul please fix this pull request and lets get it in - its rotting and the field has asked for it - so lets get it in 

Please don;t add features - if there are followup requests to add / change those should be in a seperate issue.

db/hints/manager.cc
piodul: @vladzcloudius The v3 brings back the check, so marking as resolved.

api/storage_proxy.cc
piodul: I'm not sure if I understand this comment, and if it still applies...

There are two API endpoints for settings hints configuration at the moment: `set_hinted_handoff_enabled` and `set_hinted_handoff_enabled_by_dc_list`. First takes a "true/false" flag and enables/disables hints completely, and the second one enables hints selectively for a given list of DCs. Therefore there is some kind of separation on the HTTP API level.

The YAML configuration does mix up enabling/disabling hints altogether in a single option, but that's because we inherited it from Cassandra (see my comment [here](https://github.com/scylladb/scylla/pull/6916#discussion_r468422079)). We decided to change it another time.

service/storage_proxy.hh
piodul: The v3 avoids this problem for now by allowing to change the config through HTTP API only.

db/hints/host_filter.cc
piodul: Dropped the default case in v3, and added a `throw std::logic_error` at the end of the function.

db/hints/manager.cc
piodul: This shouldn't be a problem in v3 - changing configuration is now only possible through HTTP, it causes appropriate hint managers to be stopped/started, and the HTTP endpoint for changing configuration is now blocking and waits for the operation to complete.

piodul:
@avikivity @vladzcloudius 
Updated to v3:
- I removed the config live update feature and kept the HTTP API only. There was some seemingly unresolved discussion about HTTP API/config hot-reload interaction which would require additional changes to the configuration framework, so let's restrict it to HTTP API for now.
- Updating HH configuration is now blocking and causes appropriate endpoint managers to be started or stopped - now, configuration change affects both hint writing and sending.

I also addressed an issue which was present in previous versions - hint directories would be always created and verified, regardless if the path to the hints directory is valid. To be backwards compatible, now hint directories are initialized lazily (either at startup when HH is enabled, or the first time it is enabled).

v2 can be found here: https://github.com/piodul/scylla/tree/allow-changing-hinted-handoff-configuration-in-runtime-v2

piodul:
Ninja-update: removed one commit which was supposed to be included in a separate PR.

piodul:
@avikivity @vladzcloudius ping

utils/directories.cc
```diff
@@ -64,47 +62,46 @@ future<> directories::touch_and_lock(fs::path path) {
...
-                return distributed_loader::verify_owner_and_mode(path).handle_exception([](auto ep) {
-                    startlog.error("Failed owner and mode verification: {}", ep);
-                    return make_exception_future<>(ep);
+future<> directories::create_and_verify(directories::set dir_set) {
+    return do_with(std::vector<file_lock>(), [this, dir_set = std::move(dir_set)] (std::vector<file_lock>& locks) {
+        return parallel_for_each(dir_set.get_paths(), [this, &locks] (fs::path path) {
+            return touch_and_lock(path).then([path = std::move(path), developer_mode = _developer_mode, &locks] (file_lock lock) {
+                locks.emplace_back(std::move(lock));
+                return disk_sanity(path, developer_mode).then([path = std::move(path)] {
+                    return distributed_loader::verify_owner_and_mode(path).handle_exception([](auto ep) {
```
avikivity, line 97: Unrelated and minor: this call to distributed_loader breaks layering, calling from a more-or-less generic utility into application code.

No action needed.

db/hints/resource_manager.cc
```diff
@@ -65,15 +65,17 @@ size_t resource_manager::sending_queue_length() const {
...
     : _shard_managers(managers)
     , _per_device_limits_map(per_device_limits_map)
+    , _operation_lock(operation_lock)
 {}
 
 void space_watchdog::start() {
     _started = seastar::async([this] {
         while (!_as.abort_requested()) {
             try {
+                const auto units = get_units(_operation_lock, 1).get();
```
avikivity, line 78: Also unrelated and no action needed: better to put long running lambdas into named functions, so that if we have a stack trace with this code, it is attributed to the named function rather than start().

db/hints/resource_manager.hh
```diff
@@ -119,10 +120,45 @@ class resource_manager {
...
+    shared_ptr<service::storage_service> _ss_ptr;
+
+    enum class state {
+        running,
+        replay_allowed,
+    };
+    using state_set = enum_set<super_enum<state,
+        state::running,
+        state::replay_allowed>>;
+
```
avikivity, line 139: Shouldn't stop() be adjusted to consider the state?

avikivity:
Looks reasonable, though I admit I did not follow all the details.

What happens if we shut down while in the initialization phase?

What happens if hints directory creation fails?

piodul:
@avikivity

> What happens if we shut down while in the initialization phase?

Initialization does some operations on the filesystem which should be idempotent and retryable:

1. Create, validate and lock hints directories (`create_and_verify` from `directories.cc`)

   Directories needed by hints are created if they do not exist, and a `file_lock` is acquired on each of them.

   If we crash during that step, we can safely retry creating directories later. As for the locks, they use the `lockf` function underneath, so I guess they are released if the process exits. If we encounter an exception in `create_and_verify`, the `file_lock` objects that we managed to create in this invocation are dropped.

2. Hint segments rebalancing (`manager::rebalance`)

   The hints directory contains a single subdirectory for each shard. Each subdirectory keeps a list of segments with hints. In this step, we list all segments in all subdirectories and try to move them between shard subdirectories so that each one of them gets floor(num_segments / num_shards) or ceil(num_segments / num_shards).

   After that, we remove unnecessary hint subdirectories - if the node was re-run with a lower number of shards, some subdirectories will be now empty and can be safely removed.

Those steps mimic what we have been already doing for hints during initialization. 

> What happens if hints directory creation fails?

If the process described above fails at startup, we just stop the initialization and return a non-zero exit code. Some directories may be left, and some segments may be rearranged, but restarting Scylla shouldn't have any problems with retrying.

If this fails when we try to enable hints manager through the HTTP API for the first time, then the process should be safe to retry. The only thing that needs any rollback is dropping the file locks (which we do here), other things (creating directories, rebalancing segments) should be safely retryable.

db/hints/resource_manager.hh
piodul: I'm not sure what you mean. `resource_manager::stop()` calls `unset_running()`, so the running bit is cleared when we stop.

avikivity:
We still have a conflict between REST and configuration file. The user must update the configuration file if they use REST, or on restart, the change is lost.

avikivity:
(not that I have a suggestion on how to fix that)

vladzcloudius (changes requested):

db/hints/resource_manager.cc
```diff
@@ -186,19 +192,38 @@ future<> resource_manager::start(shared_ptr<service::storage_proxy> proxy_ptr, s
...
+        // Already registered
+        return make_ready_future<>();
+    }
+    if (!running()) {
+        return make_ready_future<>();
+    }
+
+    return m.start(_proxy_ptr, _gossiper_ptr, _ss_ptr).then([this, &m] {
+        return with_semaphore(_operation_lock, 1, [this, &m] () {
+            return prepare_per_device_limits().then([this, &m] {
```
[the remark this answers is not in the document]
piodul: > I'm not sure I understand this code: all managers are started from resource_manager::start().
> The same is about calling prepare_per_device_limits().
> And the call sequence in storage_proxy::start_hints_manager(..) is that we first call resource_manager::register_manager(...) and then we call resource_manager::start().

The idea is that now, `resource_manager::register_manager(...)` can also be called _after_ the `resource_manager` is started. This will happen if you didn't enable regular hints at boot time, but later tried to do this via HTTP API.

I adjusted the `resource_manager::register_manager(...)` method so that registering of a hints manager is possible after boot. It also automatically starts and allows hints replay, if needed - we need the `resource_manager::state` enum because of that.

It looks like Avi merged this PR already, so I'll send a followup which add comments with explanation.

> prepare_per_device_limits() doesn't seem to be idempotent, nor is manager::start() (it's actually a bug to call it twice).

You are right about `prepare_per_device_limits()`. I mistakenly thought it was idempotent because it checks if manager's device_id is already present in the `_per_device_limits_map`. I didn't notice that managers are unconditionally appended to the `per_device_limits::managers` field.

The impact of this seems to be that the `resource_manager` will count the space used by view hints segments twice, if we enable the regular hints manager in runtime. I'll fix this in the followup, too.

The `hints::manager` shouldn't be started twice. `resource_manager::register_manager(...)` does nothing if it sees that the hints manager you are trying to register was registered before.

db/hints/resource_manager.cc
```diff
@@ -176,8 +178,12 @@ void space_watchdog::on_timer() {
 }
 
 future<> resource_manager::start(shared_ptr<service::storage_proxy> proxy_ptr, shared_ptr<gms::gossiper> gossiper_ptr, shared_ptr<service::storage_service> ss_ptr) {
-    return parallel_for_each(_shard_managers, [proxy_ptr, gossiper_ptr, ss_ptr](manager& m) {
-        return m.start(proxy_ptr, gossiper_ptr, ss_ptr);
+    _proxy_ptr = std::move(proxy_ptr);
+    _gossiper_ptr = std::move(gossiper_ptr);
+    _ss_ptr = std::move(ss_ptr);
+    set_running();
```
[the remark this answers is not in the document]
piodul: The `resource_manager` is started when Scylla boots. If `resource_manager::start()` fails, then the boot process is stopped.

If somebody is fast enough to try to enable hints through HTTP API before `resource_manager::start()` is finished, then we might have a race.

To be safe, I'll put `set_running()` at the end and will protect `resource_manager::start()` with the `_operation_lock` semaphore.

piodul:
@avikivity @vladzcloudius sent a followup here: https://github.com/scylladb/scylla/pull/7660

PR #6917, opened 2020-07-24
gms: Return full schema_features from cluster_schema_features

### LINKED ISSUE #6893: SSTable <name> uses CDCPartitioner partitioner which is different than Murmur3Partitioner partitioner used by the database
Steps to reproduce:

1. start node with CDC enabled
2. create a table with CDC enabled
3. insert data into the table
4. nodetool flush
5. restart node

result:
```
ERROR 2020-07-21 14:37:12,462 [shard 0] init - Startup failed: std::runtime_error (Exception while populating keyspace 'ks' with column family 't_scylla_cdc_log' from file 'tmp/ks/t_scylla_cdc_log-d8f601b2cb4e11eaace5000000000000': std::runtime_error (SSTable tmp/ks/t_scylla_cdc_log-d8f601b2cb4e11eaace5000000000000/mc-3-big-Data.db uses com.scylladb.dht.CDCPartitioner partitioner which is different than org.apache.cassandra.dht.Murmur3Partitioner partitioner used by the database))
```
By the way, it's funny that none of our tests catched that (I catched it by accident)

haaawk:
when it's called before gossip is enabled.

Without this change schemas loaded from disk at startup are stripped
from new fields guarded by feature flags.

Fixes #6893

### DIFF
--- a/gms/feature_service.cc
+++ b/gms/feature_service.cc
@@ -25,6 +25,7 @@
 #include "db/config.hh"
 #include "gms/feature.hh"
 #include "gms/feature_service.hh"
+#include "gms/gossiper.hh"
 
 namespace gms {
 constexpr std::string_view features::RANGE_TOMBSTONES = "RANGE_TOMBSTONES";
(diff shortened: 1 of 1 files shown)


gms/feature_service.cc
```diff
@@ -230,6 +231,9 @@ void feature::enable() {
 }
 
 db::schema_features feature_service::cluster_schema_features() const {
+    if (!get_local_gossiper().is_enabled()) {
+        return db::schema_features::full();
+    }
```
avikivity, line 236: Seems very dangerous. Features should be monotonicically increasing with time. With the change, they can decrease.

The idea around features is that you don't use something before you are sure everyone understands it. If you remove features, you violated this principle.

haaawk: Fair point. I guess I could just make sure that the codepath run before gossip is started to not remove anything from the schema at all when it loads tables schemas from disk.

Issue #6868, opened 2020-07-17
Segmentation Fault and coredump on other 2 nodes, when decommission process was interrupted with hardboot on 3rd node


aleksbykov:
*Installation details*
Scylla version (or git commit hash): 4.2.rc1-0.20200709.51d4efc3212 with build-id fe5f4c9282b21b22bbcc986f8c43836f6d6cf32d
Cluster size: 6
OS (RHEL/CentOS/Ubuntu/AWS AMI): ami-03396516d7bd6a5fa

job [link](https://jenkins.scylladb.com/view/scylla-4.2/job/scylla-4.2/job/longevity/job/longevity-1tb-7days-test/6/) longevity-1Tb-7d failed after 1d10h. 

Before issue over cluster runs different nemesises(actions) without any issues: TerminateAndReplaceNode, AbortRepair, Major compaction, Drainer.
CLuster was under stress command:
```
"cassandra-stress write         cl=QUORUM n=1100200300 -schema 'replication(factor=3)                               compaction(strategy=LeveledCompactionStrategy)'    -port jmx=6868 -mode cql3 native                     -rate threads=1000 -col 'size=FIXED(200) n=FIXED(5)' -pop seq=1..1100200300",
"cassandra-stress write         cl=QUORUM n=50000000   -schema 'replication(factor=3) compression=LZ4Compressor     compaction(strategy=SizeTieredCompactionStrategy)' -port jmx=6868 -mode cql3 native compression=lz4     -rate threads=50 -pop seq=1..50000000 -log interval=5",
"cassandra-stress write         cl=QUORUM n=50000000   -schema 'replication(factor=3) compression=SnappyCompressor  compaction(strategy=SizeTieredCompactionStrategy)' -port jmx=6868 -mode cql3 native compression=snappy  -rate threads=50 -pop seq=1..50000000 -log interval=5",
"cassandra-stress write         cl=QUORUM n=50000000   -schema 'replication(factor=3) compression=DeflateCompressor compaction(strategy=SizeTieredCompactionStrategy)' -port jmx=6868 -mode cql3 native compression=none    -rate threads=50 -pop seq=1..50000000 -log interval=5",
"cassandra-stress write         cl=QUORUM n=50000000   -schema 'replication(factor=3) compression=ZstdCompressor    compaction(strategy=SizeTieredCompactionStrategy)' -port jmx=6868 -mode cql3 native compression=none    -rate threads=50 -pop seq=1..50000000 -log interval=5"

[...]

Issue #6853, opened 2020-07-16
Segmentation fault and coredump detected on node, while another node running repair process second time after repair was interrupted 1st time


aleksbykov:
*Installation details*
Scylla version (or git commit hash): 4.2.rc1-0.20200709.51d4efc3212 with build-id fe5f4c9282b21b22bbcc986f8c43836f6d6cf32d
Cluster size: 6
OS (RHEL/CentOS/Ubuntu/AWS AMI): ami-0d2b45972f9c49ab3

Job [link](https://jenkins.scylladb.com/job/scylla-4.2/job/longevity/job/longevity-cdc-100gb-4h-test/10/) is running with 
 6 nodes Scylla cluster and enabled CDC.
Cluster was under stress of next c-s commands for testing cdc functionality
```
stress_cmd=cassandra-stress user no-warmup profile=/tmp/cdc_profile.yaml ops'(insert=2,read1=1,update_number=1,update_name=1,delete1=1)' cl=QUORUM duration=240m -port jmx=6868 -mode cql3 native -rate threads=100 -node 10.0.136.206 -errors skip-unsupported-columns

stress_cmd=cassandra-stress user no-warmup profile=/tmp/cdc_profile_preimage.yaml ops'(insert=2,read1=1,update_number=1,update_name=1,delete1=1)' cl=QUORUM duration=240m -port jmx=6868 -mode cql3 native -rate threads=100 -node 10.0.136.206 -errors skip-unsupported-columns

stress_cmd=cassandra-stress user no-warmup profile=/tmp/cdc_profile_postimage.yaml ops'(insert=2,read1=1,update_number=1,update_name=1,delete1=1)' cl=QUORUM duration=240m -port jmx=6868 -mode cql3 native -rate threads=100 -node 10.0.136.206 -errors skip-unsupported-columns

stress_cmd=cassandra-stress user no-warmup profile=/tmp/cdc_profile_preimage_postimage.yaml ops'(insert=2,read1=1,update_number=1,update_name=1,delete1=1)' cl=QUORUM duration=240m -port jmx=6868 -mode cql3 native -rate threads=100 -node 10.0.136.206 -errors skip-unsupported-columns

```

Scylla was stopped on node3, several sstables files were removed.
```
sdcm.nemesis.ChaosMonkey: Selected files for destroy: /var/lib/scylla/data/cdc_test/test_table_postimage-81b2ec80c63311ea9ac8000000000003/mc-3922-*
[...]

PR #6932, opened 2020-07-28
repair: Fix race between create_writer and wait_for_writer_done

asias:
We saw scylla hit user after free in repair with the following procedure during tests:

- n1 and n2 in the cluster

- n2 ran decommission

- n2 sent data to n1 using repair

- n2 was killed forcely

- n1 tried to remove repair_meta for n1

- n1 hit use after free on repair_meta object

This was what happened on n1:

1) data was received -> do_apply_rows was called -> yield before create_writer() was called

2) repair_meta::stop() was called -> wait_for_writer_done() / do_wait_for_writer_done was called
   with _writer_done[node_idx] not engaged

3) step 1 resumed, create_writer() was called and _repair_writer object was referenced

4) repair_meta::stop() finished, repair_meta object and its member _repair_writer was destroyed

5) The fiber created by create_writer() at step 3 hit use after free on _repair_writer object

To fix, we should call wait_for_writer_done() after any pending
operations were done which were protected by repair_meta::_gate. This
prevents wait for writer done finishes before the writer is in the
process of being created.

Fixes: #6853
Fixes: #6868
Backports: 4.0, 4.1, 4.2

### DIFF
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -756,11 +756,12 @@ public:
 public:
     future<> stop() {
         auto gate_future = _gate.close();
-        auto writer_future = _repair_writer.wait_for_writer_done();
         auto f1 = _sink_source_for_get_full_row_hashes.close();
         auto f2 = _sink_source_for_get_row_diff.close();
         auto f3 = _sink_source_for_put_row_diff.close();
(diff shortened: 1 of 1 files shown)


bhalevy:
@scylladb/scylla-maint please merge

PR #6943, opened 2020-07-28
schema_tables: redact schema for merge schema but not for loading it

### LINKED ISSUE #6893: SSTable <name> uses CDCPartitioner partitioner which is different than Murmur3Partitioner partitioner used by the database

haaawk:
Previously redaction was added for all the paths that use
read_schema_partition_for_table.

This patch leaves the redaction only for read_tables_for_keyspaces
which is used in do_merge_schema.

Other codepaths which initiatie from database::parse_system_tables
are not affected by redaction any more and that means schemas loaded
from disk to memory are not redacted.

Fixes #6893

### DIFF
--- a/db/schema_tables.cc
+++ b/db/schema_tables.cc
@@ -779,9 +779,7 @@ read_schema_partition_for_table(distributed<service::storage_proxy>& proxy, sche
             .with_range(std::move(clustering_range))
             .build();
     auto cmd = make_lw_shared<query::read_command>(schema->id(), schema->version(), std::move(slice), query::max_rows);
-    return query_partition_mutation(proxy.local(), std::move(schema), std::move(cmd), std::move(keyspace_key)).then([&proxy] (mutation mut) {
-         return redact_columns_for_missing_features(std::move(mut), proxy.local().get_db().local().features().cluster_schema_features());
-    });
+    return query_partition_mutation(proxy.local(), std::move(schema), std::move(cmd), std::move(keyspace_key));
 }
 
 future<mutation>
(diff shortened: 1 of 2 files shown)


haaawk:
Does that matter as long as a rebooted node still redacts schema for every internode communication?

From the point of view of other nodes, it would be impossible to tell the difference, no?

tgrabiec:
If the rebooted node is the new node, it will materialize schema_ptr with a new version, one which old nodes will not be able to materialize. Requests from that node to an old node will fail.

haaawk:
Ok. Then it seems that we just shouldn't allow schema changes on mixed cluster or is there other option?

tgrabiec:
I think there is another option. We can avoid setting schema columns which are not supported by the whole cluster yet.

haaawk:
You mean on coordinator or on replicas?

tgrabiec:
On the coordinator.

slivne:
@haaawk is this still relevant if not please close

haaawk:
It is we just need to take different approach to the problem.

PR #6955, opened 2020-07-30
Revert "schema: take into account features when converting a table cr…

eliransin:
…eation to"

This reverts commit b97f466.

It turns out that the schema mechanism has a lot of nuances,
after this change, for unknown reason, it was empirically
proven that the amount of cross shard on an upgraded node was
increased significantly with a steady stress traffic, if
was so significant that the node appeared unavailable to
the coordinators because all of the requests started to fail
on smp_srvice_group semaphore.

This revert will bring back a caveat in Scylla, the caveat is
that creating a table in a mixed cluster **might** under certain
condition cause schema mismatch on the newly created table, this
make the table essentially unusable until the whole cluster has
a uniform version (rolling upgrade or rollback completion).

### DIFF
--- a/db/schema_tables.cc
+++ b/db/schema_tables.cc
@@ -780,9 +780,7 @@ read_schema_partition_for_table(distributed<service::storage_proxy>& proxy, sche
             .build();
     auto cmd = make_lw_shared<query::read_command>(schema->id(), schema->version(), std::move(slice), proxy.local().get_max_result_size(slice),
             query::row_limit(query::max_rows));
-    return query_partition_mutation(proxy.local(), std::move(schema), std::move(cmd), std::move(keyspace_key)).then([&proxy] (mutation mut) {
-         return redact_columns_for_missing_features(std::move(mut), proxy.local().get_db().local().features().cluster_schema_features());
-    });
+    return query_partition_mutation(proxy.local(), std::move(schema), std::move(cmd), std::move(keyspace_key));
 }
 
 future<mutation>


nyh:
@eliransin you asked me again for a review, but as I noted above, I cannot meaningfully review this patch because I don't understand this area. @tgrabiec @tgrabiec @avikivity if you want me - or anyone - to understand the schema upgrade, redacting, how its unit tests work etc. - please write a document about it, and maybe next time I can review it.

That being said, I feel pretty confident to commit this patch *without* reviewing it, given that it is reverting to an old and known problem - not introducing a new problem. So if for some reason @tgrabiec (who understands this area well) can't review this patch, tell me, and I'll just commit this patch.

haaawk:
@nyh this is a revert of a change @eliransin did a few days ago and that broke loading schemas to memory from disk.

For more context see https://github.com/scylladb/scylla/issues/6893#issuecomment-665031010

haaawk:
To be more precise:

>There is a problem here indeed. Suppose node A has the feature enabled but node B doesn't have it enabled yet. Node A will send requests referencing the new (unredacted) table schema version. Node B will receive the mutations and will be able to apply them, but will still redact them when materializing the schema object. As a result the table schema version requesting will fail, since we'll materialize the wrong schema version.
I now think that b97f466 is not the right way to fix the problem it aimed at fixing. Instead, schema change commands should not be setting any columns which are not supported yet, so that this redacting is not needed. Lack of the value in the new column should be interpreted as if the default value was there. @eliransin, will that fix your bug as well?

slivne:
@tgrabiec / @avikivity can one of you please review and merge the revert ...

nyh:
@haaawk yes, I understood the context, and as I said this is why I can commit this patch without review. But to review the code (meaning I understand the supposed bug it fixes and the supposed bug it re-introduces) - this I can't. Maybe this is a github bug - @eliransin asked me to "review" this patch, whereas he should have asked me to "merge" it :-)

tgrabiec:
Merged

PR #6961, opened 2020-07-31
transport: Added listener with port-based load balancing

### LINKED ISSUE #5239: connection storm when attempting to achieve shard-per-connection
When a client attempts to establish a connection per shard, it generates a large amount of connections until (randomly) it manages to reach all shards, at which point it can drop the old ones. If there are many clients connecting simultaneouly (this can happen when a node restarts) there can be a huge number of connections happening at once, increasing the load on the node.

jul-stas:
This is inspired by #6781. The idea is to make Scylla listen for CQL connections on port 9042 (where both old shard-aware and shard-unaware clients can still connect the traditional way). On top of that I added a new port, where everything works the same way, only the port from client's socket used to determine the shard No. to connect to. Desired shard No. is the result of `clientside_port % num_shards`.

The new port is configurable from scylla.yaml and defaults to 19042 (unencrypted, unless user configures encryption options and omits `native_shard_aware_transport_port_ssl` in DB config).

Two "SUPPORTED" tags are added: "SCYLLA_SHARD_AWARE_PORT" and "SCYLLA_SHARD_AWARE_PORT_SSL". For compatibility, "SCYLLA_SHARDING_ALGORITHM" is still kept.

Fixes #5239

### DIFF
--- a/transport/server.cc
+++ b/transport/server.cc
@@ -211,7 +211,7 @@ future<> cql_server::stop() {
 }
 
 future<>
-cql_server::listen(socket_address addr, std::shared_ptr<seastar::tls::credentials_builder> creds, bool keepalive) {
+cql_server::listen(socket_address addr, std::shared_ptr<seastar::tls::credentials_builder> creds, bool is_shard_aware, bool keepalive) {
     auto f = make_ready_future<shared_ptr<seastar::tls::server_credentials>>(nullptr);
     if (creds) {
         f = creds->build_reloadable_server_credentials([](const std::unordered_set<sstring>& files, std::exception_ptr ep) {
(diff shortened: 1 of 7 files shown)


transport/server.cc
```diff
@@ -1227,6 +1230,12 @@ std::unique_ptr<cql_server::response> cql_server::connection::make_supported(int
         opts.insert({"SCYLLA_SHARD", format("{:d}", this_shard_id())});
         opts.insert({"SCYLLA_NR_SHARDS", format("{:d}", smp::count)});
         opts.insert({"SCYLLA_SHARDING_ALGORITHM", dht::cpu_sharding_algorithm_name()});
+        if (_server._config.shard_aware_transport_port) {
+            opts.insert({"SCYLLA_SHARD_AWARE_PORT", format("{:d}", *_server._config.shard_aware_transport_port)});
+        }
+        if (_server._config.shard_aware_transport_port_ssl) {
+            opts.insert({"SCYLLA_SHARD_AWARE_PORT_SSL", format("{:d}", *_server._config.shard_aware_transport_port_ssl)});
+        }
```
avikivity, line 1238: docs/protocol-extensions.md

PR #6967, opened 2020-08-03
cdc: add an abstraction for building log mutations

kbr-:
This commit takes out some responsibilities of `cdc::transformer`
(which is currently a big ball of mud) into a separate class.

This class is a simple abstraction for creating entries in a CDC log
mutation.

Low-level calls to the mutation API (such as `set_cell`) inside
`cdc::transformer` were replaced by higher-level calls to the builder
abstraction, removing some duplication of logic.

tests: unit(dev)

Refs: https://github.com/scylladb/scylla/issues/5945

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -705,15 +705,126 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
     return utils::UUID_gen::get_random_time_UUID_from_micros(t);
 }
 
+class log_mutation_builder {
+    const schema& _base_schema;
+    const schema& _log_schema;
+    const column_definition& _op_col;
+    const column_definition& _ttl_col;
+
+    // The base mutation's partition key
+    std::vector<bytes> _base_pk;
+
+    // The cdc$time value of created rows
+    const bytes _tuuid;
+    // The timestamp of the created log mutation cells
+    const api::timestamp_type _ts;
+    // The ttl of the created log mutation cells
+    const ttl_opt _ttl;
+
+    // Keeps the next cdc$batch_seq_no value
+    int _batch_no = 0;
+
+    // The mutation under construction
+    mutation& _log_mut;
+
+public:
+    log_mutation_builder(mutation& log_mut, api::timestamp_type ts,
+                         const partition_key& base_pk, const schema& base_schema)
+        : _base_schema(base_schema), _log_schema(*log_mut.schema()),
+          _op_col(*_log_schema.get_column_definition(log_meta_column_name_bytes("operation"))),
+          _ttl_col(*_log_schema.get_column_definition(log_meta_column_name_bytes("ttl"))),
+          _base_pk(base_pk.explode(_base_schema)),
+          _tuuid(timeuuid_type->decompose(generate_timeuuid(ts))),
+          _ts(ts),
+          _ttl(_base_schema.cdc_options().ttl()
+                  ? std::optional{std::chrono::seconds(_base_schema.cdc_options().ttl())} : std::nullopt),
+          _log_mut(log_mut)
+    {}
+
+    // Creates a new clustering row in the mutation, assigning it the next `cdc$batch_seq_no`.
+    // The numbering of batch sequence numbers starts from 0.
+    clustering_key allocate_new_log_row() {
+        auto log_ck = clustering_key::from_exploded(_log_schema, { _tuuid, int32_type->decompose(_batch_no++) });
+        set_key_columns(log_ck, _base_schema.partition_key_columns(), _base_pk);
+        return log_ck;
+    }
+
+    // A common pattern is to allocate a row and then immediately set its `cdc$operation` column.
+    clustering_key allocate_new_log_row(operation op) {
+        auto log_ck = allocate_new_log_row();
+        set_operation(log_ck, op);
+        return log_ck;
+    }
+
+    // Each clustering key column in the base schema has a corresponding column in the log schema with the same name.
+    // This takes a base schema clustering key prefix and sets these columns
+    // according to the prefix' values for the given log row.
+    void set_clustering_columns(const clustering_key& log_ck, const clustering_key_prefix& base_ckey) {
+        set_key_columns(log_ck, _base_schema.clustering_key_columns(), base_ckey.explode(_log_schema));
+    }
+
+    // Sets the `cdc$operation` column for the given row.
+    void set_operation(const clustering_key& log_ck, operation op) {
+        _log_mut.set_cell(log_ck, _op_col, atomic_cell::make_live(
+                    *_op_col.type, _ts, _op_col.type->decompose(operation_native_type(op)), _ttl));
+    }
+
+    // Sets the `cdc$ttl` column for the given row.
+    // Warning: if the caller wants `cdc$ttl` to be null, they shouldn't call `set_ttl` with a non-null value.
+    // Calling it with a non-null value and then with a null value will keep the non-null value.
+    void set_ttl(const clustering_key& log_ck, ttl_opt ttl) {
+        if (ttl) {
+            _log_mut.set_cell(log_ck, _ttl_col, atomic_cell::make_live(
+                        *_ttl_col.type, _ts, _ttl_col.type->decompose(ttl->count()), _ttl));
+        }
+    }
+
+    // Each regular and static column in the base schema has a corresponding column in the log schema with the same name.
+    // Given a reference to such a column from the base schema, this function sets the corresponding column
+    // in the log to the given value for the given row.
+    void set_value(const clustering_key& log_ck, const column_definition& base_cdef, bytes value) {
+        auto& log_cdef = *_log_schema.get_column_definition(log_data_column_name_bytes(base_cdef.name()));
+        _log_mut.set_cell(log_ck, log_cdef, atomic_cell::make_live(*base_cdef.type, _ts, std::move(value), _ttl));
+    }
+
+    // Each regular and static column in the base schema has a corresponding column in the log schema
+    // with boolean type and the name constructed by prefixing the original name with ,,cdc$deleted_''
+    // Given a reference to such a column from the base schema, this function sets the corresponding column
+    // in the log to `true` for the given row. If not called, the column will be `null`.
+    void set_deleted(const clustering_key& log_ck, const column_definition& base_cdef) {
+        _log_mut.set_cell(log_ck, log_data_column_deleted_name_bytes(base_cdef.name()), data_value(true), _ts, _ttl);
+    }
+
+    // Each regular and static non-atomic column in the base schema has a corresponding column in the log schema
+    // whose type is a frozen `set` of keys (the types of which depend on the base type) and whose name is constructed
+    // by prefixing the original name with ,,cdc$deleted_elements_''.
+    // Given a reference to such a column from the base schema, this function sets the corresponding column
+    // in the log to the given set of keys for the given row.
+    void set_deleted_elements(const clustering_key& log_ck, const column_definition& base_cdef, bytes deleted_elements) {
+        auto& log_cdef = *_log_schema.get_column_definition(log_data_column_deleted_elements_name_bytes(base_cdef.name()));
+        _log_mut.set_cell(log_ck, log_cdef, atomic_cell::make_live(*log_cdef.type, _ts, deleted_elements, _ttl));
+    }
+
+private:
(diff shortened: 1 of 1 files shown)


cdc/log.cc
```diff
@@ -767,64 +854,21 @@ class transformer final : public change_processor {
...
     std::vector<mutation> _result_mutations;
 
-    // Keeps the next cdc$batch_seq_no value
-    int _batch_no = -1;
-    // The timestamp of changes being currently processed
-    api::timestamp_type _ts;
-    // The cdc$time value of changes being currently processed
-    bytes _tuuid;
+    std::unique_ptr<log_mutation_builder> _builder;
+
```
elcallio, line 858: Why unique_ptr instead of just member?

cdc/log.cc
```diff
@@ -675,15 +675,102 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+        }
+    }
+
+    void set_operation(const clustering_key& log_ck, operation op) {
+        _log_mut.set_cell(log_ck, _op_col, atomic_cell::make_live(
+                    *_op_col.type, _ts, _op_col.type->decompose(operation_native_type(op)), _ttl));
+    }
+
+    void set_ttl(const clustering_key& log_ck, ttl_opt ttl) {
+        if (ttl) {
```
elcallio, line 747: I think this should be `opt_set_ttl` and in turn call an unconditional `set_ttl` (non-opt arg). It is weird with setters that are internally optional imho.

cdc/log.cc
```diff
@@ -905,24 +947,12 @@ class transformer final : public change_processor {
     void generate_image(operation op, const clustering_key* ck, const one_kind_column_set* affected_columns) {
         assert(op == operation::pre_image || op == operation::post_image);
 
-        auto& res = current_mutation();
-
-        auto image_ck = allocate_new_log_row();
-        set_operation(image_ck, op);
+        assert(_builder);
+        auto image_ck = _builder->allocate_new_log_row();
+        _builder->set_operation(image_ck, op);
```
elcallio, line 952: Maybe some overloads in builder for `allocate_new_log_row(op)` and similar?

cdc/log.cc
```diff
@@ -675,15 +675,102 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+          _tuuid(timeuuid_type->decompose(generate_timeuuid(ts))),
+          _ts(ts),
+          _ttl(_base_schema.cdc_options().ttl()
+                  ? std::optional{std::chrono::seconds(_base_schema.cdc_options().ttl())} : std::nullopt),
+          _log_mut(log_mut)
+    {}
+
+    clustering_key allocate_new_log_row() {
+        const auto log_ck = clustering_key::from_exploded(_log_schema, { _tuuid, int32_type->decompose(_batch_no++) });
+        auto pk_value = _base_pk.explode(_base_schema);
```
haaawk, line 716: Would it make sense to store pk_value as a class field instead of _base_pk?

cdc/log.cc
```diff
@@ -675,15 +675,102 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+    void set_ttl(const clustering_key& log_ck, ttl_opt ttl) {
+        if (ttl) {
+            _log_mut.set_cell(log_ck, _ttl_col, atomic_cell::make_live(
+                        *_ttl_col.type, _ts, _ttl_col.type->decompose(ttl->count()), _ttl));
+        }
+    }
+
+    void set_value(const clustering_key& log_ck, const column_definition& base_cdef, bytes value) {
+        auto& log_cdef = *_log_schema.get_column_definition(log_data_column_name_bytes(base_cdef.name()));
+        _log_mut.set_cell(log_ck, log_cdef, atomic_cell::make_live(*base_cdef.type, _ts, value, _ttl));
```
haaawk, line 755: Can we std::move(value) here? If not does it make sense to take value argument by const&?

cdc/log.cc
haaawk: Probably because log_mutation_builder is not movable due to const member field, no?

haaawk: I think it would be better if we didn't need to have a unique_ptr here.

haaawk: If const fields force us to use allocation here we should just drop them. It's sad but that's what we have with C++.

kbr-: I don't like the `unique_ptr`, but I don't want to have to move-construct / move-assign the builder either.

The problem is with how the flow is currently structured: we first call `begin_timestamp()` which initializes the state, then `process_change` which uses the state. In the call of `begin_timestamp` we have to assign `_builder`.

I was thinking about two alternative approaches:
- `begin_timestamp` returns a builder, and then the caller must pass it to `process_change`. It's nice because the builder serves as a witness (proof) to `process_change` that `begin_timestamp` was called, so we have a compile-time guarantee, we don't need to `assert`.
- have a `with_new_timestamp(callback)` function instead of `begin_timestamp`, where `callback` takes a builder as a parameter. The callback would call `process_change`. Or something.

haaawk: 1. constructing builder in begin_timestamp is not worse than constructing unique_ptr to the builder
2. I agree it is error-prone to rely on the order of calls (begin_timestamp first)
3. The first of the options you presented seems more elegant but is still error-prone because a user may pass a wrong builder.
4. The second option is safer but it is also a bit cumbersome.
5. Both options expose the internals (the builder) which is not great.

Maybe option two but without passing builder to the callback would be better?

cdc/log.cc
```diff
@@ -675,15 +675,102 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+            assert (pos < pk_value.size());
+
+            auto& cdef = *_log_schema.get_column_definition(log_data_column_name_bytes(column.name()));
+            _log_mut.set_cell(log_ck, cdef, atomic_cell::make_live(*column.type, _ts, bytes_view(pk_value[pos]), _ttl));
+            ++pos;
+        }
+        return log_ck;
+    }
+
+    void set_clustering_columns(const clustering_key& log_ck, const clustering_key_prefix& base_ckey) {
```
nyh, line 728: It seems this and the previous method have almost the same code. Could they call each other, or call a common method?

cdc/log.cc
```diff
@@ -767,64 +854,21 @@ class transformer final : public change_processor {
 
     std::vector<mutation> _result_mutations;
 
-    // Keeps the next cdc$batch_seq_no value
-    int _batch_no = -1;
-    // The timestamp of changes being currently processed
-    api::timestamp_type _ts;
-    // The cdc$time value of changes being currently processed
-    bytes _tuuid;
+    std::unique_ptr<log_mutation_builder> _builder;
```
nyh, line 857: Question about the lifetime of this _builder. It holds a bunch of references, including one for a schema. How do we guarantee that all of these references remain valid? Is the "transformer" parent object holding both a _builder and the objects it points to (like schema_ptr and the various keys etc.)?

nyh:
Also, @kbr- github tells me there is a conflict in cdc/log.cc so please rebase.

haaawk:
Conflict is probably due to no delta change that got merged.

kbr-:
Rebased, but still need to address review issues.

cdc/log.cc
kbr-: I think of it as follows:
- if `ttl` is non-null, we set `cdc$ttl` to its value
- if it's null, we set `cdc$ttl` to `null`, and the fact that the setter doesn't need to do anything is an implementation detail.

kbr-: To make it completely kosher the setter should actually somehow set it to `null` if the passed ttl is null (e.g. by removing a cell if it exists). Because what if we call `set_ttl` with non-null and then want to revert? But meh.

elcallio: Then please comment on the logic so the person needing to explicitly null something knows the if is because of said assumption.

kbr-:
After dumb rebasing the code doesn't compile because `adjust_or_delete_deltas()` uses `_op_col`, which I moved from `transformer` into `builder`.

Hence question (@haaawk @elcallio what do you think?):
- should I bring `_op_col` back to `transformer`?
- or should I move `adjust_or_delete_deltas()` logic into `_builder`? It would then adjust just the current mutation (`_log_mut`). `begin_timestamp` would call it each time we replace a builder, and there would be one more call in `finish()`. While I don't like the fact that we'd have to remember calling `adjust_or_delete` so many times, it does feel like the logic belongs to the builder.

cdc/log.cc
kbr-: Can't guarantee it in compile-time (it's not Rust unfortunately), but yes, the idea is that `transformer` keeps those things alive at least as long as `_builder`.

haaawk:
> * ould then adjust just the current mutation (`_log_mut`). `begin_timestamp` would call it each time we replace a builder, and there would be one more call in `finish()`. While I don't like the fact that we'd have to remember calling `adjust_or_delete` so many times, it does feel like the logic belongs to the builder.

There's a third option. Just define _op_col inside adjust_or_delete_deltas or in some context around it but as minimal as possible.

cdc/log.cc
kbr-: It would

cdc/log.cc
kbr-: We can

cdc/log.cc
kbr-: What do you mean "without passing builder to the callback"?

`process_change` needs to get the builder somehow. It either gets it from state (so we have to overwrite the state variable, which requires a move/copy assignment operator, or there must be a pointer), or from a parameter (in which case we could pass a reference).

haaawk: I guess I was trying to encapsulate the builder inside this class. You're right that it would require assignment to a field.

I guess what I was trying to say is that:
1. assignment per se is not a problem
2. the problem is that we have no control over the order user calls member function and they can forget to call begin_timestamp

so I would either
1. just use assignment instead of unique_ptr since unique_ptr brings no value. That would leave code in a state it was before no better no worse. Which is ok for refactoring, I guess.
or
2. Think of a way to make sure that begin_timestamp is always called

The first approach you've listed does not solve it because you could do:
```
auto builder = begin_timestamp(...)
process_changes(builder)
process_changes(builder)
```

The second approach is a bit better but it leads to a more twisted code. I guess we could go with this second approach but I'm not super sure it's that much better so I would vote for just going with assignment for now until we put some more thoughts into this and maybe find some more elegant solution.

kbr-: I just found a solution.
https://en.cppreference.com/w/cpp/utility/optional/emplace

cdc/log.cc
kbr-: I documented the methods, and included a comment about calling `set_ttl` with null/non-null.

cdc/log.cc
kbr-: Done

cdc/log.cc
kbr-: Done

cdc/log.cc
kbr-: I see one potential problem: if someone moves the `transformer` after constructing the `builder`, some references might get invalidated.

Although now that I think of it, none of them would:
- schemas are kept alive by other entities
- column definitions refer to things inside the schemas
- _log_mut does refer to an element in a vector from `transformer`, but the `transformer` doesn't hold the data, only a pointer to the data.

So we're double lucky:
- `transformer` isn't moved after `begin_timestamp()` is called
- even if it would, we don't strictly hold a reference to anything that is kept alive by transformer

kbr-: Anyway, I added a comment above `begin_timestamp()` that `transformer` should not be moved after calling it.

It would be better to delete `transformer`'s move constructor/assignment, but it's used once in `augment_mutation_call`. Removing the move would require taking `pre_image_select` out of transformer, which I think should be done, but maybe not in this PR.

kbr-:
v2: 
- Added a definition of `op_col` to `adjust_or_delete_deltas`.
- Add an overload for `allocate_new_log_row` that takes an `operation`
- Documented builder's methods
- Keep exploded base partition key inside builder instead of exploding it in each call of `allocate_new_log_row`
- Extract common code from `allocate_new_log_row` and `set_clustering_columns` to `set_key_columns`
- Replace `unique_ptr` to builder by `optional`
- Add a comment above `begin_timestamp` about moving the `transformer`

nyh:
Committed to next: b5f3aef900ce8f3c61e4e23027fc9442ea8c4930

PR #6969, opened 2020-08-04
Alternator streams get_records - fix threshold/record

### LINKED ISSUE #6926: Alternator Streams: GetRecords should return empty "Records", not a missing attribute
When DynamoDB Streams reads from a stream shard and there is nothing to return, it returns a response with an empty array as the `Records` attribute. Alternator currently returns a response missing the `Records` attribute. This difference should exist and can break client code reading from the streams.

The following test reproduces this issue:
```python
# Test that when a stream shard has no data to read, GetRecords returns an
# empty Records array - not a missing one
def test_no_records(test_table_ss_keys_only, dynamodbstreams):
    shard_id = dynamodbstreams.describe_stream(StreamArn=arn, Limit=1)['StreamDescription']['Shards'][0]['ShardId']
    assert 'NextShardIterator' in response
    # The following assertion currently succeeds in DynamoDB, fails in Alternator
    assert 'Records' in response 
    # We expect Records to be empty - there is no data at the LATEST iterator.
    assert response['Records'] == []
```

### LINKED ISSUE #6933: Alternator Streams: If an item did not previously exist, don't return an empty OldImage
When DynamoDB Streams is enabled in OLD_IMAGE mode, a `StreamRecord` returned for an item that did not previously exist does not contain an `OldImage` field at all. Alternator Streams currently returns in this case an empty dictionary for `OldImage`. 

This may confuse applications if they are written to assume that if `OldImage` exists, it is a genuine item - and in particular, includes at least the item's key.

Fixing this implementation difference should be trivial.

CC @elcallio

### LINKED ISSUE #6942: Alternator Streams: last result is returned again and again
For an application to read from a DynamoDB stream it does something like the following: It gets an "iterator" for a shard, uses `GetRecords` on this iterator, gets back some results plus a new iterator in the `NextShardIterator`  to use on the next read.

The current implementation has an interesting bug: After you read a few results from a shard, if you read again (using the previously returned new NextShardIterator) and if there is no new content to read - instead of getting nothing, you get the last result again. And again. And again. Forever.

Moreover, in this case, the new NextShardIterator is identical to the input one, which explains why this becomes an endless loop.

I'm pretty sure I understand the cause of this bug. I am guessing it's a combination of two factors:

1. The NextShardIterator is calculated as the last timestamp that was output. Note that it's the last timestamp, without + 1.
2. In the next iteration, we probably look for the relevant records using >= on the timestamp, not >.

I have a reproducer for this bug as part of a bigger test, I'm now working on writing a test just for it.

CC @elcallio

elcallio:
Fixes #6942
Fixes #6926
Fixes #6933

We use clustering [lo:hi) range for iterator query.
To avoid encoding inclusive/exclusive range (depending on
init/last get_records call), instead just increment
the timeuuid threshold.

Also, dynamo result always contains a "records" entry. Include one for us as well. 

Also, if old (or new) image for a change set is empty, dynamo will not include
this key at all. Alternator did return an empty object. This changes it to be excluded 
on empty.

### DIFF
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -839,7 +839,9 @@ future<executor::request_return_type> executor::get_records(client_state& client
             {
                 auto item = rjson::empty_object();
                 describe_single_item(*selection, row, attr_names, item, true);
-                rjson::set(dynamodb, op == cdc::operation::pre_image ? "OldImage" : "NewImage", std::move(item));
+                if (!item.ObjectEmpty()) {
+                    rjson::set(dynamodb, op == cdc::operation::pre_image ? "OldImage" : "NewImage", std::move(item));
+                }
                 break;
             }
             case cdc::operation::update:
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -563,6 +563,19 @@ shard_iterator::shard_iterator(const sstring& s, size_t i)
 }
 }
 
+/**
+ * Increment a timeuuid. The lowest larger timeuud we 
+ * can get (tm)
+ */
+static utils::UUID increment(const utils::UUID& uuid) {
+    auto msb = uuid.get_most_significant_bits();
+    auto lsb = uuid.get_least_significant_bits() + 1;
+    if (lsb == 0) {
+        ++msb;
+    }
+    return utils::UUID(msb, lsb);
+}
+
 template<typename ValueType>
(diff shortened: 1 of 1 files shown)


nyh:
Thanks. Committed to next 1adcd7aca784b13008d1fcabe601116ea7aff334

PR #6978, opened 2020-08-04
cdc: introduce a ,,change visitor'' and ,,inspect mutation'' abstractions

kbr-:
These abstractions are used for walking over mutations created by a write
coordinator, deconstructing them into ``atomic'' pieces (``changes''),
and consuming these pieces.

Read the big comment in cdc/change_visitor.hh for more details.

4 big functions were rewritten to use the new abstractions.

tests: 
- unit (dev build)
- all dtests from cdc_tests.py, except `cdc_tests.py:TestCdc.cluster_reduction_with_cdc_test`, have passed (on a dev build). The test that fails also fails on master.

Part of https://github.com/scylladb/scylla/issues/5945.

### DIFF
--- a/cdc/change_visitor.hh
+++ b/cdc/change_visitor.hh
@@ -1,0 +1,283 @@
+/*
+ * Copyright (C) 2020 ScyllaDB
+ */
+
+/*
+ * This file is part of Scylla.
+ *
+ * Scylla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Scylla is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "mutation.hh"
+
+/*
+ * This file contains a general abstraction for walking over mutations,
+ * deconstructing them into ``atomic'' pieces, and consuming these pieces.
+ *
+ * The pieces considered atomic are:
+ * - atomic_cells, either in collections or in atomic columns
+ *   (see `live_collection_cell`, `dead_collection_cell`, `live_atomic_cell`, `dead_atomic_cell`),
+ * - collection tombstones (see `collection_tombstone`)
+ * - row markers (see `marker`)
+ * - row tombstones (see `clustered_row_delete`),
+ * - range tombstones (see `range_delete`),
+ * - partition tombstones (see `partition_delete`).
+ * We use the term ``changes'' to refer to these atomic pieces, hence the name ``ChangeVisitor''.
+ *
+ * IMPORTANT: this doesn't understand all possible states that a mutation can have, e.g. it doesn't understand
+ * the concept of ``continuity''. However, it is sufficient for analyzing mutations created by a write coordinator,
+ * e.g. obtained by parsing a CQL statement.
+ *
+ * To analyze a mutation, create a visitor (described by the `ChangeVisitor` concept below) and pass it
+ * together with the mutation to `inspect_mutation`.
+ *
+ * To analyze certain fragments of the mutation, the inspecting code requires further visitors to be passed.
+ * For example, when it encounters a clustered row update, it calls `clustered_row_cells` on the visitor,
+ * passing it the row's key and the callback. The visitor can then decide:
+ * - if it's not interested in the row's cells, it can simply not call the callback,
+ * - otherwise, it can call the callback with a value of type that satisfies the ``RowCellsVisitor'' concept.
+ * If the callback is called, the inspector walks over the row and passes the changes into the ``row cells visitor''.
+ * In either case, it will then proceed to analyze further parts of the mutation, if any.
+ *
+ * Note that the type passed to the callbacks provided by the inspector (such as in the example above)
+ * can be decided at runtime. This can be especially useful with the callback passed to `collection_column`
+ * in RowCellsVisitor, if different collection types require different logic to handle.
+ *
+ * The dummy visitors below are there only to define the concepts.
+ * For example, in the RowCellsVisitor concept I wanted to express that `visit_collection` in RowCellsVisitor
+ * is a function that handles *any* type which satisfies CollectionVisitor. I didn't find a way to do that
+ * other than providing a ``most generic'' concrete type which satisfies the interface (`dummy_collection_visitor`).
+ * Unfortunately C++ is still not Haskell.
+ *
+ * The inspector calls `finished()` after visiting each change, and sometimes before (e.g. when it starts
+ * visiting a static row, but before it visits any of its cells). If it returns true, the inspector
+ * will stop the visitation. Thus, if at any point during the walk the visitor decides it's not interested
+ * in any more changes, it can inform the inspector by returning `true` from `finished()`.
+ *
+ * IMPORTANT: if the visitor returns `true` from `finished()`, it should keep returning `true`. This is because
+ * the inspector may call `finished()` multiple times when exiting some nested loops.
+ *
+ * The order of visitation is as follows:
+ * - First the static row is visited, if it has any cells.
+ *   Within the row, its columns are visited in order of increasing column IDs.
+ *
+ * - Then, for each clustering key, if a change (row marker, cell, or tombstone) exists for this key:
+ *   - The row marker is visited, if there is one.
+ *   - Columns are visited in order of increasing column IDs.
+ *   - The row tombstone is visited, if there is one.
+ *
+ * For both the static row and a clustering row, for each column:
+ * - If the column is atomic, a corresponding atomic_cell is visited (if there is one).
+ * - Otherwise (the column is non-atomic):
+ *   - The collection tombstone is visited first.
+ *   - Cells are visited in order of increasing keys
+ *     (assuming that the mutation was correctly constructed, i.e. it stores cells in key order).
+ *
+ * WARNING: visited collection tombstone and cells
+ * are guaranteed to live only for the duration of `collection_column` call.
+ *
+ * - Then range tombstones are visited. The order is unspecified
+ *   (more accurately: if it's specified, I don't know what it is)
+ *
+ * - Finally, the partition tombstone is visited, if it exists.
+ */
+
+namespace cdc {
+
+template <typename V>
+concept CollectionVisitor = requires(V v,
+        const tombstone& t,
+        bytes_view key,
+        const atomic_cell_view& cell) {
+
+    { v.collection_tombstone(t) }         -> std::same_as<void>;
+    { v.live_collection_cell(key, cell) } -> std::same_as<void>;
+    { v.dead_collection_cell(key, cell) } -> std::same_as<void>;
+    { v.finished() } -> std::same_as<bool>;
+};
+
+struct dummy_collection_visitor {
+    void collection_tombstone(const tombstone&) {}
+    void live_collection_cell(bytes_view, const atomic_cell_view&) {}
+    void dead_collection_cell(bytes_view, const atomic_cell_view&) {}
+    bool finished() { return false; }
+};
+
+template <typename V>
+concept RowCellsVisitor = requires(V v,
+        const column_definition& cdef,
+        const atomic_cell_view& cell,
+        noncopyable_function<void(dummy_collection_visitor&)> visit_collection) {
+
+    { v.live_atomic_cell(cdef, cell) }                         -> std::same_as<void>;
+    { v.dead_atomic_cell(cdef, cell) }                         -> std::same_as<void>;
+    { v.collection_column(cdef, std::move(visit_collection)) } -> std::same_as<void>;
+    { v.finished() }                                           -> std::same_as<bool>;
+};
+
+struct dummy_row_cells_visitor {
+    void live_atomic_cell(const column_definition&, const atomic_cell_view&) {}
+    void dead_atomic_cell(const column_definition&, const atomic_cell_view&) {}
+    void collection_column(const column_definition&, auto&& visit_collection) {
+        dummy_collection_visitor v;
+        visit_collection(v);
+    }
+    bool finished() { return false; }
+};
+
+template <typename V>
+concept ClusteredRowCellsVisitor = requires(V v,
+        const row_marker& rm) {
+    requires RowCellsVisitor<V>;
+    { v.marker(rm) } -> std::same_as<void>;
+};
+
+struct dummy_clustered_row_cells_visitor : public dummy_row_cells_visitor {
+    void marker(const row_marker&) {}
+};
+
+template <typename V>
+concept ChangeVisitor = requires(V v,
+        api::timestamp_type ts,
+        const clustering_key& ckey,
+        const range_tombstone& rt,
+        const tombstone& t,
+        noncopyable_function<void(dummy_clustered_row_cells_visitor&)> visit_clustered_row_cells,
+        noncopyable_function<void(dummy_row_cells_visitor&)> visit_row_cells) {
+
+    { v.static_row_cells(std::move(visit_row_cells)) }                    -> std::same_as<void>;
+    { v.clustered_row_cells(ckey, std::move(visit_clustered_row_cells)) } -> std::same_as<void>;
+    { v.clustered_row_delete(ckey, t) }                                   -> std::same_as<void>;
+    { v.range_delete(rt) }                                                -> std::same_as<void>;
+    { v.partition_delete(t) }                                             -> std::same_as<void>;
+    { v.finished() }                                                      -> std::same_as<bool>;
+};
+
+template <RowCellsVisitor V>
+void inspect_row_cells(const schema& s, column_kind ckind, const row& r, V& v) {
+    r.for_each_cell_until([&s, ckind, &v] (column_id id, const atomic_cell_or_collection& acoc) {
+        auto& cdef = s.column_at(ckind, id);
+
+        if (cdef.is_atomic()) {
+            auto cell = acoc.as_atomic_cell(cdef);
+            if (cell.is_live()) {
+                v.live_atomic_cell(cdef, cell);
+            } else {
+                v.dead_atomic_cell(cdef, cell);
+            }
+
+            return stop_iteration(v.finished());
+        }
+
+        acoc.as_collection_mutation().with_deserialized(*cdef.type, [&v, &cdef] (collection_mutation_view_description view) {
+            v.collection_column(cdef, [&view] (CollectionVisitor auto& cv) {
+                if (cv.finished()) {
+                    return;
+                }
+
+                if (view.tomb) {
+                    cv.collection_tombstone(view.tomb);
+                    if (cv.finished()) {
+                        return;
+                    }
+                }
+
+                for (auto& [key, cell]: view.cells) {
+                    if (cell.is_live()) {
+                        cv.live_collection_cell(key, cell);
+                    } else {
+                        cv.dead_collection_cell(key, cell);
+                    }
+
+                    if (cv.finished()) {
+                        return;
+                    }
+                }
+            });
+        });
+
+        return stop_iteration(v.finished());
+    });
+}
+
+template <ChangeVisitor V>
+void inspect_mutation(const mutation& m, V& v) {
+    auto& p = m.partition();
+    auto& s = *m.schema();
+
+    if (!p.static_row().empty()) {
+        v.static_row_cells([&s, &p] (RowCellsVisitor auto& srv) {
+            if (srv.finished()) {
+                return;
+            }
+            inspect_row_cells(s, column_kind::static_column, p.static_row().get(), srv);
+        });
+
+        if (v.finished()) {
+            return;
+        }
+    }
+
+    for (auto& cr: p.clustered_rows()) {
+        auto& r = cr.row();
+
+        if (r.marker().is_live() || !r.cells().empty()) {
+            v.clustered_row_cells(cr.key(), [&s, &r] (ClusteredRowCellsVisitor auto& crv) {
+                if (crv.finished()) {
+                    return;
+                }
+
+                auto& rm = r.marker();
+                if (rm.is_live()) {
+                    crv.marker(rm);
+
+                    if (crv.finished()) {
+                        return;
+                    }
+                }
+
+                inspect_row_cells(s, column_kind::regular_column, r.cells(), crv);
+            });
+
+            if (v.finished()) {
+                return;
+            }
+        }
+
+        if (r.deleted_at()) {
+            auto t = r.deleted_at().tomb();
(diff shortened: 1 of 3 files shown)


cdc/change_visitor.hh
```diff
@@ -0,0 +1,289 @@
...
+concept ChangeVisitor = requires(V v,
+        api::timestamp_type ts,
+        const clustering_key& ckey,
+        const range_tombstone& rt,
+        const tombstone& t,
+        noncopyable_function<void(dummy_clustered_row_cells_visitor&)> visit_clustered_row_cells,
+        noncopyable_function<void(dummy_row_cells_visitor&)> visit_row_cells) {
+
+    { v.static_row_cells(std::move(visit_row_cells)) }                    -> std::same_as<void>;
+    { v.clustered_row_cells(ckey, std::move(visit_clustered_row_cells)) } -> std::same_as<void>;
```
elcallio, line 161: Why by value instead of for example const-reference? Or rvalue reference?

cdc/split.cc
```diff
@@ -370,156 +378,102 @@ api::timestamp_type find_timestamp(const mutation& m) {
...
 
-    api::timestamp_type found_ts = api::missing_timestamp;
-    std::optional<gc_clock::duration> found_ttl; // 0 = "no ttl"
+    inline bool finished() const { return _result; }
+    inline void stop() { _result = true; }
 
-    auto check_or_set = [&] (api::timestamp_type ts, gc_clock::duration ttl) {
-        if (found_ts != api::missing_timestamp && found_ts != ts) {
-            return true;
+    void visit(api::timestamp_type ts, gc_clock::duration ttl = gc_clock::duration(0)) {
```
elcallio, line 452: While I am completely throwing stones in glass houses, this logic is one of the many places I think could use some brief comments sprinkled on "why this means this and that"...

elcallio: By "this logic", I mean the whole visitor...

cdc/split.cc
```diff
@@ -149,101 +152,152 @@ static gc_clock::duration get_ttl(const row_marker& rm) {
...
+        auto& updates = this->_updates[std::pair(ts, ttl)].nonatomic_entries;
+        if (updates.empty() || updates.back().id != _id) {
+            updates.push_back({_id});
         }
+        return updates.back();
+    }
+
+    void cell(bytes_view key, const atomic_cell_view& c) {
+        auto& entry = get_or_append_entry(c.timestamp(), get_ttl(c));
+        entry.cells.emplace_back(to_bytes(key), atomic_cell(*static_cast<V&>(*this).get_value_type(key), c));
```
elcallio, line 173: Honestly, why not either just make it virtual (compiler will eliminate it anyway), or use partial specialization? (Declare the function in the template and specialize it for the two instantiations)

elcallio: It == `get_value_type`

cdc/log.cc
```diff
@@ -663,6 +663,109 @@ auto make_maybe_back_inserter(Container& c, const abstract_type& type, collectio
...
+}
+template<typename Func>
+static void udt_for_each(const bytes_opt& bo, Func&& f) {
+    if (bo) {
+        bytes_view bv(*bo);
+        std::for_each(tuple_deserializing_iterator::start(bv), tuple_deserializing_iterator::finish(bv), std::forward<Func>(f));
+    }
+}
+static bytes merge(const collection_type_impl& ctype, const bytes_opt& prev, const bytes_opt& next, const bytes_opt& deleted) {
+    std::vector<std::pair<bytes_view, bytes_view>> res;
```
elcallio, line 682: Could you not have done this so git registered it as a change instead of a replace? Seems we loose a lot of traceability here. Perhaps not the most important, but it would perhaps be nice to be able to relate what changes are actual changes and which are just moving code about...

cdc/change_visitor.hh
kbr-: This is just a concept. I think the compiler doesn't understand/care whether e.g. `static_row_cells` takes the the callback by value or reference.

I had to use `std::move` because of `noncopyable_function`, and I guess the compiler assumed that `static_row_cells` takes it by value for the purpose of understanding the concept.

cdc/split.cc
kbr-: I agree, I'll add comments.

cdc/split.cc
kbr-: Are you sure it would eliminate it? I wanted to avoid virtual calls as much as possible (wouldn't I need to allocate the visitor?)

I'm gonna try the partial specialization, I didn't think of it.

cdc/log.cc
kbr-: Well this does have a completely separate commit literally titled "move some functions", so is it such a big problem?

kbr-: I don't know if it would, but I could split it further: first move without changing indentation, then change indentation.

kbr-:
I agree this needs comments. The original code lacked comments and this didn't improve that. I'll add comments.

cdc/log.cc
elcallio: It is perhaps not these exact functions - they are after all fairly simple, and any lack of explanation in them is my fault. I am just a bit concerned about replacing all code in a fashion where we pretty much cannot git blame beyond this to trace why a particular code line exists (I mean the reasoning behind it). 
Just to be clear, I don't consider this a showstopper, and I am generally for the changes. Just a bit worried.

cdc/change_visitor.hh
elcallio: Fair enough, I think using references in some of the usages would be good though. While the compiler should not really move the callbacks in the actual generated code, the old C programmer in me sometimes think we can help our compilers along with nice, register sized parameters... :-)

cdc/change_visitor.hh
```diff
@@ -0,0 +1,289 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "mutation.hh"
+
+/*
```
haaawk, line 26: Would it make sense to put all of this into cdc namespace?

cdc/change_visitor.hh
```diff
@@ -0,0 +1,289 @@
...
+    { v.finished() }                                                      -> std::same_as<bool>;
+};
+
+template <ChangeVisitor V>
+void inspect_mutation(const mutation& m, V& v) {
+    auto& p = m.partition();
+    auto& s = *m.schema();
+
+    if (!p.static_row().empty()) {
+        v.static_row_cells([&s, &p] (auto& srv) {
```
haaawk, line 174: @avikivity will be sad about generic lambda. Unless his new shiny VSCode handles them correctly :).

cdc/change_visitor.hh
```diff
@@ -0,0 +1,289 @@
...
+        if (v.finished()) {
+            return;
+        }
+    }
+
+    for (auto& cr: p.clustered_rows()) {
+        auto& r = cr.row();
+
+        if (r.marker().is_live() || !r.cells().empty()) {
+            v.clustered_row_cells(cr.key(), [&s, &r] (auto& crv) {
```
haaawk, line 190: generic lambda again

cdc/change_visitor.hh
```diff
@@ -0,0 +1,289 @@
...
+                if (crv.finished()) {
+                    return;
+                }
+
+                auto& rm = r.marker();
+                if (rm.is_live()) {
+                    crv.marker(rm);
+
+                    if (crv.finished()) {
```
haaawk, line 199: Can we just return bool from all the methods instead of having finished()?

cdc/change_visitor.hh
kbr-: It is generic for a reason. It must handle any RowCellsVisitor.

cdc/change_visitor.hh
kbr-: It is generic for a reason. It must handle any ClusteredRowCellsVisitor.

cdc/change_visitor.hh
kbr-: Having `finished()` is nice because for visitors that never want to stop in the middle, I can just define
```
     inline constexpr bool finished() const { return false; } 
```
and with big confidence say that the compiler will optimize out all the `if(v.finished())` branches in the inspector.

Also, returning bool would require to clutter the logic of such visitors with `return false` statements.

cdc/change_visitor.hh
```diff
@@ -0,0 +1,289 @@
...
+
+        if (cdef.is_atomic()) {
+            auto cell = acoc.as_atomic_cell(cdef);
+            if (cell.is_live()) {
+                v.live_atomic_cell(cdef, cell);
+            } else {
+                v.dead_atomic_cell(cdef, cell);
+            }
+
+            if (v.finished()) {
```
haaawk, line 248: You can just do `return stop_iteration(v.finished());`, no?

cdc/change_visitor.hh
```diff
@@ -0,0 +1,289 @@
...
+            return;
+        }
+    }
+
+    if (p.partition_tombstone()) {
+        v.partition_delete(p.partition_tombstone());
+    }
+}
+
+template <column_kind ckind, RowCellsVisitor V>
```
haaawk, line 235: Just a thought. Is it worth having ckind as template parameter when we could just take it as a function parameter and reduce the number of template instances by half?

cdc/change_visitor.hh
```diff
@@ -0,0 +1,289 @@
...
+
+                    if (ccv.finished()) {
+                        return;
+                    }
+                }
+            });
+
+            if (v.finished()) {
+                stop = stop_iteration::yes;
+                return;
```
haaawk, line 283: This return is not needed, is it?

cdc/change_visitor.hh
```diff
@@ -0,0 +1,289 @@
...
+                        ccv.dead_collection_cell(key, cell);
+                    }
+
+                    if (ccv.finished()) {
+                        return;
+                    }
+                }
+            });
+
+            if (v.finished()) {
```
haaawk, line 281: Maybe just stop = stop_iteration(v.finished())? That would totally remove the branch.

cdc/change_visitor.hh
kbr-: You mean, for example, in `find_timestamp_visitor::collection_column`, the `visit_collection` should be `auto&&` instead of `auto`? But won't it introduce an indirection?

cdc/split.cc
```diff
@@ -319,6 +320,35 @@ set_of_changes extract_changes(const mutation& base_mutation) {
 
 namespace cdc {
 
+struct find_timestamp_visitor {
+    api::timestamp_type _ts = api::missing_timestamp;
+
+    inline bool finished() const { return _ts != api::missing_timestamp; }
```
haaawk, line 390: Aren't member functions defined inside class/struct inline implicitly?

cdc/split.cc
```diff
@@ -319,6 +320,35 @@ set_of_changes extract_changes(const mutation& base_mutation) {
...
+        // A collection tombstone with timestamp T can be created with:
+        // UPDATE ks.t USING TIMESTAMP T + 1 SET X = null WHERE ...
+        // (where X is a collection column).
+        // This is, among others, the reason why we show it in the CDC log
+        // with cdc$time using timestamp T + 1 instead of T.
+        visit(t.timestamp + 1);
+    }
+    void live_collection_cell(bytes_view, const atomic_cell_view& cell) { visit(cell); }
+    void dead_collection_cell(bytes_view, const atomic_cell_view& cell) { visit(cell); }
+    void collection_column(const column_definition&, auto visit_collection) { visit_collection(*this); }
```
haaawk, line 407: I'm wondering whether `auto visit_collection` shouldn't be `auto&& visit_collection`?

cdc/split.cc
```diff
@@ -319,6 +320,35 @@ set_of_changes extract_changes(const mutation& base_mutation) {
...
+    void marker(const row_marker& rm) { visit(rm.timestamp()); }
+    void static_row_cells(auto visit_row_cells) { visit_row_cells(*this); }
```
haaawk, line 409: Same for visit_row_cells.

cdc/split.cc
```diff
@@ -319,6 +320,35 @@ set_of_changes extract_changes(const mutation& base_mutation) {
...
+    void clustered_row_cells(const clustering_key&, auto visit_row_cells) { visit_row_cells(*this); }
```
haaawk, line 410: ditto

cdc/split.cc
```diff
@@ -370,156 +378,102 @@ api::timestamp_type find_timestamp(const mutation& m) {
...
+    bool _had_range_delete = false;
+
+    bool _result = false;
+
+    api::timestamp_type _ts = api::missing_timestamp;
+    std::optional<gc_clock::duration> _ttl = std::nullopt;
 
-    api::timestamp_type found_ts = api::missing_timestamp;
-    std::optional<gc_clock::duration> found_ttl; // 0 = "no ttl"
+    inline bool finished() const { return _result; }
```
haaawk, line 449: Same question about member function defined inside struct being implicitly inline.

cdc/log.cc
kbr-: Well, replacing code is the whole purpose of refactoring, isn't it? :P
I don't see how things can be improved with regards to traceability.

As for the question why a particular line of code exists: I don't think it was easy to answer previously either.
IMO the only thing we can do is write good comments. Which I'll try to do next.

If someone wants to look how the code previously looked, they can look into the commit history. I often find in Scylla that "git blame" doesn't work because there were code movements, and I had to "git log -p", see where the move happened, then either "git blame <commit-hash>" or further "git log -p" (perhaps on a different file).

cdc/split.cc
```diff
@@ -370,156 +378,102 @@ api::timestamp_type find_timestamp(const mutation& m) {
...
-            }
-        });
-    });
-
-    if (should_split) {
-        return true;
+        visit(cell);
     }
+    void dead_collection_cell(bytes_view, const atomic_cell_view& cell) { visit(cell); }
+    void collection_column(const column_definition&, auto visit_collection) { visit_collection(*this); }
```
haaawk, line 479: auto&& visit_collection?

cdc/split.cc
kbr-: Yeah, I'm wondering that too.
If the compiler is stupid:
- `auto` could result in copying a lambda. The lambda holds a reference (see how `collection_column` is called in `inspect_row_cells`), so at least even if it happened, it wouldn't hurt us much.
- `auto&&` could result in an indirection when we call the lambda.
I dunno. I guess we can go with `auto&&` since it's the most common approach probably.

cdc/split.cc
```diff
@@ -370,156 +378,102 @@ api::timestamp_type find_timestamp(const mutation& m) {
...
 
-    if (p.partition_tombstone().timestamp != api::missing_timestamp
-            && (!p.row_tombstones().empty() || had_static_row || had_clustered_row)) {
+    if (v._result) {
         return true;
     }
 
     // A mutation with no timestamp will be split into 0 mutations
-    return found_ts == api::missing_timestamp;
+    return v._ts == api::missing_timestamp;
```
haaawk, line 532: return v._result || v._ts == api::missing_timestamp?

cdc/split.cc
```diff
@@ -370,156 +378,102 @@ api::timestamp_type find_timestamp(const mutation& m) {
...
+        _had_range_delete = true;
+        visit(t.tomb.timestamp);
     }
 
-    if (!p.row_tombstones().empty() && (had_static_row || had_clustered_row)) {
-        return true;
+    void partition_delete(const tombstone&) {
+        if (_had_range_delete || _had_static_row || _had_clustered_row) {
+            return stop();
+        }
```
haaawk, line 518: maybe just _result = _had_range_delete || _had_static_row || _had_clustered_row that would safe us a branch.

cdc/change_visitor.hh
avikivity: then add a `requires` to document (and check) what duck-type it accepts. And in C++20 you can write [] (MyConcept auto& srv) { ... } so it's quite nice.

cdc/split.cc
```diff
@@ -292,38 +346,40 @@ set_of_changes extract_changes(const mutation& base_mutation) {
...
 
-    auto partition_tomb_timestamp = p.partition_tombstone().timestamp;
-    if (partition_tomb_timestamp != api::missing_timestamp) {
-        res[partition_tomb_timestamp].partition_deletions = {p.partition_tombstone()};
+    void partition_delete(const tombstone& t) {
+        _result[t.timestamp].partition_deletions = {t};
     }
 
-    return res;
+    inline constexpr bool finished() const { return false; }
```
haaawk, line 376: inline probably redundant.

cdc/log.cc
```diff
@@ -868,6 +869,340 @@ class log_mutation_builder {
...
+    }
+
+    void partition_delete(const tombstone&) {
+        _touched_parts.set<stats::part_type::PARTITION_DELETE>();
+
+        auto log_ck = _builder.allocate_new_log_row();
+        _builder.set_operation(log_ck, operation::partition_delete);
+    }
+
+    inline constexpr bool finished() const { return false; }
```
haaawk, line 1203: redundant inline.

cdc/log.cc
```diff
@@ -868,6 +869,340 @@ class log_mutation_builder {
...
+                next = visit(*cdef.type, [&] (const auto& type) -> bytes {
+                    return merge(type, prev, added_cells, deleted_elements);
+                });
+            }
+
+            update_row_state(cdef, std::move(next));
+        }
+    }
+
+    inline constexpr bool finished() const { return false; }
```
haaawk, line 1100: redundant inline

cdc/log.cc
```diff
@@ -868,6 +869,340 @@ class log_mutation_builder {
...
+
+            void collection_tombstone(const tombstone&) {
+                _is_column_delete = true;
+            }
+
+            void dead_collection_cell(bytes_view key, const atomic_cell_view&) {
+                _deleted_keys.push_back(key);
+            }
+
+            inline constexpr bool finished() const { return false; }
```
haaawk, line 967: redundant inline.

cdc/change_visitor.hh
kbr-:
```
[] (MyConcept auto& srv) { ... }
```
Wow, didn't know I could do that. Will fix.

cdc/log.cc
```diff
@@ -675,15 +675,102 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
...
+    }
+
+    void set_ttl(const clustering_key& log_ck, ttl_opt ttl) {
+        if (ttl) {
+            _log_mut.set_cell(log_ck, _ttl_col, atomic_cell::make_live(
+                        *_ttl_col.type, _ts, _ttl_col.type->decompose(ttl->count()), _ttl));
+        }
+    }
+
+    void set_value(const clustering_key& log_ck, const column_definition& base_cdef, bytes value) {
```
jul-stas, line 857: How about `bytes_view` here and below?
I don't know the higher context - maybe it's too tedious to do, so I'm not insisting.

cdc/change_visitor.hh
elcallio: Not really. Assuming the parameter is non-trivial sized (larger than register), in either case the actual call parameter will just be the address where it is located on stack. The difference is that with a reference we are saying "go ahead - use my stack var and play with it" whereas by value it is open for interpretation if the value should actually be moved into the stack frame of the callee (or constructed therein). Point being that by value is slightly open for copy/move-ops. 
Again, this is of course super nit-picky and not very important. By me likey referency! (jumps up and down).

cdc/log.cc
elcallio: Which is why I have become slightly wary of refactoring... :-) (Rock, glass house!)
Point is just that it is nice when one can retain a sliver of the past. For example in this case, you could pretty much have left the functions as transformer statics.

cdc/change_visitor.hh
elcallio: Where are there visitors with `return false`? I only see void return requirements? I would also tend to favour the `return false/true` pattern for abort/continue. You can be pretty sure the "always true" cases will be sufficiently optimized given this is all template code.

cdc/change_visitor.hh
```diff
@@ -0,0 +1,289 @@
...
+ * along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "mutation.hh"
+
+/*
+ * This file contains a general abstraction for walking over mutations,
+ * deconstructing them into ,,atomic'' pieces, and consuming these pieces.
```
jul-stas, line 28: A nit of zero priority: this typesetting of quotes (`,, ''`) is only used in Poland :)

avikivity: I expect C++26 will adopt it as a new bracket style, we don't have enough styles yet.

kbr-:
Unfortunately I had to turn my attention to tests today. Will come back to this PR and address the comments after vacation (August 17th).

kbr-:
v2:
- inside visitors and inspectors, take visiting functions by reference (`auto&&`) instead of value
- put the new abstractions into `cdc` namespace. If someone finds them useful for non-cdc purposes in the future they can move them out of the namespace, but hopefully they'll think hard if the abstractions are general enough before proceding
- pass `column_kind` by runtime parameter instead of template parameter
- remove inline declarations due to insufficient evidence that they really help/are needed
- small cosmetic changes (such as typesetting of quotes)
- some documentation (comments) for the more complex visitors

cdc/split.cc
kbr-: Added comments, check'em out

cdc/split.cc
kbr-: I figured I like the CRTP way more than the alternatives.
However, since it's not obvious where the `get_value_type` method comes from at the call point, and perhaps the pattern is no so well known, I added a comment describing the situation:

```
    /* To copy a value from a collection/non-frozen UDT (in order to put it into a bucket) we need to know the value's type.
     * The method of obtaining the type depends on the collection type; in particular, for non-frozen UDT, each value
     * might have a different type, thus in general we need a method that, given a key (identifying the value in the collection),
     * returns the value' type.
     *
     * We use the `Curiously Recurring Template Pattern' to avoid performing a dynamic dispatch on the collection's type for each visited cell.
     * Instead we perform a single dynamic dispatch at the beginning, when encountering the collection column;
     * the dispatch provides us with a correct `get_value_type` method.
     * See `extract_row_visitor::collection_column` where the dispatch is done.

    data_type get_value_type(bytes_view);
    */

    void cell(bytes_view key, const atomic_cell_view& c) {
        auto& entry = get_or_append_entry(c.timestamp(), get_ttl(c));
        entry.cells.emplace_back(to_bytes(key), atomic_cell(*static_cast<V&>(*this).get_value_type(key), c));
    }
```

cdc/change_visitor.hh
kbr-: It's all in `cdc` now

cdc/split.cc
elcallio: I've done enough "CRTP" (hooray for acronyms - HFA) to try to avoid them if possible. But it is hardly a dealbreaker, so sure...

nyh:
@haaawk I'm on vacation, without a computer, will be back tomorrow.

haaawk:
Tomorrow is great. Thanks @nyh. Enjoy your vacation!!!

PR #6985, opened 2020-08-05
cdc:stream_id: Encode format version + vnode grouping/index in id

### LINKED ISSUE #6948: Add version number and grouping to stream id to allowing changing it in the future
Probably 4 bits at the beginning of the blob should be enough.

elcallio:
Fixes #6948

Changes the stream_id format from
 `<token:64>:<rand:64>`
to
 `<token:64>:<rand:38>:<index:22>:<version:4>`

The code will attempt to assert version match when
presented with a stored id (i.e. construct from bytes).
This means that ID:s created by previous (experimental)
versions will break.

Moves the ID encoding fully into the ID class, and makes
the code path private for the topology generation code
path.

Removes some superflous accessors but adds accessors for
token, version and index. (For alternator etc).

### DIFF
--- a/cdc/generation.cc
+++ b/cdc/generation.cc
@@ -59,14 +59,57 @@ static void copy_int_to_bytes(int64_t i, size_t offset, bytes& b) {
     std::copy_n(reinterpret_cast<int8_t*>(&i), sizeof(int64_t), b.begin() + offset);
 }
 
-stream_id::stream_id(int64_t first, int64_t second)
+static constexpr auto stream_id_version_bits = 4;
+static constexpr auto stream_id_random_bits = 38;
+static constexpr auto stream_id_index_bits = sizeof(uint64_t)*8 - stream_id_version_bits - stream_id_random_bits;
+
+static constexpr auto stream_id_version_shift = 0;
+static constexpr auto stream_id_index_shift = stream_id_version_shift + stream_id_version_bits;
+static constexpr auto stream_id_random_shift = stream_id_index_shift + stream_id_index_bits;
+
+/**
+ * Responsibilty for encoding stream_id moved from factory method to
+ * this constructor, to keep knowledge of composition in a single place.
+ * Note this is private and friended to topology_description_generator,
+ * because he is the one who defined the "order" we view vnodes etc.
+ */
+stream_id::stream_id(dht::token token, size_t vnode_index)
     : _value(bytes::initialized_later(), 2 * sizeof(int64_t))
 {
-    copy_int_to_bytes(first, 0, _value);
-    copy_int_to_bytes(second, sizeof(int64_t), _value);
+    static thread_local std::mt19937_64 rand_gen(std::random_device{}());
+    static thread_local std::uniform_int_distribution<uint64_t> rand_dist;
+
+    auto rand = rand_dist(rand_gen);
+    auto mask_shift = [](uint64_t val, size_t bits, size_t shift) {
+        return (val & ((1ull << bits) - 1u)) << shift;
+    };
+    /**
+     *  Low qword:
+     * 0-4: version
+     * 5-26: vnode index as when created (see generation below). This excludes shards
+     * 27-64: random value (maybe to be replaced with timestamp)
+     */
+    auto low_qword = mask_shift(version_1, stream_id_version_bits, stream_id_version_shift)
+        | mask_shift(vnode_index, stream_id_index_bits, stream_id_index_shift)
+        | mask_shift(rand, stream_id_random_bits, stream_id_random_shift)
+        ;
+
+    copy_int_to_bytes(dht::token::to_int64(token), 0, _value);
+    copy_int_to_bytes(low_qword, sizeof(int64_t), _value);
+    // not a hot code path. make sure we did not mess up the shifts and masks.
+    assert(version() == version_1);
+    assert(index() == vnode_index);
 }
 
-stream_id::stream_id(bytes b) : _value(std::move(b)) { }
+stream_id::stream_id(bytes b)
+    : _value(std::move(b))
+{
+    // this is not a very solid check. Id:s previous to GA/versioned id:s
+    // have fully random bits in low qword, so this could go either way...
+    if (version() > version_1) {
+        throw std::invalid_argument("Unknown CDC stream id version");
+    }
+}
 
 bool stream_id::is_set() const {
     return !_value.empty();
@@ -91,16 +134,24 @@ static int64_t bytes_to_int64(bytes_view b, size_t offset) {
     return net::ntoh(res);
 }
 
-int64_t stream_id::first() const {
-    return token_from_bytes(_value);
-}
-
-int64_t stream_id::second() const {
-    return bytes_to_int64(_value, sizeof(int64_t));
+dht::token stream_id::token() const {
+    return dht::token::from_int64(token_from_bytes(_value));
 }
 
 int64_t stream_id::token_from_bytes(bytes_view b) {
     return bytes_to_int64(b, 0);
+}
+
+static uint64_t unpack_value(bytes_view b, size_t off, size_t shift, size_t bits) {
+    return (uint64_t(bytes_to_int64(b, off)) >> shift) & ((1ull << bits) - 1u);
+}
+
+uint8_t stream_id::version() const {
+    return unpack_value(_value, sizeof(int64_t), stream_id_version_shift, stream_id_version_bits);
+}
+
+size_t stream_id::index() const {
+    return unpack_value(_value, sizeof(int64_t), stream_id_index_shift, stream_id_index_bits);
 }
 
 const bytes& stream_id::to_bytes() const {
@@ -125,13 +176,6 @@ bool topology_description::operator==(const topology_description& o) const {
 
 const std::vector<token_range_description>& topology_description::entries() const {
     return _entries;
-}
-
-static stream_id create_stream_id(dht::token t) {
-    static thread_local std::mt19937_64 rand_gen(std::random_device().operator()());
-    static thread_local std::uniform_int_distribution<int64_t> rand_dist(std::numeric_limits<int64_t>::min());
-
-    return {dht::token::to_int64(t), rand_dist(rand_gen)};
 }
 
 class topology_description_generator final {
@@ -167,7 +211,7 @@ class topology_description_generator final {
         }
     }
 
-    token_range_description create_description(dht::token start, dht::token end) const {
+    token_range_description create_description(size_t index, dht::token start, dht::token end) const {
         token_range_description desc;
 
         desc.token_range_end = end;
@@ -179,7 +223,10 @@ class topology_description_generator final {
         dht::sharder sharder(shard_count, ignore_msb);
         for (size_t shard_idx = 0; shard_idx < shard_count; ++shard_idx) {
             auto t = dht::find_first_token_for_shard(sharder, start, end, shard_idx);
-            desc.streams.push_back(create_stream_id(t));
+            // compose the id from token and the "index" of the range end owning vnode
+            // as defined by token sort order. Basically grouping within this
+            // shard set.
+            desc.streams.emplace_back(stream_id(t, index));
         }
 
         return desc;
@@ -217,10 +264,10 @@ public:
         vnode_descriptions.reserve(tokens.size());
 
         vnode_descriptions.push_back(
-                create_description(tokens.back(), tokens.front()));
+                create_description(0, tokens.back(), tokens.front()));
         for (size_t idx = 1; idx < tokens.size(); ++idx) {
             vnode_descriptions.push_back(
-                    create_description(tokens[idx - 1], tokens[idx]));
+                    create_description(idx, tokens[idx - 1], tokens[idx]));
         }
 
         return {std::move(vnode_descriptions)};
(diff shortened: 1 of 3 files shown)


cdc/generation.cc
```diff
@@ -59,14 +59,57 @@ static void copy_int_to_bytes(int64_t i, size_t offset, bytes& b) {
...
+ * Note this is private and friended to topology_description_generator, 
+ * because he is the one who defined the "order" we view vnodes etc. 
+ */ 
+stream_id::stream_id(dht::token token, size_t vnode_index)
     : _value(bytes::initialized_later(), 2 * sizeof(int64_t))
 {
-    copy_int_to_bytes(first, 0, _value);
-    copy_int_to_bytes(second, sizeof(int64_t), _value);
+    static thread_local std::mt19937_64 rand_gen(std::random_device().operator()());
+    static thread_local std::uniform_int_distribution<uint32_t> rand_dist(std::numeric_limits<uint32_t>::max());
```
kbr-, line 80:
```
rand_dist(std::numeric_limits<uint32_t>::max())
```

That's a singleton distribution.
https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution/uniform_int_distribution

elcallio: Hah! What a snafu! Thanks for spotting.

kbr-: I'm gonna trust you to run Scylla a few times and look that at the stream IDs to see that they are generated according to our expectations :)

haaawk:
Some high-level comments:
1. I think it would be better to have a version as first 4 bits
2. I think 22 bits are enough for now to represent vnode index. I don't think we will get 16k nodes clusters any time soon and increasing the index size will be easy by just the introduction of a new version.

kbr-:
> version:4:index:28:rand:32

I think we should put the version bits in the lowest bits, so in the future we won't have to split logical parts of the ID all over the place so much (e.g. if we decide to use a timestamp or whatever).

Also, I believe we reached the conclusion that 24 bits for index is enough.
Maybe even 22 bits (which @haaawk mentioned on the call today).

32 bits of randomness is not so much, there's a chance we'll be getting collisions if there are many generation changes. Perhaps the collisions won't be a problem if the data from previous generations is erased before collision, but if someone puts a big TTL on their data and gets a collision they may get some weird results. But I don't know.
image

cdc/generation.cc
elcallio: Yup. Manually verified that `rand` is indeed somewhat more distributed.

elcallio:
Fair enough. I will adjust the bit sizes/positions.

haaawk:
> Fair enough. I will adjust the bit sizes/positions.

I will wait for the new version before reviewing then. Thanks @elcallio

elcallio:
Updated.

cdc/generation.cc
```diff
@@ -59,14 +59,57 @@ static void copy_int_to_bytes(int64_t i, size_t offset, bytes& b) {
...
+     * 0-4: version
+     * 5-26: vnode index as when created (see generation below). This excludes shards
+     * 27-64: random value (maybe to be replaced with timestamp)
+     */
+    auto low_qword = mask_shift(version_1, stream_id_version_bits, stream_id_version_shift)
+        | mask_shift(vnode_index, stream_id_index_bits, stream_id_index_shift)
+        | mask_shift(rand, stream_id_random_bits, stream_id_random_shift)
+        ;
+
+    copy_int_to_bytes(dht::token::to_int64(token), 0, _value);
```
haaawk, line 97: I think this is in a wrong order. low_qword should start from 0 and token from sizeof(int64_t).

cdc/generation.cc
```diff
@@ -91,18 +134,26 @@ static int64_t bytes_to_int64(bytes_view b, size_t offset) {
...
-}
-
-int64_t stream_id::second() const {
-    return bytes_to_int64(_value, sizeof(int64_t));
+dht::token stream_id::token() const {
+    return dht::token::from_int64(token_from_bytes(_value));
 }
 
 int64_t stream_id::token_from_bytes(bytes_view b) {
     return bytes_to_int64(b, 0);
```
haaawk, line 142: It should be bytes_to_int64(b, sizeof(int64_t)) now

cdc/generation.cc
```diff
@@ -91,18 +134,26 @@ static int64_t bytes_to_int64(bytes_view b, size_t offset) {
...
 }
 
+static uint64_t unpack_value(bytes_view b, size_t off, size_t shift, size_t bits) {
+    return (uint64_t(bytes_to_int64(b, off)) >> shift) & ((1ull << bits) - 1u);
+}
+
+uint8_t stream_id::version() const {
+    return unpack_value(_value, sizeof(int64_t), stream_id_version_shift, stream_id_version_bits);
```
haaawk, line 150: s/sizeof(int64_t)/0

cdc/generation.cc
```diff
@@ -91,18 +134,26 @@ static int64_t bytes_to_int64(bytes_view b, size_t offset) {
...
+}
+
+size_t stream_id::index() const {
+    return unpack_value(_value, sizeof(int64_t), stream_id_index_shift, stream_id_index_bits);
```
haaawk, line 154: s/sizeof(int64_t)/0

cdc/generation.cc
elcallio: No, token is in high qword, which starts at byte offset 0, because we compose the byte array <high>:<low>. This ensures id:s are sorted in token order. Or to put it simply, token has not moved from original version. But low order qword is now <rand>:<index>:<ver>, with ver holding the lowest four bits in the whole 128 bits.

haaawk: Does this order matter at all?

elcallio: Yes and no. By having token (in network byte order) in the high qword of the byte array, we will effectively be sorting as if this was a tuple/uuid <token>:<low stuff>, which in turn means that stream ID:s are in token order.
This sort of matters at least in alternator, where we want to identify token ranges between generations to emulate the "resharding" in dynamo etc. Could of course just re-sort based on token, but imho all code is currently written using an implicit assumption that token order == id order. 
Also, remember, it was you who said "version" should be in lowest (first) 4 bits. That is exactly where they are now.

haaawk: I believed I said first 4 bits and I meant by this bits no. 0, 1, 2, and 3 in a blob. It felt more natural to me but if alternator needs the order to be preserved then I guess there's no way around it. Unless we do <version><token><rest> but not sure it makes sense

elcallio: It does not "need" it, since itcan always be sorted with a custom order. My point was just that it was at least one of the places where having token as msb imho makes sense. And I think it does in general as well. 
In fact, even from a debugging standpoint it makes sense, since it is much easier to see version that is `qwords{1]&0xf` compared to `(qword[0] >> 60)&0xf` when looking at a memdump or whatever. For me at least.

kbr-:
>  and I think it's a bit better to have
`<version><index><random><token>`

Why?
What if the ID becomes longer in the future? The version would then end up being somewhere in the middle of the ID.
I think it's nice that it's always in the 4 lower bits. Predictable.

cdc/generation.cc
kbr-: > I meant by this bits no. 0, 1, 2, and 3 in a blob

Yeah, but which are those? Does `0` mean the most significant or the least significant bit?

Right now the version is the first 4 least significant bits, and it makes sense to me. When we make the ID longer we would extend it "to the left" (add more significant bits).

cdc/generation.cc
```diff
@@ -59,14 +59,57 @@ static void copy_int_to_bytes(int64_t i, size_t offset, bytes& b) {
...
 
-stream_id::stream_id(bytes b) : _value(std::move(b)) { }
+stream_id::stream_id(bytes b) 
+    : _value(std::move(b)) 
+{
+    // this is not a very solid check. Id:s previous to GA/versioned id:s
+    // have fully random bits in low qword, so this could go either way...
+    if (version() > version_1) {
+        throw std::invalid_argument("Unknown CDC stream id version");
+    }
```
kbr-, line 111: I hope we won't have rolling upgrade problems...

(we probably will, because someone out there has turned on experimental on a production cluster and started writing stuff using CDC for sure, but meh)

elcallio: Yeah, I feel it is open for debate if this check should be here or not. But for non-experimental (i.e. real usage) we probably want it to exist.

elcallio:
spaced pruned.

cdc/generation.cc
haaawk: I know it is equivalent to have the first 4 bits or the last 4 bits being the version. It just feels weird. How many protocols do you know that first put a message and then the type or version of the message?

Regarding the order, I wouldn't like to depend on that. In the future we may have totally different representation which may sort totally differently

I won't insist though as long as the version is on any of the ends to be easily accessible.

elcallio:
ping?

cdc/generation.cc
haaawk: Have you tested rolling upgrade @elcallio ?

elcallio:
> 
> 
> LGTM but please make sure the check in stream_id(bytes) does not break rolling upgrade.

Do you mean remove the assert? An upgrade using CDC will most likely break with this change, but I don't really see how we prevent this, except by changing the length, which I would really like not to do...

haaawk:
> > LGTM but please make sure the check in stream_id(bytes) does not break rolling upgrade.
> 
> Do you mean remove the assert? An upgrade using CDC will most likely break with this change, but I don't really see how we prevent this, except by changing the length, which I would really like not to do...

I guess you're right and since CDC is experimental we can ask users to turn it off before upgrade.

elcallio:
To be clear: this will break upgrade if user ran experimental CDC. He must first disable it, restart cluster, upgrade, then re-enable CDC across the board. And even then, he should probably do a manual delete of the old CDC topology tables.

cdc/generation.cc
```diff
@@ -59,14 +59,57 @@ static void copy_int_to_bytes(int64_t i, size_t offset, bytes& b) {
     std::copy_n(reinterpret_cast<int8_t*>(&i), sizeof(int64_t), b.begin() + offset);
 }
 
-stream_id::stream_id(int64_t first, int64_t second)
+static constexpr auto stream_id_version_bits = 4;
+static constexpr auto stream_id_random_bits = 38;
```
nyh, line 63: I'm making some nitpicks here, but you can consider them for a followup patch - Seeing that @haaawk and @kbr- have already oked this patch I'm planning to commit it unless I find a big hole in it.
My first nitpick is that you should perhaps have a comment that the random bits vs index bits is true for version 0 (or whatever you call it) and may change in a future version.
My second nitpick is that I have no idea what "index" refers to, this probably deserves a comment.

elcallio:
We could potentially add code to the calculation (and maybe also alternator?) that checks if old id:s pass inspection and if not discards them (and deletes from tables) on init + force recalculation. But that is a little scary...

haaawk:
We could use a nodetool command to force creation of the new generation and clean up the generation tables first.

cdc/generation.hh
```diff
@@ -64,21 +64,26 @@ namespace cdc {
 class stream_id final {
     bytes _value;
 public:
+    static constexpr uint8_t version_1 = 1;
```
nyh, line 67: Silly question - why not start with 0? :-)

elcallio:
I think that is more difficult, because iirc, code to check and maybe recalculate id:s is async, early, and might hit the assert when looking at old generations. So really need those tables empty on startup or handling of the error in calculation.

elcallio:
I am unsure how much we want to ease the burden for experimental users. Seems like a political descension. 
We can of course just remove the assert, or downgrade it to a warning, or handle the potential issue elsewhere (calculation?).
We can assert on length instead, and add assert in computation code that _current_ generation must have matching version numbers.

haaawk:
@tzach @slivne What's your take on this? Can we force experimental users to do manual changes in system tables on upgrade or should we handle everything in code?

cdc/generation.hh
elcallio: Mainly because zero is the result of a = {}, so non-default values is nice.

cdc/generation.cc
elcallio: My approach to bits changing is changing the constant names to `stream_id_version_bits_v1` and `stream_id_version_bits_v2` _when_ the change happens, but keep it as a singular constant even by name until then, to not mislead someone to think there is another one somewhere until there is. 

Index == vnode index, commented elsewhere, but could maybe use a few extra.

nyh:
Committed to next. a6ad70d3da106ef34764098faadbed6b1c19e04c

haaawk:
Thanks @nyh. @elcallio could you update cdc docs with the explanation of the new format, please?

elcallio:
> 
> 
> Thanks @nyh. @elcallio could you update cdc docs with the explanation of the new format, please?

Which doc do you mean? The `docs/cdc.md` or any of the google docs? The former mentions id:s, but does not go into much detail on what bits in them represent. Should it?

haaawk:
docs/cdc.md. Would be good to have the format described with explanation what is each part used for

PR #6986, opened 2020-08-05
Implement prior feedback for tombstone-range checks

dekimir:
A couple of previous PRs received feedback on the tombstone-range check in delete_statement.  This PR implements it.

Tests: unit (dev)


avikivity:
Please title your series according to what they do, not what motivated you to write them.

Issue #6397, opened 2020-05-07
Node properties are null in system.peers for node that was decommissioned and removed


juliayakovlev:
*Installation details*
Scylla version (or git commit hash): 4.0.rc3-0.20200501.eee4c00e29
Cluster size: 9 nodes (multidc)
OS (RHEL/CentOS/Ubuntu/AWS AMI): ami-016482914844d8a49,  ami-08eeb1a787e407ef9, ami-0c6075ec8f222315b
 
Found "null" value in node properties in system.peers for node that was decommissioned and removed.
Maybe as issue [#4652](https://github.com/scylladb/scylla/issues/4652)

This is the flow before the error.

1. ```2020-05-06 13:18:12,623``` - Started random_disrupt_method grow_shrink_cluster
    During this nemesis node node 8  with IP 54.165.104.245 was decommissioned and removed:
```
2020-05-06T13:40:48+00:00  longevity-counters-multidc-refactor-db-node-42faa265-8 !INFO    | scylla: [shard 0] gossip - My status = LEFT
```

2. ```2020-05-06 14:45:19,090``` - Started random_disrupt_method restart_then_repair_node on node 9 (longevity-counters-multidc-refactor-db-node-42faa265-9, IP: [3.82.216.102 | 172.30.0.246])
    ```2020-05-06 14:50:51,086``` - Finished

3. ```2020-05-06 15:31:20,395``` - Started random_disrupt_method stop_wait_start_scylla_server on node 9
    ```2020-05-06 15:37:35,009``` - Finished

Error: SYSTEM.PEERS table  on node 9 keeps row about decommissioned node 8 with IP 54.165.104.245 but value in every column became "null":
```
'54.165.104.245 |        null |           null | null |            null |           null |         null"
```

Property for node 8 is not null in system.peers on all other nodes.

Should the info about decommissioned node be kept in the system.peers?

bentsi:
happened to me too, after decommission nemesis in alternator multi-dc scenario. Pasted details in #4652

slivne:
@asias  any explanation on this ?

asias:
> Error: SYSTEM.PEERS table on node 9 keeps row about decommissioned node 8 with IP 54.165.104.245 but value in every column became "null":
> 
> ```
> '54.165.104.245 |        null |           null | null |            null |           null |         null"
> ```
> 
> Property for node 8 is not null in system.peers on all other nodes.

[...]

Issue #4652, opened 2019-07-04
Node properties are null in system.peers in some cases


tgrabiec:
Scylla version: 3.1.0.rc2-0.20190630.acff367ea

I observed a discrepancy between gossip info in memory and in system.peers for one of the nodes:

```
$ nodetool gossipinfo

/10.0.250.129
  SCHEMA:ac26e779-ea83-3001-8dcb-1cd43291d467
```

```
$ cqlsh -e "select peer         , data_center , host_id , preferred_ip , rack , release_version , rpc_address , schema_version , supported_features  from system.peers where peer = '10.0.250.129'"

 peer         | data_center | host_id | preferred_ip | rack | release_version | rpc_address | schema_version | supported_features
--------------+-------------+---------+--------------+------+-----------------+-------------+----------------+--------------------
 10.0.250.129 |        null |    null |         null | null |            null |        null |           null |               null

(1 rows)
```

```
$ cqlsh -e "select peer, schema_version from system.peers"

 peer         | schema_version
--------------+--------------------------------------
   10.0.2.228 | ac26e779-ea83-3001-8dcb-1cd43291d467
  10.0.13.243 | ac26e779-ea83-3001-8dcb-1cd43291d467
  10.0.105.70 | ac26e779-ea83-3001-8dcb-1cd43291d467
 10.0.226.255 | ac26e779-ea83-3001-8dcb-1cd43291d467
 10.0.250.129 |                                 null
   10.0.85.71 | ac26e779-ea83-3001-8dcb-1cd43291d467
```

Not sure what conditions led to this. Other nodes have full info for 10.0.250.129.

slivne:
@roydahan can we add this to SCT to check nodetool gossipinfo / select from system.peers and detect this.

roydahan:
Yes, we can add it as a nemesis that occasionally run both and compare.

slivne:
@roydahan can we do it more aggressively - we need to find the bug .... I know we have nodetool status from time to time - can we add this two checks nodetool gossipinfo + select from system.peers

roydahan:
We have nodetool status before and every after nemesis or cluster layout operation.
We can do that.
We also can do it in a fixed interval if we want...

slivne:
@roydahan so lets do this change - I need this deployed so we can find the bug

roydahan:
@juliayakovlev is trying to reproduce by adding these queries everytime we run nodetool status in a longevity.

slivne:
[...]

PR #6988, opened 2020-08-06
storage_service: Avoid updating tokens in system.peers for nodes to be removed

asias:
Consider:

1) Start n1,n2,n3
2) Stop n3
3) Start n4 to replace n3 but list n4 as seed node
4) Node n4 finishes replacing operation
5) Restart n2
6) Run SELECT * from system.peers on node or node 1.
   cqlsh> SELECT * from system.peers ;
   peer| data_center | host_id| preferred_ip | rack  | release_version | rpc_address | schema_version| supported_features| tokens
   127.0.0.3 |null |null | null |  null | null |null |null |null |   {'-90410082611643223', '5874059110445936121'}

The replaced old node 127.0.0.3 shows in system.peers.

(Note, since commit 399d79fc6f1413f68f88e617386d9c4f54da1889 (init: do not allow
replace-address for seeds), step 3 will be rejected. Assume we use a version without it)

The problem is that n2 sees n3 is in gossip status of SHUTDOWN after
restart. The storage_service::handle_state_normal callback is called for
127.0.0.3. Since n4 is using different token as n3 (seed node does not bootstrap so
it uses new tokens instead of tokens of n3 which is being replaced), so
owned_tokens will be set. We see logs like:

    [shard 0] storage_service - handle_state_normal: New node 127.0.0.3 at token 5874059110445936121

    [shard 0] storage_service - Host ID collision for cbec60e5-4060-428e-8d40-9db154572df7 between 127.0.0.4
    and 127.0.0.3; ignored 127.0.0.3

As a result, db::system_keyspace::update_tokens
will be called to write to system.peers for 127.0.0.3 wrongly.

    if (!owned_tokens.empty()) {
	db::system_keyspace::update_tokens(endpoint, owned_tokens)
    }

To fix, we should skip calling db::system_keyspace::update_tokens if the
nodes is present in endpoints_to_remove.

Refs: #4652
Refs: #6397

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -1213,7 +1213,7 @@ void storage_service::handle_state_normal(inet_address endpoint) {
         remove_endpoint(ep);
     }
     slogger.debug("handle_state_normal: endpoint={} owned_tokens = {}", endpoint, owned_tokens);
-    if (!owned_tokens.empty()) {
+    if (!owned_tokens.empty() && !endpoints_to_remove.count(endpoint)) {
         db::system_keyspace::update_tokens(endpoint, owned_tokens).then_wrapped([endpoint] (auto&& f) {
             try {
                 f.get();


asias:
Ping.

avikivity:
@tgrabiec please review. It looks simple, but I can't tell if the new condition is leaving out some times where we should update the peers table.

tgrabiec:
@asias I don't understand how the scenario from the description is possible:

 1) Step 3 is "Start n4 to replace n3 but list n4 as seed node". How does n4 learn about the cluster if it thinks it's the only seed?
 1) Replacing a node with different tokens is not currently supported and should not be allowed, it requires support for more than one pending endpoint for a given token.

tgrabiec:
The code change itself looks ok.

asias:
> @asias I don't understand how the scenario from the description is possible:
> 
>     1. Step 3 is "Start n4 to replace n3 but list n4 as seed node". How does n4 learn about the cluster if it thinks it's the only seed?

n4 is not the only seed node. It is one of the seed nodes. What matters is that the node think itself is a seed node. 

>     2. Replacing a node with different tokens is not currently supported and should not be allowed, it requires support for more than one pending endpoint for a given token.

Yes, it is not supported. The situation happens only when 399d79fc6f1413f68f88e617386d9c4f54da1889 (init: do not allow
    replace-address for seeds) is not included and user makes an config error which makes the replacing node a seed node.

asias:
> @tgrabiec please review. It looks simple, but I can't tell if the new condition is leaving out some times where we should update the peers table.

The logic is quite simple. If it is a node to be removed, we should never store the tokens in system.peers for it.

tgrabiec:
Merged in fefa35987bdffdcd34fd817dcd65f87a619b634c

PR #6994, opened 2020-08-07
cql3: Move #include from .hh to .cc

dekimir:
restrictions.hh included fmt/ostream.h, which is expensive due to its
transitive #includes.  Replace it with fmt/core.h, which transitively
includes only standard C++ headers.

As requested by #5763 feedback:
https://github.com/scylladb/scylla/pull/5763#discussion_r443210634

Tests: unit (dev)

### DIFF
--- a/cql3/restrictions/restriction.hh
+++ b/cql3/restrictions/restriction.hh
@@ -48,7 +48,7 @@
 #include <variant>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


avikivity:
Queued as 8cae61ee6b39e9fedd1a38a2b9f624cce379026e

PR #6996, opened 2020-08-07
cql3: Move expressions to their own namespace

dekimir:
Move the classes representing CQL expressions (and utility functions
on them) from the `restrictions` namespace to a new namespace `expr`.

Most of the restriction.hh content was moved verbatim to
expression.hh.  Similarly, all expression-related code was moved from
statement_restrictions.cc verbatim to expression.cc.

As suggested in #5763 feedback
https://github.com/scylladb/scylla/pull/5763#discussion_r443210498

Tests: dev (unit)


cql3/expr/expression.cc
```diff
@@ -0,0 +1,872 @@
...
+
+#include "cql3/lists.hh"
+#include "cql3/tuples.hh"
+#include "types/list.hh"
+#include "types/map.hh"
+#include "types/set.hh"
+#include "utils/like_matcher.hh"
+
+namespace cql3 {
+namespace expr {
```
avikivity, line 36: Note (no need to change): you can write `namespace cql3::expr`, reducing some repetitions.

avikivity: Note2 (ditto): I expect we'll move this out of the cql3 namespace eventually and make it a common resource for all front-ends.

PR #6998, opened 2020-08-08
transport: Allow user to disable unencrypted native transport

### LINKED ISSUE #6997: CQL unecrypted native transport is enabled unconditionally
@gnumoreno writes:

```
We enabled ssl and are using native_transport_port_ssl: 9142
Commenting out native_transport_port: 9042 does not disable 9042. Is there a way to disable it?
```

As it turns out, we enable the unencrypted native transport unconditionally in `controller::do_start_server()`. This makes it impossible to configure Scylla to only accept encrypted native transport.

penberg:
Let users disable the unencrypted native transport too by setting the port to
zero in the scylla.yaml configuration file.

Fixes #6997

### DIFF
--- a/docs/protocols.md
+++ b/docs/protocols.md
@@ -92,13 +92,17 @@ The CQL binary protocol is Cassandra's and Scylla's main client-facing
 protocol. Scylla supports several Scylla-only extensions to this protocol,
 described in [protocol-extensions.md](protocol-extensions.md).
 
-By default scylla listens to the CQL protocol on port 9042, which can be
-configured via the `native_transport_port` configuration option, or entirely
-disabled by setting the `start_native_transport` option to 0.
-Again these option names were chosen for backward-compatibility with Cassandra
-configuration files: They refers to CQL as the "native transport", to contrast
-with the older Thrift protocol (described below) which wasn't native to
-Cassandra.
+By default, Scylla listens to the CQL protocol on port 9042, which can be
+configured via the `native_transport_port` configuration option. Scylla also
+supports the CQL protocol via TLS/SSL encryption, which can be enabled via the
+`native_transport_port_ssl` configuration option (default port is 9142). Users
+that want to only support the encrypted CQL protocol can disable the
+unencrypted default CQL protocol by setting the `native_transport_port` to 0.
+The CQL protocol support can be disabled altogether by setting the
+`start_native_transport` option to `false`. These option names were chosen for
+backward-compatibility with Cassandra configuration files: They refers to CQL
+as the "native transport", to contrast with the older Thrift protocol
+(described below) which wasn't native to Cassandra.
 
 There is also a `rpc_address` configuration option to set the IP address
 (and therefore network interface) on which Scylla should listen for the
--- a/transport/controller.cc
+++ b/transport/controller.cc
@@ -92,7 +92,11 @@ future<> controller::do_start_server() {
             std::shared_ptr<seastar::tls::credentials_builder> cred;
         };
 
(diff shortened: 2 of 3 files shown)


transport/controller.cc
```diff
@@ -92,7 +92,11 @@ future<> controller::do_start_server() {
             std::shared_ptr<seastar::tls::credentials_builder> cred;
         };
 
-        std::vector<listen_cfg> configs({{ socket_address{ip, cfg.native_transport_port()}, false }});
+        std::vector<listen_cfg> configs;
+
+        if (cfg.native_transport_port.is_set()) {
```
haaawk, line 97: I don't think this is a good idea to drop the default. This is not a backwards compatible change and can break existing deployments. If we decide to go with this, let's at least check `cfg.native_transport_port.is_set() || !cfg.native_transport_port_ssl.is_set()` to disable this only when SSL is enabled.

avikivity: Agree. We can use a port number of zero as a hacky way of disabling it.

penberg: @avikivity I don't understand what you mean here. If I change the if expression to @haaawk's suggestion, thing will work as expected, and there's no need for the zero port hack, right?

avikivity: I agreed with "I don't think this is a good idea to drop the default".

The trick with the boolean expression is too tricky. It changes a configuration that had 9042 defaulted and 9142 non-defaulted to one that doesn't have 9042 at all.

penberg:
Switched to using zero as a way to disable the port and documented the options.

@nyh @haaawk @avikivity please re-review.

docs/protocols.md
```diff
@@ -92,13 +92,17 @@ The CQL binary protocol is Cassandra's and Scylla's main client-facing
...
-configured via the `native_transport_port` configuration option, or entirely
-disabled by setting the `start_native_transport` option to 0.
-Again these option names were chosen for backward-compatibility with Cassandra
-configuration files: They refers to CQL as the "native transport", to contrast
-with the older Thrift protocol (described below) which wasn't native to
-Cassandra.
+By default, Scylla listens to the CQL protocol on port 9042, which can be
+configured via the `native_transport_port` configuration option. Scylla also
+supports the CQL protocol via TLS/SSL encryption, which can be enabled via the
+`native_transport_port_ssl` configuration option (default port is 9142). Users
```
nyh, line 98: This sentence is ambiguous. The phrase "can be enabled via..." suggests it's not enabled by default and becomes enable if you set this `native_transport_port_ssl` option - but then you go and say it is actually set to 9142 by default. So does SSL listen to port 9142 or not? Or maybe the real story is different - and other options (setting the certificate, etc.) are what actually enables ssl? I think we need a clearer (and possibly more correct) explanation here on what is really needed to enable SSL.

docs/protocols.md
```diff
@@ -92,13 +92,17 @@ The CQL binary protocol is Cassandra's and Scylla's main client-facing
...
+that want to only support the encrypted CQL protocol can disable the
+unencrypted default CQL protocol by setting the `native_transport_port` to 0.
+The CQL protocol support can be disabled altogether by setting the
+`start_native_transport` option to `false`. These option names were chosen for
+backward-compatibility with Cassandra configuration files: They refers to CQL
+as the "native transport", to contrast with the older Thrift protocol
+(described below) which wasn't native to Cassandra.
```
nyh, line 105: Also need to remove the "TODO: again mention SSL, and `native_transport_port_ssl` (default 9142)." line below.

Issue #6973, opened 2020-08-04
Scylla AMI not starting


slivne:
AMI: scylla_2020-08-03T23-11-29Z (ami-05b55eb1eb02c4a3a)

```
[scyllaadm@ip-10-99-0-196 ~]$ rpm -qa | grep scylla
scylla-conf-666.development-0.20200803.4edfdfa78da.x86_64
scylla-tools-666.development-20200803.aa7898d7711.noarch
scylla-debuginfo-666.development-0.20200803.4edfdfa78da.x86_64
scylla-python3-3.8.5-0.20200803.4edfdfa78da.x86_64
scylla-kernel-conf-666.development-0.20200803.4edfdfa78da.x86_64
scylla-tools-core-666.development-20200803.aa7898d7711.noarch
scylla-jmx-666.development-20200803.626fd75173f.noarch
scylla-machine-image-666.development-20200804.f2e30f54a6e.noarch
scylla-server-666.development-0.20200803.4edfdfa78da.x86_64
scylla-666.development-0.20200803.4edfdfa78da.x86_64
```

```
Aug 03 23:16:52 ip-172-30-0-81.ec2.internal systemd[1]: Starting Scylla Cloud Image Setup service...
Aug 03 23:16:52 ip-172-30-0-81.ec2.internal systemd[1]: Stopping Dynamically Generate Message Of The Day...
Aug 03 23:16:52 ip-172-30-0-81.ec2.internal systemd[1]: Starting Dynamically Generate Message Of The Day...
Aug 03 23:16:52 ip-172-30-0-81.ec2.internal systemd[1]: Created slice Slice used to run Scylla. Maximum priority for IO and CPU.
Aug 03 23:16:52 ip-172-30-0-81.ec2.internal systemd[1]: Starting Slice used to run Scylla. Maximum priority for IO and CPU.
Aug 03 23:16:52 ip-172-30-0-81.ec2.internal sudo[5254]: pam_unix(sudo:session): session closed for user root
Aug 03 23:16:52 ip-172-30-0-81.ec2.internal sudo[3944]: pam_unix(sudo:session): session closed for user root
Aug 03 23:16:52 ip-172-30-0-81.ec2.internal scylla_image_setup[14076]: 2020-08-03 23:16:52,916 - [scylla_configure] - INFO - Getting 'http://169.254.169.254/l
[...]

PR #7003, opened 2020-08-10
scylla_swap_setup improvement

### LINKED ISSUE #6947: UX: scylla_setup: include swap size in the prompt
*Installation details*
Scylla version (or git commit hash): 666.development-0.20200727.a7df8486b
Cluster size: 1

Current prompt: 
```
No swap is configured, it is highly recommended to setup swap on Scylla node.
Do you want to configure swapfile?', 'Answer yes to setup swapfile to Scylla, Answer no to do nothing.
```

Problem:
- The swap size is calculated automatically after user inputs the directory and confirms. User doesn't know how much space will be used.

Suggestion:
- include the swap size in the prompt

```
No swap is configured, it is highly recommended to setup swap on Scylla node.
Do you want to configure swapfile?', 'Answer yes to setup swapfile (xxG) to Scylla, Answer no to do nothing.
```

@tzach

syuu1228:
As I described at https://github.com/scylladb/scylla/issues/6973#issuecomment-669705374,
we need to avoid disk full on scylla_swap_setup, also we should provide manual configuration of swap size.

This PR provides following things:
 - use psutil to get memtotal and disk free, since it provides better APIs
 - calculate swap size in bytes to avoid causing error on low-memory environment
 - prevent to use more than 50% of disk space when auto-configured swap size, abort setup when almost no disk space available (less than 2GB)
 - add --swap-size option to specify swap size both on scylla_swap_setup and scylla_setup
 - add interactive swap size prompt on scylla_setup

Fixes #6947 
Related #6973
Related scylladb/scylla-machine-image#48

### DIFF
--- a/dist/common/scripts/scylla_setup
+++ b/dist/common/scripts/scylla_setup
@@ -98,6 +98,15 @@ def interactive_choose_swap_directory():
         else:
             return d
 
+def interactive_choose_swap_size():
+    print('Please input swap size in GB:')
+    while True:
+        s = input('> ')
+        try:
+            return int(s)
+        except Exception as e:
+            print(f'"{s}" is not valid number, input again')
+
 def do_verify_package(pkg):
     if is_debian_variant():
         res = run('dpkg -s {}'.format(pkg), silent=True, exception=False)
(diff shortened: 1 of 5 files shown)


avikivity:
Queued as caa28081fc1a47867388cba2434873c783cd3d85

PR #7005, opened 2020-08-10
database: Do not assert on replay positions if truncate does not flush

### LINKED ISSUE #6995: Crash with assertion failure when dropping a table with auto_snapshot = false
The following trivial Alternator test creates - using the Alternator API - a single partition with 10,000 items where each has a 10,000 byte string attribute - so totalling a little over 100 MB of data. The data is successfully written, but at the end of the test, the temporary table is deleted, and at that point *then* Scylla crashes with an assertion failure.
The test:
```
def test_big(scylla_only, test_table_sn):
    p = random_string()
    str = 'x' * 10240
    with test_table_sn.batch_writer() as batch:
        for i in range(10000):
            batch.put_item({'p': p, 'c': i, 's': str})
```
The end of the log:
```
INFO  2020-08-07 16:56:27,839 [shard 0] schema_tables - Dropping alternator_alternator_Test_1596808583729.alternator_Test_1596808583729 id=c794a830-d8b5-11ea-9bbe-000000000000 version=b29143f4-29d5-37fd-8c5d-61377619232e
test_scylla: database.cc:1845: database::truncate(const keyspace&, column_family&, database::timestamp_func, bool)::<lambda()>::<lambda()> mutable::<lambda()>::<lambda(db_clock::time_point)>::<lambda()>::<lambda(db::replay_position)>: 
Assertion `low_mark <= rp || rp == db::replay_position()' failed.
Aborting on shard 1.
```
There are no other warnings or errors in the log.

I don't know how Alternator factors into this bug. Alternator doesn't have any special changes to how `database::truncate()` works so it is probably a pre-existing bug in that code. On the other hand, it is possible that something specific in how that test was written - maybe that it made a lot of big writes quickly or never waited for memtable flush or something - exposed this pre-existing bug.

@elcallio "git blame" tells me you write that failing assertion. Can you please help understand what it means and what might cause it to trigger?

elcallio:
Fixes #6995

In c2c6c71 the assert on replay positions in flushed sstables discarded by
truncate was broken, by the fact that we no longer flush all sstables
unless auto snapshot is enabled.

This means the low_mark assertion does not hold, because we maybe/probably
never got around to creating the sstables that would hold said mark.

Note that the (old) change to not create sstables and then just delete
them is in itself good. But in that case we should not try to verify
the rp mark.

### DIFF
--- a/database.cc
+++ b/database.cc
@@ -1842,7 +1842,11 @@ future<> database::truncate(const keyspace& ks, column_family& cf, timestamp_fun
                             // TODO: indexes.
                             // Note: since discard_sstables was changed to only count tables owned by this shard,
                             // we can get zero rp back. Changed assert, and ensure we save at least low_mark.
-                            assert(low_mark <= rp || rp == db::replay_position());
+                            // #6995 - the assert below was broken in c2c6c71 and remained so for many years. 
+                            // We nowadays do not flush tables with sstables but autosnapshot=false. This means
+                            // the low_mark assertion does not hold, because we maybe/probably never got around to 
+                            // creating the sstables that would create them.
+                            assert(!should_flush || low_mark <= rp || rp == db::replay_position());
                             rp = std::max(low_mark, rp);
                             return truncate_views(cf, truncated_at, should_flush).then([&cf, truncated_at, rp] {
                                 // save_truncation_record() may actually fail after we cached the truncation time


nyh:
I was about to confirm that it fixes my test and commit it, but Avi beat me.
At least I can do the first part - I verified that this patch indeed fixes my test (which creates a table and deletes it, with auto_snapshot off). Thanks.

PR #7010, opened 2020-08-11
repair: Track repair_meta created on both repair follower and master

### LINKED ISSUE #7009: Track repair_meta created on both repair follower and master
It is pretty hard to find the repair_meta object when debugging a core.
We can put the repair_meta object created by both repair follower and master into a map.

asias:
It is pretty hard to find the repair_meta object when debugging a core.
This patch makes it is easier by putting repair_meta object created by
both repair follower and master into a map.

Fixes #7009

### DIFF
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -764,10 +764,7 @@ public:
         });
     }
 
-    static std::unordered_map<node_repair_meta_id, lw_shared_ptr<repair_meta>>& repair_meta_map() {
-        static thread_local std::unordered_map<node_repair_meta_id, lw_shared_ptr<repair_meta>> _repair_metas;
-        return _repair_metas;
-    }
+    static std::unordered_map<node_repair_meta_id, lw_shared_ptr<repair_meta>>& repair_meta_map();
 
     static lw_shared_ptr<repair_meta> get_repair_meta(gms::inet_address from, uint32_t repair_meta_id) {
         node_repair_meta_id id{from, repair_meta_id};
@@ -1843,6 +1840,12 @@ public:
         });
     }
 };
+
+// The repair_metas created passively, i.e., local node is the repair follower.
+static thread_local std::unordered_map<node_repair_meta_id, lw_shared_ptr<repair_meta>> _repair_metas;
+std::unordered_map<node_repair_meta_id, lw_shared_ptr<repair_meta>>& repair_meta::repair_meta_map() {
+    return _repair_metas;
+}
 
 static future<stop_iteration> repair_get_row_diff_with_rpc_stream_process_op(
         gms::inet_address from,
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -43,6 +43,7 @@
 #include <random>
 #include <optional>
 #include <boost/range/adaptors.hpp>
+#include <boost/intrusive/list.hpp>
 #include "../db/view/view_update_generator.hh"
 #include "gms/i_endpoint_state_change_subscriber.hh"
 #include "gms/gossiper.hh"
@@ -623,13 +624,21 @@ public:
     }
 };
 
+class repair_meta;
+class repair_meta_tracker;
+
+static void add_to_repair_meta_for_masters(repair_meta& rm);
+static void add_to_repair_meta_for_followers(repair_meta& rm);
+
 class repair_meta {
+    friend repair_meta_tracker;
 public:
     using repair_master = bool_class<class repair_master_tag>;
     using update_working_row_buf = bool_class<class update_working_row_buf_tag>;
     using update_peer_row_hash_sets = bool_class<class update_peer_row_hash_sets_tag>;
     using needs_all_rows_t = bool_class<class needs_all_rows_tag>;
     using msg_addr = netw::messaging_service::msg_addr;
+    using tracker_link_type = boost::intrusive::list_member_hook<bi::link_mode<boost::intrusive::auto_unlink>>;
 private:
     seastar::sharded<database>& _db;
     column_family& _cf;
@@ -674,6 +683,7 @@ private:
     sink_source_for_get_full_row_hashes _sink_source_for_get_full_row_hashes;
     sink_source_for_get_row_diff _sink_source_for_get_row_diff;
     sink_source_for_put_row_diff _sink_source_for_put_row_diff;
+    tracker_link_type _tracker_link;
 public:
     repair_stats& stats() {
         return _stats;
@@ -751,6 +761,11 @@ public:
                         return netw::get_local_messaging_service().make_sink_and_source_for_repair_put_row_diff_with_rpc_stream(repair_meta_id, addr);
                 })
             {
+            if (master) {
+                add_to_repair_meta_for_masters(*this);
+            } else {
+                add_to_repair_meta_for_followers(*this);
+            }
     }
 
 public:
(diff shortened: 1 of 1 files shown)


denesb:
I think better yet would be to just link them into an intrusive list in the debug namespace.

denesb:
Like we do with seastar threads and sstables.

repair/row_level.cc
```diff
@@ -2523,7 +2534,7 @@ class row_level_repair {
             auto schema_version = s->version();
             bool table_dropped = false;
 
-            repair_meta master(_ri.db,
+            auto rm = make_lw_shared<repair_meta>(_ri.db,
```
avikivity, line 2537: Why this change? You could keep it as it is, and register a pointer to `master` in the map.

repair/row_level.cc
```diff
@@ -1844,6 +1843,18 @@ class repair_meta {
...
+std::unordered_map<node_repair_meta_id, lw_shared_ptr<repair_meta>>& repair_meta::repair_meta_map() {
+    return _repair_metas;
+}
+
+// The repair_metas created actively, i.e., local node is the repair master.
+static thread_local std::unordered_map<node_repair_meta_id, lw_shared_ptr<repair_meta>> _repair_metas_masters;
+std::unordered_map<node_repair_meta_id, lw_shared_ptr<repair_meta>>& repair_meta::repair_meta_map_masters() {
+    return _repair_metas_masters;
+}
+
```
avikivity, line 1857: We often keep such debug information in the debug namespace, consider doing the same.

denesb: Also note that using a map has no advantage while debugging, we don't have the hash or the less functions available (or runnable) so the way to find anything in them is to iterate over all items. You might as well just use a simple `std::list` or `boost::intrusive::list`.

repair/row_level.cc
asias: I wanted the _repair_metas_masters  and _repair_metas (existing) one to be the same.

asias: Using row pointer is even better in turns of help debugging.

repair/row_level.cc
asias: This is a good point.  Using list is simpler but it is slower to remove the items. We are not creating those objects on the fast path. I think it is fine.

asias:
> I think better yet would be to just link them into an intrusive list in the debug namespace.

Can you point me the relevant code?

repair/row_level.cc
```diff
@@ -764,10 +764,9 @@ class repair_meta {
         });
     }
 
-    static std::unordered_map<node_repair_meta_id, lw_shared_ptr<repair_meta>>& repair_meta_map() {
-        static thread_local std::unordered_map<node_repair_meta_id, lw_shared_ptr<repair_meta>> _repair_metas;
```
denesb, line 768: Very happy to see this static move out the method, working with function statics is a major pain in gdb. It has a very weird syntax that is so easy to get wrong (or forget) and gdb often forgets the type, and spelling out correctly a template (with all the hidden defaults) is one of the hardest things to do in gdb.

asias: Yes. That is why I am moving this out. It is painful to work with it!

asias: btw, sometimes, thread_local is not working. so, I am considering do not use the thread_local for the repair_meta objects in debug space.

denesb:
> > I think better yet would be to just link them into an intrusive list in the debug namespace.
> 
> Can you point me the relevant code?

https://github.com/scylladb/scylla/blob/74db08165ddcc2db061b88b4a28e0fa682305655/sstables/sstables.hh#L135
https://github.com/scylladb/scylla/blob/74db08165ddcc2db061b88b4a28e0fa682305655/sstables/sstables.hh#L545
https://github.com/scylladb/scylla/blob/74db08165ddcc2db061b88b4a28e0fa682305655/sstables/sstables.cc#L93-L105
https://github.com/scylladb/scylla/blob/74db08165ddcc2db061b88b4a28e0fa682305655/sstables/sstables.cc#L3512-L3514
Note that `bi::constant_time_size<false>` parameter to `bi::list`, causing the members to be automatically removed on destruction.

repair/row_level.cc
denesb: Removing is updating 2 pointers, I don't think it gets much cheaper than that. :)

repair/row_level.cc
denesb: How do you plan on avoiding `thread_local`? We can't have all threads use the same list, that will result in corruption.

I think you can rely on `thread_local` working, it is such a fundamental building block in our code that without thread debugging there is really no debugging at all.

repair/row_level.cc
asias: We need to search the repair_meta belongs to a specific node_repair_meta_id when we remove it.

repair/row_level.cc
asias: std::vector<std::list<*repair_meta>> metas, metas[x] can only be accessed on shard x.

asias:
Changes in 238cefd8d740d5d459f10e4de3c84f946c296f3e:
- Use boost intrusive list to track repair_meta objects
- Drop _repair_metas_masters
- Export repair_meta list in debug namespace

repair/row_level.cc
```diff
@@ -2686,3 +2720,10 @@ future<> repair_service::stop() {
 repair_service::~repair_service() {
     assert(_stopped);
 }
+
+namespace debug {
+namespace repair {
+    static thread_local repair_meta_tracker* repair_meta_for_masters = &::_repair_meta_for_masters;
+    static thread_local repair_meta_tracker* repair_meta_for_followers = &::_repair_meta_for_followers;
+}
```
denesb, line 2728: I think these pointers are unnecessary, you can just wrap the `_repair_meta_for_masters` and `_repair_meta_for_followers` in `namespace debug::repair { }` and it will be enough.

asias: I wanted to avoid forward declarations for  _repair_meta_for_masters and _repair_meta_for_followers. Yes, I think it is better to put them inside debug namespace.

asias:
Changes in b280f67764f1a5f75199cfb0fffeda89ff9a7d88:
- Drop the point wrapper in debug name space.

repair/row_level.cc
asias: I've sent a new version.

repair/row_level.cc
```diff
@@ -2235,6 +2250,28 @@ future<> repair_uninit_messaging_service_handler() {
...
+public:
+    void add(repair_meta& rm) {
+        _repair_metas.push_back(rm);
+    }
+};
+
+namespace debug {
+    extern thread_local repair_meta_tracker repair_meta_for_masters;
+    extern thread_local repair_meta_tracker repair_meta_for_followers;
+}
```
denesb, line 2266: Why do we need separate forward declaration and defninition later? It's in the same file, can't we just define here and be done with it?

asias: Nice! Yes, we can do that. I was fooled by class dependency thing.

asias:
Changes in e9a520a22b8555056c77050d76c412694cc2e619:
- Drop the unnecessary forward declaration.

avikivity:
Looks fine, will wait for @denesb ack.

denesb:
LGTM.

PR #7017, opened 2020-08-11
docs/cdc.md: Add short explanation of stream ID bit composition

elcallio:
Bit layout, sort order and field usage of CDC stream ids.


nyh:
Thanks. Committed to next ed86e870ee57211d23712385498060efb758cef0

docs/cdc.md
```diff
@@ -12,6 +12,21 @@ public:
...
 
+The 128 bits are composed of:
+
+```
+128           64            27            4             0
+  | <token:64> | <random:38> | <index:22> | <version:4> |
+
+```
+
+With `version` making up the lowest 4 bits. The id is stored as bytes, and sorted on string ordering, i.e. the high qword (token) is msb in ordering.
```
haaawk, line 23: Could you please add a note about what the current version is and that version bits will be used to introduce new versions and what are the constraints on the new versions?

PR #7021, opened 2020-08-11
dist/common/scripts: take disk root mountpoint from env variable 

zimnx:
For k8s pods mountpoint of host disk may be in different location than
current hardcoded one.
ROOT_MOUNTPOINT env variable allows to provide different one.

### DIFF
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -144,10 +144,11 @@ class gcp_instance:
     def _non_root_nvmes(self):
         """get list of nvme disks from os, filter away if one of them is root"""
         nvme_re = re.compile(r"nvme\d+n\d+$")
+        mountpoint = os.environ.get('ROOT_MOUNTPOINT', '/')
 
-        root_dev_candidates = [x for x in psutil.disk_partitions() if x.mountpoint == "/"]
+        root_dev_candidates = [ x for x in psutil.disk_partitions() if x.mountpoint == mountpoint ]
         if len(root_dev_candidates) != 1:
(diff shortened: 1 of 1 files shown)


avikivity:
Environment variables are not a good way to pass parameters, they bypass layers and so it's hard to keep track of all the places that need changes.

Better to pass it as a command line parameter.

tarzanek:
none of those functions should matter with k8s
one is for AWS VMs, the other for selected GCP VMs
if those scripts/functions get run inside container, then this is wrong
@zimnx Maciek, can you share the errors you see from above with me?

I think iotune should be rerun in container, since we have no clue if k8s isn't restricting the bandwidth
I wouldn't set the ready made io_properties in container out of box, but correct me if I am wrong (even if k8s has no notion of restricting container disk bandwidth, you can still run multiple scyllas next to each other, how would they partition the single disk?)

mmatczuk:
I'd opt for a parameter the problem is that we'd need to consequently pass that from script to script. 
Can't we address that with k8s path rewrites @zimnx?

> one is for AWS VMs, the other for selected GCP VMs
@tarzanek AFAIC GKE does not allow to build custom node images (unlike EKS).
Having said that running prepare scripts in containers is a bread and butter of operator development.
It has lot's of advantages, and that's the k8s idiomatic way of doing things read DaemonSets.

zimnx:
I managed to get rid of it by modifying scylla-machine-image script.

tarzanek:
@zimnx can you point me to the changeset for scylla-machine-image? I am just curious what you did

PR #7024, opened 2020-08-12
cql: Forbid adding new fields to UDTs used in partition key columns

### LINKED ISSUE #6941: Forbid adding new fields to UDTs used in partition key columns
Corresponding C* issue:
https://issues.apache.org/jira/browse/CASSANDRA-15933

```cql
cqlsh:ks> create type ut (a int);
cqlsh:ks> create table t (pk frozen<ut> primary key);
cqlsh:ks> insert into t (pk) values ((0));
cqlsh:ks> alter type ut add b int;
cqlsh:ks> select * from t;

 pk
-----------------
 {a: 0, b: null}

(1 rows)
cqlsh:ks> insert into t (pk) values ((0, null));
cqlsh:ks> select * from t;

 pk
-----------------
 {a: 0, b: null}
 {a: 0, b: null}

(2 rows)

```

wmitros:
Changing a user type may allow adding duplicate rows to tables where
this type is used in a partitioning key. Fix by checking all types
of existing partitioning columns before allowing to add new fields to
the type.

Fixes #6941

### DIFF
--- a/cql3/statements/alter_type_statement.cc
+++ b/cql3/statements/alter_type_statement.cc
@@ -87,6 +87,15 @@ void alter_type_statement::do_announce_migration(database& db, ::keyspace& ks, b
         throw exceptions::invalid_request_exception(format("No user type named {} exists.", _name.to_string()));
     }
 
+    for (auto&& schema : ks.metadata()->cf_meta_data() | boost::adaptors::map_values) {
+        for (auto&& column : schema->partition_key_columns()) {
+            if (column.type->references_user_type(_name.get_keyspace(), _name.get_user_type_name())) {
+                throw exceptions::invalid_request_exception(format("Cannot add new field to type {} because it is used in the partition key column {} of table {}.{}",
+                    _name.to_string(), column.name_as_text(), schema->ks_name(), schema->cf_name()));
+            }
+        }
+    }
+
     auto&& updated = make_updated_type(db, to_update->second);
 
     // Now, we need to announce the type update to basically change it for new tables using this type,
--- a/test/cql/forbid_adding_to_udt_used_in_partkey_test.cql
+++ b/test/cql/forbid_adding_to_udt_used_in_partkey_test.cql
@@ -1,0 +1,7 @@
+create type ut (a int);
+create table t (pk frozen<ut> primary key);
+alter type ut add b int;
+drop table t;
+alter type ut add c int;
+create table t2 (pk frozen<list<frozen<ut>>> primary key);
+alter type ut add d int;
--- a/test/cql/forbid_adding_to_udt_used_in_partkey_test.result
+++ b/test/cql/forbid_adding_to_udt_used_in_partkey_test.result
@@ -1,0 +1,30 @@
+create type ut (a int);
+{
+	"status" : "ok"
+}
+create table t (pk frozen<ut> primary key);
+{
+	"status" : "ok"
+}
+alter type ut add b int;
+{
+	"message" : "exceptions::invalid_request_exception (Cannot add new field to type ks.ut because it is used in the partition key column pk of table ks.t)",
+	"status" : "error"
+}
+drop table t;
+{
+	"status" : "ok"
+}
+alter type ut add c int;
+{
+	"status" : "ok"
+}
+create table t2 (pk frozen<list<frozen<ut>>> primary key);
+{
+	"status" : "ok"
+}
+alter type ut add d int;
+{
+	"message" : "exceptions::invalid_request_exception (Cannot add new field to type ks.ut because it is used in the partition key column pk of table ks.t2)",
+	"status" : "error"
+}


avikivity:
What about clustering keys? Are those fine to change?

wmitros:
I've checked it empirically and there were no errors while changing clustering key type. That's probably because although both partition_key and clustering_key are just managing a series of bytes, when comparing partititon_keys all bytes need to be equal and when comparing clustering_keys one chunk of bytes needs to be a prefix of the other to be considered equal. 

While testing it, I came across a weird behaviour:
```
cqlsh:ks> create table t (pk tuple<int, int> primary key);
cqlsh:ks> insert into t (pk) values ((0));
cqlsh:ks> insert into t (pk) values ((0, null));
cqlsh:ks> select * from t;

 pk
-----------
 (0, None)
 (0, None)

(2 rows)
```
Is this an expected result of such series of statements?

avikivity:
> I've checked it empirically and there were no errors while changing clustering key type. That's probably because although both partition_key and clustering_key are just managing a series of bytes, when comparing partititon_keys all bytes need to be equal and when comparing clustering_keys one chunk of bytes needs to be a prefix of the other to be considered equal.

Maybe it's better to forbid it anyway. Okay, it works, but can we explain what it does?

There's no reason that a key should be equal to a prefix.

> 
> While testing it, I came across a weird behaviour:
> 
> ```
> cqlsh:ks> create table t (pk tuple<int, int> primary key);
> cqlsh:ks> insert into t (pk) values ((0));
> cqlsh:ks> insert into t (pk) values ((0, null));
> cqlsh:ks> select * from t;
> 
>  pk
> -----------
>  (0, None)
>  (0, None)
> 
> (2 rows)
> ```
> 
> Is this an expected result of such series of statements?

Maybe the first translates to a single-element tuple and the second to a two-element tuple with zero-length second element. As to "expected", no user will expect it. We have some backwards compatibilty requirements for tuples (allowing extension of tuples) that may explain it.

wmitros:
I've added checking the clustering columns to be safe.

cql3/statements/alter_type_statement.cc
```diff
@@ -87,6 +87,21 @@ void alter_type_statement::do_announce_migration(database& db, ::keyspace& ks, b
...
+                throw exceptions::invalid_request_exception(format("Cannot add new field to type {} because it is used in the partition key of table {}",
+                    _name.to_string(), schema->cf_name()));
+            }
+        }
+        for (auto&& column : schema->clustering_key_columns()) {
+            if (column.type->references_user_type(_name.get_keyspace(), _name.get_user_type_name())) {
+                throw exceptions::invalid_request_exception(format("Cannot add new field to type {} because it is used in the clustering key of table {}",
+                    _name.to_string(), schema->cf_name()));
+            }
+        }
```
haaawk, line 102: Maybe extract a lambda to avoid code duplication?
```
auto check_columns = [&] (auto& columns) {
        for (auto&& column : schema->partition_key_columns()) {
            if (column.type->references_user_type(_name.get_keyspace(), _name.get_user_type_name())) {
                throw exceptions::invalid_request_exception(format("Cannot add new field to type {} because it is used in the primary key of table {}",
                    _name.to_string(), schema->cf_name()));
            }
        }
}
```

haaawk: Also, in the error message I would include the keyspace of the table and the name of the column that uses the UDT.

haaawk (approved):
LGTM. I just left two ideas for improvements. Do you think we could merge it now @avikivity ?

cql3/statements/alter_type_statement.cc
```diff
@@ -87,6 +87,21 @@ void alter_type_statement::do_announce_migration(database& db, ::keyspace& ks, b
...
+            if (column.type->references_user_type(_name.get_keyspace(), _name.get_user_type_name())) {
+                throw exceptions::invalid_request_exception(format("Cannot add new field to type {} because it is used in the partition key of table {}",
+                    _name.to_string(), schema->cf_name()));
+            }
+        }
+        for (auto&& column : schema->clustering_key_columns()) {
+            if (column.type->references_user_type(_name.get_keyspace(), _name.get_user_type_name())) {
+                throw exceptions::invalid_request_exception(format("Cannot add new field to type {} because it is used in the clustering key of table {}",
+                    _name.to_string(), schema->cf_name()));
+            }
```
avikivity, line 101: These loops can cause a reactor stall in thousand-table schemas. Altering a user type is rare so it's not too bad.

Fixing it is hard. While it's easy to add yields in the loop, a concurrent update would invalidate iterators in ` ks.metadata()->cf_meta_data()` and segfault (if we're lucky). So a complete fix involves locking the entire schema, which is impossible now since multiple nodes/shards can alter tables concurrently. So this has to wait for raft-managed schema.

cql3/statements/alter_type_statement.cc
```diff
@@ -87,6 +87,21 @@ void alter_type_statement::do_announce_migration(database& db, ::keyspace& ks, b
         throw exceptions::invalid_request_exception(format("No user type named {} exists.", _name.to_string()));
     }
 
+    for (auto&& schema : ks.metadata()->cf_meta_data() | boost::adaptors::map_values) {
```
avikivity, line 90: Why just the keyspace? I think a table in one keyspace can refer to types in another.

wmitros: In `cql3_type::raw_ut::prepare_internal(const sstring& keyspace, const user_types_metadata& user_types)` we have 
```
if (_name.has_keyspace()) {
    // The provided keyspace is the one of the current statement this is part of. If it's different from the keyspace of
    // the UTName, we reject since we want to limit user types to their own keyspace (see #6643)
    if (!keyspace.empty() && keyspace != _name.get_keyspace()) {
        throw exceptions::invalid_request_exception(sprint("Statement on keyspace %s cannot refer to a user type in keyspace %s; "
                                                           "user types can only be used in the keyspace they are defined in",
                                                        keyspace, _name.get_keyspace()));
    }
```
So if I understand correctly, the current implementation does not allow this.

avikivity: Ah, I remembered differently. But you are correct.

wmitros:
I've somehow missed the test 
```
SEASTAR_TEST_CASE(test_alter_type_on_compact_storage_with_no_regular_columns_does_not_crash) {
    return do_with_cql_env_thread([] (cql_test_env& e) {
        cquery_nofail(e, "CREATE TYPE my_udf (first text);");
        cquery_nofail(e, "create table z (pk int, ck frozen<my_udf>, primary key(pk, ck)) with compact storage;");
        cquery_nofail(e, "alter type my_udf add test_int int;");
    });
} 
```
It was added in the commit 34cff6ed6b299d241ab0428494a049fd0f6dcb24 while fixing the issue https://github.com/scylladb/scylla/issues/4837
I'm not really seeing how we could modify the test while checking for the same issue, so should I remove it?

haaawk:
@avikivity @tgrabiec Do you think we should allow changing UDTs used for in clustering keys in compact storage for some reason? Is there any reason we have to?

tgrabiec:
> @avikivity @tgrabiec Do you think we should allow changing UDTs used for in clustering keys in compact storage for some reason? Is there any reason we have to?

I think we should allow changing UDTs which are part of clustering keys, as apparently there were users in the past which did that, and there is no problem with allowing that.

haaawk:
@wmitros found some issues with udt/tuples in clustering keys but maybe those can be fixed separately.

wmitros:
Considering the doubts about allowing changing UDTs in clustering keys I've reduced the patch to just UDTs in partition keys so that we can merge it and discuss the clustering keys separately

haaawk (approved):
LGTM

PR #7026, opened 2020-08-12
Support building packages on non-x86 platform

syuu1228:
Allow users to build unofficial packages for non-x86 platform.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1262,7 +1262,7 @@ def configure_seastar(build_dir, mode):
 
     dpdk = args.dpdk
     if dpdk is None:
-        dpdk = mode == 'release'
+        dpdk = platform.machine() == 'x86_64' and mode == 'release'
     if dpdk:
         seastar_cmake_args += ['-DSeastar_DPDK=ON', '-DSeastar_DPDK_MACHINE=wsm']
     if args.split_dwarf:
--- a/dist/debian/control.template
+++ b/dist/debian/control.template
@@ -16,7 +16,7 @@ Replaces: %{product}-server (<< 1.1)
 Conflicts: %{product}-server (<< 1.1)
 
 Package: %{product}-server
-Architecture: amd64
+Architecture: any
 Depends: ${shlibs:Depends}, ${misc:Depends}, %{product}-conf, %{product}-python3
 Description: Scylla database server binaries 
  Scylla is a highly scalable, eventually consistent, distributed,
@@ -25,7 +25,7 @@ Description: Scylla database server binaries
 Package: %{product}-server-dbg
 Section: debug
 Priority: extra
-Architecture: amd64
+Architecture: any
 Depends: %{product}-server (= ${binary:Version}), ${shlibs:Depends}, ${misc:Depends}
 Description: debugging symbols for %{product}-server
  Scylla is a highly scalable, eventually consistent, distributed,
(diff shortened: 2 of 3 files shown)


configure.py
```diff
@@ -1260,7 +1260,8 @@ def configure_seastar(build_dir, mode):
 
     dpdk = args.dpdk
     if dpdk is None:
-        dpdk = mode == 'release'
+        if platform.machine() == 'x86_64':
+            dpdk = mode == 'release'
```
avikivity, line 1264: This leaves `dpdk` as None sometimes. It's better to keep it true/false (maybe dpdk = platform.machine() == 'x86_64' and mode == 'release').

dist/redhat/scylla.spec
```diff
@@ -21,6 +21,10 @@ Obsoletes:	scylla-server < 1.1
 %undefine _unique_build_ids
 %global _no_recompute_build_ids 1
 
+%ifnarch x86_64
+%global _missing_build_ids_terminate_build 0
+%endif
```
avikivity, line 26: We'll have to fix it. Please file an issue so we don't forget and copy @espindola. Meanwhile we can live without the build-id.

dist/debian/control.template
```diff
@@ -16,7 +16,7 @@ Replaces: %{product}-server (<< 1.1)
 Conflicts: %{product}-server (<< 1.1)
 
 Package: %{product}-server
-Architecture: amd64
+Architecture: %{arch}
```
avikivity, line 19: I think you can write `any` here: https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-architecture

espindola:

> +%ifnarch x86_64
> +%global _missing_build_ids_terminate_build 0
> +%endif
>
> We'll have to fix it. Please file an issue so we don't forget and copy @espindola. Meanwhile we can live without the build-id.

Could we at least have a comment saying what is mising a buildid on non
x86_64? Do we know why?

Cheers,
Rafael

avikivity:
> Do we know why?

That's why we copied you.

dist/redhat/scylla.spec
syuu1228: made #7046, also find out patchelf breaking build-id

syuu1228:
@avikivity dropped build-id patch created #7046 instead, fixed dpdk patch and debian patch. Please review again

configure.py
```diff
@@ -1262,7 +1262,7 @@ def configure_seastar(build_dir, mode):
 
     dpdk = args.dpdk
     if dpdk is None:
-        dpdk = mode == 'release'
+        dpdk = platform.machine() == 'x86_64' mode == 'release'
```
avikivity, line 1265: Missing `and`

dist/debian/python3/control.template
```diff
@@ -8,7 +8,7 @@ Standards-Version: 3.9.5
 Rules-Requires-Root: no
 
 Package: %{product}-python3
-Architecture: amd64
+Architecture: any
```
avikivity, line 11: Don't forget to patch scylla-python3.git too (and let's revive that patch to use it)

configure.py
syuu1228: @avikivity Just fixed.

dist/debian/python3/control.template
syuu1228: okay

PR #7027, opened 2020-08-12
aws: update enhanced networking supported instance list

### LINKED ISSUE #6991: scylla_setup tool error on AWS instance:  c5n doesn't support enhanced networking!

*Installation details*
Scylla version (or git commit hash): 4.1
Cluster size: 3
OS (RHEL/CentOS/Ubuntu/AWS AMI): CentOS 7.6

UPDATE:  scylla_setup tool succeeds using AWS instances of type **i3en** and since that is the recommended instance type (storage optimized) perhaps the message about **c5n** instances is not relevant: users should be choosing the i3 types. 

The scylla_setup tool is incorrectly claiming AWS c5n instance type doesn't support enhanced networking. 

The **ethtool** claims the ENA driver is being used:

> ethtool -i ens5
driver: ena
version: 2.0.3K
firmware-version:
expansion-rom-version:
bus-info: 0000:00:05.0
supports-statistics: yes
supports-test: no
supports-eeprom-access: no
supports-register-dump: no
supports-priv-flags: no
`

The **modinfo** command says that ENA is enabled.  The c5n instance type definitely supports ENA.  

`
> modinfo ena
filename:       /lib/modules/3.10.0-1127.18.2.el7.x86_64/kernel/drivers/net/ethernet/amazon/ena/ena.ko.xz
version:        2.0.3K
license:        GPL
description:    Elastic Network Adapter (ENA)
author:         Amazon.com, Inc. or its affiliates
retpoline:      Y
rhelversion:    7.8
...
`

syuu1228:
Sync enhanced networking supported instance list to latest one.

Reference: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html

Fixes #6991

### DIFF
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -388,7 +388,7 @@ class aws_instance:
         instance_size = self.instance_size()
         if instance_class in ['c3', 'c4', 'd2', 'i2', 'r3']:
             return 'ixgbevf'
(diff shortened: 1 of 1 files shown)


syuu1228:
We probably better to switch blacklist style, since newer instances normally supported it.

PR #7028, opened 2020-08-12
CDC: allow "full" preimage in logs

### LINKED ISSUE #6935: Alternator Streams: OldImage is missing the old item's key
As checked by `test_streams.py::test_streams_updateitem_old_image`, DynamoDB Streams includes the entire old item - including its key - as the value of OldImage when OLD_IMAGE mode is enabled. Alternator is missing the key columns in this item. (note that OldImage should only exist and contain the key if an old item actually existed - see #6933). 

The good news is that this test confirms Alternator does include in OldImage all the other (non-key) attributes. I was worried it might include only the modified columns, because this is what our CDC's "preimage" supposedly provides. It works because in cases like this test, not involving an LSI or GSI, all non-key attributes are not actual Scylla columns, but rather just members of a map - and CDC preimage gives the entire contents of the map when it's modified - not just modified items in the map. We will need a separate test to check the LSI case (and separate issue to report it broken).

### LINKED ISSUE #7030: Alternator Streams: OldImage is missing unmodified LSI/GSI-key attributes
DynamoDB Streams includes the entire old item as the value of OldImage when OLD_IMAGE mode is enabled.
In contrast, in our CDC implementation, the "preimage" only includes the columns touched by the update - not the entire old item.
In most cases, this doesn't cause problems: Usually Alternator's attributes are not kept as separate Scylla columns - they are just elements of a single map column - and CDC includes in the preimage the entire content of that map column, not just the modified elements.  

But there is an exception to this "usually": In a table with LSI (and currently, also GSI), the key columns of the indexes are implemented (in the base table) as real Scylla columns - not as elements in a map. So CDC's "preimage" will be missing the value of these columns when they are not modified. I wrote a test to reproduce this: `test_streams.py::test_streams_updateitem_old_image_lsi`.

To fix this, we will need a new CDC mode which does a "full preimage", i.e., a preimage which includes the entire row, not just the  columns affected by the update.

elcallio:
Changes the "preimage" option from binary true/false to on/off/full (accepting true/false, and using old style notation for normal to string - for upgrade reasons), where "full" will force us to include all columns in pre image log rows. 

Adds small test (just adding the case to preimage test). 
Uses the feature in alternator

Fixes #7030

### DIFF
--- a/cdc/cdc_options.hh
+++ b/cdc/cdc_options.hh
@@ -27,15 +27,31 @@
 
 namespace cdc {
 
-enum class delta_mode {
+enum class delta_mode : uint8_t {
     off,
     keys,
     full,
 };
 
+/**
+ * (for now only pre-) image collection mode.
+ * Stating how much info to record.
+ * off == none
+ * on == changed columns
+ * full == all (changed and unmodified columns)
+ */
+enum class image_mode : uint8_t {
+    off, 
+    on,
+    full,
+};
+
+std::ostream& operator<<(std::ostream& os, delta_mode);
+std::ostream& operator<<(std::ostream& os, image_mode);
+
 class options final {
     bool _enabled = false;
-    bool _preimage = false;
+    image_mode _preimage = image_mode::off;
     bool _postimage = false;
     delta_mode _delta_mode = delta_mode::full;
     int _ttl = 86400; // 24h in seconds
@@ -47,13 +63,15 @@ public:
     sstring to_sstring() const;
 
     bool enabled() const { return _enabled; }
-    bool preimage() const { return _preimage; }
+    bool preimage() const { return _preimage != image_mode::off; }
+    bool full_preimage() const { return _preimage == image_mode::full; }
     bool postimage() const { return _postimage; }
     delta_mode get_delta_mode() const { return _delta_mode; }
     int ttl() const { return _ttl; }
 
     void enabled(bool b) { _enabled = b; }
-    void preimage(bool b) { _preimage = b; }
+    void preimage(bool b) { preimage(b ? image_mode::on : image_mode::off); }
+    void preimage(image_mode m) { _preimage = m; }
     void postimage(bool b) { _postimage = b; }
     void ttl(int v) { _ttl = v; }
 
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -293,9 +293,13 @@ future<> cdc::cdc_service::stop() {
 cdc::cdc_service::~cdc_service() = default;
 
 namespace {
-const sstring delta_mode_string_off  = "off";
-const sstring delta_mode_string_keys = "keys";
-const sstring delta_mode_string_full = "full";
+static const sstring delta_mode_string_off  = "off";
+static const sstring delta_mode_string_keys = "keys";
+static const sstring delta_mode_string_full = "full";
+
+static const std::string_view image_mode_string_on  = "on";
+static const std::string_view image_mode_string_off  = delta_mode_string_off;
+static const std::string_view image_mode_string_full = delta_mode_string_full;
 
 sstring to_string(cdc::delta_mode dm) {
     switch (dm) {
@@ -305,7 +309,25 @@ sstring to_string(cdc::delta_mode dm) {
     }
     throw std::logic_error("Impossible value of cdc::delta_mode");
 }
+
+sstring to_string(cdc::image_mode m) {
+    switch (m) {
+        case cdc::image_mode::off : return "false";
+        case cdc::image_mode::on : return "true";
+        case cdc::image_mode::full : return sstring(image_mode_string_full);
+    }
+    throw std::logic_error("Impossible value of cdc::image_mode");
+}
+
 } // anon. namespace
+
+std::ostream& cdc::operator<<(std::ostream& os, delta_mode m) {
+    return os << to_string(m);
+}
+
+std::ostream& cdc::operator<<(std::ostream& os, image_mode m) {
+    return os << to_string(m);
+}
 
 cdc::options::options(const std::map<sstring, sstring>& map) {
     if (!map.contains("enabled")) {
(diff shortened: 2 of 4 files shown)


haaawk:
This will create conflicts with https://github.com/scylladb/scylla/pull/6978, won't it?

elcallio:
I don't think it will. From what I can read, it does not touch the parts of log.cc that #6978 does.

nyh:
@elcallio you said that this fixes 6935, but does it? The problem in 6935 was that the *key* was missing in the preimage. We don't need a "full" preimage to output the key... Did you add the key to the "full preimage"? Did you want to add it (for CDC, not Alternator)?

I need to write a test for 6935 involving an LSI and its separate regular column, so I can test your patch against that as well. I'll do it.

elcallio:
If by "key" you mean something that is a partition or clustering key, no, this will not help. But if you mean "key" as in "named regular column", it will. To repeat from and add to that issue though: clustering/partition keys are always recorded for preimage. Regular columns are not, only changed ones. Streams will include all regular fields + embedded attributes from an alternator table. 
This patch ensures that any top-level attribute (key?) is included in preimage, even if it was not written to/deleted by the originating op.

nyh:
> If by "key" you mean something that is a partition or clustering key, no, this will not help.

So you can't mark this patch as fixing 6935.
I'll update this issue's first entry to say exactly which test is failing.

> But if you mean "key" as in "named regular column", it will.

I agree it should fix that case. I'll write an Alternator test to confirm this (the test will need to involve an LSI).

> To repeat from and add to that issue though: clustering/partition keys are always recorded for preimage.

So why don't we see them in the Alternator results? You removed them deliberately from the preimage?

> Regular columns are not, only changed ones. Streams will include all regular fields + embedded attributes from an alternator table.
> This patch ensures that any top-level attribute (key?) is included in preimage, even if it was not written to/deleted by the originating op.

Yes. The patch is good (at least in theory, I didn't test the actual code or review it) - I'm just saying it might not alone fix 6935.

elcallio:
Can you for clarity please specify what "key" means here? Are we talking the "Keys" column, which contain the actual table keys (i.e. partition/cluster), or are we talking property key-values?
Again, I only ran a local patch (manually fixed to run) version of the tests you wrote, and frankly, I failed to parse a few of them, so I am unsure what they even wanted to do... (danger or composing subroutines) :-/
Anyway, I can change the "fixes" to "refs" if the problem is deeper that excluded columns.

nyh:
Calle, you did fix a problem - a problem that does not have an issue or a test yet (I'm in the process of creating both and will let you know). 6935, which you supposedly "fixed", is about the actual table key attributes - the partition key and clustering key. Those are missing in the output, this is what 6935 was about, and your patch, while valuable in fixing another bug mentioned in comments down in 6935, doesn't fix the main bug of 6935. I edited now the first report of 6395 to hopefully make it clear what it is about.
Fixing 6935 is about fixing the single test `test_streams.py::test_streams_updateitem_old_image` - it's now xfail and should convert to xpass. But not with your patch. Your patch should fix a different test I'm writing right now :-)

nyh:
I created an issue https://github.com/scylladb/scylla/issues/7030 which this PR should "Fixes", not 6935 (I believe you'll need a different fix for 6935?). That issue mentions a test, which I'll send in a patch in a few minutes.

nyh:
> Again, I only ran a local patch (manually fixed to run) version of the tests you wrote, and frankly, I failed to parse a few of them, so I am unsure what they even wanted to do... (danger or composing subroutines) :-/

By the way, my new tests are already in master, so you don't need to patch or fix to run them. Each test also includes a clear (I hope...) comment about what it is testing. Also I really try to mention the name of a test in every issue which has a reproducing test - and also in reverse (mention issue numbers in comments in the test) so you don't need to guess which test goes with which issue. 

> Anyway, I can change the "fixes" to "refs" if the problem is deeper that excluded columns.

I think it's the opposite - in this patch you solved a more difficult problem than issue 6935 :-) I think you solved https://github.com/scylladb/scylla/issues/7030 which I just created now (previously it was just a comment in the middle of the discussion of 6935).

elcallio:
Commit and PR message updated.

cdc/cdc_options.hh
```diff
@@ -27,15 +27,24 @@
...
+
+std::ostream& operator<<(std::ostream& os, delta_mode);
+std::ostream& operator<<(std::ostream& os, image_mode);
+
 class options final {
     bool _enabled = false;
-    bool _preimage = false;
+    image_mode _preimage = image_mode::off;
     bool _postimage = false;
     delta_mode _delta_mode = delta_mode::full;
```
nyh, line 49: @haaawk @elcallio @kbr- a while back, you talked about the "delta_mode" option not being called "delta" (like I thought it should be called) because it's not a on/off but a an enum. So in this patch, "preimage" also becomes an enum. I wonder if the naming shouldn't be more consistent - i.e., delta/preimage or delta_mode/preimage_mode?

cdc/cdc_options.hh
```diff
@@ -47,13 +56,15 @@ public:
...
-    bool preimage() const { return _preimage; }
+    bool preimage() const { return _preimage != image_mode::off; }
+    bool full_preimage() const { return _preimage == image_mode::full; }
     bool postimage() const { return _postimage; }
     delta_mode get_delta_mode() const { return _delta_mode; }
     int ttl() const { return _ttl; }
 
     void enabled(bool b) { _enabled = b; }
-    void preimage(bool b) { _preimage = b; }
+    void preimage(bool b) { preimage(image_mode::on); }
```
nyh, line 66: This looks wrong :-( Maybe `preimage(b ? image_mode::on : image_mode::off)`?

cdc/log.cc
```diff
@@ -337,7 +376,7 @@ cdc::options::options(const std::map<sstring, sstring>& map) {
         }
     }
 
-    if (_enabled && !_preimage && !_postimage && _delta_mode == delta_mode::off) {
+    if (_enabled && !preimage() && !postimage() && get_delta_mode() == delta_mode::off) {
```
nyh, line 379: Again, for consistency, maybe `get_delta_mode()` should or can be called `delta()`.
Of course, doesn't need to be fixed in this patch.

test/boost/cdc_test.cc
```diff
@@ -517,15 +517,15 @@ SEASTAR_THREAD_TEST_CASE(test_primary_key_logging) {
 
 SEASTAR_THREAD_TEST_CASE(test_pre_post_image_logging) {
     do_with_cql_env_thread([](cql_test_env& e) {
-        auto test = [&e] (bool pre_enabled, bool post_enabled, bool with_ttl) {
+        auto test = [&e] (cdc::image_mode pre_enabled, bool post_enabled, bool with_ttl) {
```
nyh, line 520: Can you please change the patch title? I don't understand what you mean by "minute test" (minute in the sense that it's small? that it runs for a whole minute? :-)), and instead of "full" perhaps say "full preimage"?

cdc/cdc_options.hh
elcallio: I called it `image_mode` because eventually it could maybe be applied to post image as well, i.e. allow having non-full post image. Technically, delta_mode could also use image_mode, but I did not make this change (yet).

cdc/cdc_options.hh
elcallio: Oops. Yes.

cdc/log.cc
elcallio: All for that. I did not name the accessor, so did not change it here.

test/boost/cdc_test.cc
elcallio: minute == small, yes

cdc/cdc_options.hh
elcallio: Sorry, delta_mode cannot use it of course, because `keys`... doh.

elcallio:
Fixed the issues pointed out by @nyh

cdc/cdc_options.hh
nyh: I'll commit your patch with these names, but I ask you (and @haaawk) to take another look at this issue later. My comment not about the name of the type - it's about the name of the variable and user-facing configuration, which previously you called "delta_mode" because you said "delta" would have implied true/false only. But now also "preimage" can be more than true/false, so why can't we call the delta option just "delta" instead of "delta_mode"? It's not a disaster, but it's just inconsistent.

nyh:
I confirmed that this indeed fixes issue #7030, but as expected, doesn't solve issue #6935.
@haaawk @kbr- I want to commit this - unless you have any objections to the implementation.

nyh:
Merged to next: ee7291aa882d925d3de3dd3db14df443db020034

haaawk:
@nyh It would be good to wait for someone to review the cdc part of this before merging.

cdc/log.cc
```diff
@@ -293,9 +293,13 @@ future<> cdc::cdc_service::stop() {
...
-const sstring delta_mode_string_off  = "off";
-const sstring delta_mode_string_keys = "keys";
-const sstring delta_mode_string_full = "full";
+static const sstring delta_mode_string_off  = "off";
+static const sstring delta_mode_string_keys = "keys";
+static const sstring delta_mode_string_full = "full";
+
+static const std::string_view image_mode_string_on  = "on";
+static const std::string_view image_mode_string_off  = delta_mode_string_off;
+static const std::string_view image_mode_string_full = delta_mode_string_full;
```
haaawk, line 302: This is anonymous namespace, I don't think static is needed.

aleksbykov:
@haaawk was it backported to 4.3?

haaawk:
It seems so @aleksbykov

PR #7029, opened 2020-08-12
test/redis: make redis tests runnable from test.py

### LINKED ISSUE #6331: Integrate Redis tests with test.py
Let's do what Alternator did in commit https://github.com/scylladb/scylla/commit/4e2bf28b8455527237baabe629f01c775fe4e48c, and move the tests to `test/redis` and integrate them with `test.py` so we will run the Redis tests as part of our CI.

syuu1228:
Just like test/alternator, make redis-test runnable from test.py.

Fixes #6331


test/redis/run
```diff
@@ -0,0 +1,133 @@
...
+then
+    echo "Error: python3 and python3-redis must be installed to configure redis." >&2
+    exit 1
+fi
+
+
+SCYLLA_IP=127.0.0.1
+echo "Running Scylla on $SCYLLA_IP"
+
+REDIS_PORT=6379
```
avikivity, line 24: We will need to randomize the redis port in case we run with --repeat N (each scylla instance will need a different port).

nyh: Please take a look at test/alternator/run. This is already quite mature and battle-tested with parallel runs, and you don't need to have to write your own.
In particular, each run of Scylla uses a random (actually, based on the process id!) IP address. The ports can be the same. The beauty of this is that on Linux, actually localhost is 127/8 (i.e., 127.*.*.*), not just 127.0.0.1. We can use *any* of these millions of IP addresses.

nyh: I see you actually copied most of the alternator/run code, just not the SCYLLA_IP line, why did you drop this? Here is the original code:
```python
# Pick a loopback IP address for Scylla to run, in an attempt not to collide
# other concurrent runs of Scylla. CCM uses 127.0.0.<nodenum>, so if we use
# 127.1.*.* which cannot collide with it. Moreover, we'll take the last two
# bytes of the address from the current process - so as to allow multiple
# concurrent runs of this code to use a different address.
SCYLLA_IP=127.1.$(($$ >> 8 & 255)).$(($$ & 255))
```

syuu1228: @avikivity @nyh Why I dropped these lines was I could not succeeded to pass parameters from run script to testcase python scripts, but now I figure out how to do that so reverted the line.

Please review again.

PR #7034, opened 2020-08-13
Storage Proxy: Maintain metrics in response handler classes

### LINKED ISSUE #5797: some background_read counters are underflow
*Installation details*
Scylla version (or git commit hash): a8a4e584ec3332067b6b9e1756ba1bf2392e1424
Cluster size:

`scylla_storage_proxy_coordinator_background_reads{scheduling_group_name="main",shard="0",type="queue_length"} 18446744073709551616.000000`

caused by 971711a5461c09421f1c9b1b3b2d9f6209a425da

eliransin:
There is a known problem introduced by 971711a, the problem
is that when some callbacks for a created response handler
are called from another scheduling groups an imbalance is
introduced to statistics which are now instantiated per
scheduling group. This can be easily solved by "remembering"
the stats of the scheduling group where the response handler
was created. This will mask out another problem where certain
operations are done outside of the "native" scheduling group.
This is a complex issue that needs to be addressed in separate
commit (see 971711a commit message).

Fixes #5797

### DIFF
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -846,9 +846,9 @@ paxos_response_handler::begin_and_repair_paxos(client_state& cs, unsigned& conte
                     paxos::paxos_state::logger.debug("CAS[{}] Finishing incomplete paxos round {}", _id, *in_progress);
                     tracing::trace(tr_state, "Finishing incomplete paxos round {}", *in_progress);
                     if (is_write) {
-                        ++_proxy->get_stats().cas_write_unfinished_commit;
+                        ++_stats.cas_write_unfinished_commit;
                     } else {
-                        ++_proxy->get_stats().cas_read_unfinished_commit;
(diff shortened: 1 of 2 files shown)


eliransin:
An older version of the patch have already been reviewed by @gleb-cloudius and @psarna , the patch title in the mailing list is:
`[PATCH] Storage Proxy: Maintain metrics in response handler classes`
The main change since the older patch is the introduction of paxos_executor and paxos stats to the structure.

slivne:
@eliransin / @psarna / @gleb-cloudius is this till relevant if so lets review / merge

eliransin:
We will need this fix.
@gleb-cloudius can you review?

gleb-cloudius (commented):
That unfortunately enlarges each response handler by 8 bytes. Can we do better? For instance remembering scheduling group id instead of the reference? Also can a scheduling group disappear?

eliransin:
This was done already

Issue #7051, opened 2020-08-17
Hinted handoff is using very long timeout to sending some hints


piodul:
A hint that is stored on disk is associated with a target endpoint. When a hint is considered to be sent, it is usually sent directly to the target endpoint.

However, the topology might have changed since the time the hint was written, and the target endpoint is no longer a replica for the hint. In such case, the hint is sent with CL=ALL to all known current replicas.

Both cases are handled with slightly different code paths in storage_proxy. The most important difference is that in the second case a very long timeout is being used (1 hour, [link](https://github.com/scylladb/scylla/blob/75c2362c955e3491a7c08f1bc178b73e7fe89d38/service/storage_proxy.cc#L2594)). Because hints manager waits for all in-progress hint send operations before shutting down, it can cause the shutdown to last very long:

![linger](https://user-images.githubusercontent.com/4950936/90389833-8930b200-e08a-11ea-90fd-b0c2234af64c.png)

Link to a relevant discussion: https://github.com/scylladb/scylla/pull/5668#discussion_r372315184

haaawk:
Good catch @piodul

avikivity:
Will let it stew for a few more days, then backport.

avikivity:
Will let it stew some more.

avikivity:
Backported to 4.2 (stewed for too long).

PR #7053, opened 2020-08-17
hinted handoff: use default timeout for sending orphaned hints

piodul:
This patch causes orphaned hints (hints that were written towards a node
that is no longer their replica) to be sent with a default write
timeout. This is what is currently done for non-orphaned hints.

Previously, the timeout was hardcoded to one hour. This could cause a
long delay while shutting down, as hints manager waits until all ongoing
hint sending operation finish before stopping itself.

Fixes: #7051

### DIFF
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -2591,14 +2591,13 @@ future<> storage_proxy::send_hint_to_endpoint(frozen_mutation_and_schema fm_a_s,
 }
 
 future<> storage_proxy::send_hint_to_all_replicas(frozen_mutation_and_schema fm_a_s) {
-    const auto timeout = db::timeout_clock::now() + 1h;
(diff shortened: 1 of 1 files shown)


piodul:
@gleb-cloudius @vladzcloudius We discussed the problem of the long timeout here: https://github.com/scylladb/scylla/pull/5668#discussion_r372315184. As far as I understood, we were in an agreement that changing this timeout to the default write timeout is not a bad idea, so I went with this solution.

Non-orphaned hints already use the default write timeout, and I don't recall if we had any issues with it.

piodul:
Oops, I just noticed that I accidentally used a wrong, old branch name for this PR (the commits and the description is as intended). GitHub does not seem to allow changing it, so if it is a problem I can create another PR with the correct branch name.

avikivity:
@piodul the branch name is not important.

PR #7063, opened 2020-08-18
alternator_streams: Include keys in OldImage/NewImage

### LINKED ISSUE #6935: Alternator Streams: OldImage is missing the old item's key

elcallio:
Fixes #6935

DynamoDB streams for some reason duplicate the record keys
into both the "Keys" and "OldImage"/"NewImage" sub-objects
when doing GetRecords. But only if there is other data
to include.

This patch appends the pk/ck parts into old/new image
iff we had any record data.

Updated to handle keys-only updates, and distinguish creating vs. updating rows. Changes cdc to not generate preimage 
for non-existent/deleted rows, and also fixes missing operations/ttls in keys-only delta mode.

### DIFF
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -840,9 +840,8 @@ future<executor::request_return_type> executor::get_records(client_state& client
             {
                 auto item = rjson::empty_object();
                 describe_single_item(*selection, row, attr_names, item, true);
-                if (!item.ObjectEmpty()) {
-                    rjson::set(dynamodb, op == cdc::operation::pre_image ? "OldImage" : "NewImage", std::move(item));
-                }
+                describe_single_item(*selection, row, key_names, item);
+                rjson::set(dynamodb, op == cdc::operation::pre_image ? "OldImage" : "NewImage", std::move(item));
                 break;
             }
             case cdc::operation::update:
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -1504,11 +1504,6 @@ public:
 
         assert(_builder);
 
-        auto image_ck = _builder->allocate_new_log_row(op);
-        if (ck) {
-            _builder->set_clustering_columns(image_ck, *ck);
-        }
-
         const auto kind = ck ? column_kind::regular_column : column_kind::static_column;
 
         cell_map* row_state;
@@ -1516,10 +1511,25 @@ public:
             row_state = get_row_state(_clustering_row_states, *ck);
             if (!row_state) {
                 // We have no data for this row, we can stop here
+                // Empty row here means we had data, but column deletes
+                // removed it. This we should report as pk/ck only
                 return;
             }
         } else {
             row_state = &_static_row_state;
+            // if the static row state is empty and we did not touch 
+            // during processing, it either did not exist, or we did pk delete. 
+            // In those cases, no postimage for you.
+            // If we touched it, we should indicate it with an empty 
+            // postimage. 
+            if (row_state->empty() && !_touched_parts.contains(stats::part_type::STATIC_ROW)) {
+                return;
+            }
+        }
+
+        auto image_ck = _builder->allocate_new_log_row(op);
+        if (ck) {
+            _builder->set_clustering_columns(image_ck, *ck);
         }
 
         auto process_cell = [&, this] (const column_definition& cdef) {
--- a/test/alternator/test_streams.py
+++ b/test/alternator/test_streams.py
@@ -677,8 +677,7 @@ def test_streams_updateitem_keys_only_2(test_table_ss_keys_only, dynamodbstreams
 
 # Test OLD_IMAGE using UpdateItem. Verify that the OLD_IMAGE indeed includes,
 # as needed, the entire old item and not just the modified columns.
-# Currently fails in Alternator because the item's key is missing in OldImage (#6935).
-@pytest.mark.xfail(reason="Currently fails - see issue #6935")
+# Reproduces issue #6935
 def test_streams_updateitem_old_image(test_table_ss_old_image, dynamodbstreams):
     def do_updates(table, p, c):
         events = []
@@ -703,7 +702,6 @@ def test_streams_updateitem_old_image(test_table_ss_old_image, dynamodbstreams):
 # key - in this case since the item did exist, OLD_IMAGE should be returned -
 # and include just the key. This is a special case of reproducing #6935 -
 # the first patch for this issue failed in this special case.
-@pytest.mark.xfail(reason="Currently fails - see issue #6935")
 def test_streams_updateitem_old_image_empty_item(test_table_ss_old_image, dynamodbstreams):
     def do_updates(table, p, c):
         events = []
@@ -751,7 +749,6 @@ def test_table_ss_old_image_and_lsi(dynamodb, dynamodbstreams):
     yield table, arn
     table.delete()
 
-@pytest.mark.xfail(reason="Currently fails - see issue #6935, #7030")
 def test_streams_updateitem_old_image_lsi(test_table_ss_old_image_and_lsi, dynamodbstreams):
     def do_updates(table, p, c):
         events = []
@@ -774,7 +771,6 @@ def test_streams_updateitem_old_image_lsi(test_table_ss_old_image_and_lsi, dynam
 # that deleting the item results in a missing NEW_IMAGE, and that setting the
 # item to be empty has a different result - a NEW_IMAGE with just a key.
 # Reproduces issue #7107.
-@pytest.mark.xfail(reason="issue #7107")
 def test_streams_new_image(test_table_ss_new_image, dynamodbstreams):
     def do_updates(table, p, c):
         events = []
--- a/test/boost/cdc_test.cc
+++ b/test/boost/cdc_test.cc
@@ -525,7 +525,7 @@ SEASTAR_THREAD_TEST_CASE(test_pre_post_image_logging) {
 
             auto rows = select_log(e, "tbl");
 
-            BOOST_REQUIRE_EQUAL(pre_enabled == cdc::image_mode::off, to_bytes_filtered(*rows, cdc::operation::pre_image).empty());
+            BOOST_REQUIRE(to_bytes_filtered(*rows, cdc::operation::pre_image).empty());
             BOOST_REQUIRE_EQUAL(!post_enabled, to_bytes_filtered(*rows, cdc::operation::post_image).empty());
 
             auto first = to_bytes_filtered(*rows, cdc::operation::update);
@@ -564,7 +564,7 @@ SEASTAR_THREAD_TEST_CASE(test_pre_post_image_logging) {
                 sort_by_time(*rows, post_image);
 
                 if (pre_enabled != cdc::image_mode::off) {
-                    BOOST_REQUIRE_EQUAL(pre_image.size(), i + 2);
+                    BOOST_REQUIRE_EQUAL(pre_image.size(), i + 1);
 
                     val = *pre_image.back()[val_index];
                     // note: no val2 in pre-image, because we are not modifying it. 
@@ -620,7 +620,7 @@ SEASTAR_THREAD_TEST_CASE(test_pre_post_image_logging_static_row) {
 
             auto rows = select_log(e, "tbl");
 
-            BOOST_REQUIRE(to_bytes_filtered(*rows, cdc::operation::pre_image).empty() == !enabled);
+            BOOST_REQUIRE(to_bytes_filtered(*rows, cdc::operation::pre_image).empty());
             BOOST_REQUIRE(to_bytes_filtered(*rows, cdc::operation::post_image).empty() == !enabled);
 
             auto first = to_bytes_filtered(*rows, cdc::operation::update);
@@ -656,7 +656,7 @@ SEASTAR_THREAD_TEST_CASE(test_pre_post_image_logging_static_row) {
                     sort_by_time(*rows, second);
                     sort_by_time(*rows, pre_image);
                     sort_by_time(*rows, post_image);
-                    BOOST_REQUIRE_EQUAL(pre_image.size(), i + 2);
+                    BOOST_REQUIRE_EQUAL(pre_image.size(), i + 1);
 
                     s = *pre_image.back()[s_index];
                     BOOST_REQUIRE_EQUAL(data_value(last), s_type->deserialize(bytes_view(s)));
@@ -1603,8 +1603,6 @@ SEASTAR_THREAD_TEST_CASE(test_batch_with_row_delete) {
         auto set_null = data_value::make_null(s_type);
 
         const std::vector<std::vector<data_value>> expected = {
-            // Preimage for (0)
-            {int_null, udt_null, map_null, set_null, oper_ut(cdc::operation::pre_image)},
             // Update (0)
             {int32_t(1), make_user_value(udt_type, {1,2}), make_map_value(m_type, {{1,2},{3,4}}), make_set_value(s_type, {1,2,3}), oper_ut(cdc::operation::insert)},
             // Preimage for (1)
@@ -1780,10 +1778,6 @@ void test_batch_images(bool preimage, bool postimage) {
                 {"ck", "v1"},
                 {
                     {
-                        .preimage = {
-                            {int32_t(1), int_null},
-                            {int32_t(2), int_null}
-                        },
                         .postimage = {
                             {int32_t(1), int32_t(10)},
                             {int32_t(2), int32_t(20)}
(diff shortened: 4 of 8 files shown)


test/alternator/test_streams.py
```diff
@@ -678,7 +678,6 @@ def do_updates(table, p, c):
 # Test OLD_IMAGE using UpdateItem. Verify that the OLD_IMAGE indeed includes,
 # as needed, the entire old item and not just the modified columns.
 # Currently fails in Alternator because the item's key is missing in OldImage (#6935).
```
nyh, line 680: Please replace this line ("currently fails ...") with a statement that this reproduces issue #6935.

alternator/streams.cc
```diff
@@ -841,6 +841,7 @@ future<executor::request_return_type> executor::get_records(client_state& client
                 auto item = rjson::empty_object();
                 describe_single_item(*selection, row, attr_names, item, true);
                 if (!item.ObjectEmpty()) {
+                    describe_single_item(*selection, row, key_names, item);
```
nyh, line 844: You mentioned in the commit message - and also in code here - that you only need to insert the key if there is anything else in the item. But it's possible there is nothing in the item besides the key, and in this case, isn't item an empty object to which you **should** add the key? Did you check this case as well? I'll write a test for it momentarily so I can check your patch myself.

test/alternator/test_streams.py
```diff
@@ -729,7 +728,6 @@ def test_table_ss_old_image_and_lsi(dynamodb, dynamodbstreams):
     yield table, arn
     table.delete()
 
-@pytest.mark.xfail(reason="Currently fails - see issue #6935, #7030")
```
nyh, line 732: Good, but please also remove the two line comment before the  test_table_ss_old_image_and_lsi() fixture above saying "Currently fails in Alternator because the item's key is missing in OldImage (#6935) and the LSI key is also missing (#7030)." as both of these are already fixed.

nyh (changes requested):
I left some request for changing comments, but I'm most worried about the empty item case - I'll write a test momentarily to check that and report back my finding.

nyh:
As I worried, your patch fails for an OLD_IMAGE of an empty item (i.e., an item having only a key and no other columns), as the following test demonstrates. I'll send a patch with this test to the mailing list.

```python
# Above we verified that if an item did not previously exist, the OLD_IMAGE
# should be missing, but if the item did previously exist, OLD_IMAGE should
# be present and include the key. Here we confirm the special case of an
# *empty* item, which has just the key - in this case too the item did exist
# so OLD_IMAGE should be returned - and include just the key.
# Part of reproducing #6935.
def test_streams_updateitem_old_image_empty_item(test_table_ss_old_image, dynamodbstreams):
    def do_updates(table, p, c):
        events = []
        # Create an *empty* item, with nothing except a key:
        table.update_item(Key={'p': p, 'c': c})
        events.append(['*', {'p': p, 'c': c}, None, {'p': p, 'c': c}])
        table.update_item(Key={'p': p, 'c': c},
            UpdateExpression='SET y = :val1', ExpressionAttributeValues={':val1': 3})
        events.append(['MODIFY', {'p': p, 'c': c}, {'p': p, 'c': c}, {'p': p, 'c': c, 'y': 3}])
        return events
    do_test(test_table_ss_old_image, dynamodbstreams, do_updates, 'OLD_IMAGE')
```

nyh:
@elcallio sent a patch to the mailing list titled "alternator test: test for OLD_IMAGE of an empty item" with my new test (which as you can see passes on DynamoDB, and fails with your patch). Please review my test to encourage having it committed, and fix your code's logic around the empty vs missing item to make this test pass.

If there is some inherent reason why this new test cannot succeed - perhaps some bug in CDC which prevents telling the difference between an empty item and a missing one - we'll need to split https://github.com/scylladb/scylla/issues/6935 into two - closing 6935 but opening a new issue about the remaining bug on the empty item. But please let me know if this is the case (I hope it isn't).

test/alternator/test_streams.py
elcallio: fixed

test/alternator/test_streams.py
elcallio: fixed.

alternator/streams.cc
elcallio: Handling all cases turns out requires fixing cdc to not generate preimages for non-existent rows. The generation of which is weird that it came to be in the first place.

nyh:
@elcallio the first patch in this series seems to fix the same problem that https://github.com/scylladb/scylla/pull/7096 - which I committed a few minutes ago - fixes. Please confirm that it really fixes the same bug (the test changes of the two fixes appear not be the same!) and if so, drop the first patch from this series, or make it fix whatever still needs fixing after https://github.com/scylladb/scylla/pull/7096. Thanks.

elcallio:
I'll move it to a separate change set, as the only thing really needed for the series is for cdc$op et all to show up, and that is mainly because of my OCD and the `result` files not making sense otherwise. :-/

elcallio:
...once stuff goes through next of course... zzzz...

cdc/log.cc
```diff
@@ -1463,22 +1463,25 @@ class transformer final : public change_processor {
...
-            _builder->set_clustering_columns(image_ck, *ck);
-        }
-
         const auto kind = ck ? column_kind::regular_column : column_kind::static_column;
 
         cell_map* row_state;
         if (ck) {
             row_state = get_row_state(_clustering_row_states, *ck);
-            if (!row_state) {
+            if (!row_state && op == cdc::operation::pre_image) {
```
nyh, line 1471: Why is this specific to pre_image? What about post_image? It can also be missing because the operation was to delete a row, in which case there is no post_image, no?
(the Alternator tests for that is missing, by the way, I still need to write them).

cdc/log.cc
```diff
@@ -1463,22 +1463,25 @@ class transformer final : public change_processor {
...
                 return;
             }
         } else {
             row_state = &_static_row_state;
+            if (row_state->empty()) {
+                return;
+            }
+        }
+
+        auto image_ck = _builder->allocate_new_log_row(op);
```
nyh, line 1482: Why is this line not inside the if()? It's not used outside the if, is it?
Or maybe you did this way to ensure it lives a long enough time? But is is a long enough time? I'm afraid I am not familiar with this code enough to understand.

alternator/streams.cc
```diff
@@ -840,6 +840,9 @@ future<executor::request_return_type> executor::get_records(client_state& client
             {
                 auto item = rjson::empty_object();
                 describe_single_item(*selection, row, attr_names, item, true);
+                if (!item.ObjectEmpty() || op == cdc::operation::pre_image) {
```
nyh, line 843: I don't understand, why do you need the "!item.ObjectEmpty()" part of this if? Didn't you just want to do this, unconditionally, for the pre_image?
Also, I think you maybe need to check || op == cdc::operation::post_image as well... Alternator's NEW_IMAGE also needs to include the key. As I mentioned above, the Alternator tests for this are missing (I'll do that) but I'm worried it will exhibit the same bug as the preimage.

test/alternator/test_streams.py
```diff
@@ -677,8 +677,6 @@ def do_updates(table, p, c):
 
 # Test OLD_IMAGE using UpdateItem. Verify that the OLD_IMAGE indeed includes,
 # as needed, the entire old item and not just the modified columns.
-# Currently fails in Alternator because the item's key is missing in OldImage (#6935).
```
nyh, line 680: Good you removed this comment, but please replace it by a comment like "Reproduces issue #6935".

nyh (changes requested):
Thanks! looks good except the first patch (which conflicts with https://github.com/scylladb/scylla/pull/7096) and additionally, I left comments where I was worried that some of the changes you did only for preimage also need to be done for the postimage.
@haaawk please have somebody from the CDC team review the second patch, which modifies CDC proper.

cdc/log.cc
elcallio: We always generate post image, for better or for worse. Though technically it might be more reasonable to _not_ do this. Right now, this is preserving the behaviour as-is except for pre-image where I need it to change. I think follow-up...

cdc/log.cc
elcallio: ? The code path is used from two true-branches (i.e. both cases where we _don't shortcircuit image generation).

alternator/streams.cc
elcallio: NEW_IMAGE will include it iff there is preimage data. I.e. if we did in fact to a delete, and no fields exist, there will be no NEW_IMAGE, neither keys nor fields. For OLD_IMAGE, we include it as long as there is one, because maybe keys-only.

Dropping postimage  for deleted rows would let us always include the keys, but as it is, the above two conditions (see the second `!item.ObjectEmpty()` below).

test/alternator/test_streams.py
elcallio: sure

cdc/log.cc
nyh: I meant, as far as I can tell (maybe I don't see the entire code so I'm missing something), this "image_ck" variable is only used inside the if() below, so why isn't it defined inside the if?

cdc/log.cc
nyh: Ok, I'll accept this. If there is also a similar bug in the postimage, we'll discover it later when I write the test, and fix it later.

cdc/log.cc
elcallio: The var is used in lambda capture three lines below. It is at the right scope.

alternator/streams.cc
nyh: I'm withdrawing my question about post_image, we can address these problem later when I write tests for them.
But I'm still puzzled about the "!item.ObjectEmpty()" thing. Apparently it's not needed for solving the pre_image-for-the-empty-but-existing-item bug, because the "op == pre_image" solves it. So what is it for?

elcallio: We add the keys if either:
  Any other value was included (basically NEW_IMAGE case, distinguishing row delete from something that has values)
or
  It is OLD_IMAGE, since pre_image is only generated iff there acually exists what we call OLD_IMAGE (previous state). 

Then there is a check below if `!item.ObjectEmpty()` to determine if we should in fact add OLD/NEW image at all, i.e. if we had either keys and/or data.

test/alternator/test_streams.py
elcallio: Fixed

alternator/streams.cc
elcallio: Post image generation fix is included in the change set. Thus the checks can be removed.

cdc/log.cc
elcallio: Removed the check. Postimage is not consistent with pre-image, i.e. none if the row does not exist.

test/alternator/test_streams.py
```diff
@@ -769,11 +766,74 @@ def do_updates(table, p, c):
         return events
     do_test(test_table_ss_old_image_and_lsi, dynamodbstreams, do_updates, 'OLD_IMAGE')
 
-# TODO: test NEW_IMAGE and NEW_AND_OLD_IMAGES, similar to OLD_IMAGE above.
-# In fact, it can be exactly the same code, sharing one do_update() function.
-# Just make sure to keep three separate tests on the three separate tables -
-# testing NEW_AND_OLD_IMAGES is not enough - OLD_IMAGE might have a bug,
-# e.g., a missing column, that isn't missing in NEW_AND_OLD_IMAGES.
+# Tests similar to the above tests for OLD_IMAGE, just for NEW_IMAGE mode.
```
nyh, line 769: Avi already committed my patch adding these tests to next. We'll need to either wait for master to contain it and rebase over master - or if you want this in quickly, please rebase over next now.
In either case, don't forget after the rebase to delete the "xfail" I added in my patch.

test/alternator/test_streams.py
```diff
@@ -769,11 +766,74 @@ def do_updates(table, p, c):
...
+# Verify that the NEW_IMAGE includes the entire old item (including the key),
+# that deleting the item results in a missing NEW_IMAGE, and that setting the
+# item to be empty has a different result - a NEW_IMAGE with just a key.
+# Reproduces issue #7107.
+#@pytest.mark.xfail(reason="issue #7107")
```
nyh, line 774: Don't forget to remove this xfail in your rebased patch

nyh (changes requested):
Excellent, looks very good, but needs a rebase to remove that tests which I already added.
@kbr- or @haaawk please review that you're ok with the CDC part (it seemed on today's status call that you are, but please verify).

kbr-:
So we're not showing postimage for a row which got row-deleted.
Note that we're also not showing it even if that row got updated in this batch (i.e. it got both updated and then deleted).
I guess I'm fine with that, if everyone else is. @haaawk 

So LGTM

nyh:
> So we're not showing postimage for a row which got row-deleted.
> Note that we're also not showing it even if that row got updated in this batch (i.e. it got both updated and then deleted).

A "batch" means the same timestamp? In that case, a delete overrides an update of the same timestamp, so it is correct that nothing should remain. I'm not even sure if we see both in the code (wouldn't the reader drop the update if it sees a delete with the same timestamp?).

kbr-:
The mutation would contain both the updated cell and a row delete (it's kind of weird -- we could compact the cell away -- but that's how it is) and we'll process them in order.

But I agree that it makes more sense to not show it; it leads to a simple to understand rule, i.e. if the row got deleted (using either a row delete, a range delete, or a partition delete), it doesn't get a postimage.

nyh:
> it leads to a simple to understand rule, i.e. if the row got deleted (using either a row delete, a range delete, or a partition delete), it doesn't get a postimage.

Just don't forget my comment that there remains one way where a row can get deleted and still get a post-image: If a row **without** a row marker gets its last cell deleted, the row is deleted as far as Scylla is concerned, but you will generate an empty post-image. Whether this is acceptable or not, I don't know, but you should at least be aware that this is the case. I doesn't matter for Alternator because in Alternator we always have row markers.

elcallio:
rebased.

cdc/log.cc
```diff
@@ -1504,22 +1504,27 @@ class transformer final : public change_processor {
...
             row_state = get_row_state(_clustering_row_states, *ck);
             if (!row_state) {
                 // We have no data for this row, we can stop here
+                // Empty row here means we had data, but column deletes
+                // removed it. This we should report as pk/ck only
                 return;
             }
         } else {
             row_state = &_static_row_state;
+            if (row_state->empty()) {
```
haaawk, line 1520: I understand the difference between row_state == nullptr and row_state->empty() == true for clustering rows. First means there was no row or the row was deleted. Second means the row has only empty cells. Right?

What's the story with static row? Doesn't row_state->empty() == true mean removal of it's all columns but not removal of the static row itself?

elcallio: Static row is only really "removed" on PK delete. But we don't really make any distinction on whether static row actually exists (on this level), or if it was emptied. 
The above is probably not really correct. We should maybe add a state var to track if we created/processed static row at all.

elcallio: changed this to check iff we touched static row during processing, and generate empty postimage if so, and none otherwise (if the row state is empty). This should be consistent with "normal" clustered row state.

haaawk (approved):
LGTM

nyh (approved):
Thanks. I'll test and hopefully commit. I think you forgot to un-xfail one of the alternator tests, if you did I'll send a followup patch to do that.

nyh:
Thanks. Committed to next - 95afadfe21caee40ca9636266be87c1834ce8119
Indeed there is one additional test which needs to be un-xfailed. I'll do this as a separate followup patch.

nyh:
Sent to the mailing list a follow-up patch titled "alternator test: enable a passing test" to un-xfail the additional passing test.

Issue #7073, opened 2020-08-20
killed_wiped_node_cannot_join_test failed


asias:
Scylla version 666.development-0.20200817.9aa47122

```
Error Message

18 Aug 2020 10:30:43 [node4] Missing: ['A node with address 127.0.76.4 already exists, cancelling join']:
Scylla version 666.development-0.20200817.9aa47122.....
See system.log for remainder

Stacktrace

Traceback (most recent call last):
  File "/usr/lib64/python3.7/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib64/python3.7/unittest/case.py", line 645, in run
    testMethod()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/bootstrap_test.py", line 390, in killed_wiped_node_cannot_join_test
    self._wiped_node_cannot_join_test(gently=False)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/bootstrap_test.py", line 436, in _wiped_node_cannot_join_test
    node2.watch_log_for(expected_error, from_mark=mark)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/node.py", line 448, in watch_log_for
    [e.pattern for e in tofind]) + ":\n" + reads[:50] + ".....\nSee {} for remainder".format(filename))
ccmlib.node.TimeoutError: 18 Aug 2020 10:30:43 [node4] Missing: ['A node with address 127.0.76.4 already exists, cancelling join']:
Scylla version 666.development-0.20200817.9aa47122.....
See system.log for remainder
```

asias:
Fixed by: https://github.com/scylladb/scylla/pull/7074

avikivity:
@asias what's your opinion about backporting this? How risky is it? And is the bug likely to happen outside a test scenario?

avikivity:
Fix present on all active branches, not backporting.

PR #7074, opened 2020-08-20
gossip: Apply state for local node in shadow round

asias:
We saw errors in killed_wiped_node_cannot_join_test dtest:

  Aug 2020 10:30:43 [node4] Missing: ['A node with address 127.0.76.4 already exists, cancelling join']:

The test does:
  n1, n2, n3, n4
  wipe data on n4
  start n4 again with the same ip address

Without this patch, n4 will bootstrap into the cluster new tokens. We
should prevent n4 to bootstrap because there is an existing
node in the cluster.

In shadow round, the local node should apply the application state of
the node with the same ip address. This is useful to detect a node
trying to bootstrap with the same IP address of an existing node.

Tests: bootstrap_test.py
Fixes: #7073

### DIFF
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -656,9 +656,6 @@ void gossiper::apply_state_locally_without_listener_notification(std::unordered_
     for (auto& x : map) {
         const inet_address& node = x.first;
         const endpoint_state& remote_state = x.second;
-        if (node == this->get_broadcast_address()) {
-            continue;
-        }
         do_apply_state_locally(node, remote_state, false);
     }
 }


gms/gossiper.cc
```diff
@@ -656,9 +656,6 @@ void gossiper::apply_state_locally_without_listener_notification(std::unordered_
     for (auto& x : map) {
         const inet_address& node = x.first;
         const endpoint_state& remote_state = x.second;
-        if (node == this->get_broadcast_address()) {
-            continue;
-        }
```
avikivity, line 661: How does this detect the problem? Will it compare the node uuid and throw?

asias: It is checked in storage_service::check_for_endpoint_collision which calls gossiper::is_safe_for_bootstrap.  

We allow to bootstrap a new node in only two cases:

1) The node is a completely new node and no state in gossip at all
2) The node has state in gossip and it is already removed from the
cluster either by nodetool decommission or nodetool removenode

avikivity: But check_for_endpoint_collision() is only called from prepare_do_join(). How is apply_state_locally() related?

asias: Note apply_state_locally_without_listener_notification() is related, not apply_state_locally. The flow is gossip send shadow round message, apply_state_locally_without_listener_notification is called to apply states sent back from the response.

avikivity: Yes, but I don't see how it's related to check_for_endpoints_collision().

asias: I am not sure what is the "it" here. 

We allow applying states from the shadow round replies with the ip address of this local node to detect node with the ip address already exists in the cluster. check_for_endpoints_collision does the actual check. 

Oh, perhaps I see what confuses you now. check_for_endpoint_collision is only used for bootstrap collision check, not for other checks, e.g, restart a node.

avikivity: Yes. So where does this change make a difference?

asias: Without this patch, the gossip state of the node with the same ip address will be ignored. The node will always think there is no existing node of this ip address in the cluster.

avikivity: Ok, I'm not sure I understood 100% but will proceed.

PR #7080, opened 2020-08-20
Gce image support

bentsi:
- Add necessary changes to `scylla_util.py` in order to support Scylla Machine Image in GCE.
- Fixes and improvements for `curl` function.

### DIFF
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -92,8 +92,7 @@ def scyllabindir():
 
 
 # @param headers dict of k:v
-def curl(url, headers=None, byte=False, timeout=3):
-    max_retries = 5
+def curl(url, headers=None, byte=False, timeout=3, max_retries=5):
     retries = 0
     while True:
         try:
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -92,13 +92,13 @@ def scyllabindir():
 
 
 # @param headers dict of k:v
-def curl(url, headers=None, byte=False):
+def curl(url, headers=None, byte=False, timeout=3):
     max_retries = 5
     retries = 0
     while True:
         try:
             req = urllib.request.Request(url, headers=headers or {})
-            with urllib.request.urlopen(req) as res:
+            with urllib.request.urlopen(req, timeout=timeout) as res:
                 if byte:
                     return res.read()
                 else:
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -104,10 +104,10 @@ def curl(url, headers=None, byte=False):
                 else:
                     return res.read().decode('utf-8')
         except urllib.error.HTTPError:
-            logging.warn("Failed to grab %s..." % url)
+            logging.warning("Failed to grab %s..." % url)
             time.sleep(5)
             retries += 1
-            if (retries >= max_retries):
+            if retries >= max_retries:
                 raise
 
 class gcp_instance:
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -90,13 +90,14 @@ def datadir():
 def scyllabindir():
     return str(scyllabindir_p())
 
+
 # @param headers dict of k:v
-def curl(url, byte=False, headers={}):
+def curl(url, headers=None, byte=False):
     max_retries = 5
     retries = 0
     while True:
         try:
-            req = urllib.request.Request(url,headers=headers)
+            req = urllib.request.Request(url, headers=headers or {})
             with urllib.request.urlopen(req) as res:
                 if byte:
                     return res.read()
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -109,11 +109,15 @@ def curl(url, headers=None, byte=False, timeout=3, max_retries=5):
             if retries >= max_retries:
                 raise
 
+
 class gcp_instance:
     """Describe several aspects of the current GCP instance"""
 
     EPHEMERAL = "ephemeral"
     ROOT = "root"
+    GETTING_STARTED_URL = "http://www.scylladb.com/doc/getting-started-google/"
+    META_DATA_BASE_URL = "http://metadata.google.internal/computeMetadata/v1/instance/"
+    ENDPOINT_SNITCH = "GoogleCloudSnitch"
 
     def __init__(self):
         self.__type = None
@@ -139,10 +143,9 @@ class gcp_instance:
                     return True
         return False
 
-    def __instance_metadata(self, path):
-        """query GCP metadata server, recursively!"""
-        return curl("http://169.254.169.254/computeMetadata/v1/instance/" + path+"?recursive=true", headers={"Metadata-Flavor": "Google"})
-        #169.254.169.254 is metadata.google.internal
+    def __instance_metadata(self, path, recursive=False):
+        return curl(self.META_DATA_BASE_URL + path + "?recursive=%s" % str(recursive).lower(),
+                    headers={"Metadata-Flavor": "Google"})
 
     def _non_root_nvmes(self):
         """get list of nvme disks from os, filter away if one of them is root"""
@@ -245,7 +248,7 @@ class gcp_instance:
 
     m1supported="m1-megamem-96" #this is the only exception of supported m1 as per https://cloud.google.com/compute/docs/machine-types#m1_machine_types
 
-    def is_unsupported_instance(self):
+    def is_unsupported_instance_class(self):
         """Returns if this instance type belongs to unsupported ones for nvmes"""
         if self.instancetype == self.m1supported:
             return False
@@ -253,7 +256,7 @@ class gcp_instance:
             return True
         return False
 
-    def is_supported_instance(self):
+    def is_supported_instance_class(self):
         """Returns if this instance type belongs to supported ones for nvmes"""
         if self.instancetype == self.m1supported:
             return True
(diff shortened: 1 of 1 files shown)


dist/common/scripts/scylla_util.py
```diff
@@ -87,23 +87,24 @@ def datadir():
...
                 else:
                     return res.read().decode('utf-8')
         except urllib.error.HTTPError:
-            logging.warn("Failed to grab %s..." % url)
+            logging.warning("Failed to grab %s..." % url)
             time.sleep(5)
             retries += 1
-            if (retries >= max_retries):
+            if retries >= max_retries:
                 raise
```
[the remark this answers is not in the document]
bentsi: this is styling fix, no need to put the round brackets here...

dist/common/scripts/scylla_util.py
```diff
@@ -87,23 +87,24 @@ def datadir():
 def scyllabindir():
     return str(scyllabindir_p())
 
+
 # @param headers dict of k:v
-def curl(url, byte=False, headers={}):
+def curl(url, headers, byte=False):
```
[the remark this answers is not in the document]
bentsi: usually, the order of arguments is chosen by their use, importance, and common use case
example:

- curl(url)
- curl(url, headers)
and if we want to encode the output as byte string we set byte=True
- curl(url, byte=True)
- curl(url, headers, byte=True)
this is a better API.

Regarding making headers mandatory you are right, I will change it (but we can't use a mutable `{}` as a default (https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)

dist/common/scripts/scylla_util.py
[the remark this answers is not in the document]
bentsi: moved to separate commit

dist/common/scripts/scylla_util.py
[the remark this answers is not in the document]
avikivity: And what about the removal of the default?

The goal is not to confuse the maintainer with lots of unrelated changes.

dist/common/scripts/scylla_util.py
```diff
@@ -87,13 +87,14 @@ def datadir():
 def scyllabindir():
     return str(scyllabindir_p())
 
+
 # @param headers dict of k:v
-def curl(url, byte=False, headers={}):
+def curl(url, headers=None, byte=False):
```
avikivity, line 92: Why not write headers={}?

The subject of the commit is wrong, since we could pass headers before. As fat as I can tell, this commit does nothing.

dist/common/scripts/scylla_util.py
```diff
@@ -137,10 +141,9 @@ def is_gce_instance():
                     return True
         return False
 
-    def __instance_metadata(self, path):
-        """query GCP metadata server, recursively!"""
-        return curl("http://169.254.169.254/computeMetadata/v1/instance/" + path+"?recursive=true", headers={"Metadata-Flavor": "Google"})
-        #169.254.169.254 is metadata.google.internal
+    def __instance_metadata(self, path, recursive=False):
+        return curl(self.META_DATA_BASE_URL + path + "?recursive=%s" % str(recursive).lower(),
```
avikivity, line 145: Why did you remove the comment?

dist/common/scripts/scylla_util.py
```diff
@@ -575,18 +613,40 @@ def is_gentoo_variant():
...
+    for pattern, fpath in list_of_patterns_files:
+        if re.match(pattern, get_text_from_path(fpath), flags=re.IGNORECASE):
+            return True
     return False
 
 
+def is_ec2():
+    patterns_files = ((r"^Amazon EC2$", "/sys/class/dmi/id/board_vendor"),
+                      (r"^ec2.*", "/sys/hypervisor/uuid"))
+    return match_patterns_in_files(patterns_files)
```
avikivity, line 633: This looks fragile. Is there no way to do this from the metadata server?

dist/common/scripts/scylla_util.py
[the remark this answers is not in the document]
bentsi: I am fixing a bug, we can't use a mutable ({}) as a default argument to the function (https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) 
I can separate it as a different commit.

dist/common/scripts/scylla_util.py
bentsi: this is what AWS recommends. You can read more here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html

dist/common/scripts/scylla_util.py
bentsi: I will fix the commit message. It was originally done before Lubus changes that added the headers argument.

dist/common/scripts/scylla_util.py
bentsi: after the change, the information in the comment is not useful here

dist/common/scripts/scylla_util.py
avikivity: Thanks. This way I understand the change, and it can be backported independently if needed.

dist/common/scripts/scylla_util.py
```diff
@@ -87,13 +87,14 @@ def datadir():
...
+
 # @param headers dict of k:v
-def curl(url, byte=False, headers={}):
+def curl(url, headers=None, byte=False):
     max_retries = 5
     retries = 0
     while True:
         try:
-            req = urllib.request.Request(url,headers=headers)
+            req = urllib.request.Request(url, headers={} if not headers else headers)
```
avikivity, line 97: You could write this as `headers or {}`

dist/common/scripts/scylla_util.py
avikivity: The second one is not accurate. There is a 1:4000 chance of a random uuid starting with "ec2". They even document it as "probably".

avikivity: I think it's better to use the instance metadata, since we have to read it anyway.

dist/common/scripts/scylla_util.py
bentsi: right. This will look better. Fixed.

dist/common/scripts/scylla_util.py
bentsi: fixed.

slivne:
@avikivity / @syuu1228 - according to bentsi he is fixed everything - please review and approve so we can move forward

dist/common/scripts/scylla_util.py
```diff
@@ -369,6 +394,14 @@ def __init__(self):
         self._type = self.__instance_metadata("instance-type")
         self.__populate_disks()
 
+    def is_aws_instance(self):
```
fruch, line 397: why `is_gce_instance` is static, and this one isn't ?

fruch: or at least a class method

dist/common/scripts/scylla_util.py
```diff
@@ -577,18 +623,29 @@ def is_gentoo_variant():
...
 
-def is_ec2():
-    if os.path.exists('/sys/hypervisor/uuid'):
-        with open('/sys/hypervisor/uuid') as f:
-            s = f.read()
-        return True if re.match(r'^ec2.*', s, flags=re.IGNORECASE) else False
-    elif os.path.exists('/sys/class/dmi/id/board_vendor'):
-        with open('/sys/class/dmi/id/board_vendor') as f:
-            s = f.read()
-        return True if re.match(r'^Amazon EC2$', s) else False
```
syuu1228, line 588: Why do we need to remove is_ec2() to support GCE?
There are multiple scripts relies on the function, if you remove this you will need to fix them too.

dist/common/scripts/scylla_util.py
fruch: Avi, FYI doing `headers={}` in a function signature, can bite someone very hard, 
since this dict is created only once. it basically becomes a global variable.

dist/common/scripts/scylla_util.py
bentsi: the whole scylla_util lacks basic architecture and basic usage of OOP design principles (Base class in this case). I am not going to redesign and rewrite this...
I can't use static method since I need `META_DATA_BASE_URL`

bentsi: I can use classmethod

slivne:
@bentsi ping

dist/common/scripts/scylla_util.py
bentsi: right, my bad, deleted it mistakenly when resolving conflicts with Lubush changes

bentsi: fixed

dist/common/scripts/scylla_util.py
bentsi: fixed

bentsi:
@slivne all comments resolved

avikivity:
@syuu1228 @fruch please re-review.

fruch (approved):
Looks o.k.

dist/common/scripts/scylla_util.py
avikivity: Yes, unexpected for a nice language such as Python.

dist/common/scripts/scylla_util.py
```diff
@@ -577,18 +624,37 @@ def is_gentoo_variant():
...
+
+def match_patterns_in_files(list_of_patterns_files):
+    for pattern, fpath in list_of_patterns_files:
+        if re.match(pattern, get_text_from_path(fpath), flags=re.IGNORECASE):
+            return True
     return False
 
 
+def is_ec2():
+    return aws_instance.is_aws_instance()
```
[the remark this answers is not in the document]
tarzanek: also GCP actually just does dns query (but I'd love it to be optimized and use /sys or any other variable instead)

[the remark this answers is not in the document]
bentsi: @tarzanek That's how it was originally implemented but Avi was not happy about it:
comment from above:
image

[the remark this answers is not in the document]
tarzanek: so how will this method then behave on non cloud servers? or it will just wait for timeout?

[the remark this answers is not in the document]
bentsi: in the current implementation, it will wait for a timeout... but it's not critical since those functions are mainly used in the Scylla Amazon/GCE images

PR #7089, opened 2020-08-24
install.sh: show warning when systemd does not support user mode

### LINKED ISSUE #7071: offline-install: Failed to get D-Bus connection: No such file or directory on CentOS 7
*Installation details*
Scylla version (or git commit hash):  scylla-unified-package-0.20200819.2e1499ee93f.tar.gz / 666.development-0.20200819.2e1499ee93f
Cluster size: 1

### Description

The offline installation failed for a dbus connection error.

#### prepare

- curl http://downloads.scylladb.com.s3.amazonaws.com/relocatable/unstable/master/latest/scylla-unified-package-0.20200819.2e1499ee93f.tar.gz -o scylla-unified-package-0.20200819.2e1499ee93f.tar.gz
- tar xvfz scylla-unified-package-0.20200819.2e1499ee93f.tar.gz

#### install error

```
[scylla-test@amos-centos7-offline-installer offline_install]$ time ./install.sh --nonroot --prefix `pwd`/install_root
Failed to get D-Bus connection: No such file or directory

real	1m56.806s
user	0m0.978s
sys	0m7.878s
[scylla-test@amos-centos7-offline-installer offline_install]$ echo $?
1
```

```
[scylla-test@amos-centos7-offline-installer offline_install]$ ls
install_root  offline-install.sh  scylla-jmx      scylla-tools
install.sh    scylla              scylla-python3  scylla-unified-package-0.20200819.2e1499ee93f.tar.gz

[scylla-test@amos-centos7-offline-installer offline_install]$ cat /etc/redhat-release 
CentOS Linux release 7.8.2003 (Core)

[scylla-test@amos-centos7-offline-installer offline_install]$ rpm -qa |grep dbus
dbus-glib-0.100-7.el7.x86_64
python-slip-dbus-0.4.0-4.el7.noarch
dbus-libs-1.10.24-14.el7_8.x86_64
dbus-1.10.24-14.el7_8.x86_64
dbus-python-1.1.1-9.el7.x86_64
[scylla-test@amos-centos7-offline-installer offline_install]$ ps aux|grep dbus
dbus       463  0.0  0.0  66472  2196 ?        Ssl  23:31   0:00 /usr/bin/dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation
```

```
[scylla-test@amos-centos7-offline-installer install_root]$ ./bin/scylla --version
666.development-0.20200819.2e1499ee93f
```

/CC @penberg @roydahan @slivne

syuu1228:
On older distribution such as CentOS7, it does not support systemd user mode.
On such distribution nonroot mode does not work, show error message and
reject installing before copying files.

Fixes #7071

### DIFF
--- a/unified/install.sh
+++ b/unified/install.sh
@@ -38,6 +38,11 @@ EOF
     exit 1
 }
 
+check_usermode_support() {
+    user=$(systemctl --help|grep -e '--user')
+    [ -n "$user" ]
+}
+
 root=/
 housekeeping=false
 python3=/opt/scylladb/python3/bin/python3
@@ -79,6 +84,12 @@ while [ $# -gt 0 ]; do
             ;;
     esac
 done
+
+if $nonroot && ! check_usermode_support; then
+    echo "WARNING: This distribution does not support systemd user mode, please configure and launch Scylla manually."
+    echo "Press any key to continue..."
+    read
+fi
 
 if [ -z "$prefix" ]; then
     if $nonroot; then
--- a/install.sh
+++ b/install.sh
@@ -167,6 +167,11 @@ installconfig() {
     fi
 }
 
+check_usermode_support() {
+    user=$(systemctl --help|grep -e '--user')
+    [ -n "$user" ]
+}
+
 # change directory to the package's root directory
 cd "$(dirname "$0")"
 
@@ -364,7 +369,9 @@ if $nonroot; then
     # nonroot install is also 'offline install'
     touch $rprefix/SCYLLA-OFFLINE-FILE
     touch $rprefix/SCYLLA-NONROOT-FILE
-    systemctl --user daemon-reload
+    if check_usermode_support; then
+        systemctl --user daemon-reload
+    fi
     echo "Scylla non-root install completed."
 elif ! $packaging; then
     # run install.sh without --packaging is 'offline install'


avikivity:
It is too much to reject. We can install, but not install the systemd support parts. Users can still run directly.

@eyalgutkind please provide your input - should we partially support nonroot install on CentOS 7 (with full support on CentOS 8), or drop support for CentOS 7 nonroot installs completely.

eyalgutkind:
It might delay some users, But, in general I do not want to drag it. any more.
We can place a restriction supporting non-root installation on CentOS 8+ distros.

penberg:
@avikivity So let's merge this now and relax the check for CentOS 7 later if needed?

avikivity:
We're sure to have CentOS 7 users (nonroot users tend to be late adopters). The check shouldn't be complicated.

syuu1228:
Pushed v2, not reject setup just print warning and skip running systemctl --user instead.

unified/install.sh
```diff
@@ -80,6 +85,12 @@ while [ $# -gt 0 ]; do
     esac
 done
 
+if $nonroot && ! check_usermode_support; then
+    echo "WARNING: This distribution does not support systemd user mode, need to configure & launch Scylla manually."
```
penberg, line 89: Minor nit:

```WARNING: This distribution does not support systemd user mode, please configure and launch Scylla manually```

syuu1228: @penberg Fixed the message.

avikivity:
Dequeued, breaks next. Maybe it fails in a container without systemd because it waits for keypress, or maybe something else.

PR #7091, opened 2020-08-24
Do verify package for offline

syuu1228:
To verify offline installed image with do_verify_package() in scylla_setup, we introduce is_offline() function works just like is_nonroot() but the install not with --nonroot.

### DIFF
--- a/dist/common/scripts/scylla_setup
+++ b/dist/common/scripts/scylla_setup
@@ -107,8 +107,14 @@ def interactive_choose_swap_size():
         except Exception as e:
             print(f'"{s}" is not valid number, input again')
 
+pkg_files = {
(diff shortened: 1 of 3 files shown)


dist/common/scripts/scylla_setup
```diff
@@ -107,8 +107,14 @@ def interactive_choose_swap_size():
         except Exception as e:
             print(f'"{s}" is not valid number, input again')
 
+pkg_files = {
+        'scylla-jmx': scylladir_p() / 'jmx/scylla-jmx',
+        'scylla-tools': scylladir_p() / 'share/cassandra/bin/cqlsh'
+        }
```
avikivity, line 113: Don't this need to be adjusted for $PRODUCT? I suppose it can be fixed later since the problem already exists.

PR #7092, opened 2020-08-24
cdc: Remove stored (postimage) data when doing row delete

### LINKED ISSUE #6900: CDC: within a batch, partition deletes and range deletes do not affect postimage
A CQL batch that both

1. inserts or updates a row `X` with timestamp `T`,
2. deletes that row using row delete or partition delete that covers the row `X`, also using timestamp `T`,

has wrong postimage for the row `X`. Current logic shows the value set by 1) and ignores the effect of 2).

It is wrong, because range/partition delete wins with the insert/update, and cancels its effect - therefore postimage in this case does not reflect the actual state of the row after the batch is applied.

Example with insert + range delete:

```cql
cqlsh:ks> create table tbl (pk int, ck int, v int, primary key(pk, ck)) with cdc = {'enabled': true, 'preimage': true, 'postimage': true};
cqlsh:ks> begin batch insert into tbl (pk, ck, v) values (1, 1, 100); delete from tbl where pk = 1 and ck >= 1 and ck <= 2; apply batch;
cqlsh:ks> select * from tbl;

 pk | ck | v
----+----+---

(0 rows)
cqlsh:ks> select * from tbl_scylla_cdc_log;

 cdc$stream_id                      | cdc$time                             | cdc$batch_seq_no | cdc$deleted_v | cdc$operation | cdc$ttl | ck | pk | v
------------------------------------+--------------------------------------+------------------+---------------+---------------+---------+----+----+------
 0xc74d27bc17a40015893bffbb671eb09d | 6743d17a-cc0f-11ea-8c1b-e371909c58d6 |                0 |          null |             0 |    null |  1 |  1 | null
 0xc74d27bc17a40015893bffbb671eb09d | 6743d17a-cc0f-11ea-8c1b-e371909c58d6 |                1 |          null |             2 |    null |  1 |  1 |  100
 0xc74d27bc17a40015893bffbb671eb09d | 6743d17a-cc0f-11ea-8c1b-e371909c58d6 |                2 |          null |             5 |    null |  1 |  1 | null
 0xc74d27bc17a40015893bffbb671eb09d | 6743d17a-cc0f-11ea-8c1b-e371909c58d6 |                3 |          null |             7 |    null |  2 |  1 | null
 0xc74d27bc17a40015893bffbb671eb09d | 6743d17a-cc0f-11ea-8c1b-e371909c58d6 |                4 |          null |             9 |    null |  1 |  1 |  100

(5 rows)
```

Example with insert + partition delete:

```cql
cqlsh:ks> create table tbl (pk int, ck int, v int, primary key(pk, ck)) with cdc = {'enabled': true, 'preimage': true, 'postimage': true};
cqlsh:ks> begin batch insert into tbl (pk, ck, v) values (1, 1, 100); delete from tbl where pk = 1; apply batch;
cqlsh:ks> select * from tbl;

 pk | ck | v
----+----+---

(0 rows)
cqlsh:ks> select * from tbl_scylla_cdc_log;

 cdc$stream_id                      | cdc$time                             | cdc$batch_seq_no | cdc$deleted_v | cdc$operation | cdc$ttl | ck   | pk | v
------------------------------------+--------------------------------------+------------------+---------------+---------------+---------+------+----+------
 0xc74d27bc17a40015893bffbb671eb09d | bd56e1f6-cc0f-11ea-67fc-ead359feab5f |                0 |          null |             0 |    null |    1 |  1 | null
 0xc74d27bc17a40015893bffbb671eb09d | bd56e1f6-cc0f-11ea-67fc-ead359feab5f |                1 |          null |             2 |    null |    1 |  1 |  100
 0xc74d27bc17a40015893bffbb671eb09d | bd56e1f6-cc0f-11ea-67fc-ead359feab5f |                2 |          null |             4 |    null | null |  1 | null
 0xc74d27bc17a40015893bffbb671eb09d | bd56e1f6-cc0f-11ea-67fc-ead359feab5f |                3 |          null |             9 |    null |    1 |  1 |  100

(4 rows)
```

In both cases, it can be seen that the inserted row does NOT appear in base table, but the postimage suggests otherwise (it should be `null`, but is `100`).

elcallio:
Fixes #6900

Clustered range deletes did not clear out the "row_states" data associated
with affected rows (might be many).

Adds a sweep through and erases relevant data. Since we do pre- and
postimage in "order", this should only affect postimage.

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -885,6 +885,10 @@ public:
           _log_mut(log_mut)
     {}
 
+    const schema& base_schema() const {
+        return _base_schema;
+    }
+
     // Creates a new clustering row in the mutation, assigning it the next `cdc$batch_seq_no`.
     // The numbering of batch sequence numbers starts from 0.
     clustering_key allocate_new_log_row() {
@@ -1296,12 +1300,31 @@ struct process_change_visitor {
             auto log_ck = _builder.allocate_new_log_row(end_operation);
             _builder.set_clustering_columns(log_ck, rt.end);
         }
+
+        // #6900 - remove stored row data (for postimage)
+        // if it falls inside the clustering range of the
+        // tombstone.
+        if (!_enable_updating_state) {
+            return;
+        }
+
+        bound_view::compare cmp(_builder.base_schema());
+        auto sb = rt.start_bound();
+        auto eb = rt.end_bound();
+
+        std::erase_if(_clustering_row_states, [&](auto& p) {
+            auto& ck = p.first;
+            return cmp(sb, ck) && !cmp(eb, ck);
+        });
     }
 
     void partition_delete(const tombstone&) {
         _touched_parts.set<stats::part_type::PARTITION_DELETE>();
 
         auto log_ck = _builder.allocate_new_log_row(operation::partition_delete);
+        if (_enable_updating_state) {
+            _clustering_row_states.clear();
+        }
     }
 
     constexpr bool finished() const { return false; }
--- a/test/cql/cdc_batch_delete_postimage.cql
+++ b/test/cql/cdc_batch_delete_postimage.cql
@@ -1,0 +1,21 @@
+-- do range delete in batch
+create table ks.t (pk int, ck int, v int, primary key(pk, ck)) with cdc = {'enabled': true, 'preimage': true, 'postimage': true};
+
+begin batch insert into ks.t (pk, ck, v) values (1, 1, 100); delete from t where pk = 1 and ck >= 1 and ck <= 2; apply batch;
+
+select "cdc$batch_seq_no", "cdc$operation", pk, ck, v from t_scylla_cdc_log;
+
+-- do pk delete in batch
+create table ks.t2 (pk int, ck int, v int, primary key(pk, ck)) with cdc = {'enabled': true, 'preimage': true, 'postimage': true};
+
+begin batch insert into ks.t2 (pk, ck, v) values (1, 1, 100); delete from t2 where pk = 1; apply batch;
+
+select "cdc$batch_seq_no", "cdc$operation", pk, ck, v from t2_scylla_cdc_log;
+
+-- do range delete in batch, but not matcing ck
+create table ks.t3 (pk int, ck int, v int, primary key(pk, ck)) with cdc = {'enabled': true, 'preimage': true, 'postimage': true};
+
+-- do range delete in batch
+begin batch insert into ks.t3 (pk, ck, v) values (1, 1, 100); delete from t3 where pk = 1 and ck >= 2 and ck <= 3; apply batch;
+
+select "cdc$batch_seq_no", "cdc$operation", pk, ck, v from t3_scylla_cdc_log;
--- a/test/cql/cdc_batch_delete_postimage.result
+++ b/test/cql/cdc_batch_delete_postimage.result
@@ -1,0 +1,141 @@
+-- do range delete in batch
+create table ks.t (pk int, ck int, v int, primary key(pk, ck)) with cdc = {'enabled': true, 'preimage': true, 'postimage': true};
+{
+	"status" : "ok"
+}
+
+begin batch insert into ks.t (pk, ck, v) values (1, 1, 100); delete from t where pk = 1 and ck >= 1 and ck <= 2; apply batch;
+{
+	"status" : "ok"
+}
+
+select "cdc$batch_seq_no", "cdc$operation", pk, ck, v from t_scylla_cdc_log;
+{
+	"rows" : 
+	[
+		{
+			"cdc$batch_seq_no" : "0",
+			"cdc$operation" : "0",
+			"ck" : "1",
+			"pk" : "1"
+		},
+		{
+			"cdc$batch_seq_no" : "1",
+			"cdc$operation" : "2",
+			"ck" : "1",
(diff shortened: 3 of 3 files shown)


cdc/log.cc
```diff
@@ -1296,6 +1300,23 @@ struct process_change_visitor {
...
+        bound_view::compare cmp(_builder.base_schema());
+        auto sb = rt.start_bound();
+        auto eb = rt.end_bound();
+        auto i = _clustering_row_states.begin();
+        auto e = _clustering_row_states.end();
+        while (i != e) {
+            auto& ck = i->first;
+            if (cmp(sb, ck) && !cmp(eb, ck)) {
+                i = _clustering_row_states.erase(i);
+                continue;
```
haaawk, line 1316: wouldn't _clustering_row_states.erase_if be of use here?

cdc/log.cc
```diff
@@ -1296,6 +1300,23 @@ struct process_change_visitor {
             auto log_ck = _builder.allocate_new_log_row(end_operation);
             _builder.set_clustering_columns(log_ck, rt.end);
         }
+
+        // #6900 - remove stored row data (for postimage)
+        // if it falls inside the clustering range of the
+        // tombstone.
+        bound_view::compare cmp(_builder.base_schema());
```
kbr-, line 1307: `if (_enable_updating_state)`

cdc/log.cc
```diff
@@ -1296,6 +1300,23 @@ struct process_change_visitor {
...
+            auto& ck = i->first;
+            if (cmp(sb, ck) && !cmp(eb, ck)) {
+                i = _clustering_row_states.erase(i);
+                continue;
+            }
+            ++i;
+        }
     }
 
     void partition_delete(const tombstone&) {
```
kbr-, line 1322: https://github.com/scylladb/scylla/issues/6900 also mentioned partition deletes

cdc/log.cc
elcallio: There is no `erase_if`member function. There is `std::erase_if(std::unordered_map, pred)`in c++20 though. Sorry, I am so old school when it comes to maps.

kbr-: Generally we should avoid "raw loops"

kbr-:
Needs tests: both for range deletes and partition delete

haaawk (approved):
LGTM

haaawk:
@piodul please have a look.

cdc/log.cc
kbr-: @elcallio there is `erase_if` for vectors too
https://en.cppreference.com/w/cpp/container/vector/erase2

and there is `remove_if` for iterators

elcallio: Yeah, I know. But I am old enough to feel weird about std routines that don't operate on iterators only.

cdc/log.cc
elcallio: true

cdc/log.cc
elcallio: Right.

elcallio: fixed

cdc/log.cc
elcallio: fixed

cdc/log.cc
elcallio: fixed

elcallio:
> 
> 
> Needs tests: both for range deletes and partition delete

Added cql test.

kbr- (approved):

test/cql/cdc_batch_delete_postimage.cql
```diff
@@ -0,0 +1,21 @@
...
+select "cdc$batch_seq_no", "cdc$operation", pk, ck, v from t_scylla_cdc_log;
+
+-- do pk delete in batch
+create table ks.t2 (pk int, ck int, v int, primary key(pk, ck)) with cdc = {'enabled': true, 'preimage': true, 'postimage': true};
+
+begin batch insert into ks.t2 (pk, ck, v) values (1, 1, 100); delete from t2 where pk = 1; apply batch;
+
+select "cdc$batch_seq_no", "cdc$operation", pk, ck, v from t2_scylla_cdc_log;
+
+-- do range delete in batch, but not matcing ck
```
piodul, line 15: Nit: s/matcing/matching

piodul (commented):
You should also mention in PR title/PR desc/commit desc that this change also fixes partition deletes.

Otherwise, LGTM.

cdc/log.cc
```diff
@@ -1296,12 +1300,31 @@ struct process_change_visitor {
...
+        // tombstone.
+        if (!_enable_updating_state) {
+            return;
+        }
+
+        bound_view::compare cmp(_builder.base_schema());
+        auto sb = rt.start_bound();
+        auto eb = rt.end_bound();
+
+        std::erase_if(_clustering_row_states, [&](auto& p) {
```
haaawk, line 1315: @avikivity won't be happy about generic lambda ;)

haaawk: Unless VSCode can handle them now :D

haaawk (approved):
LGTM

nyh:
Thanks, committed to next (5ed3d6892da2fa9d191faf55b35a01ba76c776d7).
I see that despite the two approvals, there are a few unresolved nitpicks, so if you want to address those please send a followup patch. Thanks.

PR #7096, opened 2020-08-24
cdc: Preserve metadata columns when geting only keys for delta

### LINKED ISSUE #7095: cdc delta == keys does not produce cdc$operation nor cdc$ttl
If cdc::delta is set to "keys" (keys only), the result is postprocessed to remove certain columns or rows (inefficient - with the latest refactor, it should be easy to just not create it in the first place). 

However, even `cdc$operation` and `cdc$ttl` is removed. That does not seem right.

haaawk:
Fixes #7095

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -1424,9 +1424,8 @@ private:
                     // The case of `get_delta_mode() == delta_mode::keys`:
                     it->row().cells().remove_if([this, log_s = m.schema()] (column_id id, atomic_cell_or_collection& acoc) {
                         const auto& log_cdef = log_s->column_at(column_kind::regular_column, id);
-                        // We can surely remove "cdc$*" columns.
                         if (is_cdc_metacolumn_name(log_cdef.name_as_text())) {
-                            return true;
+                            return false;
                         }
                         const auto* base_cdef = _schema->get_column_definition(log_cdef.name());
                         // Remove columns from delta that correspond to non-PK/CK columns in the base table.
--- a/test/cql/cdc_delta_modes_test.result
+++ b/test/cql/cdc_delta_modes_test.result
@@ -54,6 +54,8 @@ select "cdc$batch_seq_no", "cdc$operation", "cdc$ttl", pk, ck, i1 from tb2_scyll
 		},
 		{
 			"cdc$batch_seq_no" : "1",
+			"cdc$operation" : "2",
+			"cdc$ttl" : "2222",
 			"ck" : "22",
 			"pk" : "2"
 		},


haaawk:
I'm updating the tests as well - stay tuned.

haaawk:
Done all the tests pass now.

elcallio:
I think a better fix is to not generate the value parts in the first place, and remove the filtering. With @kbr- :s visitor pattern it is quite straight forward. I have it included in a patch set. I'll send a PR tomorrow.

haaawk:
> I think a better fix is to not generate the value parts in the first place, and remove the filtering. With @kbr- :s visitor pattern it is quite straight forward. I have it included in a patch set. I'll send a PR tomorrow.

Is it going to be an if in every visitor method?

haaawk:
Let's fix the bug with this simple PR and leave the refactoring for the future @elcallio. Refactoring is not needed to fix the bug and simple change will be easier to backport.

elcallio:
More or less. However, if you are worried about branches, I would perhaps not worry about that being a performance issue.

haaawk:
I'm mostly worried about maintainability. Look how simple it was to fix the issue now. I knew where to look almost instantly. Additional ifs in every branch of the logic (handling of each cell type) will certainly contribute to the spaghetti. Let's think for a while if there is no better approach. Maybe the logic could be part of the builder and we can somehow avoid spreading this logic to almost all builder's functions.

One way or the other, let's deal with one problem at a time. This PR fixes the issue. And then let's get another PR to refactor. It will be much more tricky to refactor it in the right way which balances speed and maintainability. I would like to really think it through because I feel that performance and maintainability are equally important in the CDC at the moment. We don't even know how big of a problem the current implementation is and if it's a problem at all. Would be good to see some flame graphs and based on them assess whether the optimization is worth it. It may well be but unless we measure we won't even know if it makes anything faster or not. The optimization certainly has a price of increasing complexity of the code. We need to assess if the gain is worth the price and try to make the price as low as possible.

Sorry for being such a killjoy but it became really hard to introduce changes in CDC without breaking something on the way. We need to fight this or it will bite us back sooner or later.

nyh:
Thanks. Committed to next (de153075f845926c061fa3414f0eedaf8ae19624). I agree with @haaawk that it makes sense to quickly fix a big now, even if you plan to do a better fix later as part of a bigger rewrite. It's even more important to fix the wrong tests now, so they don't continue to enshrine wrong behavior.

And if I'm on the topic of tests: @haaawk @kostja this patch demonstrates what I think is the main downside of the idea of diff-based regression tests, which try to validate that the output doesn't change instead of verifying that it is correct. This testing framework makes it too easy to do what happened in this case: Somebody wrote a test that had "wrong" output (in this case, a missing `cdc$operation` column) but instead of noticing that this output is actually wrong, it was saved as the gold standard for the next runs to check, and from there on, it looks as if this test is "passing".

haaawk:
Thanks, @nyh. 

Regarding the tests: I actually think cql tests are a step in the right direction because they remove boiler-plate which allows us to focus on the meat.

> but instead of noticing that this output is actually wrong, it was saved as the gold standard for the next runs to check

this could have happened with regular unit tests as well. I don't think it's the fault of CQL tests. BTW @jul-stas thought it is the right behavior and he wrote the test together with the change that introduced the behavior.

slivne:
@haaawk is this still relevant if not lets close

PR #7097, opened 2020-08-24
ninja test: add --repeat and --timeout parameters

yaronkaikov:
Adding missing parameters to ninja test:
--repeat: number of times to repeat each test, default = 3
--timeout: set total timeout in sec for each test


penberg:
Queued as 7d5ad1d9da078edd3565f942313554a8cf5f0875

PR #7100, opened 2020-08-24
reloc: remove build_reloc.sh

syuu1228:
Now we uses ninja-build to build relocatable package, no script calling
build_reloc.sh anymore, so we can drop it.


avikivity:
I see 

```
$ git grep build_reloc
configure.py:          command = cd $reloc_dir && ./reloc/build_reloc.sh --nodeps
dist/docker/redhat/README.md:2. Run `./reloc/build_reloc.sh --mode dev`
reloc/build_deb.sh:    echo "Run ./reloc/build_reloc.sh first."
reloc/build_rpm.sh:    echo "Run ./reloc/build_reloc.sh first."
```

PR #7105, opened 2020-08-25
scylla-gdb.py: Add scylla repairs command

### LINKED ISSUE #7103: scylla-gdb.py: Add scylla repairs helper
It will be helpful if we have a scylla repairs command that prints all the pending repair_meta objects and at which step the master and followers are.

asias:
This series adds scylla repairs command to help debug repair.

Fixes #7103

### DIFF
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -2314,7 +2314,7 @@ class row_level_repair {
     std::vector<gms::inet_address> _all_live_peer_nodes;
     column_family& _cf;
 
-    // All particular peer nodes not including the node itself.
+    // All peer nodes and the node itself.
     std::vector<gms::inet_address> _all_nodes;
 
     // Repair master and followers will propose a sync boundary. Each of them
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -380,6 +380,8 @@ private:
     std::optional<evictable_reader_handle> _reader_handle;
     // Current partition read from disk
     lw_shared_ptr<const decorated_key_with_hash> _current_dk;
+    uint64_t _reads_issued = 0;
+    uint64_t _reads_finished = 0;
 
 public:
     repair_reader(
@@ -431,7 +433,11 @@ public:
 
     future<mutation_fragment_opt>
     read_mutation_fragment() {
-        return _reader(db::no_timeout);
+        ++_reads_issued;
+        return _reader(db::no_timeout).then([this] (mutation_fragment_opt mfopt) {
+            ++_reads_finished;
+            return mfopt;
+        });
     }
 
     void on_end_of_stream() {
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -634,6 +634,7 @@ public:
 
 class repair_meta;
 class repair_meta_tracker;
+class row_level_repair;
 
 static void add_to_repair_meta_for_masters(repair_meta& rm);
 static void add_to_repair_meta_for_followers(repair_meta& rm);
@@ -693,6 +694,7 @@ private:
     sink_source_for_get_row_diff _sink_source_for_get_row_diff;
     sink_source_for_put_row_diff _sink_source_for_put_row_diff;
     tracker_link_type _tracker_link;
+    row_level_repair* _row_level_repair_ptr;
 public:
     repair_stats& stats() {
         return _stats;
@@ -730,7 +732,8 @@ public:
             uint32_t repair_meta_id,
             streaming::stream_reason reason,
             shard_config master_node_shard_config,
-            size_t nr_peer_nodes = 1)
+            size_t nr_peer_nodes = 1,
+            row_level_repair* row_level_repair_ptr = nullptr)
             : _db(db)
             , _messaging(ms)
             , _cf(cf)
@@ -771,6 +774,7 @@ public:
                     [&ms] (uint32_t repair_meta_id, netw::messaging_service::msg_addr addr) {
                         return ms.local().make_sink_and_source_for_repair_put_row_diff_with_rpc_stream(repair_meta_id, addr);
                 })
+            , _row_level_repair_ptr(row_level_repair_ptr)
             {
             if (master) {
                 add_to_repair_meta_for_masters(*this);
@@ -2605,7 +2609,8 @@ public:
                     repair_meta_id,
                     _ri.reason,
                     std::move(master_node_shard_config),
-                    _all_live_peer_nodes.size());
+                    _all_live_peer_nodes.size(),
+                    this);
 
             // All nodes including the node itself.
             _all_nodes.insert(_all_nodes.begin(), master.myip());
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -65,6 +65,42 @@ distributed<db::system_distributed_keyspace>* _sys_dist_ks;
 distributed<db::view::view_update_generator>* _view_update_generator;
 sharded<netw::messaging_service>* _messaging;
 
+enum class repair_state : uint16_t {
+    unknown,
+    row_level_start_started,
+    row_level_start_finished,
+    get_estimated_partitions_started,
+    get_estimated_partitions_finished,
+    set_estimated_partitions_started,
+    set_estimated_partitions_finished,
+    get_sync_boundary_started,
+    get_sync_boundary_finished,
+    get_combined_row_hash_started,
+    get_combined_row_hash_finished,
+    get_row_diff_with_rpc_stream_started,
+    get_row_diff_with_rpc_stream_finished,
+    get_row_diff_and_update_peer_row_hash_sets_started,
+    get_row_diff_and_update_peer_row_hash_sets_finished,
+    get_full_row_hashes_with_rpc_stream_started,
+    get_full_row_hashes_with_rpc_stream_finished,
+    get_full_row_hashes_started,
+    get_full_row_hashes_finished,
+    get_row_diff_started,
+    get_row_diff_finished,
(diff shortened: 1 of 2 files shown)


repair/row_level.cc
```diff
@@ -695,7 +732,22 @@ class repair_meta {
     sink_source_for_put_row_diff _sink_source_for_put_row_diff;
     tracker_link_type _tracker_link;
     row_level_repair* _row_level_repair_ptr;
+    std::vector<repair_node_state> _all_nodes;
```
denesb, line 735: `_all_node_states` or just `_node_states`.

repair/row_level.cc
```diff
@@ -1576,7 +1636,10 @@ class repair_meta {
         rlogger.debug("<<< Finished Row Level Repair (Follower): local={}, peers={}, repair_meta_id={}, keyspace={}, cf={}, range={}",
             utils::fb_utilities::get_broadcast_address(), from, repair_meta_id, ks_name, cf_name, range);
         auto rm = get_repair_meta(from, repair_meta_id);
-        return remove_repair_meta(from, repair_meta_id, std::move(ks_name), std::move(cf_name), std::move(range));
+        rm->set_repair_state_for_local_node(repair_state::row_level_stop_started);
+        return remove_repair_meta(from, repair_meta_id, std::move(ks_name), std::move(cf_name), std::move(range)).then([rm] {
+            rm->set_repair_state_for_local_node(repair_state::row_level_stop_finished);
```
denesb, line 1641: Is the `rm` alive here still?

scylla-gdb.py
```diff
@@ -3789,6 +3789,25 @@ def invoke(self, arg, for_tty):
             f = reference_wrapper(f).get()
```
denesb, line 3789: Please add an example printout to the commit log.

scylla-gdb.py
```diff
@@ -3789,6 +3789,25 @@ def invoke(self, arg, for_tty):
...
             gdb.write('%s: %s\n' % (f['_name'], f['_enabled']))
 
+class scylla_repairs(gdb.Command):
```
denesb, line 3792: Please document the command, see e.g. `scylla fiber` for an example.

denesb (changes requested):
Very good, this should help debugging repairs a lot. I had just one major comment (I suspect a possible use-after-free).

scylla-gdb.py
```diff
@@ -656,8 +656,9 @@ class boost_intrusive_list_printer(gdb.printing.PrettyPrinter):
     def __init__(self, val):
         self.val = intrusive_list(val)
     def to_string(self):
-        items = [str(v) for v in self.val]
-        return 'boost::intrusive::list of size {} = [{}]'.format(len(items), ', '.join(items))
+        items = ['@' + str(v.address) + '=' + str(v) for v in self.val]
+        ptrs = [str(v.address) for v in self.val]
+        return 'boost::intrusive::list of size {} = [{}] = [{}]'.format(len(items), ', '.join(ptrs), ', '.join(items))
```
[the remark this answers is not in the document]
asias: Yes, but I wanted to have a lists of the pointers printed before the full repair_meta objects  are printed which are pretty verbose and long.

scylla-gdb.py
```diff
@@ -656,8 +656,9 @@ class boost_intrusive_list_printer(gdb.printing.PrettyPrinter):
     def __init__(self, val):
         self.val = intrusive_list(val)
     def to_string(self):
-        items = [str(v) for v in self.val]
-        return 'boost::intrusive::list of size {} = [{}]'.format(len(items), ', '.join(items))
+        items = ['@' + str(v.address) + '=' + str(v) for v in self.val]
```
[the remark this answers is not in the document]
asias: Yes, it is hex.

repair/row_level.cc
asias: I prefer to keep the word all. I will change it to _all_node_states.

scylla-gdb.py
asias: ok.

scylla-gdb.py
asias: ok.

repair/row_level.cc
asias: I think it is alive. The call rm->set_repair_state_for_local_node does not yield.   We should be able to access rm inside the lambda: [rm] { rm->foo() }.

asias:
Changes in f30895ad22420f960192c8c135bd6adc0e5f0c3a
  - Added docuemnt for scylla repairs cmd
  - Added example in commit log
  - Renamed _all_nodes to _all_node_states
  - Rebased to master
  - Use struct for repair_node_state

repair/row_level.cc
```diff
@@ -65,6 +65,43 @@ distributed<db::system_distributed_keyspace>* _sys_dist_ks;
...
+    row_level_stop_started,
+    row_level_stop_finished,
+};
+
+class repair_node_state {
+public:
+    gms::inet_address node;
+    repair_state state = repair_state::unknown;
+    explicit repair_node_state(gms::inet_address n) : node(n) { }
+};
```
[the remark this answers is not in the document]
asias: OK. I do not mind changing it.

scylla-gdb.py
```diff
@@ -3789,6 +3789,49 @@ def invoke(self, arg, for_tty):
...
+         _db = @0x7fffe538c9f0,
+         _messaging = @0x7fffe538ca90,
+        _cf = @0x6000066f0000,
+
+       ....
+
+      (repair_meta*) for follower: addr = 0x60000432a808, table = myks2.standard1, ip = 127.0.0.1, states = ['127.0.0.1->repair_state::get_sync_boundary_started', '127.0.0.2->repair_state::unknown'], repair_meta = {
+        db = @0x7fffe538c9f0,
+        messaging = @0x7fffe538ca90,
+        _cf = @0x6000066f0000,
```
denesb, line 3814: I think printing the entire `repair_meta` object is too much, its enough to print the address, table, ip and state, if someone wants to have a look they can print the object themselves. Something like this:
```
(repair_meta*) 0x60000432a808 kind=follower table=myks2.standard1 ip=127.0.0.1 states=[]
```
Having it all in single line helps readability a lot.

repair/row_level.cc
denesb: Won't `remove_repair_meta()` destroy it? We are in a continuation attached to that here.

scylla-gdb.py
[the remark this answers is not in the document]
denesb: Ok.

scylla-gdb.py
asias: I thought about only printing some "headers", but when we want to see other fields on all repair_meta objects, we need to loop those  repair_meta manually one by one. There can be a lot of repair_meta objects. We can redirect the output to a file and grep easily. So, I think it is better to print all details.

repair/row_level.cc
asias:
```
 914     static future<>
 915     remove_repair_meta(const gms::inet_address& from,
 916             uint32_t repair_meta_id,
 917             sstring ks_name,
 918             sstring cf_name,
 919             dht::token_range range) {
 920         node_repair_meta_id id{from, repair_meta_id};
 921         auto it = repair_meta_map().find(id);
 922         if (it == repair_meta_map().end()) {
 923             rlogger.warn("remove_repair_meta: repair_meta_id {} for node {} does not exist", id.repair_meta_id, id.ip);
 924             return make_ready_future<>();
 925         } else {
 926             auto rm = it->second;
 927             repair_meta_map().erase(it);
 928             rlogger.debug("remove_repair_meta: Stop repair_meta_id {} for node {} started", id.repair_meta_id, id.ip);
 929             return rm->stop().then([rm, id] {
 930                 rlogger.debug("remove_repair_meta: Stop repair_meta_id {} for node {} finished", id.repair_meta_id, id.ip);
 931             });
 932         }
 933     }
```
It only removes the lw_shared_ptr from repair_meta_map(), we still have a reference to rm, so it will not be destroyed.

scylla-gdb.py
denesb: We should have a flag to toggle between the two modes (summary/full). Can be a follow-up.

repair/row_level.cc
denesb: I see.

denesb (approved):

repair/row_level.cc
```diff
@@ -695,7 +731,22 @@ class repair_meta {
     sink_source_for_put_row_diff _sink_source_for_put_row_diff;
     tracker_link_type _tracker_link;
     row_level_repair* _row_level_repair_ptr;
+    std::vector<repair_node_state> _all_node_states;
 public:
+    std::vector<repair_node_state>& all_nodes() {
+        return _all_node_states;
+    }
+    void set_repair_state(repair_state state, gms::inet_address node) {
+        for (auto& ns : all_nodes()) {
```
avikivity, line 740: Won't this be slow in large clusters? A node can have has many peers as there are tokens.

asias: It only depends on the RF not the number of nodes in cluster. Also set_repair_state is called on the stop code path, which is a slow path.

scylla-gdb.py
asias: Yes.

repair/row_level.cc
avikivity: Ah, this is per-vnode and not per-repair?

asias: Yes.

PR #7121, opened 2020-08-26
cdc: Remove post-filterings for keys-only/off cdc delta generation

### LINKED ISSUE #7128: cdc: Remove delta_mode::off
CDC information is not useful without at least `delta==keys`, since pre/post image data contains no info on _what_ is actually done in base table operations. 
The addition was a bad (by me - shame!) idea, and we should remove it before we go GA with cdc, so no one uses it by accident.

elcallio:
Refs #7095
Fixes #7128

CDC delta!=full both relied on post-filtering
to remove generated log row and/or cells. This is inefficient.
Instead, simply check if the data should be created in the
visitors.

v2:
    * Fixed delta logs rows created (empty) even when delta == off
v3:
    * Killed delta == off (patch in series)

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -976,6 +976,13 @@ static ttl_opt get_ttl(const row_marker& rm) {
     return rm.is_expiring() ? std::optional{rm.ttl()} : std::nullopt;
 }
 
+/**
+ * Returns whether we should generate cdc delta values (beyond keys)
+ */
+static bool generate_delta_values(const schema& s) {
+    return s.cdc_options().get_delta_mode() == cdc::delta_mode::full;
+}
+
 /* Visits the cells and tombstones of a single base mutation row and constructs corresponding delta-row cells
  * for the corresponding log mutation.
  *
@@ -1007,13 +1014,16 @@ struct process_row_visitor {
     // We need to keep a reference to it since we might insert new row_states during the visitation.
     row_states_map& _clustering_row_states;
 
+    const bool _generate_delta_values = true; 
+
     process_row_visitor(
             const clustering_key& log_ck, stats::part_type_set& touched_parts, log_mutation_builder& builder,
             bool enable_updating_state, const clustering_key* base_ck, cell_map* row_state,
-            row_states_map& clustering_row_states)
+            row_states_map& clustering_row_states, bool generate_delta_values)
         : _log_ck(log_ck), _touched_parts(touched_parts), _builder(builder),
           _enable_updating_state(enable_updating_state), _base_ck(base_ck), _row_state(row_state),
-          _clustering_row_states(clustering_row_states)
+          _clustering_row_states(clustering_row_states),
+          _generate_delta_values(generate_delta_values)
     {}
 
     void update_row_state(const column_definition& cdef, bytes_opt value) {
@@ -1031,7 +1041,9 @@ struct process_row_visitor {
         bytes value = get_bytes(cell);
 
         // delta
-        _builder.set_value(_log_ck, cdef, value);
+        if (_generate_delta_values) {
+            _builder.set_value(_log_ck, cdef, value);
+        }
 
         // images
         if (_enable_updating_state) {
@@ -1041,8 +1053,9 @@ struct process_row_visitor {
 
     void dead_atomic_cell(const column_definition& cdef, const atomic_cell_view&) {
         // delta
-        _builder.set_deleted(_log_ck, cdef);
-
+        if (_generate_delta_values) {
+            _builder.set_deleted(_log_ck, cdef);
+        }
         // images
         if (_enable_updating_state) {
             update_row_state(cdef, std::nullopt);
@@ -1174,16 +1187,18 @@ struct process_row_visitor {
         }
 
         // delta
-        if (is_column_delete) {
-            _builder.set_deleted(_log_ck, cdef);
-        }
+        if (_generate_delta_values) {
+            if (is_column_delete) {
+                _builder.set_deleted(_log_ck, cdef);
+            }
 
-        if (deleted_elements) {
-            _builder.set_deleted_elements(_log_ck, cdef, *deleted_elements);
-        }
+            if (deleted_elements) {
+                _builder.set_deleted_elements(_log_ck, cdef, *deleted_elements);
+            }
 
-        if (added_cells) {
-            _builder.set_value(_log_ck, cdef, *added_cells);
+            if (added_cells) {
+                _builder.set_value(_log_ck, cdef, *added_cells);
+            }
         }
 
         // images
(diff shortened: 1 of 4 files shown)


test/cql/cdc_delta_modes_test.result
```diff
@@ -24,7 +24,7 @@ select "cdc$batch_seq_no", "cdc$operation", pk, ck, i1 from tb2_scylla_cdc_log w
 			"pk" : "1"
 		},
 		{
-			"cdc$batch_seq_no" : "1",
+			"cdc$batch_seq_no" : "2",
```
nyh, line 27: Hmm, this change in the test indicates the previous code had a bug, which you fixed now?

nyh (approved):
Looks good to me, but please have bigger CDC experts than me review it too.

test/cql/cdc_delta_modes_test.result
elcallio: Nope. That shows that there is a bug in my code. We are inc:ing seq:no one time to many. Woops.

elcallio: fixed.

cdc/log.cc
```diff
@@ -840,6 +840,13 @@ utils::UUID generate_timeuuid(api::timestamp_type t) {
     return utils::UUID_gen::get_random_time_UUID_from_micros(t);
 }
 
+/**
+ * Returns whether we should generate cdc delta values (beyond keys)
+ */
+static bool generate_delta_values(const schema& s) {
+    return s.cdc_options().get_delta_mode() == cdc::delta_mode::full;
```
haaawk, line 847: I think this should be decided once when builder is created and not checked every time we want to add something to the result

cdc/log.cc
```diff
@@ -1031,7 +1038,9 @@ struct process_row_visitor {
         bytes value = get_bytes(cell);
 
         // delta
-        _builder.set_value(_log_ck, cdef, value);
+        if (generate_delta_values(_builder.base_schema())) {
```
haaawk, line 1041: I think it would be nicer to have this check hidden inside the builder.

haaawk: All those checks.

cdc/log.cc
```diff
@@ -1226,14 +1238,14 @@ struct process_change_visitor {
                 _enable_updating_state, nullptr, &_static_row_state, _clustering_row_states);
         visit_row_cells(v);
 
+        _builder.set_operation(log_ck, operation::update);
```
haaawk, line 1241: What's this and how is this related to this commit?

haaawk:
`cdc: Remove cdc delta_mode::off` looks good to me.

I've got mixed fillings about `cdc: Remove post-filterings for keys-only/off cdc delta generation`. Does it make any difference? Could you generate flamegraphs with and without this commit so that we could see if it's worth doing at all?

cdc/log.cc
elcallio: rebase error. removing.

cdc/log.cc
haaawk: Or when the visitor is created if we decide not to hide checks inside the builder.

elcallio: I find it hard to believe that an if per value is gonna cost more than for example even more templating -> code bloat -> icache misses and flushes. 

If a profile really shows that the checks adds measurable cost, we can consider making it embedded (templated), but until then I think that is a premature opt.

elcallio:
@haaawk I assume by flame graph, you mean profile runs. Possibly, but running what? Do we have a standard benchmark for this?
Again, I think you are  _way_ to worried about a few mostly-true branches. Premature opts are the enemy of all things. Except wasting time and code bloat.

cdc/log.cc
haaawk: What I mean is that the visitor (or the builder) should have `bool add_full_deltas = s.cdc_options().get_delta_mode() == cdc::delta_mode::full` set in the constructor and then the checks should be just `if (add_full_deltas) { ...`

haaawk:
> Premature opts are the enemy of all things.

@elcallio exactly my point. I'm saying that `cdc: Remove post-filterings for keys-only/off cdc delta generation` is a premature opts

elcallio:
Yes and no. It has the benefit of simplifying delta==keys, which is good (though to be fair, once delta==off is removed, even the filtrering becomes fairly easy). 
But the filter approach also has the downside of using potentially way more memory than simply not generating the memory in the first place. Which I think has way more performance issues, as it also contaminates caches. (As the saying goes: cache is king!)
How about a compromise: I'll move the check to a local var (delta==off gone makes all code nicer) - that saves two pointer chases at least, and should make the branches virtually costless.

cdc/log.cc
kbr-: `const bool` even! Good for readability.

cdc/log.cc
```diff
@@ -1311,11 +1323,10 @@ struct process_change_visitor {
 
     void partition_delete(const tombstone&) {
         _touched_parts.set<stats::part_type::PARTITION_DELETE>();
-
-        auto log_ck = _builder.allocate_new_log_row(operation::partition_delete);
         if (_enable_updating_state) {
             _clustering_row_states.clear();
         }
+        auto log_ck = _builder.allocate_new_log_row(operation::partition_delete);
```
kbr-, line 1329: why the shuffling around?

elcallio: gah, rebase again. fixing.

haaawk:
I'm not worried about the performance of your change (even though a bool variable computed once would be good). I'm more worried about mixing the logic that used to be fully separated with everything else. I don't know how to fix it now. On one hand, pushing the checks down to the builder would hide it from higher abstraction levels but then the builder wouldn't be a single purpose class for building the result any more.

cdc/log.cc
elcallio: Right, ok, agree. Fixing.

kbr-:
Have faith @haaawk. I have some further refactoring plans which might fix your worries.

elcallio:
Moved the check to outermost struct. Now super const member vars.

cdc/log.cc
```diff
@@ -1242,20 +1250,12 @@ struct process_change_visitor {
...
 
-            clustering_row_cells_visitor(
-                    const clustering_key& log_ck, stats::part_type_set& touched_parts, log_mutation_builder& builder,
-                    bool enable_updating_state, const clustering_key* base_ck, cell_map* row_state,
-                    row_states_map& clustering_row_states)
-                : process_row_visitor(
-                        log_ck, touched_parts, builder,
-                        enable_updating_state, base_ck, row_state, clustering_row_states)
-            {}
+            using process_row_visitor::process_row_visitor;
```
kbr-, line 1258: Thanks, I forgot this feature exists.

kbr- (approved):

haaawk (approved):
LGTM

haaawk:
Thanks @elcallio

nyh:
@elcallio please rebase, github tells me there are conflicts in the test files

elcallio:
rebased

nyh:
Thanks. Merged to next: f3dfb6e01196c3792e720c4ff6973a4ea53b138d

PR #7124, opened 2020-08-26
cdc: rename CDC description tables... again

kbr-:
Commit a6ad70d3da106ef34764098faadbed6b1c19e04c changed the format of
stream IDs: the lower 8 bytes were previously generated randomly, now
some of them have semantics. In particular, the least significant byte
contains a version (stream IDs might evolve with further releases).

This is a backward-incompatible change: the code won't properly handle
stream IDs with all lower 8 bytes generated randomly. To protect us from
subtle bugs, the code has an assertion that checks the stream ID's
version.

This means that if an experimental user used CDC before the change and
then upgraded, they might hit the assertion when a node attempts to
retrieve a CDC generation with old stream IDs from the CDC description
tables and then decode it.
In effect, the user won't even be able to start a node.

Similarly as with the case described in
d89b7a0548a9ec60efc5bf456de2396f69341856, the simplest fix is to rename
the tables. This fix must get merged in before CDC goes out of
experimental.

Now, if the user upgrades their cluster from a pre-rename version, the
node will simply complain that it can't obtain the CDC generation
instead of preventing the cluster from working. The user will be able to
use CDC after running checkAndRepairCDCStreams.

Since a new table is added to the system_distributed keyspace, the
cluster's schema has changed, so sstables and digests need to be
regenerated for schema_digest_test.


elcallio:
Could we maybe call the tables something like `cdc_topology_streams` and `cdc_topology_generations` instead or just `..._2`? To make it look a little less lame... :-/

kbr-:
This "topology" part is also lame :P

śr., 26 sie 2020 o 15:22 Calle Wilund <notifications@github.com> napisał(a):

elcallio:
Very much so. But it was the best, slightly topical, I could come up with.

kbr-:
how about cdc_generation_descriptions and cdc_streams_descriptions?

śr., 26 sie 2020 o 15:53 Calle Wilund <notifications@github.com> napisał(a):

PR #7129, opened 2020-08-26
cdc: fix deadlock inside check_and_repair_cdc_streams

kbr-:
check_and_repair_cdc_streams, in case it decides to create a new CDC
generation, updates the STATUS application state so that other nodes
gossiped with pick up the generation change.

The node which runs check_and_repair_cdc_streams also learns about a
generation change: the STATUS update causes a notification change.
This happens during add_local_application_state call
which caused the STATUS update; it would lead to calling
handle_cdc_generation, which detects a generation change and calls
add_local_application_state with the new generation's timestamp.

Thus, we get a recursive add_local_application_state call. Unforunately,
the function takes a lock before doing on_change notifications, so we
get a deadlock.

This commit prevents the deadlock.
We update the local variable which stores the generation timestamp
before updating STATUS, so handle_cdc_generation won't consider
the observed generation to be new, hence it won't perform the recursive
add_local_application_state call.

Fixes https://github.com/scylladb/scylla/issues/7127.

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -856,12 +856,14 @@ future<> storage_service::check_and_repair_cdc_streams() {
             cdc_log.error("Aborting CDC generation repair due to missing STATUS");
             return;
         }
+        // Update _cdc_streams_ts first, so that do_handle_cdc_generation (which will get called due to the status update)
+        // won't try to update the gossiper, which would result in a deadlock inside add_local_application_state
+        _cdc_streams_ts = new_streams_ts;
         _gossiper.add_local_application_state({
                 { gms::application_state::CDC_STREAMS_TIMESTAMP, versioned_value::cdc_streams_timestamp(new_streams_ts) },
                 { gms::application_state::STATUS, *status }
         }).get();
         db::system_keyspace::update_cdc_streams_timestamp(new_streams_ts).get();
-        _cdc_streams_ts = new_streams_ts;
     });
 }
 


service/storage_service.cc
```diff
@@ -856,12 +856,14 @@ future<> storage_service::check_and_repair_cdc_streams() {
             cdc_log.error("Aborting CDC generation repair due to missing STATUS");
```
haaawk, line 856: Please explain in a commit message what's going on.

haaawk (approved):
LGTM. Would it be possible to test it in a reasonable way?

kbr-:
Updated commit message.

@haaawk maybe I can get a "unit" test that checks it, lemme see

kbr-:
I figured a single-node-dtest will be better.

kbr-:
Dtest: https://github.com/scylladb/scylla-dtest/pull/1656

@nyh could you look at this?

nyh:
Committed to next: f655fc86d09ea0d6549c09384d5f51a0f5a5e04e

kbr-:
Thanks

kbr-:
@nyh this should also be backported to 4.2

nyh:
Ok. The process is that once the patch reaches master, the issue you mentioned as "Fixes" will get closed, and the issue will get the "backport candidate" label. Eventually one of the maintainers (Avi, me, etc.) will get to that issue, notice you wrote there (not here) it should be backported to 4.2, and do it.
If you want this specific backport to be done more promptly, please wait until your patch reaches master and the issue gets the "backport candidate" tag, and then remind me of this issue.

kbr-:
@nyh it looks like the fix did not reach master. What happened?

kbr-:
Github tells me that https://github.com/scylladb/scylla/commit/f655fc86d09ea0d6549c09384d5f51a0f5a5e04e is not a part of any branch

nyh:
I'm committing it again. I had a problem with compiling master because of fmt changes, but @avikivity beat me to fixing it ;-)

nyh:
Pushed to next again. 42fb4fe37ce745612b9d60ec9ecb5718747603d0

nyh:
Oops, just realised the "Fixes" line is missing on the patch (you only put it on the pull request header... it is not used in a single-patch commit....). So we'll have to manually close the issue, it won't happen automatically :-(

nyh:
Next promotion is failing because of a (most likely) different patch, I sent a mail with the patch which I suspect. In the meantime, this patch is not in master yet. We need to keep following this, so I'm not closing this PR otherwise I'll surely forget to follow it.

nyh:
Finally reached master. I'll close this PR and close the issue manually now.

PR #7132, opened 2020-08-27
offline installation: fix default value of python3/sysconfdir and a small cleanup

amoskong:
1. scylla setup scripts doesn't work, because python3 path isn't correct in creating relocatable packages if user doesn't set python3 path in running unified/install.sh. This PR fixed the problem by setting the default python3 value according to the install mode (root/nonroot).

2. another patch cleaned the tmp scylla.yaml after installation

### DIFF
--- a/unified/install.sh
+++ b/unified/install.sh
@@ -40,8 +40,6 @@ EOF
 
 root=/
 housekeeping=false
-python3=/opt/scylladb/python3/bin/python3
-sysconfdir=/etc/sysconfig
 nonroot=false
 
 while [ $# -gt 0 ]; do
@@ -88,6 +86,23 @@ if [ -z "$prefix" ]; then
     fi
 fi
 rprefix=$(realpath -m "$root/$prefix")
+
+if [ -f "/etc/os-release" ]; then
+    . /etc/os-release
+fi
+
+if [ -z "$sysconfdir" ]; then
+    sysconfdir=/etc/sysconfig
+    if ! $nonroot; then
+        if [ "$ID" = "ubuntu" ] || [ "$ID" = "debian" ]; then
+            sysconfdir=/etc/default
+        fi
+    fi
+fi
+
+if [ -z "$python3" ]; then
+    python3=$prefix/python3/bin/python3
+fi
 
 scylla_args=()
 args=()
--- a/install.sh
+++ b/install.sh
@@ -42,8 +42,6 @@ EOF
 
 root=/
 housekeeping=false
-python3=/opt/scylladb/python3/bin/python3
-sysconfdir=/etc/sysconfig
 nonroot=false
 packaging=false
 upgrade=false
@@ -179,6 +177,20 @@ if [ -z "$prefix" ]; then
 fi
 
 rprefix=$(realpath -m "$root/$prefix")
+
+if [ -f "/etc/os-release" ]; then
+    . /etc/os-release
+fi
+
+if [ -z "$sysconfdir" ]; then
+    sysconfdir=/etc/sysconfig
+    if ! $nonroot; then
+        if [ "$ID" = "ubuntu" ] || [ "$ID" = "debian" ]; then
+            sysconfdir=/etc/default
+        fi
+    fi
+fi
+
 if [ -z "$python3" ]; then
     python3=$prefix/python3/bin/python3
 fi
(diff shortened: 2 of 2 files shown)


penberg:
@syuu1228 Please review this.

unified/install.sh
```diff
@@ -88,6 +95,22 @@ if [ -z "$prefix" ]; then
     fi
 fi
 
+if [ -z "$python3" ]; then
+    if $nonroot; then
+        python3=~/scylladb/python3/bin/python3
+    else
+        python3=/opt/scylladb/python3/bin/python3
```
avikivity, line 102: should use "$prefix/" as a prefix, so users can override the default for all files with just --prefix.

unified/install.sh
```diff
@@ -88,6 +95,22 @@ if [ -z "$prefix" ]; then
...
+fi
+
+if [ -z "$sysconfdir" ]; then
+    if [ "$ID" = "ubuntu" ] || [ "$ID" = "debian" ]; then
+        sysconfdir=/etc/default
+    else
+        sysconfdir=/etc/sysconfig
+    fi
+fi
+
```
avikivity, line 113: How does this work in nonroot mode?

amoskong: In nonroot mode, $rprefix (real patch of prefix) will be added for access sysconfig files.
So it's not a problem.

unified/install.sh
amoskong: good catch! fixed.

unified/install.sh
amoskong: For example, in install.sh (https://github.com/scylladb/scylla/blob/master/install.sh#L186), it get the final real paths by following code:

```
if ! $nonroot; then
    retc="$root/etc"
    rsysconfdir="$root/$sysconfdir"
    rsystemd="$rusr/lib/systemd/system"
    ....
else
    retc="$rprefix/etc"
    rsysconfdir="$rprefix/$sysconfdir"
    rsystemd="$retc/systemd"
    ....
fi
```

unified/install.sh
```diff
@@ -88,6 +86,31 @@ if [ -z "$prefix" ]; then
...
```
syuu1228, line 113: Current script designed to use /etc/sysconfig even on Debian variants when --sysconfdir is not specified, on offline & nonroot install.
We can change this, but we should apply this on install.sh not just unified/install.sh, just like --python3 option.
And also install.sh on scylla-tools & scylla-jmx & scylla-python3, too.

unified/install.sh
```diff
@@ -88,6 +86,31 @@ if [ -z "$prefix" ]; then
...
+if [ -f "/etc/os-release" ]; then
+    . /etc/os-release
+elif [ -f "/etc/arch-release" ]; then
+    export ID=arch
+else
+    echo "/etc/os-release missing."
+    exit 1
+fi
```
syuu1228, line 104: We probably don't need /etc/arch-release since we just want to detect Debian variants.
Also, We don't have to abort setup when /etc/os-release missing, we just able to apply /etc/sysconfig as default.
So I guess
```
if [ -f "/etc/os-release" ]; then
    . /etc/os-release
fi
```
is enough for this.
If there is no file, ID is empty, sysconfdir will be /etc/sysconfig.

syuu1228 (changes requested):

unified/install.sh
```diff
@@ -88,6 +86,31 @@ if [ -z "$prefix" ]; then
...
 fi
 
+if [ -z "$python3" ]; then
+    if $nonroot; then
+        python3=$prefix/python3/bin/python3
+    else
+        python3=$prefix/python3/bin/python3
+    fi
+fi
+
```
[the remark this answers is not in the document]
amoskong: Seriously it's not a bug, just some UX problem.
I will fix both install.sh and unified/install.sh

unified/install.sh
amoskong: OK. fixed.

unified/install.sh
amoskong: Let's continue using /etc/sysconfig for root install of debian-like distro.
I will change this back.

syuu1228 (approved):
LGTM

avikivity:
Do those patches fix bugs? If so add Fixes #xxxx to individual commits.

amoskong:
> Do those patches fix bugs? If so add Fixes #xxxx to individual commits.

The second patch fixed https://github.com/scylladb/scylla/issues/7130, just updated the commit log.

unified/install.sh
avikivity: Yes.

avikivity: We should move away from /etc/sysconfig and /etc/default and put all such files into /etc/scylla. But not in this series.

PR #7135, opened 2020-08-27
storage_service.cc: get_natural_endpoints should translate key

### LINKED ISSUE #7134: nodetool getendpoints return random 3 nodes instead of expected PR nodes for PK
nodetool getendpoints does not translate the key correctly and return the wrong nodes.

amnonh:
The get_natural_endpoints returns the list of nodes holding a key.

There is a variation of the method that gets the key as string, the
current implementation just cast the string to bytes_view, which will
not work. Instead, this patch changes the implementation to use
from_nodetool_style_string to translate the key (in a nodetool like
format) to a token.
    
Fixes #7134

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -2985,8 +2985,9 @@ storage_service::get_all_ranges(const std::vector<token>& sorted_tokens) const {
 std::vector<gms::inet_address>
 storage_service::get_natural_endpoints(const sstring& keyspace,
         const sstring& cf, const sstring& key) const {
-    sstables::key_view key_view = sstables::key_view(bytes_view(reinterpret_cast<const signed char*>(key.c_str()), key.size()));
-    dht::token token = _db.local().find_schema(keyspace, cf)->get_partitioner().get_token(key_view);
+    auto schema = _db.local().find_schema(keyspace, cf);
+    partition_key pk = partition_key::from_nodetool_style_string(schema, key);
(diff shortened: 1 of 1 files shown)


service/storage_service.cc
```diff
@@ -2985,8 +2985,9 @@ storage_service::get_all_ranges(const std::vector<token>& sorted_tokens) const {
 std::vector<gms::inet_address>
 storage_service::get_natural_endpoints(const sstring& keyspace,
         const sstring& cf, const sstring& key) const {
-    sstables::key_view key_view = sstables::key_view(bytes_view(reinterpret_cast<const signed char*>(key.c_str()), key.size()));
-    dht::token token = _db.local().find_schema(keyspace, cf)->get_partitioner().get_token(key_view);
+    auto schema = _db.local().find_schema(keyspace, cf);
+    partition_key pk = partition_key::from_nodetool_style_string(schema, key);
```
avikivity, line 2989: It looks like the problem was that the key was converted using the wrong method, no?

amnonh: Exactly, it was cast to bytes_view but that would not work for integer or compound key

avikivity: ok, please adjust the changelog to be more clear.

amnonh: @avikivity I've updated the commit log to make it clearer what the change is

PR #7143, opened 2020-08-29
cql3: Use one expression in restrictions_filter

dekimir:
Replace complex logic in restrictions_filter::do_filter with a simple
evaluation of a single expression compiled from the restrictions.

This approach separates compilation from evaluation, allowing us to
easily add optimizations (eg, binding markers, compiling LIKE patterns
into regular expressions, constant folding) in the future.

Tests: unit (dev)

### DIFF
--- a/cql3/selection/selection.cc
+++ b/cql3/selection/selection.cc
@@ -42,6 +42,8 @@
 #include <boost/range/adaptors.hpp>
 #include <boost/range/algorithm/equal.hpp>
 #include <boost/range/algorithm/transform.hpp>
+#include <ranges>
+#include <seastar/util/variant_utils.hh>
 
 #include "cql3/selection/selection.hh"
 #include "cql3/selection/selector_factories.hh"
@@ -386,17 +388,61 @@ std::unique_ptr<result_set> result_set_builder::build() {
     return std::move(_result_set);
 }
 
-result_set_builder::restrictions_filter::restrictions_filter(::shared_ptr<restrictions::statement_restrictions> restrictions,
+namespace {
+
+using namespace expr;
+
+/// A visitor that copies an expression, but replaces token operator and unselected-column operators with the
+/// TRUE literal.
+struct drop_token_and_unselected_columns {
+    const selection& sel;
+
+    expression operator()(const conjunction& c) const {
+        auto r = c.children | std::views::transform([this] (const expression& e) { return std::visit(*this, e); });
+        return conjunction{std::vector<expression>(r.begin(), r.end())};
+    }
+
+    expression operator()(const binary_operator& binop) const {
+        return seastar::visit(binop.lhs,
+                [] (const token&) { return expression(true); },
+                [&] (const column_value& cv) {
+                    return sel.has_column(*cv.col) ? expression(binop) : expression(true);
+                },
+                [&] (auto&& default_case) { return expression(binop); });
+    }
+
+    expression operator()(bool b) const { return b; }
+};
+
+/// Compiles an expression equivalent to the given restrictions.  A filter can evaluate this expression on each
+/// row to satisfy the restrictions.
+expression compile(
+        const restrictions::statement_restrictions& restrictions,
+        const query_options& options,
+        const selection& sel) {
+    auto ex = restrictions.get_partition_key_restrictions()->expression;
+    if (restrictions.ck_restrictions_need_filtering()) {
+        ex = make_conjunction(std::move(ex), restrictions.get_clustering_columns_restrictions()->expression);
+    }
+    for (auto r : restrictions.get_non_pk_restriction()) {
+        ex = make_conjunction(std::move(ex), r.second->expression);
+    }
+    // TODO: Bind markers here instead of in do_filter, which redoes it for each row.
+    return std::visit(drop_token_and_unselected_columns{sel}, ex);
+}
+
+} // anonymous namespace
+
+result_set_builder::restrictions_filter::restrictions_filter(const restrictions::statement_restrictions& restrictions,
         const query_options& options,
         uint64_t remaining,
         schema_ptr schema,
         uint64_t per_partition_limit,
+        const selection& sel,
         std::optional<partition_key> last_pkey,
         uint64_t rows_fetched_for_last_partition)
-    : _restrictions(restrictions)
+    : _expression(compile(restrictions, options, sel))
     , _options(options)
-    , _skip_pk_restrictions(!_restrictions->pk_restrictions_need_filtering())
-    , _skip_ck_restrictions(!_restrictions->ck_restrictions_need_filtering())
     , _remaining(remaining)
     , _schema(schema)
     , _per_partition_limit(per_partition_limit)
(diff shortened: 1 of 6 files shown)


cql3/selection/selection.cc
```diff
@@ -386,17 +388,61 @@ std::unique_ptr<result_set> result_set_builder::build() {
...
 }
 
-result_set_builder::restrictions_filter::restrictions_filter(::shared_ptr<restrictions::statement_restrictions> restrictions,
+namespace {
+
+using namespace expr;
+
+/// A visitor that copies an expression, but replaces token operator and unselected-column operators with the
+/// TRUE literal.
+struct drop_token_and_unselected_columns {
```
avikivity, line 397: How feasible is it to combine the drop with the extraction of the primary addressing path (primary key invtervals)?

These two parts have to be kept in sync, otherwise we can have something dropped that isn't taken care of by the other side (missing data) or not drop something from both paths (efficiently loss). IOW we must keep it either-or.

Something like

struct query_plan {
    expression primary_key_lookup;
    expression filter;
};

query_plan split_expression_to_primary_ley_lookup_and_filter(expression);

cql3/selection/selection.cc
```diff
@@ -386,17 +388,61 @@ std::unique_ptr<result_set> result_set_builder::build() {
...
+    const selection& sel;
+
+    expression operator()(const conjunction& c) const {
+        auto r = c.children | std::views::transform([this] (const expression& e) { return std::visit(*this, e); });
+        return conjunction{std::vector<expression>(r.begin(), r.end())};
```
avikivity, line 402: I am filled with indignation that a vector(const auto std::range) consructor doesn't exist. We could then write 

```c++
    return conjunction{c.children | std::views::transform([this] (const expression& e) { return std::visit(*this, e); })};
```

avikivity: We could even have written

```c++
    return conjunction{c.children | std::views::transform(std::bind_front(std::visit, std::ref(*this)))};
```

Except that std::visit is a template and so a second-class object. We'd be getting close to 5X bloat from the equivalent Haskell or lisp.

cql3/selection/selection.cc
```diff
@@ -386,17 +388,61 @@ std::unique_ptr<result_set> result_set_builder::build() {
...
+    }
+
+    expression operator()(const binary_operator& binop) const {
+        return seastar::visit(binop.lhs,
+                [] (const token&) { return expression(true); },
```
avikivity, line 407: What about something like `token(x) != :marker`? It still has to be filtered. Or maybe it's translated to two intervals? Or not supported at all?

avikivity: Maybe I'm missing why we want to drop tokens and unselected columns from the expression.

cql3/selection/selection.cc
```diff
@@ -386,17 +388,61 @@ std::unique_ptr<result_set> result_set_builder::build() {
...
+    }
+
+    expression operator()(bool b) const { return b; }
+};
+
+/// Compiles an expression equivalent to the given restrictions.  A filter can evaluate this expression on each
+/// row to satisfy the restrictions.
+expression compile(
+        const restrictions::statement_restrictions& restrictions,
+        const query_options& options,
```
avikivity, line 421: If we get query_options here, how is this compilation? It's performed on every query.

avikivity:
I need some more high-level explanation in comments about the goal of the new functions. They're either too low-level (drop_this_and_that) or too abstract (compile). I need something that describes the immediate goals of the functions.

I suppose I could read the patch as a transformation of the existing code, but it helps to understand what it does, and it will help future readers too.

cql3/selection/selection.cc
```diff
@@ -386,17 +388,61 @@ std::unique_ptr<result_set> result_set_builder::build() {
...
+        return conjunction{std::vector<expression>(r.begin(), r.end())};
+    }
+
+    expression operator()(const binary_operator& binop) const {
+        return seastar::visit(binop.lhs,
+                [] (const token&) { return expression(true); },
+                [&] (const column_value& cv) {
+                    return sel.has_column(*cv.col) ? expression(binop) : expression(true);
+                },
+                [&] (auto&& default_case) { return expression(binop); });
```
avikivity, line 411: These defaults are dangerous, if you add a new varient member, it is easy to miss a visitor that you should have modified.

cql3/selection/selection.cc
avikivity: In fact you don't use options, which suggests the caller could be during prepare (doesn't have to be in this patch).

cql3/selection/selection.cc
```diff
@@ -386,17 +388,61 @@ std::unique_ptr<result_set> result_set_builder::build() {
...
+/// Compiles an expression equivalent to the given restrictions.  A filter can evaluate this expression on each
+/// row to satisfy the restrictions.
+expression compile(
+        const restrictions::statement_restrictions& restrictions,
+        const query_options& options,
+        const selection& sel) {
+    auto ex = restrictions.get_partition_key_restrictions()->expression;
+    if (restrictions.ck_restrictions_need_filtering()) {
+        ex = make_conjunction(std::move(ex), restrictions.get_clustering_columns_restrictions()->expression);
+    }
```
avikivity, line 426: Why not do it unconditionally? wouldn't rhs be true if the condition is false?

cql3/selection/selection.cc
```diff
@@ -386,17 +388,61 @@ std::unique_ptr<result_set> result_set_builder::build() {
...
+    for (auto r : restrictions.get_non_pk_restriction()) {
+        ex = make_conjunction(std::move(ex), r.second->expression);
+    }
+    // TODO: Bind markers here instead of in do_filter, which redoes it for each row.
+    return std::visit(drop_token_and_unselected_columns{sel}, ex);
```
avikivity, line 431: I'm puzzled by this due to missing high-level infromation. Here you merge several expressions parts into one. I'd expect the direction to be opposite: the grammar generates one big expression (doesn't today) and the planner works to split it according to what we can execute.

avikivity:
While I'm puzziled by the new code, it's a demonstration of how we improved - the existing code is impenetrable, the new code can be understood enough to ask questions about.

cql3/selection/selection.cc
dekimir: We wouldn't even get to code generation, because bind_front can't pick the right overload: https://godbolt.org/z/ozPvad

avikivity: Yes, because it's a template, as I mentioned.

cql3/selection/selection.cc
dekimir: Interesting question but irrelevant to do_filter().  This code only deals with post-fetch filtering, so the primary addressing path is irrelevant.  Its only aim is to ensure correct filtering of rows already fetched.

cql3/selection/selection.cc
dekimir: > What about something like `token(x) != :marker`?

Forbidden.
`InvalidRequest: Error from server: code=2200 [Invalid query] message="Unsupported "!=" relation: token(p) != 2"` 

> Maybe I'm missing why we want to drop tokens and unselected columns from the expression.

Tokens are used to prepare the fetch command; those subexpressions can never be false on the fetched rows.  Unselected columns will not be fetched, so we cannot get their values to evaluate the filter expression.  See old code:

https://sourcegraph.com/github.com/scylladb/scylla/-/blob/cql3/selection/selection.cc#L430

cql3/selection/selection.cc
dekimir: > If we get query_options here, how is this compilation? It's performed on every query.

That doesn't make it not-compilation. :) There are a few phases in every query's life:
1. preparation
2. execution
 2.1. compute bounds
 2.2. fetch rows
 2.3. create restrictions_filter
 2.4. for each fetched row, apply the restrictions_filter

The code above is in phase 2.3, which happens on every query but not every row.  By compiling (or optimizing, I don't care what we call it) the query here, we avoid repetitive tasks in 2.4.  And we need query_options to turn markers into concrete values (eg, so we can compile regexes for LIKE patterns).

> In fact you don't use options

I'll take `options` out of this patch, then add it right back in the next one, which replaces LIKE nodes with regex match.

cql3/selection/selection.cc
dekimir: True in general, but this particular visitor's contract is that it only transforms tokens and some single-column restrictions.  I don't want to have to rewrite it on every irrelevant IR modification.

cql3/selection/selection.cc
dekimir: > Why not do it unconditionally?

Because then tests fail. :)  This is equivalent to the [old code](https://sourcegraph.com/github.com/scylladb/scylla/-/blob/cql3/selection/selection.cc#L399:7&tab=references).

cql3/selection/selection.cc
dekimir: I share that vision, but it'll take a few more steps to get there.  Right now I have to deal with multiple expressions coming from upstream, and I want to streamline and optimize their evaluation.  Tomorrow when the upstream hands me a single expression, I'll drop the merging but keep the optimization.

cql3/selection/selection.cc
avikivity: But how does it know which parts of the expression are safe to drop? I think it only knows it by relying on knowledge of how the primary addressing path works.

cql3/selection/selection.cc
avikivity: Can we have an expression on an unselected column? I can think of two cases:

 1. WHERE clause, but constraint is on a pk/ck column so moved to slice and not added to selection
 2. WHERE clause, constraint on regular column, added to selection even if not used.

Is this correct? Either way, please add a comment explaining why we drop unselected columns and why it's safe.

cql3/selection/selection.cc
avikivity: Ok. I think we'll want to move this to the prepare phase eventually, and have an extra, optional step for expressions that have non-literal LIKE patterns. We don't want to penalize all other expressions for that.

How far are we from generating expressions directly from the parser?

cql3/selection/selection.cc
avikivity: Ok.

cql3/selection/selection.cc
avikivity: But that old code is impossible to understand. I suppose I'll just have to suffer until it's gone.

cql3/selection/selection.cc
avikivity: Ok.

cql3/selection/selection.cc
dekimir: Correct. Perhaps it would be more logical to do this split upstream and only pass the filter here. Let me see how hard that is to code...

avikivity: I'm fine with not doing everything at once. What I want is
 - understand the path to enlightenment
 - add comments wherever we're doing something strange/weird/suboptimal, even if we're just keeping the status quo of the OO-style code

cql3/selection/selection.cc
dekimir: This is roughly correct (though note that some pk restrictions still require filtering, and it makes a difference if there are unrestricted primary columns). I'll add a comment.

cql3/selection/selection.cc
dekimir: > How far are we from generating expressions directly from the parser?

I was going to work on that next, after this patch and [translating LIKE to compiled regex](https://github.com/dekimir/scylla/commits/match_regex-in-ast2).  So I'm guessing start in a couple of days, and finish... IDK when, depends on landmines I uncover along the way. ;)

cql3/selection/selection.cc
dekimir: You and me both.  Sometimes I skip the rabbit hole of understanding and settle for mimicking, like here.  It's more efficient. :(

slivne:
@dekimir is a new version beeing worked - or did we move this o a different pull request - if this is not relevant please close

dekimir:
@slivne We're adopting an [alternative approach](#7217), which seems to me a well-engineered solution (despite turning out to be more difficult in the short term).

PR #7144, opened 2020-08-29
unified: add uninstall.sh

### LINKED ISSUE #7076: offline installer: uninstall isn't supported
*Installation details*
Scylla version (or git commit hash): latest master: scylla-unified-package-0.20200819.2e1499ee93f.tar.gz
Cluster size: 1
OS (RHEL/CentOS/Ubuntu/AWS AMI): CentOS/Debian/Ubuntu

### Description

Original https://github.com/scylladb/scylla/wiki/Offline-Installer is based on RPM/DEB, so we can easily uninstall scylla by removing the RPM/DEB packages.

The new offline installer (unified package) doesn't support to uninstall.

/CC @slivne @roydahan @penberg

syuu1228:
Provide an uninstaller for offline & nonroot installation.

Fixes #7076

### DIFF
--- a/unified/build_unified.sh
+++ b/unified/build_unified.sh
@@ -65,5 +65,6 @@ for pkg in $PKGS; do
     fi
 done
 ln -f unified/install.sh build/"$MODE"/unified/
+ln -f unified/uninstall.sh build/"$MODE"/unified/
 cd build/"$MODE"/unified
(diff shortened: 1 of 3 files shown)


syuu1228:
We probably don't want to add multiple uninstall.sh for each packages, implemented uninstaller in unified package.

PR #7149, opened 2020-08-31
repair: Add  progress metrics for replace decommission removenode

### LINKED ISSUE #1244: repair: need a method to monitor progress

asias:
This series follows "repair: Add progress metrics for node ops #6842" and adds the metrics for the remaining node operations, i.e., replace, decommission and removenode. 

Fixes #1244, #6733

### DIFF
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -64,6 +64,8 @@ public:
                 [this] { return rebuild_finished_percentage(); }),
             sm::make_gauge("decommission_finished_percentage", sm::description("Number of finished percentage for decommission operation on this shard."),
                 [this] { return decommission_finished_percentage(); }),
+            sm::make_gauge("removenode_finished_percentage", sm::description("Number of finished percentage for removenode operation on this shard."),
+                [this] { return removenode_finished_percentage(); }),
             sm::make_gauge("repair_finished_percentage", sm::description("Number of finished percentage for repair operation on this shard."),
                 [this] { return repair_finished_percentage(); }),
         });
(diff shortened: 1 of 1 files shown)


repair/repair.cc
```diff
@@ -1470,6 +1479,8 @@ static future<> do_repair_ranges(lw_shared_ptr<repair_info> ri) {
             }).then([ri] {
                 if (ri->reason == streaming::stream_reason::bootstrap) {
                     _node_ops_metrics.bootstrap_finished_ranges++;
+                } else if (ri->reason == streaming::stream_reason::replace) {
```
amnonh, line 1500: In general, I would prefer having the different types split by label and not by name.

Also, it's not a bug but I think you can implement this with an array based on the enum index, it would replace all those ifs and will be easier to extend if you'll have new reasons in the future.

Both comments are personal taste, about the metrics I may suggest such a change in the future on top of you patch if it would be merged as-is.

asias: The ifs can be replaced by a switch. I not using switch right now because not all operations are supported in the first part of this series. We have an agreement we can do the refactor as a follow up during the previous review.

amnonh: What I meant that if you had an std::array<uint64_t, 6> instead of multiple metrics, you could just use the enum as an index to that array.

One more thing, I've tried to find in the code and maybe I missed it. Are the total and finished ever cleared?
I'm trying to understand if we are looking at repair progress, is it the last repair, or is it an ever-growing repair?
if it's the last, it will be practically 1.

For operations that happen just once in a node life cycle, this is fine.

asias: Sorry for the delay. I am not sure why my comments though email did not reach here.

So, yes total and finished are cleared when the operation is started, e.g., bootstrap, rebuild. Repair is a bit different, it counts all pending repair jobs. When a job is finished, it is removed form the total and finished. This is because we can have multiple repairs jobs in parallel.

Yes, using array is one alternative, since it does not give much gain, I prefer we could merge as it is.

asias:
@avikivity  Ping for merge.

psarna (approved):

PR #7152, opened 2020-08-31
lsa: decay reserves

### LINKED ISSUE #325: Make allocation_section decay the reserves over time
Right now the reserves only grow. It's better to reduce them with time so that they adapt to changing workloads.

Potential risk with the current approach is that temporary spike in memory requirements may hurt performance onward.

avikivity:
The log-structured allocator (LSA) reserves memory when performing
operations, since its operations are performed with reclaiming disabled
and if it runs out, it cannot evict cache to gain more. The amount of
memory to reserve is remembered across calls so that it does not have
to repeat the fail/increase-reserve/retry cycle for every operation.

However, we currently lack decaying the amount to reserve. This means
that if a single operation increased the reserve in the distant past,
all current operations also require this large reserve. Large reserves
are expensive since they can cause large amounts of cache to be evicted.

This patch adds reserve decay. The time-to-decay is inversely proportional
to reserve size: 10GB/reserve. This means that a 20MB reserve is halved
after 500 operations (10GB/20MB) while a 20kB reserve is halved after
500,000 operations (10GB/20kB). So large, expensive reserves are decayed
quickly while small, inexpensive reserves are decayed slowly to reduce
the risk of allocation failures and exceptions.

Fixes #325.

### DIFF
--- a/utils/logalloc.cc
+++ b/utils/logalloc.cc
@@ -2259,6 +2259,34 @@ allocating_section::guard::~guard() {
     shard_segment_pool.set_emergency_reserve_max(_prev);
 }
 
+void allocating_section::maybe_decay_reserve() {
+    // The decay rate is inversely proportional to the reserve
+    // (every (s_segments_per_decay/_lsa_reserve) allocations).
+    //
+    // If the reserve is high, it is expensive since we may need to 
+    // evict a lot of memory to satisfy the reserve. Hence, we are
+    // willing to risk a more frequent bad_alloc in order to decay it.
+    // The cost of a bad_alloc is also lower compared to maintaining
+    // the reserve.
+    //
+    // If the reserve is low, it is not expensive to maintain, so we
+    // decay it at a lower rate.
+
+    _remaining_lsa_segments_until_decay -= _lsa_reserve;
+    if (_remaining_lsa_segments_until_decay < 0) {
+        _remaining_lsa_segments_until_decay = s_segments_per_decay;
+        _lsa_reserve = std::max(s_min_lsa_reserve, _lsa_reserve / 2);
+        llogger.debug("Decaying LSA reserve in section {} to {} segments", this, _lsa_reserve);
+    }
+
+    _remaining_std_bytes_until_decay -= _std_reserve;
+    if (_remaining_std_bytes_until_decay < 0) {
+        _remaining_std_bytes_until_decay = s_bytes_per_decay;
+        _std_reserve = std::max(s_min_std_reserve, _std_reserve / 2);
+        llogger.debug("Decaying standard allocator head-room in section {} to {} [B]", this, _std_reserve);
+    }
+}
+
 void allocating_section::reserve() {
   try {
     shard_segment_pool.set_emergency_reserve_max(std::max(_lsa_reserve, _minimum_lsa_emergency_reserve));
@@ -2287,10 +2315,10 @@ void allocating_section::reserve() {
 void allocating_section::on_alloc_failure(logalloc::region& r) {
     r.allocator().invalidate_references();
     if (shard_segment_pool.allocation_failure_flag()) {
-        _lsa_reserve *= 2; // FIXME: decay?
+        _lsa_reserve *= 2;
         llogger.debug("LSA allocation failure, increasing reserve in section {} to {} segments", this, _lsa_reserve);
     } else {
-        _std_reserve *= 2; // FIXME: decay?
+        _std_reserve *= 2;
         llogger.debug("Standard allocator failure, increasing head-room in section {} to {} [B]", this, _std_reserve);
     }
     reserve();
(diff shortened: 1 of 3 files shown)


utils/logalloc.hh
```diff
@@ -685,16 +685,24 @@ struct reclaim_lock {
 // also allocate LSA memory. The object learns from failures how much it
 // should reserve up front in order to not cause allocation failures.
 class allocating_section {
-    size_t _lsa_reserve = 10; // in segments
-    size_t _std_reserve = 1024; // in bytes
+    // Do not decay below these minimal values
+    static constexpr size_t s_min_lsa_reserve = 10;
```
tgrabiec, line 689: Is there some new convention behind the `s_` name prefix (s = static)?

utils/logalloc.cc
```diff
@@ -2259,7 +2259,36 @@ allocating_section::guard::~guard() {
...
+    // evict a lot of memory to satisfy the reserve. Hence, we are
+    // willing to risk a more frequent bad_alloc in order to decay it.
+    // The cost of a bad_alloc is also lower compared to maintaining
+    // the reserve.
+    //
+    // If the reserve is low, it is not expensive to maintain, so we
+    // decay it at a lower rate.
+
+    _remaining_lsa_segments_until_decay -= _lsa_reserve;
+    if (_remaining_lsa_segments_until_decay < 0) {
```
tgrabiec, line 2276: _remaining_lsa_segments_until_decay is unsigned, how can it go below 0?

utils/logalloc.cc
```diff
@@ -2259,7 +2259,36 @@ allocating_section::guard::~guard() {
...
+        _remaining_lsa_segments_until_decay = s_segments_per_decay;
+        _lsa_reserve = std::max(s_min_lsa_reserve, _lsa_reserve / 2);
+        llogger.debug("Decaying LSA reserve in section {} to {} segments", this, _lsa_reserve);
+    }
+
+    _remaining_std_bytes_until_decay -= _std_reserve;
+    if (_remaining_std_bytes_until_decay < 0) {
+        _remaining_std_bytes_until_decay = s_segments_per_decay;
```
tgrabiec, line 2284: Should be `s_bytes_per_decay`

utils/logalloc.cc
```diff
@@ -2259,7 +2259,36 @@ allocating_section::guard::~guard() {
...
+        _std_reserve = std::max(s_min_std_reserve, _std_reserve / 2);
+        llogger.debug("Decaying standard allocator head-room in section {} to {} [B]", this, _std_reserve);
+    }
+}
+
 void allocating_section::reserve() {
+  maybe_decay_reserve();
```
tgrabiec, line 2291: with_reclaiming_disabled() calls reserve() when retrying, in which case we should not attempt to decay, otherwise they may cancel each other and in the extreme case we'll not progress.

I suggest to call it once per with_reclaiming_disabled.

utils/logalloc.cc
```diff
@@ -2259,7 +2259,36 @@ allocating_section::guard::~guard() {
     shard_segment_pool.set_emergency_reserve_max(_prev);
 }
 
+void allocating_section::maybe_decay_reserve() {
+    // The decay rate is inversely proportional to the reserve
+    // (every (s_segments_per_decay/_lsa_reserve) allocations).
+    //
```
tgrabiec, line 2265: Approach LGTM

tgrabiec (changes requested):

utils/logalloc.hh
avikivity: Yes, not formalized.

utils/logalloc.cc
avikivity: Oops, meant to change it to be signed, but forgot.

utils/logalloc.cc
avikivity: Yes.

utils/logalloc.cc
avikivity: You mean, if the number of attempts is equal to the number of times until decay? I think it's unlikely, but will change to reduce the risk of confusion.

utils/logalloc.cc
avikivity: Thanks. I'll try to add a unit test for the next iteration.

utils/logalloc.hh
avikivity: (and not new, see s_shard_id, s_rpc_client_idx_table)

avikivity:
Update: 

 - changed _remaining_std_bytes_until_decay, _remaining_lsa_segments_until_decay to be signed, so that the checks that they are < 0 make sense
 - decay std reserve by the correct constant, not s_segments_per_decay
 - moved decay to with_reclaiming_disabled() to prevent chance of decay happening more than once per operation
 - removed FIXME: decay comments
 - add unit test

tgrabiec (approved):

avikivity:
Queued.

avikivity:
Merged as 7ac59dcc9837658bb9fc1ff06dc4c95fdabef6f8

PR #7160, opened 2020-09-02
Alternator streams - fix table description and sequence number

elcallio:
Stream descriptions, as returned from create, update and describe stream was missing "latest" stream arn. 

Shard descriptions sequence number (for us timeuuid:s) were formatted wrongly. The spec states they should be numeric only. 

Both these defects break Kinesis operations. 

v2: 
* Fixed sequence number in getrecords

### DIFF
--- a/alternator/executor.cc
+++ b/alternator/executor.cc
@@ -785,6 +785,23 @@ static future<> wait_for_schema_agreement(service::migration_manager& mm, db::ti
     });
 }
 
+static void verify_billing_mode(const rjson::value& request) {
+        // Alternator does not yet support billing or throughput limitations, but
+    // let's verify that BillingMode is at least legal.
+    std::string billing_mode = get_string_attribute(request, "BillingMode", "PROVISIONED");
+    if (billing_mode == "PAY_PER_REQUEST") {
+        if (rjson::find(request, "ProvisionedThroughput")) {
+            throw api_error::validation("When BillingMode=PAY_PER_REQUEST, ProvisionedThroughput cannot be specified.");
+        }
+    } else if (billing_mode == "PROVISIONED") {
+        if (!rjson::find(request, "ProvisionedThroughput")) {
+            throw api_error::validation("When BillingMode=PROVISIONED, ProvisionedThroughput must be specified.");
+        }
+    } else {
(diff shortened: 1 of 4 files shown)


nyh:
Thanks!
As you know I'm a sucker for writing Alternator tests, so I'll write tests for those issues, and once I verify they pass I'll commit this series (and send the tests as a patch).

elcallio:
Hang on. Found one more place where wrong type is serialized.

elcallio:
Updated.

nyh:
By the way, note that `test_streams.py::test_streams_after_sequence_number` is a fairly new test - I added it to make sure the fix for the 0x80 bug doesn't break things (and it didn't) - so it is possible your development version didn't have this test which is why you didn't notice it failing.

nyh:
@elcallio FYI I just sent to the mailing list a patch titled "[RFC PATCH] alternator test: add reproducing tests for several issues" which adds test failures (not even xfails) which this patch corrects, which is good - and you can use that patch to confirm your patches continue to fix these problems. 
But note that as I said above, you patch breaks a **different**, already-existing, test, `test_streams.py::test_streams_after_sequence_number`, so it probably still has a bug that needs to be fixed.

elcallio:
I've pushed a fix for the `test_streams_after_sequence_number` issue - turns out the compiler bound bmp::number constructor call without warning, but completely wrong. Even I think that is scary template behaviour...

elcallio:
I've added a few more changes to fix the tests you RFC:d to the mailing list. Note that these tests are not included here. Should I add them as a patch? Or do you prefer to commit them yourself?

alternator/executor.cc
```diff
@@ -1043,6 +1044,10 @@ future<executor::request_return_type> executor::update_table(client_state& clien
         }
     }
 
+    if (rjson::find(request, "BillingMode")) {
```
nyh, line 1047: Nitpick: I think you should call verify_billing_mode unconditionally, because it also checks the case that billingmode is missing and some other attributes are not allowed. But in any case, none of this is supported right now, so it doesn't matter :-)

nyh (approved):

nyh:
Thans. Committed to next: f76c519c1fc9f511383024b83f3b015833a65736

PR #7168, opened 2020-09-04
Update HACKING.md

tarzanek:
update as per https://github.com/scylladb/scylla-jmx/blob/master/README.md#compiling

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -225,7 +225,7 @@ scylla-jmx (https://github.com/scylladb/scylla-jmx). It can be build
 with
 
 ```bash
-$ mvn package
+$ mvn --file scylla-jmx-parent/pom.xml package
 ```
 
 and must be started with


HACKING.md
```diff
@@ -225,7 +225,7 @@ scylla-jmx (https://github.com/scylladb/scylla-jmx). It can be build
 with
 
 ```bash
-$ mvn package
+$ mvn --file scylla-jmx-parent/pom.xml package
```
nyh, line 228: What is scylla-jmx-parent? Since scylla-jmx is now, if I understand correctly, a submodule, you don't need to mention https://github.com/scylladb/scylla-jmx nor choose a name to clone it under, you can just use scylla-jmx/pom.xml.
Also, please update the lines below which call to run "./scripts/scylla-jmx", but this should probably be changed to "scylla-jmx/scripts/scylla-jmx".

nyh (changes requested):

HACKING.md
tarzanek: Nadav, scylla-jmx isn't really a submodule, just a component that can be part of a build
above is old way how it was built
I changed the build process so scylla-jmx can be broken to smaller parts (namely the API wrapper and client is interesting) and leveraged in scylla-java-tools repo by introducing a parent project to wrap the pieces

I didn't realize this howto also shows how to build the scylla-jmx repo, so just wanted to make both howtos in sync
(see https://github.com/scylladb/scylla-jmx/blob/master/README.md#compiling )
I don't even mind by having here a pointer to above, rather than duplicating the actual instruction ... hmm ... that is even better idea I guess

tarzanek: so I guess I will drop this PR and open a new one with a pointer to how to build rather than duplicating the build info here, what do you think @nyh ?

nyh: I just realized that "scylla-jmx" indeed isn't a submodule today - I have this directory because I checkout out an older branch which did have it. So today, I have it in "toos/jmx". So is that the directory we should list?
Please check out a branch new working directory, try to follow your own instructions, and see if works. I tried your instructions and quickly realized that "scylla-jmx-parent/pom.xml" doesn't exist :-)

tarzanek: thnx, will recheck with new state after Pekkas presso on new integrated building

PR #7169, opened 2020-09-04
utf8: Print invalid UTF-8 character position

### LINKED ISSUE #3016: Improve validation error messages
If utf8 field validation fails, the CQL client will get:
```
marshaling error: Conversion failed
```

It's cryptic. We should mention that validation failed, for which table and which column. If possible, what was the offending position in the string.

avelanarius:
#### Summary
Before this change, if there was a UTF-8 validation error, Scylla provided no further information about location of invalid character inside the offending string.

Added new `validate_with_error_position` function, which returns a position of invalid UTF-8 byte in a string. Modified all "invalid UTF-8" exception messages to present a location information to user (below - in bold highlighted added new text):

- Invalid UTF-8 string inside prepared statement argument:
`exceptions::invalid_request_exception: Exception while binding column b: marshaling error: Validation failed - non-UTF8 character in a UTF8 string,` **`at byte offset 0`**
- Server-side message returned to the Java Cassandra client:
`Exception in thread "main" com.datastax.oss.driver.api.core.servererrors.ProtocolError: Cannot decode string as UTF8` **`, invalid character at byte offset 45`**

Fixes #3016 

#### Why `validate_with_error_position` is done in two passes?
At first I considered changing existing functions to return a error position instead of bool error/no error. One of the existing implementations is a SIMD branchfree (in the main loop) method, which goes through the entire string accumulating state in `error` vector (no short-circuit!). In order to return a position I would have to:

a) change the loop to short-circuit, eliminating the branchfree aspect of the implementation
or
b) somehow accumulate the position alongside/in the `error` vector, adding additional operations which might make it slower
https://github.com/scylladb/scylla/blob/6fadba56cc18bebe6648e6aec09be1b0b93f4817/utils/utf8.cc#L461-L462 

As (probably) most strings are valid UTF-8, in those cases only a one (SIMD optimized) pass is performed. If there is an UTF8 error, in current implementation a slower naive pass is performed to find the error position - I think a good tradeoff.

### DIFF
--- a/utils/utf8.cc
+++ b/utils/utf8.cc
@@ -57,7 +57,9 @@ namespace utils {
 namespace utf8 {
 
 // 3x faster than boost utf_to_utf
-static inline bool validate_naive(const uint8_t *data, size_t len) {
+static inline std::optional<size_t> validate_naive(const uint8_t *data, size_t len) {
+    size_t pos = 0;
+
     while (len) {
         size_t bytes;
         const uint8_t byte1 = data[0];
@@ -99,20 +101,21 @@ static inline bool validate_naive(const uint8_t *data, size_t len) {
                          (byte1 == 0xF4 && byte2 <= 0x8F))) {
                     bytes = 4;
                 } else {
-                    return false;
+                    return pos;
                 }
             } else {
-                return false;
+                return pos;
             }
         } else {
-            return false;
+            return pos;
         }
 
         len -= bytes;
+        pos += bytes;
         data += bytes;
     }
 
-    return true;
+    return std::nullopt;
 }
 
 #if defined(__aarch64__)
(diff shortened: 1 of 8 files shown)


utils/utf8.cc
```diff
@@ -491,16 +494,25 @@ bool validate(const uint8_t *data, size_t len) {
...
 
 #else
 // No SIMD implementation for this arch, fallback to naive method
 bool validate(const uint8_t *data, size_t len) {
-    return validate_naive(data, len);
+    return validate_naive(data, len) == -1;
 }
 #endif
 
+ssize_t validate_with_error_position(const uint8_t *data, size_t len) {
```
penberg, line 507: You could use `std::optional<size_t>` as the return type and indicate lack of error with `std::nullopt`.

penberg (changes requested):

utils/utf8.cc
avikivity: I agree that for the public interface it's better to avoid magic return values.

avelanarius: Changed to `std::optional` in amended commit.

Found that previously I forgot to fix one unit test (checking for hardcoded exception message) - it is now fixed and all unit tests pass correctly.

https://github.com/scylladb/scylla/pull/7169/files#diff-43be90fff306ebda1a2855a78a22daa1L577-R577

PR #7173, opened 2020-09-05
Add support passing python3 dependencies from main repo to scylla-python3 script

syuu1228:
We don't want to update scylla-python3 submodule for every python3 dependency
update, bring python3 package list to python3-dependencies.txt, pass it on
package building time.

See #6702
See scylladb/scylla-python3#6

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -27,10 +27,6 @@ else
     exit 1
 fi
 
-bash seastar/install-dependencies.sh
-bash tools/jmx/install-dependencies.sh
-bash tools/java/install-dependencies.sh
-
 debian_base_packages=(
     liblua5.3-dev
     python3-pyparsing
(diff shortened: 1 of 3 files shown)


syuu1228:
This requires https://github.com/scylladb/scylla-python3/pull/8 to make it work.

install-dependencies.sh
```diff
@@ -103,6 +92,7 @@ fedora_packages=(
     fakeroot
     file
     dpkg-dev
+    $(cat ./python3-dependencies.txt)
```
avikivity, line 95: This breaks tools/toolchain/Dockerfile:

```
ADD ./install-dependencies.sh ./
ADD ./seastar/install-dependencies.sh ./seastar/
```

Now we don't have that file.

In order to avoid splitting the information across many files, you can keep the variable in install-dependencies.sh, and add a flag to extract it: `./install-dependencies --print-python3-runtime-packages` to extract just that list.

syuu1228:
@avikivity Dropped python3-dependencies.txt, merged python3 package list into install-dependencies.sh.
Please review again.

avikivity:
Queued as 5d75ecb069f7db44b0457a634edf9ed596d3de95 (with the submodule update)

PR #7178, opened 2020-09-06
storage_proxy: add a separate smp_group for hints

### LINKED ISSUE #7177: Possible priority inversion between hints and normal write in storage_proxy
Hints writes are handled by storage_proxy in the exact same way
regular writes are, which in turn means that the same smp service
group is used for both. The problem is that it can lead to a priority
inversion where writes of the lower priority  kind occupies a lot of
the semaphores units making the higher priority writes wait for an
empty slot.
The affected queries and hints are those which originated in a non
shard aware clients since only then they will be queued on the cross
shard semaphore a.k.a smp_service_group

eliransin:
Hints writes are handled by storage_proxy in the exact same way
regular writes are, which in turn means that the same smp service
group is used for both. The problem is that it can lead to a priority
inversion where writes of the lower priority  kind occupies a lot of
the semaphores units making the higher priority writes wait for an
empty slot.
This series adds a separate smp group for hints as well as a field
to pass the correct smp group to mutate_locally functions, and
then uses this field to properly classify the writes.

Fixes #7177

### DIFF
--- a/service/storage_proxy.cc
+++ b/service/storage_proxy.cc
@@ -1824,10 +1824,10 @@ storage_proxy::response_id_type storage_proxy::unique_response_handler::release(
 }
 
 future<>
-storage_proxy::mutate_locally(const mutation& m, tracing::trace_state_ptr tr_state, db::commitlog::force_sync sync, clock_type::time_point timeout) {
+storage_proxy::mutate_locally(const mutation& m, tracing::trace_state_ptr tr_state, db::commitlog::force_sync sync, clock_type::time_point timeout, smp_service_group smp_grp) {
     auto shard = _db.local().shard_of(m);
     get_stats().replica_cross_shard_ops += shard != this_shard_id();
-    return _db.invoke_on(shard, {_write_smp_service_group, timeout},
+    return _db.invoke_on(shard, {smp_grp, timeout},
             [s = global_schema_ptr(m.schema()),
              m = freeze(m),
              gtr = tracing::global_trace_state_ptr(std::move(tr_state)),
@@ -1838,24 +1838,29 @@ storage_proxy::mutate_locally(const mutation& m, tracing::trace_state_ptr tr_sta
 }
 
 future<>
-storage_proxy::mutate_locally(const schema_ptr& s, const frozen_mutation& m, tracing::trace_state_ptr tr_state, db::commitlog::force_sync sync, clock_type::time_point timeout) {
+storage_proxy::mutate_locally(const schema_ptr& s, const frozen_mutation& m, tracing::trace_state_ptr tr_state, db::commitlog::force_sync sync, clock_type::time_point timeout,
+        smp_service_group smp_grp) {
     auto shard = _db.local().shard_of(m);
     get_stats().replica_cross_shard_ops += shard != this_shard_id();
-    return _db.invoke_on(shard, {_write_smp_service_group, timeout},
+    return _db.invoke_on(shard, {smp_grp, timeout},
             [&m, gs = global_schema_ptr(s), gtr = tracing::global_trace_state_ptr(std::move(tr_state)), timeout, sync] (database& db) mutable -> future<> {
         return db.apply(gs, m, gtr.get(), sync, timeout);
     });
 }
 
 future<>
-storage_proxy::mutate_locally(std::vector<mutation> mutations, tracing::trace_state_ptr tr_state, clock_type::time_point timeout) {
-    return do_with(std::move(mutations), [this, timeout, tr_state = std::move(tr_state)] (std::vector<mutation>& pmut) mutable {
-        return parallel_for_each(pmut.begin(), pmut.end(), [this, tr_state = std::move(tr_state), timeout] (const mutation& m) mutable {
-            return mutate_locally(m, tr_state, db::commitlog::force_sync::no, timeout);
+storage_proxy::mutate_locally(std::vector<mutation> mutations, tracing::trace_state_ptr tr_state, clock_type::time_point timeout, smp_service_group smp_grp) {
+    return do_with(std::move(mutations), [this, timeout, tr_state = std::move(tr_state), smp_grp] (std::vector<mutation>& pmut) mutable {
+        return parallel_for_each(pmut.begin(), pmut.end(), [this, tr_state = std::move(tr_state), timeout, smp_grp] (const mutation& m) mutable {
+            return mutate_locally(m, tr_state, db::commitlog::force_sync::no, timeout, smp_grp);
         });
     });
 }
 
+future<> 
+storage_proxy::mutate_locally(std::vector<mutation> mutation, tracing::trace_state_ptr tr_state, clock_type::time_point timeout) {
+        return mutate_locally(std::move(mutation), tr_state, timeout, _write_smp_service_group);
+}
 future<>
 storage_proxy::mutate_hint(const schema_ptr& s, const frozen_mutation& m, tracing::trace_state_ptr tr_state, clock_type::time_point timeout) {
     auto shard = _db.local().shard_of(m);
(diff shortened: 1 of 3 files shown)


service/storage_proxy.cc
```diff
@@ -4861,26 +4863,26 @@ void storage_proxy::init_messaging_service() {
...
                 /* forward_fn */ [] (shared_ptr<storage_proxy>& p, netw::messaging_service::msg_addr addr, clock_type::time_point timeout, const frozen_mutation& m,
                         gms::inet_address reply_to, unsigned shard, response_id_type response_id,
                         std::optional<tracing::trace_info> trace_info) {
                     return p->_messaging.send_mutation(addr, timeout, m, {}, reply_to, shard, response_id, std::move(trace_info));
                 });
     };
-    ms.register_mutation(receive_mutation_handler);
-    ms.register_hint_mutation(receive_mutation_handler);
+    ms.register_mutation(std::bind_front<>(receive_mutation_handler, write_hint::no));
+    ms.register_hint_mutation(std::bind_front<>(receive_mutation_handler, write_hint::yes));
```
avikivity, line 4886: Maybe just pass a pointer to the service group, instead of yes/no and then make a choice later? Saves a choice.

More complicated is to make smp_service_group scheduling_group aware.

service/storage_proxy.cc
```diff
@@ -1837,20 +1838,21 @@ storage_proxy::mutate_locally(const mutation& m, tracing::trace_state_ptr tr_sta
...
-    return do_with(std::move(mutations), [this, timeout, tr_state = std::move(tr_state)] (std::vector<mutation>& pmut) mutable {
-        return parallel_for_each(pmut.begin(), pmut.end(), [this, tr_state = std::move(tr_state), timeout] (const mutation& m) mutable {
-            return mutate_locally(m, tr_state, db::commitlog::force_sync::no, timeout);
+storage_proxy::mutate_locally(std::vector<mutation> mutations, tracing::trace_state_ptr tr_state, clock_type::time_point timeout, write_hint mutation_is_hint) {
+    return do_with(std::move(mutations), [this, timeout, tr_state = std::move(tr_state), mutation_is_hint] (std::vector<mutation>& pmut) mutable {
+        return parallel_for_each(pmut.begin(), pmut.end(), [this, tr_state = std::move(tr_state), timeout, mutation_is_hint] (const mutation& m) mutable {
+            return mutate_locally(m, tr_state, db::commitlog::force_sync::no, timeout, mutation_is_hint);
         });
     });
 }
```
piodul, line 1858: (Writing here because I can't attach this comment to the right place)

You should update the smp_service_group in the mutate_hint function, too.

service/storage_proxy.cc
eliransin: I need to check it but the mutate_locally functions are public while the smp service groups are private in storage proxy,
the question is if it justifies the split.
As an alternative to what I did here I can:
1. make the groups public ( as mutate locally is used outside of storage_proxy)
2. make a private mutate_locally functions that are being called with default smp groups by a public counterparts
3. Add an array of smp_groups that is indexed by an enum, this enum will be passed to the mutate_locally function

I prefer option 3 as it will externalize the concept without externalizing storage_proxy's internals.

service/storage_proxy.cc
eliransin: Thanks @piodul - I missed this one :slightly_smiling_face:

service/storage_proxy.cc
avikivity: If we have a scheduling group aware smp_service_group it will also work for workload prioritization.

avikivity: I think option 2 is reasonable, it's nicer to have a different function rather than booleans that slightly affect behavior. Option 3 would be good if we didn't have the extra complication of workload prioritization.

eliransin: I don't have any specific problem with option 2, but I don't understand what does workload prioritization have got to do 
with it. It is not a good solution to create a semaphore per scheduling group because it will either increase the amount of 
cross shard operations that are active or will split the existing cross shard operations quota in a way that will not fully 
utilize them. I have just published a proposal [here](https://docs.google.com/document/d/1KJd-VRD_ZinlVbLDA6_qYfFjshcLAN9EaksoX0vB8xo/edit?usp=sharing) for a priority aware semaphore which should solve this problem for workload prioritization and for other cases like hints vs normal writes.

avikivity: Workload prioritization suffers from the same issue. If service level X stuffs the smp_service_group with requests, which are then handled slowly because X has few shares, then service level Y will block on the queue even though it has plenty of shares. 

If we have a priority aware sempahore then we can use it here, too, no?

eliransin: Ahh sure, I probably misunderstood you.
Will start working on an RFC for priority aware semaphore ASAP.

service/storage_proxy.cc
```diff
@@ -4862,26 +4867,26 @@ void storage_proxy::init_messaging_service() {
...
+    auto receive_mutation_handler = [] (smp_service_group smp_grp, const rpc::client_info& cinfo, rpc::opt_time_point t, frozen_mutation in, std::vector<gms::inet_address> forward,
             gms::inet_address reply_to, unsigned shard, storage_proxy::response_id_type response_id, rpc::optional<std::optional<tracing::trace_info>> trace_info) {
         tracing::trace_state_ptr trace_state_ptr;
         auto src_addr = netw::messaging_service::get_source(cinfo);
 
         utils::UUID schema_version = in.schema_version();
         return handle_write(src_addr, t, schema_version, std::move(in), std::move(forward), reply_to, shard, response_id,
                 trace_info ? *trace_info : std::nullopt,
-                /* apply_fn */ [] (shared_ptr<storage_proxy>& p, tracing::trace_state_ptr tr_state, schema_ptr s, const frozen_mutation& m,
+                /* apply_fn */ [&smp_grp] (shared_ptr<storage_proxy>& p, tracing::trace_state_ptr tr_state, schema_ptr s, const frozen_mutation& m,
```
nyh, line 4878: I think you need to copy smp_grp, not to use a reference to the local variable here.

service/storage_proxy.hh
```diff
@@ -468,12 +468,29 @@ public:
...
+    // Applies mutation on this node.
+    // Resolves with timed_out_error when timeout is reached.
+    future<> mutate_locally(const mutation& m, tracing::trace_state_ptr tr_state, db::commitlog::force_sync sync, clock_type::time_point timeout = clock_type::time_point::max()) {
+        return mutate_locally(m, tr_state, sync, timeout, _write_smp_service_group);
+    }
+    // Applies mutation on this node.
+    // Resolves with timed_out_error when timeout is reached.
+    future<> mutate_locally(const schema_ptr& s, const frozen_mutation& m, tracing::trace_state_ptr tr_state, db::commitlog::force_sync sync, clock_type::time_point timeout = clock_type::time_point::max()) {
+        return mutate_locally(s, m, tr_state, sync, timeout, _write_smp_service_group);
+    }
```
nyh, line 493: Function overloading is one of the least favorite features of C++ for me, so I'm not very happy with having 4 functions called mutate_locally. But I won't insist.

nyh (approved):
Looks good to me, though I left a few comments. I see you have long discussions with Avi about this, however, so I let him have the honor of committing it when he's ok with this.

nyh (changes requested):
Oh, I forgot that I saw what I suspect to be a use-after-free bug (where you pass a reference to a copy), please look into that.

service/storage_proxy.cc
eliransin: I will change it here and in another place where smp_group is passed down to a lambda.

service/storage_proxy.hh
eliransin: I started with a different approach but @avi likes this better https://github.com/scylladb/scylla/pull/7178#discussion_r484195057

eliransin:
pushed a new branch:
Changes are to copy the smp_service_group around instead of passing a reference.

eliransin:
> Oh, I forgot that I saw what I suspect to be a use-after-free bug (where you pass a reference to a copy), please look into that.

It was indeed the case so I just changed the code to copy it around since it is small enough so it doesn't matter.

nyh (approved):

service/storage_proxy.cc
avikivity: I'll merge this, and we can replace with the super duper semaphore when we have it.

PR #7180, opened 2020-09-06
github: modernize issue templates

avikivity:
Have separate templates for bug reports, enhancements, and internals.
Automatically assign labels based on the type of issue.


slivne:
@tzach sent a parallel pull request not long ago - @tzach ?

.github/ISSUE_TEMPLATE/enhancement.md
```diff
@@ -0,0 +1,8 @@
+---
+name: Enhancement Request
+about: I'd like a feature to be implemented
+labels: enhancement
```
nyh, line 4: I think the "enhancement" label is redundant. With your templates, every issue will have either "bug" or "enhancement", never both. So it's enough to just have "bug", and no label for enhancements.
It's also easier to change your mind later when there is just one label ("bug"). If here are two labels (bug and enhancement) and you decide that an enhancement is really a bug, you need to do two changes (drop the enhancement label, add the bug label) instead of one.

nyh (changes requested):

.github/ISSUE_TEMPLATE/enhancement.md
avikivity: There are also issues that have neither. 

It's also more natural to select all enhancements, than all non-bugs.

nyh: With your templates, there will be no issue that has neither. All your templates have either one or the other.

tzach:
@slivne here it is https://github.com/scylladb/scylla/pull/6326
I closed it after a few months of inactivity.

Its similar to what @avikivity created, only with even more issues types (templates)

1. bug-issue
2. performance-issue
3. feature-request
4. installation-issue
5. build-issue
6. security
The last refer you the the new SECURITY.md file.

It also include cool unicode per type, but I'm not sure everyone like it ;)

slivne:
So please @Avi Kivity <avi@scylladb.com> review what tzach shared as well -
lets align on whats better.

nyh:
If you ask me, the better thing is not Avi's suggestion and not Tzach's - it is what we already have, after we **remove** the annoying warning and check box from the beginning.

In other words, I think we should have just one type of issue where users are just "encouraged" to fill details about their version. In my opinion and experience, any attempt of having users try to triage the bug for you - divide it into 7 (or 3) categories, and so on, is doomed for failure - all that will happen is that you'll make these labels useless because nobody can trust them anyway - a developer will anyway need to check these labels and often correct them - which is more work (you need to remove incorrect labels, and add new ones).

What's wrong with us (developers) putting these labels ourselves when we look at the bug - something we need to do anyway?

avikivity:
> If you ask me, the better thing is not Avi's suggestion and not Tzach's - it is what we already have, after we **remove** the annoying warning and check box from the beginning.
> 
> In other words, I think we should have just one type of issue where users are just "encouraged" to fill details about their version. In my opinion and experience, any attempt of having users try to triage the bug for you - divide it into 7 (or 3) categories, and so on, is doomed for failure - all that will happen is that you'll make these labels useless because nobody can trust them anyway - a developer will anyway need to check these labels and often correct them - which is more work (you need to remove incorrect labels, and add new ones).
> 
> What's wrong with us (developers) putting these labels ourselves when we look at the bug - something we need to do anyway?

It's not done, that's what's wrong. What's wrong with auto-labeling a bug as a bug?

.github/ISSUE_TEMPLATE/enhancement.md
avikivity: Yes. That's the goal. So if you see an issue with no label, you have don't have to wonder whether it's uncategorized, or an enhancement. You know it's an uncategorized.

nyh: No, after your patch you will never see an issue which is "uncategorized", it will always have one of these two. Even if the bug reporter doesn't really know what's the difference, so he just picked one of the two at random. So now if I see the label "enhancement", I have to wonder even harder (and every time, not just once) whether this is really an "enhancement", because this label might have been set by a random user who knows nothing about our development process.
Anyway, I'm not a fan of the new templates, but if you are, and find another fan to commit it, I won't fight it either :-)

nyh:
> > What's wrong with us (developers) putting these labels ourselves when we look at the bug - something we need to do anyway?
> 
> It's not done, that's what's wrong. What's wrong with auto-labeling a bug as a bug?

It's ok to have the user label things as "bug" vs "enhancement", but this means that this label can no longer be trusted to be correct, and in particular should not be a deciding factor of whether to backport a patch or not. We cannot base any of our development or backport processes on a tag which might have been set by a user who knows nothing of this process. So if it's nice for you to have this informational, untrustworthy, labels, then ok. I personally prefer to have only trustworthy tags, or none at all.

denesb:
We can add an `awaiting_triage` label, so we can filter out untriaged issues.

nyh:
@denesb I guess my opinion isn't popular here, but my opinion (and past experience) is that the more automatic labels you add, the less useful they become. Let's say that you add an "awaiting_triage" label - who is officially supposed to do this triage, and remove this label? If I (who don't see myself as the "triage boss" in Scylla) look at the issue myself, should I remove this tag when I'm done looking? Or shouldn't I? At the end, you'll have a random subset of issues with this tag, and another random subset without this tag, and nobody will know what it means.
My belief is that tags are only useful when written by people heavily invested in that tag: e.g., I care which issues have the "alternator tag" (and upset by any mistagging), Avi care about backport tags, and Shlomi cares about milestone tags. Although we can allow random contributors to also add these tags (and we do), the result will not be helpful. For example, imagine that we ask users to pick a "subsystem" for their issues (like many projects do). I'm guessing that we'll see many issues reported for Alternator just because it's first in the alphabetical list ;-)

.github/ISSUE_TEMPLATE/enhancement.md
avikivity: We will still see uncategorized issues, because many still exist.

avikivity:
Right now I triage all issues wrt backports (except those labeled "enhancement" which I believe @penberg filters out). It's pretty boring.

penberg:
@avikivity Yes, the backport queue bot indeed ignores issues labeled as "enhancement".

slivne:
@avikivity / @tzach please agree on whats needed for users - we are running months without an agreed solution and its hurting us - its better to make a small step then keep the current status

Issue #7153, opened 2020-08-31
scylla_setup failing with dependency errors


kolasani83:
Iam trying to install scylla 4.1.4 on ubuntu 16 server which has no access to internet.
So i down loaded scylla packages on system with internet access and installing them with dpkg -i *

openjdk-8-jre-headless_8u265-b01-0ubuntu2~16.04_amd64.deb
apt-transport-https_1.2.32ubuntu0.1_amd64.deb
ca-certificates-java_20160321ubuntu1_all.deb
libxtst6.deb
libxi6.deb
libfontconfig1_2.11.94-0ubuntu1.1_amd64.deb
libpcsclite1_1.8.14-1ubuntu1.16.04.1_amd64.deb
libxrender1.deb
ntp.deb
ntpdate.deb

scylla-manager-client_2.1.0-0.20200707.c650b46d-0ubuntu1~xenial_amd64.deb
scylla-manager-agent_2.1.0-0.20200707.c650b46d-0ubuntu1~xenial_amd64.deb
scylla-jmx_4.1.4-20200809.d2dd8d027ce-1_all.deb
scylla-python3_3.7.7-0.20200809.76f4bc4c6f0-1_amd64.deb
scylla-kernel-conf_4.1.4-0.20200809.76f4bc4c6f0-1_amd64.deb
scylla-conf_4.1.4-0.20200809.76f4bc4c6f0-1_amd64.deb
scylla-server_4.1.4-0.20200809.76f4bc4c6f0-1_amd64.deb
scylla_4.1.4-0.20200809.76f4bc4c6f0-1_amd64.deb
scylla-tools-core_4.1.4-20200809.4492e12926d-1_all.deb
scylla-tools_4.1.4-20200809.4492e12926d-1_all.deb

when i ran scylla_setup iam getting below error.

root@ip-10-23-198-218:/tmp/openjdk_package# scylla_setup --no-raid-setup --no-enable-service  --no-sysconfig-setup
This EC2 instance is optimized for Scylla.
WARN  2020-08-31 17:03:37,763 [shard 0] iotune - Available space on filesystem at /var/tmp/mnt: 124 MB: is less than recommended: 10 GB
INFO  2020-08-31 17:03:37,763 [shard 0] iotune - /var/tmp/mnt passed sanity checks
This is a supported kernel version.
You current Scylla release is 4.1.4 while the latest patch release is 4.1.5, update for the latest bug fixes and improvements
Reading package lists... Done
Building dependency tree
[...]

PR #7185, opened 2020-09-07
scylla_ntp_setup: don't install ntpd package when it's already exists

syuu1228:
Don't install ntpd package when it's already exists.
Related with #7153

### DIFF
--- a/dist/common/scripts/scylla_ntp_setup
+++ b/dist/common/scripts/scylla_ntp_setup
@@ -41,7 +41,8 @@ if __name__ == '__main__':
     args = parser.parse_args()
 
     if is_debian_variant():
-        run('apt-get install -y ntp ntpdate')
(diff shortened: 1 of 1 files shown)


dist/common/scripts/scylla_ntp_setup
```diff
@@ -41,7 +41,8 @@ if __name__ == '__main__':
     args = parser.parse_args()
 
     if is_debian_variant():
-        run('apt-get install -y ntp ntpdate')
+        if not os.path.exists('/usr/sbin/ntpd') or not os.path.exists('/usr/sbni/ntpdate'):
```
avikivity, line 44: s/sbin/sbin/

amoskong: s/sbni/sbin/

syuu1228: Fixed this.

penberg:
Queued as 4deb245198f303938d8e69ba1189ae31ce93d9ab

PR #7189, opened 2020-09-07
db: Propagate enable_cache to system keyspaces

### LINKED ISSUE #2909: --enable-cache has no effect for system keyspace
It's always enabled.

avelanarius:
Make `enable_cache` configuration option also affect caching of system keyspaces. Fixes #2909.

It makes `--enable-cache` command-line option change caching of system keyspaces (`system`, `system_schema`). 

Manually verified that after setting `--enable-cache false`, no calls to caching were performed (added debug prints) - previously caching was still performed in system keyspaces.


haaawk (approved):
LGTM

Issue #7166, opened 2020-09-04
init - Startup failed: std::runtime_error (Could not find tokens for 10.0.0.155 to replace) during large-partition-4d test


aleksbykov:
*Installation details*
Scylla version (or git commit hash): Scylla version 4.2.rc3-0.20200825.989d8fe63 with build-id e7d653bede6105c8264984a606cd5334967ca553
Cluster size: 4 nodes
OS (RHEL/CentOS/Ubuntu/AWS AMI): ami-0637c7ab2b923d84e(en-central-1)

*Hardware details (for performance issues)*          Delete if unneeded
Platform (physical/VM/cloud instance type/docker): AWS
instance type: i3en.3xlarge 

During job [longevity-large-partition-4d](https://jenkins.scylladb.com/view/scylla-4.2/job/scylla-4.2/job/longevity/job/longevity-large-partition-4days-test/5/) Scylla on node4 during nemesis: RestartThenRepairNode failed to start with error:
```
2020-09-01T23:02:55+00:00  longevity-large-partitions-4d-4-2-db-node-7a748b66-6 !ERR     | scylla: [shard 0] init - Startup failed: std::runtime_error (Could not find tokens for 10.0.0.155 to replace)
```
It is similar to issue https://github.com/scylladb/scylla/issues/6648

Next nemesises(actions) were run before failed RestartThenRepairNode nemesis:
```
2020-09-01 14:44:19.768: (DisruptionEvent Severity.NORMAL): type=start name=NoCorruptRepair node=Node longevity-large-partitions-4d-4-2-db-node-7a748b66-3 [3.120.147.241 | 10.0.2.223] (seed: False) duration=None
2020-09-01 15:11:26.153: (DisruptionEvent Severity.NORMAL): type=end name=NoCorruptRepair node=Node longevity-large-partitions-4d-4-2-db-node-7a748b66-3 [3.120.147.241 | 10.0.2.223] (seed: False) duration=1626
2020-09-01 15:41:47.367: (DisruptionEvent Severity.NORMAL): type=start name=ManagementRepair node=Node longevity-large-partitions-4d-4-2-db-node-7a748b66-2 [3.120.179.232 | 10.0.3.165] (seed: False) duration=None
[...]

PR #7202, opened 2020-09-09
storage_service: Fix a TOKENS update race for replace operation

asias:
In commit 7d86a3b208659e92489216e71df13fb33b1a4c3a (storage_service:
Make replacing node take writes), application state of TOKENS of the
replacing node is added into gossip and propagated to the cluster after
the initial start of gossip service. This can cause a race below

1. The replacing node replaces the old dead node with the same ip address
2. The replacing node starts gossip without application state of the TOKENS
3. Other nodes in the cluster replace the application states of old dead node's
   version with the new replacing node's version
4. replacing node dies
5. replace operation is performed again, the TOKENS application state is
   not preset and replace operation fails.

To fix, we can always add TOKENS application state when the
gossip service starts.

Fixes: #7166
Backports: 4.1 and 4.2

### DIFF
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -354,6 +354,9 @@ void storage_service::prepare_to_join(
         app_states.emplace(gms::application_state::CDC_STREAMS_TIMESTAMP, versioned_value::cdc_streams_timestamp(_cdc_streams_ts));
         app_states.emplace(gms::application_state::STATUS, versioned_value::normal(my_tokens));
     }
+    if (replacing_a_node_with_same_ip || replacing_a_node_with_diff_ip) {
+        app_states.emplace(gms::application_state::TOKENS, versioned_value::tokens(_bootstrap_tokens));
+    }
     slogger.info("Starting up server gossip");
 
     auto generation_number = db::system_keyspace::increment_and_get_generation().get0();


tgrabiec:
> 
> 
> 1. The replacing node replaces the old dead node with the same ip address
> 2. The replacing node starts gossip without application state of the TOKENS
> 3. Other nodes in the cluster replace the application states of old dead node's
>    version with the new replacing node's version
> 4. replacing node dies
> 5. replace operation is performed again, the TOKENS application state is
>    not preset and replace operation fails.

I don't understand some things about this scenario:

1. "replace operation is performed again" -> do you restart the same node which died, or start a new node?

2. You say that replace fails because TOKENS application state is not present, but why is it not present? How is the "performed again" scenario different from performed for the first time? What expects TOKENS to be present?

3. Shouldn't we always cleanup after the first replacing node died before proceeding with step 5. Then the "performed again" case is the same as "performed for the first time".

asias:
> > 1. The replacing node replaces the old dead node with the same ip address
> > 2. The replacing node starts gossip without application state of the TOKENS
> > 3. Other nodes in the cluster replace the application states of old dead node's
> >    version with the new replacing node's version
> > 4. replacing node dies
> > 5. replace operation is performed again, the TOKENS application state is
> >    not preset and replace operation fails.
> 
> I don't understand some things about this scenario:
> 
>     1. "replace operation is performed again" -> do you restart the same node which died, or start a new node?

Either start the same replacing node again or start a new node to replace. It is the same. What we actually saw in the report is the starting the same node. 
> 
>     2. You say that replace fails because TOKENS application state is not present, but why is it not present? How is the "performed again" scenario different from performed for the first time? What expects TOKENS to be present?

It is replaced by the application states propagated to the cluster by the second boot. The replacing node advertise bigger generation so other nodes will use it to update the existing one with older generation. 


The fist replace operation has the full gossip application state of the dead node.

The "performed again" has the gossip application states of what the first replace left. 

Replacing node needs TOKENS to know what tokens it should use.


> 
>     3. Shouldn't we always cleanup after the first replacing node died before proceeding with step 5. Then the "performed again" case is the same as "performed for the first time".

What to cleanup? They are using the same ip address. We can not tell if the application states were created by the replacing node or the dead node to be replaced.

tgrabiec (approved):

asias:
In master already.

Issue #7197, opened 2020-09-09
Gossiper: make sure to inform other hosts that this host is alive


vladzcloudius:
HEAD: f0b45d652016e84ff6d998d26d9d0839b48b48a7

*Description*
Today, if gossiper fails to send the echo request to other nodes due to some overload or stalls on other shards it will keep on trying and failing (timeout) until the original reason for a failure goes away - and this may take some time. 

All that time other nodes in the cluster will conclude that those node(s) are DOWN, and the cluster will become unavailable, while in fact they are pretty much alive and there are just a few shards that are overloaded.

We may want to develop strategies inside the gossiper that would make it more resilient in such events, e.g.:
 * Try to send a slim echo message that doesn't require getting info from other shards.
 * Migrate the main gossiper task to another shard and try to operate from there.

vladzcloudius:
@avikivity @asias FYI

asias:
I think we can optimize to get rid of the submit_to(0) call in the echo message handler. I will send a patch.

Migrate gossiper tasks to other shards is quite complicated change because gossip is per node and we store information on shard 0 only. 

Gossip echo message timeout in 3 seconds which is pretty long. Do you see a lot of such timeout due to loaded system in the field?

dorlaor:
Vlad observed such issues on the field, it's possible to find them on github

tzach:
@asias @vladzcloudius can we close this issue?

vladzcloudius:
@tzach Is this issue "fixed" by patches above? @asias Could you, please, clarify?

denesb:
Reassigning to @emdroid (cluster team). Emil, please close if stale or doesn't align with your plans for gossip.

PR #7210, opened 2020-09-10
gossiper: Run echo message handler on all shards

asias:
Echo message does not need to access gossip internal states. 
Yes, set_last_processed_message_at() is called but it is not really used. 
So we can run it on all shards and avoid forwarding to shard zero.

This makes gossip marking node up more robust when shard zero is loaded.

Refs: #7197

### DIFF
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -418,7 +418,6 @@ future<> gossiper::handle_ack2_msg(gossip_digest_ack2 msg) {
 }
 
 future<> gossiper::handle_echo_msg() {
-    set_last_processed_message_at();
     return make_ready_future<>();
 }
 
@@ -491,9 +490,7 @@ future<> gossiper::init_messaging_service_handler(bind_messaging_port do_bind) {
         return messaging_service::no_wait();
     });
     _messaging.register_gossip_echo([] {
-        return smp::submit_to(0, [] {
-            return gms::get_local_gossiper().handle_echo_msg();
-        });
+        return gms::get_local_gossiper().handle_echo_msg();
     });
     _messaging.register_gossip_shutdown([] (inet_address from) {
         // In a new fiber.
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -159,14 +159,6 @@ gossiper::gossiper(abort_source& as, feature_service& features, const locator::t
     });
 }
 
-void gossiper::set_last_processed_message_at() {
-    set_last_processed_message_at(now());
-}
-
-void gossiper::set_last_processed_message_at(clk::time_point tp) {
-    _last_processed_message_at = tp;
-}
-
 /*
  * First construct a map whose key is the endpoint in the GossipDigest and the value is the
  * GossipDigest itself. Then build a list of version differences i.e difference between the
@@ -211,7 +203,6 @@ void gossiper::do_sort(utils::chunked_vector<gossip_digest>& g_digest_list) {
 future<> gossiper::handle_syn_msg(msg_addr from, gossip_digest_syn syn_msg) {
     logger.trace("handle_syn_msg():from={},cluster_name:peer={},local={},partitioner_name:peer={},local={}",
         from, syn_msg.cluster_id(), get_cluster_name(), syn_msg.partioner(), get_partitioner_name());
-    this->set_last_processed_message_at();
     if (!this->is_enabled()) {
         return make_ready_future<>();
     }
@@ -306,7 +297,6 @@ private:
 future<> gossiper::handle_ack_msg(msg_addr id, gossip_digest_ack ack_msg) {
     logger.trace("handle_ack_msg():from={},msg={}", id, ack_msg);
 
-    this->set_last_processed_message_at();
     if (!this->is_enabled() && !this->is_in_shadow_round()) {
         return make_ready_future<>();
     }
@@ -404,7 +394,6 @@ future<> gossiper::do_send_ack2_msg(msg_addr from, utils::chunked_vector<gossip_
 // - on_alive callbacks
 future<> gossiper::handle_ack2_msg(gossip_digest_ack2 msg) {
     logger.trace("handle_ack2_msg():msg={}", msg);
-    set_last_processed_message_at();
     if (!is_enabled()) {
         return make_ready_future<>();
     }
@@ -422,7 +411,6 @@ future<> gossiper::handle_echo_msg() {
 }
 
 future<> gossiper::handle_shutdown_msg(inet_address from) {
-    set_last_processed_message_at();
     if (!is_enabled()) {
         logger.debug("Ignoring shutdown message from {} because gossip is disabled", from);
         return make_ready_future<>();
@@ -1415,7 +1403,6 @@ void gossiper::mark_alive(inet_address addr, endpoint_state& local_state) {
     // Do it in the background.
     (void)_messaging.send_gossip_echo(id).then([this, addr] {
         logger.trace("Got EchoMessage Reply");
-        set_last_processed_message_at();
         return seastar::async([this, addr] {
             // After sending echo message, the Node might not be in the
             // endpoint_state_map anymore, use the reference of local_state
--- a/gms/gossiper.hh
+++ b/gms/gossiper.hh
@@ -222,8 +222,6 @@ private:
 
     bool _in_shadow_round = false;
 
-    clk::time_point _last_processed_message_at = now();
-
     std::unordered_map<inet_address, clk::time_point> _shadow_unreachable_endpoints;
     utils::chunked_vector<inet_address> _shadow_live_endpoints;
 
@@ -239,9 +237,6 @@ private:
     future<> replicate(inet_address, application_state key, const versioned_value& value);
 public:
     explicit gossiper(abort_source& as, feature_service& features, const locator::token_metadata& tokens, netw::messaging_service& ms, db::config& cfg);
-
-    void set_last_processed_message_at();
-    void set_last_processed_message_at(clk::time_point tp);
 
     void check_seen_seeds();
 


psarna (approved):

avikivity:
Does it really work? our rpc tries to connect equal number shards, so a message from shard 0 will be received by shard 0 and processed by it.

tgrabiec:
I have a question about semantics. What is "node UP" supposed to mean?

In case of unresponsive shards, should we consider a node with a single unresponsive shard as unavailable? We use the UP/DOWN status for request routing. Requests which go to the slow shard would be stalled, which would negate the purpose of the UP/DOWN thing. In this interpretation, this patch is a step in the wrong direction.

asias:
> Does it really work? our rpc tries to connect equal number shards, so a message from shard 0 will be received by shard 0 and processed by it.

It is not guaranteed that a connection on shard x from node A will always be on shard x on node B, isn't it? If shard 0 connection goes to non-zero shard, we save an unnecessary sumbit_to. If it goes to shard zero, we have nothing to lose.

asias:
> I have a question about semantics. What is "node UP" supposed to mean?

A node UP is local node's view about another node. If other node does not update heart beat for a while, a node is marked as down. When new heart beat comes, local node mark the other node up after sending the echo message to confirm.

> In case of unresponsive shards, should we consider a node with a single unresponsive shard as unavailable? 

UP and DOWN is per node not per shard, which I think it is good enough because a single shard's unresponseness is supposed to be a bug we should fix.


> We use the UP/DOWN status for request routing. 

UP and DOWN is not only used for request routing, it is also used to decide if a node is down and we should kill all the streaming or repair sessions with that node. 

> Requests which go to the slow shard would be stalled, which would negate the purpose of the UP/DOWN thing. In this interpretation, this patch is a step in the wrong direction.

UP/DOWN is not used to decide if a node is fast or slow. Even if currently, a down can be caused by stall of shard zero on one node, what about other shards if we use it to present slowness.

dorlaor:
The shard liveness vs node liveness is indeed an interesting relation.
We may need to solve both of them. This code attempts to solve the issue
that a single shard that runs gossip becomes temporarily overloaded,
resulting
in a failure to be part of the cluster.

The question is whether the best behaviour is to expel the node from the
cluster
and thus limit the pain of the whole cluster or keep it as a member, keep
the
cluster membership intact while we risk that the member may slow the rest
of
the cluster.

Assuming the overload is due to hot partitions, the remaining replicas will
even
suffer more, thus it's better to keep the membership. In case it's due to a
hardware
issue, we better expel the node.

Currently we saw in the field that expelling results in pain.

tgrabiec:
> > I have a question about semantics. What is "node UP" supposed to mean?
> 
> A node UP is local node's view about another node. If other node does not update heart beat for a while, a node is marked as down. When new heart beat comes, local node mark the other node up after sending the echo message to confirm.
> 
> > In case of unresponsive shards, should we consider a node with a single unresponsive shard as unavailable?
> 
> UP and DOWN is per node not per shard, which I think it is good enough because a single shard's unresponseness is supposed to be a bug we should fix.

I don't think "a bug we should fix" should be a deciding factor. The cause of unavailability doesn't matter. If a node crashes due to a bug we should fix, it doesn't change the fact that the node is down.

Even if the flag is per node, the meaning is ambiguous. Does UP mean:
  1) all shards are responsive
  2) there exists at least one shard which is responsive
  3) the scylla process is running
  4) the machine on which scylla process should be running is powered on

Maybe different parts of scylla need different meaning.

For example, @kostja had some ideas about reducing raft group keepalives by relying on failure detector. That calls for meaning 1.

> 
> > We use the UP/DOWN status for request routing.
> 
> UP and DOWN is not only used for request routing, it is also used to decide if a node is down and we should kill all the streaming or repair sessions with that node.


Is your argument here that it's better to not kill streaming when one of the shards is unavailable because (and now I'm guessing) we would lose progress, which is expensive, and we hope the shard will come back up and we will be able to continue so it's better to wait than to kill streaming. 

The same could be said about a node which is detected DOWN, it may come back up, e.g. when the network partition heals. Why do we kill streaming then? We could wait as well.



> 
> > Requests which go to the slow shard would be stalled, which would negate the purpose of the UP/DOWN thing. In this interpretation, this patch is a step in the wrong direction.
> 
> UP/DOWN is not used to decide if a node is fast or slow. Even if currently, a down can be caused by stall of shard zero on one node, what about other shards if we use it to present slowness.

That would be an argument for having UP/DOWN with shard granularity instead of node granularity.

And if we keep it with node granularity, then it's also an argument for keeping the meaning 2, where UP is weak. But again, this makes the meaning of UP weaker, so anything which relies on that will function less effectively. For example, we could always mark every node as UP and your concern will be satisfied as well, but then UP-based routing knows nothing, as if the feature didn't exist.

gleb-cloudius:
> > In case of unresponsive shards, should we consider a node with a single unresponsive shard as unavailable?
> 
> UP and DOWN is per node not per shard, which I think it is good enough because a single shard's unresponseness is supposed to be a bug we should fix.
> 
A single shard's unresponsiveness may indeed be a bug, but it does not mean that a node with a shard that is stuck should be reported as to be alive. It is not for request serving purposes.

avikivity:
> > Does it really work? our rpc tries to connect equal number shards, so a message from shard 0 will be received by shard 0 and processed by it.
> 
> It is not guaranteed that a connection on shard x from node A will always be on shard x on node B, isn't it? 

It's not, but in the vast majority of deployments, it is so.

> If shard 0 connection goes to non-zero shard, we save an unnecessary sumbit_to. If it goes to shard zero, we have nothing to lose.

We don't lose anything (apart from the call you removed, which may be unneeded), but we also don't gain much. It doesn't fix the problem.

avikivity:
> > > I have a question about semantics. What is "node UP" supposed to mean?
> > 
> > 
> > A node UP is local node's view about another node. If other node does not update heart beat for a while, a node is marked as down. When new heart beat comes, local node mark the other node up after sending the echo message to confirm.
> > > In case of unresponsive shards, should we consider a node with a single unresponsive shard as unavailable?
> > 
> > 
> > UP and DOWN is per node not per shard, which I think it is good enough because a single shard's unresponseness is supposed to be a bug we should fix.
> 
> I don't think "a bug we should fix" should be a deciding factor. The cause of unavailability doesn't matter. If a node crashes due to a bug we should fix, it doesn't change the fact that the node is down.
> 
> Even if the flag is per node, the meaning is ambiguous. Does UP mean:
> 
> 1. all shards are responsive

I think this. If a shard is unresponsive, we should isolate the node. It doesn't help that other shards still respond.

Of course we should before marking ourselves as unresponsive we should be extra sure that the problem is not correctable.


> 2. there exists at least one shard which is responsive
> 3. the scylla process is running
> 4. the machine on which scylla process should be running is powered on
> 
> Maybe different parts of scylla need different meaning.
> 
> For example, @kostja had some ideas about reducing raft group keepalives by relying on failure detector. That calls for meaning 1.

Yes.

asias:
> > > Does it really work? our rpc tries to connect equal number shards, so a message from shard 0 will be received by shard 0 and processed by it.
> > 
> > 
> > It is not guaranteed that a connection on shard x from node A will always be on shard x on node B, isn't it?
> 
> It's not, but in the vast majority of deployments, it is so.
> 
> > If shard 0 connection goes to non-zero shard, we save an unnecessary sumbit_to. If it goes to shard zero, we have nothing to lose.
> 
> We don't lose anything (apart from the call you removed, which may be unneeded), but we also don't gain much. It doesn't fix the problem.


You can see I used "Refs" instead of "Fixes".

asias:
> > > I have a question about semantics. What is "node UP" supposed to mean?
> > 
> > 
> > A node UP is local node's view about another node. If other node does not update heart beat for a while, a node is marked as down. When new heart beat comes, local node mark the other node up after sending the echo message to confirm.
> > > In case of unresponsive shards, should we consider a node with a single unresponsive shard as unavailable?
> > 
> > 
> > UP and DOWN is per node not per shard, which I think it is good enough because a single shard's unresponseness is supposed to be a bug we should fix.
> 
> I don't think "a bug we should fix" should be a deciding factor. The cause of unavailability doesn't matter. If a node crashes due to a bug we should fix, it doesn't change the fact that the node is down.

It can be a deciding factor, for instance, if the only reason that can cause a shard to be unresponsive is a bug. We can assume if one shard is responding the other shard will respond too. The current shard zero only gossip builds on this assumption.

>
> Even if the flag is per node, the meaning is ambiguous. Does UP mean:
> 
>     1. all shards are responsive
>

No such guarantee.

>     2. there exists at least one shard which is responsive

Currently, what it really means is shard zero is live in the granularity of seconds. 

> 
>     3. the scylla process is running

No. Scylla process can run but gossip is stopped. So 3 is not true. 
> 
>     4. the machine on which scylla process should be running is powered on

NO.

> 
> Maybe different parts of scylla need different meaning.
> 
> For example, @kostja had some ideas about reducing raft group keepalives by relying on failure detector. That calls for meaning 1.

If we need all shards liveness, we need to extend current gossip or use something else, e.g., the plain raft keepalive. 
> > > We use the UP/DOWN status for request routing.
> > 
> > 
> > UP and DOWN is not only used for request routing, it is also used to decide if a node is down and we should kill all the streaming or repair sessions with that node.
> 
> Is your argument here that it's better to not kill streaming when one of the shards is unavailable because (and now I'm guessing) we would lose progress, which is expensive, and we hope the shard will come back up and we will be able to continue so it's better to wait than to kill streaming.

No. If gossip detects a node is down, we kill the stream and we want to fail fast. In the past, we ignore gossip down for streaming and we removed that.

> 
> The same could be said about a node which is detected DOWN, it may come back up, e.g. when the network partition heals. Why do we kill streaming then? We could wait as well.

No, we trust gossip's detection.

> 
> > > Requests which go to the slow shard would be stalled, which would negate the purpose of the UP/DOWN thing. In this interpretation, this patch is a step in the wrong direction.
> > 
> > 
> > UP/DOWN is not used to decide if a node is fast or slow. Even if currently, a down can be caused by stall of shard zero on one node, what about other shards if we use it to present slowness.
> 
> That would be an argument for having UP/DOWN with shard granularity instead of node granularity.
> 
> And if we keep it with node granularity, then it's also an argument for keeping the meaning 2, where UP is weak. But again, this makes the meaning of UP weaker, so anything which relies on that will function less effectively. For example, we could always mark every node as UP and your concern will be satisfied as well, but then UP-based routing knows nothing, as if the feature didn't exist.

My questions is what cases would cause one of the shard is responding in seconds while other nodes are not responding other than we have bugs that stalls the shard. Yes, current gossip does not guarantee all shards are responding, but you are exaggerating "as if the feature does not exist".

asias:
> > > > I have a question about semantics. What is "node UP" supposed to mean?
> > > 
> > > 
> > > A node UP is local node's view about another node. If other node does not update heart beat for a while, a node is marked as down. When new heart beat comes, local node mark the other node up after sending the echo message to confirm.
> > > > In case of unresponsive shards, should we consider a node with a single unresponsive shard as unavailable?
> > > 
> > > 
> > > UP and DOWN is per node not per shard, which I think it is good enough because a single shard's unresponseness is supposed to be a bug we should fix.
> > 
> > 
> > I don't think "a bug we should fix" should be a deciding factor. The cause of unavailability doesn't matter. If a node crashes due to a bug we should fix, it doesn't change the fact that the node is down.
> > Even if the flag is per node, the meaning is ambiguous. Does UP mean:
> > 
> > 1. all shards are responsive
> 
> I think this. If a shard is unresponsive, we should isolate the node. It doesn't help that other shards still respond.
> 
> Of course we should before marking ourselves as unresponsive we should be extra sure that the problem is not correctable.

If we really want all shards to be responsive we can

1) internally, we set up watch dog timers for all non-zero shards. If one of them are not updating, shard zero that runs gossip for the node should not update the heartbeat

2) The echo handler message should be executed on all shards.

This makes gossip more strict about shards liveness and goes to the opposite direction of what Vlad wanted in #7197.

> 
> > 1. there exists at least one shard which is responsive
> > 2. the scylla process is running
> > 3. the machine on which scylla process should be running is powered on
> > 
> > Maybe different parts of scylla need different meaning.
> > For example, @kostja had some ideas about reducing raft group keepalives by relying on failure detector. That calls for meaning 1.
> 
> Yes.

dorlaor:
IMO the problem has more levels. We shouldn't take a while node down once a
single
shard is loaded but we also need to be careful not to affect the cluster in
a negative way
because of it.

Today the situation isn't symmetrical, shard !=0 can get stalled while
gossip won't be
affected in the way shard 0 is (IIUC after a while, gossip does get info
from all shards).
So the concern that we leave a bad node in the cluster is still there for
other shards.

Thus this patch makes shard 0 less special and we become a bit more
resilient.
It shouldn't be the last improvement we do in this space.

I think that it will be good to consider a health state per shard as a
node-internal state.
When we reach a situation where high priority tasks like the gossip cannot
run within
x seconds, we should enter a degraded state, pause repair/compaction and
even aggressively
drop the low priority CQL operations. Like an emergency load shedding.
We should do it for a limited amount of time/repetitions, so in case there
is a problem that
doesn't go away, we'll eventually remove ourselves from the cluster.

dorlaor:
I went to read the history of the 2 customer issues that popped related to gossip + shard-0.
On one hand there are many issues with large partitions and unpaged queries which will be
resolved by the users plus are now better handled by Scylla. 
On the other hand, there is a theory about loading shard-0 by tracing.

It seems that in terms of urgency this isn't super high and it will be good to try to emulate such
an issue in a test environment. I do think that shard fairness is a good principle and also the 
node-overload state that can help us out here.

Another thing is to properly log things - it will be nice that a node will notice it's overloaded (either
a single shard or multiple) and log it so the admin will notice.

avikivity:
Different consumers need different levels of confidence.

Streaming can throw a lot of work away (not so much today since it is resumable), so it needs to be very sure that the node is down. On the other hand, it doesn't care about the detection latency.

OLTP read/write can tolerate a node disappearing for a few seconds, since it (usually) has other replicas. On the other hand, it needs fast detection latency, because user response is degraded while we send to a node that is down.

Both need to be sure that all shards are operational.

So how about this:
 - gossip reponds with a confidence level about whether a node is up or down, not just a bool
 - different consumers check at different levels, for example streaming can insist on 15 seconds worth of checks
 - we check all shards. it can be with invoke_on_all(), I don't think it has significant latency since it runs on a private scheduling_group
 - we apply some hysteresis. For example, for streaming, wait until the last 15 echo requests fail before declaring a node down, but also wait for 15 successful echo responses before declaring it up again.

tgrabiec:
> > > > I have a question about semantics. What is "node UP" supposed to mean?
> > > 
> > > 
> > > A node UP is local node's view about another node. If other node does not update heart beat for a while, a node is marked as down. When new heart beat comes, local node mark the other node up after sending the echo message to confirm.
> > > > In case of unresponsive shards, should we consider a node with a single unresponsive shard as unavailable?
> > > 
> > > 
> > > UP and DOWN is per node not per shard, which I think it is good enough because a single shard's unresponseness is supposed to be a bug we should fix.
> > 
> > 
> > I don't think "a bug we should fix" should be a deciding factor. The cause of unavailability doesn't matter. If a node crashes due to a bug we should fix, it doesn't change the fact that the node is down.
> 
> It can be a deciding factor, for instance, if the only reason that can cause a shard to be unresponsive is a bug. We can assume if one shard is responding the other shard will respond too. The current shard zero only gossip builds on this assumption.

Categorizing an event as a bug doesn't permit us to assume that it can't happen because bugs do happen. For example, every crash is a bug, doesn't mean we can assume it won't happen and not fsync the commit log and lose data on a crash.

How is "shard zero only gossip" building on this assumption? I thought it doesn't need to care if other shards are responsive. It's like in Cassandra where the gossip thread could be running fine even though other threads are deadlocked.

> 
> > Even if the flag is per node, the meaning is ambiguous. Does UP mean:
> > ```
> > 1. all shards are responsive
> > ```
> 
> No such guarantee.
> 
> > ```
> > 2. there exists at least one shard which is responsive
> > ```
> 
> Currently, what it really means is shard zero is live in the granularity of seconds.
> 
> > ```
> > 3. the scylla process is running
> > ```
> 
> No. Scylla process can run but gossip is stopped. So 3 is not true.
> 
> > ```
> > 4. the machine on which scylla process should be running is powered on
> > ```
> 
> NO.
> 
> > Maybe different parts of scylla need different meaning.
> > For example, @kostja had some ideas about reducing raft group keepalives by relying on failure detector. That calls for meaning 1.
> 
> If we need all shards liveness, we need to extend current gossip or use something else, e.g., the plain raft keepalive.
> 
> > > > We use the UP/DOWN status for request routing.
> > > 
> > > 
> > > UP and DOWN is not only used for request routing, it is also used to decide if a node is down and we should kill all the streaming or repair sessions with that node.
> > 
> > 
> > Is your argument here that it's better to not kill streaming when one of the shards is unavailable because (and now I'm guessing) we would lose progress, which is expensive, and we hope the shard will come back up and we will be able to continue so it's better to wait than to kill streaming.
> 
> No. If gossip detects a node is down, we kill the stream and we want to fail fast. In the past, we ignore gossip down for 
streaming and we removed that.
> 
> > The same could be said about a node which is detected DOWN, it may come back up, e.g. when the network partition heals. Why do we kill streaming then? We could wait as well.
> 
> No, we trust gossip's detection.

Trust to detect what?

Gossip doesn't tell you whether it's better to abort streaming or wait. So maybe we shouldn't abort right away on node DOWN if that would lose a lot of work. 

Of course, it would be best to make the process resumable so there is no tradeoff and no need to wait.

The point is, the problem you describe we would have with advertising DOWN when a single shard is temporarily down, we already have, when a node becomes temporarily partitioned from the cluster. So maybe it should be solved in another way, outside gossip.

> 
> > > > Requests which go to the slow shard would be stalled, which would negate the purpose of the UP/DOWN thing. In this interpretation, this patch is a step in the wrong direction.
> > > 
> > > 
> > > UP/DOWN is not used to decide if a node is fast or slow. Even if currently, a down can be caused by stall of shard zero on one node, what about other shards if we use it to present slowness.
> > 
> > 
> > That would be an argument for having UP/DOWN with shard granularity instead of node granularity.
> > And if we keep it with node granularity, then it's also an argument for keeping the meaning 2, where UP is weak. But again, this makes the meaning of UP weaker, so anything which relies on that will function less effectively. For example, we could always mark every node as UP and your concern will be satisfied as well, but then UP-based routing knows nothing, as if the feature didn't exist.
> 
> My questions is what cases would cause one of the shard is responding in seconds while other nodes are not responding other than we have bugs that stalls the shard. 

You can categorize all such events as bugs, I don't think it changes anything.

tgrabiec:
> 
> > ```
> > 2. there exists at least one shard which is responsive
> > ```
> 
> Currently, what it really means is shard zero is live in the granularity of seconds.
> 
> > ```
> > 3. the scylla process is running
> > ```
> 
> No. Scylla process can run but gossip is stopped. So 3 is not true.

This meaning applies to when the node is UP (if UP then ...). If gossip is stopped then UP will be false, so "Scylla process can run but gossip is stopped" doesn't contradict that "if UP then scylla process is running". 

If "not UP", then we don't know.

asias:
> Categorizing an event as a bug doesn't permit us to assume that it can't happen because bugs do happen. For example, every crash is a bug, doesn't mean we can assume it won't happen and not fsync the commit log and lose data on a crash.

I never assumed the bug will not happen. The question is the frequency and the possibility such event. If scylla stalls 10s of seconds, it is a very serious bug.  If we think it is important to let other nodes know such even and mark this node down in case of such stall, we do not even need to modify any of gossip, we could just shutdown the node itself and isolate from the rest of the cluster. 

Shard zero is supposed to be the busiest shard because we have more work to do on shard zero. If scylla stalls, shard zero is supposed to have much better chance to than other shards. 

> How is "shard zero only gossip" building on this assumption? I thought it doesn't need to care if other shards are responsive. It's like in Cassandra where the gossip thread could be running fine even though other threads are deadlocked.

The UP or DOWN concept is per node. The assumption is that one shard is up or down, so do the other shards. We use shard zero's liveness to present the whole node.

asias:
> > > The same could be said about a node which is detected DOWN, it may come back up, e.g. when the network partition heals. Why do we kill streaming then? We could wait as well.
> > 
> > 
> > No, we trust gossip's detection.
> 
> Trust to detect what?

Detect a node down.
> 
> Gossip doesn't tell you whether it's better to abort streaming or wait. So maybe we shouldn't abort right away on node DOWN if that would lose a lot of work.

Sure gossip does not tell.  The fact is no one does and we do not really care because we can resume from where it is left. 
> 
> Of course, it would be best to make the process resumable so there is no tradeoff and no need to wait.
> 
> The point is, the problem you describe we would have with advertising DOWN when a single shard is temporarily down, we already have, when a node becomes temporarily partitioned from the cluster. So maybe it should be solved in another way, outside gossip.

Sorry,  it is not clear to me what you meant here.

asias:
> > No. Scylla process can run but gossip is stopped. So 3 is not true.
> 
> This meaning applies to when the node is UP (if UP then ...). If gossip is stopped then UP will be false, so "Scylla process can run but gossip is stopped" doesn't contradict that "if UP then scylla process is running".
> 
> If "not UP", then we don't know.

Exactly, if we don't know what "not UP" means, how is this useful at all.

asias:
> Different consumers need different levels of confidence.
> 
> Streaming can throw a lot of work away (not so much today since it is resumable), so it needs to be very sure that the node is down. On the other hand, it doesn't care about the detection latency.
> 
> OLTP read/write can tolerate a node disappearing for a few seconds, since it (usually) has other replicas. On the other hand, it needs fast detection latency, because user response is degraded while we send to a node that is down.
> 
> Both need to be sure that all shards are operational.
> 
> So how about this:
> 
>     * gossip reponds with a confidence level about whether a node is up or down, not just a bool
> 
>     * different consumers check at different levels, for example streaming can insist on 15 seconds worth of checks

We already have a sensitive parameter phi in gossip failure detector. Tuning it  affects how fast we detect a node down. If we need different levels, we can expose this internal to gossip consumer. 

> 
>     * we check all shards. it can be with invoke_on_all(), I don't think it has significant latency since it runs on a private scheduling_group

Updating a node's heart beat or not is the source of this node's liveness information. As I mentioned here https://github.com/scylladb/scylla/pull/7210#issuecomment-692411955, 

if we want to ensure a node is considered up, we could update the heartbeat only if all shards do not stall. 


> 
>     * we apply some hysteresis. For example, for streaming, wait until the last 15 echo requests fail before declaring a node down, but also wait for 15 successful echo responses before declaring it up again.

I want to clarify that we do not use echo request to detect a node down at all. We only use heartbeat update to detect a node up or down. The echo request is an extra step when a node mark a node up with the heartbeat update to verify that a node is really up.

asias:
> I went to read the history of the 2 customer issues that popped related to gossip + shard-0.
> On one hand there are many issues with large partitions and unpaged queries which will be
> resolved by the users plus are now better handled by Scylla.
> On the other hand, there is a theory about loading shard-0 by tracing.

All tracing runs on shard zero only too? Could someone confirm or deny?

> 
> It seems that in terms of urgency this isn't super high and it will be good to try to emulate such
> an issue in a test environment. I do think that shard fairness is a good principle and also the
> node-overload state that can help us out here.

The root of the problem is per node and per shard granularity. If a item is per node, we can only select one of the shard to do so. Unavoidably, we could cause shard unfairness if the amount of work is not ignorable. We can perform a round-robin policy to run the work on all shards, this cause more copying from the write-only shard to read-only shard. 

I have an idea to replace the shared<gossiper> semantics to the following:

- We have at most two copies of gossiper object at any time, no matter the number of shards in the system. 
- At any time, one shard owns the write-only copy, all other shards perform read access to the read-only shard. 

This saves memory and copying from one shard to all other nodes. It also makes it much easier to run gossiper run any shards.


What do you think?  We can apply such trick to any per shard object like token_metadata as well.


> 
> Another thing is to properly log things - it will be nice that a node will notice it's overloaded (either
> a single shard or multiple) and log it so the admin will notice.

We have stall detectors for this purpose. Anything else we should log?

tgrabiec:
> > Categorizing an event as a bug doesn't permit us to assume that it can't happen because bugs do happen. For example, every crash is a bug, doesn't mean we can assume it won't happen and not fsync the commit log and lose data on a crash.
> 
> I never assumed the bug will not happen. The question is the frequency and the possibility such event. 

Possibility is relevant in a sense that if you think the bug is so improbable that it shouldn't happen, then we can assume that it won't happen. That's not the case here because such "bugs" happened already many times in practice, so we cannot assume they won't happen.

You wouldn't submit this patch if you thought such things cannot happen in practice, and don't need to be handled.

> If scylla stalls 10s of seconds, it is a very serious bug. If we think it is important to let other nodes know such even and mark this node down in case of such stall, we do not even need to modify any of gossip, we could just shutdown the node itself and isolate from the rest of the cluster.

We shouldn't shutdown the node in such a case, that makes the situation worse. It would make the node unavailable for longer, or even not be able to make progress.

Marking the node as "down" doesn't mean the node is actually down. It's not even meant to mean that. It's badly named. It rather detects subjective unavailability of a node.


> > > > The same could be said about a node which is detected DOWN, it may come back up, e.g. when the network partition heals. Why do we kill streaming then? We could wait as well.
> > > 
> > > 
> > > No, we trust gossip's detection.
> > 
> > 
> > Trust to detect what?
> 
> Detect a node down.

That's not what it detects currently. Gossip may show DOWN even if the node is not down.

> 
> > Gossip doesn't tell you whether it's better to abort streaming or wait. So maybe we shouldn't abort right away on node DOWN if that would lose a lot of work.
> 
> Sure gossip does not tell. The fact is no one does and we do not really care because we can resume from where it is left.
> 
> > Of course, it would be best to make the process resumable so there is no tradeoff and no need to wait.
> > The point is, the problem you describe we would have with advertising DOWN when a single shard is temporarily down, we already have, when a node becomes temporarily partitioned from the cluster. So maybe it should be solved in another way, outside gossip.
> 
> Sorry, it is not clear to me what you meant here.

I meant that the following problem which you brought up is moot in the context of the proposed change in gossip behavior:

> UP and DOWN is not only used for request routing, it is also used to decide if a node is down and we should kill all the streaming or repair sessions with that node.

tgrabiec:
> > > No. Scylla process can run but gossip is stopped. So 3 is not true.
> > 
> > 
> > This meaning applies to when the node is UP (if UP then ...). If gossip is stopped then UP will be false, so "Scylla process can run but gossip is stopped" doesn't contradict that "if UP then scylla process is running".
> > If "not UP", then we don't know.
> 
> Exactly, if we don't know what "not UP" means, how is this useful at all.

It still could be, if you have 3 nodes and 2 are UP and one is "not UP", you may prefer the UP nodes because they have a higher likelihood of being able to serve the request.

"not UP" is actually stronger than "don't know". If UP means X, then "not UP" means "not X || we cannot contact the node to establish that".

If some operation requires X to succeed, then from "not UP" you can derive that it cannot succeed on that node because either X doesn't hold or you won't be able to send the request to the node because the network is down. Because if "we cannot contact the node to establish that" then there is a high chance that we also cannot contact the node to send the operation.

avikivity:
Yes, so if we can score nodes by the probablity+confidence they are alive we can select the two highest nodes for quorum reads.

Actions with large consequences would require high confidence in the results before taking the action.

asias:
Patch is merged.

PR #7224, opened 2020-09-13
dist/common/scripts: skip internet access on offline installation

syuu1228:
We need to skip internet access on offline installation.
To do this we need following changes:
 - prevent running yum/apt for each script
 - set default "NO" for scripts it requires package installation
 - set default "NO" for scripts it requires internet access, such as NTP

See #7153

### DIFF
--- a/dist/common/scripts/scylla_kernel_check
+++ b/dist/common/scripts/scylla_kernel_check
@@ -31,12 +31,7 @@ if __name__ == '__main__':
         sys.exit(1)
 
     if not shutil.which('mkfs.xfs'):
-        if is_debian_variant():
-            run('apt-get install -y xfsprogs')
-        elif is_gentoo_variant():
-            run('emerge -uq sys-fs/xfsprogs')
-        elif is_redhat_variant():
-            run('yum install -y xfsprogs')
+        pkg_install('xfsprogs')
 
     makedirs('/var/tmp/mnt')
     run('dd if=/dev/zero of=/var/tmp/kernel-check.img bs=1M count=128', silent=True)
--- a/dist/common/scripts/scylla_ntp_setup
+++ b/dist/common/scripts/scylla_ntp_setup
@@ -43,7 +43,7 @@ if __name__ == '__main__':
 
     if is_debian_variant():
         if not shutil.which('ntpd') or not shutil.which('ntpdate'):
-            run('apt-get install -y ntp ntpdate')
+            apt_install('ntp ntpdate')
         with open('/etc/ntp.conf') as f:
             conf = f.read()
         if not re.match(r'^server ', conf, flags=re.MULTILINE):
@@ -59,7 +59,7 @@ if __name__ == '__main__':
 
     if is_gentoo_variant():
         if not shutil.which('ntpd'):
-            run('emerge -uq net-misc/ntp')
+            emerge_install('net-misc/ntp')
         with open('/etc/ntp.conf') as f:
             conf = f.read()
         match = re.search(r'^server\s+(\S*)(\s+\S+)?', conf, flags=re.MULTILINE)
@@ -80,7 +80,7 @@ if __name__ == '__main__':
     if is_redhat_variant():
         if int(distro.major_version()) >= 8:
             if not shutil.which('chronyd'):
-                run('dnf install -y chrony')
+                yum_install('chrony')
             with open('/etc/chrony.conf') as f:
                 conf = f.read()
             if args.subdomain:
@@ -94,7 +94,7 @@ if __name__ == '__main__':
             run('chronyc makestep')
         else:
             if not shutil.which('ntpd') or not shutil.which('ntpdate'):
-                run('yum install -y ntp ntpdate')
+                yum_install('ntp ntpdate')
             with open('/etc/ntp.conf') as f:
                 conf = f.read()
             if args.subdomain:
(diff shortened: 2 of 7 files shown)


syuu1228:
Pushed new version, added package remove functions (will use another PR)

dist/common/scripts/scylla_ntp_setup
```diff
@@ -26,6 +26,7 @@ import argparse
 import subprocess
 import re
 import distro
+import shutil
```
avikivity, line 29: Needs to be added to fedora_python3_packages.

dist/common/scripts/scylla_ntp_setup
```diff
@@ -41,7 +42,7 @@ if __name__ == '__main__':
     args = parser.parse_args()
 
     if is_debian_variant():
-        if not os.path.exists('/usr/sbin/ntpd') or not os.path.exists('/usr/sbin/ntpdate'):
+        if not shutil.which('ntpd') or not shutil.which('ntpdate'):
```
avikivity, line 45: Why? Please explain in the changelog.

dist/common/scripts/scylla_kernel_check
```diff
@@ -31,12 +31,7 @@ if __name__ == '__main__':
         sys.exit(1)
 
     if not shutil.which('mkfs.xfs'):
-        if is_debian_variant():
-            run('apt-get install -y xfsprogs')
-        elif is_gentoo_variant():
-            run('emerge -uq sys-fs/xfsprogs')
-        elif is_redhat_variant():
-            run('yum install -y xfsprogs')
+        pkg_install('xfsprogs')
```
avikivity, line 34: I think this breaks gentoo.

Maybe we can do something like this:

    pkg_install('xfsprogs', {'gentoo': 'sys-fs/xfsprogs'})

To provide exceptional names when they aren't the same.

avikivity: Actually, `emerge -uq xfsprogs` works (well I see it start, I'm not going to wait for it to finish).

The idea still makes sense, but not needed here.

dist/common/scripts/scylla_setup
```diff
@@ -169,11 +169,36 @@ if __name__ == '__main__':
...
+            default_no_raid_setup = True
+
+        if not shutil.which('coredumpctl'):
+            default_no_coredump_setup = True
+
+        if is_debian_variant() and not shutil.which('cpufreq-set'):
+            default_no_cpuscaling_setup = True
+        elif not shutil.which('cpupower'):
+            default_no_cpuscaling_setup = True
+
```
avikivity, line 195: We should print something for what we skip, so the user is aware the installation is partial.

Let's also create a setup log (not in this series) so the user can report problems easily.

penberg:
Slightly off-topic and certainly not something to address in this pull request, but: why do the script have to install _anything_ in the first place? Should not these packages be part of our .rpm and .deb dependencies?

dist/common/scripts/scylla_kernel_check
penberg: I like this approach. I think we could (in a future patch) simplify the code by consolidating the different code paths to use this proposed variant of `pkg_install()` instead of also having OS-specific paths where we call `apt`, `emerge`, and so on.

penberg (approved):
lgtm

dist/common/scripts/scylla_ntp_setup
syuu1228: It's standard library, already used on other scripts

dist/common/scripts/scylla_kernel_check
syuu1228: It's better to passing category name to emerge command that is true, but the command should work unless there are same name packages between different category.
If emerge detects multiple packages, it causes error like this:
```
6b5e41425118 / # emerge -uq mysql

 * IMPORTANT: 7 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.



!!! The short ebuild name "mysql" is ambiguous. Please specify
!!! one of the following fully-qualified ebuild names instead:

    acct-group/mysql
    acct-user/mysql
    dev-db/mysql
    dev-haskell/mysql
    virtual/mysql
```
So current code should work in limited condition: it's only package name in all categories.
If there are two or more, then we have to pass package name with category.

But maybe it's not clean approach, we can have dict to resolve full package name for emerge, something like this `{'xfsprogs': 'sys-fs/xfsprogs'}`

dist/common/scripts/scylla_setup
syuu1228: Added messages to warn some setup are skipped.

dist/common/scripts/scylla_ntp_setup
syuu1228: Added commit message.

syuu1228:
Rebased with latest next, added changes where pointed out on earlier comments.

dist/common/scripts/scylla_ntp_setup
avikivity: ok

dist/common/scripts/scylla_kernel_check
avikivity: Yes, a side dict for translating gentoo package names is a good idea.

avikivity:
> Slightly off-topic and certainly not something to address in this pull request, but: why do the script have to install _anything_ in the first place? Should not these packages be part of our .rpm and .deb dependencies?

This is for the tarball, no? tarballs don't have dependencies.

PR #7229, opened 2020-09-14
scylla_ntp_setup: use chrony on all distributions

syuu1228:
To simplify scylla_ntp_setup, use chrony on all distributions.

### DIFF
--- a/dist/common/scripts/scylla_ntp_setup
+++ b/dist/common/scripts/scylla_ntp_setup
@@ -36,70 +36,28 @@ if __name__ == '__main__':
         sys.exit(1)
 
     parser = argparse.ArgumentParser(description='Optimize NTP client setting for Scylla.')
-    if is_redhat_variant():
-        parser.add_argument('--subdomain',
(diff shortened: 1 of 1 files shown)


avikivity:
Is chrony available everywhere?

I remember something about chrony being targeted at desktops while ntp is for servers, but maybe it's wrong.

syuu1228:
> Is chrony available everywhere?
Looks like all supported distributions have it:
Debian: 8, 9, 10 Ubuntu: 16.04, 18.04, 20.04 CentOS: 7, 8

> I remember something about chrony being targeted at desktops while ntp is for servers, but maybe it's wrong.
Not sure about this, but at least RHEL7/8 and SLES15 shipped with chrony by default, and also RHEL8/SLES15 dropped ntpd, I guess it should be fine.

avikivity:
Agree it should be fine.

syuu1228:
Rebased, able to merge now.

avikivity:
Conflicts (probably the scylla_util simplification).

syuu1228:
@avikivity rebased.

PR #7231, opened 2020-09-14
[DONOTMERGE] scylla_cpuscaling_setup: use scylla-cpupower.service on all distributions

### LINKED ISSUE #7230: scylla_cpuscaling_setup: Got warning when installing scylla-cpupower.service
Got following warning message when installing scylla-cpupower.service on Fedora 32.
The service actually only for Amazon Linux 2, the distribution does not shows up such warning,
but looks like [Install] section is needed anyway. 

```
[syuu@monster ~]$ sudo systemctl enable scylla-cpupower
The unit files have no installation config (WantedBy=, RequiredBy=, Also=,
Alias= settings in the [Install] section, and DefaultInstance= for template
units). This means they are not meant to be enabled using systemctl.
 
Possible reasons for having this kind of units are:
• A unit may be statically enabled by being symlinked from another unit's
  .wants/ or .requires/ directory.
• A unit's purpose may be to act as a helper for some other unit which has
  a requirement dependency on it.
• A unit may be started when needed via activation (socket, path, timer,
  D-Bus, udev, scripted systemctl call, ...).
• In case of template units, the unit is meant to be enabled with some
  instance name specified.
```

syuu1228:
This PR requires to merge #7224, do not merge now.

----

Use scylla-cpupower.service on all distributions, to simplify the setup script
and also to minimize difference of Scylla running environment between distributions.

Also contain 'add Install section' patch to fix #7230

### DIFF
--- a/dist/common/scripts/scylla_coredump_setup
+++ b/dist/common/scripts/scylla_coredump_setup
@@ -46,7 +46,7 @@ if __name__ == '__main__':
 # Other distributions can use systemd-coredump, so setup it
     else:
         if is_debian_variant():
-            run('apt-get install -y systemd-coredump')
+            apt_install('systemd-coredump')
         conf_data = '''
 [Coredump]
 Storage=external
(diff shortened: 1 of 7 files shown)


avikivity:
Well there is some advantage to using the standard cpupower.service if it exists, it is easier to interact with for users who are familiar with it.

PR #7233, opened 2020-09-14
scylla_setup: drop hugepages package installation

### LINKED ISSUE #7182: scylla_setup doesn't support to skip install of hugepages or libhugetlbfs-bin package
*Installation details*
Scylla version (or git commit hash):  latest master (b3f00685ec6c04c67fc9db809fbb63cb1696881e)
Cluster size: 1
OS (RHEL/CentOS/Ubuntu/AWS AMI): Ubuntu 20.04

hugepages or libhugetlbfs-bin package is installed for Ubuntu 20.04 and later.
scylla_setup should support to disable it, we have some users that run scylla_setup without network.

image

@slivne @penberg

syuu1228:
hugepages and libhugetlbfs-bin packages is only required for DPDK mode,
and unconditionally installation causes error on offline mode, so drop it.

Fixes #7182

### DIFF
--- a/dist/common/scripts/scylla_setup
+++ b/dist/common/scripts/scylla_setup
@@ -502,11 +502,3 @@ if __name__ == '__main__':
         if selinux_reboot_required:
             print('Please restart your machine before using ScyllaDB, as you have disabled')
             print(' SELinux.')
-
-        if distro.id() == 'ubuntu':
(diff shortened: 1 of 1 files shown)


syuu1228:
Note: #7224 will conflict with this PR, need to rebase after merging this.

PR #7235, opened 2020-09-15
Remove _pending_ranges and _pending_ranges_map in token_metadata

asias:
This PR removes  _pending_ranges and _pending_ranges_map in token_metadata.
This removal of  makes copying of token_metadata faster and reduces the chance to cause reactor stall.
    
Refs: #7220

### DIFF
--- a/locator/token_metadata.cc
+++ b/locator/token_metadata.cc
@@ -68,7 +68,6 @@ private:
     std::unordered_map<inet_address, inet_address> _replacing_endpoints;
 
     std::unordered_map<sstring, std::unordered_multimap<range<token>, inet_address>> _pending_ranges;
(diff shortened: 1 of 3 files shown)


locator/token_metadata.cc
```diff
@@ -1355,24 +1349,22 @@ void token_metadata_impl::set_pending_ranges(const sstring& keyspace_name,
         _pending_ranges_interval_map[keyspace_name] +=
```
bhalevy, line 1349: As a follow-up, we don't need to lookup `keyspace_name` in `_pending_ranges_interval_map` for every entry.
We can build the keyspace interval map and when done move it to `_pending_ranges_interval_map[keyspace_name]`

bhalevy (approved):
Looks good to me.
See comment with suggested follow up

PR #7241, opened 2020-09-15
reconcilable_result_builder: don't aggrevate out-of-memory condition during recovery

### LINKED ISSUE #7240: reconcilable_result_builder can trigger OOM failure during error recovery
```c++
class reconcilable_result_builder {
    const schema& _schema;
    const query::partition_slice& _slice;

    utils::chunked_vector<partition> _result;
    uint32_t _live_rows{};

    bool _has_ck_selector{};
    bool _static_row_is_alive{};
    uint32_t _total_live_rows = 0;
    query::result_memory_accounter _memory_accounter;
    stop_iteration _stop;
    bool _short_read_allowed;
    stdx::optional<streamed_mutation_freezer> _mutation_consumer;
```

Let's say we have a reconcilable_result_builder that ate all memory in its `_result` member. We can out of memory, so we destroy it. C++ destroys last member first, so _memory_accounter is destroyed before _result. This causes requests that wait for memory to be resumed before _result is destroyed, aggravating the OOM, and in one observed instance, causing a crash on bad_alloc when increasing the task queue size.

avikivity:
Consider an unpaged query that consumes all of available memory, despite
fea5067dfa339ddf36d86704a8bdbb671efb27cf which limits them (perhaps the
user raised the limit, or this is a system query). Eventually we will see a
bad_alloc which will abort the query and destroy this reconcilable_result_builder.

During destruction, we first destroy _memory_accounter, and then _result.
Destroying _memory_accounter resumes some continuations which can then
allocate memory synchronously when increasing the task queue to accomodate
them. We will then crash. Had we not crashed, we would immediately afterwards
release _result, freeing all the memory that we would ever need.

Fix by making _result the last member, so it is freed first.

Fixes #7240.

### DIFF
--- a/mutation_query.hh
+++ b/mutation_query.hh
@@ -140,15 +140,16 @@ class reconcilable_result_builder {
     const schema& _schema;
     const query::partition_slice& _slice;
 
-    utils::chunked_vector<partition> _result;
-    uint64_t _live_rows{};
-
     bool _return_static_content_on_partition_with_no_rows{};
     bool _static_row_is_alive{};
     uint64_t _total_live_rows = 0;
     query::result_memory_accounter _memory_accounter;
     stop_iteration _stop;
     std::optional<streamed_mutation_freezer> _mutation_consumer;
+
+    uint64_t _live_rows{};
+    // make this the last member so it is destroyed first. #7240
+    utils::chunked_vector<partition> _result;
 public:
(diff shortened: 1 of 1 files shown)


tgrabiec:
Queued as 9421cfd

tgrabiec (approved):

PR #7246, opened 2020-09-16
alternator: Use vnode token ranges for shards instead of 1:1 stream I…

elcallio:
…D mapping

CDC stream ids are tied to (scylla) shards per vnode. This means very
many ids per cluster, even with a low number of nodes.

Dynamodb typically expects the number of "shards" in a stream to be much lower.

This patch makes each stream shard a vnode (using the embedded "index" classifier
of cdc stream id), and uses token range query instead of id = <value> (or id in
{ ...values }). This means we can still keep alternator streams stateless (all
info is encoded in the various json identifiers).

There are caveats though: To avoid querying into an old generation, we also
need to make the timestamp restrictions for query adjust according to a shards
creation and ev. expiry time. This is a little problematic for an iterator
with a shards from the latest generation, if a new generation is created
right in the middle of operations.

We assume this is rare, and (mis-)use the local gossip app state for cdc
generation timestamp to check if it seems we expired. If so, we simply
throw. It should be rare enough.

For old shard iterators, we simply adjust the high threshold for query.
This in turn means that we can include writes made to a new generation
in a new generation query, since we ignore confidence overlap. However,
this is not a problem since we similarly filter out this data from the
new generation query.

Queries to an old generation might be slower though, since they might
hit more than one current vnode. But again, this should be rare, and
adjust itself quickly (catching up).

An added benefit is that we can more easily say that iterators have
hit end-of-data, since end timestamp is now embedded.

Also fixes string length restrictions on ShardId:s. Dynamo define them
as max 65 chars, which we broke even before.

Note that the magic string markers for ShardId and ShardIterator
remain the same, since alternator streams are not GA yet, thus
a format change is ok.

### DIFF
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -87,16 +87,18 @@ struct rapidjson::internal::TypeHelper<ValueType, utils::UUID>
     : public from_string_helper<ValueType, utils::UUID>
 {};
 
+static db_clock::time_point as_timepoint(const utils::UUID& uuid) {
+    return db_clock::time_point{std::chrono::milliseconds(utils::UUID_gen::get_adjusted_timestamp(uuid))};
+}
+
 /**
  * Note: scylla tables do not have a timestamp as such, 
  * but the UUID (for newly created tables at least) is 
  * a timeuuid, so we can use this to fake creation timestamp
  */
 static sstring stream_label(const schema& log_schema) {
-    auto& uuid = log_schema.id();
-    auto ms = utils::UUID_gen::get_adjusted_timestamp(uuid);
-    std::chrono::system_clock::time_point ts{std::chrono::milliseconds(ms)};
-    auto tt = std::chrono::system_clock::to_time_t(ts);
+    auto ts = as_timepoint(log_schema.id());
+    auto tt = db_clock::to_time_t(ts);
     struct tm tm;
     ::localtime_r(&tt, &tm);
     return seastar::json::formatter::to_json(tm);
@@ -218,40 +220,79 @@ future<alternator::executor::request_return_type> alternator::executor::list_str
     return make_ready_future<executor::request_return_type>(make_jsonable(std::move(ret)));
 }
 
+/**
+ * Note: because "end_time" will be zero for non-closed generation
+ * shards, but non-zero once said generation expires, we will in 
+ * fact give "multiple versions" of the same shard, i.e. a shard
+ * will change ID (as presented to the user).
+ * 
+ * The dynamo specs do not forbid this, but we are (probably)
+ * expected to allow a shard that is still valid (not 24h+ old)
+ * to still be queried, hence we "repair" info in iterator 
+ * query later on.
+ * 
+ * The alternative would be to either be stateful, or query
+ * end timestamp on get_shard_iterator. But we cannot currently 
+ * do this in any really efficient way for a single shard, so
+ * let's be mutable instead.
+ * 
+ */
 struct shard_id {
     static constexpr auto marker = 'H';
+    static constexpr auto hex_width = sizeof(uint64_t)*8/4;
 
-    utils::UUID table;
-    db_clock::time_point time;
-    cdc::stream_id id;
+    db_clock::time_point start_time;
+    db_clock::time_point end_time; // zero if open
+    dht::token lo_token; // range start, inclusive (first id of vnode)
+    dht::token hi_token; // range end, inclusive (last id of vnode)
 
-    shard_id(utils::UUID tab, db_clock::time_point t, cdc::stream_id i)
-        : table(tab)
-        , time(t)
-        , id(i)
-    {}
+    shard_id(db_clock::time_point start,  db_clock::time_point end, dht::token lo, dht::token hi)
+        : start_time(start)
+        , end_time(end)
+        , lo_token(lo)
+        , hi_token(hi)
+    {
+        // cdc generations cannot really have start==end, but for our purposes
+        // it is still passable/valid, so allow it.
+        assert(end_time == db_clock::time_point{} || start_time < end_time);
+    }
     shard_id(const sstring&);
 
+    bool has_expired() const {
+        return end_time != db_clock::time_point{};
+    }
+    bool is_wraparound() const {
+        return lo_token > hi_token;
+    }
+
+    // dynamo specifies shardid as max 65 chars. Thus we cannot use separators, 
+    // instead we have to rely on 16 width hex
     friend std::ostream& operator<<(std::ostream& os, const shard_id& id) {
         boost::io::ios_flags_saver fs(os);
-        return os << marker << id.table << std::hex
-            << ':' << id.time.time_since_epoch().count()
-            << ':' << id.id.to_bytes()
+        return os << marker << std::hex << std::setfill('0') 
+            << std::setw(hex_width) << id.start_time.time_since_epoch().count()
+            << std::setw(hex_width) << id.end_time.time_since_epoch().count()
+            << std::setw(hex_width) << (id.lo_token.is_minimum() ? 0ull : dht::token::to_int64(id.lo_token))
+            << std::setw(hex_width) << (id.hi_token.is_maximum() ? 0ull : dht::token::to_int64(id.hi_token))
             ;
     }
 };
 
 shard_id::shard_id(const sstring& s) {
-    auto i = s.find(':');
-    auto j = s.find(':', i + 1);
-
-    if (s.at(0) != marker || i == sstring::npos || j == sstring::npos) {
+    if (s.at(0) != marker || s.size() != (1 + hex_width * 4)) {
         throw std::invalid_argument(s);
     }
 
-    table = utils::UUID(s.substr(1, i));
-    time = db_clock::time_point(db_clock::duration(std::stoull(s.substr(i + 1, j), nullptr, 16)));
-    id = cdc::stream_id(from_hex(s.substr(j + 1)));
+    start_time = db_clock::time_point(db_clock::duration(std::stoull(s.substr(1, hex_width), nullptr, 16)));
+    end_time = db_clock::time_point(db_clock::duration(std::stoull(s.substr(1 + hex_width, hex_width), nullptr, 16)));
+    auto lo = std::stoull(s.substr(1 + 2*hex_width, hex_width), nullptr, 16);
+    auto hi = std::stoull(s.substr(1 + 3*hex_width), nullptr, 16);
+    lo_token = lo == 0 ? dht::minimum_token() : dht::token::from_int64(lo);
+    hi_token = hi == 0 ? dht::maximum_token() : dht::token::from_int64(hi);
+
+    if (end_time != db_clock::time_point{} && end_time < start_time) {
+        throw std::invalid_argument(format("Invalid ShardID {}. Start={}, End={}", s, start_time, end_time));
+    }
 }
 
 struct sequence_number {
@@ -359,52 +400,26 @@ using namespace std::string_literals;
 
 /**
  * This implementation of CDC stream to dynamo shard mapping
- * is simply a 1:1 mapping, id=shard. Simple, and fairly easy
- * to do ranged queries later.
+ * is making a shard equal to a vnode token range, by using the 
+ * cdc stream id "index" grouping. 
  * 
- * Downside is that we get a _lot_ of shards. And with actual
- * generations, we'll get hubba-hubba loads of them. 
+ * We stil get a lot of shards, but fewer than with 1:1 mapping
+ * cdc stream -> shard. This means our queries will cross actual
+ * scylla shards, but that is not neccessarily a bad thing.
  * 
- * Improvement ideas: 
- * 1.) Prune expired streams that cannot have records (check ttls)
- * 2.) Group cdc streams into N per shards (by token range)
- * 
- * The latter however makes it impossible to do a simple 
- * range query with limit (we need to do staggered querying, because
- * of how GetRecords work).
- * The results of a select from cdc log on more than one cdc stream,
- * where timestamp < T will typically look like: 
- * 
- * <shard 0> <ts0> ...
- * ....
- * <shard 0> <tsN> ...
- * <shard 1> <ts0p> ...
- * ...
- * <shard 1> <tsNp> ...
- *
- * Where tsN < timestamp and tsNp < timestamp, but range [ts0-tsN] and [ts0p-tsNp]
- * overlap. 
- * 
- * Now, if we have limit=X, we might end on a row inside shard 0.
- * For our next query, we would in the simple case just say "timestamp > ts3 and timestamp < T",
- * but of course this would then miss rows in shard 1 and above. 
- * 
- * However if we instead set the conditions to 
- * 
- *   "(token(stream_id) = token(shard 0) and timestamp > ts3 and timestamp < T)
- *      or (token(stream_id) > token(shard 0) and timestamp < T)"
- * 
- * we can get the correct data. Question is however how well this will perform, if at 
- * all.
- * 
- * For now, go simple, but maybe consider if this latter approach would work
- * and perform. 
- * 
+ * Note we could possibly reduce the number of shards by simply 
+ * splitting the id/token range into N (chosen by some critera, 
+ * probably ring size) and hope the range queries crossing vnode
+ * bounds are not that bad. To do so, it is a fairly easy change
+ * in the code below.
  */
 
 static std::chrono::seconds confidence_interval(const database& db) {
     return std::chrono::seconds(db.get_config().alternator_streams_time_window_s());
 }
+
+// Dynamo docs says no data shall live longer than 24h.
+static constexpr auto dynamo_streams_max_window = 24h;
 
 future<executor::request_return_type> executor::describe_stream(client_state& client_state, service_permit permit, rjson::value request) {
     _stats.api_operations.describe_stream++;
@@ -479,50 +494,125 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
     // or on expired...
     // TODO: maybe add secondary index to topology table to enable this?
     return _sdks.cdc_get_versioned_streams({ tm.count_normal_token_owners() }).then([this, &db, schema, shard_start, limit, ret = std::move(ret), stream_desc = std::move(stream_desc)](std::map<db_clock::time_point, cdc::streams_version> topologies) mutable {
-        auto i = topologies.begin();
+
+        // filter out cdc generations older than the table or now() - dynamo_streams_max_window (24h)
+        auto low_ts = std::max(as_timepoint(schema->id()), db_clock::now() - dynamo_streams_max_window);
+
+        auto i = topologies.lower_bound(low_ts);
+        // need first gen _intersecting_ the timestamp.
+        if (i != topologies.begin()) {
+            i = std::prev(i);
+        }
+
         auto e = topologies.end();
         auto prev = e;
-        auto threshold = db_clock::now() - confidence_interval(db);
         auto shards = rjson::empty_array();
 
         std::optional<shard_id> last;
+
+        // i is now at the youngest generation we include. make a mark of it.
+        auto first = i;
+
+        // if we're a paged query, skip to the generation where we left of.
+        if (shard_start) {
+            i = topologies.find(shard_start->start_time);
+        }
+
+        // need a prev even if we are skipping stuff
+        if (i != first) {
+            prev = std::prev(i);
+            std::stable_sort(prev->second.streams.begin(), prev->second.streams.end(), [](auto& id1, auto& id2) { 
+                return id1.token() < id2.token(); 
+            });
+        }
 
         for (; limit > 0 && i != e; prev = i, ++i) {
             auto& [ts, sv] = *i;
 
             last = std::nullopt;
 
-            for (auto& id : sv.streams()) {
-                if (shard_start && shard_start->id != id) {
-                    continue;
+            auto get_expiration = [&](auto iter) {
+                if (++iter != e) {
+                    return iter->first;
                 }
-                if (shard_start && shard_start->id == id) {
-                    shard_start = std::nullopt;
-                    continue;
-                }
+                return db_clock::time_point{};
+            };
+
+            auto lo = sv.streams.begin();
+            auto end = sv.streams.end();
+
+            // first sort by index.
+            std::stable_sort(lo, end, [](auto& id1, auto& id2) {
+                return id1.index() < id2.index();
+            });
+
+            if (shard_start) {
+                // find next shard position
+                // first id set is the wraparound range. skip it here.
+                lo = std::find_if(lo, end, [](const cdc::stream_id& id) { return id.index() != 0; });
+                lo = std::upper_bound(lo, end, shard_start->hi_token, [](const dht::token& t, const cdc::stream_id& id) {
+                    return t < id.token();
+                });
+                shard_start = std::nullopt;
+            }
+
+            for (auto hi = lo; lo != end; lo = hi) {
+                auto& id = *lo++;
+
+                hi = std::find_if(lo, end, [i = id.index()](const cdc::stream_id& id) {
+                    return id.index() != i;
+                });
+                auto t = id.token();
+                auto et = std::prev(hi)->token();
+
+                last.emplace(ts, get_expiration(i), t, et);
 
                 auto shard = rjson::empty_object();
 
                 if (prev != e) {
-                    auto token = id.token();
-                    auto& pids = prev->second.streams();
-                    auto pid = std::upper_bound(pids.begin(), pids.end(), token, [](const dht::token& t, const cdc::stream_id& id) {
+                    auto& pids = prev->second.streams;
+                    // maybe re-sort by token, so we can find the previous shard bounds
+                    // if prev is sorted by index, we can just rotate it slightly to make it sorted by
+                    // token.
+                    if (pids.front().token() > pids.back().token()) {
+                        auto nf = std::find_if(pids.begin(), pids.end(), [t = pids.back().token()](const cdc::stream_id& id) {
+                            return id.token() < t;
+                        });
+                        std::rotate(pids.begin(), nf, pids.end());
+                    }
+                    auto pid = std::upper_bound(pids.begin(), pids.end(), t, [](const dht::token& t, const cdc::stream_id& id) {
                         return t < id.token();
                     });
+                    if (pid != pids.begin()) {
+                        pid = std::prev(pid);
+                    }
                     if (pid != pids.end()) {
-                        rjson::set(shard, "ParentShardId", shard_id(schema->id(), prev->first, *pid));
+                        auto index = pid->index();
+                        auto cmp = [index](const cdc::stream_id& id) {
+                            return id.index() != index;
+                        };
+                        // reverse iteration: base will point to elem matching index
+                        auto plo = index == 0
+                            ? std::find_if(pids.rbegin(), pids.rend(), cmp).base()
+                            : std::find_if(std::make_reverse_iterator(pid), pids.rend(), cmp).base()
+                            ;
+                        auto phi = index == 0
+                            ? std::find_if(pids.begin(), pids.end(), cmp)
+                            : std::find_if(pid, pids.end(), cmp)
+                            ;
+                        auto lt = plo->token();
+                        auto et = std::prev(phi)->token();
+                        rjson::set(shard, "ParentShardId", shard_id(prev->first, get_expiration(prev), lt, et));
                     }
                 }
 
-                last = shard_id(schema->id(), ts, id);
-                rjson::set(shard, "ShardId", *last);
-
                 auto range = rjson::empty_object();
                 rjson::set(range, "StartingSequenceNumber", sequence_number(utils::UUID_gen::min_time_UUID(ts.time_since_epoch().count())));
-                if (sv.expired() && *sv.expired() < threshold) {
-                    rjson::set(range, "EndingSequenceNumber", sequence_number(utils::UUID_gen::max_time_UUID((*sv.expired() + confidence_interval(db)).time_since_epoch().count())));
+                if (last->has_expired()) {
+                    rjson::set(range, "EndingSequenceNumber", sequence_number(utils::UUID_gen::min_time_UUID(last->end_time.time_since_epoch().count())));
                 }
 
+                rjson::set(shard, "ShardId", *last);
                 rjson::set(shard, "SequenceNumberRange", std::move(range));
                 rjson::push_back(shards, std::move(shard));
                 
@@ -585,44 +675,39 @@ struct shard_iterator {
     static auto constexpr inclusive_marker = 'I';
     static auto constexpr exclusive_marker = 'i';
 
-    shard_id shard;
+    static auto constexpr uuid_str_length = 36;
+
+    utils::UUID table;
     utils::UUID threshold;
+    shard_id shard;
     bool inclusive;
 
-    shard_iterator(shard_id i, utils::UUID th, bool inclusive)
-        : shard(i)
-        , threshold(th)
-        , inclusive(inclusive)
-    {}
-    shard_iterator(utils::UUID arn, db_clock::time_point ts, cdc::stream_id i, utils::UUID th, bool inclusive)
-        : shard(arn, ts, i)
-        , threshold(th)
-        , inclusive(inclusive)
-    {}
     shard_iterator(const sstring&);
 
+    shard_iterator(utils::UUID t, shard_id i, utils::UUID th, bool inclusive)
+        : table(t)
+        , threshold(th)
+        , shard(i)
+        , inclusive(inclusive)
+    {}
     friend std::ostream& operator<<(std::ostream& os, const shard_iterator& id) {
-        boost::io::ios_flags_saver fs(os);
-        return os << (id.inclusive ? inclusive_marker : exclusive_marker) << std::hex << id.threshold
+        return os << (id.inclusive ? inclusive_marker : exclusive_marker) << std::hex
+            << id.table
+            << ':' << id.threshold
             << ':' << id.shard
             ;
     }
-private:
-    shard_iterator(const sstring&, size_t);
 };
 
-shard_iterator::shard_iterator(const sstring& s) 
-    : shard_iterator(s, s.find(':'))
-{}
-
-shard_iterator::shard_iterator(const sstring& s, size_t i)
-    : shard(s.substr(i + 1))
-    , threshold(s.substr(1, i))
+shard_iterator::shard_iterator(const sstring& s)
+    : table(s.substr(1, uuid_str_length))
+    , threshold(s.substr(2 + uuid_str_length, uuid_str_length))
+    , shard(s.substr(3 + 2 * uuid_str_length))
+    , inclusive(s.at(0) == inclusive_marker)
 {
-    if ((s.at(0) != inclusive_marker && s.at(0) != exclusive_marker) || i == sstring::npos) {
+    if (s.at(0) != inclusive_marker && s.at(0) != exclusive_marker) {
         throw std::invalid_argument(s);
     }
-    inclusive = (s[0] == inclusive_marker);
 }
 }
 
@@ -663,56 +748,87 @@ future<executor::request_return_type> executor::get_shard_iterator(client_state&
         sid = rjson::get<shard_id>(request, "ShardId");
     } catch (...) {
     }
-    if (!schema || !cdc::get_base_table(db, *schema) || !is_alternator_keyspace(schema->ks_name()) || sid->table != schema->id()) {
+    if (!schema || !cdc::get_base_table(db, *schema) || !is_alternator_keyspace(schema->ks_name())) {
         throw api_error::resource_not_found("Invalid StreamArn");
     }
-
-    utils::UUID threshold;
-    bool inclusive_of_threshold;
-
-    switch (type) {
-        case shard_iterator_type::AT_SEQUENCE_NUMBER:
-            threshold = *seq_num;
-            inclusive_of_threshold = true;
-            break;
-        case shard_iterator_type::AFTER_SEQUENCE_NUMBER:
-            threshold = *seq_num;
-            inclusive_of_threshold = false;
-            break;
-        case shard_iterator_type::TRIM_HORIZON:
-            threshold = utils::UUID{};
-            inclusive_of_threshold = true;
-            break;
-        case shard_iterator_type::LATEST:
-            threshold = utils::UUID_gen::min_time_UUID((db_clock::now() - confidence_interval(db)).time_since_epoch().count());
-            inclusive_of_threshold = true;
-            break;
+    if (!sid) {
+        throw api_error::resource_not_found("Invalid ShardId");
     }
 
-    shard_iterator iter(*sid, threshold, inclusive_of_threshold);
+    auto f = make_ready_future<shard_id>(*sid);
 
-    auto ret = rjson::empty_object();
-    rjson::set(ret, "ShardIterator", iter);
+    if (!sid->has_expired()) {
+        // see comment in get_records. Need to check if this (open)
+        // shard is in fact still valid.
+        auto& ep = utils::fb_utilities::get_broadcast_address();
+        auto ts = cdc::get_streams_timestamp_for(ep, _ss.gossiper());
+        if (ts != db_clock::time_point{} &&  ts != sid->start_time) {
+            // The shard expired between user queried it and 
+            // this call. Client probably does not expect a shard ID to 
+            // "change"/disappear like this sort of does (replaced with
+            // one with an end time in it), so we'll simply repair it
+            // in place and give an iterator that knows its end time.
+            // Hopefully the client will then read until end-of-shard
+            // data using this iterator, and can move on to next
+            // describe_stream result set next.
+            const auto& tm = _proxy.get_token_metadata();
+            f = _sdks.get_next_stream_timestamp({ tm.count_normal_token_owners() }, sid->start_time).then([sid = *sid](db_clock::time_point ts) mutable {
+                sid.end_time = ts;
+                return sid;
+            });
+        }
+    }
 
-    return make_ready_future<executor::request_return_type>(make_jsonable(std::move(ret)));
+    return f.then([&db, stream_arn, schema, type, seq_num = std::move(seq_num)](shard_id sid) {
+        if (sid.has_expired() && (db_clock::now() - dynamo_streams_max_window) <= sid.end_time) {
+            throw api_error::trimmed_data_access_exception("Expired shard");
+        }
+
+        utils::UUID threshold;
+        bool inclusive_of_threshold;
+
+        switch (type) {
+            case shard_iterator_type::AT_SEQUENCE_NUMBER:
+                threshold = *seq_num;
+                inclusive_of_threshold = true;
+                break;
+            case shard_iterator_type::AFTER_SEQUENCE_NUMBER:
+                threshold = *seq_num;
+                inclusive_of_threshold = false;
+                break;
+            case shard_iterator_type::TRIM_HORIZON:
+                threshold = utils::UUID_gen::min_time_UUID(sid.start_time.time_since_epoch().count());
+                inclusive_of_threshold = true;
+                break;
+            case shard_iterator_type::LATEST:
+                threshold = utils::UUID_gen::min_time_UUID((db_clock::now() - confidence_interval(db)).time_since_epoch().count());
+                inclusive_of_threshold = true;
+                break;
+        }
+
+        shard_iterator iter(stream_arn, sid, threshold, inclusive_of_threshold);
+
+        auto ret = rjson::empty_object();
+        rjson::set(ret, "ShardIterator", iter);
+
+        return make_ready_future<executor::request_return_type>(make_jsonable(std::move(ret)));
+    });
 }
 
 struct event_id {
-    cdc::stream_id stream;
+    utils::UUID stream;
     utils::UUID timestamp;
 
     static const auto marker = 'E';
 
-    event_id(cdc::stream_id s, utils::UUID ts)
+    event_id(utils::UUID s, utils::UUID ts)
         : stream(s)
         , timestamp(ts)
     {}
     
     friend std::ostream& operator<<(std::ostream& os, const event_id& id) {
         boost::io::ios_flags_saver fs(os);
-        return os << marker << std::hex << id.stream.to_bytes()
-            << ':' << id.timestamp
-            ;
+        return os << marker << id.stream << ':' << id.timestamp;
     }
 };
 }
@@ -738,24 +854,70 @@ future<executor::request_return_type> executor::get_records(client_state& client
     auto& db = _proxy.get_db().local();
     schema_ptr schema, base;
     try {
-        auto& log_table = db.find_column_family(iter.shard.table);
+        auto& log_table = db.find_column_family(iter.table);
         schema = log_table.schema();
         base = cdc::get_base_table(db, *schema);
     } catch (...) {        
     }
 
     if (!schema || !base || !is_alternator_keyspace(schema->ks_name())) {
-        throw api_error::resource_not_found(boost::lexical_cast<std::string>(iter.shard.table));
+        throw api_error::resource_not_found(boost::lexical_cast<std::string>(iter.table));
     }
 
     tracing::add_table_name(trace_state, schema->ks_name(), schema->cf_name());
 
+    auto f = make_ready_future<shard_iterator>(iter);
+
+    if (!iter.shard.has_expired()) {
+        // ugh. figure out if we have end-of-shard
+        // We try to avoid the extra query by using the
+        // gossip state as a cache check - i.e. know if
+        // topology has changed and invalidated the current
+        // iterator. This is far from exact, so we could still
+        // generate duplicate records at generation changes
+        // unless our confidence window is large enough.
+        auto& ep = utils::fb_utilities::get_broadcast_address();
+        auto ts = cdc::get_streams_timestamp_for(ep, _ss.gossiper());
+        if (ts != db_clock::time_point{} &&  ts != iter.shard.start_time) {
+            // Sigh. there seem to be at least one new generation after this.
+            // Do a query for our actual end timestamp (i.e. start of generation
+            // after use), and update the iterator.
+            // We could just say "expired iterator". This would cause kinesis et all 
+            // to get a new iterator with the same shard ID though, which is why
+            // we replace end time in get_shard_iterator if needed. But we can just as
+            // well save the round trip and update the iterator ourselves.
+            const auto& tm = _proxy.get_token_metadata();
+            f = _sdks.get_next_stream_timestamp({ tm.count_normal_token_owners() }, iter.shard.start_time).then([iter](db_clock::time_point ts) mutable {
+                iter.shard.end_time = ts;
+                return iter;
+            });
+        }
+    }
+
+    return f.then([this, schema, base, limit, &client_state, permit = std::move(permit)](shard_iterator iter) mutable {
+        return do_get_records(client_state, std::move(permit), schema, base, limit, iter);
+    }).finally([this, start_time] {
+        _stats.api_operations.get_records_latency.add(std::chrono::steady_clock::now() - start_time);
+    });
+}
+
+future<executor::request_return_type> executor::do_get_records(client_state& client_state, service_permit permit, schema_ptr schema, schema_ptr base, size_t limit, shard_iterator iter) {
     db::consistency_level cl = db::consistency_level::LOCAL_QUORUM;
-    partition_key pk = iter.shard.id.to_partition_key(*schema);
 
-    dht::partition_range_vector partition_ranges{ dht::partition_range::make_singular(dht::decorate_key(*schema, pk)) };
+    auto lo_pk_bound = dht::partition_range::bound(dht::ring_position::starting_at(iter.shard.lo_token));
+    auto hi_pk_bound = dht::partition_range::bound(dht::ring_position::ending_at(iter.shard.hi_token));
 
+    auto partition_ranges = iter.shard.is_wraparound()
+        ? dht::partition_range_vector{ dht::partition_range::make_starting_with(std::move(lo_pk_bound)), dht::partition_range::make_ending_with(std::move(hi_pk_bound))}
+        : dht::partition_range_vector{ dht::partition_range::make(std::move(lo_pk_bound), std::move(hi_pk_bound)) }
+        ;
+
+    auto& db = _proxy.get_db().local();
     auto high_ts = db_clock::now() - confidence_interval(db);
+    if (iter.shard.has_expired() && high_ts > iter.shard.end_time) {
+        high_ts = iter.shard.end_time;
+    }
+
     auto high_uuid = utils::UUID_gen::min_time_UUID(high_ts.time_since_epoch().count());
     auto lo = clustering_key_prefix::from_exploded(*schema, { iter.threshold.serialize() });
     auto hi = clustering_key_prefix::from_exploded(*schema, { high_uuid.serialize() });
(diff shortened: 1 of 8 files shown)


nyh:
> This patch makes each stream shard a vnode (using the embedded "index" classifier
> of cdc stream id), and uses token range query instead of id = (or id in
> { ...values }).

So, if I understand correctly, the number of stream shards will go down by a factor of the number of cores per node, but will still be the product of num_tokens (256 by default) and the number of nodes, which is still a huge number, just less huge?
I guess for some setups this will indeed make a big difference. In my small test setups the fact 256 usually dominates, but I agree in other setups, every factor we eliminate will help.

> There are caveats though: To avoid querying into an old generation

I am not sure I understand what you are trying to avoid here -
Are you trying to avoid the waste of time (i.e., performance issue) to keep querying an old closed-long-ago "stream" (in the CDC terminology) during every read for recent data, even though it will never have new data, so for a read starting at some given time, you want to be able to guess which streams still existed at that time?

> , we also
> need to make the timestamp restrictions for query adjust according to a shards
> creation and ev. expiry time. This is a little problematic for an iterator
> with a shards from the latest generation, if a new generation is created
> right in the middle of operations.

I didn't understand why this is a problem - isn't it ok (and in fact, exactly the plan) that the client continues to read just the old generation's streams until they end - and the *next* request will see also (or only) the new streams?
 
> We assume this is rare, and (mis-)use the local gossip app state for cdc
> generation timestamp to check if it seems we expired. If so, we simply
> throw. It should be rare enough.

I didn't understand this. Maybe I'll understand as I read the code.
 
> For old shard iterators, we simply adjust the high threshold for query.
> This in turn means that we can include writes made to a new generation
> in a new generation query, since we ignore confidence overlap. However,
> this is not a problem since we similarly filter out this data from the
> new generation query.

I didn't understand this paragraph either :(

alternator/streams.cc
```diff
@@ -491,34 +508,107 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
+                hi = std::find_if(lo, end, [i = id.index()](const cdc::stream_id& id) {
+                    return id.index() != i;
+                });
+
+                auto t = id.token();
+
+                auto et = hi != end ? hi->token() : dht::maximum_token();
+                if (et < t) {
+                    assert(id.index() == 0);
+                    // wraparound id. index 0. just treat is as (min, end] instead.
```
nyh, line 579: Oh, I think I got my answer to the wraparound question above.

alternator/streams.cc
```diff
@@ -589,44 +679,40 @@ struct shard_iterator {
...
-{}
-
-shard_iterator::shard_iterator(const sstring& s, size_t i)
-    : shard(s.substr(i + 1))
-    , threshold(s.substr(1, i))
+shard_iterator::shard_iterator(const sstring& s)
+    : table(s.substr(1, uuid_str_length))
+    , threshold(s.substr(2 + uuid_str_length, uuid_str_length))
+    , shard(s.substr(3 + 2 * uuid_str_length))
+    , inclusive(s[0] == inclusive_marker)
```
nyh, line 711: If s comes from the user, you can't do s[0], it might not exist. This is why the old code used s.at(0) before thinking of using s[0]. You can use s.at(0) here, and s[0] in the constructor function body, after you already know s.at(0) didn't throw.

alternator/streams.cc
```diff
@@ -702,21 +791,19 @@ future<executor::request_return_type> executor::get_shard_iterator(client_state&
...
     static const auto marker = 'E';
 
-    event_id(cdc::stream_id s, utils::UUID ts)
+    event_id(utils::UUID s, utils::UUID ts)
         : stream(s)
         , timestamp(ts)
     {}
     
     friend std::ostream& operator<<(std::ostream& os, const event_id& id) {
         boost::io::ios_flags_saver fs(os);
```
nyh, line 805: You don't need flags_saver any more because you removed the std::hex usage.

alternator/streams.cc
```diff
@@ -742,24 +829,51 @@ future<executor::request_return_type> executor::get_records(client_state& client
...
+    }
+
+    auto f = make_ready_future<shard_iterator>(iter);
+    if (!iter.shard.has_expired()) {
+        // ugh. figure out if we have end-of-shard
+        // We try to avoid the extra query by using the
+        // gossip state as a cache check - i.e. know if
+        // topology has changed and invalidated the current
+        // iterator. This is far from exact, so we could still
+        // generate duplicate records at generation changes
```
nyh, line 849: Is this a bug? a FIXME? or acceptable?

alternator/streams.cc
```diff
@@ -742,24 +829,51 @@ future<executor::request_return_type> executor::get_records(client_state& client
...
+        auto& ep = utils::fb_utilities::get_broadcast_address();
+        auto ts = cdc::get_streams_timestamp_for(ep, _ss.gossiper());
+        if (ts != db_clock::time_point{} &&  ts != iter.shard.start_time) {
+            // this is not really correct, since this should actually mean the
+            // iterator is more than 15 minutes old. But it sort of fits our
+            // use case here, namely the client managed to get and use an iterator
+            // inbetween topology change -> cdc generation change.
```
nyh, line 856: I'm a bit worried about this (ab)use of ExpiredIteratorException... A "normal" application which queries a stream every, say, 10 seconds, does not expect to ever see an ExpiredIteratorException (which suggests the application didn't update the iterator for 15 minutes), so the client code might not be prepared to deal with this case. It would be good if we can find an example stream-using application (I think you have one in Java) and check what it does in this case.

cdc/generation.hh
```diff
@@ -142,6 +142,9 @@ public:
     const std::vector<stream_id>& streams() const {
         return _streams;
     }
+    std::vector<stream_id>& streams() {
+        return _streams;
+    }
```
nyh, line 147: Why do you want a variant that returns a non-constant vector? Aren't you afraid that the user will modify it?

test/alternator/test_streams.py
```diff
@@ -359,10 +359,8 @@ def test_get_records(dynamodb, dynamodbstreams):
             while iterators:
                 iter = iterators.pop(0)
                 response = dynamodbstreams.get_records(ShardIterator=iter, Limit=1000)
-                next = response['NextShardIterator']
-
-                if next != '':
-                    next_iterators.append(next)
+                if 'NextShardIterator' in response:
+                    next_iterators.append(response['NextShardIterator'])
```
nyh, line 363: I think this change is already included in a patch I sent, maybe you will need to rebase.

elcallio:
> 
> 
> > This patch makes each stream shard a vnode (using the embedded "index" classifier
> > of cdc stream id), and uses token range query instead of id = (or id in
> > { ...values }).
> 
> So, if I understand correctly, the number of stream shards will go down by a factor of the number of cores per node, but will still be the product of num_tokens (256 by default) and the number of nodes, which is still a huge number, just less huge?
> I guess for some setups this will indeed make a big difference. In my small test setups the fact 256 usually dominates, but I agree in other setups, every factor we eliminate will help.

Yes.

> 
> > There are caveats though: To avoid querying into an old generation
> 
> I am not sure I understand what you are trying to avoid here -
> Are you trying to avoid the waste of time (i.e., performance issue) to keep querying an old closed-long-ago "stream" (in the CDC terminology) during every read for recent data, even though it will never have new data, so for a read starting at some given time, you want to be able to guess which streams still existed at that time?
> 
Since cdc generation is more or less equal to "shard" for us, we want to ensure reading from a given "old" shard produces the data we've determined belongs to it, and nothing more. The means we have to determine this are the creation and invalidation timestamps of the cdc generations, which are the start and end sequence numbers in the stream descriptions shards. 

> > , we also
> > need to make the timestamp restrictions for query adjust according to a shards
> > creation and ev. expiry time. This is a little problematic for an iterator
> > with a shards from the latest generation, if a new generation is created
> > right in the middle of operations.
> 
> I didn't understand why this is a problem - isn't it ok (and in fact, exactly the plan) that the client continues to read just the old generation's streams until they end - and the _next_ request will see also (or only) the new streams?
> 
With cdc stream id 1:1 mapped to shard ID we could just do "threshold" queries on the ID. When we change to doing token range queries we have to use exact start end (if available) end timestamps to bound the data queries so we get only records associated with the given shard. 

> > We assume this is rare, and (mis-)use the local gossip app state for cdc
> > generation timestamp to check if it seems we expired. If so, we simply
> > throw. It should be rare enough.
> 
> I didn't understand this. Maybe I'll understand as I read the code.

Almost assumes this. :-) The latest (current) cdc generation does not have and end bound (timestamp). But at any time a topology change could invalidate this generation and make the actual shard iterator used contain "stale" info. 

We use the gossip info on current cdc generation to do a quicker check iff we suddenly expired. This is not exact, as there is a delay writing to this value, but it is faster than  querying into the cdc state tables and getting the latest gen from there. An about as exact. Since reading from the current generation is assumed to be the common case we want this quick, and since we assume adding nodes is relatively rare, we simply say "expired iterator" if we are suddenly stale. 


> 
> > For old shard iterators, we simply adjust the high threshold for query.
> > This in turn means that we can include writes made to a new generation
> > in a new generation query, since we ignore confidence overlap. However,
> > this is not a problem since we similarly filter out this data from the
> > new generation query.
> 
> I didn't understand this paragraph either :(

We use the shard iterators shards end time as a high bound for data query.

alternator/streams.cc
```diff
@@ -91,16 +91,18 @@ struct rapidjson::internal::TypeHelper<ValueType, utils::UUID>
     : public from_string_helper<ValueType, utils::UUID>
 {};
 
+static db_clock::time_point as_timepoint(const utils::UUID& uuid) {
+    return db_clock::time_point{} + std::chrono::milliseconds(utils::UUID_gen::get_adjusted_timestamp(uuid));
```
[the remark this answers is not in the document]
elcallio: Sorry, sort of slunk in there from previous code that was raw uin64:s.

alternator/streams.cc
```diff
@@ -224,38 +226,53 @@ future<alternator::executor::request_return_type> alternator::executor::list_str
...
-
-    shard_id(utils::UUID tab, db_clock::time_point t, cdc::stream_id i)
-        : table(tab)
-        , time(t)
-        , id(i)
-    {}
+    static constexpr auto hex_width = sizeof(uint64_t)*8/4;
+
+    db_clock::time_point start_time;
+    db_clock::time_point end_time; // zero if open
```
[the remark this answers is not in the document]
elcallio: Well, we need to communicate the info in the request/response data, and I don't think the dynamo docs explicitly says that a superseded shard retains its ID for next describe stream query, so I think it is ok.

alternator/streams.cc
```diff
@@ -224,38 +226,53 @@ future<alternator::executor::request_return_type> alternator::executor::list_str
...
+    dht::token lo_token; // inclusive
+    dht::token hi_token; // exclusive
+
+    shard_id(db_clock::time_point start,  db_clock::time_point end, dht::token lo, dht::token hi)
+        : start_time(start)
+        , end_time(end)
+        , lo_token(lo)
+        , hi_token(hi)
+    {
+        assert(lo_token < hi_token);
```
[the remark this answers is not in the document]
elcallio: In describe stream. It verifies sorting assumptions there.

alternator/streams.cc
```diff
@@ -224,38 +226,53 @@ future<alternator::executor::request_return_type> alternator::executor::list_str
...
-    table = utils::UUID(s.substr(1, i));
-    time = db_clock::time_point(db_clock::duration(std::stoull(s.substr(i + 1, j), nullptr, 16)));
-    id = cdc::stream_id(from_hex(s.substr(j + 1)));
+    start_time = db_clock::time_point(db_clock::duration(std::stoull(s.substr(1, hex_width), nullptr, 16)));
+    end_time = db_clock::time_point(db_clock::duration(std::stoull(s.substr(1 + hex_width, hex_width), nullptr, 16)));
+    auto lo = std::stoull(s.substr(1 + 2*hex_width, hex_width), nullptr, 16);
+    auto hi = std::stoull(s.substr(1 + 3*hex_width), nullptr, 16);
+    lo_token = lo == 0 ? dht::minimum_token() : dht::token::from_int64(lo);
+    hi_token = hi == 0 ? dht::maximum_token() : dht::token::from_int64(hi);
+    assert(lo_token < hi_token);
```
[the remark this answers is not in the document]
elcallio: We verify that the UUID is a table with cdc. As for time and ranges, we do not really verify. But changing to an exception sounds right. And should the user manage to send bogus data that still passes the checks, the worst that happens is that he makes weird data queries.

alternator/streams.cc
```diff
@@ -491,34 +508,107 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
 
         std::optional<shard_id> last;
 
+        if (shard_start) {
+            i = std::find_if(i, e, [&](auto& p) {
+                return p.first == shard_start->start_time;
```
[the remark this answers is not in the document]
elcallio: Generations are sorted by time. We find the generation corresponding to the exclusive start ID.

alternator/streams.cc
elcallio: Yes. We generate two ranges that are not fully in line with what the stream ID says, (min, lo] (hi, max]. To be able to have sorted token ranges.

alternator/streams.cc
elcallio: We already did substr(1, ...), which will throw if the string is smaller that 2, so the [] there is already checked. But pro format, sure.

alternator/streams.cc
elcallio: true

alternator/streams.cc
elcallio: I think it is an acceptable trade-off between correct and efficient. As I describe in reply to  your earlier comment, we need to check that the current CDC generation wasn't replaced. Doing a query into cdc tables is to expensive (me thinks), so we use this var as a check. Neither check (expensive or cheaper) would be 100% correct, since we may not see a new cdc gen until some time after it appears.

alternator/streams.cc
elcallio: At least things like kinesis _has_ to be prepared for it, since its very async nature means it can't really be sure when it is gonna actually perform a task, such as reading from an iterator. 
Again, the above case should be rare, I "expired" seemed the best mapping to the case of an iterator actually going stale. 
The dynamo docs says that the exception is thrown when "The shard iterator has expired and can no longer be used to retrieve stream records. A shard iterator expires 15 minutes after it is retrieved using the GetShardIterator action.". 
I read that as "it will absolutely expire after 15m, but it probably could at other times as well".

cdc/generation.hh
elcallio: I want to modify it. Could maybe change it to a by-move-value or so, but in the stream description case, I have these by-value, and need to sort the ids. Could copy the vector, but that seems expensive.

test/alternator/test_streams.py
elcallio: Sure. But should merge ok.

elcallio:
Comments fixed. Rebased. Fixed some additional bugs I found.

elcallio:
ping?

alternator/streams.cc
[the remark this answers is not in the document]
nyh: I still don't understand... Do you mean that the existing client will continue to use the "open" shard id, and eventually report that it is closed, and just next time you do list the shards, you'll get a different one? If that's what you do, it's ok, of course.

alternator/streams.cc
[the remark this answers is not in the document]
nyh: Ok, so please change this assert() to an exception.

alternator/streams.cc
nyh: You're right, I forgot that substr() throws.

alternator/streams.cc
nyh: I'm sorry, I guess I'm dense, I still don't understand... If this is "not 100% correct", isn't it... A bug? Which needs to be fixed? (and I didn't yet understand what this bug means).
Is this a bug caused by a CDC bug, or a new bug we're introducing especially for Alternator user?

alternator/streams.cc
nyh: Ok, I guess we can do this for now. I'm not convinced, though, that your reading of the doc is the correct one. I can also read it as "A shard iterator may expire in 15 minutes so don't think you can hold it for hours" - but an application which thinks it only holds iterators for 10 seconds may not be prepared to handle this event because it never thinks it will run for anything close to 15 minutes. But let's ignore this issue for now.

cdc/generation.hh
nyh: I see. Ok. Let's just get someone from the CDC team (@haaawk @kbr- are the usual suspects) to check that they are ok with this change, as it changes an API not specific to Alternator.

nyh:
@elcallio I looked at the code again, and besides the small things I noted, I don't see any obvious problem with the code, but I have a bigger problem: I simply don't understand the code well enough to feel comfortable about it... There are a lot of weird new cases which I don't understand. Do you feel that you tested it well? Do you think our existing tests have good coverage of all the different cases, or we need to write more tests?

One thing our tests definitely do *not* test well is the case of shards being closed, which seems a big part of your new code... My approach to testing closed shards - by stopping the stream - doesn't work in Alternator (because the CDC table is deleted...). So we need to come up with a different approach how to cause shard-closings. Perhaps using dtest and adding an additional node to the cluster will shift data around and force shards to be closed? Does this sound a workable test plan? Do you have a simpler idea how we can force shards to be closed with just one node?

alternator/streams.cc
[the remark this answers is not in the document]
elcallio: It is fixed. Githib is showing this diff here, but the correct one on "files changed" page. No idea why.

alternator/streams.cc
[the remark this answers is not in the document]
elcallio: Pretty much. The client will use the latest set of shards until either they do describe stream again, or if the topology changes, the current set of iterators will be invalidated. When doing describestream, a new generation will be "latest", so previous "latest" gen will have changed ID:s. 

But as you say this, I should probably also give "invalidad shard" if someone tries to get an interator using a stale shardid as well. This is currently not checked.

alternator/streams.cc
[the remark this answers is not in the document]
elcallio: It is. But again, for some reason the diff above is not updated.

elcallio:
My testing so far was simply manually adding a node to force having multiple generations. But note that since I explicitly filter out shards that are older than the table we're querying on we tend to _not_ get closed shards in unit tests. 
Yes, a d-test style environment where we can add a node between creating and doing one describe_table and actually using the shards returned would be better testing. 
Right now the unit tests only verifies that the code still holds up after/without generations created.

elcallio:
An even better question is how do you test this in origin (invalidated shards/iterators). 

Two other approaches exist as well: 
1.) We could "repair" iterators iff expired, by doing a query into generations table and find the actual endpoint (which might be different from the new start point we use for invalidation detection, if two or more new generations managed to show up). 
I was doing this initially but opted against it since it made the code more complex and I felt this is a rare case. 
2.) We could go stateful. Add expiring caches and whatnot to map ID:s -> actual cdc stream ID sets, etc. Would prevent "changing" shard IDs in describe_stream, and would allow iterators to operate on IN expressions instead (avoid time bounds). Potentially more exact, but much more complicated and potentially bug prone. Plus stateful which is the antithesis of a REST service. I am not for this. 

Note that the approach in this patch set has one additional plus that might not be obvious: If we decide that the number of shards we present just is not workable for a customer, it is trivial to change describe_stream to simply divide the ID set into N more or less equal ranges to produce N shards instead. And hope cross-node queries will not be a bottleneck.

elcallio:
Actually, reading the kinesis code, I think we need to implement case 1.) - i.e. inplace-repair shard/iterator (find actual end time) iff expired, to fully fit into how the Kinesis library operates.

elcallio:
Updated: rebased (properly this time). Also removed the throwing and instead bite the bullet and do shard/iterator repair when expired. This to fit into the work pattern of Kinesis, so a "normal" worker/replication case works across topology changes.

kbr- (changes requested):

alternator/streams.cc
```diff
@@ -486,47 +480,130 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
         auto i = topologies.begin();
         auto e = topologies.end();
         auto prev = e;
-        auto threshold = db_clock::now() - confidence_interval(db);
         auto shards = rjson::empty_array();
 
         std::optional<shard_id> last;
 
+        // filter out cdc generations older than the table or now() - 24h
+        auto low_ts = std::max(as_timepoint(schema->id()), db_clock::now() - 24h);
```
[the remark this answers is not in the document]
elcallio: 24h is literally hard-coded specified in dynamodb docs. Should maybe be a constant, but will not change.

alternator/streams.cc
```diff
@@ -486,47 +480,130 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
+        while (i != e) {
+            auto j = i;
+            if (++j != e && j->first < low_ts) {
+                i = j;
+                continue;
+            }
+            // next is younger.
+            break;
+        }
```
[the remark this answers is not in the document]
elcallio: Why not? Easier to read, easier to change. And yes, this is literally lower_bound, however, since we're using map iterators (forward only), the normal implementation of lower bound (which is using binary search internally) will be much more expensive in most cases. Not that it matters much, but it is still quite unnecessary.

alternator/streams.cc
```diff
@@ -486,47 +480,130 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
 
+        // filter out cdc generations older than the table or now() - 24h
+        auto low_ts = std::max(as_timepoint(schema->id()), db_clock::now() - 24h);
+        while (i != e) {
+            auto j = i;
+            if (++j != e && j->first < low_ts) {
+                i = j;
+                continue;
+            }
+            // next is younger.
```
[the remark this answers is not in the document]
elcallio: Sure, but checked later.

alternator/streams.cc
```diff
@@ -486,47 +480,130 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
+            break;
+        }
+
+        auto first = i;
```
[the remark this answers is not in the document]
elcallio: It is the first. That we check.

alternator/streams.cc
```diff
@@ -486,47 +480,130 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
+
+        if (shard_start) {
```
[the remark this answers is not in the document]
elcallio: ì`, as conventions go, is the iterator used to extract data. It serves exactly one purpose. It is the generation being processed. We move it forwards to skip irrelevant/already processed generations, but it is still the same purpose iterator.

alternator/streams.cc
```diff
@@ -224,38 +226,59 @@ future<alternator::executor::request_return_type> alternator::executor::list_str
...
-        : table(tab)
-        , time(t)
-        , id(i)
-    {}
+    static constexpr auto hex_width = sizeof(uint64_t)*8/4;
+
+    db_clock::time_point start_time;
+    db_clock::time_point end_time; // zero if open
+    dht::token lo_token; // inclusive
+    dht::token hi_token; // exclusive
```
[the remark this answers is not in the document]
elcallio: Yes. lo_token == start of range (inclusive), hi_token == end of range (exclusive). hi_token is typically the next shards lo_token.

alternator/streams.cc
```diff
@@ -482,47 +476,130 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
-                if (shard_start && shard_start->id == id) {
-                    shard_start = std::nullopt;
-                    continue;
+                return exp.value_or(db_clock::time_point{});
+            };
+
+            auto lo = sv.streams().begin();
+            auto end = sv.streams().end();
+
+            // operating on a copy. ok to sort. this order is also preserved
```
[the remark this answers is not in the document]
elcallio: By copy here we mean the whole topology set. We query it from tables, and get it as a value (`topologies`, above). So we can freely modify the data in this set to fit our iteration (i.e. sort it).

alternator/streams.cc
```diff
@@ -482,47 +476,130 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
         for (; limit > 0 && i != e; prev = i, ++i) {
-            auto& [ts, sv] = *i;
+            auto [ts, sv] = *i;
 
             last = std::nullopt;
 
-            for (auto& id : sv.streams()) {
-                if (shard_start && shard_start->id != id) {
-                    continue;
+            // TODO: get rid of this once cdc generations
```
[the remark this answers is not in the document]
elcallio: Then why is there an "expired" field in the topology? It seems to me it would be more practical (if slightly redundant) if the generation management set/updated this table column when creating the new generation. To me, extracting it like this feels like a workaround that deserves the todo. 

The code below simply allows for the field to be set. If you want to change it to be a permanent solution, then the column should go.

alternator/streams.cc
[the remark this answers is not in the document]
kbr-: Well you do make another copy here:
```
             auto [ts, sv] = *i; 
```

alternator/streams.cc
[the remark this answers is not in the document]
elcallio: And dynamodb docs says data older than 24h should _not_ be retained/returned.

alternator/streams.cc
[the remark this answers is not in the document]
elcallio: Woops. That was a real typo. Its supposed to be ref.

alternator/streams.cc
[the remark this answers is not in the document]
kbr-: `map::lower_bound` takes a key, not an iterator, and is logarithmic in the size of the container.

alternator/streams.cc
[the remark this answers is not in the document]
kbr-: It's a good and obvious convention for maybe a 10-line algorithm where `i` is used as a single loop iterator, but this function has at least a hundred lines.

It's hard to keep track of all these variables. Some of them serve a single purpose but then aren't used for the rest of the function (`low_ts`, `first`), but you don't know that before you read the entire function.

Also, when I read code, I first want to know **what** it does, and only then **how** it does it. Here it's totally not clear what the code is supposed to achieve before you read it entirely. It could be made easier to read if parts of the code had its own names and documentation. For example, before you enter the `for` loop below, you could have something like
`auto [prev, i] = find_first_relevant_generation(topologies)`
where the meaning of "relevant" would be described in a comment (i.e. it may still contain data for this table). And suddenly I only need to read the name of the function and a 2-line comment to know what `prev` and `i` is, I don't even need to read the function's body. Then, when I want to review the function itself, I know what it's supposed to do so it's easier for me to reason about its correctness and easier for me to catch potential bugs.

alternator/streams.cc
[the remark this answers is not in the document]
kbr-: So why is end of range exclusive here, if vnodes have the end of range inclusive? Maybe it doesn't matter for correctness, but the inconsistency is weird.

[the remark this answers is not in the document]
kbr-: Unless the `hi_token` here is the vnode's end + 1, and `lo_token` is the vnode's start - 1, but then it's even more weird.

alternator/streams.cc
elcallio: Correct as in "temporarily exact", i.e. just because we pass this check does not 100% mean we are not being replaced, just that we don't see it yet. I'll update the comment.

alternator/streams.cc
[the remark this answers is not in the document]
elcallio: Ooh, newfangled c++14 methods. Forgot those were added.

alternator/streams.cc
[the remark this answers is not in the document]
elcallio: Otoh, creating a subroutine for code not run more than once is rather silly, where a better comment would suffice. I'll opt for the latter.

alternator/streams.cc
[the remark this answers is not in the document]
elcallio: Because it is based on info in stream_id:s. lo_token is the first (start) token of the range determined by the set of ID:s having index N, and hi_token is the first token of N+1. Thus the (min, max] range.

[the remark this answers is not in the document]
elcallio: However, if I misunderstood the code in generations.cc so that a stream_id:s token is an exclusive start, then the comment and the selects in get_records need adjusting, otherwise they will be subtly ineffective.

alternator/streams.cc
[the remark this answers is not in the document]
kbr-: It's used for a generation garbage-collection mechanism that exists in my head but not in the code yet. The mechanism doesn't guarantee that `it->second.expired() == std::next(it)->first` because it doesn't need it and doesn't care.

`G.expired = X` means the following: if `now() - X > max{table.cdc.ttl for all tables with cdc enabled}`, then `G` can be safely removed (because no table can possibly have data in `G` anymore). In particular, if `G.expired = X` is OK, then  `G.expired = X + 10 days` is also OK.

The mechanisms works as follows:
- someone sees at some point that generation `G` is not operating anymore, i.e. there's a newer generation and everybody is writing to that new generation, and not to `G` anymore.
- he sets `G.expired` to `now()`. Note that whoever does it doesn't need to be a dedicated process, it can be done "on the occasion".
- someone else periodically queries `time, expired` from CDC description tables. (say, once a day)
- that someone checks the condition `now() - expired > max{table.cdc.ttl for all tables with cdc enabled}` for each entry queried. If the condition is true, he removes the entry.
Note that the condition may always remain false if there exists at least one table with `ttl = 0` (no data expiration). I guess it can be made weaker by also comparing table creation times, but still there might be a case where we simply cannot ever GC a generation.

alternator/streams.cc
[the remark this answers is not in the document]
kbr-: That's why I gave you a link, you could have clicked on it :(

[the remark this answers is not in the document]
elcallio: Old eyes do not see well. Thought you linked to std::lower_bound, because you said "stl algo", not "std::map member".

alternator/streams.cc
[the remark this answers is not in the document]
kbr-: > creating a subroutine for code not run more than once is rather silly

Why? It is known that a human's short-term memory can only keep 5 to 9 items at once. So I'd argue that if a part of the code is easy to be named and separated to a logically-self-contained function (in this case it doesn't modify any state, it just searches for a thing in a set of things), it's good to do it so it won't pollute our short-term memory when we read the rest of the code.

alternator/streams.cc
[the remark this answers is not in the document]
elcallio: Unfortunately, neither of the std::algo or member functions works here, because it is neither upper or lower bound as defined, it is the last element _not greater_ (i.e. equal or lower), which either means convoluted reverse iterators + inverted comparator etc, or similar. Imho much less readable than a simple loop.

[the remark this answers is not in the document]
elcallio: But I guess we can make it with an additional `--`... if it makes you happy.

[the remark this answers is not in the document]
kbr-: I thought it's the last element which is *lower*, because the code breaks when the next element is `>= low_ts`. You're saying *lower or equal*, in which case it should break when next element is `> low_ts`, i.e. this line:
```
 if (++j != e && j->first < low_ts) {
```
should instead be
```
 if (++j != e && j->first <= low_ts) {
```

In any case:
1. if you meant *lower*, you could do `i = std::prev(topologies.lower_bound(low_ts))`
2. if you meant *lower or equal* (which the code does now), you could do  `i = std::prev(topologies.upper_bound(low_ts))`

alternator/streams.cc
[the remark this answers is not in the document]
kbr-: There may be a problem.

The token inside stream_id is not supposed to describe neither the start nor the end of any vnode, it is just defined so that all IDs with the same index fall into the same vnode in such a way that the token of each ID is mapped to a particular shard on one of the nodes.

So if you define ranges using the tokens of stream IDs by taking token of one stream ID in each group as a boundary, they will cover the entire token ring, but they won't match vnodes. You'll get something like the following picture:
image
where the green color denotes vnode boundaries, and red color denotes your boundaries that you defined using stream IDs.

The information about vnode boundaries is contained inside the internal description table, where generations are lists of tuples of the following type:
```
struct token_range_description {
    dht::token token_range_end;
    std::vector<stream_id> streams;
    uint8_t sharding_ignore_msb;
```
the `token_range_end` is the info we're looking for. It is not guaranteed that any of the streams' IDs contains `token_range_end` as its token.

But not in the user-facing table which has just a set of streams, this information is lost.

db/system_distributed_keyspace.cc
```diff
@@ -394,5 +394,27 @@ system_distributed_keyspace::cdc_get_versioned_streams(context ctx) {
     });
 }
 
+future<db_clock::time_point>
+system_distributed_keyspace::get_next_stream_timestamp(context ctx, db_clock::time_point low) {
+    // I love that you cannot order by pk.
```
haaawk, line 399: As much as I enjoy the humor. I don't think it's a good idea to leave it in the code.

db/system_distributed_keyspace.cc
```diff
@@ -394,5 +394,27 @@ system_distributed_keyspace::cdc_get_versioned_streams(context ctx) {
...
+            quorum_if_many(ctx.num_token_owners),
+            internal_distributed_timeout_config,
+            {},
+            false
+    ).then([low] (::shared_ptr<cql3::untyped_result_set> cql_result) {
+        std::optional<db_clock::time_point> high;
+        db_clock::time_point res = low;
+        for (auto& row : *cql_result) {
+            auto ts = row.get_as<db_clock::time_point>("time");
+            if (ts > low && (!high || ts < *high)) {
```
haaawk, line 411: If you define high as `db_clock::time_point high = // something real big` then the check could be simplified to `if (ts > low && ts < high)`.

In fact, the whole function could look like this:
```
db_clock::time_point res = // something real big;
for (auto& row : *cql_result) {
    auto ts = row.get_as<db_clock::time_point>("time");
    res = (ts > low && ts < res) ? ts : res;
}
return res == <real big thing from above> ? low : res;

```

cdc/generation.hh
haaawk: Actually. streams_version is used just by Alternator. I would rather move the whole class out of the CDC namespace and encapsulate it in Alternator. That way there won't be any confusion.

alternator/streams.cc
```diff
@@ -479,50 +476,135 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
         auto shards = rjson::empty_array();
 
         std::optional<shard_id> last;
 
+        // i is now at the youngest generation we include. make a mark of it.
+        auto first = i;
+
+        // if we're a paged query, skip to the generation where we left of.
+        if (shard_start) {
+            i = std::find_if(i, e, [&](auto& p) {
```
haaawk, line 500: generic lambda will probably make @avikivity unhappy

alternator/streams.cc
```diff
@@ -479,50 +476,135 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
     // or on expired...
     // TODO: maybe add secondary index to topology table to enable this?
     return _sdks.cdc_get_versioned_streams({ tm.count_normal_token_owners() }).then([this, &db, schema, shard_start, limit, ret = std::move(ret), stream_desc = std::move(stream_desc)](std::map<db_clock::time_point, cdc::streams_version> topologies) mutable {
-        auto i = topologies.begin();
+
+        // filter out cdc generations older than the table or now() - dynamo_streams_max_window (24h)
```
haaawk, line 480: Why is this needed to group streams by vnodes?

haaawk:
I have to say that this is a surprisingly large change for something which seems quite simple (just group the streams).
The size makes it really hard to understand and review.
I know I'm asking for additional work, and I'm sorry for that, but do you think it would be possible to split the work into smaller commits with more localized commit messages, @elcallio?

alternator/streams.cc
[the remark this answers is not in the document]
haaawk: I think we need this in the comment. This is the knowledge that's required to understand this code. At least for me it was.

alternator/streams.cc
```diff
@@ -220,38 +222,59 @@ future<alternator::executor::request_return_type> alternator::executor::list_str
...
-
-    shard_id(utils::UUID tab, db_clock::time_point t, cdc::stream_id i)
-        : table(tab)
-        , time(t)
-        , id(i)
-    {}
+    static constexpr auto hex_width = sizeof(uint64_t)*8/4;
+
+    db_clock::time_point start_time;
+    db_clock::time_point end_time; // zero if open
```
haaawk, line 228: What does this start_time and end_time represent logically? Is start_time the time the CDC generation that this shard belongs to was opened? Is end_time the time this CDC generation was closed?

alternator/streams.cc
```diff
@@ -220,38 +222,59 @@ future<alternator::executor::request_return_type> alternator::executor::list_str
...
-    time = db_clock::time_point(db_clock::duration(std::stoull(s.substr(i + 1, j), nullptr, 16)));
-    id = cdc::stream_id(from_hex(s.substr(j + 1)));
+    start_time = db_clock::time_point(db_clock::duration(std::stoull(s.substr(1, hex_width), nullptr, 16)));
+    end_time = db_clock::time_point(db_clock::duration(std::stoull(s.substr(1 + hex_width, hex_width), nullptr, 16)));
+    auto lo = std::stoull(s.substr(1 + 2*hex_width, hex_width), nullptr, 16);
+    auto hi = std::stoull(s.substr(1 + 3*hex_width), nullptr, 16);
+    lo_token = lo == 0 ? dht::minimum_token() : dht::token::from_int64(lo);
+    hi_token = hi == 0 ? dht::maximum_token() : dht::token::from_int64(hi);
+
+    if (end_time != db_clock::time_point{} && end_time < start_time) {
```
haaawk, line 272: Super nit: s/</<=/

alternator/streams.cc
haaawk: Ok. Nevermind. I think I got it.

alternator/streams.cc
```diff
@@ -479,50 +476,135 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
         auto prev = e;
-        auto threshold = db_clock::now() - confidence_interval(db);
         auto shards = rjson::empty_array();
 
         std::optional<shard_id> last;
 
+        // i is now at the youngest generation we include. make a mark of it.
+        auto first = i;
+
+        // if we're a paged query, skip to the generation where we left of.
```
haaawk, line 498: How is shard_id related to the state of paging?

alternator/streams.cc
```diff
@@ -479,50 +476,135 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
+            auto i1 = id1.index();
+            auto i2 = id2.index();
+            return i1 < i2 || (i1 == i2 && id1.token() < id2.token());
+        };
+
+        // need a prev even if we are skipping stuff
+        if (i != first) {
+            auto j = i;
+            --j;
+            std::sort(j->second.streams().begin(), j->second.streams().end(), cmp);
```
haaawk, line 516: Why are we sorting only when i != first? Don't we need to sort prev when i == first?

alternator/streams.cc
```diff
@@ -479,50 +476,135 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
+        auto low_ts = std::max(as_timepoint(schema->id()), db_clock::now() - dynamo_streams_max_window);
+
+        auto i = topologies.lower_bound(low_ts);
+        // need first gen _intersecting_ the timestamp.
+        if (i != topologies.begin()) {
+            --i;
+        }
+
         auto e = topologies.end();
         auto prev = e;
```
haaawk, line 490: Is this still correct? shouldn't prev be std::prev(i) when i != topologies.begin()?

alternator/streams.cc
```diff
@@ -479,50 +476,135 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
+            };
+
+            auto lo = sv.streams().begin();
+            auto end = sv.streams().end();
+
+            // operating on a copy. ok to sort. this order is also preserved 
+            // for the "next" check below.
+            std::sort(lo, end, cmp);
+
+            assert(std::is_sorted(std::find_if(lo, end, [](auto& id) { return id.index() == 1; }), end, [](auto& id1, auto& id2) {
```
haaawk, line 545: Probably best to avoid generic lambdas.

alternator/streams.cc
elcallio: Yes. Though "closed" is defined as "next generation was opened".

alternator/streams.cc
elcallio: No, I purposely allow for end to be equal to start. While our cdc generation will not do this, it is a valid state for the ID here, so it is acceptable. It will not yield data, but acceptable.

alternator/streams.cc
elcallio: It is technically not, but it filters out useless data, so helps further reduce shard bloat. Could technically have been a separate patch, yes.

alternator/streams.cc
elcallio: No, because i - 1 a.) could be begin() - 1, and b.) has not been sorted/processed, so not a valid .

alternator/streams.cc
elcallio: Dynamo paging. We only return 100 shards each call, and must resume from last returned shard + 1 on subsequent calls, based on the `ExclusiveStartShardID` - hence "paging".

alternator/streams.cc
elcallio: In this tiny case, he can deal with it.

alternator/streams.cc
elcallio: If prev is set because we did a full iteration, it is already sorted (we just processed it). It is only when we are skipping and need to "recreate" prev that we need to pre-emptively sort.  If we are on the first lap, prev is == end() (no prev), so no sorting.

alternator/streams.cc
elcallio: Again, based on the size of the lambda, me thinks it is warranted.

cdc/generation.hh
elcallio: It still needs code in system_distributed. Moving it to alternator expands header entanglement.

db/system_distributed_keyspace.cc
elcallio: This is not humour. I am dead serious.

db/system_distributed_keyspace.cc
elcallio: timepoint::max(), sure.

alternator/streams.cc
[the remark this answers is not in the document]
elcallio: We are not really interested in the actual vnode range as such, we are interested in a range such that all the ID:s for this vnode are inside it. Using lowest == inclusive start, and next ID start as exclusive high should fulfill this criteria. 
It could be a problem though if the query will still consider querying _into_ another vnode (though no ID:s should match), i.e. the lowest "start" of ID:s for the "next" vnode is "higher" than vnode range start. 
But ID token is defined as "first token for (scylla) shard" of a given vnode. Logically it would seem that you cannot have a token lower than this that hits the same shard (and by extension vnode).

alternator/streams.cc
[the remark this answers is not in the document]
elcallio: Fair enough, will kill it and always use next:s timestamp.

alternator/streams.cc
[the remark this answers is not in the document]
kbr-: > It could be a problem though if the query will still consider querying into another vnode

I'd assume that range scans don't know whether or not there will exist a partition with data in the other node, so it will have to look into it. Btw. why the decision to use partition scans and not IN queries?

> But ID token is defined as "first token for (scylla) shard" of a given vnode. Logically it would seem that you cannot have a token lower than this that hits the same shard (and by extension vnode).

The vnode may start inside shard 1 say, to the smallest stream ID, which is defined to be the "first token for shard 0", will not be on the vnode boundary.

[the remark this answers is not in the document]
elcallio: Your code in generation splits the ring into vnodes by token boundaries, then for each such boundary range find all shard boundaries and generate an id for it. While it is not the same thing as vnode boundary, I fail to see how we could have data at  ID ranges outside it for a given vnode. So I still think it should be valid ranges to query.

As for scan vs IN, the latter would require an iterator to know all the ID:s contained in a vnode, and we need to encode this in (hopefully shorter than) 4096 chars. Otherwise we need statefulness, or secondary queries. Whereas a token range is easily encoded in 128bits.

elcallio:
Added more comments, stdlib usage, versioned_streams now pure pod instead to avoid false encapsulation. Etc.

alternator/streams.cc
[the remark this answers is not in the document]
kbr-: I meant a situation like the following: say vnode is `(100, 200]`, and shards have the following ranges:
`(90, 130]` - shard 1
`(130, 170]` - shard 0
`(170, 210]` - shard 1
(I'm not sure which boundaries are closed and which are opened, but let's assume it's the same as for vnodes)

Then, if I understand correctly:
- there will be two stream IDs in this vnode with tokens `131` (for shard 0) and `171` (for shard 1)
- your code will perform a range query `(170, 212)` or something like that (because the first stream ID in the following vnode will have token 211), so we perform a cross-vnode scan.

another approach would be to query `(130, 172)` (or `[131, 171]` or whatever) and simply give up trying to cover the entire token ring with our ranges, because we don't really need to if we just want to cover all stream IDs. Then there won't be any cross-vnode scans.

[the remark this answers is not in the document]
elcallio: Right, that was what I was a bit worried about - we will potentially query the "wrong" vnode and get zilch. . But as you say, we can eliminate the cross-vnode scan by simply changing the shard range to  be lowest-id-start-inclusive to highest-id-start-inclusive. This will include all ID:s generated for the vnode, and is actually even easier to build.

kbr-:
@elcallio I have another issue. If we perform range scans, we'll be getting results from multiple generations. I understand that you want to use `start_time` and `end_time` to filter out these results. However, there is a non-zero period of time between introducing a new generation and updating `end_time` in the "streams iterator" (or whatever it's called). Thus, if we have an iterator to the latest generation and a new generation is introduced, we may get some stuff from the new generation. How do we deal with that problem?

elcallio:
We don't. We simply acknowledge that if you read data too close to the generation switch-over, you can get duplicate records. But assuming the confidence window is big enough, we should in general see the switch-over before doing a query to this time, so assuming proper config this should be a rare occurrence.

kbr-:
Okay. Also note that there is some time (`2 x ring_delay`) between introducing the next generation and nodes actually writing to it. If we're careful, we should always manage to update `end_time` during this period.

Still, I wonder if such range scans won't be a problem for performance: the replica will have to consider a lot of partitions that are not relevant for the query (from other generations). We do at least try to filter out sstables based on clustering key ranges (https://github.com/scylladb/scylla/issues/4042, https://github.com/scylladb/scylla/issues/4442) but I don't know if it works for range scans or only for single-partition queries (the two cases use different code paths). If it doesn't, that would mean we'll be opening a **huge** number of sstables and prefetch a prefix of each into memory, which would cause OOMs as soon as there are too many generations.

elcallio:
I am very much hoping the clustering filtering works, otherwise it might indeed be troublesome. But that should be for an IN query as well, no?
We are still assuming that generation changes are relatively rare compared to reads. And given that there is quite a large overhead in doing a single-id query (esp. with the double layers caused by alternator, and the actual query most likely not hitting same node as coordinator), trying to eliminate some shards should hopefully be worthwhile. 
There is still one really important part missing from the alternator query also, and that is query limit. Right now we do not use it, because we have no way of telling if we have read all the associated rows for a single CDC event, nor how many CDC rows we should expect (given the dynamo query limit). So there are more than one hurdle towards efficiency here.

kbr-:
> But that should be for an IN query as well, no?

IN queries use single-partition reads underneath, and I'm sure these do sstable filtering based on clustering key.

elcallio:
Updated with more proper id token handling (use first/last as inclusive bounds), and handle wraparound fully correct now (wraparound split range).

alternator/streams.cc
```diff
@@ -479,50 +494,125 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
+            auto end = sv.streams.end();
+
+            // first sort by index.
+            std::stable_sort(lo, end, [](auto& id1, auto& id2) {
+                return id1.index() < id2.index();
+            });
+
+            if (shard_start) {
+                // find next shard position
+                // first id set is the wraparound range. skip it here.
```
kbr-, line 551: Isn't that an implementation detail? Are we sure this will be guaranteed forever? If yes, then probably a unit test should check it somehow.

alternator/streams.cc
```diff
@@ -479,50 +494,125 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
-                    auto pid = std::upper_bound(pids.begin(), pids.end(), token, [](const dht::token& t, const cdc::stream_id& id) {
+                    auto& pids = prev->second.streams;
+                    // maybe re-sort by token, so we can find the previous shard bounds
+                    // if prev is sorted by index, we can just rotate it slightly to make it sorted by
+                    // token.
+                    if (pids.front().token() > pids.back().token()) {
+                        auto nf = std::find_if(pids.begin(), pids.end(), [t = pids.back().token()](const cdc::stream_id& id) {
+                            return id.token() < t;
+                        });
+                        std::rotate(pids.begin(), nf, pids.end());
```
kbr-, line 581: Maybe using a single sort which compares by index first and token second in the first place would be simpler than all of this.

But wouldn't sorting by just the token automatically result in sorting by index as well?

alternator/streams.cc
elcallio: It is using inside knowledge, but otoh it is in the same source code base as said knowledge, so...

alternator/streams.cc
elcallio: Then we get zero == range that covers the rest of the id:s (being the wraparound). Could add code that "finds" the wraparound, but it seems rather like false abstraction to me. And a pointless search of a rather large array once the cluster becomes larger.
I think we can just define it as "index 0 == wraparound". Easier.

elcallio:
ping?

kbr- (approved):
I don't see any more issues with the change, so LGTM

elcallio:
I suddenly do. I just realized this does not work. The problem is the record request limit (default 100) and sequence numbers. 
1.) While not explicitly stated, since the API exists we must assume that Dynamo expects each records sequence number to be monotonically growing. This will not always be true in a result based on multiple cdc streams, since cluster ordering will only be per PK. 
2.) Iterators - the code also assumes monotonic timestamp uuids. We have no way of specifying a "next" iterator when selecting from multiple ID:s. 
These problems of course only exists because we might have to limit the number of returned records in response. 2.) could be dealt with by serializing a paging state into the iterator (how large would that make it?). But 1.) is more difficult.

test/alternator/test_streams.py
```diff
@@ -845,9 +848,11 @@ def test_streams_no_records(test_table_ss_keys_only, dynamodbstreams):
     table, arn = test_table_ss_keys_only
     # Get just one shard - any shard - and its LATEST iterator. Because it's
     # LATEST, there will be no data to read from this iterator.
-    shard_id = dynamodbstreams.describe_stream(StreamArn=arn, Limit=1)['StreamDescription']['Shards'][0]['ShardId']
+    shard = dynamodbstreams.describe_stream(StreamArn=arn, Limit=1)['StreamDescription']['Shards'][0]
+    shard_id = shard['ShardId']
     iter = dynamodbstreams.get_shard_iterator(StreamArn=arn, ShardId=shard_id, ShardIteratorType='LATEST')['ShardIterator']
     response = dynamodbstreams.get_records(ShardIterator=iter)
+    assert 'NextShardIterator' in response or 'EndingSequenceNumber' in shard['SequenceNumberRange']
```
nyh, line 855: This is a weird thing to test here. Nothing this test did closed any shards, in DynamoDB the first case will always be true (there is a NextShardIterator), but in Alternator it depends on whether any closed shards exist - i.e., I expect a 1-node setup to always have the first case, and only a multi-node setup maybe reach the second case. Did you check a multi-node setup?
Moreover, the EndingSequenceNumber test is on the shard, not the iterator. What happens if at the time we got the shard iterator, there it wasn't closed yet, but now, a few seconds later (potentially), it did get closed? In that case, this test will fail - because there is no NextShardIterator, and no EndingSequenceNumber.
I'll try to think how to add in test_streams_closed_read a test for EndingSequenceNumber. It won't help now (this test fails on Alternator) but may help in the future.
In general I  prefer tests which ensure that the situation they are trying to test actually happens. E.g., test_streams_closed_read tried to ensure a closed shard happens - but doesn't work on Alternator.

nyh (changes requested):
I left a few small comments and only one place where I think there is an error (you dropped an ios flags save which I think you shouldn't have).
Other than that, I still have a problem that I don't really understand this patch, but given that Kamil did, and I hope you tested it well with various generation-changing events, I tend to accept this patch, and probably will.
There are a few unresolved discussions in this PR, would be nice if you could make sure there is nothing left to be done. Thanks!

nyh:
> I suddenly do. I just realized this does not work. The problem is the record request limit (default 100) and sequence numbers.
> 1.) While not explicitly stated, since the API exists we must assume that Dynamo expects each records sequence number to be monotonically growing.

I was thinking about this earlier today. I'm not even sure this is true even today, because we take the timeuuid with its bat-shit-crazy ordering rules and convert it into a number, and that number is probably not monotonic. I don't know if it actually matters, although it would be nice if it was monotonic :-(

> This will not always be true in a result based on multiple cdc streams, since cluster ordering will only be per PK.

But doesn't the "sequence number" also include the generation number? We obviously need it there, so that AFTER_SEQUENCE_NUMBER works as expected.

> 2.) Iterators - the code also assumes monotonic timestamp uuids. We have no way of specifying a "next" iterator when selecting from multiple ID:s.

I didn't understand this one.

> These problems of course only exists because we might have to limit the number of returned records in response.

It's not *just* that. The DynamoDB API also allows you to read 200 items, remember the 34th item's "sequence number", and then ask to read again starting at this 34th item and read 166 items again.

> 2.) could be dealt with by serializing a paging state into the iterator (how large would that make it?). But 1.) is more difficult.

elcallio:
> 
> 
> > I suddenly do. I just realized this does not work. The problem is the record request limit (default 100) and sequence numbers.
> > 1.) While not explicitly stated, since the API exists we must assume that Dynamo expects each records sequence number to be monotonically growing.
> 
> I was thinking about this earlier today. I'm not even sure this is true even today, because we take the timeuuid with its bat-shit-crazy ordering rules and convert it into a number, and that number is probably not monotonic. I don't know if it actually matters, although it would be nice if it was monotonic :-(
> 
I don't think it need to be monotonic as a decimal number, but once used as for example a parameter to get_shard_iterator, it needs to be monotonic in that context. 
And while it is not 100% explicit in the dynamo spec, it is at least implied that inside a shard, records are returned in creation order. 
If we define a shard as multiple cdc$streams, this might not be true, depending on how we page and if we merge-sort or not across these streams. Today we do not. 

> > This will not always be true in a result based on multiple cdc streams, since cluster ordering will only be per PK.
> 
> But doesn't the "sequence number" also include the generation number? We obviously need it there, so that AFTER_SEQUENCE_NUMBER works as expected.
> 
Sequence only need to be unique within the context of a shard. You cannot use a sequence number as "after seq no" to another shard and expect consistent results (if any). Problem again is if you have multiple cdc streams, and don't ensure all data is actually ordered by timestamp/seq no.

We could do query across ID:s (and use query limit x N id:s), then merge-sort records. But that would explode both query size (again, we don't yet use limits - which is very bad), as well as potentially in-memory usage. 

> > 2.) Iterators - the code also assumes monotonic timestamp uuids. We have no way of specifying a "next" iterator when selecting from multiple ID:s.
> 
> I didn't understand this one.
> 
Basically, the iterator need two more optional fields, "PK position + in-partition CK", if paging across multiple streams is to work, since the timeuuid threshold is only valid inside a single stream. Over many, you need to restrict second call by PK + in-pk CK or PK in <rest-of-range>. Once you reach end-of-pk-range, you can go back to just clustering order. 

> > These problems of course only exists because we might have to limit the number of returned records in response.
> 
> It's not _just_ that. The DynamoDB API also allows you to read 200 items, remember the 34th item's "sequence number", and then ask to read again starting at this 34th item and read 166 items again.

Yes.
> 
> > 2.) could be dealt with by serializing a paging state into the iterator (how large would that make it?). But 1.) is more difficult.

elcallio:
Closing this PR. The still valid parts will be moved to a different PR.

PR #7252, opened 2020-09-17
docs/docker-hub.md: add quickstart section with --smp 1

sitano:
Also provide formula to calculate proper value for aio-max-nr.

### DIFF
--- a/docs/docker-hub.md
+++ b/docs/docker-hub.md
@@ -7,6 +7,35 @@ ScyllaDB is released under the GNU Affero General Public License version 3 and t
 
 ![logo](http://www.scylladb.com/wp-content/uploads/mascot_medium.png)
 
+# Quick start
+
+To startup a Scylla single-node cluster in developer mode execute:
+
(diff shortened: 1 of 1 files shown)


sitano:
However, the formula is not correct because it does not work even for 8 cores that must require 128*8*8 = 8192 free aio-nr items.

bentsi:
@avikivity Scylla [page](https://hub.docker.com/r/scylladb/scylla) on Docker hub doesn't reflect the changes in this PR.
How is the page updated? Manually or we have some automatic process to update it?

CC: @lauranovich 
image

avikivity:
It's not updated at all.

bentsi:
@guy9 FYI: Docker examples in the Scylla University won't run out of the box, need somehow to notify users that the `aio-max-nr` should be increased to the recommended value.

guy9:
@bentsi can you elaborate on this? Is this on Fedora? A specific version?

PR #7268, opened 2020-09-22
build: use consistent version-release strings across submodules

avikivity:
Instead of relying on SCYLLA-VERSION-GEN to be consistently
updated in each submodule, propagate the top-level product-version-release
to all submodules. This reduces the churn required for each release,
and makes the release strings consistent (previously, the git hash in each
was different).

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1754,7 +1754,7 @@ with open(buildfile_tmp, 'w') as f:
         build dist-server: phony dist-server-tar dist-server-rpm dist-server-deb
 
         rule build-submodule-reloc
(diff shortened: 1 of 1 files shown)


avikivity:
Marked as draft since it depends on the submodules, but otherwise complete. Can be merged once the submodules are updated (in rebase mode only).

penberg:
@avikivity I think all the prerequisites are merged now, no?

avikivity:
Yes.

avikivity:
@penberg ping

PR #7269, opened 2020-09-22
build: prepare for clang

avikivity:
This series prepares the build system for clang support. It deals with the different sets of warnings accepted by clang and gcc, and with detecting clang 10 as a supported compiler.

It's still not possible to build with clang after this, but we're another step closer.


avikivity:
Dequeued, breaks release build.

avikivity:
Updated: patch 2 did not add `-Wno-error=stack-usage=` correctly. Our warning detection doesn't work well with negative warnings, and -Wstack-usage= is special because of the trailing =.

Simplified by adding the -Wno-error flag where we add -Wstack-usage.

PR #7275, opened 2020-09-24
Gossip echo message improvement

asias:
This series improves gossip echo message handling in a loaded cluster. 

Refs: #7197

### DIFF
--- a/message/messaging_service.cc
+++ b/message/messaging_service.cc
@@ -1008,7 +1008,7 @@ future<> messaging_service::unregister_gossip_echo() {
     return unregister_handler(netw::messaging_verb::GOSSIP_ECHO);
 }
 future<> messaging_service::send_gossip_echo(msg_addr id) {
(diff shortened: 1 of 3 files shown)


message/messaging_service.cc
```diff
@@ -1008,7 +1008,7 @@ future<> messaging_service::unregister_gossip_echo() {
     return unregister_handler(netw::messaging_verb::GOSSIP_ECHO);
 }
 future<> messaging_service::send_gossip_echo(msg_addr id) {
-    return send_message_timeout<void>(this, messaging_verb::GOSSIP_ECHO, std::move(id), 3000ms);
+    return send_message_timeout<void>(this, messaging_verb::GOSSIP_ECHO, std::move(id), 15000ms);
```
vladzcloudius, line 1011: The best would have been to have it user configurable too, @asias

PR #7281, opened 2020-09-25
gossiper: Use dedicated gossip scheduling group

### LINKED ISSUE #7154: Set gossiper scheduling class explicitly
*Installation details*
Scylla version: 58104d17e08a94b5c339da370cd2bf9abb276209

*Description*
We are not initializing the scheduling class of the gossiper timer explicitly. In this case the scheduling group ID is default-initialized to 0 which happens to come out the same as the ID of the "maintenance" scheduling group *because* we initialize this class the first in the main.cc: https://github.com/scylladb/scylla/blob/master/main.cc#L587 and https://github.com/scylladb/scylla/blob/master/gms/gossiper.cc#L140 - this is very fragile and we'd rather initialize the gossiper scheduling class explicitly in order to avoid surprises if/when scheduling groups numbers change.

asias:
Gossip currently runs inside the default (main) scheduling group. It is
fine to run inside default scheduling group. From time to time, we see
many tasks in main scheduling group and we suspect gossip. It is best
we can move gossip to a dedicated scheduling group, so that we can catch
bugs that leak tasks to main group more easily.

After this patch, we can check:

scylla_scheduler_time_spent_on_task_quota_violations_ms{group="gossip",shard="0"}

Fixes #7154

### DIFF
--- a/database.hh
+++ b/database.hh
@@ -1217,6 +1217,7 @@ struct database_config {
     seastar::scheduling_group memory_compaction_scheduling_group;
     seastar::scheduling_group statement_scheduling_group;
     seastar::scheduling_group streaming_scheduling_group;
+    seastar::scheduling_group gossip_scheduling_group;
     size_t available_memory;
 };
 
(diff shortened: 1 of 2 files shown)


avikivity:
What about messaging_service? It needs adjustments to run the message handlers in the gossip group.

What about nodetool enablegossip/disablegossip?

asias:
> What about messaging_service? It needs adjustments to run the message handlers in the gossip group.
> 
> What about nodetool enablegossip/disablegossip?

When gossip is disabled and enabled again, the verbs are registered. Here 
https://github.com/scylladb/scylla/pull/7282#issuecomment-700362542, I think we will register the verb with the correct scheduling group for gossip verbs. 

btw, I am closing this pr because we have a duplicated one https://github.com/scylladb/scylla/pull/7282

PR #7282, opened 2020-09-25
gossiper: Use dedicated gossip scheduling group

### LINKED ISSUE #7154: Set gossiper scheduling class explicitly

asias:
Gossip currently runs inside the default (main) scheduling group. It is
fine to run inside default scheduling group. From time to time, we see
many tasks in main scheduling group and we suspect gossip. It is best
we can move gossip to a dedicated scheduling group, so that we can catch
bugs that leak tasks to main group more easily.

After this patch, we can check:

scylla_scheduler_time_spent_on_task_quota_violations_ms{group="gossip",shard="0"}

Fixes #7154

### DIFF
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -123,12 +123,13 @@ public:
     void on_restart(inet_address, endpoint_state) override {}
 };
 
-gossiper::gossiper(abort_source& as, feature_service& features, const locator::token_metadata& tokens, netw::messaging_service& ms, db::config& cfg)
+gossiper::gossiper(abort_source& as, feature_service& features, const locator::token_metadata& tokens, netw::messaging_service& ms, db::config& cfg, gossip_config gcfg)
         : _abort_source(as)
         , _feature_service(features)
         , _token_metadata(tokens)
         , _messaging(ms)
         , _cfg(cfg)
+        , _gcfg(gcfg)
         , _fd(cfg.phi_convict_threshold(),
                 std::chrono::milliseconds(cfg.fd_initial_value_ms()),
                 std::chrono::milliseconds(cfg.fd_max_interval_ms())) {
@@ -137,7 +138,7 @@ gossiper::gossiper(abort_source& as, feature_service& features, const locator::t
         return;
     }
 
-    _scheduled_gossip_task.set_callback([this] { run(); });
+    _scheduled_gossip_task.set_callback(_gcfg.gossip_scheduling_group, [this] { run(); });
     // half of QUARATINE_DELAY, to ensure _just_removed_endpoints has enough leeway to prevent re-gossip
     fat_client_timeout = quarantine_delay() / 2;
     /* register with the Failure Detector for receiving Failure detector events */
--- a/database.hh
+++ b/database.hh
@@ -1218,6 +1218,7 @@ struct database_config {
     seastar::scheduling_group memory_compaction_scheduling_group;
     seastar::scheduling_group statement_scheduling_group;
     seastar::scheduling_group streaming_scheduling_group;
+    seastar::scheduling_group gossip_scheduling_group;
     size_t available_memory;
 };
 
--- a/gms/gossiper.hh
+++ b/gms/gossiper.hh
@@ -64,6 +64,7 @@
 #include <seastar/core/condition-variable.hh>
 #include <seastar/core/metrics_registration.hh>
 #include <seastar/core/abort_source.hh>
+#include <seastar/core/scheduling.hh>
 
 namespace db {
 class config;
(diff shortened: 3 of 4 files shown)


denesb:
We need to check how this interacts with scheduling-groups based read semaphore selection.

denesb:
Does gossip do any queries?

denesb:
We use the system semamphore as the fallback for queries running in an unknown secheduling group. So this should be fine.

denesb:
Note that messaging service will drop the gossip scheduling group when using the gossip verbs, the remote will be using the default group. We implemented scheduling group conservation only for statement verbs.

asias:
> Does gossip do any queries?

Gossip itself does not do any queries, but the gossip listeners (on_change), writes to system table. Is such query a concern?

asias:
> Note that messaging service will drop the gossip scheduling group when using the gossip verbs, the remote will be using the default group. We implemented scheduling group conservation only for statement verbs.

I did not follow the recent changes about scheduling. 

Looking at the code, 

Gossip verbs in do_get_rpc_client_idx are having index of zero.

In messaging_service::initial_scheduling_info, index zero are using scheduling group _scheduling_config.gossip which we is passed from main.cc where we set it scfg.gossip.

 566     auto sched_infos = std::vector<scheduling_info_for_connection_index>({
 567         { _scheduling_config.gossip, "gossip" },
 568         { _scheduling_config.streaming, "streaming", },
 569     }); 

And when the verb is registered:

 272 // Register a handler (a callback lambda) for verb
 273 template <typename Func>
 274 void register_handler(messaging_service* ms, messaging_verb verb, Func&& func) {
 275     ms->rpc()->register_handler(verb, ms->scheduling_group_for_verb(verb), std::move(func));
 276 }

it is using the scheduling group for the verb which is gossip for gossip verbs.

So, what is the problem here?

denesb:
Looks like we should be fine then.

denesb:
> > Does gossip do any queries?
> 
> Gossip itself does not do any queries, but the gossip listeners (on_change), writes to system table. Is such query a concern?

No, I just wasn't sure on whether it will use the right semaphore, but it looks like it will fall back to the one for `default`, so its fine. What I wanted to ensure is that such queries will *not* use the user semaphore.

asias:
> > > Does gossip do any queries?
> > 
> > 
> > Gossip itself does not do any queries, but the gossip listeners (on_change), writes to system table. Is such query a concern?
> 
> No, I just wasn't sure on whether it will use the right semaphore, but it looks like it will fall back to the one for `default`, so its fine. What I wanted to ensure is that such queries will _not_ use the user semaphore.

OK. I see.

avikivity:
> Does gossip do any queries?

Yes, it interacts with the system.peers table.

avikivity:
@asias we need to put the main gossip worker (is it a thread?) into this group too

asias:
> @asias we need to put the main gossip worker (is it a thread?) into this group too

The main loop is driven by a timer.

```
gossiper::gossiper() {
_scheduled_gossip_task.set_callback([this] { run(); });
}

gossiper::run() {
do something
_scheduled_gossip_task.arm(INTERVAL);
}

```

Is the scheduling group class decided by the caller's scheduling group who creates gossiper objects? Or who calls _scheduled_gossip_task.arm()?

avikivity:
By the scheduling group of set_callback()'s caller.

asias:
> By the scheduling group of set_callback()'s caller.

OK. Then we need to call with_scheduling_group inside run().

asias:
Changes in b19950a2f56000902aba804121e4f08e681a9757:
- Run gossip loop callback inside gossip scheduling group

gms/gossiper.cc
```diff
@@ -762,9 +763,10 @@ future<gossiper::endpoint_permit> gossiper::lock_endpoint(inet_address ep) {
 // - failure_detector
 // - on_remove callbacks, e.g, storage_service -> access token_metadata
 void gossiper::run() {
-   // Run it in the background.
-  (void)seastar::with_semaphore(_callback_running, 1, [this] {
-    return seastar::async([this, g = this->shared_from_this()] {
```
avikivity, line 767: Why not do it in the timer itself? set_callback() accepts a scheduling group.

asias: OK. I will get back to this when I finish other higher priority issues.

asias: Done. Sent v2.

asias:
Changes in 565af97e56acc0d15cdef5115ce4f0ef713409ce:
- Set scheduling group when setting the timer callback

PR #7292, opened 2020-09-28
[RFC] tracing: Fix slow batches of prepared statements

### LINKED ISSUE #5843: "No query parameter set for a session requesting a slow_query_log record" error while tracing slow batch queries
*Installation details*
Scylla version (or git commit hash): 6c7aa1823826817b38268871a0633a818cd98495
Cluster size:

`trace_keyspace_helper::make_slow_query_mutation_data()` assumes there is `query` parameter, but for batch there are `query[0], query[1]...`. See `trace_keyspace_helper::make_slow_query_mutation_data()`

jul-stas:
`trace_keyspace_helper::make_slow_query_mutation_data` expected a "query" key in its parameters, which does not appear in case of batches of prepared statements. This is example of failing `record.parameters`:
```
...{"query[0]" : "INSERT INTO ks.tbl (pk, i) values (?, ?);"},
{"query[1]" : "INSERT INTO ks.tbl (pk, i) values (?, ?);"}...
```

In such case Scylla records no trace and says:
```
ERROR 2020-09-28 10:09:36,696 [shard 3] trace_keyspace_helper - No "query" parameter set for a session requesting a slow_query_log record
```

The fix is to concatenate all the batched queries into a single string and trace this string. It's not perfect, because the traced mega-query does not have the structure of a batch. But it's something. So the error is gone and the trace would now look like this:
```
cqlsh> select * from system_traces.node_slow_log;
...
01915bd7-016a-11eb-9043-000000000002 | 127.0.0.1 |     3 |                                                                           \n    INSERT INTO ks.tbl (pk, i) values (?, ?); \n    INSERT INTO ks.tbl (pk, i) values (?, ?);  | 2020-09-28 09:07:16.939000+0000 |      336 |                                                                                                              {'batch_endpoints': '/127.0.0.1', 'consistency_level': 'LOCAL_ONE', 'param[0][0]': '1', 'param[0][1]': '2', 'param[1][0]': '3', 'param[1][1]': '4', 'query[0]': '\n    INSERT INTO ks.tbl (pk, i) values (?, ?);', 'query[1]': '\n    INSERT INTO ks.tbl (pk, i) values (?, ?);', 'serial_consistency_level': 'SERIAL'} | 019141b0-016a-11eb-9043-000000000002 | 127.0.0.1 |                   {'ks.tbl'} | anonymous
...
```

Fixes #5843

### DIFF
--- a/tracing/trace_keyspace_helper.cc
+++ b/tracing/trace_keyspace_helper.cc
@@ -289,10 +289,22 @@ cql3::query_options trace_keyspace_helper::make_slow_query_mutation_data(const o
     const session_record& record = session_records.session_rec;
     auto millis_since_epoch = std::chrono::duration_cast<std::chrono::milliseconds>(record.started_at.time_since_epoch()).count();
 
-    // query command is stored on a parameters map with a 'query' key
-    auto query_str_it = record.parameters.find("query");
-    if (query_str_it == record.parameters.end()) {
-        throw std::logic_error("No \"query\" parameter set for a session requesting a slow_query_log record");
+    sstring query_str;
+    // query command is stored on a parameters map with a 'query' key or 'query[N]' keys
+    if (auto query_str_it = record.parameters.find("query"); query_str_it == record.parameters.end()) {
+        for (int i = 0; /* nothing */; ++i) {
+            // Extract the values under "query[0]", "query[1]", etc. in case of batched prepared statements
+            auto ith_query_iter = record.parameters.find(format("query[{}]", i));
+            if (ith_query_iter == record.parameters.end()) {
+                break;
+            }
+            query_str += ith_query_iter->second + " ";
+        }
+        if (query_str.empty()) {
+            throw std::logic_error("No \"query\" nor \"query[0]\" parameter set for a session requesting a slow_query_log record");
+        }
+    } else {
+        query_str = query_str_it->second;
     }
 
     // parameters map
@@ -313,7 +325,7 @@ cql3::query_options trace_keyspace_helper::make_slow_query_mutation_data(const o
         cql3::raw_value::make_value(uuid_type->decompose(session_records.session_id)),
         cql3::raw_value::make_value(timestamp_type->decompose(millis_since_epoch)),
         cql3::raw_value::make_value(timeuuid_type->decompose(start_time_id)),
-        cql3::raw_value::make_value(utf8_type->decompose(query_str_it->second)),
+        cql3::raw_value::make_value(utf8_type->decompose(query_str)),
         cql3::raw_value::make_value(int32_type->decompose(elapsed_to_micros(record.elapsed))),
         cql3::raw_value::make_value(make_map_value(my_map_type, map_type_impl::native_type(std::move(parameters_values_vector))).serialize()),
         cql3::raw_value::make_value(inet_addr_type->decompose(record.client.addr())),


tracing/trace_keyspace_helper.cc
```diff
@@ -289,10 +289,23 @@ cql3::query_options trace_keyspace_helper::make_slow_query_mutation_data(const o
     const session_record& record = session_records.session_rec;
     auto millis_since_epoch = std::chrono::duration_cast<std::chrono::milliseconds>(record.started_at.time_since_epoch()).count();
 
-    // query command is stored on a parameters map with a 'query' key
+    sstring query_str;
+    // query command is stored on a parameters map with a 'query' key or 'query[N]' keys
     auto query_str_it = record.parameters.find("query");
```
haaawk, line 294: Super nit: This variable could be defined in smaller scope now (if scope), no?

haaawk (approved):
LGTM

tracing/trace_keyspace_helper.cc
```diff
@@ -289,10 +289,23 @@ cql3::query_options trace_keyspace_helper::make_slow_query_mutation_data(const o
...
     if (query_str_it == record.parameters.end()) {
-        throw std::logic_error("No \"query\" parameter set for a session requesting a slow_query_log record");
+        for (int i = 0; /* nothing */; ++i) {
+            // Extract the values under "query[0]", "query[1]", etc. in case of batched prepared statements
+            auto ith_query_iter = record.parameters.find(format("query[{}]", i));
+            if (ith_query_iter == record.parameters.end()) {
+                break;
+            }
+            query_str += ith_query_iter->second + " ";
```
psarna, line 302: What is the maximum expected size of a batch statement? Can it be, say, a few thousand? I'm asking because a few thousand string allocations can be a strain on the memory allocator, and each string's `operator+=` would allocate. In that case, it would perhaps make sense to use a string builder (e.g. stringstream) instead of adding the strings one by one.

Also, I assume a common case is that we have a single `"query"` param instead of a batch, in which case we should perhaps elide allocations altogether instead of always creating a new string with a trailing space - so perhaps a single query deserves a separate code path which just forwards the string without new allocations.

Slow query tracing is not in any way on the fast path, but on the other hand it's often ran on strained clusters in order to investigate what's wrong with them, so I think it makes sense to minimize the overhead here.

haaawk: We've seen batches of tens of thousands so definitely worth doing.

tracing/trace_keyspace_helper.cc
jul-stas: done!

tracing/trace_keyspace_helper.cc
psarna: If it can be tens of thousands, then one more thing to consider on top of my previous comments is whether we want to construct the whole string... my point being that 50k requests, 100B each will create a 5MB string. The question is whether we need the whole string to investigate the error or maybe a shorter summary would suffice - e.g. a couple of first params and the total number of statements. I don't know the answer to that question, maybe someone from the field has an opinion. Anyway, it's probably a topic for a follow-up later, maybe worth leaving a TODO in the code.

psarna: (by the way, processing tens of thousands strings in a non-preemptive loop is most likely a reactor stall candidate)

vladzcloudius (changes requested):
We already have all queries strings and corresponding parameters in `query[]` and `params[][]` arrays. No need to add a useless `" ".join(query)` string.

I think we should special case batches and not require 'query' key to be set for them.

Keep in mind that a `batch` can be sent using `QUERY` opcode so you want to handle this with caution.

jul-stas:
> No need to add a useless `" ".join(query)` string.
Useless, yes! Then how about this: let's remove the `throw`, default to empty string and go home? No allocations, no stalls, no code.

> Keep in mind that a `batch` can be sent using `QUERY` opcode so you want to handle this with caution.
I'm testing both cases but +1

psarna:
On Mon, Sep 28, 2020, 17:00 Juliusz Stasiewicz <notifications@github.com>
wrote:

> No need to add a useless " ".join(query) string.
>
> Useless, yes! Then how about this: let's remove the throw, default to
> empty string and go home? No allocations, no stalls, no code.
>
That sounds very good.

vladzcloudius:
> On Mon, Sep 28, 2020, 17:00 Juliusz Stasiewicz ***@***.***> wrote: No need to add a useless " ".join(query) string. Useless, yes! Then how about this: let's remove the throw, default to empty string and go home? No allocations, no stalls, no code.
> That sounds very good.

The check makes sense to non-batch queries, so, please, keep it.
Storing an empty string makes no sense at all. Simply don't store anything like I suggested.

PR #7296, opened 2020-09-29
consistency level: fix wrong quorum calculation whe RF = 0

### LINKED ISSUE #6905: (LOCAL_/EACH_)QUORUM consistency calculation is broken when RF=0
*Installation details*
Scylla version: 5e1fa13d088c5fdea7c4626474e367f81d5d1f4d

*Description*
Our current implementation of `db::quorum_for(...)`, `db::local_quorum_for(...)` is going to return 1 when RF for the corresponding replication unit is 0.

One of the cases where this causes a problem is when there is DC in the replication list of a KS that uses `NetworkTopologyStrategy` with RF=0, this DC is not longer present in the cluster (e.g. has been decommissioned) and there comes a write with CL=EACH_QUORUM.

The bug will make every such write fail with `write unavailable` because `assure_sufficient_live_nodes_each_quorum(...)` would always return `false` and this is because `dc_live(0) < dc_block_for(1)` for the DC with RF=0 as described above.

eliransin:
We used to calculate the number of endpoints for quorum and local_quorum
unconditionally as ((rf / 2) + 1). This formula doesn't take into
account the corner case where RF = 0, in this situation quorum should
also be 0.
This commit adds the missing corner case.

Tests: Unit Tests (dev)
Fixes #6905

### DIFF
--- a/db/consistency_level.cc
+++ b/db/consistency_level.cc
@@ -61,7 +61,8 @@ namespace db {
 logging::logger cl_logger("consistency");
 
 size_t quorum_for(const keyspace& ks) {
-    return (ks.get_replication_strategy().get_replication_factor() / 2) + 1;
+    size_t replication_factor = ks.get_replication_strategy().get_replication_factor();
+    return replication_factor ? (replication_factor / 2) + 1 : 0;
 }
 
 size_t local_quorum_for(const keyspace& ks, const sstring& dc) {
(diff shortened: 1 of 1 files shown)


gleb-cloudius:
Do you have unit test or dtest to reproduce the bug and check that it is fixed?

eliransin:
> Do you have unit test or dtest to reproduce the bug and check that it is fixed?

No. Your are right, I will add one (for each case).

kostja:
Second a request for unit test (please atake a look at CQL tests), otherwise LGTM.

PR #7306, opened 2020-09-30
transport: Delay NEW_NODE until CQL listen started

### LINKED ISSUE #7301: NEW_NODE should be sent after listening for CQL clients has started
After adding a new node to the cluster, Scylla sends a `NEW_NODE` event. Some clients (for example Java Driver 3.x - both Scylla and [DataStax version](https://github.com/datastax/java-driver/blob/373cb519310208f93a3b08b70f913d3e95fd16eb/driver-core/src/main/java/com/datastax/driver/core/Cluster.java#L2398-L2418)) immediately try to connect to such new node, however they (sometimes) fail, because Scylla has not started listening for CQL clients yet.

There is a discrepancy between Scylla and Cassandra: Cassandra waits for CQL listening to start (on a new node) before sending `NEW_NODE` to clients. Therefore in Cassandra clients can immediately connect to the new node described in the `NEW_NODE` event. I have manually verified this waiting behaviour by adding a 20s sleep in Cassandra 3.11.8 before starting CQL listening (`Thread.sleep(20000)` before line 161 in [Server.java](https://github.com/apache/cassandra/blob/8b29b698630960a0ebb2c695cc5b21dee4686d09/src/java/org/apache/cassandra/transport/Server.java#L161)).

*Installation details*
Scylla version (or git commit hash): 666.development-0.20200929.1adf2cc84
Cluster size: 1 (2 after adding second node)

avelanarius:
After adding a new node to the cluster, Scylla sends a `NEW_NODE` event to CQL clients. Some clients immediately try to connect to the new node, however they fail as the node has not yet started listening to CQL requests.

In contrast, Apache Cassandra waits for the new node to start its CQL server before sending `NEW_NODE` event. In practice this means that `NEW_NODE` and `UP` events will be sent "jointly" after new node is `UP`.

This change is implemented in the same manner as in Apache Cassandra code:

- [`endpointsPendingJoinedNotification` field in `EventNotifier`](https://github.com/apache/cassandra/blob/79e693e16e2152097c5b27d2d7aaa1763e34f594/src/java/org/apache/cassandra/transport/Server.java#L589-L591)
- [`onUp`](https://github.com/apache/cassandra/blob/79e693e16e2152097c5b27d2d7aaa1763e34f594/src/java/org/apache/cassandra/transport/Server.java#L653-L659)
- [`onJoinCluster`](https://github.com/apache/cassandra/blob/79e693e16e2152097c5b27d2d7aaa1763e34f594/src/java/org/apache/cassandra/transport/Server.java#L635-L641)

Fixes #7301.

### DIFF
--- a/transport/event_notifier.cc
+++ b/transport/event_notifier.cc
@@ -248,6 +248,16 @@ void cql_server::event_notifier::on_drop_aggregate(const sstring& ks_name, const
 
 void cql_server::event_notifier::on_join_cluster(const gms::inet_address& endpoint)
 {
+    if (!gms::get_local_gossiper().is_cql_ready(endpoint)) {
+        _endpoints_pending_joined_notification.insert(endpoint);
+        return;
+    }
+
+    send_join_cluster(endpoint);
+}
+
+void cql_server::event_notifier::send_join_cluster(const gms::inet_address& endpoint)
+{
     for (auto&& conn : _topology_change_listeners) {
         using namespace cql_transport;
         if (!conn->_pending_requests_gate.is_closed()) {
@@ -268,6 +278,10 @@ void cql_server::event_notifier::on_leave_cluster(const gms::inet_address& endpo
 
 void cql_server::event_notifier::on_up(const gms::inet_address& endpoint)
 {
+    if (_endpoints_pending_joined_notification.erase(endpoint)) {
+        send_join_cluster(endpoint);
+    }
+
     bool was_up = _last_status_change.contains(endpoint) && _last_status_change.at(endpoint) == event::status_change::status_type::UP;
     _last_status_change[endpoint] = event::status_change::status_type::UP;
     if (!was_up) {
--- a/transport/server.hh
+++ b/transport/server.hh
@@ -273,6 +273,12 @@ class cql_server::event_notifier : public service::migration_listener,
     std::unordered_map<gms::inet_address, event::status_change::status_type> _last_status_change;
     service::migration_notifier& _mnotifier;
     bool _stopped = false;
+
+    // We want to delay sending NEW_NODE CQL event to clients until the new node
+    // has started listening for CQL requests.
+    std::unordered_set<gms::inet_address> _endpoints_pending_joined_notification;
+
+    void send_join_cluster(const gms::inet_address& endpoint);
 public:
     future<> stop();
     event_notifier(service::migration_notifier& mn);


haaawk:
Could you have a look please, @asias?

transport/event_notifier.cc
```diff
@@ -268,6 +273,10 @@ void cql_server::event_notifier::on_leave_cluster(const gms::inet_address& endpo
 
 void cql_server::event_notifier::on_up(const gms::inet_address& endpoint)
 {
+    if (_endpoints_pending_joined_notification.erase(endpoint)) {
+        on_join_cluster(endpoint);
```
haaawk, line 277: Do we have a guarantee that on_up is always called and never missed? @asias may know

haaawk (approved):
LGTM. The purpose is clear. The code seems right to me but I would like someone more experience in that part of the code (like @asias ) to voice their opinion.

transport/server.hh
```diff
@@ -271,6 +271,11 @@ class cql_server::event_notifier : public service::migration_listener,
     std::unordered_map<gms::inet_address, event::status_change::status_type> _last_status_change;
     service::migration_notifier& _mnotifier;
     bool _stopped = false;
+
+    // We want to delay sending NEW_NODE CQL event to clients until the new node
+    // has started listening for CQL requests.
+    // See https://github.com/scylladb/scylla/issues/7301
```
penberg, line 277: I don't think we should put links to issues for stuff that we actually fixed. It's sufficient that it's mentioned in the commitlog.

transport/event_notifier.cc
penberg, line 277: Please extract the notification logic from `on_join_cluster()` into it's own function and call that from here. It's confusing if the top-level callback functions start calling each other.

penberg: It's supposed to be called, yes. But I can't claim that there are no bugs, of course.

penberg (changes requested):

transport/event_notifier.cc
avelanarius: Fixed in amended commit: https://github.com/scylladb/scylla/pull/7306/commits/35f19c6d3d2259feb2adb282c65a6da345c055d7

avelanarius:
Regarding the question whether `on_up` will get called:

**On the new node side:** After the new node started listening for CQL clients, it sets `cql_ready`:
https://github.com/scylladb/scylla/blob/fd1dd0eac7a303ceaa9f3ff643c4848506b0c85c/transport/controller.cc#L139-L147

This information is then passed to the gossiper:
https://github.com/scylladb/scylla/blob/fd1dd0eac7a303ceaa9f3ff643c4848506b0c85c/transport/controller.cc#L195-L197

(With my basic understanding how gossip works) The gossiper has now changed its local state with new `RPC_READY` value. When "existing" nodes gossip with the new node, they will be sent this new local state (sending done in `do_send_ack2_msg`). Even if sending this message fails, my understanding is that eventually other node will gossip again with new node and learn about new `RPC_READY`.

https://github.com/scylladb/scylla/blob/fd1dd0eac7a303ceaa9f3ff643c4848506b0c85c/gms/gossiper.cc#L372-L387

**On the existing node side:** `storage_service` is a gossip subscriber. When `ack2` gossip message reaches existing node, `storage_service::on_change` will get called and in case of `RPC_READY` this happens: https://github.com/scylladb/scylla/blob/fd1dd0eac7a303ceaa9f3ff643c4848506b0c85c/service/storage_service.cc#L1429-L1432

`notify_cql_change` calls `notify_up` which calls `on_up` on `event_notifier`. 

Whether the gossiper works the way I described and has no bugs - I can't be 100% sure.

haaawk:
@penberg Do you have any other comments?

PR #7311, opened 2020-09-30
transport/config: fix cross-shard use of updateable_value

piodul:
Recently, the cql_server_config::max_concurrent_requests field was
changed to be an updateable_value, so that it is updated when the
corresponding option in Scylla's configuration is live-reloaded.
Unfortunately, due to how cql_server is constructed, this caused
cql_server instances on all shards to store an updateable_value which
pointed to an updateable_value_source on shard 0. Unsynchronized
cross-shard memory operations ensue.

The fix changes the cql_server_config so that it holds a function which
creates an updateable_value appropriate for the given shard. This
pattern is similar to another, already existing option in the config:
get_service_memory_limiter_semaphore.

This fix can be reverted if updateable_value becomes safe to use across
shards.

Tests: unit(dev)

Fixes: #7310

### DIFF
--- a/transport/controller.cc
+++ b/transport/controller.cc
@@ -69,7 +69,9 @@ future<> controller::do_start_server() {
         cql_transport::cql_server_config cql_server_config;
         cql_server_config.timeout_config = make_timeout_config(cfg);
         cql_server_config.max_request_size = service::get_local_storage_service()._service_memory_total;
-        cql_server_config.max_concurrent_requests = cfg.max_concurrent_requests_per_shard;
+        cql_server_config.get_max_concurrent_requests_updateable_value = [ss = std::ref(service::get_storage_service())] () -> utils::updateable_value<uint32_t> {
+            return ss.get().local().db().local().get_config().max_concurrent_requests_per_shard;
+        };
         cql_server_config.get_service_memory_limiter_semaphore = [ss = std::ref(service::get_storage_service())] () -> semaphore& { return ss.get().local()._service_memory_limiter; };
         cql_server_config.allow_shard_aware_drivers = cfg.enable_shard_aware_drivers();
         cql_server_config.sharding_ignore_msb = cfg.murmur3_partitioner_ignore_msb_bits();
--- a/transport/server.hh
+++ b/transport/server.hh
@@ -105,7 +105,7 @@ struct cql_query_state {
 struct cql_server_config {
     ::timeout_config timeout_config;
     size_t max_request_size;
-    utils::updateable_value<uint32_t> max_concurrent_requests;
+    std::function<utils::updateable_value<uint32_t> ()> get_max_concurrent_requests_updateable_value;
     std::function<semaphore& ()> get_service_memory_limiter_semaphore;
     sstring partitioner_name;
     unsigned sharding_ignore_msb;
@@ -125,6 +125,7 @@ private:
     distributed<cql3::query_processor>& _query_processor;
     cql_server_config _config;
     size_t _max_request_size;
+    utils::updateable_value<uint32_t> _max_concurrent_requests;
     semaphore& _memory_available;
     seastar::metrics::metric_groups _metrics;
     std::unique_ptr<event_notifier> _notifier;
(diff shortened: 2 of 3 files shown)


psarna (approved):

bhalevy (approved):
LGTM

transport/controller.cc
```diff
@@ -69,7 +69,9 @@ future<> controller::do_start_server() {
         cql_transport::cql_server_config cql_server_config;
         cql_server_config.timeout_config = make_timeout_config(cfg);
         cql_server_config.max_request_size = service::get_local_storage_service()._service_memory_total;
-        cql_server_config.max_concurrent_requests = cfg.max_concurrent_requests_per_shard;
+        cql_server_config.get_max_concurrent_requests = [ss = std::ref(service::get_storage_service())] () -> utils::updateable_value<uint32_t> {
+            return ss.get().local().db().local().get_config().max_concurrent_requests_per_shard;
+        };
```
avikivity, line 74: updateable_value was supposed to be shard-safe exactly for this use-case!

avikivity:
```c++
// A T that can be updated at runtime; uses updateable_value_base to track
// the source as the object is moved or copied. Copying across shards is supported.
template <typename T>
class updateable_value : public updateable_value_base {
```

transport/server.hh
```diff
@@ -105,7 +105,7 @@ struct cql_query_state {
 struct cql_server_config {
     ::timeout_config timeout_config;
     size_t max_request_size;
-    utils::updateable_value<uint32_t> max_concurrent_requests;
+    std::function<utils::updateable_value<uint32_t> ()> get_max_concurrent_requests;
```
avikivity, line 108: It's sad to call a virtual function just to read a constant from memory,

piodul: OK, I see why it is confusing.

This function returns an updateable_value for the given shard, not just uint32_t. It's only used to initialize the new field `_max_concurrent_requests`, which itself is also an updateable value. This causes `_max_concurrent_requests` attach to the config parameter. When checking if we are above the limit of concurrent requests, we only look at the `_max_concurrent_requests` value, we don't call this std::function.

Perhaps it's better to change it to a more descriptive name like `get_max_concurrent_requests_updateable_value` (ugh)

transport/controller.cc
piodul: Well, the current implementation is not :(

The bug is already on master, so I wanted to provide a simple fix fast. If `updateable_value` was supposed to be shard safe then I agree it should be fixed too, but I think that it will require more effort than this.

piodul:
v2: changed the name of `get_max_concurrent_requests` to `get_max_concurrent_requests_updateable_value` in hope that it's a little more clear what this function is supposed to be for.

transport/controller.cc
avikivity: What about reverting 4b856cf62d839b6d555703f6ba902ff57a132898? We can undo the revert after updateable_value is fixed.

transport/server.hh
avikivity: No, I did not spend enough effort to read it.

transport/controller.cc
piodul: If we revert, then we will lose the ability to hot-reload the param. If we apply this PR we can later revert it after updateable_value is fixed.

piodul: Regardless from what we do with this PR, I'll create a separate issue about updateable_value not being shard-safe.

piodul: https://github.com/scylladb/scylla/issues/7316

piodul:
@haaawk FYI

transport/controller.cc
avikivity: Ok. I'll merge.

haaawk:
This has been already merged - right @piodul ? https://github.com/scylladb/scylla/commit/bfbf02a65705feba28f8163e4c347d22fae75548

Can we close this PR then?

PR #7326, opened 2020-10-04
install.sh: logging to scylla-server.log when journalctl --user does not work

### LINKED ISSUE #7131: nonroot install: can't find scylla log by journalctl --user -xe
*Installation details*
Scylla version (or git commit hash): scylla-unified-package-666.development-0.20200825.f7c5c48df.tar.gz
Cluster size: 1
OS (RHEL/CentOS/Ubuntu/AWS AMI): CentOS8

After nonroot installation on CentOS8.
Scylla-server can be started. But I can't any scylla log in 'jounalctl --user -xe'.

syuu1228:
On some environment such as CentOS8, journalctl --user -xe does not work
since journald is running in volatile mode.
The issue cannnot fix in non-root mode, as a workaround we should logging to
a file instead of journal.

Also added scylla_logrotate to ExecStartPre which rename previous log file,
since StandardOutput=file:/path/to/file will erase existing file when service
restarted.

Fixes #7131

### DIFF
--- a/dist/common/scripts/scylla_logrotate
+++ b/dist/common/scripts/scylla_logrotate
@@ -1,0 +1,31 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 ScyllaDB
+#
+
+#
+# This file is part of Scylla.
+#
+# Scylla is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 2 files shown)


penberg:
@avikivity What do you think about the workaround of writing to a file?

avikivity:
It's reasonable.

avikivity:
Please make sure that std::cout is line buffered even when redirected to a file.

PR #7333, opened 2020-10-05
test: database_test: fix threading confusion

avikivity:
database_test contains several instances of calling do_with_cql_test_env()
with a function that expects to be called in a thread. This mostly works
because there is an internal thread in do_with_cql_test_env(), but is not
guaranteed to.

Fix by switching to the more appropriate do_with_cql_test_env_thread().


bhalevy (approved):
LGTM

PR #7334, opened 2020-10-05
table: fix race and exception handling in on_compaction_completion()

### LINKED ISSUE #7331: Shutting down database hangs in dtest-debug
Seen consistently since https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug/604/testReport
Scylla version f1fcf4f13915c2d3ed532a3a32fc51a8ce7c08ac

https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug/604/artifact/logs-debug.2/1601137898758_repair_additional_test.RepairAdditionalTest.repair_option_pr_dc_host_test/node1.log
```
INFO  2020-09-26 16:14:03,637 [shard 0] init - Shutting down database
Scylla version 666.development-0.20200924.f1fcf4f13 with build-id 1b71d370bdf7e15b911b8e16e18b291737d79b35 starting ...
...
INFO  2020-09-26 16:25:46,848 [shard 0] init - Shutting down database
Scylla version 666.development-0.20200924.f1fcf4f13 with build-id 1b71d370bdf7e15b911b8e16e18b291737d79b35 starting ...
```

https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug/604/artifact/logs-debug.2/1601139448436_repair_additional_test.RepairAdditionalTest.repair_option_pr_multi_dc_test/node1.log
```
INFO  2020-09-26 16:32:42,395 [shard 0] init - Shutting down database
Scylla version 666.development-0.20200924.f1fcf4f13 with build-id 1b71d370bdf7e15b911b8e16e18b291737d79b35 starting ...
```

But not only in those dtests, for example:
https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug/610/artifact/logs-debug.2/1601850222686_bootstrap_test.TestBootstrap.start_stop_test/node1.log
```
INFO  2020-10-04 22:21:36,140 [shard 0] init - Shutting down database
<EOF>
```

Note that this looks similar, but different than https://github.com/scylladb/scylla/issues/7183

avikivity:
Fix a race condition in on_compaction_completion() that can prevent shutdown, as well as an exception handling error. See individual patches for details.

Fixes #7331.

### DIFF
--- a/table.cc
+++ b/table.cc
@@ -991,9 +991,20 @@ table::on_compaction_completion(sstables::compaction_completion_desc& desc) {
         }
     }
 
-    auto new_compacted_but_not_deleted = _sstables_compacted_but_not_deleted;
-    // rebuilding _sstables_compacted_but_not_deleted first to make the entire rebuild operation exception safe.
-    new_compacted_but_not_deleted.insert(new_compacted_but_not_deleted.end(), desc.old_sstables.begin(), desc.old_sstables.end());
+    // Precompute before so undo_compacted_but_not_deleted can be sure not to throw
+    std::unordered_set<sstables::shared_sstable> s(
+           desc.old_sstables.begin(), desc.old_sstables.end());
+    _sstables_compacted_but_not_deleted.insert(_sstables_compacted_but_not_deleted.end(), desc.old_sstables.begin(), desc.old_sstables.end());
+    // After we are done, unconditionally remove compacted sstables from _sstables_compacted_but_not_deleted,
+    // or they could stay forever in the set, resulting in deleted files remaining
+    // opened and disk space not being released until shutdown.
+    auto undo_compacted_but_not_deleted = defer([&] {
+        auto e = boost::range::remove_if(_sstables_compacted_but_not_deleted, [&] (sstables::shared_sstable sst) {
+            return s.contains(sst);
+        });
+        _sstables_compacted_but_not_deleted.erase(e, _sstables_compacted_but_not_deleted.end());
+        rebuild_statistics();
+    });
 
     _cache.invalidate([this, &desc] () noexcept {
         // FIXME: this is not really noexcept, but we need to provide strong exception guarantees.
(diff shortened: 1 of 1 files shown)


table.cc
```diff
@@ -1022,15 +1030,6 @@ table::on_compaction_completion(sstables::compaction_completion_desc& desc) {
...
-    // unconditionally remove compacted sstables from _sstables_compacted_but_not_deleted,
-    // or they could stay forever in the set, resulting in deleted files remaining
-    // opened and disk space not being released until shutdown.
-    std::unordered_set<sstables::shared_sstable> s(
-           desc.old_sstables.begin(), desc.old_sstables.end());
-    auto e = boost::range::remove_if(_sstables_compacted_but_not_deleted, [&] (sstables::shared_sstable sst) {
-        return s.contains(sst);
-    });
-    _sstables_compacted_but_not_deleted.erase(e, _sstables_compacted_but_not_deleted.end());
     rebuild_statistics();
```
bhalevy, line 1033: The call here to rebuild_statistics expects `_sstables_compacted_but_not_deleted` to be already updated so it should also go into the deferred action.

By the way, on the gate closed error path, these sstables aren't really deleted.  I'm not sure they are being tracked correctly, but since this happening on the shutdown path it doesn't matter.

bhalevy (changes requested):

bhalevy:
I tested `repair_additional_test:RepairAdditionalTest.repair_option_pr_multi_dc_test` with this change in debug mode and it indeed fixes https://github.com/scylladb/scylla/issues/7331

table.cc
avikivity: Good catch, will update.

Probably, we should delete those sstables unconditionally. The time to delete isn't worth the extra special case. But I won't do that in this series.

avikivity:
> I tested `repair_additional_test:RepairAdditionalTest.repair_option_pr_multi_dc_test` with this change in debug mode and it indeed fixes #7331

I will add a Fixes tag.

avikivity:
Updated:
 - added Fixes: #7331 (thanks @bhalevy for testing)
 - made final rebuild_statistics) happen after we update the sstable list

bhalevy (approved):
LGTM

eliransin:
LGTM also (not that anyone asked me :slightly_smiling_face: )

PR #7348, opened 2020-10-06
Alternator streams: Fix shard lengths, parenting, expiration, filter useless ones and improve paging

### LINKED ISSUE #7344: Alternator shard ID:s are to long
Docs says max 65 chars. We encode (potentially) 87 chars. 
Suggest removing the stream arn from shard_id, since it is redundant.

### LINKED ISSUE #7345: Alternator endsequencenumber is never set.
The code looks at "expired" in cdc stream gen from table query, but this is neither ever set nor does it really represent the "end" of a generation.

 EndSequenceNumber should be next generations StartSequenceNumber, with maybe a confidence interval overlap.

### LINKED ISSUE #7346: Alternator shards parent is not set very correctly
The code relies on the sort order of cdc streamid:s from query. Should 
a.) explicitly sort these so we can find token bounds
b.) Select parent so that it is the largest token < child token (i.e. start of range covering my range. (though this is somewhat beside the point).

### LINKED ISSUE #7347: Alternator shard ID:s should be filtered by time
Since we cannot have data read from any generation that expired more than 24h ago, nor any generation expired before the stream was created, we should filter out these generations from describe_streams.

Without this, we would get an ever larger, unbounded, set of shards as topology changes in the cluster.
Old code tries to do something like this, but looked at the wrong info (expired - never set - and not what we are looking for), and also was not very efficient.

elcallio:
The remains of the defunct #7246. 

Fixes #7344
Fixes #7345
Fixes #7346
Fixes #7347

Shard ID length is now within limits.
Shard end sequence number should be set when appropriate.
Shard parent is selected a bit more carefully (sorting)
Shards are filtered by time to exclude cdc generations we cannot get data from (too old)
Shard paging improved

### DIFF
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -1029,6 +1029,7 @@ void executor::add_stream_options(const rjson::value& stream_specification, sche
         cdc::options opts;
         opts.enabled(true);
         opts.set_delta_mode(cdc::delta_mode::keys);
+        opts.ttl(std::chrono::duration_cast<std::chrono::seconds>(dynamodb_streams_max_window).count());
 
         auto type = rjson::get_opt<stream_view_type>(stream_specification, "StreamViewType").value_or(stream_view_type::KEYS_ONLY);
         switch (type) {
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -87,16 +87,18 @@ struct rapidjson::internal::TypeHelper<ValueType, utils::UUID>
     : public from_string_helper<ValueType, utils::UUID>
 {};
 
+static db_clock::time_point as_timepoint(const utils::UUID& uuid) {
+    return db_clock::time_point{std::chrono::milliseconds(utils::UUID_gen::get_adjusted_timestamp(uuid))};
+}
+
 /**
  * Note: scylla tables do not have a timestamp as such, 
  * but the UUID (for newly created tables at least) is 
  * a timeuuid, so we can use this to fake creation timestamp
  */
 static sstring stream_label(const schema& log_schema) {
-    auto& uuid = log_schema.id();
-    auto ms = utils::UUID_gen::get_adjusted_timestamp(uuid);
-    std::chrono::system_clock::time_point ts{std::chrono::milliseconds(ms)};
-    auto tt = std::chrono::system_clock::to_time_t(ts);
+    auto ts = as_timepoint(log_schema.id());
+    auto tt = db_clock::to_time_t(ts);
     struct tm tm;
     ::localtime_r(&tt, &tm);
     return seastar::json::formatter::to_json(tm);
@@ -406,6 +408,9 @@ static std::chrono::seconds confidence_interval(const database& db) {
     return std::chrono::seconds(db.get_config().alternator_streams_time_window_s());
 }
 
+// Dynamo docs says no data shall live longer than 24h.
+static constexpr auto dynamodb_streams_max_window = 24h;
+
 future<executor::request_return_type> executor::describe_stream(client_state& client_state, service_permit permit, rjson::value request) {
     _stats.api_operations.describe_stream++;
 
@@ -479,10 +484,18 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
     // or on expired...
     // TODO: maybe add secondary index to topology table to enable this?
     return _sdks.cdc_get_versioned_streams({ tm.count_normal_token_owners() }).then([this, &db, schema, shard_start, limit, ret = std::move(ret), stream_desc = std::move(stream_desc)](std::map<db_clock::time_point, cdc::streams_version> topologies) mutable {
-        auto i = topologies.begin();
+
+        // filter out cdc generations older than the table or now() - dynamodb_streams_max_window (24h)
+        auto low_ts = std::max(as_timepoint(schema->id()), db_clock::now() - dynamodb_streams_max_window);
+
+        auto i = topologies.lower_bound(low_ts);
+        // need first gen _intersecting_ the timestamp.
+        if (i != topologies.begin()) {
+            i = std::prev(i);
+        }
+
         auto e = topologies.end();
         auto prev = e;
-        auto threshold = db_clock::now() - confidence_interval(db);
         auto shards = rjson::empty_array();
 
         std::optional<shard_id> last;
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -223,21 +223,20 @@ future<alternator::executor::request_return_type> alternator::executor::list_str
 struct shard_id {
     static constexpr auto marker = 'H';
 
-    utils::UUID table;
     db_clock::time_point time;
     cdc::stream_id id;
 
-    shard_id(utils::UUID tab, db_clock::time_point t, cdc::stream_id i)
-        : table(tab)
-        , time(t)
+    shard_id(db_clock::time_point t, cdc::stream_id i)
+        : time(t)
         , id(i)
     {}
     shard_id(const sstring&);
 
+    // dynamo specifies shardid as max 65 chars. 
     friend std::ostream& operator<<(std::ostream& os, const shard_id& id) {
         boost::io::ios_flags_saver fs(os);
-        return os << marker << id.table << std::hex
-            << ':' << id.time.time_since_epoch().count()
+        return os << marker << std::hex  
+            << id.time.time_since_epoch().count()
             << ':' << id.id.to_bytes()
             ;
     }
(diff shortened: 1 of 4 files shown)


alternator/streams.cc
```diff
@@ -479,10 +481,18 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
     // or on expired...
     // TODO: maybe add secondary index to topology table to enable this?
     return _sdks.cdc_get_versioned_streams({ tm.count_normal_token_owners() }).then([this, &db, schema, shard_start, limit, ret = std::move(ret), stream_desc = std::move(stream_desc)](std::map<db_clock::time_point, cdc::streams_version> topologies) mutable {
-        auto i = topologies.begin();
+
+        // filter out cdc generations older than the table or now() - dynamo_streams_max_window (24h)
+        auto low_ts = std::max(as_timepoint(schema->id()), db_clock::now() - dynamo_streams_max_window);
+
+        auto i = topologies.lower_bound(low_ts);
```
nyh, line 479: Does this work correctly when paging through the list of shards and time goes by while we're having fun?
Maybe the non-first pages don't start with this "i" and start somewhere else? (I don't know if it does...).

alternator/streams.cc
```diff
@@ -690,7 +685,7 @@ future<executor::request_return_type> executor::get_shard_iterator(client_state&
             inclusive_of_threshold = false;
             break;
         case shard_iterator_type::TRIM_HORIZON:
-            threshold = utils::UUID{};
+            threshold = utils::UUID_gen::min_time_UUID(sid->time.time_since_epoch().count());
```
nyh, line 720: Why is this change needed? In case by freak chance we reuse the same stream token for a new generation as was already used in an older generation?

alternator/streams.cc
```diff
@@ -595,44 +592,39 @@ struct shard_iterator {
     static auto constexpr inclusive_marker = 'I';
     static auto constexpr exclusive_marker = 'i';
 
-    shard_id shard;
+    static auto constexpr uuid_str_length = 36;
```
nyh, line 627: I don't have a better place to leave this comment so I'll randomly leave it hereI:
If the purpose of this patch is to fix the length of shard ids, which need to be shorter than some length, please add a test which verifies this length - which should fail before this patch and pass after (and of course also on DynamoDB). It doesn't need to be a whole new test - can be an additional assert in one of the many existing tests which already deal with shard ids (please with a comment explaining what is being tested).

alternator/streams.cc
```diff
@@ -673,9 +665,12 @@ future<executor::request_return_type> executor::get_shard_iterator(client_state&
         sid = rjson::get<shard_id>(request, "ShardId");
     } catch (...) {
     }
-    if (!schema || !cdc::get_base_table(db, *schema) || !is_alternator_keyspace(schema->ks_name()) || sid->table != schema->id()) {
+    if (!schema || !cdc::get_base_table(db, *schema) || !is_alternator_keyspace(schema->ks_name())) {
         throw api_error::resource_not_found("Invalid StreamArn");
     }
+    if (!sid) {
+        throw api_error::resource_not_found("Invalid ShardId");
+    }
```
nyh, line 705: I wanted to suggest that you add tests which try to pass invalid shard ids or arns, and check we get ResourceNotFound as expected (and run the same test against DynamoDB to confirm that this is the right error to return in this case). But when I tried to look, I found a test you already wrote, test_get_shard_iterator(), which seems to suggest that a bad ARN should throw ValidationException, not ResourceNotFound... What am I missing?

alternator/streams.cc
```diff
@@ -357,44 +357,35 @@ namespace alternator {
...
  * is simply a 1:1 mapping, id=shard. Simple, and fairly easy
  * to do ranged queries later.
  * 
  * Downside is that we get a _lot_ of shards. And with actual
  * generations, we'll get hubba-hubba loads of them. 
  * 
- * Improvement ideas: 
- * 1.) Prune expired streams that cannot have records (check ttls)
- * 2.) Group cdc streams into N per shards (by token range)
+ * We would like to group cdc streams into N per shards
```
nyh, line 366: Adding a comment here because Github doesn't allow me to add comments on the commit message :-(
The word "paging" in the commit message is very ambigous. We do paging of almost everything in Alternator and Scylla, and it's not clear which one is intended here. I didn't understand what "efficient collection skip when doing paging" refers to in the commit message.
Also, please instead of talking about the "sanity" of the parent-child relationship, can you please say a few words about how this parent-child relationship actually works in Scylla? Do we even have this notion?

alternator/streams.cc
```diff
@@ -357,44 +357,35 @@ namespace alternator {
 using namespace std::string_literals;
 
 /**
- * This implementation of CDC stream to dynamo shard mapping
```
nyh, line 361: I think you removed this line by mistake?

alternator/streams.cc
```diff
@@ -405,6 +396,9 @@ static std::chrono::seconds confidence_interval(const database& db) {
     return std::chrono::seconds(db.get_config().alternator_streams_time_window_s());
 }
 
+// Dynamo docs says no data shall live longer than 24h.
+static constexpr auto dynamo_streams_max_window = 24h;
```
nyh, line 400: Oh, found it :-) The patches are in the wrong order (not bisectable).
As I mentioned in my earlier comment, please rename this constant (alternator, probably, or dynamodb if you must?) and please also think how we can use the same constant when configuring the CDC log TTL.

alternator/streams.cc
```diff
@@ -494,40 +488,78 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
+
+        // for parent-child stuff we need id:s to be sorted by token
+        static auto cmp = [](const cdc::stream_id& id1, const cdc::stream_id& id2) {
+            return id1.token() < id2.token();
+        };
+
+        // need a prev even if we are skipping stuff
+        if (i != first) {
+            prev = std::prev(i);
+            std::stable_sort(prev->second.streams.begin(), prev->second.streams.end(), cmp);
```
nyh, line 507: I don't understand this parent-child stuff or what you are sorting or why so it's hard for me to review the validity of this change. Were you able to demonstrate a bug in the old code somehow and this change fixes it? Or how did you test this?

test/alternator/test_streams.py
```diff
@@ -845,9 +848,11 @@ def test_streams_no_records(test_table_ss_keys_only, dynamodbstreams):
     table, arn = test_table_ss_keys_only
     # Get just one shard - any shard - and its LATEST iterator. Because it's
     # LATEST, there will be no data to read from this iterator.
-    shard_id = dynamodbstreams.describe_stream(StreamArn=arn, Limit=1)['StreamDescription']['Shards'][0]['ShardId']
+    shard = dynamodbstreams.describe_stream(StreamArn=arn, Limit=1)['StreamDescription']['Shards'][0]
+    shard_id = shard['ShardId']
     iter = dynamodbstreams.get_shard_iterator(StreamArn=arn, ShardId=shard_id, ShardIteratorType='LATEST')['ShardIterator']
     response = dynamodbstreams.get_records(ShardIterator=iter)
+    assert 'NextShardIterator' in response or 'EndingSequenceNumber' in shard['SequenceNumberRange']
```
nyh, line 855: I had a comment about this line in the previous PR. I'm don't remember what it was ;-)

nyh (changes requested):
Thanks. Most of the patches look good, but I did have some questions or doubts on some thing so please take a look. Thanks.

alternator/streams.cc
elcallio: If someone used describe_streams and waits with getting page N until after cdc generations expire (i.e. shards effectively disappear - which will happen in DynamoDB as well), I would assume that he could potentially miss parts of the sequence. But in that case, we will resume at first point > exclusiveshardid in request.

alternator/streams.cc
elcallio: Actually, on second thought, this should probably go. I put it there to handle potentially changing to range-based-shards eventually, but it is probably better to just retain the zero-uuid threshold so we don't miss any data in the time of generation overlap.

alternator/streams.cc
elcallio: ARN == table id. `sid` above is shard id. That should generate resourcenotfound. And is tested.

alternator/streams.cc
elcallio: Since we cannot force generation changes, we cannot really test this. Even if we can, the parent-child relationship will be ambiguous at best, so I don't know if we can verify it really. The code is mainly just trying to provide a better matching of parent-child via token range of ID:s. Old code used _almost_ but not quite sorted ID:s. 
I've added some more code comments on this in next iteration.

alternator/streams.cc
elcallio: I've updated the comment to include more info.

alternator/streams.cc
elcallio: See line 515

alternator/streams.cc
elcallio: test assert added.

nyh (approved):

Issue #6946, opened 2020-07-28
system.clients is missing information with regards to connected clients


slivne:
Scylla 4.0.X

```
cat /etc/scylla/scylla.yaml  | grep ^auth
authenticator: PasswordAuthenticator
authorizer: CassandraAuthorizer
```

```
[root@e40a07114766 /]# cqlsh -u cassandra -p cassandra
Connected to  at 172.17.0.3:9042.
[cqlsh 5.0.1 | Cassandra 3.0.8 | CQL spec 3.3.1 | Native protocol v4]
Use HELP for help.
cassandra@cqlsh> select * from system.clients ;

 address    | port  | client_type | connection_stage | driver_name | driver_version | hostname | protocol_version | shard_id | ssl_cipher_suite | ssl_enabled | ssl_protocol | username
------------+-------+-------------+------------------+-------------+----------------+----------+------------------+----------+------------------+-------------+--------------+-----------
 172.17.0.3 | 60618 |         cql |             null |        null |           null |     null |                0 |        0 |             null |        null |         null | anonymous
 172.17.0.3 | 60620 |         cql |             null |        null |           null |     null |                0 |        1 |             null |        null |         null | anonymous
```

- too many fields are null all of them should have values.
- username is incorrect its anonymous it should be cassandra 
- protocol version is incorrect its not version 0

tzach:
@amnonh we can include this info in the connect table in the monitoring dashboard

haaawk:
Can we close this already @jul-stas ?

jul-stas:
The encryption info is missing and the last time I talked to @elcallio (> year ago) I learned that it's not trivial to get this info. We can give it a try, but I'm not proficient with crypto unless I'll have to be.

slivne:
Aside of encryption and hostname all should be populated if this is not the case open a bug. 

pushing these based on requests

jul-stas:
I'm closing this in favor of this: #9216

avikivity:
Usually it's better to close the new bugs, to preserve the historical record.

avikivity:
And it's not possible to backport issues that have no commits attached.

jul-stas:
I edited #7349 by adding "Fixes #6946", but apparently it didn't work (too late?)
What can I do instead?

avikivity:
Ah, then add "Fixed by <commit_hash> here, and add the backport candidate label (mimicking what the robots would do if you added the Fixes # thing in time).

[...]

PR #7349, opened 2020-10-06
transport/system_ks: Add more info to `system.clients`

jul-stas:
This patch fills the following columns in `system.clients` table:
* `connection_stage`
* `driver_name`
* `driver_version`
* `protocol_version`

It also improves:
* `client_type` - distinguishes cql from thrift just in case
* `username` - now it displays correct username iff `PasswordAuthenticator` is configured.

What is still missing:
* SSL params (I'll happily get some advice here)
* `hostname` - I didn't find it in tested drivers

Refs #6946

### DIFF
--- a/connection_notifier.cc
+++ b/connection_notifier.cc
@@ -20,18 +20,12 @@
  */
 
 #include "connection_notifier.hh"
-#include "db/query_context.hh"
 #include "cql3/constants.hh"
 #include "database.hh"
-#include "service/storage_proxy.hh"
 
 #include <stdexcept>
 
-namespace db::system_keyspace {
-extern const char *const CLIENTS;
-}
-
-static sstring to_string(client_type ct) {
+sstring to_string(client_type ct) {
     switch (ct) {
         case client_type::cql: return "cql";
         case client_type::thrift: return "thrift";
@@ -52,12 +46,12 @@ future<> notify_new_client(client_data cd) {
             cd.username.value_or("anonymous")).discard_result();
 }
 
-future<> notify_disconnected_client(gms::inet_address addr, client_type ct, int port) {
+future<> notify_disconnected_client(net::inet_address addr, int port, client_type ct) {
     // FIXME: consider prepared statement
     const static sstring req
             = format("DELETE FROM system.{} where address=? AND port=? AND client_type=?;",
                      db::system_keyspace::CLIENTS);
-    return db::execute_cql(req, addr.addr(), port, to_string(ct)).discard_result();
+    return db::execute_cql(req, std::move(addr), port, to_string(ct)).discard_result();
 }
 
 future<> clear_clientlist() {
--- a/connection_notifier.cc
+++ b/connection_notifier.cc
@@ -30,18 +30,27 @@ sstring to_string(client_type ct) {
         case client_type::cql: return "cql";
         case client_type::thrift: return "thrift";
         case client_type::alternator: return "alternator";
-        default: throw std::runtime_error("Invalid client_type");
     }
+    throw std::runtime_error("Invalid client_type");
+}
+
+static sstring to_string(client_connection_stage ccs) {
+    switch (ccs) {
+        case client_connection_stage::established: return connection_stage_literal<client_connection_stage::established>;
+        case client_connection_stage::authenticating: return connection_stage_literal<client_connection_stage::authenticating>;
+        case client_connection_stage::ready: return connection_stage_literal<client_connection_stage::ready>;
+    }
+    throw std::runtime_error("Invalid client_connection_stage");
 }
 
 future<> notify_new_client(client_data cd) {
     // FIXME: consider prepared statement
     const static sstring req
-            = format("INSERT INTO system.{} (address, port, client_type, shard_id, protocol_version, username) "
-                     "VALUES (?, ?, ?, ?, ?, ?);", db::system_keyspace::CLIENTS);
+            = format("INSERT INTO system.{} (address, port, client_type, connection_stage, shard_id, protocol_version, username) "
+                     "VALUES (?, ?, ?, ?, ?, ?, ?);", db::system_keyspace::CLIENTS);
     
     return db::execute_cql(req,
-            std::move(cd.ip), cd.port, to_string(cd.ct), cd.shard_id,
+            std::move(cd.ip), cd.port, to_string(cd.ct), to_string(cd.connection_stage), cd.shard_id,
             cd.protocol_version.has_value() ? data_value(*cd.protocol_version) : data_value::make_null(int32_type),
             cd.username.value_or("anonymous")).discard_result();
 }
--- a/connection_notifier.hh
+++ b/connection_notifier.hh
@@ -20,9 +20,17 @@
  */
 #pragma once
 
-#include "gms/inet_address.hh"
+#include "db/query_context.hh"
+
+#include <seastar/net/inet_address.hh>
 #include <seastar/core/sstring.hh>
+#include "seastarx.hh"
+
 #include <optional>
+
+namespace db::system_keyspace {
+extern const char *const CLIENTS;
+}
 
 enum class client_type {
     cql = 0,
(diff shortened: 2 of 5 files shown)


connection_notifier.hh
```diff
@@ -20,19 +20,46 @@
...
+
+template <changed_column column> constexpr const char* column_literal = "";
+template <> constexpr const char* column_literal<changed_column::username> = "username";
+template <> constexpr const char* column_literal<changed_column::connection_stage> = "connection_stage";
+template <> constexpr const char* column_literal<changed_column::driver_name> = "driver_name";
+template <> constexpr const char* column_literal<changed_column::driver_version> = "driver_version";
+template <> constexpr const char* column_literal<changed_column::hostname> = "hostname";
+template <> constexpr const char* column_literal<changed_column::protocol_version> = "protocol_version";
+
 // Representation of a row in `system.clients'. std::optionals are for nullable cells.
```
avikivity, line 71: Maybe add another template for the type? So if the caller passes the wrong type, we don't get a runtime error.

connection_notifier.hh
```diff
@@ -52,6 +79,17 @@ struct client_data {
...
+struct notify_client_change {
+    template <typename T>
+    future<> operator()(net::inet_address addr, int port, client_type ct, T&& value) {
+        const static sstring req
+                = format("UPDATE system.{} SET {}=? WHERE address=? AND port=? AND client_type=?;",
+                        db::system_keyspace::CLIENTS, column_literal<column_enum_val>);
+
+        return db::execute_cql(req, std::forward<T>(value), std::move(addr), port, to_string(ct)).discard_result();
+    }
+};
```
avikivity, line 106: Nice. We may yet evolve a compile-time query generator.

transport/server.cc
```diff
@@ -579,11 +579,15 @@ future<> cql_server::connection::shutdown()
     return make_ready_future<>();
 }
 
+std::tuple<net::inet_address, int, client_type> cql_server::connection::make_client_key(const service::client_state& cli_state) {
+    return std::make_tuple(cli_state.get_client_address().addr(),
+            cli_state.get_client_port(),
+            cli_state.is_thrift() ? client_type::thrift : client_type::cql);
+}
```
avikivity, line 592: How can this be thrift, here in transport/?

avikivity: Ah, client_state is neutral.

avikivity:
Pretty nice. But please make the cover letter refer to what it does, not to what it doesn't do.

transport/server.cc
```diff
@@ -371,6 +371,11 @@ cql_server::connection::read_frame() {
                 _version = current_version;
                 throw exceptions::protocol_exception(format("Invalid or unsupported protocol version: {:d}", client_version));
             }
+
+          auto client_state_notification_f = std::apply(notify_client_change<changed_column::protocol_version>{},
+                  std::tuple_cat(make_client_key(_client_state), std::make_tuple(_version)));
+
+          return client_state_notification_f.then_wrapped([this] (future<>) {
```
haaawk, line 378: Shouldn't we check the resulting future for exceptions? Otherwise, we will be sweeping it under the carpet

transport/server.cc
```diff
@@ -761,18 +763,37 @@ future<std::unique_ptr<cql_server::response>> cql_server::connection::process_st
...
+                std::tuple_cat(make_client_key(_client_state), std::make_tuple(driver_ver_opt->second)));
+    }
+    if (auto driver_name_opt = options.find("DRIVER_NAME"); driver_name_opt != options.end()) {
+        _client_state.set_driver_name(driver_name_opt->second);
+        client_state_notification_f = client_state_notification_f.then([ck = make_client_key(_client_state), dn = driver_name_opt->second] {
+            return std::apply(notify_client_change<changed_column::driver_name>{},
+                    std::tuple_cat(std::move(ck), std::forward_as_tuple(dn)));
+        });
+    }
+
```
haaawk, line 780: This may issue two updates, no? Maybe we could make it only a single update. It will make the code more complex so probably not worth doing. This is not performance-critical code, is it?

transport/server.cc
```diff
@@ -761,18 +763,37 @@ future<std::unique_ptr<cql_server::response>> cql_server::connection::process_st
...
-    if (a.require_authentication()) {
-        return make_ready_future<std::unique_ptr<cql_server::response>>(make_autheticate(stream, a.qualified_java_name(), trace_state));
+    std::unique_ptr<cql_server::response> res;
+    if (auto& a = client_state.get_auth_service()->underlying_authenticator(); a.require_authentication()) {
+        res = make_autheticate(stream, a.qualified_java_name(), trace_state);
+    } else {
+        res = make_ready(stream, trace_state);
     }
-    return make_ready_future<std::unique_ptr<cql_server::response>>(make_ready(stream, trace_state));
+    return client_state_notification_f.then_wrapped([res = std::move(res)] (future<>) mutable {
```
haaawk, line 802: Probably not a good idea to ignore the resulting future without checking for exceptions.

connection_notifier.cc
```diff
@@ -34,14 +34,23 @@ sstring to_string(client_type ct) {
     }
 }
 
+static sstring to_string(client_connection_stage ccs) {
+    switch (ccs) {
+        case client_connection_stage::established: return connection_stage_literal<client_connection_stage::established>;
+        case client_connection_stage::authenticating: return connection_stage_literal<client_connection_stage::authenticating>;
+        case client_connection_stage::ready: return connection_stage_literal<client_connection_stage::ready>;
+        default: throw std::runtime_error("Invalid client_connection_stage");
```
haaawk, line 42: Won't the compiler complain about switch not covering all the enum values?
I think adding default destroys this compile-time safety, no?
Probably better to throw after the switch.

connection_notifier.hh
```diff
@@ -57,17 +57,28 @@ template <> constexpr const char* column_literal<changed_column::driver_version>
...
+enum class client_connection_stage {
+    established = 0,
+    authenticating,
+    ready,
+};
+
+template <client_connection_stage ccs> constexpr const char* connection_stage_literal = "";
+template <> constexpr const char* connection_stage_literal<client_connection_stage::established> = "ESTABLISHED";
+template <> constexpr const char* connection_stage_literal<client_connection_stage::authenticating> = "AUTHENTICATING";
+template <> constexpr const char* connection_stage_literal<client_connection_stage::ready> = "READY";
```
haaawk, line 69: Would it make sense to use string_view here?

transport/server.cc
```diff
@@ -1146,7 +1154,12 @@ cql_server::connection::process_register(uint16_t stream, request_reader in, ser
         auto et = parse_event_type(event_type);
         _server._notifier->register_event(et, this);
     }
-    return make_ready_future<std::unique_ptr<cql_server::response>>(make_ready(stream, std::move(trace_state)));
+    auto client_state_notification_f = std::apply(notify_client_change<changed_column::connection_stage>{},
+            std::tuple_cat(make_client_key(_client_state), std::make_tuple(connection_stage_literal<client_connection_stage::ready>)));
+
+    return client_state_notification_f.then_wrapped([res = make_ready(stream, std::move(trace_state))] (future<>) mutable {
```
haaawk, line 1160: Probably worth checking for exceptions on resulting future.

connection_notifier.hh
```diff
@@ -20,19 +20,46 @@
...
+};
+
+template <changed_column column> constexpr const char* column_literal = "";
+template <> constexpr const char* column_literal<changed_column::username> = "username";
+template <> constexpr const char* column_literal<changed_column::connection_stage> = "connection_stage";
+template <> constexpr const char* column_literal<changed_column::driver_name> = "driver_name";
+template <> constexpr const char* column_literal<changed_column::driver_version> = "driver_version";
+template <> constexpr const char* column_literal<changed_column::hostname> = "hostname";
+template <> constexpr const char* column_literal<changed_column::protocol_version> = "protocol_version";
+
```
haaawk, line 59: Would it make sense to use string_view here?

jul-stas: While technically possible (`string_view` has trivial dtor and constexpr methods), I don't see much of an advantage in doing this.
The literals are short, they don't have much in common (so pooling them won't win us much). And `const char*` can be used as template param if we ever decide to build a full-blown compile-time query generator.

haaawk: ok

connection_notifier.hh
jul-stas: Do you mean templating `const char*`?

transport/server.cc
jul-stas: It slows down connection handshake, so not really critical.
I already tried variadic `notify_client_change`, but it gets awfully complex. So if you don't insist, I'd happily stay away from it.

transport/server.cc
jul-stas: Done

connection_notifier.cc
jul-stas: Done

transport/server.cc
jul-stas: Done

transport/server.cc
jul-stas: Done

transport/server.cc
haaawk: Ok. Don't bother then.

connection_notifier.hh
haaawk: @avikivity ?

haaawk (approved):
LGTM

haaawk:
@avikivity Could you please clarify https://github.com/scylladb/scylla/pull/7349#discussion_r500473548 or maybe merge if you find the PR acceptable?

connection_notifier.hh
avikivity: I see all the types are text, so not much point.

avikivity: Actually, we also have inet_address. I thought to have two template types for column_literal, one for the ID, one for the type, so if you use the wrong type you get a compile time error. But it's not worth the effort.

PR #7358, opened 2020-10-07
utils: array-search: deinline, working around clang bug

avikivity:
Clang has a bug processing inline ifuncs with intrinsics[1].
Since ifuncs can't be inlined anyway (they are always dispatched
via a function pointer that is determined based on the CPU
features present), nothing is gained by inlining them. Deinlining
therefore reduces compile time and works around the clang bug.

[1] https://bugs.llvm.org/show_bug.cgi?id=47691


nyh:
I'll commit, but have one comment on pre-existing code which you can perhaps consider for a followup:

The source file includes a warning that one version (the most important version...) of the function "doesn't take @size argument into account and expects all the elements above it to be less than any possible value.". I don't understand this phrase at all (what are elements "above it", and what does it mean for them to be "less than any possible value", and how does any of this relate to "doesn't take @size argument into account").
But in any case, if this warning is important, and users of this function need to know about it, the warning should also appear in the header file - not just in the source file.

avikivity:
The comment was there originally, I didn't add it. I can move it to the header.

avikivity:
Do you want me to move before you commit or after?

nyh:
> Do you want me to move before you commit or after?

I committed already because as you said, you just moved the pre-existing comment with the code, so it's not a reason to delay this patch.
Note that beyond the fact that some sort of comment should be in the header file, not just source file, my second problem was that I couldn't actually understand this comment :-( So if you did, maybe you can rephrase it somehow...

avikivity:
I have some vague understanding - it's a requirement that the values from &array[size]  to &array[capacity-1] must be numeric_limits<>::min.

avikivity:
This is because the SIMD instructions used do not support masking. It could have been done correctly with AVX-512.

nyh:
> I have some vague understanding - it's a requirement that the values from &array[size] to &array[capacity-1] must be numeric_limits<>::min.

Wow, this is an extremely unnatural requirement that never happens "by accident", so definitely needs to appear in a comment.
But I still don't understand how capacity is relevant there. Maybe it's not capacity but something like align_up(size, 16) or something like that? Which can be even higher than capacity?

avikivity:
No, it's capacity (which is required to be a multiple of 4)

avikivity:
(the align_up requirement was pushed to capacity)

avikivity:
(the code always reads and compares element in the range [size, capacity), so they have this extra requirement.

nyh:
> No, it's capacity (which is required to be a multiple of 4)
Ok, now I finally understand what the comment was trying to say:

1. All values between size and capacity-1 must be set to `numeric_limits<...>::min`.
2. capacity must be a multiple of 4.

PR #7368, opened 2020-10-09
cql3/statements/batch_statement.cc: improve batch size warning message

### LINKED ISSUE #7367: Make batchlog size warning clearer
*Installation details*
HEAD: 6c133e36d851cf64682438922a02d545f410975b

*Description*
Our current warning about a BATCH being too big is confusing:

```
Batch of prepared statements for XX is of size YYY, exceeding specified WARN/FAIL threshold of ZZZ by NNN.
```

This message makes a reader think that he/she's being warned about the batch of the size YYY, where YYY is the number of statements in the batch, while in fact this is about the amount of memory the batch is about to write.

Let's add the 'bytes' word to make it clear.

vladzcloudius:
Make the warning message clearer:
 * Include the number of statements in the batch.
 * Be clear that the warning is about the batch size in bytes.

Fixes #7367

Signed-off-by: Vlad Zolotarov <vladz@scylladb.com>

### DIFF
--- a/cql3/statements/batch_statement.cc
+++ b/cql3/statements/batch_statement.cc
@@ -233,8 +233,8 @@ void batch_statement::verify_batch_size(service::storage_proxy& proxy, const std
             for (auto&& m : mutations) {
                 ks_cf_pairs.insert(m.schema()->ks_name() + "." + m.schema()->cf_name());
             }
-            return format("Batch of prepared statements for {} is of size {:d}, exceeding specified {} threshold of {:d} by {:d}.",
-                    join(", ", ks_cf_pairs), size, type, threshold, size - threshold);
+            return format("Batch of {:d} prepared statements for {} is of size {:d} bytes, exceeding specified {} threshold of {:d} by {:d}.",
+                    mutations.size(), join(", ", ks_cf_pairs), size, type, threshold, size - threshold);
         };
         if (size > fail_threshold) {
             _logger.error(error("FAIL", fail_threshold).c_str());


cql3/statements/batch_statement.cc
```diff
@@ -233,8 +233,8 @@ void batch_statement::verify_batch_size(service::storage_proxy& proxy, const std
             for (auto&& m : mutations) {
                 ks_cf_pairs.insert(m.schema()->ks_name() + "." + m.schema()->cf_name());
             }
-            return format("Batch of prepared statements for {} is of size {:d}, exceeding specified {} threshold of {:d} by {:d}.",
-                    join(", ", ks_cf_pairs), size, type, threshold, size - threshold);
+            return format("Batch of {:d} prepared statements for {} is of size {:d} bytes, exceeding specified {} threshold of {:d} by {:d}.",
+                    mutations.size(), join(", ", ks_cf_pairs), size, type, threshold, size - threshold);
         };
```
avikivity, line 238: mutations.size() is not the number of statements, it's the number of tables affected. Two statements that affect one table result in one mutation.

gleb-cloudius: Two statements that affect different partitions in one table will result in two mutations.

avikivity: Right. But to statements that affect one partition will create one mutation.

gleb-cloudius: Yes. For Cassundra this will be true even if partitions belong to different tables in the same keyspace, but not for us.

vladzcloudius: Sending a v2

PR #7369, opened 2020-10-10
cql3: use larger stack for do_with_cql_parser() in debug mode

avikivity:
Our cql parser uses large amounts of stack, and can overflow it
in debug mode with clang. To prevent this stack overflow,
temporarily use a larger (1MB) stack.


cql3/util.cc
```diff
@@ -19,9 +19,20 @@
...
+void __sanitizer_start_switch_fiber(void** fake_stack_save, const void* stack_bottom, size_t stack_size);
+void __sanitizer_finish_switch_fiber(void* fake_stack_save, const void** stack_bottom_old, size_t* stack_size_old);
+}
+
+#endif
+
 namespace cql3::util {
 
-void do_with_parser_impl(const sstring_view& cql, noncopyable_function<void (cql3_parser::CqlParser& parser)> f) {
+static void do_with_parser_impl_impl(const sstring_view& cql, noncopyable_function<void (cql3_parser::CqlParser& parser)> f) {
```
nyh, line 35: A new rule of programming: any problem can be solved by adding another layer of "impl" :-)

nyh (commented):
Looks good to me, but I wonder if this could look less like black magic if you created a general purpose function like `run_with_big_stack(func, stack_size)` and use that, instead of mixing the cql parser code with the how-to-run-with-a-big-stack ugliness. Maybe you'll need to use this `run_with_big_stack` for something else as well, beyond the CQL parser you found now.

avikivity:
> Looks good to me, but I wonder if this could look less like black magic if you created a general purpose function like `run_with_big_stack(func, stack_size)` and use that, instead of mixing the cql parser code with the how-to-run-with-a-big-stack ugliness. Maybe you'll need to use this `run_with_big_stack` for something else as well, beyond the CQL parser you found now.

It's the natural extension, but premature abstraction also has its problems.

For example, we may choose to move such complicated tasks to an alien thread (which naturally won't have a restricted stack) and the number of users of the new abstraction will drop rather than increase.

nyh:
> > Looks good to me, but I wonder if this could look less like black magic if you created a general purpose function like `run_with_big_stack(func, stack_size)` and use that, instead of mixing the cql parser code with the how-to-run-with-a-big-stack ugliness. Maybe you'll need to use this `run_with_big_stack` for something else as well, beyond the CQL parser you found now.
> 
> It's the natural extension, but premature abstraction also has its problems.
> 
> For example, we may choose to move such complicated tasks to an alien thread (which naturally won't have a restricted stack) and the number of users of the new abstraction will drop rather than increase.

Ok, then. I'll commit as-is, and if someone encounters the same need later, they'll just do this abstraction later.

PR #7372, opened 2020-10-10
utils: log_heap: relax check for clang's sanitizer

avikivity:
b1e78313fe7ed58 added a check for ubsan to squelch a false positive,
but that check doesn't work with clang. Relax it to check for debug
mode, so clang doesn't hit the same false positive as gcc did.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -252,7 +252,7 @@ def find_headers(repodir, excluded_dirs):
 
 modes = {
     'debug': {
-        'cxxflags': '-DDEBUG -DDEBUG_LSA_SANITIZER -DSCYLLA_ENABLE_ERROR_INJECTION',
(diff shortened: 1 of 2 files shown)


nyh:
How is "DEBUG" a relevant for what you wanted to test?
I checked in config.py, and we set DEBUG even in "sanitize" build mode, but this feels accidental to me, and maybe in the future we'll want to enable sanitizer without DEBUG, no?
If we can't trust gcc or clang to have a common ifdef (does clang really have no such ifdef?), we can add a new "-DSANITIZER" ourselvees in configurepy in sanitize and debug modes. I feel that hitching onto DEBUG isn't the right thing.

avikivity:
I could not find a define for clang, only discussions that it's needed.

I'll add a new flag for us.

avikivity:
Updated with a SANITIZE macro

nyh:
Thanks. I think that's more future-proof. Committed.

PR #7373, opened 2020-10-10
utils: fragmented_temporary_buffer: don't add to potentially null pointers

avikivity:
Offsetting a null pointer is undefined, and clang's ubsan complains.

Rearrange the arithmetic so we never offset a null pointer. A function
is introduced for the remaining contiguous bytes so it can cast the result
to size_t, avoiding a compare-of-different-signedness warning from gcc.


avikivity:
Updated: rearranged even more to avoid a compare-of-different-signedness warning from gcc

PR #7376, opened 2020-10-10
test: network_topology_strategy_test: fix overflow in d2t()

avikivity:
d2t() scales a fraction in the range [0, 1] to the range of
a biased token (same as unsigned long). But x86 doesn't support
conversion to unsigned, only signed, so this is a truncating
conversion. Clang's ubsan correctly warns about it.

Fix by reducing the range before converting, and expanding it
afterwards.

### DIFF
--- a/test/boost/network_topology_strategy_test.cc
+++ b/test/boost/network_topology_strategy_test.cc
@@ -133,7 +133,10 @@ void endpoints_check(
 }
 
 auto d2t = [](double d) -> int64_t {
-    return static_cast<unsigned long>(d*(std::numeric_limits<unsigned long>::max()));
+    // Double to unsigned long conversion will overflow if the
+    // input is greater than numeric_limits<long>::max(), so divide by two and
+    // multiply again later.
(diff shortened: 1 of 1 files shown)


nyh:
I'll commit, because this is just for a test that doesn't need any great precision, but what I don't like about this trick is that I think that this trick loses a bit (literally) of precision that the original version had:
For example, we start with the double 2^-64 (which should be representable exactly in floating point), we expect the resulting integer to be 1, but with your formula it will return 0, just like it will return for the double 0.

avikivity:
double only has 53 bits of precision, so no bits were lost.

nyh:
`2^-64` needs only one bit of precision... But you're right about fractions closer to 1.
Again, I am not suggesting there is any practical implications for this difference for this specific test.

avikivity:
In fact before the patch we lost half the range - our tokens would be in the range (-2^63, -1] instead of (-2^63, 2^63).

The difference between gcc and clang wasn't the result, it was the warning in ubsan.

PR #7380, opened 2020-10-10
test: sstable_datafile_test: sstable_run_based_compaction_test: prevent use of uninitialized variable observer

avikivity:
The variable 'observer' (an std::optional) may be left uninitialized
if 'incremental_enabled' is false. However, it is used afterwards
with a call to disconnect, accessing garbage.

Fix by accessing it via the optional wrapper. A call to optional::reset()
destroys the observable, which in turn calls disconnect().


raphaelsc:
thanks for working on this @avikivity

PR #7381, opened 2020-10-10
configure.py: fix run in NixOS build sandbox

conferno:
configure.py fails to run on NixOS sandbox.
There are 3 issues:
1. `/etc/os-release` does not exist
2. `subprocess.check_output(['pkg-config'], env=None)` effectively removes `$PATH`, so `pkg-config` cannot be found
3. `args.user_cflags += '-I abseil'` concatenates strings without any space


avikivity:
@conferno Please see our [cla](https://www.scylladb.com/open-source/contributor-agreement/) requirement. We also require a name on the patch Author: line, not an alias.

PR #7389, opened 2020-10-11
utils: big_decimal: work around clang difficulty with boost::cpp_int(string_view) constructor

avikivity:
Clang has some difficulty with the boost::cpp_int constructor from string_view.
In fact it is a mess of enable_if<>s so a human would have trouble too.

Work around it by converting to std::string. This is bad for performance, but
this constructor is not going to be fast in any case.

Hopefully a fix will arrive in clang or boost.

### DIFF
--- a/utils/big_decimal.cc
+++ b/utils/big_decimal.cc
@@ -26,6 +26,21 @@
 
 #include <regex>
 
+#ifdef __clang__
+
+// Clang or boost have a problem navigating the enable_if maze
+// that is cpp_int's constructor. It ends up treating the
+// string_view as binary and "0" ends up 48.
+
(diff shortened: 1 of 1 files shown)


nyh:
Thanks. Committed. I wonder if the Alternator Streams bug you got with clang 10 (but not clang 11) is similar to the issue you fixed here. I have a vague recollection that @elcallio also had trouble in the past with the cpp_int constructor (with gcc) and it caused parsing bugs in the past.

elcallio:
Quite so. It got erroneously dispatched because of the said enable_if:s not catching the users (mine) intent.
The result was borked sequence number parsing -> could very well be the bug reported.

PR #7392, opened 2020-10-11
build: use c++20 for all C++ files, not just those that use the seastar flags

avikivity:
A few source files (like those generated by antlr) don't build with seastar,
and so don't inherit all of its flags. They then use the compiler default dialect,
not C++20. With gcc that's just fine, since gcc supports concepts in earlier dialects,
but clang requires C++20.

Fix by forcing --std=c++20 for all files.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1507,7 +1507,7 @@ with open(buildfile_tmp, 'w') as f:
         configure_args = {configure_args}
         builddir = {outdir}
         cxx = {cxx}
-        cxxflags = {user_cflags} {warnings} {defines}
+        cxxflags = --std=gnu++20 {user_cflags} {warnings} {defines}
         ldflags = {linker_flags} {user_ldflags}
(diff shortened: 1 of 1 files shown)


avikivity:
This somehow broke the build with gcc, so I dequeued it.

nyh:
On Mon, Oct 12, 2020 at 9:44 AM Avi Kivity <notifications@github.com> wrote:

> This somehow broke the build with gcc, so I dequeued it.
>
Interesting. Like for all your patches, I compile-tested this patch and it
didn't seem to break anything. But I think I did not try to build all the
tests, so maybe it broke one of them.

avikivity:
yes, it broke view_build_test.

avikivity:
Fixed. Changed to use --std=gnu++20.

I couldn't figure out from the error message what extension we were using. Generally I prefer not to use extensions unless strictly needed.

PR #7394, opened 2020-10-11
types: don't linearize ascii during validation

### LINKED ISSUE #7393: ascii validation of large cells causes large allocations
A very large cell (megabytes) is kept fragmented until it is validated, at which point it is linearized. This causes a large allocation that can stall.

avikivity:
ascii has no inter-byte dependencies and so can
be validated fragment by fragment, reducing large
contiguous allocations.

Fixes #7393.

### DIFF
--- a/types.cc
+++ b/types.cc
@@ -1430,11 +1430,12 @@ struct validate_visitor {
         }
     }
     void operator()(const ascii_type_impl&) {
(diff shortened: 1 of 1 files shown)


raphaelsc (approved):
LGTM

denesb (approved):

PR #7399, opened 2020-10-12
scylla_raid_setup: use sysfs to detect existing RAID volume

### LINKED ISSUE #7383: On clean ubuntu 20.04 image, scylla_setup fails on  '/dev/md0 is already using'
*Installation details*
Scylla version (or git commit hash): 4.1.7
Cluster size: 4
OS (RHEL/CentOS/Ubuntu/AWS AMI): Ubuntu 20.04
https://www.googleapis.com/compute/v1/projects/ubuntu-os-cloud/global/images/family/ubuntu-2004-lts

Seem like /dev/md0 is already in used in those images, and there is no way to change this parameter vi `scylla_setup`
only if you call with `--no-raid-setup` and run `scylla_raid_setup` on you own

```
23:57:40  Command: 'sudo /usr/lib/scylla/scylla_setup --nic ens5 --disks /dev/nvme0n1,/dev/nvme0n2,/dev/nvme0n3,/dev/nvme0n4,/dev/nvme0n5,/dev/nvme0n6,/dev/nvme0n7,/dev/nvme0n8  --setup-nic-and-disks'
23:57:40  Exit code: 1
23:57:40  Stdout:
23:57:40  ntp-systemd-netif.service is a disabled or a static unit, not starting it.
23:57:40  Setting up libevent-core-2.1-7:amd64 (2.1.11-stable-1) ...
23:57:40  Setting up libevent-pthreads-2.1-7:amd64 (2.1.11-stable-1) ...
23:57:40  Setting up sntp (1:4.2.8p12+dfsg-3ubuntu4) ...
23:57:40  Processing triggers for systemd (245.4-4ubuntu3.2) ...
23:57:40  Processing triggers for man-db (2.9.1-1) ...
23:57:40  Processing triggers for libc-bin (2.31-0ubuntu9.1) ...
23:57:40  8 Oct 20:57:22 ntpdate[26322]: adjust time server 91.189.91.157 offset -0.000273 sec
23:57:40  /dev/md0 is already using
```

**EDIT**:
I know we don't have an official support for this version, let's call it a heads up :), since we were trying it anyhow

syuu1228:
We may not able to detect existing RAID volume by device file existance,
we should use sysfs instead to make sure it's running.

Fixes #7383

### DIFF
--- a/dist/common/scripts/scylla_raid_setup
+++ b/dist/common/scripts/scylla_raid_setup
@@ -73,7 +73,8 @@ if __name__ == '__main__':
             print('{} is busy'.format(disk))
             sys.exit(1)
 
-    if os.path.exists(args.raiddev):
+    raiddevname = os.path.basename(args.raiddev)
+    if os.path.exists(f'/sys/block/{raiddevname}/md/array_state'):
         print('{} is already using'.format(args.raiddev))
         sys.exit(1)
 


fruch:
@syuu1228 and the process after that would work regardless `/dev/md0` exist or not ?

penberg:
@fruch Yes, we want Ubuntu 20.04 support, not sure what else is missing.

penberg:
@syuu1228 Can you elaborate on why probing sysfs is the right approach? I see `/proc/mdstat` mentioned elsewhere. @avikivity any preferences on RAID detection?

avikivity:
I think /sys is  preferable to /proc, since less parsing is involved.

penberg:
Ok, @syuu1228 can you point us to the source (e.g. documentation) that says this is the right thing to do to detect the presence of a RAID volume?

syuu1228:
Why I choose sysfs is it's better to avoid bit complicated text parsing on /proc/mdstat if it's possible, since it might cause another bug, I want to perform detection by os.path.exists() instead.

After RAID started sysfs creates /sys/block/${raiddevice}/md/ and also number of files inside, and 'array_state' file
in the directory looks good to detect RAID because it's always there.

There is a manual of mdraid: https://www.kernel.org/doc/Documentation/admin-guide/md.rst
It says:
```
md devices appear in sysfs (``/sys``) as regular block devices,
e.g.::

   /sys/block/md0

Each ``md`` device will contain a subdirectory called ``md`` which
contains further md-specific information about the device.

All md devices contain:

(skip descriptions of each files)

  array_state
     This file contains a single word which describes the current
     state of the array.  In many cases, the state can be set by
     writing the word for the desired state, however some states
     cannot be explicitly set, and some transitions are not allowed.

     Select/poll works on this file.  All changes except between
     Active_idle and active (which can be frequent and are not
     very interesting) are notified.  active->active_idle is
     reported if the metadata is externally managed.

     clear
         No devices, no size, no level

         Writing is equivalent to STOP_ARRAY ioctl

     inactive
         May have some settings, but array is not active
         all IO results in error

         When written, doesn't tear down array, but just stops it

     suspended (not supported yet)
         All IO requests will block. The array can be reconfigured.

         Writing this, if accepted, will block until array is quiessent

     readonly
         no resync can happen.  no superblocks get written.

         Write requests fail

     read-auto
         like readonly, but behaves like ``clean`` on a write request.

     clean
         no pending writes, but otherwise active.

         When written to inactive array, starts without resync

         If a write request arrives then
         if metadata is known, mark ``dirty`` and switch to ``active``.
         if not known, block and switch to write-pending

         If written to an active array that has pending writes, then fails.
     active
         fully active: IO and resync can be happening.
         When written to inactive array, starts with resync

     write-pending
         clean, but writes are blocked waiting for ``active`` to be written.

     active-idle
         like active, but no writes have been seen for a while (safe_mode_delay).
```

PR #7401, opened 2020-10-12
data/cell: fix value_writer use before definition

avikivity:
Clang parses templates more eagerly than gcc, so it fails on
some forward-declared templates. In this case, value_writer
was forward-declared and then used in data::cell. As it also
uses some definitions local to data::cell, it cannot be
defined before it as well as after it.

To solve the problem, we define it as a nested class so
it can use other local definitions, yet be defined before it
is used. No code changes.


nyh:
I'll commit. It's sad how the already small cell_impl.hh became even smaller, but I am assuming you considered all the options and this is the best you came up with. We can revisit the size of cell.hh (which is used throughout the project) later :-(

avikivity:
Note that there is no impact on compile time, since data/cell.h includes cell-impl.hh.

nyh:
> Note that there is no impact on compile time, since data/cell.h includes cell-impl.hh.

Oh.... So why does cell-impl.hh even exist?

denesb:
I guess for the usual reason, to move the definitions into a separate file so that the declarations are easier to read.

PR #7187, opened 2020-09-07
install.sh: show warning nonroot mode when systemd does not support user mode

syuu1228:
This is v2 of 7089, removed key press waiting since it likely breaked next build

----

On older distribution such as CentOS7, it does not support systemd user mode.
On such distribution nonroot mode does not work, show warning message and
skip running systemctl --user.

Fixes #7071

PR #7402, opened 2020-10-12
install.sh: allow --packaging with nonroot mode

syuu1228:
Since scylla-ccm wants to skip systemctl, we need to support --packaging
in nonroot mode too.

Related: #7187


syuu1228:
This is prepare to fix syntax error while merging #7187.

We also need to do `if ! $packaging; then check_usermode_support; fi` in #7187, and call install.sh with --packaging in ccm, then we can drop sed hack to remove systemctl in ccm.

see: https://github.com/scylladb/scylla-ccm/blob/next/ccmlib/scylla_repository.py#L241

PR #7405, opened 2020-10-12
install.sh: allow --packaging with nonroot mode

syuu1228:
Since scylla-ccm wants to skip systemctl, we need to support --packaging
in nonroot mode too.

Related: https://github.com/scylladb/scylla-ccm/pull/250
Related: scylladb/scylla-ccm#249
Related: #7187

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -393,7 +393,9 @@ if $nonroot; then
     # nonroot install is also 'offline install'
     touch $rprefix/SCYLLA-OFFLINE-FILE
     touch $rprefix/SCYLLA-NONROOT-FILE
-    systemctl --user daemon-reload
+    if ! $packaging; then
+        systemctl --user daemon-reload
+    fi
     echo "Scylla non-root install completed."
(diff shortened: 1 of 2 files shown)


penberg:
@syuu1228 What are the semantics of the `--packaging` option? The help text is not that helpful...

```
  --packaging               use install.sh for packaging
```

avikivity:
"just copy the file files, don't worry about issuing commands to complete integration" (since these will be done at the real install time by rpm/deb)

avikivity:
Maybe we should rename it --only-files

syuu1228:
@avikivity @penberg Merged a PR for scylla-ccm https://github.com/scylladb/scylla-ccm/pull/250 to avoid breaking jenkins build, we can merge this now.
After merging this, relocatable package will be version 2.1.
From relocatable package version 2.1, we can use ./install.sh --packaging to skip running systemctl non scylla-ccm, no need to rewrite install.sh by sed on scylla-ccm anymore.

Issue #6964, opened 2020-08-02
Node was overloaded: Too many in flight hints during rolling upgrade


ShlomiBalalis:
*Installation details*
Scylla version (or git commit hash): 4.1.3-0.20200729.9002592ee0c initialization completed.
Cluster size: 4
OS (RHEL/CentOS/Ubuntu/AWS AMI): https://www.googleapis.com/compute/v1/projects/debian-cloud/global/images/family/debian-9

In the rolling upgrade test of the master branch, running with debian 9, the stress failed during the upgrade of one of node#1 (when the node was down and in the stage of the upgrade) with the following error:
```
2020-07-31 04:21:18.637: (CassandraStressLogEvent Severity.ERROR): type=IOException regex=java\.io\.IOException line_number=8300 node=Node rolling-upgrade-master-debian-stret-loader-node-a5343403-0-1 [35.231.58.153 | 10.142.15.193] (seed: False)
java.io.IOException: Operation x10 on key(s) [503639373034334c4e30]: Error executing: (OverloadedException): Queried host (rolling-upgrade-master-debian-stret-db-node-a5343403-0-2.c.skilled-adapter-452.internal/10.142.0.79:9042) was overloaded: Too many in flight hints: 10485768
2020-07-31 04:21:18.652: (CassandraStressLogEvent Severity.ERROR): type=IOException regex=java\.io\.IOException line_number=8373 node=Node rolling-upgrade-master-debian-stret-loader-node-a5343403-0-1 [35.231.58.153 | 10.142.15.193] (seed: False)
WARN  04:21:18,593 Host rolling-upgrade-master-debian-stret-db-node-a534java.io.IOException: Operation x10 on key(s) [503636383635364d3830]: Error executing: (OverloadedException): Queried host (rolling-upgrade-master-debian-stret-db-node-a5343403-0-2.c.skilled-adapter-452.internal/10.142.0.79:9042) was overloaded: Too many in flight hints: 10485768
2020-07-31 04:21:18.665: (CassandraStressLogEvent Severity.ERROR): type=IOException regex=java\.io\.IOException line_number=8395 node=Node rolling-upgrade-master-debian-stret-loader-node-a5343403-0-1 [35.231.58.153 | 10.142.15.193] (seed: False)
[...]

PR #7407, opened 2020-10-12
hints: don't read hint files when it's not allowed to send

piodul:
When there are hint files to be sent and the target endpoint is DOWN,
end_point_hints_manager works in the following loop:

- It reads the first hint file in the queue,
- For each hint in the file it decides that it won't be sent because the
  target endpoint is DOWN,
- After realizing that there are some unsent hints, it decides to retry
  this operation after sleeping 1 second.

This causes the first segment to be wholly read over and over again,
with 1 second pauses, until the target endpoint becomes UP or leaves the
cluster. This causes unnecessary I/O load in the streaming scheduling
group.

This patch adds a check which prevents end_point_hints_manager from
reading the first hint file at all when it is not allowed to send hints.

First observed in #6964

Tests:
- unit(dev)
- hinted handoff dtests

### DIFF
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -825,7 +825,7 @@ void manager::end_point_hints_manager::sender::send_hints_maybe() noexcept {
     int replayed_segments_count = 0;
 
     try {
-        while (replay_allowed() && have_segments()) {
+        while (replay_allowed() && have_segments() && can_send()) {
             if (!send_one_file(*_segments_to_replay.begin())) {
                 break;
             }


nyh (approved):
Looks good to me. I'll commit.

haaawk (approved):
LGTM

nyh:
Committed. @piodul, should we backport this patch to older releases (4.2, 4.1, 4.0)? We should backport if this issue really caused a noticeable performance degradation, but if this is just a cleanup and the performance impact is tiny, we better not unnecessarily backport.

By the way, in general (for next time), please create an issue which your PR "fixes" (not just "fixed noticed in..."). The the closing of such an issue will remind us to consider it for backporting. See `docs/backport.md` for more information on how we track and do backports.

vladzcloudius:
> Committed. @piodul, should we backport this patch to older releases (4.2, 4.1, 4.0)? We should backport if this issue really caused a noticeable performance degradation, but if this is just a cleanup and the performance impact is tiny, we better not unnecessarily backport.

I doubt the impact of this change is measurable, @nyh 

> 
> By the way, in general (for next time), please create an issue which your PR "fixes" (not just "fixed noticed in..."). The the closing of such an issue will remind us to consider it for backporting. See `docs/backport.md` for more information on how we track and do backports.

piodul:
@nyh Thanks for committing. I'll remember to create an issue for backports next time.

@vladzcloudius I think that the impact will be measurable. In issue #6964, you can see that the bug wasted around 40-50 MB/s worth of IO when the cluster should be doing nothing at all. I also saw that the CPU usage was at 100% all the time when it was happening.

Personally, I think that we should backport to all versions that we still support. The bug was present since 4.1 for sure (as evidenced in #6964), but looking at the code of 4.0 it should happen there as well. What do you think?

vladzcloudius:
> @nyh Thanks for committing. I'll remember to create an issue for backports next time.
> 
> @vladzcloudius I think that the impact will be measurable. In issue #6964, you can see that the bug wasted around 40-50 MB/s worth of IO when the cluster should be doing nothing at all. I also saw that the CPU usage was at 100% all the time when it was happening.

Are doubt that waking every second, opening a file and reading a single record is going to waste 100% CPU. This makes no sense.

If you claim it's measurable - let's measure it and see some numbers. Something like before and after the change. If we see that what you've described above is caused by these 1s wakeups - we should dig deeper. We must have some terrible bug in the I/O backend probably.

> 
> Personally, I think that we should backport to all versions that we still support. The bug was present since 4.1 for sure (as evidenced in #6964), but looking at the code of 4.0 it should happen there as well. What do you think?

I don't think this deserves a backport to any release unless we see that it fixes anything.
AFAIKT this patch fixes nothing unless proven otherwise.

We don't backport cleanups.

vladzcloudius:
> > @piodul
> > I doubt that waking every second, opening a file and reading a single record is going to waste 100% CPU. This makes no sense.
> 
> @vladzcloudius Reading a single record? The `send_one_file` function does not stop on the first record if it detects that it cannot send. It only marks the `segment_replay_failed` flag and continues. Here is the relevant part from the code:
> 
> ```c++
> commitlog::read_log_file(fname, manager::FILENAME_PREFIX, service::get_local_streaming_priority(), [this, secs_since_file_mod, &fname, ctx_ptr] (commitlog::buffer_and_replay_position buf_rp) mutable {
>     auto&& [buf, rp] = buf_rp;
>     // Check that we can still send the next hint. Don't try to send it if the destination host
>     // is DOWN or if we have already failed to send some of the previous hints.
>     if (!draining() && ctx_ptr->segment_replay_failed) {
>         return make_ready_future<>();
>     }
> 
>     // Break early if stop() was called or the destination node went down.
>     if (!can_send()) {
>         ctx_ptr->segment_replay_failed = true;
>         return make_ready_future<>();
>     }
> 
>     return flush_maybe().finally([this, ctx_ptr, buf = std::move(buf), rp, secs_since_file_mod, &fname] () mutable {
>         return send_one_hint(std::move(ctx_ptr), std::move(buf), rp, secs_since_file_mod, fname);
>     });
> }, _last_not_complete_rp.pos, &_db.extensions()).get();
> ```
> 
> There is nothing here which prevents `read_log_file` from reading the rest of the file and invoking the callback on every other record. Perhaps returning an exception would work, but in this case there is no need to open the file at all if we are sure that we won't do anything useful with the data.

@piodul 
Right, so it reads the whole 32MB file and moves to the next entry right away. Still this doesn't explain 100% CPU utilization.

vladzcloudius:
Here how looks a CPU usage of a single shard scylla process (4.0.something) when there is nothing do by spin on hints:

image

So, definitely not a 100%.

From the perf-top I see that most (~28%) of the time is spent in calculating CRC32.

When we recall that this all runs in a streaming scheduling class this all is very unlikely going to be measurable if we try to express this overhead in query processing performance.

Like I mentioned a few posts above.

PR #7410, opened 2020-10-12
alternator::streams: Ensure shards are reported in string lexical order

### LINKED ISSUE #7409: Alternator stream shards must be reported in lexicographically sorted order.
While never specified, this is assumed and required by the kinesis adapter, which wholly ignores the "lastevaluatedshardid" field, and just looks at - not the last sent - but the last sent as sorted - shard in list.

This happened to work before, but was broken by me in the parent-management PR.

### LINKED ISSUE #7424: AWS kinesis expects SequenceNumbers to be bigint comparable (growing)
Since out stream positions are timeuuids, their bits (views as 128 bits unsigned or whatnot), is _not_ the sort order. Timeuuid msb is somewhat weirdly mangled (see UUID_gen.hh), and thus comparing them like this will potentially run into cases where we report two sequentially growing sequence numbers that kinesis will read as out-of-order. 

We can work around this by de-mangle the msb when converting the values in alternator streams. Since timeuuid sorting is still sort by tuple(time, random lsb), we can reconstruct the time|lsb 128bit int instead.

elcallio:
Fixes #7409

AWS kinesis Java sdk requires/expects shards to be reported in
lexical order, and even worse, ignores lastevalshard. Thus not
upholding said order will break their stream intropection badly.

Added asserts to unit tests.

Also Fixes #7424 (added - since the basic problem is more or less the same)

Ensure sequencenumbers are bigint lexical.

### DIFF
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -519,14 +519,21 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
         // stored order, which is not neccesarily token order,
         // so the finding of "closest" token boundary (using upper bound)
         // could give somewhat weird results.
-        static auto cmp = [](const cdc::stream_id& id1, const cdc::stream_id& id2) {
+        static auto token_cmp = [](const cdc::stream_id& id1, const cdc::stream_id& id2) {
             return id1.token() < id2.token();
+        };
+
+        // #7409 - shards must be returned in lexicographical order,
+        // normal bytes compare is string_traits<int8_t>::compare.
+        // thus bytes 0x8000 is less than 0x0000. By doing unsigned
+        // compare instead we inadvertently will sort in string lexical.
+        static auto id_cmp = [](const cdc::stream_id& id1, const cdc::stream_id& id2) {
+            return compare_unsigned(id1.to_bytes(), id2.to_bytes()) < 0;
         };
 
         // need a prev even if we are skipping stuff
         if (i != first) {
             prev = std::prev(i);
-            std::stable_sort(prev->second.streams.begin(), prev->second.streams.end(), cmp);
         }
 
         for (; limit > 0 && i != e; prev = i, ++i) {
@@ -537,15 +544,19 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
             auto lo = sv.streams.begin();
             auto end = sv.streams.end();
 
-            // first sort by index. (see above)
-            std::stable_sort(lo, end, cmp);
+            // #7409 - shards must be returned in lexicographical order,
+            std::sort(lo, end, id_cmp);
 
             if (shard_start) {
                 // find next shard position
-                lo = std::upper_bound(lo, end, shard_start->id.token(), [](const dht::token& t, const cdc::stream_id& id) {
-                    return t < id.token();
-                });
+                lo = std::upper_bound(lo, end, shard_start->id, id_cmp);
                 shard_start = std::nullopt;
+            }
+
+            if (lo != end && prev != e) {
+                // We want older stuff sorted in token order so we can find matching
+                // token range when determining parent shard.
+                std::stable_sort(prev->second.streams.begin(), prev->second.streams.end(), token_cmp);
             }
 
             auto expired = [&]() -> std::optional<db_clock::time_point> {
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -270,7 +270,17 @@ struct sequence_number {
 
         using namespace boost::multiprecision;
 
-        uint128_t hi = uint64_t(num.uuid.get_most_significant_bits());
+        /**
+         * #7424 - aws sdk assumes sequence numbers are
+         * monotonically growing bigints. 
+         *
+         * Timeuuids viewed as msb<<64|lsb are _not_,
+         * but they are still sorted as
+         *  timestamp() << 64|lsb
+         * so we can simpy unpack the mangled msb
+         * and use as hi 64 in our "bignum".
+         */
+        uint128_t hi = uint64_t(num.uuid.timestamp());
         uint128_t lo = uint64_t(num.uuid.get_least_significant_bits());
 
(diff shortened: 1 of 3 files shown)


alternator/streams.cc
```diff
@@ -519,14 +519,26 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
+        // normal bytes compare is string_traits<int8_t>::compare.
+        // thus bytes 0x8000 is less than 0x0000. By doing unsigned
+        // compare instead we inadvertently will sort in string lexical.
+        using lexical_view = std::basic_string_view<uint8_t>;
+        static auto make_lexical_view = [](const bytes& b) {
+            return lexical_view(reinterpret_cast<const uint8_t*>(b.data()), b.size());
+        };
+        static auto id_cmp = [](const cdc::stream_id& id1, const cdc::stream_id& id2) {                    
+            return make_lexical_view(id1.to_bytes()) < make_lexical_view(id2.to_bytes());
+        };
```
nyh, line 535: All of this looks correct, but in bytes.hh we have a function `compare_unsigned()` to do exactly the comparison you need (you only need "<" so you need `compare_unsigned(..,...) < 0`).
We had a bug like this in Alternator in the past - see commit f6b1f45d696acc0fff5922e0bd3c31e7c6348894
That being said, I'm ok with your re-implementation, it seems simple enough (I don't know about the performance). Your call.

alternator/streams.cc
```diff
@@ -593,6 +602,12 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
 
                 last = std::nullopt;
             }
+
+            if (limit > 0) {
+                // #7409 - shards must be returned in lexicographical order,
+                // so only sort for prevs sake - i.e. finding parent
+                std::stable_sort(sv.streams.begin(), sv.streams.end(), token_cmp);
```
nyh, line 609: I don't understand what I'm seeing here... Didn't you already sort the same array by id_cmp? Why do we need to resort it using token_cmp?  The comment seems to contradict what you do (you say "must be returned in lexicographical order" and then sort differently)...
Also, what does the "limit > 0" test mean?

test/alternator/test_streams.py
```diff
@@ -484,12 +484,25 @@ def list_shards(dynamodbstreams, arn):
...
     assert len(set(shards)) == len(shards)
+    # 7409 - kinesis required shards to be in lexical order.
+    # verify.
+    assert shards == sorted(shards)
+
+    # special test: ensure we get nothing more if we ask for starting at the last
+    # of the last
+    response = dynamodbstreams.describe_stream(StreamArn=arn,
+        ExclusiveStartShardId=shards[-1])['StreamDescription']
+    assert not 'Shards' in response or len(response['Shards']) == 0
```
nyh, line 504: Please assert an "or" of two options unless there is a very good reason. If DynamoDB returns missing 'Shards' we should too, and if DynamoDB returns empty 'Shards' we should to. This test suggests to me maybe Alternator and DynamoDB return a different thing here, and it's not a good thing.

nyh (changes requested):

alternator/streams.cc
elcallio: Missed that one. Updated to use it.

test/alternator/test_streams.py
elcallio: The or should not be there. Removed.

alternator/streams.cc
elcallio: We need to return shards in "byte representation" order (lexical). But when we process next generation and this is "prev", we want things in token order, so we can quickly find the parent of a shard (if any). 
The limit check is simply to bypass the sort if we are with this paged call - no need to sort if we are not processing anything else.

elcallio: Simplified.

alternator/streams.cc
```diff
@@ -270,7 +270,17 @@ struct sequence_number {
 
         using namespace boost::multiprecision;
 
-        uint128_t hi = uint64_t(num.uuid.get_most_significant_bits());
+        /**
+         * #7424 - aws sdk assumes sequence numbers are
+         * monotonically growing bigints.
```
haaawk, line 275: timestamps in cdc$time are non-decreasing is that enough?

elcallio: Timestamps are monotonically growing, and timeuuids are sorted as such, _but_ their bit representation if viewed just as 128 unsigned bytes are _not_, because the msb portion is bitswapped around (reflected in the comparator). What we do here is simply unswap the bits so we get the "actual" value <timestamp>|lsb as a 128 bit number.

haaawk: it is possible for two rows in CDC Log to have cdc$time that translates to the same timestamp. cdc$time is a timeuuid especially because of that. To resolve conflicts when two operations have the same timestamp. Thus, timestamps in cdc$time are non-decreasing but not strictly growing.

haaawk: Maybe the resulting 128-bit values after the transformation of msb will be sorted in the same way for two timeuuids that share the same timestamp but I don't know. That's why I'm asking. Plus the non-decreasing vs strictly growing thing.

haaawk (approved):
LGTM in general. Apart from one question.

alternator/streams.cc
elcallio: Yes, that is the effect. We just ensure the 128-bit int is the bigint representation of how the timeuuid was sorted (the effective bit comparison order).

nyh:
@elcallio @haaawk please note I already merged this series - so if there are things we want to further improve, let's do it as a followup patch.

slivne:
@elcallio / @haaawk - nyh has merged something - if this one is nore relevant please close

nyh:
I don't know what happened - as you can see above, I did merge this series into "next", but somehow my merge did not reach master (so github also didn't close this PR), and yet, the patches did reach master. I guess some sort of dequeuing incident?
I'm closing this PR.

PR #7417, opened 2020-10-13
cql3/statements/batch_statement.cc: improve batch size warning message

### LINKED ISSUE #7367: Make batchlog size warning clearer

vladzcloudius:
Make the warning message clearer:
 * Include the number of partitions affected by the batch.
 * Be clear that the warning is about the batch size in bytes.

Fixes #7367

Signed-off-by: Vlad Zolotarov <vladz@scylladb.com>

### DIFF
--- a/cql3/statements/batch_statement.cc
+++ b/cql3/statements/batch_statement.cc
@@ -233,8 +233,8 @@ void batch_statement::verify_batch_size(service::storage_proxy& proxy, const std
             for (auto&& m : mutations) {
                 ks_cf_pairs.insert(m.schema()->ks_name() + "." + m.schema()->cf_name());
             }
-            return format("Batch of prepared statements for {} is of size {:d}, exceeding specified {} threshold of {:d} by {:d}.",
-                    join(", ", ks_cf_pairs), size, type, threshold, size - threshold);
+            return format("Batch modifying {:d} partitions in {} is of size {:d} bytes, exceeding specified {} threshold of {:d} by {:d}.",
+                    mutations.size(), join(", ", ks_cf_pairs), size, type, threshold, size - threshold);
         };
         if (size > fail_threshold) {
             _logger.error(error("FAIL", fail_threshold).c_str());


cql3/statements/batch_statement.cc
```diff
@@ -233,8 +233,8 @@ void batch_statement::verify_batch_size(service::storage_proxy& proxy, const std
             for (auto&& m : mutations) {
                 ks_cf_pairs.insert(m.schema()->ks_name() + "." + m.schema()->cf_name());
             }
-            return format("Batch of prepared statements for {} is of size {:d}, exceeding specified {} threshold of {:d} by {:d}.",
-                    join(", ", ks_cf_pairs), size, type, threshold, size - threshold);
+            return format("Batch modifying {:d} partitions in {} is of size {:d} bytes, exceeding specified {} threshold of {:d} by {:d}.",
+                    mutations.size(), join(", ", ks_cf_pairs), size, type, threshold, size - threshold);
```
raphaelsc, line 237: @vladzcloudius not necessarily unique partitions right? if the batch contains multiple updates on different rows of same partition. maybe those mutations are already in compacted form?!

vladzcloudius: @raphaelsc Mutations to the same partitions are already squashed into a single partition before this line here: https://github.com/scylladb/scylla/blob/master/cql3/statements/batch_statement.cc#L201

The call flow goes as follows: 

```
   batch_statement::do_execute(..):
          get_mutations()   <--- squashes mutations for the same partition into a single 'mutation'
          execute_without_conditions()
                validate_batch_size()  <----- the message is here
```

raphaelsc: thanks for the explanation, @vladzcloudius

PR #7422, opened 2020-10-13
cql3: expression: switch from range_bound to interval_bound to avoid clang class template argument deduction woes

avikivity:
Clang does not implement P1814R0 (class template argument deduction
for alias templates), so it can't deduce the template arguments
for range_bound, but it can for interval_bound, so switch to that.
Using the modern name rather than the compatibility alias is preferred
anyway.


avikivity:
@nyh ping

avikivity:
@nyh ping

avikivity:
@nyh ping

Issue #7364, opened 2020-10-08
Reading range tombstones causes memory pressure on LSA 


slivne:
Split from: #7336 

Scylla version : 4.2.rc5

```
Oct 02 18:29:08 longevity-200gb-48h-verify-limited--db-node-1b9d6ec2-4 scylla[1450]:  [shard 10] seastar_memory - oversized allocation: 1970176 bytes. This is non-fatal, but could lead to latency and/or fragmentation issues. Please report
                                                                                        0x333dc10
                                                                                        0x333e099
                                                                                        0x2e348b3
                                                                                        0x2e393eb
                                                                                        0xfac9d7
                                                                                        0xffc3ad
                                                                                        0x100251b
                                                                                        0x1004249
                                                                                        0x100497b
                                                                                        0x100435c
                                                                                        0x11a6ea7
                                                                                        0xff3ad8
                                                                                        0xff409d
                                                                                        0x1168c0c
                                                                                        0x1169eda
                                                                                        0x116b585
                                                                                        0x116dafd
                                                                                        0x116e07a
                                                                                        0x116f97c
                                                                                        0x1174933
                                                                                        0x1174f07
                                                                                        0x117558d
                                                                                        0x115b32f
                                                                                        0x115f2d1
                                                                                        0x2e83d97
                                                                                        0x2e8410e
                                                                                        0x2ebb1ed
                                                                                        0x2ecabea
                                                                                        0x2e4ee6d
                                                                                        /opt/scylladb/libreloc/libpthread.so.0+0x9431
                                                                                        /opt/scylladb/libreloc/libc.so.6+0x101912
                                                                                        --------
                                                                                        seastar::smp_message_queue::async_work_item<(anonymous namespace)::shard_reader::do_fill_buffer(std::chrono::time_point<seastar::lowres_clock, std::ch
Oct 02 18:29:08 longevity-200gb-48h-verify-limited--db-node-1b9d6ec2-4 scylla[1450]:  [shard 10] seastar_memory - oversized allocation: 1970176 bytes. This is non-fatal, but could lead to latency and/or fragmentation issues. Please report
                                                                                        0x333dc10
                                                                                        0x333e099
                                                                                        0x2e348b3
                                                                                        0x2e393eb
                                                                                        0xfac9d7
                                                                                        0xffc3ad
                                                                                        0x100251b
                                                                                        0x1004249
                                                                                        0x100497b
                                                                                        0x100435c
                                                                                        0x11a6ea7
[...]

PR #7433, opened 2020-10-14
partition_version: Change range_tombstones() to return chunked_vector

elcallio:
Refs #7364

The number of tombstones can be large. As a stopgap measure to
just returning a source range (with keepalive), we can at least
alleviate the problem by using a chunked vector.


haaawk (approved):
LGTM

avikivity:
Dequeued, tests don't build.

elcallio:
Ugh. How strange/embarrassing. Rebased and fixed.

PR #7436, opened 2020-10-14
cdc: Add an "end-of-record" column to

### LINKED ISSUE #7435: CDC log should include an end of record marker
Since a CDC "event" is broken up into potentially several log rows, each representing a sub-modification or pre/post state of a given change/source mutation, a reader cannot really be sure if he has actually read a full event unless he explicitly sees the next timestamp set being. 

Combined with paging and query limits this makes it more difficult to implement a CDC reader/translator without resorting to over-reads and/or stateful caching between calls. 

By adding an end-of-row marker to the last row in a given timestamp group we can always know if we are in fact at an event boundary, even if we don't have "row spill" to check, which makes it easier to write a forward-only reader/translator.

elcallio:
Fixes #7435

Adds an "eor" (end-of-record) column to cdc log. This is non-null only on
last-in-timestamp group rows, i.e. end of a singular source "event".

A client can use this as a shortcut to knowing whether or not he has a
full cdc "record" for a given source mutation (single row change).

### DIFF
--- a/cdc/log.cc
+++ b/cdc/log.cc
@@ -519,6 +519,7 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID> 
     b.with_column(log_meta_column_name_bytes("batch_seq_no"), int32_type, column_kind::clustering_key);
     b.with_column(log_meta_column_name_bytes("operation"), data_type_for<operation_native_type>());
     b.with_column(log_meta_column_name_bytes("ttl"), long_type);
+    b.with_column(log_meta_column_name_bytes("end_of_batch"), boolean_type);
     b.set_caching_options(caching_options::get_disabled_caching_options());
     auto add_columns = [&] (const schema::const_iterator_range_type& columns, bool is_data_col = false) {
         for (const auto& column : columns) {
@@ -880,12 +881,24 @@ public:
         return _base_schema;
     }
 
+    clustering_key create_ck(int batch) const {
+        return clustering_key::from_exploded(_log_schema, { _tuuid, int32_type->decompose(batch) });
+    }
+
     // Creates a new clustering row in the mutation, assigning it the next `cdc$batch_seq_no`.
     // The numbering of batch sequence numbers starts from 0.
     clustering_key allocate_new_log_row() {
-        auto log_ck = clustering_key::from_exploded(_log_schema, { _tuuid, int32_type->decompose(_batch_no++) });
+        auto log_ck = create_ck(_batch_no++);
         set_key_columns(log_ck, _base_schema.partition_key_columns(), _base_pk);
         return log_ck;
+    }
+
+    bool has_rows() const {
+        return _batch_no != 0;
+    }
+
+    clustering_key last_row_key() const {
+        return create_ck(_batch_no - 1);
     }
 
     // A common pattern is to allocate a row and then immediately set its `cdc$operation` column.
@@ -944,6 +957,11 @@ public:
         _log_mut.set_cell(log_ck, log_cdef, atomic_cell::make_live(*log_cdef.type, _ts, deleted_elements, _ttl));
     }
 
+    void end_record() {
+        if (has_rows()) {
+            _log_mut.set_cell(last_row_key(), log_meta_column_name_bytes("end_of_batch"), data_value(true), _ts, _ttl);
+        }
+    }
 private:
     void set_key_columns(const clustering_key& log_ck, schema::const_iterator_range_type columns, const std::vector<bytes>& key) {
         size_t pos = 0;
@@ -1514,6 +1532,11 @@ public:
             ._generate_delta_values = generate_delta_values(_builder->base_schema())
         };
         cdc::inspect_mutation(m, v);
+    }
+
+    void end_record() override {
+        assert(_builder);
+        _builder->end_record();
     }
 
     // Takes and returns generated cdc log mutations and associated statistics about parts touched during transformer's lifetime.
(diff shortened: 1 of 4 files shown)


cdc/log.cc
```diff
@@ -519,6 +519,7 @@ static schema_ptr create_log_schema(const schema& s, std::optional<utils::UUID>
     b.with_column(log_meta_column_name_bytes("batch_seq_no"), int32_type, column_kind::clustering_key);
     b.with_column(log_meta_column_name_bytes("operation"), data_type_for<operation_native_type>());
     b.with_column(log_meta_column_name_bytes("ttl"), long_type);
+    b.with_column(log_meta_column_name_bytes("eor"), boolean_type);
```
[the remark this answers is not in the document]
elcallio: I purposefully picked as short a column name as I could, but if you want to make it more verbatim, how about "end_of_batch"?

test/boost/cdc_test.cc
```diff
@@ -369,7 +370,7 @@ SEASTAR_THREAD_TEST_CASE(test_cdc_log_schema) {
         assert_has_column(cdc::log_data_column_deleted_elements_name("c_typ"), set_type_impl::get_instance(short_type, false));
 
         // Check if we missed something
-        BOOST_REQUIRE_EQUAL(required_column_count, log_schema->all_columns_count());
+        BOOST_REQUIRE_GE(required_column_count, log_schema->all_columns_count());
```
haaawk, line 373: Why is this needed?
assert_has_column(cdc::log_meta_column_name("eor"), boolean_type); should have increased required_column_count by 1, no?

elcallio: It is not, but I found the check pernickety. Have _more_ columns than those required should not really be a problem.

cdc/log.cc
```diff
@@ -944,6 +948,11 @@ class log_mutation_builder {
         _log_mut.set_cell(log_ck, log_cdef, atomic_cell::make_live(*log_cdef.type, _ts, deleted_elements, _ttl));
     }
 
+    void end_record() {
```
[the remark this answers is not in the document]
elcallio: ? Look at the call chain. the caller (split.cc) only has _source_ ck.

cdc/log.cc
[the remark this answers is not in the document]
haaawk: end of batch is fine

test/boost/cdc_test.cc
haaawk: I think it's better to be precise. Otherwise we may miss errors when one column is added but tests are not updated and then another column is added and tests are just updated by +1. Then the test wouldn't catch the second column not working.

cdc/log.cc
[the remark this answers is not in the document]
haaawk: Ok. I see this is needed.

haaawk (approved):

cdc/log.cc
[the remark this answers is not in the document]
elcallio: Done

test/boost/cdc_test.cc
elcallio: Reverted

kbr- (approved):

kbr- (commented):
wait, the new version seems to do some new stuff, lemme review again

elcallio:
I changed the impl a tiny bit to requests from  @haaawk

cdc/log.cc
```diff
@@ -944,8 +957,13 @@ class log_mutation_builder {
...
+        if (has_rows()) {
+            _log_mut.set_cell(last_row_key(), log_meta_column_name_bytes("end_of_batch"), data_value(true), _ts, _ttl);
+        }
+    }
 private:
-    void set_key_columns(const clustering_key& log_ck, schema::const_iterator_range_type columns, const std::vector<bytes>& key) {
+    void set_key_columns(const clustering_key& log_ck, schema::const_iterator_range_type columns, const std::vector<bytes>& key) const {
```
kbr-, line 966: But it's not const, it modifies the state.

cdc/log.cc
```diff
@@ -880,12 +881,24 @@ class log_mutation_builder {
         return _base_schema;
     }
 
+    clustering_key create_ck(int batch) const {
+        auto log_ck = clustering_key::from_exploded(_log_schema, { _tuuid, int32_type->decompose(batch) });
+        set_key_columns(log_ck, _base_schema.partition_key_columns(), _base_pk);
```
kbr-, line 886: this is definitely not `const`.
I see what you wanted to do here, leme suggest the following: move the `set_key_columns` call to `allocate_new_log_row`. Then `create_ck` is indeed `const` (it's just a calculation), and `set_key_columns` doesn't need to be `const`, as it should not be.

cdc/log.cc
```diff
@@ -880,12 +881,24 @@ class log_mutation_builder {
...
-        set_key_columns(log_ck, _base_schema.partition_key_columns(), _base_pk);
-        return log_ck;
+        return create_ck(_batch_no++);
+    }
+
+    bool has_rows() const {
+        return _batch_no != 0;
+    }
+
+    clustering_key last_row_key() const {
```
kbr-, line 900: One would presume that this is a pure calculation, but it actually modifies the state now.

cdc/log.cc
elcallio: true.

cdc/log.cc
elcallio: Fixed.

cdc/log.cc
elcallio: Fixed.

kbr- (approved):

haaawk:
@nyh I think the PR is ready for you to review and merge if you find it appropriate. Thanks.

elcallio:
@nyh ping?

PR #7445, opened 2020-10-16
treewide: don't capture structured bindings in lambdas

avikivity:
Clang does not yet implement p1091r3, which allows lambdas
to capture structured bindings. To accomodate it, don't
use structured bindings for variables that are later
captured.

Hopefully, most of these lambda captures will be replaces with
coroutines.

### DIFF
--- a/cql3/statements/select_statement.cc
+++ b/cql3/statements/select_statement.cc
@@ -529,7 +529,9 @@ indexed_table_select_statement::do_execute_base_query(
 
     base_query_state query_state{cmd->get_row_limit() * queried_ranges_count, std::move(ranges_to_vnodes)};
     return do_with(std::move(query_state), [this, &proxy, &state, &options, cmd, timeout] (auto&& query_state) {
-        auto& [merger, ranges_to_vnodes, concurrency] = query_state;
+        auto& merger = query_state.merger;
+        auto& ranges_to_vnodes = query_state.ranges_to_vnodes;
+        auto& concurrency = query_state.concurrency;
         return repeat([this, &ranges_to_vnodes, &merger, &proxy, &state, &options, &concurrency, cmd, timeout]() {
             // Starting with 1 range, we check if the result was a short read, and if not,
             // we continue exponentially, asking for 2x more ranges than before
@@ -881,7 +883,10 @@ lw_shared_ptr<const service::pager::paging_state> indexed_table_select_statement
     if (!results->row_count() || *results->row_count() == 0) {
         return std::move(paging_state);
     }
-    auto [last_base_pk, last_base_ck] = result_view.get_last_partition_and_clustering_key();
+
+    auto&& last_partition_and_clustering_key = result_view.get_last_partition_and_clustering_key();
+    auto& last_base_pk = std::get<0>(last_partition_and_clustering_key);
+    auto& last_base_ck = std::get<1>(last_partition_and_clustering_key);
 
     bytes_opt indexed_column_value = _used_index_restrictions->value_for(*cdef, options);
 
--- a/db/hints/manager.cc
+++ b/db/hints/manager.cc
@@ -759,7 +759,8 @@ bool manager::end_point_hints_manager::sender::send_one_file(const sstring& fnam
 
     try {
         commitlog::read_log_file(fname, manager::FILENAME_PREFIX, service::get_local_streaming_priority(), [this, secs_since_file_mod, &fname, ctx_ptr] (commitlog::buffer_and_replay_position buf_rp) mutable {
-            auto&& [buf, rp] = buf_rp;
+            auto& buf = buf_rp.buffer;
+            auto& rp = buf_rp.position;
             // Check that we can still send the next hint. Don't try to send it if the destination host
             // is DOWN or if we have already failed to send some of the previous hints.
             if (!draining() && ctx_ptr->segment_replay_failed) {
--- a/test/boost/mutation_reader_test.cc
+++ b/test/boost/mutation_reader_test.cc
@@ -2121,7 +2121,7 @@ SEASTAR_THREAD_TEST_CASE(test_foreign_reader_destroyed_with_pending_read_ahead) 
     do_with_cql_env([] (cql_test_env& env) -> future<> {
         const auto shard_of_interest = (this_shard_id() + 1) % smp::count;
         auto s = simple_schema();
-        auto [remote_control, remote_reader] = smp::submit_to(shard_of_interest, [gs = global_simple_schema(s)] {
+        auto remote_control_remote_reader = smp::submit_to(shard_of_interest, [gs = global_simple_schema(s)] {
             using control_type = foreign_ptr<std::unique_ptr<puppet_reader::control>>;
             using reader_type = foreign_ptr<std::unique_ptr<flat_mutation_reader>>;
 
@@ -2133,6 +2133,9 @@ SEASTAR_THREAD_TEST_CASE(test_foreign_reader_destroyed_with_pending_read_ahead) 
 
             return make_ready_future<std::tuple<control_type, reader_type>>(std::tuple(std::move(control), std::move(reader)));
         }).get0();
+
+        auto& remote_control = std::get<0>(remote_control_remote_reader);
+        auto& remote_reader = std::get<1>(remote_control_remote_reader);
 
         {
             auto reader = make_foreign_reader(s.schema(), tests::make_permit(), std::move(remote_reader));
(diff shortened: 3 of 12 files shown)


avikivity:
Note: @roydahan benchmarked clang vs gcc and found no regressions, so I now feel comfortable pushing the uglier clang compatibility patches.

db/hints/manager.cc
```diff
@@ -759,7 +759,8 @@ bool manager::end_point_hints_manager::sender::send_one_file(const sstring& fnam
 
     try {
         commitlog::read_log_file(fname, manager::FILENAME_PREFIX, service::get_local_streaming_priority(), [this, secs_since_file_mod, &fname, ctx_ptr] (commitlog::buffer_and_replay_position buf_rp) mutable {
-            auto&& [buf, rp] = buf_rp;
+            auto& buf = buf_rp.buffer;
+            auto& rp = buf_rp.position;
```
haaawk, line 763: Shouldn't it be auto&& ?

test/manual/partition_data_test.cc
```diff
@@ -61,7 +61,11 @@ BOOST_AUTO_TEST_CASE(test_atomic_cell) {
     };
 
     for (auto tc : cases) {
-        auto [live, fixed_size, value, expiring, counter_update] = tc;
+        auto& live = tc.live;
+        auto& fixed_size = tc.fixed_size;
+        auto& value = tc.value;
+        auto& expiring = tc.expiring;
+        auto& counter_update = tc.counter_update;
```
haaawk, line 68: Shouldn't it be auto instead of auto&?

test/boost/mutation_reader_test.cc
```diff
@@ -2134,6 +2134,9 @@ SEASTAR_THREAD_TEST_CASE(test_foreign_reader_destroyed_with_pending_read_ahead)
             return make_ready_future<std::tuple<control_type, reader_type>>(std::tuple(std::move(control), std::move(reader)));
         }).get0();
 
+        auto& remote_control = std::get<0>(remote_control_remote_reader);
+        auto& remote_reader = std::get<1>(remote_control_remote_reader);
```
haaawk, line 2138: Shouldn't it be auto instead of auto& ?

haaawk:
This change makes me sad.

test/manual/partition_data_test.cc
avikivity: No, why?

db/hints/manager.cc
avikivity: The right hand side is an lvalue, and in auto&& x = lvalue 'auto&&' resolves to 'auto&'.

test/boost/mutation_reader_test.cc
avikivity: No, why?

avikivity:
> This change makes me sad.

Yes. very unfortunate. But many of those sites will disappear with coroutines.

db/hints/manager.cc
haaawk: Right.

test/boost/mutation_reader_test.cc
haaawk: Previously remote_control and remote_reader were lvalues and now they are lvalue references. Maybe it's ok to change it as well. I was just pointing out the place that, I believe, has changed.

test/manual/partition_data_test.cc
haaawk: Again, just pointing out the change in behavior. Variables were lvalues and now are lvalue references. It's probably fine to change it like this.

cql3/statements/select_statement.cc
```diff
@@ -881,7 +883,10 @@ lw_shared_ptr<const service::pager::paging_state> indexed_table_select_statement
     if (!results->row_count() || *results->row_count() == 0) {
         return std::move(paging_state);
     }
-    auto [last_base_pk, last_base_ck] = result_view.get_last_partition_and_clustering_key();
+
+    auto&& last_partition_and_clustering_key = result_view.get_last_partition_and_clustering_key();
+    auto& last_base_pk = std::get<0>(last_partition_and_clustering_key);
+    auto& last_base_ck = std::get<1>(last_partition_and_clustering_key);
```
nyh, line 889: Is the change of "auto" before the patch to "auto&" after the patch, necessary? I don't know how to review it before going back to look at the entire code surrounding it.

test/boost/mutation_test.cc
```diff
@@ -1944,7 +1944,9 @@ SEASTAR_THREAD_TEST_CASE(test_external_memory_usage) {
     };
 
     for (auto i = 0; i < 16; i++) {
-        auto [ m, size ] = generate();
+        auto m_and_size = generate();
+        auto&& m = m_and_size.first;
+        auto&& size = m_and_size.second;
```
nyh, line 1949: How is "auto&&" in this case different from just "auto&"?

test/boost/querier_cache_test.cc
```diff
@@ -213,8 +213,10 @@ class test_querier_cache {
         const auto cache_key = make_cache_key(key);
 
         auto querier = make_querier<Querier>(range);
-        auto [dk, ck] = querier.consume_page(dummy_result_builder{}, row_limit, std::numeric_limits<uint32_t>::max(),
+        auto dk_ck = querier.consume_page(dummy_result_builder{}, row_limit, std::numeric_limits<uint32_t>::max(),
                 gc_clock::now(), db::no_timeout, query::max_result_size(std::numeric_limits<uint64_t>::max())).get0();
+        auto&& dk = dk_ck.first;
+        auto&& ck = dk_ck.second;
```
nyh, line 219: Again I'm a bit confused about this idiom. Isn't it more natural to do something like
```
auto&& dc_ck = some_function();    // extend temporary lifetime
auto& dk = ...; // pointer to some part of that temporary
auto& ck = ...;
```

test/manual/partition_data_test.cc
nyh: Yes, these sort of changes bothered me too. I have no practical way to review this change without trying to read the rest of the file and what is actually done with these references - which I assume you (Avi) did, but I didn't.

thrift/server.cc
```diff
@@ -232,7 +232,8 @@ thrift_server::do_accepts(int which, bool keepalive) {
     // Future is waited on indirectly in `stop()` (via `_stop_gate`).
     (void)with_gate(_stop_gate, [&, this] {
         return _listeners[which].accept().then([this, which, keepalive] (accept_result ar) {
-            auto&& [fd, addr] = ar;
+            auto&& fd = ar.connection;
+            auto&& addr = ar.remote_address;
```
nyh, line 236: The more of these changes I read, the more confused I get... What does it mean to put an lvalue in an "auto&&"? Can't it just be written as a "auto&"?

nyh (approved):
Looks mostly good, but at many places, I'm not sure how decide whether to use "auto", "auto&" or "auto&&'. In some places it seems like you are slightly changing the previous behavior - and I hope you reviewed such changes carefully (I didn't). In other places, it seemed to me you used "auto&&" where just "auto&" would have said the same thing, and so on. So I left a few comments. But your code is probably correct anyway so I'm calling this an "Approve".

test/manual/partition_data_test.cc
avikivity: The variables were actually structured bindings, which isn't the same as a free standing variable (this is why it couldn't be captured in a lambda). A reference is the closest thing to a structured binding. Note also the container of the data is in a regular variable, so should have the same lifetime.

test/boost/mutation_reader_test.cc
avikivity: The were not variables, they were structured bindings.

`remote_control_remote_reader` is a variable in the same scope, so the references don't change the lifetime.

cql3/statements/select_statement.cc
avikivity: It's not necessary. As far as I can tell there is no difference in behavior. `auto&&` resolves to `auto`.

test/boost/mutation_test.cc
avikivity: It's not different.

test/boost/querier_cache_test.cc
avikivity: It's the same thing. I prefer `auto&&` too in general.

thrift/server.cc
avikivity: Yes, it can be written as `auto&`.

PR #7449, opened 2020-10-17
utf8: don't linearize cells for validation

### LINKED ISSUE #7448: utf8 validation of large cells causes latency spikes
When validating utf8 values, we first linearize them. This causes latency spikes if they are large.

avikivity:
Currently, we linearize large UTF8 cells in order to validate them. This can cause large latency spikes if the cell is large.

This series changes UTF8 validation to work on fragmented buffers. This is somewhat tricky since the validation routines are optimized for single-instruction-multiple-data (SIMD) architectures.

The unit tests are expanded to cover the new functionality.

Fixes #7448.

### DIFF
--- a/utils/utf8.cc
+++ b/utils/utf8.cc
@@ -56,21 +56,29 @@ namespace utils {
 
 namespace utf8 {
 
-// 3x faster than boost utf_to_utf
-static inline std::optional<size_t> validate_naive(const uint8_t *data, size_t len) {
-    size_t pos = 0;
+struct codepoint_status {
+    size_t bytes_validated;
+    bool error;
+    uint8_t more_bytes_needed;
+};
 
-    while (len) {
-        size_t bytes;
-        const uint8_t byte1 = data[0];
-
+static
+codepoint_status
+inline
+evaluate_codepoint(const uint8_t* data, size_t len) {
+    const uint8_t byte1 = data[0];
+    static const uint8_t len_from_first_nibble[16] = { 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 3, 4 };
+    auto codepoint_len = len_from_first_nibble[byte1 >> 4];
+    if (codepoint_len > len) {
+        return codepoint_status{.more_bytes_needed = uint8_t(codepoint_len - len)};
+    } else {
         if (byte1 <= 0x7F) {
             // 00..7F
-            bytes = 1;
+            return codepoint_status{.bytes_validated = codepoint_len};
         } else if (len >= 2 && byte1 >= 0xC2 && byte1 <= 0xDF &&
                 (int8_t)data[1] <= (int8_t)0xBF) {
             // C2..DF, 80..BF
-            bytes = 2;
+            return codepoint_status{.bytes_validated = codepoint_len};
         } else if (len >= 3) {
             const uint8_t byte2 = data[1];
 
@@ -87,7 +95,7 @@ static inline std::optional<size_t> validate_naive(const uint8_t *data, size_t l
                      (byte1 == 0xED && byte2 <= 0x9F) ||
                      // EE..EF, 80..BF, 80..BF
                      (byte1 >= 0xEE && byte1 <= 0xEF))) {
-                bytes = 3;
+                return codepoint_status{.bytes_validated = codepoint_len};
             } else if (len >= 4) {
                 // Is byte4 between 0x80 ~ 0xBF
                 const int byte4_ok = (int8_t)data[3] <= (int8_t)0xBF;
@@ -99,20 +107,31 @@ static inline std::optional<size_t> validate_naive(const uint8_t *data, size_t l
                          (byte1 >= 0xF1 && byte1 <= 0xF3) ||
                          // F4, 80..8F, 80..BF, 80..BF
                          (byte1 == 0xF4 && byte2 <= 0x8F))) {
-                    bytes = 4;
+                    return codepoint_status{.bytes_validated = codepoint_len};
                 } else {
-                    return pos;
+                    return codepoint_status{.error = true};
                 }
             } else {
-                return pos;
+                return codepoint_status{.error = true};
             }
         } else {
+            return codepoint_status{.error = true};
+        }
+    }
+}
+
+// 3x faster than boost utf_to_utf
+static inline std::optional<size_t> validate_naive(const uint8_t *data, size_t len) {
+    size_t pos = 0;
+
+    while (len) {
+        auto cs = evaluate_codepoint(data, len);
+        pos += cs.bytes_validated;
+        data += cs.bytes_validated;
+        len -= cs.bytes_validated;
+        if (cs.error || cs.more_bytes_needed) {
             return pos;
         }
-
-        len -= bytes;
-        pos += bytes;
-        data += bytes;
     }
 
     return std::nullopt;
--- a/utils/utf8.cc
+++ b/utils/utf8.cc
@@ -56,6 +56,8 @@ namespace utils {
 
 namespace utf8 {
 
+using namespace internal;
+
 struct codepoint_status {
     size_t bytes_validated;
     bool error;
@@ -136,12 +138,6 @@ static inline std::optional<size_t> validate_naive(const uint8_t *data, size_t l
 
     return std::nullopt;
 }
-
-struct partial_validation_results {
-    bool error;
-    size_t unvalidated_tail;
-    size_t bytes_needed_for_tail;
-};
 
 static
 partial_validation_results
@@ -235,9 +231,8 @@ alignas(16) static const uint8_t s_range_adjust_tbl[] = {
 };
 
 // 2x ~ 4x faster than naive method
-static
 partial_validation_results
-validate_partial(const uint8_t *data, size_t len) {
+internal::validate_partial(const uint8_t *data, size_t len) {
     if (len >= 16) {
         uint8x16_t prev_input = vdupq_n_u8(0);
         uint8x16_t prev_first_len = vdupq_n_u8(0);
@@ -415,9 +410,8 @@ alignas(16) static const int8_t s_ef_fe_tbl[] = {
 };
 
 // 5x faster than naive method
-static
 partial_validation_results
-validate_partial(const uint8_t *data, size_t len) {
+internal::validate_partial(const uint8_t *data, size_t len) {
     if (len >= 16) {
         __m128i prev_input = _mm_set1_epi8(0);
         __m128i prev_first_len = _mm_set1_epi8(0);
@@ -545,7 +539,6 @@ validate_partial(const uint8_t *data, size_t len) {
 
 #else
 // No SIMD implementation for this arch, fallback to naive method
-static
 partial_validation_results
 validate_partial(const uint8_t *data, size_t len) {
     return validate_partial_naive(data, len);
--- a/utils/utf8.cc
+++ b/utils/utf8.cc
@@ -137,6 +137,29 @@ static inline std::optional<size_t> validate_naive(const uint8_t *data, size_t l
     return std::nullopt;
 }
 
+struct partial_validation_results {
+    bool error;
+    size_t unvalidated_tail;
+    size_t bytes_needed_for_tail;
+};
+
+static
+partial_validation_results
+validate_partial_naive(const uint8_t *data, size_t len) {
+    while (len) {
+        auto cs = evaluate_codepoint(data, len);
+        data += cs.bytes_validated;
+        len -= cs.bytes_validated;
+        if (cs.error) {
+            return partial_validation_results{.error = true};
+        }
+        if (cs.more_bytes_needed) {
+            return partial_validation_results{.unvalidated_tail = len, .bytes_needed_for_tail = cs.more_bytes_needed};
+        }
+    }
+    return partial_validation_results{};
+}
+
 #if defined(__aarch64__)
 #include <arm_neon.h>
 
@@ -212,7 +235,9 @@ alignas(16) static const uint8_t s_range_adjust_tbl[] = {
 };
 
 // 2x ~ 4x faster than naive method
-bool validate(const uint8_t *data, size_t len) {
+static
+partial_validation_results
+validate_partial(const uint8_t *data, size_t len) {
     if (len >= 16) {
         uint8x16_t prev_input = vdupq_n_u8(0);
         uint8x16_t prev_first_len = vdupq_n_u8(0);
@@ -308,7 +333,7 @@ bool validate(const uint8_t *data, size_t len) {
 
         // Delay error check till loop ends
         if (vmaxvq_u8(error)) {
-            return false;
+            return partial_validation_results{.error = true};
         }
 
         // Find previous token (not 80~BF)
@@ -328,8 +353,8 @@ bool validate(const uint8_t *data, size_t len) {
         len += lookahead;
     }
 
-    // Check for no error position in remaining bytes with naive method
-    return !validate_naive(data, len);
+    // Continue with remaining bytes with naive method
+    return validate_partial_naive(data, len);
 }
 
 #elif defined(__x86_64__)
@@ -390,7 +415,9 @@ alignas(16) static const int8_t s_ef_fe_tbl[] = {
 };
 
 // 5x faster than naive method
-bool validate(const uint8_t *data, size_t len) {
+static
+partial_validation_results
+validate_partial(const uint8_t *data, size_t len) {
     if (len >= 16) {
         __m128i prev_input = _mm_set1_epi8(0);
         __m128i prev_first_len = _mm_set1_epi8(0);
@@ -494,7 +521,7 @@ bool validate(const uint8_t *data, size_t len) {
         int error_reduced =
             _mm_movemask_epi8(_mm_cmpeq_epi8(error, _mm_set1_epi8(0)));
         if (error_reduced != 0xFFFF) {
-            return false;
+            return partial_validation_results{.error = true};
         }
 
         // Find previous token (not 80~BF)
@@ -512,17 +539,23 @@ bool validate(const uint8_t *data, size_t len) {
         len += lookahead;
     }
 
-    // Check for no error position in remaining bytes with naive method
-    return !validate_naive(data, len);
+    // Continue with remaining bytes with naive method
+    return validate_partial_naive(data, len);
 }
 
 #else
 // No SIMD implementation for this arch, fallback to naive method
-bool validate(const uint8_t *data, size_t len) {
-    // Check for no error position
-    return !validate_naive(data, len);
+static
+partial_validation_results
+validate_partial(const uint8_t *data, size_t len) {
+    return validate_partial_naive(data, len);
 }
 #endif
+
+bool validate(const uint8_t* data, size_t len) {
+    auto pvr = validate_partial(data, len);
+    return !pvr.error && !pvr.unvalidated_tail;
+}
 
 std::optional<size_t> validate_with_error_position(const uint8_t *data, size_t len) {
     // First pass - validate data (using optimized code)
--- a/utils/utf8.hh
+++ b/utils/utf8.hh
@@ -26,6 +26,7 @@
 
 #include <cstdint>
 #include "bytes.hh"
+#include "fragment_range.hh"
 
 namespace utils {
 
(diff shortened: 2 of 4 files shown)


utils/utf8.cc
```diff
@@ -56,22 +56,30 @@ namespace utils {
 
 namespace utf8 {
```
haaawk, line 58: Can't comment on commit message so posting here:
s/that **is** spans multiple buffers/that spans multiple buffers

utils/utf8.cc
```diff
@@ -56,22 +56,30 @@ namespace utils {
...
+    bool error;
+    uint8_t more_bytes_needed;
+};
 
+static
+codepoint_status
+inline
+evaluate_codepoint(const uint8_t* data, size_t len) {
+    auto byte1 = data[0];
+    static uint8_t len_from_first_nibble[16] = { 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 3, 4 };
```
haaawk, line 72: std::array? and const?

utils/utf8.cc
```diff
@@ -56,22 +56,30 @@ namespace utils {
...
+    size_t bytes_validated;
+    bool error;
+    uint8_t more_bytes_needed;
+};
 
+static
+codepoint_status
+inline
+evaluate_codepoint(const uint8_t* data, size_t len) {
+    auto byte1 = data[0];
```
haaawk, line 71: const?

utils/utf8.cc
```diff
@@ -56,22 +56,30 @@ namespace utils {
...
+    if (codepoint_len > len) {
+	return codepoint_status{.more_bytes_needed = uint8_t(codepoint_len - len)};
+    } else {
         if (byte1 <= 0x7F) {
             // 00..7F
-            bytes = 1;
-        } else if (len >= 2 && byte1 >= 0xC2 && byte1 <= 0xDF &&
+            return codepoint_status{.bytes_validated = codepoint_len};
+        } else if (byte1 >= 0xC2 && byte1 <= 0xDF &&
                 (int8_t)data[1] <= (int8_t)0xBF) {
```
haaawk, line 81: Shouldn't we also check (int8_t)data[1] >= 0x80?

utils/utf8.cc
```diff
@@ -56,22 +56,30 @@ namespace utils {
...
             // C2..DF, 80..BF
-            bytes = 2;
-        } else if (len >= 3) {
+            return codepoint_status{.bytes_validated = codepoint_len};
+        } else {
             const uint8_t byte2 = data[1];
 
             // Is byte2, byte3 between 0x80 ~ 0xBF
```
haaawk, line 87: This comment says that we check bytes are within [0x80, 0xBF] but the check below compares only against 0xBF. Shouldn't we also compare against 0x80?

utils/utf8.cc
```diff
@@ -87,8 +95,9 @@ static inline std::optional<size_t> validate_naive(const uint8_t *data, size_t l
                      (byte1 == 0xED && byte2 <= 0x9F) ||
                      // EE..EF, 80..BF, 80..BF
                      (byte1 >= 0xEE && byte1 <= 0xEF))) {
-                bytes = 3;
-            } else if (len >= 4) {
+                return codepoint_status{.bytes_validated = codepoint_len};
+            }
+            {
                 // Is byte4 between 0x80 ~ 0xBF
```
haaawk, line 103: Again, compares only against 0xBF. 0x80 is never checked.

utils/utf8.cc
```diff
@@ -87,8 +95,9 @@ static inline std::optional<size_t> validate_naive(const uint8_t *data, size_t l
                      (byte1 == 0xED && byte2 <= 0x9F) ||
                      // EE..EF, 80..BF, 80..BF
                      (byte1 >= 0xEE && byte1 <= 0xEF))) {
-                bytes = 3;
-            } else if (len >= 4) {
+                return codepoint_status{.bytes_validated = codepoint_len};
```
haaawk, line 100: I have to admit, I don't understand those checks. For example `(byte1 == 0xE0 && byte2 >= 0xA0)` why is `byte2 >= 0xA0` needed? I would understand `byte2 >= 0x80` but have no idea what `0xA0` is.
Same with the comparison of byte1 against `0xE1`, `0xED`, and `0xEE`. I thought that meaningful values of byte1 are `0x00 - 0x7F`, `0xC0-0xDF`, `0xE0-0xEF`, and `0xF0-0xF7`?

utils/utf8.hh
```diff
@@ -64,6 +65,54 @@ inline std::optional<size_t> validate_with_error_position(bytes_view string) {
     return validate_with_error_position(data, len);	
 }
 
+std::optional<size_t> validate_with_error_position_fragmented(const FragmentRange auto& fr) {
+    uint8_t partial_codepoint[4];
```
haaawk, line 69: std::array?

utils/utf8.hh
```diff
@@ -64,6 +65,54 @@ inline std::optional<size_t> validate_with_error_position(bytes_view string) {
...
+                // We accumulated a codepoint that straddled two oe more fragments,
+                // validate it now.
+                auto pvr = internal::validate_partial(partial_codepoint, partial_filled);
+                if (pvr.error) {
+                    return {bytes_validated};
+                }
+                bytes_validated += partial_filled;
+                partial_filled = partial_more_needed = 0;
+            }
+            if (!len) {
```
haaawk, line 93: Super nit: it is guaranteed that after a while loop above either partition_more_needed == 0 or len == 0 so here you can just do else and inside this else do assert (!len).

haaawk: Ok actually no. It is also possible for both partial_more_needed and len to be zero so `else` wouldn't work.

haaawk (approved):
LGTM I just have a couple of questions about the first commit. Not really related to this change.

utils/utf8.cc
nyh: I don't understand this either. For 3 bytes, byte0 should have binary 1110..., in other words between 0xE0 and 0xEF. What's all these special cases for 0xED and 0xEE?
All this code seems to me like maybe we're overthinking things... IIRC, in UTF-8, the first byte determines the length (that's the codepoint_len thing), but then - the rest of the bytes just need to begin with bits 10, i.e., between 0x80 and 0xBF.  That's it. Why do we have any other case?

utils/utf8.cc
```diff
@@ -56,22 +56,30 @@ namespace utils {
...
 
+static
+codepoint_status
+inline
+evaluate_codepoint(const uint8_t* data, size_t len) {
+    auto byte1 = data[0];
+    static uint8_t len_from_first_nibble[16] = { 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 3, 4 };
+    auto codepoint_len = len_from_first_nibble[byte1 >> 4];
+    if (codepoint_len > len) {
+	return codepoint_status{.more_bytes_needed = uint8_t(codepoint_len - len)};
```
nyh, line 75: I think there's a bug here. Even if codepoint_len > len, you still have "len" to check. I think you need to check that all the bytes that you do have after the first one are in the range 0x80 and 0xBF, and if not, set "error".
In fact, that's *all* you need to do - it seems to me most of the various cases below are not really needed (more on this below), and moreover, the whole idea of remembering the end of the buffer is not needed. See more on that too.

utils/utf8.cc
```diff
@@ -99,20 +108,26 @@ static inline std::optional<size_t> validate_naive(const uint8_t *data, size_t l
...
+static inline std::optional<size_t> validate_naive(const uint8_t *data, size_t len) {
+    size_t pos = 0;
+
+    while (len) {
+        auto cs = evaluate_codepoint(data, len);
+        pos += cs.bytes_validated;
+        data += cs.bytes_validated;
+        len -= cs.bytes_validated;
+        if (cs.error || cs.more_bytes_needed) {
+            return pos;
```
nyh, line 131: I didn't understand what you're trying to do here or why.
It seems to me you are trying to return the end of the last full UTF-8 codepoint, perhaps so we can take the last few bytes plus a few bytes from the next buffer to create the last codepoint. But I think that's not needed! The only thing we need to check is that the next "more_bytes_needed" bytes in the next buffer are between 0x80 and 0xBF. That's it! We don't need to remember what the old buffer contained! We just need to remember more_bytes_needed and that's it.

nyh (changes requested):
I think the validator was missing some important checks and didn't understand other checks it did - see my review. If any of the potential bugs I raised are true, it also means the tests missed them so they aren't exhaustive enough.
Moreover, I think the fragmented utf8-checking should actually be simpler than the way you did it. If a UTF-8 "code point" is split between buffers, there is no reason to merge the two halfs. The first half includes the first byte - which indicates the code's **length**, and all the rest of the bytes simply need to begin by bits 10 - i.e., be between 0x80 and 0xBF. In other words, in the second buffer, we just need to check that the first M (calculated in the previous buffer) bytes are between 0x80 and 0xBF - there is no need to keep any byte data from the previous buffer.

utils/utf8.cc
```diff
@@ -56,22 +56,30 @@ namespace utils {
...
             // 00..7F
-            bytes = 1;
-        } else if (len >= 2 && byte1 >= 0xC2 && byte1 <= 0xDF &&
+            return codepoint_status{.bytes_validated = codepoint_len};
+        } else if (byte1 >= 0xC2 && byte1 <= 0xDF &&
                 (int8_t)data[1] <= (int8_t)0xBF) {
             // C2..DF, 80..BF
-            bytes = 2;
-        } else if (len >= 3) {
+            return codepoint_status{.bytes_validated = codepoint_len};
```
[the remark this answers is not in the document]
haaawk: That's my understanding too.

utils/utf8.cc
avikivity: fixed

utils/utf8.cc
avikivity: std::array gives nothing here IMO. It's useful to wrap an array in a regular object without special behavior when passing it around, especially to templates. Here it's just obfuscation.

Added const.

utils/utf8.cc
avikivity: Why? const it useful on interfaces, it tells the reader of the interface that the callee does not change the value.

How does it help here?

utils/utf8.cc
avikivity: There is no (int8_t)data[1] that can be < (int8_t)0x80 - that's the minimum int8_t value. A nice trick by the original author.

In any case it is not related to this refactoting.

utils/utf8.cc
avikivity: All int8_t values are >= 0x80, that is the minimum value.

utils/utf8.cc
avikivity: Again, the codeeis tricky but correct.

utils/utf8.cc
avikivity: First, these comments don't concern the refactoring. I did not change the logic, just the interfaces.

Second, the code is very careful and correct. While the first byte determines the length, the following bytes have to follow a distinct pattern, and if they don't, the sequence is invalid. They don't just have to fit the range 0x80-0xbf.

I recommend you study the code, not the changes, if you want to understand the validation logic. My series does not change validation in any way (at least not intentionally), it just makes it work for scattered buffers.

utils/utf8.hh
avikivity: why?

utils/utf8.hh
avikivity: Correct.

utils/utf8.cc
avikivity: > I think the validator was missing some important checks and didn't understand other checks it did - see my review.

This series does not introduce a validator, so comments about the validator do not apply.

> If any of the potential bugs I raised are true, it also means the tests missed them so they aren't exhaustive enough.
> Moreover, I think the fragmented utf8-checking should actually be simpler than the way you did it. If a UTF-8 "code point" is split between buffers, there is no reason to merge the two halfs. The first half includes the first byte - which indicates the code's **length**, and all the rest of the bytes simply need to begin by bits 10 - i.e., be between 0x80 and 0xBF. 

That's not true. 

Edit: actually it is true. So maybe it can be simplified.

>In other words, in the second buffer, we just need to check that the first M (calculated in the previous buffer) bytes are between 0x80 and 0xBF - there is no need to keep any byte data from the previous buffer.

utils/utf8.cc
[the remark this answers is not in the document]
avikivity: Yes, I used the variable so the compiler can consolidate all the return statements.

utils/utf8.cc
avikivity: Ah, I think it's protecting against overlong sequences.

utils/utf8.cc
avikivity: No. UTF8 forbids overlong sequences,  0x80-0xbf is too weak.

utils/utf8.cc
```diff
@@ -56,22 +56,30 @@ namespace utils {
...
+    static uint8_t len_from_first_nibble[16] = { 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 3, 4 };
+    auto codepoint_len = len_from_first_nibble[byte1 >> 4];
+    if (codepoint_len > len) {
+	return codepoint_status{.more_bytes_needed = uint8_t(codepoint_len - len)};
+    } else {
         if (byte1 <= 0x7F) {
             // 00..7F
-            bytes = 1;
-        } else if (len >= 2 && byte1 >= 0xC2 && byte1 <= 0xDF &&
+            return codepoint_status{.bytes_validated = codepoint_len};
```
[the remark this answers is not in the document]
avikivity: This is only called on buffer edges. The main buffer is evaluated using SSE code.

Note the original author updated the code, and there are now contributed AVX2 implementations. If we want to optimize, we should merge from there.

avikivity:
Updated:
 - fixed grammar in changelog
 - added const to len_from_first_nibble array

I did not make any of the simplifications suggested. First, they are wrong, the extra checks are needed to prevent forbidden overlong sequences. Second, the focus of this series is to remove linearization from the validation path, not to optimize the validation code.

utils/utf8.cc
avikivity: Edit2: it's not true.The sequence f0 80 80 80 codes for codepoint 0, but is illegal.

haaawk: I was under the impression that in modern C++ std::array should always replace old arrays when the size is known. Isn't std::array safer?

utils/utf8.cc
haaawk: when a variable is const a reader knows it will never change. When it's not const a reader has to read the whole code to know.
If something is const like byte1 in our case why not mark it as such?

utils/utf8.cc
haaawk: Nice trick indeed.

utils/utf8.cc
haaawk: It's not that all int8_t values are >= 0x80. It's more that every value that's < 0x80 in uint8_t is bigger than 0xBF in int8_t. Otherwise, the check wouldn't work.

But you're right everything's fine.

utils/utf8.cc
haaawk: True.

utils/utf8.hh
haaawk: better safety and more idiomatic for modern c++

utils/utf8.cc
avikivity: No, it just wraps a regular C array.

It's safer when you pass it around, since C arrays have strange rules where they decay into a pointer and lose information about the size.

utils/utf8.cc
avikivity: Ok, I'll add const. I think it's noise in small functions, but if it helps someone, it's worth it.

utils/utf8.cc
avikivity: godbolt thinks it is:

   https://godbolt.org/z/9rKah7

0xbf checks the other side of the interval. The trick reduced a closed interval into a half-open interval.

avikivity:
Updated:
 - made byte1 in evaluate_codepoint() const uint8_t instead of auto

utils/utf8.cc
haaawk: Compiler optimization is understandable. 0x80 is the minimal value of int8_t so there's no point checking that something is smaller than 0x80. But it's all about the conversion from uint8_t to int8_t. Values that are < 0x80 in uint8_t after conversion to int8_t become > than 0xBF and that's why this trick works. If values that are < 0x80 in uint8_t after conversion to int8_t were taking values smaller than 0xBF then the trick wouldn't work. The `<= 0xBF` check serves two purposes:
1. eliminate all values that in uint8_t are bigger than 0xBF because they are still bigger than 0xBF after conversion to int8_t
2. eliminate all values that in uint8_t are smaller than 0x80 because after conversion to int8_t they become bigger than 0xBF

What I'm saying is that the fact that 0x80 is a minimal value for int8_t does not make the trick correct. What does is the way uint8_t is converted to int8_t. It's based on the fact that the conversion turns every x < 0x80 into y > 0xBF. If it wasn't true the trick wouldn't work even with 0x80 being the minimal value of int8_t.

utils/utf8.cc
haaawk: Fair enough. It's not safer. It still seems more idiomatic in modern c++. Because why would anyone use C arrays when there is std::array and std::vector?

nyh: > Edit2: it's not true.The sequence f0 80 80 80 codes for codepoint 0, but is illegal.

I didn't think about this. But not all is lost. Basically, leading zeros are not allowed. So it's not enough to remember remaining_bytes for the current codepoint, we also need to remember is_second_byte which means whether the next byte to be read is the first one between 0x80 and 0xBF - and if it is the first one, it must not be 0x80.

Anyway, I see now that https://tools.ietf.org/html/rfc3629 has in "Syntax of UTF-8 Byte Sequences" the same weird ranges that I don't understand. Maybe I'm missing something important, but I don't understand yet what :-(

utils/utf8.cc
nyh:
```
const int byte2_ok = (int8_t)byte2 <= (int8_t)0xBF;
```
Wow, what a trick... Basically, 0xBF is a negative number, so for byte2 to be smaller than it, it needs to be negative too (i.e., the unsigned version is >= 0x80), and must also be <= 0xBF. This should have been in a separate function with clear comments explaining how/why it works. I agree it's pre-existing code, so you don't *have* to fix it too, but maybe it's a good opportunity since we understand it now...

utils/utf8.cc
nyh: Oh, I *think* maybe I understand now... If you have a 4-byte sequence, and only got 3, you don't check these 3 at all. You just collect them to later, and work on the 4 together, later. Is that true? If so, than the fact the code returns more_bytes_needed immediately, without checking anything, is correct, and there is no bug.

utils/utf8.cc
nyh: I see. I didn't understand this earlier. I am still not sure how such "overlong" sequences can be caused by anything except leading zeros, i.e., 0x80, although I see now https://tools.ietf.org/html/rfc3629#section-10 has another example, and I guess this code was carefully written with that in mind, we we can't try to "improve" or "simplify" it as I have been suggesting to do :-(

nyh (approved):
Ok, I withdraw all my requests and hopes for "simplification" of this code, due to https://tools.ietf.org/html/rfc3629#section-10 :-(

utils/utf8.cc
haaawk: I agree it's very hacky. Hard to spot.

utils/utf8.cc
avikivity: Yes, but I don't want to invent a new mechanism for validating the buffer-straddling codepoint. There's no real benefit, and there is a risk of getting the validation wrong.

utils/utf8.cc
avikivity: Yes, checking is performed when we have a full codepoint.

utils/utf8.hh
```diff
@@ -64,6 +65,54 @@ inline std::optional<size_t> validate_with_error_position(bytes_view string) {
...
+        auto len = frag.size();
+        if (partial_more_needed) {
+            // Tiny loop (often zero iterations), don't call memcpy
+            while (partial_more_needed && len) {
+                partial_codepoint[partial_filled++] = *data++;
+                --partial_more_needed;
+                --len;
+            }
+            if (!partial_more_needed) {
+                // We accumulated a codepoint that straddled two oe more fragments,
```
denesb, line 84: s/oe/or/

denesb (approved):

utils/utf8.cc
avikivity: The same trick is used in the SIMD code, where it's not possible to extract.

utils/utf8.hh
avikivity: Fixed.

avikivity:
Updated: fixed a typo in a comment.

avikivity:
I dequeued it, some debug mode test fails.

avikivity:
Updated:
 - restored patch 1's checks for len. Checking codepoint_len isn't sufficient, because for invalid first bytes have codepoint_len == 1, but the code structure still reads bytes 2..4.
 - rearranged patch 1 to have a smaller diff, so it can more easily be verified
 - replaced a tab that snuck in with spaces

PR #7452, opened 2020-10-19
repair: Add strict mode for removenode ops

### LINKED ISSUE #7359: Repair-based streaming is "best effort" which leads to linearizability violation during nodetool removenode
Recently Scylla introduced repair-based streaming - a feature which allows to stream a token range from all available replicas, thus increasing consistency of the new node during streaming. However, repair-based streaming does not insist on using at least a QUORUM of replicas, which leads to lost data if streaming is accompanied with temporary or permanent network failure.

How to repeat:

1. Create a cluster of 5 nodes.
2. Create a table with replication factor = 3
3. Stop or partition away node 1.
4. Insert 100 records into the table, ensuring all nodes store data ranges using QUORUM or LWT.
5. Start the stopped node or resolve the network partitioning.
6. Kill and remove with nodetool removenode node 2 concurrently with stopping or partioning away node 3. This leads to node 1 being used exclusively to stream some ranges, thus replicating missing ranges to a quorum of nodes.
7. Observe removenode passes successfully, even though some token ranges have lost quorum. The linearizability is permanently violated, since once removenode completes the troubled ranges will begin accepting new QUORUM or LWT writes, based on empty history, which corrupts data even if node 2 later rejoins the cluster.
8. Read data with SERIAL or QUORUM and observe linearizability violation. Some keys are missing.

Suggested fix:
- nodetool removenode should block or retry until networking partition heals and not continue in absence of quorum.
- the current behavior should be moved behind a switch, something like --best-effort

Please note that once we begin streaming from a quorum, we do not need the concept of "mandatory" node,
i.e. removenode becomes less different from decommission, and https://github.com/scylladb/scylla/issues/7087 could be fixed by reverting the change that broke it.

asias:
Currently, nodetool removenode operation runs in best effort mode which
means if a node that owns the range after the removenode operation is
down during the operation, the removenode node operation will continue
to succeed without requiring that node to perform data syncing. This can
cause data consistency issues. For example,

Five nodes in the cluster, RF = 3, for a range, n1, n2, n3 is the old
replicas, n2 is being removed, after the removenode operation, the new
replicas are n1, n5, n3. If n3 is down during the removenode operation,
only n1 will be used to sync data with the new owner n5. This will break
QUORUM read consistency if n1 happens to miss some writes.

To make removenode safe by default, we should fail the removenode
operation if the operation is not safe. If users want to continue the
unsafe removenode anyway, a best_effort_to_removenode option can be used
to switch to old best effort mode.

Fixes #7359


asias:
@kostja FYI

nyh (commented):
Looks mostly good, I just left a few minor comments, but I do have a question about the premise of this patch: You implemented the "best effort" option as a configuration. Who would ever configure this option? It sounds like it should have been a nodetool parameter, that you decide to use on-the-spot when initiating the nodetool command - not something you plan in advance while configuring your cluster.

service/storage_service.cc
```diff
@@ -2835,8 +2858,9 @@ future<> storage_service::keyspace_changed(const sstring& ks_name) {
 }
 
 void storage_service::init_messaging_service() {
-    _messaging.local().register_replication_finished([] (gms::inet_address from) {
-        return get_local_storage_service().confirm_replication(from);
+    _messaging.local().register_replication_finished([] (gms::inet_address from, rpc::optional<bool> failed_opt) {
+        bool failed = failed_opt ? *failed_opt: false;
```
bhalevy, line 2862: this is `failed_opt.value_or(false)`

slivne (commented):
best_effort_to_removenode is a config option -

We need to make this a flag controlled from nodetool / rest api - not a config option

service/storage_service.cc
```diff
@@ -2197,7 +2198,12 @@ future<> storage_service::removenode(sstring host_id_string) {
                     if (ss._gossiper.is_alive(ep)) {
                         ss._replicating_nodes.emplace(ep);
                     } else {
-                        slogger.warn("Endpoint {} is down and will not receive data for re-replication of {}", ep, endpoint);
+                        if (best_effort) {
+                            slogger.warn("Endpoint {} is down and will not receive data for re-replication of {}", ep, endpoint);
+                        } else {
+                            throw std::runtime_error(format("New owner {} for keyspace {} range {} is not alive for removenode operation to remove node {}",
```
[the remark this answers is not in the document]
asias: Say node X owns range R. Node X is removed. Some of existing nodes need to be the "new owner" for the ranges R. Here we check those "new owner" are alive. Not check if the new owners are able to retrieve data from some other nodes.

service/storage_service.cc
```diff
@@ -2498,13 +2515,15 @@ future<> storage_service::restore_replica_count(inet_address endpoint, inet_addr
         streamer->add_rx_ranges(keyspace_name, std::move(ranges_per_endpoint));
     }
     streamer->stream_async().then_wrapped([this, streamer, notify_endpoint] (auto&& f) {
+        bool failed = false;
         try {
             f.get();
-            return this->send_replication_notification(notify_endpoint);
+            return this->send_replication_notification(notify_endpoint, failed);
```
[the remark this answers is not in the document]
asias: To make it is easier to know what the true or false stands for.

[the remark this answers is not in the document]
bhalevy: Maybe we need to define a `bool_class failed`.

asias:
> best_effort_to_removenode is a config option -
> 
> We need to make this a flag controlled from nodetool / rest api - not a config option

We do not have a good way to communicate this option from the removenode coordinator to rest of the nodes that needs to perform the operation.

avikivity:
I don't understand, isn't the safe version of removenode called `nodetool decommission`?

If the node is up, why would one use removenode?

nyh:
> I don't understand, isn't the safe version of removenode called `nodetool decommission`?
> 
> If the node is up, why would one use removenode?

I don't think the point is whether the node is up or not. The question is what happens if an *additional* node is down. In that case, if I understood correctly today the removenode operation will succeed with loss of consistency - and we want it to fail (the operator should bring the *additional* node back up before removing the first node).

avikivity:
> > I don't understand, isn't the safe version of removenode called `nodetool decommission`?
> > If the node is up, why would one use removenode?
> 
> I don't think the point is whether the node is up or not. The question is what happens if an _additional_ node is down. In that case, if I understood correctly today the removenode operation will succeed with loss of consistency - and we want it to fail (the operator should bring the _additional_ node back up before removing the first node).

Ah, it makes sense now.

db/config.cc
```diff
@@ -686,6 +686,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
     , replace_address_first_boot(this, "replace_address_first_boot", value_status::Used, "", "Like replace_address option, but if the node has been bootstrapped successfully it will be ignored. Same as -Dcassandra.replace_address_first_boot.")
     , override_decommission(this, "override_decommission", value_status::Used, false, "Set true to force a decommissioned node to join the cluster")
     , enable_repair_based_node_ops(this, "enable_repair_based_node_ops", liveness::LiveUpdate, value_status::Used, true, "Set true to use enable repair based node operations instead of streaming based")
+    , best_effort_to_removenode(this, "best_effort_to_removenode", liveness::LiveUpdate, value_status::Used, false, "Set true to perform removenode operations in best effort mode.")
```
avikivity, line 689: The name of the variable and the explanation give no hint about what the flag does and why one would or would not use it.

We should make it part of the nodetool command flags, not a node configuration flag.

asias: I agree. It is hard to explain and for user to understand the option because it is quite complicated. 

I mentioned here: https://github.com/scylladb/scylla/pull/7452#issuecomment-711720737

Existing nodes know it needs to sync data for the removenode operation by gossip status REMOVING_TOKEN. We can extend REMOVING_TOKEN to contain a flag if we want safe or best effort mode. This is hack solution though.

A big picture of this problem: we may want to drop the removenode operation completely and use more safe alternative replace + decommission. 

removenode is very complicated operation than replace and decommission. It is unsafe. It is hard for users, especially for new users, to choose if they should removenode or replacing. From time to time, they choose removenode operation where replacing is a better option. 

What do you think?

avikivity: If a node is dead, replacenode may not be an option. It's easy in the cloud where there is an infinite supply of nodes, but on non-cloud, if a node dies, you may not be able to find a replacement.

asias: Yes. It is a good point user might not find a replacement node in some extreme cases. He can delay the replace operation until a new node is available or use the unsafe removenode ops (we give user big warning. Users might change the mind if they know the operation is unsafe). For most common cases, we can always use the replace + decommission operation. We should not sacrifice the uncommon cases for common cases.

kostja (changes requested):
@asias I think we need to patch add in the same way. Basically no streaming can happen in absense of majority. I tested the patch with Jepsen and it still fails, although now it takes it much longer to fail (2400 seconds instead of just 300 as it used to be earlier), and there is only one violation rather than many.

asias:
> @asias I think we need to patch **add node** in the same way. Basically no streaming can happen in absense of majority. I tested the patch with Jepsen and it still fails, although now it takes it much longer to fail (2400 seconds instead of just 300 as it used to be earlier), and there is only one violation rather than many.

In what cases did you see the failure? Is it removenode or bootstrap or replace ops? When bootstrap a new node, we only stream from one node that is losing the range if number of nodes > RF. When number of nodes < RF, it is streamed from 
nr_nodes_to_sync_with = std::min<size_t>(rf_in_local_dc / 2 + 1, old_endpoints_in_local_dc.size()). Is this your case? We can not require quorum in this case because otherwise we can not add a new node with RF = 3 when there is only one existing node in the cluster.

slivne:
@asias

> > 
> > best_effort_to_removenode is a config option -
> > 
> > We need to make this a flag controlled from nodetool / rest api - not a config option
> > 
> 
> We do not have a good way to communicate this option from the removenode coordinator to rest of the nodes that needs to perform the operation.
> 

- are all the nodes in the cluster supposed to be aware of this or only the node that starts the remove ?, I assume its all the nodes in the cluster
- the initiating node communicates to all the other nodes in the cluster that we are going to remove the missing node from the ring and going todo a lot of streaming - how do we communicate that - can't we transfer this information in that command.

Its a problematic case - to force the user todo a rolling restart to be able to remove a missing the node. The cluster is in a degraded mode (we are missing a node) if we need to restart nodes to change this flag value the availability is going to be hurt.

If we can't transfer this value in the same command that sends all that we are going to remove the node we have to make this option configurable at runtime without forcing to restart the node (yet I would still prefer - this would be sent from the node that initiates the removenode operation).

slivne:
@kostja 

> @asias I think we need to patch add node in the same way. Basically no streaming can happen in absense of majority. I tested the patch with Jepsen and it still fails, although now it takes it much longer to fail (2400 seconds instead of just 300 as it used to be earlier), and there is only one violation rather than many.

When we add a node we are streaming from a specific replica that we are taking the ownership from there is no majority involved.

asias:
> @asias
> 
> > > best_effort_to_removenode is a config option -
> > > We need to make this a flag controlled from nodetool / rest api - not a config option
> > 
> > 
> > We do not have a good way to communicate this option from the removenode coordinator to rest of the nodes that needs to perform the operation.
> 
>     * are all the nodes in the cluster supposed to be aware of this or only the node that starts the remove ?, I assume its all the nodes in the cluster
> 
>     * the initiating node communicates to all the other nodes in the cluster that we are going to remove the missing node from the ring and going todo a lot of streaming - how do we communicate that - can't we transfer this information in that command.
> 
> 
> Its a problematic case - to force the user todo a rolling restart to be able to remove a missing the node. The cluster is in a degraded mode (we are missing a node) if we need to restart nodes to change this flag value the availability is going to be hurt.
> 
> If we can't transfer this value in the same command that sends all that we are going to remove the node we have to make this option configurable at runtime without forcing to restart the node (yet I would still prefer - this would be sent from the node that initiates the removenode operation).

No need to restart a node. We can reload a value from config file.

However, I have an idea to make change the removenode to be completely a synchronous operation.  It has multiple benefits. It will take more time though since it is a big change.

asias:
> @kostja
> 
> > @asias I think we need to patch add node in the same way. Basically no streaming can happen in absense of majority. I tested the patch with Jepsen and it still fails, although now it takes it much longer to fail (2400 seconds instead of just 300 as it used to be earlier), and there is only one violation rather than many.
> 
> When we add a node we are streaming from a specific replica that we are taking the ownership from there is no majority involved.

There are exceptions: If number of nodes < RF, it is not guaranteed we stream from the node losing the range. See https://github.com/scylladb/scylla/pull/7452#issuecomment-722121348.

slivne:
@asias what is left to get this code in

asias:
> @asias what is left to get this code in

I do not know if you and avi think it is ok to introduce the option. 

I have a plan to simplify the removenode operation but it is a big change and require more time. My recommendation is to keep the fix in this patch simple and switch to the new removenode method later.

avikivity:
I think it's okay but not with a config option, rather a flag added to the REST API. It's very cumbersome to edit the config, SIGHUP,  issue the removenode again, and edit the config again to restore the original setting). Don't forget to SIGHUP the second time.

asias:
> I think it's okay but not with a config option, rather a flag added to the REST API. It's very cumbersome to edit the config, SIGHUP, issue the removenode again, and edit the config again to restore the original setting). Don't forget to SIGHUP the second time.

okay.

asias:
@avikivity @slivne @kostja I managed to have a draft  patch that implements the new removenode operation idea, without the need to config the cluster system wide.

Here is the PR:  https://github.com/scylladb/scylla/pull/7626

asias:
Closed in favor of new PR: #7626

PR #7454, opened 2020-10-19
database, streaming: remove remnants of memtable-base streaming

avikivity:
Commit e5be3352cf7 ("database, streaming, messaging: drop
streaming memtables") removed streaming memtables; this removes
the mechanisms to synchronize them: _streaming_flush_gate and
_streaming_flush_phaser. The memory manager for streaming is removed,
and its 10% reserve is evenly distributed between memtables and
general use (e.g. cache).

Note that _streaming_flush_phaser and _streaming_flush_date are
no longer used to syncrhonize anything - the gate is only used
to protect the phaser, and the phaser isn't used for anything.

### DIFF
--- a/database.cc
+++ b/database.cc
@@ -171,8 +171,7 @@ database::database(const db::config& cfg, database_config dbcfg, service::migrat
     , _cfg(cfg)
     // Allow system tables a pool of 10 MB memory to write, but never block on other regions.
     , _system_dirty_memory_manager(*this, 10 << 20, cfg.virtual_dirty_soft_limit(), default_scheduling_group())
-    , _dirty_memory_manager(*this, dbcfg.available_memory * 0.45, cfg.virtual_dirty_soft_limit(), dbcfg.statement_scheduling_group)
-    , _streaming_dirty_memory_manager(*this, dbcfg.available_memory * 0.10, cfg.virtual_dirty_soft_limit(), dbcfg.streaming_scheduling_group)
+    , _dirty_memory_manager(*this, dbcfg.available_memory * 0.50, cfg.virtual_dirty_soft_limit(), dbcfg.statement_scheduling_group)
     , _dbcfg(dbcfg)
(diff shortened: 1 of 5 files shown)


avikivity:
Update: update scylla-gdb.py

avikivity:
@tgrabiec ping

avikivity:
@tgrabiec ping

streaming/stream_transfer_task.cc
```diff
@@ -230,15 +230,6 @@ future<> stream_transfer_task::execute() {
             }
             return send_mutation_fragments(std::move(si));
         });
-    }).then([this, plan_id, cf_id, id] {
-        sslog.debug("[Stream #{}] SEND STREAM_MUTATION_DONE to {}, cf_id={}", plan_id, id, cf_id);
-        return session->ms().send_stream_mutation_done(id, plan_id, _ranges,
```
tgrabiec, line 235: Don't we still need to make this call, so that receive_task_completed() is called on the other node?

avikivity: I think we do, and I remember considering it. Maybe I only considered the server side and missed the client side.

avikivity: btw, @asias please review too

asias: We still need to send STREAM_MUTATION_DONE on the client side.

asias: Otherwise looks good to me.

avikivity:
Updated:
 - restored sending STREAM_MUTATION_DONE
 - tested using dtest's bootstrap_test.py (with repair_based_node_operations = false)

tgrabiec (approved):

asias (approved):

avikivity:
@tgrabiec looks like it was approved unanimously

PR #7456, opened 2020-10-19
treewide: adjust for missing aggregate template type deduction and parenthesized aggregate construction

avikivity:
Clang does not implement P0960R3 and P1816R0, so constructions
of aggregates (structs with no constructors) have to used braced initialization
and cannot use class template argument deduction. This series makes the
adjustments.


denesb (approved):
Will a clang build be part of gating?

avikivity:
There will only be a clang build.

psarna (approved):

PR #7458, opened 2020-10-19
Docker: pass extra arguments to scylla

dkropachev:
Currently there is no way to pass scylla arguments from docker-entrypoint to scylla,
but time to time it is needed.
Example: https://github.com/scylladb/scylla-operator/issues/177

### DIFF
--- a/dist/docker/redhat/scyllasetup.py
+++ b/dist/docker/redhat/scyllasetup.py
@@ -5,7 +5,7 @@ import os
 
 
 class ScyllaSetup:
-    def __init__(self, arguments):
+    def __init__(self, arguments, extra_arguments):
         self._developerMode = arguments.developerMode
         self._seeds = arguments.seeds
         self._cpuset = arguments.cpuset
@@ -30,6 +30,7 @@ class ScyllaSetup:
         self._endpointSnitch = arguments.endpointSnitch
         self._replaceAddressFirstBoot = arguments.replaceAddressFirstBoot
         self._io_setup = arguments.io_setup
+        self._extra_args = extra_arguments
 
     def _run(self, *args, **kwargs):
         logging.info('running: {}'.format(args))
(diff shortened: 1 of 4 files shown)


dist/docker/redhat/scyllasetup.py
```diff
@@ -151,5 +152,22 @@ def arguments(self):
 
         args += ["--blocked-reactor-notify-ms 999999999"]
 
+        argument = None
+        for arg in self._extra_args:
```
penberg, line 156: Is this transformation really needed? Can't we just append `extra_arguments` to `args`?

penberg (commented):
Nice improvement!

dist/docker/redhat/scyllasetup.py
dkropachev: In base cases we can drop it in favor of simple .split .

Tested with: 
--new_arg1=val --new_arg2 val --new_arg3="val" --new_arg4 "val" --new_arg5
Result:
SCYLLA_DOCKER_ARGS="--overprovisioned --listen-address 127.0.1.1 --rpc-address 127.0.1.1 --seed-provider-parameters seeds=127.0.1.1 --blocked-reactor-notify-ms 999999999 --new_arg1=val --new_arg2 val --new_arg3=val --new_arg4 val --new_arg5"

penberg:
Queued to `next`. Please send a follow up to document this in `docs/docker-hub.md`.

dkropachev:
> Queued to `next`. Please send a follow up to document this in `docs/docker-hub.md`.

Pushed another commit to put it into the docs

mmatczuk:
This is a move in a good direction, we also need a way to specify which services to run i.e. how do I run scylla only the entrypoint cannot be changed.

avikivity:
Suppose someone makes a spelling error in an argument. Now, parse_known_args() will ignore them and they will be passed to scylla. Scylla will start, fail, and be restarted by the supervisor.

PR #7466, opened 2020-10-20
build: supply linker flags only to the linker, not the compiler

avikivity:
Clang complains if it sees linker-only flags when called for compilation,
so move the compile-time flags from cxx_ld_flags to cxxflags, and remove
cxx_ld_flags from the compiler command line.

The linker flags are also passed to Seastar so that the build-id and
interpreter hacks still apply to iotune.


avikivity:
@penberg ping

PR #7467, opened 2020-10-20
build: relocatable package: exclude tools/python3

avikivity:
python3 has its own relocatable package, no need to include it
in scylla-package.tar.gz.

Python has its own relocatable package, so packaging it in scylla-package.ta


avikivity:
@penberg ping

PR #7468, opened 2020-10-20
scylla_boot_setup: use kvm-clock for clock source on AWS Nitro VMs

### LINKED ISSUE #7444: scylla_bootparam_setup uses TSC clocksource even on AWS Nitro VMs where it is not recommended
According to [AWS Support](https://aws.amazon.com/premiumsupport/knowledge-center/manage-ec2-linux-clock-source/) and the [relevant optimization procedure](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking-os.html), scylla_bootparam_setup should only switch to the TSC clock source on Xen VMs, not Nitro VMs. I'm not sure how you tell you're on a Nitro VM, but checking if `/sys/devices/system/clocksource/clocksource0/available_clocksource` contains `kvm-clock` should be sufficient (and then I guess set the clocksource to kvm-clock instead of tsc).

If there's a reason why TSC is better than the recommended clock on Nitro VMs, then it would be helpful to document that.

syuu1228:
We currently set 'clocksource=tsc' on bootparameter statically on our AMI,
since it is recommended for AWS Xen Hypervisor.
However, Amazon added AWS Nitro Hypervisor, which is recommended to use
kvm-clock as clocksource.
We need to support both hypervisor on single AMI image, so we need to configure
clocksource on first boot of the instance, not AMI creation type.

To do this, we need to detect which type of hypervisor we are on,
and set optimal clocksource to /sys/devices/system/clocksource/clocksource0/current_clocksource, not just change bootparameter settings.

Fixes #7444

### DIFF
--- a/dist/common/scripts/scylla_bootparam_setup
+++ b/dist/common/scripts/scylla_bootparam_setup
@@ -27,11 +27,16 @@ import argparse
 import subprocess
 from scylla_util import *
 
+def set_clocksource(source):
+    print(f'Set {source} as clock source')
+    with open('/sys/devices/system/clocksource/clocksource0/current_clocksource', 'w') as f:
+        f.write(source)
+
 if __name__ == '__main__':
     if os.getuid() > 0:
         print('Requires root permission.')
         sys.exit(1)
-    parser = argparse.ArgumentParser(description='Optimize boot parameter settings for Scylla.')
+    parser = argparse.ArgumentParser(description='Optimize clocksource settings for Scylla.')
     parser.add_argument('--ami', action='store_true', default=False,
                         help='setup AMI instance')
     args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


avikivity:
There are five commits here.

syuu1228:
@avikivity sorry, rebased

syuu1228:
I just realized that we have added clocksource tuneup feature on perftune, and also there is a option to enable it on scylla_setup:
https://github.com/scylladb/seastar/commit/9357a7b2eba9e6d51d7cdfc1a03a070cbddd8d2d#diff-3308b1dbea6a75007234eacc16bcf645c9245ab6e28b1c216dd77771838d30b5
https://github.com/scylladb/scylla/commit/d7795b1efae939cf1e37122c6cb664fe8968aa2c#diff-50a6af38badc886bac68b8724c68cab0a3de945c91331d22da73be9428b67b45
So my PR is duplicated code with this one, I suppose we just need to drop bootparam modification and turn on clocksource tune on perftune in AMI creation.

syuu1228:
I think I need to close this due to https://github.com/scylladb/scylla/pull/7468#issuecomment-714575429, I'll make another PR

PR #7475, opened 2020-10-22
cql3: Check for timestamp correctness in USING TIMESTAMP statements 

### LINKED ISSUE #5619: USING TIMESTAMP value range is not checked

*Installation details*
Scylla version (or git commit hash): 2019.1.2-0.20190814.2772d525c

*Description*

CQL allows to specify `USING TIMESTAMP` clause or to specify optional default timestamp for a query in the cql protocol (that might be set using client drivers). The value for either of these is documented as microseconds since UNIX epoch.

When issuing CQL statements with `USING TIMESTAMP` clause, Scylla should return a query error in case the value is too far in the future. This can catch bugs when a client uses wrong units.

For example, if a single client uses nanosecond timestamp instead of microsecond timestamp as other clients (or Scylla coordinators) use, the client would overwrite even future updates by other clients, as the timestamps by the misbehaving client would be more than 50000 years in the future. Reverting the misbehaving client to use microseconds does not fix the issue, since data with the far-future timestamps has already been written, so to recover from that kind of client mistake, one has to either build a new table from scratch with the current timestamps or switch all other clients to use timestamp in nanoseconds.

I'm not sure what the limit value should be, but updates more than a few days in the future seem suspicious. If update timestamps in the future are a supported feature, the limit might need to be configurable.

StarostaGit:
In certain CQL statements it's possible to provide a custom timestamp via the USING TIMESTAMP clause. Those values are accepted in microseconds, however, there's no limit on the timestamp (apart from type size constraint) and providing a timestamp in a different unit like nanoseconds can lead to creating an entry with a timestamp way ahead in the future, thus compromising the table. To avoid this, this change introduces a sanity check for modification and batch statements that raises an error when a timestamp of more than 3 days into the future is provided.

Test: unit (dev)

Fixes #5619

### DIFF
--- a/cql3/statements/batch_statement.cc
+++ b/cql3/statements/batch_statement.cc
@@ -38,6 +38,7 @@
  */
 
 #include "batch_statement.hh"
+#include "cql3/util.hh"
 #include "raw/batch_statement.hh"
 #include "db/config.hh"
 #include "db/consistency_level_validations.hh"
@@ -259,6 +260,7 @@ static thread_local inheriting_concrete_execution_stage<
 
 future<shared_ptr<cql_transport::messages::result_message>> batch_statement::execute(
         service::storage_proxy& storage, service::query_state& state, const query_options& options) const {
+    cql3::util::validate_timestamp(options, _attrs);
     return batch_stage(this, seastar::ref(storage), seastar::ref(state),
                        seastar::cref(options), false, options.get_timestamp(state));
 }
--- a/cql3/statements/modification_statement.cc
+++ b/cql3/statements/modification_statement.cc
@@ -44,6 +44,7 @@
 #include "cql3/statements/raw/modification_statement.hh"
 #include "cql3/statements/prepared_statement.hh"
 #include "cql3/restrictions/single_column_restriction.hh"
+#include "cql3/util.hh"
 #include "validation.hh"
 #include "db/consistency_level_validations.hh"
 #include <seastar/core/shared_ptr.hh>
@@ -258,6 +259,7 @@ static thread_local inheriting_concrete_execution_stage<
 
 future<::shared_ptr<cql_transport::messages::result_message>>
 modification_statement::execute(service::storage_proxy& proxy, service::query_state& qs, const query_options& options) const {
+    cql3::util::validate_timestamp(options, attrs);
     return modify_stage(this, seastar::ref(proxy), seastar::ref(qs), seastar::cref(options));
 }
 
--- a/test/boost/cql_query_test.cc
+++ b/test/boost/cql_query_test.cc
@@ -4754,3 +4754,17 @@ SEASTAR_TEST_CASE(test_counter_column_added_into_non_counter_table) {
                 exceptions::configuration_exception);
     });
 }
+
+SEASTAR_THREAD_TEST_CASE(test_invalid_using_timestamps) {
+    do_with_cql_env_thread([] (cql_test_env& e) {
+        auto now_nano = std::chrono::duration_cast<std::chrono::nanoseconds>(db_clock::now().time_since_epoch()).count();
+        e.execute_cql("CREATE TABLE tbl (a int, b int, PRIMARY KEY (a))").get();
+
+        BOOST_REQUIRE_THROW(e.execute_cql(format("INSERT INTO tbl (a, b) VALUES (1, 1) USING TIMESTAMP {}", now_nano)).get(), exceptions::invalid_request_exception);
+
+        e.execute_cql("INSERT INTO tbl (a, b) VALUES (1, 1)").get();
+        BOOST_REQUIRE_THROW(e.execute_cql(format("UPDATE tbl USING TIMESTAMP {} SET b = 10 WHERE a = 1", now_nano)).get(), exceptions::invalid_request_exception);
+        BOOST_REQUIRE_THROW(e.execute_cql(format("DELETE b FROM tbl USING TIMESTAMP {} WHERE a = 1", now_nano)).get(), exceptions::invalid_request_exception);
+        BOOST_REQUIRE_THROW(e.execute_cql(format("BEGIN BATCH USING TIMESTAMP {} INSERT INTO TBL (a, b) VALUES (2, 2); APPLY BATCH", now_nano)).get(), exceptions::invalid_request_exception);
+    }).get();
+}
--- a/cql3/util.cc
+++ b/cql3/util.cc
@@ -119,5 +119,19 @@ void do_with_parser_impl(const sstring_view& cql, noncopyable_function<void (cql
 
 #endif
 
+void validate_timestamp(const query_options& options, const std::unique_ptr<attributes>& attrs) {
+    if (attrs->is_timestamp_set()) {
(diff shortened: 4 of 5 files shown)


avikivity:
Please use Fixes instead of Closes, both because it carries more meaning, and because our tooling looks for that keyword.

cql3/statements/modification_statement.cc
```diff
@@ -529,6 +529,19 @@ modification_statement::validate(service::storage_proxy&, const service::client_
     if (is_view()) {
         throw exceptions::invalid_request_exception("Cannot directly modify a materialized view");
     }
+
+    if (attrs->is_timestamp_set()) {
+        static constexpr int64_t MAX_DIFFERENCE = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::days(3)).count();
+        auto now = std::chrono::duration_cast<std::chrono::microseconds>(db_clock::now().time_since_epoch()).count();
+        cql3::query_options query_options(std::vector<cql3::raw_value>{cql3::raw_value::make_unset_value()});
+
```
avikivity, line 537: This looks like a wasteful way to get a timestamp, including an allocation for a vector.

avikivity: validate() runs on prepare time, while USING TIMESTAMP can have a value that is only available during execution time (`USING TIMESTAMP :my_variable`). Not sure how this works.

StarostaGit: Fair point, moved the checks to right before execution

StarostaGit:
Addressed feedback and added a unit test

avikivity:
The commit changelog has no description of the problem or of the solution.

haaawk:
> The commit changelog has no description of the problem or of the solution.

@StarostaGit please just copy PR description to the commit message.

cql3/statements/batch_statement.cc
```diff
@@ -173,6 +173,21 @@ void batch_statement::validate(service::storage_proxy& proxy, const service::cli
     }
 }
 
+void batch_statement::validate_timestamp(const query_options& options) const {
+    if (_attrs->is_timestamp_set()) {
+        static constexpr int64_t MAX_DIFFERENCE = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::days(3)).count();
+        auto now = std::chrono::duration_cast<std::chrono::microseconds>(db_clock::now().time_since_epoch()).count();
+        cql3::query_options query_options(std::vector<cql3::raw_value>{cql3::raw_value::make_unset_value()});
```
haaawk, line 180: Where is this used?

test/boost/cql_query_test.cc
```diff
@@ -4754,3 +4754,19 @@ SEASTAR_TEST_CASE(test_counter_column_added_into_non_counter_table) {
                 exceptions::configuration_exception);
     });
 }
+
+SEASTAR_TEST_CASE(test_invalid_using_timestamps) {
```
haaawk, line 4758: SEASTAR_THREAD_TEST_CASE is preferred.

haaawk (commented):
LGTM

cql3/statements/batch_statement.cc
StarostaGit: leftover from previous approach

StarostaGit:
change test type and populated the commit changelog

haaawk (approved):
LGTM

haaawk:
@avikivity would you mind taking another look, please?

avikivity:
Please add the Fixes tag to the commit. In single-patch series, the cover letter doesn't get into the log, so we lose that information.

cql3/statements/modification_statement.cc
```diff
@@ -531,6 +532,20 @@ modification_statement::validate(service::storage_proxy&, const service::client_
...
+        auto timestamp = get_timestamp(now, options);
+
+        if (timestamp - now > MAX_DIFFERENCE) {
+            throw exceptions::invalid_request_exception("Cannot provide a timestamp more than 3 days into the future. If this was not intended, "
+            "make sure the timestamp is in microseconds");
+        }
+    }
+}
+
```
avikivity, line 548: This looks like code duplication. I think you can move it to query_options.

Perhaps it can be called even earlier, from transport/server.cc, when the frame is parsed and the timestamp extracted.

StarostaGit: While query_options are indeed populated early, server.cc is not aware of the exact type of the statement and (at least to my knowledge) checking what was provided in the `USING TIMESTAMP` clause directly would require ugly checks and casting.

Now, as for the code duplication and moving this to query_options - apart from them, we also determine the timestamp based on the attributes of the statement and delegating that to query_options (which is used to populate the statement, not the other way around) would create a very counter-intuitive back and forth (pass the options to statement -> get timestamp using options -> pass that timestamp back to options for validation).
I can extract that code to a separate function, standalone in some namespace, or perhaps a static somewhere up the statement class hierarchy tree. But, is it really worth it? I'd argue that it's a small piece of code in only 2 places (I don't think `USING TIMESTAMP` will appear in many more statement types soon) and it provides a natural separation of concerns as it is right now.

StarostaGit:
Updated the commit message to include the **Fixes** clause.

cql3/statements/modification_statement.cc
haaawk: I had the same observation as @avikivity but couldn't convince myself one way or the other. Fundamentally this code does a well-separated check so it wouldn't hurt to wrap it into a well-named and well-commented free function.

haaawk: I would avoid doing it in a class hierarchy.

avikivity: My suggestion to move to transport or query_options is nonsensical, you don't have access to batch_statement::_attrs there.

But a free function to reduce duplication is a good idea. It's a small amount of code but a free function is not adding any complexity (and de-duplicates a constant so we don't have to change it twice if we ever want to).

StarostaGit:
Moved the check to a free function in `cql3::util` namespace

cql3/statements/modification_statement.hh
```diff
@@ -220,6 +220,7 @@ private:
     bool applies_only_to_static_columns() const {
         return _sets_static_columns && !_sets_regular_columns && _regular_conditions.empty();
     }
+
```
haaawk, line 223: Please remove from the PR.

cql3/statements/modification_statement.cc
```diff
@@ -258,6 +259,10 @@ static thread_local inheriting_concrete_execution_stage<
 
 future<::shared_ptr<cql_transport::messages::result_message>>
 modification_statement::execute(service::storage_proxy& proxy, service::query_state& qs, const query_options& options) const {
+    if (attrs->is_timestamp_set()) {
+        cql3::util::validate_timestamp(get_timestamp(0, options));
+    }
```
haaawk, line 264: I would put the if inside the validate_timestamp function. Just take attrs and options as argument of validate_timestamp.

haaawk: get_timestamp calls -> _attrs->get_timestamp(0, options)

StarostaGit: done

haaawk (approved):
LGTM

PR #7479, opened 2020-10-23
config_file: read_from_yaml support flag shorthands

### LINKED ISSUE #7478: Config file workdir option not working

*Installation details*
Scylla version (or git commit hash): 4.2 5fcc1f20

Setting workdir param

```
# Directory where Scylla should store all its files, which are commitlog,
# data, hints, view_hints and saved_caches subdirectories. All of these
# subs can be overriden by the respective options below.
# If unset, the value defaults to /var/lib/scylla
workdir: /home/michal/scylla/
```

results in 

```
WARN  2020-10-22 15:25:45,067 [shard 0] init - Unknown option : workdir
```

and config option is not applied.

mmatczuk:
This patch adds support for long_flag_name,short_flag_name syntax used in workdir,W.

Without this patch workdir option specified in config file results in

```
WARN  2020-10-22 15:25:45,067 [shard 0] init - Unknown option : workdir
```

Fixes #7478

### DIFF
--- a/utils/config_file.cc
+++ b/utils/config_file.cc
@@ -221,6 +221,16 @@ sstring utils::hyphenate(const std::string_view& v) {
     return result;
 }
 
+sstring utils::trim_to_first_comma(const std::string_view& v) {
+    auto idx = v.find(',');
+    if (idx == 0) {
+        return v.data();
+    } else {
+        sstring result(v.substr(0, idx));
+        return result;
+    }
+}
+
 utils::config_file::config_file(std::initializer_list<cfg_ref> cfgs)
     : _cfgs(cfgs)
 {}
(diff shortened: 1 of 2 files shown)


utils/config_file.cc
```diff
@@ -221,6 +221,16 @@ sstring utils::hyphenate(const std::string_view& v) {
     return result;
 }
 
+sstring utils::trim_to_first_comma(const std::string_view& v) {
+    auto idx = v.find(',');
+    if (idx == 0) {
```
avikivity, line 226: find() returns npos, not 0, when a match is not found (https://en.cppreference.com/w/cpp/string/basic_string_view/find)

utils/config_file.cc
```diff
@@ -221,6 +221,16 @@ sstring utils::hyphenate(const std::string_view& v) {
     return result;
 }
 
+sstring utils::trim_to_first_comma(const std::string_view& v) {
```
avikivity, line 224: You may as well return a string_view.

PR #7492, opened 2020-10-27
Implement managed_bytes and managed_bytes_view apis

bhalevy:
Based on top of https://github.com/scylladb/scylla/pull/7490

Implements various methods unimplemented in skeleton implementation.


bhalevy:
sorry, wrong repo.  not yet...

PR #7498, opened 2020-10-28
alternator::streams: Use end-of-record info in get_records

### LINKED ISSUE #7496: Use CDC "end of record" marker in Alternator Streams
In issue #7435 fixed by commit 46ea8c9b8b067ccebacb1a3be020c3c8068dc684, we added to CDC an "end of record" marker (a boolean column) which can allow the CDC log reader to make sure that it got the full change (composed of multiple rows in the log). 
The CDC log reader in Alternator Streams should make use of this new flag, to avoid returning to the reader of the Alternator Stream a change event based on partial information.  

CC @elcallio

elcallio:
Fixes #7496

Since cdc log now has an end-of-batch/record marker that tells
us explicitly that we've read the last row of a change, we
can use this instead of timestamp checks + limit extra to
ensure we have complete records.

Note that this does not try to fulfill user query limit
exact. To do this we would need to add a loop and potentially
re-query if quried rows are not enough. But that is a
separate exercise, and superbly suited for coroutines!

### DIFF
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -885,8 +885,17 @@ future<executor::request_return_type> executor::get_records(client_state& client
     auto partition_slice = query::partition_slice(
         std::move(bounds)
         , {}, std::move(regular_columns), selection->get_query_options());
+
+	auto& opts = base->cdc_options();
+	auto mul = 2; // key-only, allow for delete + insert
+    if (opts.preimage()) {
+        ++mul;
+    }
+    if (opts.postimage()) {
+        ++mul;
+    }
     auto command = ::make_lw_shared<query::read_command>(schema->id(), schema->version(), partition_slice, _proxy.get_max_result_size(partition_slice),
-            query::row_limit(limit * 4));
+            query::row_limit(limit * mul));
 
     return _proxy.query(schema, std::move(command), std::move(partition_ranges), cl, service::storage_proxy::coordinator_query_options(default_timeout(), std::move(permit), client_state)).then(
             [this, schema, partition_slice = std::move(partition_slice), selection = std::move(selection), start_time = std::move(start_time), limit, key_names = std::move(key_names), attr_names = std::move(attr_names), type, iter, high_ts] (service::storage_proxy::coordinator_query_result qr) mutable {       
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -851,6 +851,7 @@ future<executor::request_return_type> executor::get_records(client_state& client
 
     static const bytes timestamp_column_name = cdc::log_meta_column_name_bytes("time");
     static const bytes op_column_name = cdc::log_meta_column_name_bytes("operation");
+    static const bytes eor_column_name = cdc::log_meta_column_name_bytes("end_of_batch");
 
     auto key_names = boost::copy_range<std::unordered_set<std::string>>(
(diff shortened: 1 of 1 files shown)


alternator/streams.cc
```diff
@@ -884,8 +885,17 @@ future<executor::request_return_type> executor::get_records(client_state& client
     auto partition_slice = query::partition_slice(
         std::move(bounds)
         , {}, std::move(regular_columns), selection->get_query_options());
+    auto& opts = base->cdc_options();
```
haaawk, line 888: That's an unrelated optimization, no? Or at least not directly related to EOR?

alternator/streams.cc
```diff
@@ -932,15 +947,8 @@ future<executor::request_return_type> executor::get_records(client_state& client
...
-            if (timestamp && timestamp != ts) {
-                maybe_add_record();
-                if (limit == 0) {
-                    break;
-                }
-            }
-
-            timestamp = ts;
+            // end_of_batch is boolean, but always null or true.
+            auto eor = row[eor_index].has_value();
```
haaawk, line 951: Just for the sake of being save in the future, I would check the value anyway in case we change the behavior and the false value actually starts appearing.

alternator/streams.cc
```diff
@@ -993,9 +1001,13 @@ future<executor::request_return_type> executor::get_records(client_state& client
                 rjson::set(record, "eventName", "REMOVE");
                 break;
             }
-        }
-        if (limit > 0 && timestamp) {
-            maybe_add_record();
+            if (eor) {
+                maybe_add_record();
```
haaawk, line 1005: I don't understand the change of this logic.
Previously, `maybe_add_record` was called every time `limit > 0` because `timestamp` was always set in line 943.
Now, `maybe_add_record` is only called when `eor` is true.

Could you please explain why this is ok @elcallio ? I'm probably missing something because I don't understand how `timestamp` could have ever been empty at this point.

haaawk:
Conceptually I think I understand the change. I also think that it's correct and correctly implemented but the result of the change modifies the behavior of the code. It seems to change not only how things are done (as the description would suggest) but also what is done. I may be wrong but it seems that `maybe_add_record` is now called less often. This got me a bit confused.

I would also put a loosely related `query::row_limit` change in a separate commit just for the sake of making it absolutely clear to reviewers that it's not coupled with the rest of the change.

All of this may just be me not understanding the code so please correct me when I'm wrong, @elcallio.

alternator/streams.cc
elcallio: In a sense maybe

alternator/streams.cc
elcallio: The diff is tricking you. There was two `maybe_add_record` calls before. One above, when checking if we switched timestamp, one post-loop that checked if we should seal the record being built (hoping for the best - that it was complete). 
Now there is only one, which after processing data in the cdc row, closes and adds the record iff `eor==true`, all inside the loop. We no longer have to even bother about the timestamp changing.

elcallio:
@haaawk  issues addressed...

nyh (approved):

haaawk (approved):
LGTM

haaawk:
@nyh I think we can merge this, no?

nyh:
@haaawk I already did, 5 hours ago. See the automatic comment from Github about the merge, above. This PR will automatically close once that patch will reach master.

PR #7499, opened 2020-10-28
test: increase quota for tests to 6GB

avikivity:
test.py estimates the amount of memory needed per test
in order not to overload the machine, but it underestimates
badly and so machines with many cores but not a lot of memory
fail the tests (in debug mode principally) due to running out
of memory.

Increase the estimate from 2GB per test to 6GB.

### DIFF
--- a/test.py
+++ b/test.py
@@ -509,7 +509,7 @@ def parse_cmd_line():
     """ Print usage and process command line options. """
     all_modes = ['debug', 'release', 'dev', 'sanitize']
     sysmem = os.sysconf('SC_PAGE_SIZE') * os.sysconf('SC_PHYS_PAGES')
-    testmem = 2e9
+    testmem = 6e9
     cpus_per_test_job = 1
     default_num_jobs_mem = ((sysmem - 4e9) // testmem)
     default_num_jobs_cpu = multiprocessing.cpu_count() // cpus_per_test_job


nyh:
@avikivity normally, Seastar has full control on how much memory it uses, so if we ask for 2GB, it will take exactly 2GB. How can it reach 6GB? Maybe the problem is specific to debug mode because it does not use the Seastar allocator, so can allocate much more than desired? If this is the case, shouldn't we leave the estimate for other modes at 2GB and only increase the estimate for debug mode? 

@avikivity @kostja I also want to add that it's bad that the number 2G appears separately in three places in the code: UnitTestSuite.add_test(), CqlTest.__init__() and the inappropriately-named parse_cmd_line(). A better approach would be for individual tests to be able to estimate their own memory use - and unit tests (except debug mode) and CqlTest would choose their own memory usage and also report the same number as the estimate. Unit tests in debug mode would need to report a higher number as the estimate.

avikivity:
Yes, it's debug mode.

nyh:
> Yes, it's debug mode.

So why not only consider only debug runs to take 6 GB? If someone is running *only* dev tests, for example, none of them will ever take 6 GB.

nyh:
> > Yes, it's debug mode.
> 
> So why not only consider only debug runs to take 6 GB? If someone is running _only_ dev tests, for example, none of them will ever take 6 GB.

Looking at the test.py, I see it isn't ready for using the actual memory usage and cpu usage of jobs it ran. I'll create a new issue about doing this.

nyh:
I opened https://github.com/scylladb/scylla/issues/7501. I guess we can commit this PR first, and only later do the right thing - perhaps what I described in that issue.

PR #7502, opened 2020-10-29
dist: redhat: don't pull in kernel package

avikivity:
We require a kernel that is at least 3.10.0-514, because older
kernel have an XFS related bug that causes data corruption. However
this Requires: clause pulls in a kernel even in Docker installation,
where it (and especially the associated firmware) occupies a lot of
space.

Change to a Conflicts: instead. This prevents installation when
the really old kernel is present, but doesn't pull it in for the
Docker image.

### DIFF
--- a/dist/redhat/scylla.spec
+++ b/dist/redhat/scylla.spec
@@ -51,7 +51,7 @@ Group:          Applications/Databases
 Summary:        The Scylla database server
 License:        AGPLv3
 URL:            http://www.scylladb.com/
-Requires: kernel >= 3.10.0-514
+Conflicts: kernel < 3.10.0-514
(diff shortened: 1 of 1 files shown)


nyh (approved):

nyh:
Committed this patch (bd73898a5c80d14015d759dea0debebbda78b484). Beyond being useful, this patch is also **correct** - Scylla does not actually depend on the Kernel package. It will not need the kernel executable, or header files, or anything else in the kernel package. So Scylla doesn't need to insist that any kernel package is installed - just that a supported kernel is running (and this can only be checked at runtime).

espindola:
LGTM. Should we do something similar for the debian package?

Issue #5304, opened 2019-11-19
Consolidate limit number of pending gossip ACK and ACK2 messages handling


asias:
In commit f32ae00510dcf11ce402c4551ede076b7baec15f and 15148182ab400e147831598bef4dcb42268b6a40, we have similar logic to process the incoming pending messages. It would be nice to consolidate them with a helper to reduce code duplication.

PR #7503, opened 2020-10-29
[RFC] gossiper: Consolidate limiters for pending ACK and ACK2

michoecho:
This is my attempt at #5304. But honestly, I think that this piece of code will be better off without such refactoring. One small piece of duplication is cleaner than a template in this case. Especially since there are some small differences (log messages) that need to be specialized for both types of messages, which muddles the code. Perhaps #5304 should simply be closed without changes? Or maybe you could point me to a solution more elegant than just stuffing both pieces of similar code into a template?

### DIFF
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -124,7 +124,9 @@ public:
 };
 
 gossiper::gossiper(abort_source& as, feature_service& features, const locator::token_metadata& tokens, netw::messaging_service& ms, db::config& cfg)
-        : _abort_source(as)
+        : _syn_queue(*this)
+        , _ack_queue(*this)
+        , _abort_source(as)
         , _feature_service(features)
         , _token_metadata(tokens)
         , _messaging(ms)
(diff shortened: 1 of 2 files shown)


avikivity:
@michoecho first, your patch has to explain what problem it is solving, and how it is solving it in the changelog. Look at `git log` output from this repository.

When I review I first review the changelog to see if it something that should be done, and then I look at the patch itself to see how it is done.

avikivity:
I agree that adding a template to eliminate just this duplication is overkill, since it's just two messages you're consolidating. Perhaps a non-template de-duplicating with smaller scope would be better.

tgrabiec:
I agree that templating the whole thing is not the right direction.

The main problem here is not the duplication but intertwining high-lever and low-level logic in one method. It would help to extract the low level logic into a synchronizer like `serialized_action`.

michoecho:
Closing due to inactivity and doubtful value of the refactoring.

PR #7505, opened 2020-10-29
Restrict default access to system keyspace

### LINKED ISSUE #3607: system.large_partitions access should be restricted to operator
The large_partitions table contains partition keys, which are potentially sensitive. They should be restricted to the superuser by default.

(in fact the entire system keyspace should prevent writes by non-superusers).

StarostaGit:
This change marks the whole system keyspace as a restricted resource
in the default authorizer so that it can only be accessed by the superuser.
Later, permissions can still be granted to other users if the need arises.

Fixes #3607

### DIFF
--- a/auth/default_authorizer.cc
+++ b/auth/default_authorizer.cc
@@ -334,7 +334,7 @@ future<> default_authorizer::revoke_all(const resource& resource) const {
 }
 
 const resource_set& default_authorizer::protected_resources() const {
-    static const resource_set resources({ make_data_resource(meta::AUTH_KS, PERMISSIONS_CF) });
+    static const resource_set resources({ make_data_resource(meta::AUTH_KS, PERMISSIONS_CF), make_data_resource(db::system_keyspace_name()) });
     return resources;
 }
 


avikivity:
Don't the drivers query the system keyspace?

elcallio:
At least the java driver does.

nyh:
> At least the java driver does.

I think this has to be true for all drivers. Unfortunately, CQL doesn't have any introspection features - commands you have in cqlsh like "DESCRIBE KEYSPACE", "DESCRIBE TABLE", etc., are not really part of the CQL protocol, but rather implemented in cqlsh or the driver by fetching schema information from the system tables. Drivers also need to read these schemas even if the user didn't explicitly asked to show the schema.
So some of these tables need to be readable by the client, and moreover - we had to be very careful in making them backward-compatible (not to change their format) so they can still be readable by older clients.
So we should probably consider the different tables in this keyspace individually, and can't make a blanket decision for all of them.

By the way, all of this demonstrates that with the current design, table-based ACL might not be good enough. It doesn't let us to allow a role to only to describe his own tables and not others. It doesn't let us to allow a role to only look at large partitions in his own tables and not other tables. And so on.

elcallio:
You could still make all the system tables read-only for any external (non-scylla) connection.

StarostaGit:
The associated issue (#3607) has been closed

PR #7509, opened 2020-10-29
test: sstable_3_x_test: reduce stack usage in thread- local storage initialization

avikivity:
gcc collects all the initialization code for thread-local storage
and puts it in one giant function. In combination with debug mode,
this creates a very large stack frame that overflows the stack
on aarch64.

Work around the problem by placing each initializer expression in
its own function, thus reusing the stack.


haaawk (approved):
LGTM

Issue #5578, opened 2020-01-09
imr: switch to run-time (open-coded) description of structures


denesb:
When the IMR infrastructure was originally proposed there was a long discussion on whether the description of structures (e.g. cells) should be at least partially compile-time (via template meta-programming) or whether it should be entirely runtime (open-coded) [[1]](https://groups.google.com/d/topic/scylladb-dev/wsGtRT3pL-E/discussion).

The arguments for-against the compile-time description were:

Pros:
* Human readable definition of the building-blocks, available up-ahead and statically.
* More efficient code.
* Easier to change anything as the compiler generates the code.
* Easier way forward to LLVM JIT compiled (de)serialization.

Cons:
* Complex, very hard to understand and modify implementation (template meta-programming).
* Low cost/benefit ratio, as most things are known only at runtime.
* Poor diagnostics.
* Slow compile times.

After some back-and-forth we were convinced that the the pros outweigh the cons. Indeed compile times didn't prove to be an issue and most building blocks of cells could be nicely described compile-time with `imr::variant` providing the necessary runtime switching between the possible options. We were also told that tools will likely catch up soon with templates. The sad reality is that they still haven't. Also the complexity and unreadability of the implementation remained a serious issue, exacerbated with the original author no longer being with us.
After some mileage with the IMR code additional problems surfaced:
* Debugability is horrible as `gdb` fails to provide accurate source code mapping inside the `imr` code, one is left to either debug ASM or try to guess where the problem is based on code addresses of the frames.
* Although the definitions of new building blocks can be done easily, things quickly become very tricky as it is very easy to write context objects that only handle parts of the fields, producing corrupt data. Generally it is easy to make mistakes as template meta-programming is not type-safe. This was not yet seen in the wild (AFAIC), I only bumped into this when experimenting with moving containers to IMR.

IMR was supposed to become the infrastructure for all our building blocks, from cells all the way to rows, keys and maybe even partitions. However development has halted as developers are reluctant to touch the code and maintainers are even more reluctant to make even more of our building blocks this complex.

To sum it up, after reconsidering the question, with some mileage under our belt, it seems that runtime description of types would be a better option for such mundane reasons as readability, maintainability and debugability. There are no surprises here, we've known/suspected all the pros/cons from the beginning, it is just their cost/benefit ration that we miscalculated.

kbr-:
> Complex, very hard to understand and modify implementation (template meta-programming)

Does the mere fact that it uses template meta-programming make it complex?
[...]

PR #7514, opened 2020-10-30
imr: switch back to open-coded description of structures

denesb:
Commit aab6b0ee27a7bc3fdf1d879d25291f807b4795a5 introduced the controversial new IMR format, which relied on a very template-heavy infrastructure to generate serialization and deserialization code via template meta-programming. The promise was that this new format, beyond solving the problems the previous open-coded representation had (working on linearized buffers), will speed up migrating other components to this IMR format, as the IMR infrastructure reduces code bloat, makes the code more readable via declarative type descriptions as well as safer. However, the results were almost the opposite. The template meta-programming used by the IMR infrastructure proved very hard to understand. Developers don't want to read or modify it. Maintainers don't want to see it being used anywhere else. In short, nobody wants to touch it.

This commit does a conceptual revert of aab6b0ee27a7bc3fdf1d879d25291f807b4795a5. A verbatim revert is not possible because related code evolved a lot since the merge. Also, going back to the previous code would mean we regress as we'd revert the move to fragmented buffers. So this revert is only conceptual, it changes the underlying infrastructure back to the previous open-coded one, but keeps the fragmented buffers, as well as the interface of the related components (to the extent possible).

Fixes: #5578

TODO:
* Tests

Review the top 2 commits only. The rest will be folded into https://github.com/scylladb/scylla/pull/7490.


denesb:
This series was sponsored by the IMR removal hackathon, Scylla Engineering Virtual Summit 2020.

utils/managed_bytes.hh
```diff
@@ -458,10 +458,20 @@ public:
...
     iterator end() { return iterator(); }
 
-    const_iterator begin(size_t offset = 0) const;
+    const_iterator begin() const {
+        if (external()) {
+            return const_iterator(_u.ptr->data, _u.ptr->data + _u.ptr->frag_size, _u.ptr->next);
+        }
+        return const_iterator(_u.small.data, _u.small.data + _u.small.size, nullptr);
+    }
     const_iterator end() const { return const_iterator(); }
```
avikivity, line 475: begin/end are dangerous, since they are slow and people might use them acceidentally.

Perhaps we can have something like

```c++
    template <typename Func>
    requires std::invocable<Func, managed_bytes::iterator, managed_bytes::iterator> 
          && std::invocable<Func, bytes_view::iterator, bytes_view::iterator> 
    auto managed_bytes::with_iterator_pair(Func func) {
         if (is_fragmented()) {
             return func(slow_begin(), slow_end());
         } else {
             return func(single_fragmented_begin(), single_fragment_end());
         }
    }
```

michoecho:
I have corrected a couple of errors introduced in this change. I'm investigating the remaining ones:
https://github.com/denesb/scylla/pull/9

denesb:
Closing in favor of https://github.com/scylladb/scylla/pull/8106.

PR #7524, opened 2020-11-01
scylla_swap_setup: when swap already configured, exit with return code 0

bentsi:
Fixes: https://github.com/scylladb/scylla/issues/7522
If scylla_swap_setup is run when the swap already exists the scripts exits with error code 1. We shouldn't fail the script in such case.

@penberg @yaronkaikov

### DIFF
--- a/dist/common/scripts/scylla_swap_setup
+++ b/dist/common/scripts/scylla_swap_setup
@@ -46,7 +46,7 @@ if __name__ == '__main__':
 
     if swap_exists():
         print('swap already configured, exiting setup')
-        sys.exit(1)
+        sys.exit(0)
 
     diskfree = psutil.disk_usage(args.swap_directory).free
     if args.swap_size:


penberg:
@bentsi The commit message needs to explain the motivation for this change. In this case, the issue does not explain this either.

penberg:
@bentsi Yes, I understand what you are fixing, but I don't understand _why_. Looks like some consumer of the script does not like the return value, and errors out, right? But why is _that_ consumer even calling `scylla_setup_swap` if swap is already set up? Also, why is it ok to silently return? What if the swap configuration is not what `scylla_setup_swap` would have done?

Basically, what I am after here is that why is this the correct place to fix this? The risk here is that we break existing systems.

/cc @avikivity

bentsi:
@penberg right, now I see the issue: we run `scylla_setup` without `--no-swap-setup` when we create GCE image...
But if the script is intended to run by a customer separately, I don't think it is a correct behavior to fail when the swap exists.
@syuu1228 The function `swap_exists()` should probably check not only if the swap exists but has the recommended size
I am closing this.

PR #7528, opened 2020-11-02
repair: Use single writer for all followers

### LINKED ISSUE #7525: Reduce number of sstables generated by repair
Currently, repair master create one writer for each follower to write
rows from follower to sstables. That are RF - 1 writers in total. Each
writer creates 1 sstable for the range repaired, usually a vnode range.
Those sstables for a given vnode range are disjoint. 

To reduce the compaction work, we can create one writer for all the
followers. This reduces the number of sstables generated by repair
significantly to one per vnode range from RF - 1 per vnode range.

asias:
repair: Use single writer for all followers

Currently, repair master create one writer for each follower to write
rows from follower to sstables. That are RF - 1 writers in total. Each
writer creates 1 sstable for the range repaired, usually a vnode range.
Those sstables for a given vnode range are disjoint.

To reduce the compaction work, we can create one writer for all the
followers. This reduces the number of sstables generated by repair
significantly to one per vnode range from RF - 1 per vnode range.

Fixes #7525

### DIFF
--- a/repair/row_level.cc
+++ b/repair/row_level.cc
@@ -336,11 +336,14 @@ private:
     }
 };
 
+using is_dirty_on_master = bool_class<class is_dirty_on_master_tag>;
+
 class repair_row {
     std::optional<frozen_mutation_fragment> _fm;
     lw_shared_ptr<const decorated_key_with_hash> _dk_with_hash;
     std::optional<repair_sync_boundary> _boundary;
     std::optional<repair_hash> _hash;
+    is_dirty_on_master _dirty_on_master;
     lw_shared_ptr<mutation_fragment> _mf;
 public:
     repair_row() = default;
@@ -348,12 +351,17 @@ public:
             std::optional<position_in_partition> pos,
             lw_shared_ptr<const decorated_key_with_hash> dk_with_hash,
             std::optional<repair_hash> hash,
+            is_dirty_on_master dirty_on_master,
             lw_shared_ptr<mutation_fragment> mf = {})
             : _fm(std::move(fm))
             , _dk_with_hash(std::move(dk_with_hash))
             , _boundary(pos ? std::optional<repair_sync_boundary>(repair_sync_boundary{_dk_with_hash->dk, std::move(*pos)}) : std::nullopt)
             , _hash(std::move(hash))
+            , _dirty_on_master(dirty_on_master)
             , _mf(std::move(mf)) {
+    }
+    lw_shared_ptr<mutation_fragment>& get_mutation_fragment_ptr () {
+        return _mf;
     }
     mutation_fragment& get_mutation_fragment() {
         if (!_mf) {
@@ -393,6 +401,9 @@ public:
             throw std::runtime_error("empty hash");
         }
         return *_hash;
+    }
+    is_dirty_on_master dirty_on_master() const {
+        return _dirty_on_master;
     }
 };
 
@@ -740,6 +751,7 @@ private:
     tracker_link_type _tracker_link;
     row_level_repair* _row_level_repair_ptr;
     std::vector<repair_node_state> _all_node_states;
+    is_dirty_on_master _dirty_on_master = is_dirty_on_master::no;
 public:
     std::vector<repair_node_state>& all_nodes() {
         return _all_node_states;
@@ -1152,7 +1164,7 @@ private:
             return stop_iteration::no;
         }
         auto hash = do_hash_for_mf(*_repair_reader.get_current_dk(), mf);
-        repair_row r(freeze(*_schema, mf), position_in_partition(mf.position()), _repair_reader.get_current_dk(), hash);
+        repair_row r(freeze(*_schema, mf), position_in_partition(mf.position()), _repair_reader.get_current_dk(), hash, is_dirty_on_master::no);
         rlogger.trace("Reading: r.boundary={}, r.hash={}", r.boundary(), r.hash());
         _metrics.row_from_disk_nr++;
         _metrics.row_from_disk_bytes += r.size();
@@ -1383,14 +1395,57 @@ private:
             // and _working_row_buf are ordered.
             utils::merge_to_gently(_working_row_buf, row_diff,
                  [this] (const repair_row& x, const repair_row& y) { return _cmp(x.boundary(), y.boundary()) < 0; });
+            for (auto& r : row_diff) {
+                thread::maybe_yield();
+                _working_row_buf_combined_hash.add(r.hash());
+            }
         }
         if (update_hash_set) {
             _peer_row_hash_sets[node_idx] = boost::copy_range<repair_hash_set>(row_diff |
                     boost::adaptors::transformed([] (repair_row& r) { thread::maybe_yield(); return r.hash(); }));
         }
-        do_apply_rows(std::move(row_diff), node_idx, update_buf).get();
+        // Repair rows in row_diff will be flushed to disk by flush_rows_in_working_row_buf,
+        // so we skip calling do_apply_rows here.
+        _dirty_on_master = is_dirty_on_master::yes;
+    }
+public:
+    // Must run inside a seastar thread
+    void flush_rows_in_working_row_buf() {
+        if (_dirty_on_master) {
+            _dirty_on_master = is_dirty_on_master::no;
+        } else {
+            return;
+        }
+        auto cmp = position_in_partition::tri_compare(*_schema);
+        lw_shared_ptr<mutation_fragment> last_mf;
+        lw_shared_ptr<const decorated_key_with_hash> last_dk;
+        for (auto& r : _working_row_buf) {
+            thread::maybe_yield();
+            if (!r.dirty_on_master()) {
+                continue;
+            }
+            _repair_writer->create_writer(_db, node_idx);
+            auto mf = r.get_mutation_fragment_ptr();
+            const auto& pk = r.get_dk_with_hash()->dk.key();
+            if (last_mf && last_dk &&
+                    cmp(last_mf->position(), mf->position()) == 0 &&
+                    pk.legacy_equal(*_schema, last_dk->dk.key()) &&
+                    last_mf->mergeable_with(*mf)) {
+                last_mf->apply(*_schema, std::move(*mf));
+            } else {
+                if (last_mf && last_dk) {
+                    _repair_writer->do_write(std::move(last_dk), std::move(*last_mf)).get();
+                }
+                last_mf = mf;
+                last_dk = r.get_dk_with_hash();
+            }
+        }
+        if (last_mf && last_dk) {
+            _repair_writer->do_write(std::move(last_dk), std::move(*last_mf)).get();
+        }
     }
 
+private:
     future<>
     apply_rows_on_follower(repair_rows_on_wire rows) {
         if (rows.empty()) {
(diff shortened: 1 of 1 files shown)


asias:
@avikivity @raphaelsc FYI, this reduces the number of sstables generated by repair / node operations significantly.

repair/row_level.cc
```diff
@@ -342,18 +342,24 @@ class repair_row {
...
+            lw_shared_ptr<mutation_fragment> mf = {},
+            bool flush_to_disk = false)
             : _fm(std::move(fm))
             , _dk_with_hash(std::move(dk_with_hash))
             , _boundary(pos ? std::optional<repair_sync_boundary>(repair_sync_boundary{_dk_with_hash->dk, std::move(*pos)}) : std::nullopt)
             , _hash(std::move(hash))
-            , _mf(std::move(mf)) {
+            , _mf(std::move(mf))
+            , _flush_to_disk(flush_to_disk) {
+    }
```
avikivity, line 360: I don't understand this flush_to_disk. Is it needed to distinguish between rows that are unequal on the repair master before the operation and after the operation, and so need to be written?

repair/row_level.cc
```diff
@@ -1383,14 +1393,60 @@ class repair_meta {
...
+            auto mf = r.get_mutation_fragment_ptr();
+            if (last_mf && cmp(last_mf->position(), mf->position()) == -1 && last_mf->mergeable_with(*mf)) {
+                last_mf->apply(*_schema, std::move(*mf));
+            } else {
+                last_mf = mf;
+                mfs.push_back(dk_and_mf{r.get_dk_with_hash(), mf});
+            }
+        }
+        while (!mfs.empty()) {
+            auto& dk_mf = mfs.front();
```
avikivity, line 1430: It's nicer to write the mf (it is really a dangerous acronym) in the loop above. As soon as we discover that mf is not mergeable with last_mf, we can write last_mf. This makes the writing parallel with the computation.

repair/row_level.cc
asias: It is used to distinguish if the row is recieved from peers. Repair master only receive rows from peers that the master does not have. So if a row is recieved from master, it needs to write to disk.

repair/row_level.cc
asias: Why is mf a dangerous acronym? I will take a look. I thought about it too because it saves a loop.

avikivity: https://www.youtube.com/watch?v=c0LBi1MHoaU

asias:
Changes in v2 (cc711bbbc1bcf29e7b308132d63a389b8f28f11c):
- write the row in the the same loop
- compare partition key as well

repair/row_level.cc
asias: It is a terrible interpretation of this lovely data structure ;-)

repair/row_level.cc
```diff
@@ -1383,14 +1393,58 @@ class repair_meta {
...
+            thread::maybe_yield();
+            if (!r.flush_to_disk()) {
+                continue;
+            }
+            _repair_writer->create_writer(_db, node_idx);
+            auto mf = r.get_mutation_fragment_ptr();
+            const auto& pk = r.get_dk_with_hash()->dk.key();
+            if (last_mf &&
+                last_dk &&
+                cmp(last_mf->position(), mf->position()) == -1 &&
```
denesb, line 1421: I don't understand why we are merging when `last_mf < mf`?

repair/row_level.cc
```diff
@@ -1383,14 +1393,58 @@ class repair_meta {
...
+                pk.legacy_equal(*_schema, last_dk->dk.key()) &&
+                last_mf->mergeable_with(*mf)) {
```
denesb, line 1423: Please indent these lines with double indtent, really hard to see where the if header ends and where the body starts.

repair/row_level.cc
asias: It is an accident. Inside vim and ctrl-x. I will fix it up.

repair/row_level.cc
asias: I prefer to align to first the item in if because it looks nicer. I agree double indent is easier to read though.

asias:
Changes in v3 (074a96e6066116dc132a6d82c236735cf3f0c41c):
- Double indent inside if
- Fix cmp == 0

repair/row_level.cc
```diff
@@ -1383,14 +1393,57 @@ class repair_meta {
             // and _working_row_buf are ordered.
             utils::merge_to_gently(_working_row_buf, row_diff,
                  [this] (const repair_row& x, const repair_row& y) { return _cmp(x.boundary(), y.boundary()) < 0; });
+            for (auto& r : row_diff) {
+                thread::maybe_yield();
+                _working_row_buf_combined_hash.add(r.hash());
```
nyh, line 1389: I'm afraid I don't understand what this patch is doing.
If we (the master) get modifications from multiple remote nodes, how do we "combine" these modifications? What's the purpose of this "combined hash", and how can we "add" the hash of one diff - we may, for example, get the same missing column from two different remote nodes, we can't add them twice...
I'm probably missing something on how the new algorithm works.

repair/row_level.cc
```diff
@@ -1383,14 +1393,57 @@ class repair_meta {
...
+        }
+        auto cmp = position_in_partition::tri_compare(*_schema);
+        lw_shared_ptr<mutation_fragment> last_mf;
+        lw_shared_ptr<const decorated_key_with_hash> last_dk;
+        for (auto& r : _working_row_buf) {
+            thread::maybe_yield();
+            if (!r.flush_to_disk()) {
+                continue;
+            }
+            _repair_writer->create_writer(_db, node_idx);
```
nyh: I'm not familiar with this code, so maybe this is a stupid question, but I'll ask anyway: Why create a write for each row in the loop?

repair/row_level.cc
```diff
@@ -524,14 +524,14 @@ class repair_writer : public enable_lw_shared_from_this<repair_writer> {
     uint64_t _estimated_partitions;
     size_t _nr_peer_nodes;
     // Needs more than one for repair master
```
nyh, line 526: This comment is no longer correct? (this patch changes the vector to a single one)

repair/row_level.cc
asias: The combined hash is a hash of hashes in _working_row_buf. It is used to check if the rows inside _working_row_buf are identical. 

The new algorithm will not get the same row from more than one node.

repair/row_level.cc
asias: It creates write only if the writer is not created.

repair/row_level.cc
asias: Yes. We should drop this comment. Thanks.

asias:
Changes in v4 (b42356844702bf4ec2b74ca6410a56f468151bf5)
- Drop a comment about that we need multiple writers

repair/row_level.cc
avikivity: Ok. Suggestions for readability:
 - use bool_class so it's clear on the caller side
 - rename to dirty_on_master or dirty_locally, this informs us about the row, not what you want to do with it. `flush_to_disk` doesn't tell us why
 - undefault the parameter, defaults are often dangerous, and here there is no natural reason for a row to be not-dirty

repair/row_level.cc
avikivity: A better name is maybe_create_writer() then (but let's not do it now).

repair/row_level.cc
```diff
@@ -1383,14 +1393,57 @@ class repair_meta {
...
+            thread::maybe_yield();
+            if (!r.flush_to_disk()) {
+                continue;
+            }
+            _repair_writer->create_writer(_db, node_idx);
+            auto mf = r.get_mutation_fragment_ptr();
+            const auto& pk = r.get_dk_with_hash()->dk.key();
+            if (last_mf && last_dk &&
+                    cmp(last_mf->position(), mf->position()) == 0 &&
+                    pk.legacy_equal(*_schema, last_dk->dk.key()) &&
```
avikivity, line 1420: why legacy_equal?

repair/row_level.cc
```diff
@@ -1383,14 +1393,57 @@ class repair_meta {
...
+                    last_mf->mergeable_with(*mf)) {
+                last_mf->apply(*_schema, std::move(*mf));
```
avikivity, line 1422: This changes last_mf in place. I expect it's unused other than here, but please verify.

repair/row_level.cc
```diff
@@ -1383,14 +1393,57 @@ class repair_meta {
...
+                    _repair_writer->do_write(std::move(last_dk), std::move(*last_mf)).get();
+                }
+                last_mf = mf;
+                last_dk = r.get_dk_with_hash();
+            }
+        }
+        if (last_mf && last_dk) {
+            _repair_writer->do_write(std::move(last_dk), std::move(*last_mf)).get();
+        }
     }
```
avikivity, line 1434: This whole thing duplicates combined_reader. On one hand it's duplication, on the other hand we'd have to write a reader to consume from the repair buffers and filter out non-dirty rows, so maybe that's even more work.

repair/row_level.cc
asias: Sure.

repair/row_level.cc
asias: Do we have better alternative now?

repair/row_level.cc
asias: The only reason we attach a mf to repair_row is that we will write it to disk. Later, we only use the frozen version to send from master to followers. So it is fine to apply in place.

repair/row_level.cc
asias: I considered using combined reader. But we can not use it because master gets data from peer nodes one after another, so the first item in the queue is not always present which combined reader needs.

repair/row_level.cc
asias: Ok. Done. Fixed in latest version.

asias:
Changes in v5 db28efb28a4e610207b5418f8d54e72a73cebf5e:
- Change flush to disk to dirty on master
- Use bool class
- Remove the default for dirty on master
- Rebase to master

repair/row_level.cc
avikivity: combined_reader will wait for data to arrive.

repair/row_level.cc
tgrabiec: > The new algorithm will not get the same row from more than one node.

Why is it ok? Each node may contain different writes to the same row and repair should merge all of them.

repair/row_level.cc
asias: Yes it can wait. However, say we have 3 followers. The master is able to start to write only after it receive all data from 2 of the followers. One follower can have up to 32MB of data.

repair/row_level.cc
asias: Nice. I will have a follow up patch to change both usage to the new comparator.

repair/row_level.cc
asias: same row = same row with same content

repair/row_level.cc
tgrabiec: How do we avoid waiting under the same scenario with this PR?

repair/row_level.cc
tgrabiec: I don't get it. Followers can have a given row (identified by primary key) with different content. How can you avoid merging the rows from multiple followers if you direct them into a single fragment stream? Rows cannot repeat in the stream.

repair/row_level.cc
tgrabiec: We probably don't have this covered by unit tests, do we?

PR #7529, opened 2020-11-02
moving avarage rate: Keep computed rates in zero until they are meaningful

### LINKED ISSUE #4469: multi_mvs_on_different_populated_base_tables_test failed: Infinite double value is not supported
scylla ee9f3388f6b5e29e682bf78c085b8495ca839602
Seen in [dtest-release/109/testReport/materialized_views_test/TestMaterializedViews/multi_mvs_on_different_populated_base_tables_test](http://jenkins.cloudius-systems.com:8080/view/master/job/scylla-master/job/dtest-release/109/testReport/materialized_views_test/TestMaterializedViews/multi_mvs_on_different_populated_base_tables_test/):
```
Error Message
Nodetool command '/jenkins/workspace/scylla-master/dtest-release/scylla/resources/cassandra/bin/nodetool -h localhost -p 7488 cfstats' failed; exit status: 1; stdout: nodetool: Scylla API server HTTP GET to URL '/column_family/metrics/write_latency/moving_average_histogram/ks:tm_table0_mv_0' failed: Infinite double value is not supported

Stacktrace
  File "/usr/lib64/python2.7/unittest/case.py", line 367, in run
    testMethod()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/materialized_views_test.py", line 742, in multi_mvs_on_different_populated_base_tables_test
    increase_rows=10, populated_table=True)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/materialized_views_test.py", line 754, in _multi_mvs_on_different_base_tables_multi_ks
    run_in_parallel(proc_functions)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/scylla_tools.py", line 846, in run_in_parallel
    results = [task.result() for task in tasks]
  File "/usr/lib/python2.7/site-packages/concurrent/futures/_base.py", line 455, in result
    return self.__get_result()
  File "/usr/lib/python2.7/site-packages/concurrent/futures/thread.py", line 63, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/materialized_views_test.py", line 780, in _multi_mvs_on_different_base_tables
    func()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/materialized_views_test.py", line 760, in _prefill_base_tables
    base_table.prefill_table(prefill)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/scylla_tools.py", line 389, in prefill_table
    flush_by_node(self.cluster)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/scylla_tools.py", line 201, in flush_by_node
    node.flush()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/scylla_node.py", line 840, in flush
    self._wait_no_pending_flushes()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/scylla_node.py", line 833, in _wait_no_pending_flushes
    result = wait_for(no_pending_flushes, timeout=wait_timeout, step=1.0)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/scylla_node.py", line 46, in wait_for
    output = func()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/scylla_node.py", line 823, in no_pending_flushes
    stdout, _ = self.nodetool('cfstats')
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/node.py", line 714, in nodetool
    raise NodetoolError(" ".join(args), exit_status, stdout, stderr)
```

This exception is generated here:
```
sstring formatter::to_json(double d) {
    if (std::isinf(d)) {
        throw out_of_range("Infinite double value is not supported");
```

/cc @eliransin @psarna @nyh

eliransin:
When computing moving average rates too early after startup, the
rate can be infinite, this is simply because the sample interval
since the system started is too small to generate meaningful results.
Here we check for this situation and keep the rate at 0 if it happens
to signal that there are still no meaningful results.
This incident is unlikely to happen since it can happen only during a
very small time window after restart, so we add a hint to the compiler
to optimize for that in order to have a minimum impact on the normal
usecase.

Fixes #4469

### DIFF
--- a/utils/histogram.hh
+++ b/utils/histogram.hh
@@ -256,9 +256,15 @@ public:
 
     rate_moving_average rate() const {
         rate_moving_average res;
-        if (_count > 0) {
-            double elapsed = std::chrono::duration_cast<std::chrono::seconds>(latency_counter::now() - start_time).count();
+        double elapsed = std::chrono::duration_cast<std::chrono::seconds>(latency_counter::now() - start_time).count();
+        // We condition also in elapsed because it can happen that the call
+        // for the rate calculation was performed too early and will not yield
+        // meaningful results (i.e mean_rate is infinity) so the best thing is
+        // to return 0 as it best reflects the state.
+        if ((_count > 0) && (elapsed >= 1.0)) [[likely]] {
             res.mean_rate = (_count / elapsed);
+        } else {
+            res.mean_rate = 0;
         }
         res.count = _count;
         for (int i = 0; i < 3; i++) {


eliransin:
@amnonh can you review please. This is something we discussed a long time ago.

utils/histogram.hh
```diff
@@ -258,7 +258,17 @@ public:
         rate_moving_average res;
         if (_count > 0) {
```
amnonh, line 259: Typically, _count will be greater than zero, you can move the duration calculation outside of the if, and add elapsed > 1.0 (or any other value for that matter) to be part of the if condition.

amnonh: all in all, it looks good and cool use of C++20 attributes.

eliransin: @amnonh 
Just to see that we are on the same page, do you think that returning 0 until we have an interval of at least one second or any finite an reasonably large duration (0.5,0.25...) is an acceptable solution? (this will ensure that _count/elapse will never be infinite). If so , unfortunately I will have to drop the cool c++ 20 feature or to add a [[likely]]
attribute to the new consolidation condition :slightly_smiling_face:  but do you agree to that solution?

utils/histogram.hh
```diff
@@ -258,7 +258,17 @@ public:
...
             double elapsed = std::chrono::duration_cast<std::chrono::seconds>(latency_counter::now() - start_time).count();
-            res.mean_rate = (_count / elapsed);
+            double mean_rate = (_count / elapsed);            
+            [[unlikely]]
+            if ((elapsed < 1.0) && std::isinf(mean_rate)) {
```
avikivity, line 263: This is wrong, it says the if statement is unlikely (both branches).

eliransin: Thanks, I will move it to the appropriate place.

PR #7531, opened 2020-11-02
build: switch to clang as the default compiler

avikivity:
Clang brings us working support for coroutines, which are
needed for Raft and for code simplification.

perf_simple_query as well as full system tests show no
significant performance regression.

Test: unit(dev, release, debug)

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -478,9 +478,9 @@ arg_parser.add_argument('--ldflags', action='store', dest='user_ldflags', defaul
                         help='Extra flags for the linker')
 arg_parser.add_argument('--target', action='store', dest='target', default=default_target_arch(),
                         help='Target architecture (-march)')
-arg_parser.add_argument('--compiler', action='store', dest='cxx', default='g++',
+arg_parser.add_argument('--compiler', action='store', dest='cxx', default='clang++',
                         help='C++ compiler path')
-arg_parser.add_argument('--c-compiler', action='store', dest='cc', default='gcc',
+arg_parser.add_argument('--c-compiler', action='store', dest='cc', default='clang',
                         help='C compiler path')
 add_tristate(arg_parser, name='dpdk', dest='dpdk',
                         help='Use dpdk (from seastar dpdk sources) (default=True for release builds)')


gleb-cloudius:
On Mon, Nov 02, 2020 at 04:05:45AM -0800, Avi Kivity wrote:
> Clang brings us working support for coroutines, which are
> needed for Raft and for code simplification.
> 
But Raft will not compile without: https://github.com/scylladb/scylla-dev/commit/c45ac68eb3cc235d32f7aeb5f2d7db41927169c2

--
			Gleb.

haaawk:
Should this change include a new entry in install-dependencies.sh with clang dependency?

nyh (commented):
Sic transit gloria mundi!
@avikivity I can commit this and we'll deal with the fallout later, but I wonder if we shouldn't have discussed this a bit on the mailing list before this patch. Some questions I have:
1. With what version of clang are we known to work? For example, I have clang 10.0.1 (on Fedora 32). Is this new enough, or must we tell everyone to upgrade?
2. Did you test performance of Scylla on micro and macro benchmarks to see clang's optimizer is on par with gcc's? 
3. Can you say something about the build time - is clang faster, slower, or same as gcc in that respect?

nyh:
> Should this change include a new entry in install-dependencies.sh with clang dependency?

Avi already did this, in commit 17226f2f6cf3f21821f4044bca93712017f747a8, a few days ago.

avikivity:
> Sic transit gloria mundi!
> @avikivity I can commit this and we'll deal with the fallout later, but I wonder if we shouldn't have discussed this a bit on the mailing list before this patch. Some questions I have:
> 
> 1. With what version of clang are we known to work? For example, I have clang 10.0.1 (on Fedora 32). Is this new enough, or must we tell everyone to upgrade?

Earlier I tested with clang 10. Later iterations used 11, so maybe a dependency on 11 crept in. There may be problems in release mode related to `-mllvm -inline-threshold=2500`, but we can remove them if we detect clang 10.

> 2. Did you test performance of Scylla on micro and macro benchmarks to see clang's optimizer is on par with gcc's?

Yes, I noted this in the changelog.

> 3. Can you say something about the build time - is clang faster, slower, or same as gcc in that respect?

Yes, clang is faster, slower, or the same as gcc wrt compile time (but I don't know which).

PR #7539, opened 2020-11-03
Mark CDC as GA

haaawk:
CDC is ready to be a non-experimental feature so remove the experimental flag for it.
Also, guard Alternator Streams with their own experimental flag. Previously, they were using CDC experimental flag as they depend on CDC.

Tests: unit(dev)

### DIFF
--- a/cdc/generation.cc
+++ b/cdc/generation.cc
@@ -181,12 +181,12 @@ const std::vector<token_range_description>& topology_description::entries() cons
 class topology_description_generator final {
     const db::config& _cfg;
     const std::unordered_set<dht::token>& _bootstrap_tokens;
-    const locator::token_metadata& _token_metadata;
+    const locator::token_metadata_ptr _tmptr;
     const gms::gossiper& _gossiper;
 
     // Compute a set of tokens that split the token ring into vnodes
     auto get_tokens() const {
-        auto tokens = _token_metadata.sorted_tokens();
+        auto tokens = _tmptr->sorted_tokens();
         auto it = tokens.insert(
                 tokens.end(), _bootstrap_tokens.begin(), _bootstrap_tokens.end());
         std::sort(it, tokens.end());
@@ -201,7 +201,7 @@ class topology_description_generator final {
         if (_bootstrap_tokens.contains(end)) {
             return {smp::count, _cfg.murmur3_partitioner_ignore_msb_bits()};
         } else {
-            auto endpoint = _token_metadata.get_endpoint(end);
+            auto endpoint = _tmptr->get_endpoint(end);
             if (!endpoint) {
                 throw std::runtime_error(
                         format("Can't find endpoint for token {}", end));
@@ -235,11 +235,11 @@ public:
     topology_description_generator(
             const db::config& cfg,
             const std::unordered_set<dht::token>& bootstrap_tokens,
-            const locator::token_metadata& token_metadata,
+            const locator::token_metadata_ptr tmptr,
             const gms::gossiper& gossiper)
         : _cfg(cfg)
         , _bootstrap_tokens(bootstrap_tokens)
-        , _token_metadata(token_metadata)
+        , _tmptr(std::move(tmptr))
         , _gossiper(gossiper)
     {}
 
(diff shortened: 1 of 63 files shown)


avikivity:
Dequeued. @bhalevy reports that it fails in dtest murmur3_partitioner_test.

slivne:
@haaawk ping

cdc/generation.cc
```diff
@@ -310,7 +310,14 @@ db_clock::time_point make_new_cdc_generation(
     auto ts = db_clock::now() + (
             (for_testing || ring_delay == milliseconds(0)) ? milliseconds(0) : (
                 2 * ring_delay + duration_cast<milliseconds>(generation_leeway)));
-    sys_dist_ks.insert_cdc_topology_description(ts, std::move(gen), { tm.count_normal_token_owners() }).get();
+    try {
+        sys_dist_ks.insert_cdc_topology_description(ts, std::move(gen), { tm.count_normal_token_owners() }).get();
+    } catch (...) {
```
kbr-, line 315: We only want to make this optional in the replace-node or rolling-upgrade case, but not in the node-bootstrap case. For node bootstrap changing the generation is mandatory.

kbr-: Hence I think this `try`+`catch` should be outside the function, in the `if (!_cdc_streams_ts) {` path (line 552 in storage service).

service/storage_service.cc
```diff
@@ -892,24 +897,19 @@ void storage_service::bootstrap() {
...
-                    _sys_dist_ks.local(), get_ring_delay(), _for_testing);
-        } else {
-            // We should not be able to join the cluster if other nodes support CDC but we don't.
-            // The check should have been made somewhere in prepare_to_join (`check_knows_remote_features`).
-            assert(!_feature_service.cluster_supports_cdc());
-        }
+        _cdc_streams_ts = cdc::make_new_cdc_generation(db().local().get_config(),
+                _bootstrap_tokens, _token_metadata, _gossiper,
+                _sys_dist_ks.local(), get_ring_delay(), _for_testing);
+        assert(_cdc_streams_ts);
```
kbr-, line 912: This assert may fire in normal circumstances (not due to a bug), so it should not be an assert.

kbr-: Instead, the node should fail bootstrapping with a message. We should just let the exception from `make_new_cdc_generation` fly out in this case.

cdc/generation.cc
haaawk: done

service/storage_service.cc
haaawk: done

haaawk:
I fixed @kbr- comments and the failing partitioner test.
BYO is running here: https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/1065/

Nothing changed in the previous version. Only a new commit was added to fix the failing test.

avikivity:
@nyh since you shepherded the work maybe you should have the honor of committing the GA series.

haaawk:
Please wait for BYO to finish though.

test/boost/cdc_test.cc
```diff
@@ -46,9 +46,6 @@ static cql_test_config mk_cdc_test_config() {
     auto ext = std::make_shared<db::extensions>();
```
kbr-, line 46: This function should not be necessary at all now.
It was introduced in b130b76274fe7414b01ada9b41554c78540454e1 to speed up tests.
(I don't recall if the slowdown in startup due to enabling CDC wasn't fixed at some point)

haaawk: Sounds great but also deserves a PR on its own ;).

haaawk:
First BYO failed due to some flaky dtest (migration_test.py:TTLWithMigrate.big_table_with_ttls_test). Here's another run -> https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/1067/

haaawk:
All BYO tests have passed. @nyh @avikivity would you mind merging?

avikivity:
I'll leave it to @nyh who shepherded the whole process.

PR #7542, opened 2020-11-03
alternator::streams: Workaround fix for apparent code gen bug in seq_…

### LINKED ISSUE #7325: Alternator Streams test fails when building on clang
@avikivity reported on the mailing list tha scylla built with clang (many unapplied patched needed) passes all unit tests except `alternator/run`, which has one failure - `test_streams_after_sequence_number`:
```
______________________ test_streams_after_sequence_number 
...
>                   assert len(response['Records']) == 1
E                   AssertionError: assert 2 == 1
E                    +  where 2 = len([{'dynamodb':
{'ApproximateCreationDateTime': datetime.datetime(2020, 10, 4, 13, 40,
8, tzinfo=tzlocal()), 'Keys':
{'c...d67e574682740000d1:f8de1e30-062d-11eb-6e12-418927c0cf0b',
'eventName': 'MODIFY', 'eventSource': 'scylladb:alternator'}])

test_streams.py:1019: AssertionError
```

I tried to explain the context of this test and the failure:

> This test checks Alternator Streams (i.e., CDC) . The test writes two items to the table, and reads two items from the stream. This part worked well. Then, it asks to read from the same stream again, but not from the start - it gives the sequence number of the first change, and asks to read *after* it. The result should be just one item, but for some reason we got two so the test failed.
>
> Calle, if I remember correctly, in the past we had a similar bug when your sequence number serialization/deserialization code had a bug and relied on some implied constructor which didn't do what you thought it did - maybe there is a similar bug in the clang build?
>
> The relevant code is probably in alternator/streams.cc, struct sequence_number, and possibly struct shard_iterator, e.g., shard_iterator::shard_iterator(const sstring& s, size_t i). I don't know how to debug it further in my mind without adding printouts and seeing whether the sequence number was printed wrong, parsed wrong, or what actually went wrong...

CC @elcallio

elcallio:
…number

Fixes #7325

When building with clang on fedora32, calling the string_view constructor
of bignum generates broken ID:s (i.e. parsing borks). Creating a temp
std::string fixes it.

### DIFF
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -290,7 +290,9 @@ struct sequence_number {
 sequence_number::sequence_number(std::string_view v) 
     : uuid([&] {
         using namespace boost::multiprecision;
-        uint128_t tmp{v};
+        // workaround for weird clang 10 bug when calling constructor with
+        // view directly.
+        uint128_t tmp{std::string(v)};
         // see above
         return utils::UUID_gen::get_time_UUID_raw(uint64_t(tmp >> 64), uint64_t(tmp & std::numeric_limits<uint64_t>::max()));
     }())


nyh:
Wow, in 10 seconds I would have sent my own patch which is already in my repository ;-) I'll consider this one instead.....

haaawk (approved):
LGTM

haaawk:
Thanks @elcallio for saving the world :)

PR #7554, opened 2020-11-05
dbuild: Replace stray use of `docker` with `$tool`

michoecho:
Instead of invoking `$tool`, as is done everywhere else in dbuild,
kill_it() invoked `docker` explicitly. This was slightly breaking the
script for DBUILD_TOOL other than `docker`.


avikivity:
Queued, but please add your full name in github configuration. ./scripts/pull_github_pr.sh uses it for multi-patch series.

michoecho:
Done.

PR #7557, opened 2020-11-05
configure.py: Use product prefix in dependencies

penberg:
The submodules now generate packages with a product prefix:

  - https://github.com/scylladb/scylla-jmx/commit/6174a479249518a720709c6a1d59f5bfa5780431

  - https://github.com/scylladb/scylla-tools-java/commit/e97c1060476085d25f9709b94df8ab48eac03c39

  - https://github.com/scylladb/scylla-python3/commit/1763a1a8348dac7d2d8e4930a5550ffec6537cf1

Let's adjust configure.py to generate dependencies with the product
prefix too to unbreak non-"scylla" product builds.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1286,6 +1286,8 @@ if status != 0:
     print('Version file generation failed')
     sys.exit(1)
 
+file = open(f'{outdir}/SCYLLA-PRODUCT-FILE', 'r')
+scylla_product = file.read().strip()
 file = open(f'{outdir}/SCYLLA-VERSION-FILE', 'r')
 scylla_version = file.read().strip()
 file = open(f'{outdir}/SCYLLA-RELEASE-FILE', 'r')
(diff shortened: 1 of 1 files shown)


avikivity:
Looks reasonable.

penberg:
@avikivity @eliransin ping

eliransin (commented):
I think this is not enough for products other than scylla, have you tested it with a product name
different than scylla (lets say the randomly chosen product name scylla-enterprise? :wink: )
The script unified/build_unified.sh for example will fail if not taken care of.
If you tested it and for some reason this script doesn't gets called during the build so LGTM. If not so you can either test it, or give me an hour or two and I think I can publish something that will seamlessly work when merged to other products.

penberg:
I am closing as we merged a better pull request from @eliransin: https://github.com/scylladb/scylla/pull/7581

Issue #6403, opened 2020-05-08
Identify nodes by UUID for coordination and messaging instead of IPs


tgrabiec:
Scylla allows replacing a dead node with a node which has the same IP as the dead node.

During replace operation we don't want messages targeted to the old node to arrive to the new node. This could violate consistency invariants, as the new node may not yet be fully synchronized. Refs https://github.com/scylladb/scylla/issues/4648

At the same time, we do want some messages to arrive to the new node. For example, data synchronization algorithm requires that current requests arrive to the new node while it's being synchronized.

I propose that we identify nodes by host IDs in the ring instead of IPs. Messaging layer will allow sending to a host ID instead of to an IP (it will still allow to send by IP). Messaging layer will keep the translation. When a connection is established between nodes, messaging layer learns about node's host ID during early negotiation before any message can be sent through that connection. This will make sure that we never send any message to the wrong node.

This will also remove the need to have special-casing in the coordinator to disambiguate between IPs. This means we won't need `abstract_replication_strategy::get_natural_endpoints_without_node_being_replaced()`.

dorlaor:
It's a good idea. Can you pls dig into C*'s JIRA to see if they
have such an enhancement request? Worth to contribute
such a request if one doesn't exist.

If this has effects on our configuration file, worth to plan it and
also have a compatibility/transition mode with IPs

kostja:
Do you suggest UUIDs are specified by the user or created transparently to the user by Scylla? In the latter case, how do you suggest obsolete UUIDs are expired?

tgrabiec:
@kostja Transparently. We already assign such IDs (see `system_keyspace::get_local_host_id()`). 

By expiring, do you mean garbage collecting unused ID->IP mappings? We only need to track IDs which are present in the ring (token_metadata), only those will be contacted by coordinators. We already persist this mapping in system tables. When a node is removed from the ring, the mapping goes away.

There'd be also an in-memory mapping kept by messaging service, this goes away when a node disconnects.

dorlaor:
https://issues.apache.org/jira/browse/CASSANDRA-15823

kbr-:
I support this cause.

We could take this even further: decouple the messaging layer completely from other parts of the system, make it pluggable, like `raft::server` has a pluggable `rpc` interface.

This could allow us to do fun stuff like starting multiple Scylla nodes that communicate through a fake network... in a single process. Would make it possible to run cluster tests inside the unit testing framework!

[...]

PR #7561, opened 2020-11-06
storage_service, gossip: Do not announce bootstrapping nodes with inc…

tgrabiec:
…omplete state

Currently, bootstrapping (new node or replacing existing node) has the following phases:

  1) start gossiper
  2) wait for gossip to settle
  3) set STATUS, TOKENS, etc

This is problematic when replacing a node with the same ip because
right after step 1 the old nodes will mark the replacing node as alive
and can start sending traffic to the node. However, the node is not
yet marked as doing the replacing by the old nodes, and should not
receive requests. It is not yet prepared for receiving requests
until after step 3. For example, old nodes should not treat the new 
node as the old replica and send read requests to it because it has no data yet.

To fix the problem, change the gossiper to start in the shadow mode,
so that the bootstrapping node can learn about existing nodes but
existing nodes cannot learn about the bootstrapping node. We run in
this state until after step 3. This way existing nodes will see the
replacing node as alive only after it decides to come out of the shadow,
after its state is settled.

Tested using ccm, replacing a node with both the same and different address.

### DIFF
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -204,7 +204,7 @@ void gossiper::do_sort(utils::chunked_vector<gossip_digest>& g_digest_list) {
 future<> gossiper::handle_syn_msg(msg_addr from, gossip_digest_syn syn_msg) {
     logger.trace("handle_syn_msg():from={},cluster_name:peer={},local={},partitioner_name:peer={},local={}",
         from, syn_msg.cluster_id(), get_cluster_name(), syn_msg.partioner(), get_partitioner_name());
-    if (!this->is_enabled()) {
+    if (!this->is_enabled() || _in_long_shadow_round) {
         return make_ready_future<>();
     }
 
@@ -317,7 +317,7 @@ future<> gossiper::handle_ack_msg(msg_addr id, gossip_digest_ack ack_msg) {
     assert(_msg_processing > 0);
 
     return f.then([this, from = id, ack_msg_digest = std::move(g_digest_list), mp = std::move(mp), g = this->shared_from_this()] () mutable {
-        if (this->is_in_shadow_round()) {
+        if (this->is_in_shadow_round() || _in_long_shadow_round) {
             this->finish_shadow_round();
             // don't bother doing anything else, we have what we came for
             return make_ready_future<>();
@@ -676,6 +676,10 @@ future<> gossiper::apply_state_locally(std::map<inet_address, endpoint_state> ma
     });
 }
 
+void gossiper::come_out_of_shadow() {
+    _in_long_shadow_round = false;
+}
+
 // Runs inside seastar::async context
 void gossiper::remove_endpoint(inet_address endpoint) {
     // do subscribers first so anything in the subscriber that depends on gossiper state won't get confused
@@ -778,9 +782,11 @@ void gossiper::run() {
 
             logger.trace("My heartbeat is now {}", endpoint_state_map[br_addr].get_heart_beat_state().get_heart_beat_version());
             utils::chunked_vector<gossip_digest> g_digests;
-            this->make_random_gossip_digest(g_digests);
+            if (!_in_long_shadow_round) {
+                this->make_random_gossip_digest(g_digests);
+            }
 
-            if (g_digests.size() > 0) {
+            if (g_digests.size() > 0 || _in_long_shadow_round) {
                 gossip_digest_syn message(get_cluster_name(), get_partitioner_name(), g_digests);
 
                 if (_endpoints_to_talk_with.empty()) {
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -590,6 +590,7 @@ void storage_service::join_token_ring(int delay) {
     replicate_to_all_cores().get();
     // start participating in the ring.
     set_gossip_tokens(_bootstrap_tokens, _cdc_streams_ts);
+    _gossiper.come_out_of_shadow();
     set_mode(mode::NORMAL, "node is now in normal status", true);
 
     if (_token_metadata.sorted_tokens().empty()) {
@@ -918,7 +919,8 @@ void storage_service::bootstrap() {
             { gms::application_state::CDC_STREAMS_TIMESTAMP, versioned_value::cdc_streams_timestamp(_cdc_streams_ts) },
             { gms::application_state::STATUS, versioned_value::bootstrapping(_bootstrap_tokens) },
         }).get();
-
+        set_mode(mode::JOINING, "Coming out of the shadow");
+        _gossiper.come_out_of_shadow();
         set_mode(mode::JOINING, format("sleeping {} ms for pending range setup", get_ring_delay().count()), true);
         _gossiper.wait_for_range_setup().get();
     } else {
@@ -931,6 +933,7 @@ void storage_service::bootstrap() {
             { gms::application_state::CDC_STREAMS_TIMESTAMP, versioned_value::cdc_streams_timestamp(_cdc_streams_ts) },
             { gms::application_state::STATUS, versioned_value::hibernate(true) },
         }).get();
+        _gossiper.come_out_of_shadow();
         set_mode(mode::JOINING, format("Wait until peer nodes know the bootstrap tokens of local node"), true);
         _gossiper.wait_for_range_setup().get();
         auto replace_addr = db().local().get_replace_address();
@@ -2043,6 +2046,7 @@ future<> storage_service::start_gossiping(bind_messaging_port do_bind) {
                 ss._gossiper.start_gossiping(utils::get_generation_number(), gms::bind_messaging_port(bool(do_bind))).then([&ss] {
                     ss._initialized = true;
                 }).get();
+                ss._gossiper.come_out_of_shadow();
             }
         });
     });
--- a/gms/gossiper.hh
+++ b/gms/gossiper.hh
@@ -115,6 +115,9 @@ struct gossip_config {
  * of the three above mentioned messages updates the Failure Detector with the liveness information.
  * Upon hearing a GossipShutdownMessage, this module will instantly mark the remote node as down in
  * the Failure Detector.
+ *
+ * Initially, this gossiper will stay silent about what he knows, including himself.
+ * In order to start spreading gossip, call come_out_of_shadow().
  */
 class gossiper : public i_failure_detection_event_listener, public seastar::async_sharded_service<gossiper>, public seastar::peering_sharded_service<gossiper> {
 public:
@@ -226,6 +229,7 @@ private:
     std::map<inet_address, clk::time_point> _expire_time_endpoint_map;
 
     bool _in_shadow_round = false;
+    bool _in_long_shadow_round = true;
 
     std::unordered_map<inet_address, clk::time_point> _shadow_unreachable_endpoints;
     utils::chunked_vector<inet_address> _shadow_live_endpoints;
@@ -524,6 +528,10 @@ public:
     bool is_in_shadow_round() const;
 
     void goto_shadow_round();
+
+    // After this call the gossiper starts gossipping about the current node with other nodes.
+    // Before this, the current node is learning from other nodes but stays silent about himself.
+    void come_out_of_shadow();
 
 public:
     void add_expire_time_for_endpoint(inet_address endpoint, clk::time_point expire_time);


service/storage_service.cc
```diff
@@ -918,7 +919,8 @@ void storage_service::bootstrap() {
             { gms::application_state::CDC_STREAMS_TIMESTAMP, versioned_value::cdc_streams_timestamp(_cdc_streams_ts) },
             { gms::application_state::STATUS, versioned_value::bootstrapping(_bootstrap_tokens) },
         }).get();
-
+        set_mode(mode::JOINING, "Coming out of the shadow");
```
bhalevy, line 922: Why only here and not in all places we come out of shadow?

gms/gossiper.cc
```diff
@@ -317,7 +317,7 @@ future<> gossiper::handle_ack_msg(msg_addr id, gossip_digest_ack ack_msg) {
     assert(_msg_processing > 0);
 
     return f.then([this, from = id, ack_msg_digest = std::move(g_digest_list), mp = std::move(mp), g = this->shared_from_this()] () mutable {
-        if (this->is_in_shadow_round()) {
+        if (this->is_in_shadow_round() || _in_long_shadow_round) {
```
bhalevy, line 320: This is confusing since if the gossiper is not `_in_shadow_round` then there's no need to call `finish_shadow_round`.
It'd be better to leave this unchanged and add:
```
if (_in_long_shadow_round) {
    // explain why we return early
    return make_ready_future<>();
}
```

bhalevy (changes requested):
In general, the change is quite scary. It is not clear to me what needs to be done `_in_shadow_round`, what in _in_long_shadow_round`, and what if neither, or both.
I think this will make maintaining the code potentially harder of what it is today.

asias:
This change has nothing to do with shadow round at all. It is confusing to call it shadow round to confuse with the old shadow round logics in the gossip message handers. It introduce a new gossip working mode, that is silent mode which listens only. 

It is much better than what you proposed earlier (making current verb based shadow round to call the gossip handlers and preparing a state before we can start gossip). 

However, it adds a new gossip working mode. Yes, this can work. But I think it is worse than introducing a new gossip status.

Without this patch, we can get rid of all the special existing shadow round "if" in gossip message handler after some grace period of being compatible with old version.  Now, this patch adds more and we need to keep it forever. It is hard to reason about when we need to be in silent mode and when come out. It is easier to have only one normal gossip mode.

kbr-:
I don't understand the difference between the old shadow round and this new "long shadow" mode.
@asias or @tgrabiec could you explain?
@tgrabiec why can't we use the old shadow round? (maybe I'll understand if the difference is made clearer)

could you argue why the decision about introducing a silent mode? as @asias said, it looks like this could also be solved by the node simply broadcasting a different STATUS (not `BOOT` or `NORMAL`, which cause other nodes to send data to this node).

service/storage_service.cc
tgrabiec: Other places have had other log messages nearby so adding this seemed unnecessary. But we could do it for consistency.

tgrabiec:
> In general, the change is quite scary. It is not clear to me what needs to be done `_in_shadow_round`, what in _in_long_shadow_round`, and what if neither, or both.
> I think this will make maintaining the code potentially harder of what it is today.

It's maybe a documentation issue.

_in_shadow_round  and _in_long_shadow_round are mutually exclusive. The former runs before the gossiper is started, the latter runs after it is started.

_long_shadow_round is like normal gossip except that the current node doesn't actively gossip information, which means that it:
   1) doesn't respond to syns
   2) sends empty syns

The existing _in_shadow_round mode is more intrusive, because it also prevents listeners from being called, so the state we collect is shallow.

asias:
> I don't understand the difference between the old shadow round and this new "long shadow" mode.
> @asias or @tgrabiec could you explain?
> @tgrabiec why can't we use the old shadow round? (maybe I'll understand if the difference is made clearer)
> 
> also could you argue why the decision of introducing a silent mode? as @asias said, it looks like this could also be solved by the node simply broadcasting a different STATUS (not `BOOT` or `NORMAL`, which cause other nodes to send data to this node).

I sent the PR here: https://github.com/scylladb/scylla/pull/7416 before this PR.

tgrabiec:
> I don't understand the difference between the old shadow round and this new "long shadow" mode.
> @asias or @tgrabiec could you explain?

The main difference is that the old shadow round doesn't run listeners, it only collects states inside the gossiper. So doesn't update token_metadata, for example, which is updated by listeners installed by the storage_service.

Also, the old shadow round is run before the gossiper is actually started, so communication is explicit.

> @tgrabiec why can't we use the old shadow round? (maybe I'll understand if the difference is made clearer)

We can't use the old shadow round because it doesn't run listeners.

> 
> also could you argue why the decision of introducing a silent mode? as @asias said, it looks like this could also be solved by the node simply broadcasting a different STATUS (not `BOOT` or `NORMAL`, which cause other nodes to send data to this node).

The main reason is that of simplicity and safety. With this patch, correctness depends on fewer things, the fact that you call `come_out_of_shadow()` at the proper time during boot (after initializing), and that the long shadow round is correct. The two are easy to prove because they're localized.

If you don't advertise yourself, other nodes won't call you. The fix doesn't care about what subsystems are initialized, and doesn't depend on who may try to contact the node and needs to be told not to. With the alternative approach which introduces a new global state, the correctness depends on different parts properly interpreting the new state. The proof now needs to be global, and the system becomes more complex due to increased implicit coupling. I am not sure that the alternative fix in the current shape is correct.

Also, introducing new node state complicates lifecycle management. What if we fail before actually moving to HIBERNATE? The replacing node is already in the ring, so needs to be dealt with. Can you replace a replacing node in such a state?

tgrabiec:
> This change has nothing to do with shadow round at all. It is confusing to call it shadow round to confuse with the old shadow round logics in the gossip message handers. It introduce a new gossip working mode, that is silent mode which listens only.

I think both modes belong to the category of a "shadow" round. Being in a shadow means that you can see others but others can't see you. All modes which have this property can be called shadow modes.

There are of course differences between the two modes, one calls the listeners and the other does not, for example. This can be captured by making the name more specific. I attempted this by adding "long" to the name.

> 
> It is much better than what you proposed earlier (making current verb based shadow round to call the gossip handlers and preparing a state before we can start gossip).

That's not what I was proposing, you must have misunderstood me.

> 
> However, it adds a new gossip working mode. Yes, this can work. But I think it is worse than introducing a new gossip status.
> 
> Without this patch, we can get rid of all the special existing shadow round "if" in gossip message handler after some grace period of being compatible with old version. Now, this patch adds more and we need to keep it forever. It is hard to reason about when we need to be in silent mode and when come out. It is easier to have only one normal gossip mode.

It's better to have one gossip mode than two. I think the benefits of this patch outweigh the cost of those ifs. Those ifs protect the rest of the system, and you avoid ifs in the layers which would need to be prepared for handling partially bootstrapped nodes.

We can get rid of all shadow rounds after we implement safe topology changes.

kbr-:
@tgrabiec I'm wondering... suppose the replacing node has the same IP address as the node being replaced. Suppose it's in the long shadow round.

One of the coordinator nodes attempts to make a read from the node being replaced, because it still thinks that it's UP (the race you discussed in the other PR). But the read is routed to the new node because it has the same IP.
Won't the read reach the replacing node?

What happens in that case?

tgrabiec:
> @tgrabiec I'm wondering... suppose the replacing node has the same IP address as the node being replaced. Suppose it's in the long shadow round.
> 
> One of the coordinator nodes attempts to make a read from the node being replaced, because it still thinks that it's UP (the race you discussed in the other PR). But the read is routed to the new node because it has the same IP.
> Won't the read reach the replacing node?
> 
> What happens in that case?

That's another, known, problem. Under the normal conditions, the coordinator should not send any requests to this node until it jumps to the HIBERNATE state because it will be still marked as DOWN. 

There can be races. This should be solved by having the bootstrapping node register replica-side RPC verbs only after it's initialized.

There is also a more general idea to get rid of the special casing needed to handle replacing nodes: #6403

service/storage_service.cc
```diff
@@ -590,6 +590,7 @@ void storage_service::join_token_ring(int delay) {
     replicate_to_all_cores().get();
     // start participating in the ring.
     set_gossip_tokens(_bootstrap_tokens, _cdc_streams_ts);
```
kbr-, line 592: Note that this can potentially perform a write:
```
            _cdc_streams_ts = cdc::make_new_cdc_generation(db().local().get_config(),
```
(line 574), e.g. if we're doing a rolling upgrade. So the node might need to coordinate a write to other nodes. Can it do it if other nodes didn't see this node in gossip (even with status UNKNOWN)?

btw. this should only happen either during rolling upgrade (and perhaps replace if the replaced node didn't have a cdc generation timestamp, but I don't know if that ever actually happens), so the other nodes should know something about this node from an earlier time, even though now it is in shadow. I.e. they should have some status of this node; in case of rolling upgrade it will probably be DN.
So the nodes would receive a write from a DN node. I wonder if that's ok.

Could you perform the following test with this change:
1. bootstrap two node cluster with cdc experimental flag disabled
2. stop one node, enable cdc flag, restart
The restarted node should try to create a CDC generation during this long shadow round and send it to both nodes.

kbr-: Correction: it will only try to create a CDC generation if the following check passes:
```
bool should_propose_first_generation(const gms::inet_address& me, const gms::gossiper& g) {
    auto my_host_id = g.get_host_id(me);
    auto& eps = g.get_endpoint_states();
    return std::none_of(eps.begin(), eps.end(),
            [&] (const std::pair<gms::inet_address, gms::endpoint_state>& ep) {
        return my_host_id < g.get_host_id(ep.first);
    });
}
```

so we can be sure that one of the two nodes will attempt this.

So the test should be as follows:
1. bootstrap two node cluster with cdc experimental flag disabled
2. stop one node, enable cdc flag, restart
3. stop second node, enable cdc flag, restart

the write will happen in either step 2 or 3 depending on how the host ids were generated

kbr-: also my question is whether this check is now at all correct... does the node have its own endpoint state while in shadow? This comment:
> Initially, this gossiper will stay silent about what he knows, including himself

if I understand correctly, it means that it won't even know about itself?

tgrabiec:
On Mon, Nov 9, 2020 at 3:30 PM Kamil Braun <notifications@github.com> wrote:

> *@kbr-* commented on this pull request.
>
> In service/storage_service.cc
> <https://github.com/scylladb/scylla/pull/7561#discussion_r519855686>:
>
> > @@ -590,6 +590,7 @@ void storage_service::join_token_ring(int delay) {
>      replicate_to_all_cores().get();
>      // start participating in the ring.
>      set_gossip_tokens(_bootstrap_tokens, _cdc_streams_ts);
>
> Note that this can potentially perform a write:
>
>             _cdc_streams_ts = cdc::make_new_cdc_generation(db().local().get_config(),
>
> (line 574), e.g. if we're doing a rolling upgrade. So the node might need
> to coordinate a write to other nodes. Can it do it if other nodes didn't
> see this node in gossip (even with status UNKNOWN)?
>
Coordinators don't have to be part of the ring, as far as I know.

We could come out earlier, if you think that would be a problem. But then,
is it safe to come out before _cdc_streams_ts are initialized?

tgrabiec:
On Mon, Nov 9, 2020 at 3:38 PM Kamil Braun <notifications@github.com> wrote:

> *@kbr-* commented on this pull request.
>
> In service/storage_service.cc
> <https://github.com/scylladb/scylla/pull/7561#discussion_r519860966>:
>
> > @@ -590,6 +590,7 @@ void storage_service::join_token_ring(int delay) {
>      replicate_to_all_cores().get();
>      // start participating in the ring.
>      set_gossip_tokens(_bootstrap_tokens, _cdc_streams_ts);
>
> also my question is whether this check is now at all correct... does the
> node have its own endpoint state while in shadow? This comment:
>
> Initially, this gossiper will stay silent about what he knows, including
> himself
>
> if I understand correctly, it means that it won't even know about itself?
>

The node which is in the shadow will know all the things about itself which
it knew before this change, it will just not tell anyone.

>

kbr-:
> Coordinators don't have to be part of the ring, as far as I know.

Yes. Question is whether the coordinator doesn't have to be considered "UP" for replicas to receive the mutation. Maybe not, I see no reason why we would want to enforce this. The test I proposed should immediately give us the answer.

> We could come out earlier, if you think that would be a problem. But then,
is it safe to come out before _cdc_streams_ts are initialized?

During normal bootstrapping procedure (when the node first joins), it's not safe to show ourselves as `NORMAL` before `_cdc_streams_ts` is initialized, because the moment other nodes see us as `NORMAL`, they will consider us a fully joined member of the cluster and will remember us as a peer, but formally the bootstrap can still fail since initializing `cdc_streams_ts` requires making a write, in which case the admin is allowed to abandon the bootstrap.

But we can show ourselves as `UNKNOWN` for the write (that's what happens today AFAIU).

For rolling upgrade we already were `NORMAL`, so the "damage" is already done... Failing the CDC generation write means that the upgrade fails and the admin has to keep retrying until the node successfully starts in order to finish the upgrade. So it shouldn't matter if we show ourselves earlier.

For replacing it's basically the same as for rolling upgrade (although replacing requires streaming data, so retrying a replace is costly). Also we could remove the cdc_streams initialization from this code path completely in the replacing case; it should have been initialized earlier (during `bootstrap()`), and if it wasn't, CDC probably doesn't work anyway and the fact of replacing a node won't change it. So yes, it also doesn't matter if we show ourselves earlier.

---

Another potential issue that popped into my mind: rolling upgrade may create new distributed tables (such as the ones in `system_distributed` keyspace). Again, I think it shouldn't matter whether or not the coordinating node is considered UP or not, it simply sends a mutation to other nodes. But what about schema digest exchange, which requires gossiping? Does the fact that we cannot exchange schema digests somehow conflict with sending a table creation request?

---

Also there's a simpler test that can be performed: simply bootstrap a node with CDC enabled from the start. This will attempt to create a CDC generation in the `bootstrap()` function, which will happen in the long shadow mode with this change. It will confirm whether or not a write can be performed during the shadow mode.

tgrabiec:
On Mon, Nov 9, 2020 at 4:57 PM Kamil Braun <notifications@github.com> wrote:

> Coordinators don't have to be part of the ring, as far as I know.
>
> Yes. Question is whether the coordinator doesn't have to be considered
> "UP" for replicas to receive the mutation. Maybe not, I see no reason why
> we would want to enforce this. The test I proposed should immediately give
> us the answer.
>
> We could come out earlier, if you think that would be a problem. But then,
> is it safe to come out before _cdc_streams_ts are initialized?
>
> During normal bootstrapping procedure (when the node first joins), it's
> not safe to show ourselves as NORMAL before _cdc_streams_ts is
> initialized, because the moment other nodes see us as NORMAL, they will
> consider us a fully joined member of the cluster and will remember us as a
> peer, but formally the bootstrap can still fail since initializing
> cdc_streams_ts requires making a write, in which case the admin is
> allowed to abandon the bootstrap.
>
But we can show ourselves as UNKNOWN for the write (that's what happens
> today AFAIU).
>
> For rolling upgrade we already were NORMAL, so the "damage" is already
> done... Failing the CDC generation write means that the upgrade fails and
> the admin has to keep retrying until the node successfully starts in order
> to finish the upgrade. So it shouldn't matter if we show ourselves earlier.
>
For replacing it's basically the same as for rolling upgrade (although
> replacing requires streaming data, so retrying a replace is costly). Also
> we could remove the cdc_streams initialization from this code path
> completely in the replacing case; it should have been initialized earlier
> (during bootstrap()), and if it wasn't, CDC probably doesn't work anyway
> and the fact of replacing a node won't change it. So yes, it also doesn't
> matter if we show ourselves earlier.
>
> Another potential issue that popped into my mind: rolling upgrade may
> create new distributed tables (such as the ones in system_distributed
> keyspace). Again, I think it shouldn't matter whether or not the
> coordinating node is considered UP or not, it simply sends a mutation to
> other nodes. But what about schema digest exchange, which requires
> gossiping? Does the fact that we cannot exchange schema digests somehow
> conflict with sending a table creation request?
>

We should come out before initiating schema changes, otherwise schema pull
won't work.

This was already the case for all bootstrapping (and replacing) nodes,
other nodes won't pull from them even if they see the mismatch.

This need not block this patch. For the restart case, I will move
come_out_of_shadow() to the place such that the behavior will not change
compared to what we have now.



>
> Also there's a simpler test that can be performed: simply bootstrap a node
> with CDC enabled from the start. This will attempt to create a CDC
> generation in the bootstrap() function, which will happen in the long
> shadow mode with this change. It will confirm whether or not a write can be
> performed during the shadow mode.
>

avikivity:
@kbr- what is the conclusion?

@asias please review

asias:
> @kbr- what is the conclusion?
> 
> @asias please review

I already did. It is bad idea to introduce a new gossip mode and bring back shadow round. It is much better to add a new state.

kbr-:
> @kbr- what is the conclusion?

I don't see any problems other than the fact that it needs to be tested on how it plays together with the cdc generation write made on node joining and on rolling upgrade.

Also I don't have a strong opinion on which way should we choose: adding a new gossip shadow mode, or adding a new state. The gossip mode has some nice theoretical properties that Tomek explained, but it's a new mechanism and may have unexpected consequences. Adding a new state is something that we know how to do and already did many times.
Maybe the path we choose doesn't matter that much in the long run: we will need to redesign the join procedure anyway when we start using Raft for topology changes.

gms/gossiper.cc
tgrabiec: Ok

avikivity:
Adding a state is easy, but don't we have to check that state on every request? Better to prepare in advance before joining.

asias:
> Adding a state is easy, but don't we have to check that state on every request? Better to prepare in advance before joining.

We have to check if there is a replacing node without this patch already. This is how replacing node takes write works. The check is a bool flag check, it is almost nothing comparing to other work we need to do in the request.

avikivity:
> > Adding a state is easy, but don't we have to check that state on every request? Better to prepare in advance before joining.
> 
> We have to check if there is a replacing node without this patch already. This is how replacing node takes write works. The check is a bool flag check, it is almost nothing comparing to other work we need to do in the request.

Yes, it is, but we keep piling on more and more easy stuff. It never ends.

We should try to move more work to prepare stages which are easier to make race-free.

asias:
> > > Adding a state is easy, but don't we have to check that state on every request? Better to prepare in advance before joining.
> > 
> > 
> > We have to check if there is a replacing node without this patch already. This is how replacing node takes write works. The check is a bool flag check, it is almost nothing comparing to other work we need to do in the request.
> 
> Yes, it is, but we keep piling on more and more easy stuff. It never ends.
> 
> We should try to move more work to prepare stages which are easier to make race-free.

The new state I proposed is the prepare stage. It makes the intention loud and clear. 

1) Prepare the replacing operation  2) Start the replacing operation.

mykaul:
@bhalevy - is this still relevant? What's the next step?

tgrabiec:
> I'm not sure this is still required after #12250 @tgrabiec, @asias: what do you think?

#12250 is not enough on its own, only full https://github.com/scylladb/scylladb/issues/6403 would make it obsolete. 

If I remember correctly, @asias implemented a different workaround for this problem, so I'm closing this.

PR #7564, opened 2020-11-06
dht: filter out empty ranges from selective_token_range_sharder

### LINKED ISSUE #5374: selective_token_range_sharder can generate range with empty tokens
On a node with msb = 12, shard count = 8,

1)
sharder = selective_token_range_sharder((9211550087832928255, 9212113037786349567], 2 )
sharder.next() will return (9211550087832928255, 9211550087832928256)

This range is empty because both side is exclusive. 

2)
sharder = selective_token_range_sharder((9211550087832928255, 9212113037786349567], 3 )
sharder.next() will return (9211550087832928255, 9212113037786349567]

This range is correct.

Example output of the test:

```
INFO  2019-11-28 16:15:58,871 [shard 0] repair - Repair 1 out of 1 ranges, id=12, shard=0, keyspace=myks3, table={standard1}, range=(9211550087832928255, 9212113037786349567], range_shard={}, skipped on this shard 
INFO  2019-11-28 16:15:58,871 [shard 1] repair - Repair 1 out of 1 ranges, id=12, shard=1, keyspace=myks3, table={standard1}, range=(9211550087832928255, 9212113037786349567], range_shard={}, skipped on this shard
INFO  2019-11-28 16:15:58,871 [shard 2] repair - Repair 1 out of 1 ranges, id=12, shard=2, keyspace=myks3, table={standard1}, range=(9211550087832928255, 9212113037786349567], range_shard={(9211550087832928255, 9211550087832928256)} 
INFO  2019-11-28 16:15:58,871 [shard 3] repair - Repair 1 out of 1 ranges, id=12, shard=3, keyspace=myks3, table={standard1}, range=(9211550087832928255, 9212113037786349567], range_shard={[9211550087832928256, 9212113037786349567]}                      
INFO  2019-11-28 16:15:58,871 [shard 4] repair - Repair 1 out of 1 ranges, id=12, shard=4, keyspace=myks3, table={standard1}, range=(9211550087832928255, 9212113037786349567], range_shard={}, skipped on this shard                                         
INFO  2019-11-28 16:15:58,871 [shard 5] repair - Repair 1 out of 1 ranges, id=12, shard=5, keyspace=myks3, table={standard1}, range=(9211550087832928255, 9212113037786349567], range_shard={}, skipped on this shard                                         
INFO  2019-11-28 16:15:58,871 [shard 6] repair - Repair 1 out of 1 ranges, id=12, shard=6, keyspace=myks3, table={standard1}, range=(9211550087832928255, 9212113037786349567], range_shard={}, skipped on this shard       
INFO  2019-11-28 16:15:58,871 [shard 7] repair - Repair 1 out of 1 ranges, id=12, shard=7, keyspace=myks3, table={standard1}, range=(9211550087832928255, 9212113037786349567], range_shard={}, skipped on this shard        
```

michoecho:
Sometimes selective_token_range_sharder generated ranges such as
(x, x+1). They contain no tokens and generating them was a waste
of resources (for example, repair would start an instance on the
empty range).
This patch filters out ranges of the form (x, x+1).

Fixes #5374

### DIFF
--- a/dht/i_partitioner.cc
+++ b/dht/i_partitioner.cc
@@ -179,6 +179,23 @@ unsigned shard_of(const schema& s, const token& t) {
     return s.get_sharder().shard_of(t);
 }
 
+static bool token_range_is_not_empty(const dht::token_range& r) {
+    bool is_bounded = r.start() && !r.start()->value().is_minimum() && r.end() && !r.end()->value().is_maximum();
+    if (is_bounded) {
+        // token_range is a nonwrapping_interval, so (end() >= start()), so
+        // (r.end() && !r.end()->value().is_maximum()) implies !r.start()->value().is_maximum()
+        // (r.start() && !r.start()->value().is_minimum()) implies !r.end()->value().is_minimum()
+        uint64_t start_raw = r.start()->value().raw();
+        uint64_t end_raw = r.end()->value().raw();
+        return end_raw - start_raw >= (!r.start()->is_inclusive() + !r.end()->is_inclusive());
+    } else {
+        const static dht::token_range full_range = dht::token_range::make(
+                range_bound<dht::token>(token::from_int64(std::numeric_limits<int64_t>::min() + 1), true),
+                range_bound<dht::token>(token::from_int64(std::numeric_limits<int64_t>::max()), true));
+        return r.overlaps(full_range, dht::token_comparator());
+    }
+}
+
 std::optional<dht::token_range>
 selective_token_range_sharder::next() {
     if (_done) {
@@ -191,7 +208,7 @@ selective_token_range_sharder::next() {
         auto intersection = _range.intersection(std::move(candidate), dht::token_comparator());
         _start_token = _sharder.token_for_next_shard(end_token, _shard);
         _start_boundary = range_bound<dht::token>(_start_token);
-        if (intersection) {
+        if (intersection && token_range_is_not_empty(*intersection)) {
             return *intersection;
         }
     }
--- a/dht/sharder.hh
+++ b/dht/sharder.hh
@@ -129,6 +129,7 @@ public:
 
 // Incrementally divides a `partition_range` into sub-ranges wholly owned by a single shard.
 // Unlike ring_position_range_sharder, it only returns result for a shard number provided by the caller.
+// Skips ranges which contain no tokens.
 class selective_token_range_sharder {
     const sharder& _sharder;
     dht::token_range _range;
--- a/test/boost/partitioner_test.cc
+++ b/test/boost/partitioner_test.cc
@@ -506,6 +506,43 @@ SEASTAR_THREAD_TEST_CASE(test_selective_token_range_sharder) {
     return test_something_with_some_interesting_ranges_and_sharder_with_token_range(do_test_selective_token_range_sharder);
 }
 
+SEASTAR_THREAD_TEST_CASE(test_selective_token_range_sharder_empty_ranges) {
+    auto input_sharder = dht::sharder(2, 0);
+    auto make_bound = [] (dht::token t, bool inclusive) {
+        return range_bound<dht::token>(std::move(t), inclusive);
+    };
+
+    dht::token min_regular_token = dht::token::from_int64(std::numeric_limits<int64_t>::min() + 1);
+    dht::token max_regular_token = dht::token::from_int64(std::numeric_limits<int64_t>::max());
+
+    // (shard, range) pairs
+    std::pair<int, dht::token_range> empty_cases[] = {
+            {0, dht::token_range::make_ending_with(make_bound(dht::minimum_token(), true))},
+            {0, dht::token_range::make_ending_with(make_bound(min_regular_token, false))},
+            {1, dht::token_range::make_starting_with(make_bound(dht::maximum_token(), true))},
+            {1, dht::token_range::make_starting_with(make_bound(max_regular_token, false))},
+            {1, dht::token_range::make_ending_with(make_bound(token_from_long(0), false))},
+            {0, dht::token_range::make_starting_with(make_bound(token_from_long(-1), false))},
+    };
+    std::pair<int, dht::token_range> nonempty_cases[] = {
+            {0, dht::token_range::make_ending_with(make_bound(min_regular_token, true))},
+            {1, dht::token_range::make_starting_with(make_bound(max_regular_token, true))},
+            {0, dht::token_range::make_open_ended_both_sides()},
+            {1, dht::token_range::make_ending_with(make_bound(token_from_long(0), true))},
+            {0, dht::token_range::make_starting_with(make_bound(token_from_long(-1), true))},
+            {1, dht::token_range::make(make_bound(token_from_long(-1), false), make_bound(token_from_long(1), false))},
+    };
+
+    for (const auto& [shard, range] : empty_cases) {
+        auto result = dht::selective_token_range_sharder(input_sharder, range, shard).next();
+        BOOST_REQUIRE(!result);
+    }
+    for (const auto& [shard, range] : nonempty_cases) {
+        auto result = dht::selective_token_range_sharder(input_sharder, range, shard).next();
+        BOOST_REQUIRE(result);
+    }
+}
+
 SEASTAR_THREAD_TEST_CASE(test_find_first_token_for_shard) {
     const unsigned cpu_count = 3;
     const unsigned ignore_msb_bits = 10;


michoecho:
I didn't understand the semantics of token_kind::after_all_keys, token_kind::before_all_keys, and one-sided ranges in the context of this task. I have already spent way too long on this issue, so please tell me what should be done with those if you know.

haaawk:
I think what you do now is ok. We can assume a range starting with before_all_keys or/and ending with after_all_keys is not empty. Examples you gave should never appear in practice and if they do we will pay a little price of a single empty range that's not filtered out. Just please add a comment that we're ignoring them on purpose.

haaawk:
Didn't you need to change any of the existing tests? If there are no tests that cover the issue you're fixing, please add some.

dht/i_partitioner.cc
```diff
@@ -179,6 +179,16 @@ unsigned shard_of(const schema& s, const token& t) {
...
 }
 
+static bool token_range_is_empty(const dht::token_range& r) {
+    return r.start()
+        && r.end()
+        && !r.start()->is_inclusive()
+        && !r.end()->is_inclusive()
+        && r.start()->value()._kind == dht::token_kind::key
+        && r.end()->value()._kind == dht::token_kind::key
+        && static_cast<uint64_t>(r.end()->value()._data) - static_cast<uint64_t>(r.start()->value()._data) == 1;
```
haaawk, line 189: I think it would be better to use https://github.com/scylladb/scylla/blob/1b46a0a5f114f50a295be388ceecd808283e87d5/dht/token.hh#L138

dht/i_partitioner.cc
```diff
@@ -179,6 +179,16 @@ unsigned shard_of(const schema& s, const token& t) {
...
+}
+
```
haaawk, line 191: Maybe it would be better to use https://github.com/scylladb/scylla/blob/1b46a0a5f114f50a295be388ceecd808283e87d5/dht/token.hh#L163

and just do:
```
return r.start()
    && r.end()
    && !r.start()->is_inclusive()
    && !r.end()->is_inclusive()
    && r.end().raw() - r.start().raw() < 2;
```

haaawk: The question is whether r can be a wrap-around like (10, 5).

haaawk:
> I think what you do now is ok. We can assume a range starting with before_all_keys or/and ending with after_all_keys is not empty. Examples you gave should never appear in practice and if they do we will pay a little price of a single empty range that's not filtered out. Just please add a comment that we're ignoring them on purpose.

I take it back. Using `token::raw()` should solve all the corner cases.

dht/i_partitioner.cc
michoecho: `dht::token_range` is `nonwrapping_interval<token>`, so it can't be a wrap-around.
But the problem with raw() is that it collates `after_all_keys` and int64max, which are different things. In particular, (int64max-1, after_all_keys) is definitely not empty, but it fulfils your condition.

haaawk: In practice, it is empty, no? what token can be placed in (int64max -1 , int64max) ?

haaawk: ok I see your point. int64max should be inside (int64max-1, after_all_keys).

haaawk: Then the logic should be:
1. if start is before all keys and r.end().raw() - r.start().raw() == 0 then empty
2. if end is after all keys and r.end().raw() - r.start().raw() == 0 then empty
3. Otherwise if neither start nor end are special r.end().raw() - r.start().raw() < 2
How about something like this?

haaawk: so
```
return r.start()
    && r.end()
    && !r.start()->is_inclusive()
    && !r.end()->is_inclusive()
    && r.end().raw() - r.start().raw() < ((r.start().is_minimum() || r.end().is_maximum()) ? 0 : 2);
```

michoecho:
v2 changes:
- Excluded subranges of (-inf, min_int + 1) and (max_int, +inf) as empty.
- Added a unit test.

avikivity:
> I can't figure out the semantics of token_kind::after_all_keys, token_kind::before_all_keys, and one-sided ranges in the context of #5374.
> 
> For example, are [before_all_keys, before_all_keys], (int64max, after_all_keys], (-∞, before_all_keys] empty?

[before_all_keys, after_all_keys] == infinite range (the opposite of empty).

Note that before_all_keys/after_all_keys are redundant and can be removed since they are always used in an interval, and so can be replaced with an open bound. However, that is quite difficult since before_all/after_all are passed in idl. Perhaps we can start by never generating them, but still accepting them.

dht/i_partitioner.cc
```diff
@@ -191,7 +205,7 @@ selective_token_range_sharder::next() {
         auto intersection = _range.intersection(std::move(candidate), dht::token_comparator());
```
avikivity, line 205: We might be able to make it work on a lower level (and so on more cases) by changing the token compare functions, so that intersection() returns a disengaged optional.

Although, interval will not be able to distinguish between an open interval and min/max tokens. So it's probably not workable.

dht/i_partitioner.cc
```diff
@@ -179,6 +179,20 @@ unsigned shard_of(const schema& s, const token& t) {
     return s.get_sharder().shard_of(t);
 }
 
+static bool token_range_is_not_empty(const dht::token_range& r) {
+    bool is_bounded = r.start() && !r.start()->value().is_minimum() && r.end() && !r.end()->value().is_maximum();
```
avikivity, line 183: What if we somehow get is_maximum() in start() or is_minimum() in end()?

dht/i_partitioner.cc
```diff
@@ -179,6 +179,20 @@ unsigned shard_of(const schema& s, const token& t) {
...
+    if (is_bounded) {
+        uint64_t start_raw = r.start()->value().raw();
+        uint64_t end_raw = r.end()->value().raw();
+        return end_raw - start_raw >= (!r.start()->is_inclusive() + !r.end()->is_inclusive());
```
avikivity, line 187: You were careful to avoid overflow in two separate ways. Let's celebrate it with a comment, this isn't trivial.

dht/i_partitioner.cc
michoecho: This is a nonwrapping_interval, so if start() is maximum, then end() must be either maximum() or empty/open.
Ditto for the second case.

michoecho: In other words, `!r.start()->value().is_maximum()` is implied by `r.end() && !r.end()->value().is_maximum()`

avikivity:
Looks good, left minor comments.

dht/i_partitioner.cc
michoecho: I don't think it can be done while keeping the interval classes generic.

dht/i_partitioner.cc
```diff
@@ -179,6 +179,20 @@ unsigned shard_of(const schema& s, const token& t) {
...
+    } else {
+        const static dht::token_range full_range = dht::token_range::make(
+                range_bound<dht::token>(token::from_int64(std::numeric_limits<int64_t>::min() + 1), true),
+                range_bound<dht::token>(token::from_int64(std::numeric_limits<int64_t>::max()), true));
```
haaawk, line 191: Maybe use maximum_token() instead of token::from_int64?

dht/i_partitioner.cc
```diff
@@ -179,6 +179,20 @@ unsigned shard_of(const schema& s, const token& t) {
...
 
+static bool token_range_is_not_empty(const dht::token_range& r) {
+    bool is_bounded = r.start() && !r.start()->value().is_minimum() && r.end() && !r.end()->value().is_maximum();
+    if (is_bounded) {
+        uint64_t start_raw = r.start()->value().raw();
+        uint64_t end_raw = r.end()->value().raw();
+        return end_raw - start_raw >= (!r.start()->is_inclusive() + !r.end()->is_inclusive());
+    } else {
+        const static dht::token_range full_range = dht::token_range::make(
+                range_bound<dht::token>(token::from_int64(std::numeric_limits<int64_t>::min() + 1), true),
```
haaawk, line 190: Maybe use range_bound<dht::token>(minimum_token(), false)

haaawk (approved):
LGTM left two suggestions

dht/i_partitioner.cc
avikivity: Agree, if we remove before_all/after_all _and_ specialize `interval` for integral ranges, _and_ communicate to `interval` that `token` is integral, we might pull it off. Quite a large task.

dht/i_partitioner.cc
avikivity: Right. Please add a comment.

dht/i_partitioner.cc
michoecho: That will give different results. We want (int64max, +inf) to be empty. Using maximum_token() (which is after_all_keys) will give us a non-empty intersection (int64max, after_all_keys) here.

dht/i_partitioner.cc
michoecho: We want (-inf, int64min) to be empty, but your bound would result in a (-inf, before_all_keys) intersection.

dht/i_partitioner.cc
michoecho: Done.

dht/i_partitioner.cc
haaawk: Ok.

dht/i_partitioner.cc
haaawk: Ok.

dht/i_partitioner.cc
haaawk: Let's leave it as a future improvement, shall we?

michoecho:
Closing due to inactivity and doubts about the issue itself. There is some work happening with simplifying the representation of intervals, perhaps this can be revisited when that work is done.

PR #7565, opened 2020-11-06
Add TransitionalAuthorizer and TransitionalAuthenticator to config descriptions

StarostaGit:
We allow setting the transitional auth as one of the options in scylla.yaml, but don't mention it at all in the field's description. Let's change that.


haaawk (approved):
LGTM. @psarna could you have a look please?

PR #7567, opened 2020-11-07
scylla_util.py: fix exception handling in curl

bentsi:
Retry mechanism didn't work when URLError happend.
Example of exception: urllib.error.URLError: <urlopen error [Errno 101] Network is unreachable>
Catching URLError instead of HTTP since URLError is a base excption for all exceptions in
the urllib module


avikivity:
Please add a Fixes tag (and file an issue) so it can be backported.

bentsi:
@avikivity Done.

PR #7576, opened 2020-11-09
install.sh: set PATH for relocatable CLI tools in python thunk

### LINKED ISSUE #7350: relocatable scripts lack proper paths

*Installation details*
Scylla version (or git commit hash): 4.1
Cluster size:

Disks: (SSD/HDD, count)

@tarzanek
this command isn't working, I think it's not using the corrects PATH and the relocatable python, from within the scylla_prepare I think it does work
```
/opt/scylladb/scripts/perftune.py --tune net --nic eth0 --get-cpu-mask
```

- above binary is expecting to find hwloc in relocatable paths, but it's not finding it
within scylla-server service is most probably able to find it, but I think this command should work also as standalone

I am opening this bug as a placeholder to open a discussion (or get a fix) to get the tools from scylla work again even with relocatable binaries under different user than scylla (e.g. root)

syuu1228:
We currently set PATH for relocatable CLI tools in scylla_util.run() and
scylla_util.out(), but it doesn't work for perftune.py, since it's not part of
Scylla, does not use scylla_util module.
We can set PATH in python thunk instead, it can set PATH for all python scripts.

Fixes #7350

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -144,7 +144,7 @@ DEBIAN_SSL_CERT_FILE="/etc/ssl/certs/ca-certificates.crt"
 if [ -f "\${DEBIAN_SSL_CERT_FILE}" ]; then
   c=\${DEBIAN_SSL_CERT_FILE}
 fi
-PYTHONPATH="\${d}:\${d}/libexec:\$PYTHONPATH" PATH="\${d}/$pythonpath:\${PATH}" SSL_CERT_FILE="\${c}" exec -a "\$0" "\${d}/libexec/\${b}" "\$@"
+PYTHONPATH="\${d}:\${d}/libexec:\$PYTHONPATH" PATH="\${d}/../bin:\${d}/$pythonpath:\${PATH}" SSL_CERT_FILE="\${c}" exec -a "\$0" "\${d}/libexec/\${b}" "\$@"
 EOF
     chmod +x "$install"
 }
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -133,6 +133,7 @@ relocate_python3() {
 #!/usr/bin/env bash
 [[ -z "\$LD_PRELOAD" ]] || { echo "\$0: not compatible with LD_PRELOAD" >&2; exit 110; }
 export LC_ALL=en_US.UTF-8
+export DEBIAN_FRONTEND=noninteractive
```
avikivity, line 136: Why move this? it is unrelated.

install.sh
```diff
@@ -144,7 +145,7 @@ DEBIAN_SSL_CERT_FILE="/etc/ssl/certs/ca-certificates.crt"
 if [ -f "\${DEBIAN_SSL_CERT_FILE}" ]; then
   c=\${DEBIAN_SSL_CERT_FILE}
 fi
-PYTHONPATH="\${d}:\${d}/libexec:\$PYTHONPATH" PATH="\${d}/$pythonpath:\${PATH}" SSL_CERT_FILE="\${c}" exec -a "\$0" "\${d}/libexec/\${b}" "\$@"
+PYTHONPATH="\${d}:\${d}/libexec:\$PYTHONPATH" PATH="\${d}/../bin:\${d}/$pythonpath:\${PATH}" SSL_CERT_FILE="\${c}" exec -a "\$0" "\${d}/libexec/\${b}" "\$@"
```
avikivity, line 148: I see install.sh relocates perftune.py,, so this should work.

install.sh
syuu1228: Dropped on new version.

PR #7582, opened 2020-11-10
secondary_index: Fix TOKEN() handling when combined with index-supported restriction

### LINKED ISSUE #7043: scylla vs cassandra compatibility issue (spark connector filtering on secondary index)

avelanarius:
Fixes `TOKEN(...)` restrictions handling in `SELECTs` when combined with global secondary index supported restriction, for example:
```
CREATE TABLE t(pk int, ck int, v int, PRIMARY KEY(pk, ck));
CREATE INDEX ON t(v);
SELECT * FROM t WHERE v = 5 AND token(pk) < 0 ALLOW FILTERING;
```

This PR does not change when `ALLOW FILTERING` clause is required, because there are multiple opinions on when it should be required. 

This functionality is only supported on new indexes (those with `token_v2` column computation). In case of running on a old index, the old behaviour is not changed (returning the same invalid data).

Adds unit tests to validate various `SELECT`s with `TOKEN` restrictions.

Adds new `global_index_prepare_data` structure which stores data related to global index that was computed on `prepare` stage.

Fixes #7043.

### DIFF
--- a/cql3/expr/expression.cc
+++ b/cql3/expr/expression.cc
@@ -863,6 +863,26 @@ expression replace_column_def(const expression& expr, const column_definition* n
         }, expr);
 }
 
+expression replace_token(const expression& expr, const column_definition* new_cdef) {
+    return std::visit(overloaded_functor{
+            [] (bool b){ return expression(b); },
+            [&] (const conjunction& conj) {
+                const auto applied = conj.children | transformed(
+                        std::bind(replace_token, std::placeholders::_1, new_cdef));
+                return expression(conjunction{std::vector(applied.begin(), applied.end())});
+            },
+            [&] (const binary_operator& oper) {
+                return std::visit(overloaded_functor{
+                        [&] (const column_value& col) { return expr; },
+                        [&] (const std::vector<column_value>& cvs) { return expr; },
+                        [&] (const token&) { 
+                            return expression(binary_operator{column_value{new_cdef}, oper.op, oper.rhs});
+                        },
+                    }, oper.lhs);
+            },
+        }, expr);
+}
+
 std::ostream& operator<<(std::ostream& s, oper_t op) {
     switch (op) {
     case oper_t::EQ:
--- a/cql3/restrictions/statement_restrictions.hh
+++ b/cql3/restrictions/statement_restrictions.hh
@@ -414,7 +414,23 @@ public:
     }
 
     bool ck_restrictions_need_filtering() const {
-        return _partition_key_restrictions->has_unrestricted_components(*_schema) || _clustering_columns_restrictions->needs_filtering(*_schema);
+        return !_clustering_columns_restrictions->empty() && 
+            (_partition_key_restrictions->has_unrestricted_components(*_schema) 
+            //
+            // An index table's clustering key is (token(pk), pk, ck) of the base table. 
+            // If pk is specified, we can calculate token; but if token is specified, we CANNOT calculate pk. 
+            // (Currently token and pk restrictions cannot mix together.) 
+            //
+            // Therefore, if both token and ck are restricted, the clustering slice of the index 
+            // table is not continuous (see below), and we may need to filter out some of the rows fetched.
+            //
+            // Example: for restriction on token(pk1, pk2) and ck1
+            // (token(pk1, pk2), pk1, pk2, ck1, ck2)
+            //  ^ restricted               ^ restricted
+            // 
+            //  In this example clustering slice is not continuous (missing restrictions on pk).
+            || (_uses_secondary_indexing && has_token(_partition_key_restrictions->expression)) 
+            || _clustering_columns_restrictions->needs_filtering(*_schema));
     }
 
     /**
--- a/cql3/statements/select_statement.cc
+++ b/cql3/statements/select_statement.cc
@@ -46,7 +46,6 @@
 #include "cql3/functions/as_json_function.hh"
 #include "cql3/selection/selection.hh"
 #include "cql3/util.hh"
-#include "cql3/restrictions/single_column_primary_key_restrictions.hh"
 #include "cql3/selection/selector_factories.hh"
 #include <seastar/core/shared_ptr.hh>
 #include "query-result-reader.hh"
@@ -837,8 +836,24 @@ indexed_table_select_statement::prepare(database& db,
         // Fill in the global_index_prepare_data structure
         const column_definition& token_cdef = *view_schema->clustering_key_columns().begin();
 
+        const bool is_token_v2_index = token_cdef.type == long_type;
+        const auto& partition_key_restriction_expression = restrictions->get_partition_key_restrictions()->expression;
+        // Global posting list partition slice according to the token restriction is only supported
+        // with new token_column_computation type (computation name: token_v2)
+        bool is_token_partition_slice = has_token(partition_key_restriction_expression) && is_token_v2_index;
+
+        ::shared_ptr<restrictions::single_column_clustering_key_restrictions> posting_list_clustering_restrictions;
+        if (is_token_partition_slice) {
+            posting_list_clustering_restrictions = ::make_shared<restrictions::single_column_clustering_key_restrictions>(view_schema, false);
+
+            auto token_restriction = ::make_shared<restrictions::single_column_restriction>(token_cdef);
+            token_restriction->expression = expr::replace_token(partition_key_restriction_expression, &token_cdef);
+
+            posting_list_clustering_restrictions->merge_with(token_restriction);
+        }
+
         global_index_data.emplace(global_index_prepare_data{
-            token_cdef
+            token_cdef, is_token_partition_slice, std::move(posting_list_clustering_restrictions)
         });
     }
 
@@ -1132,8 +1147,8 @@ query::partition_slice indexed_table_select_statement::get_partition_slice_for_g
 
     if (!_restrictions->has_partition_key_unrestricted_components()) {
         auto single_pk_restrictions = dynamic_pointer_cast<restrictions::single_column_partition_key_restrictions>(_restrictions->get_partition_key_restrictions());
-        // Only EQ restrictions on base partition key can be used in an index view query
         if (single_pk_restrictions && single_pk_restrictions->is_all_eq()) {
+            // EQ restrictions on base partition key
             auto clustering_restrictions = ::make_shared<restrictions::single_column_clustering_key_restrictions>(_view_schema, *single_pk_restrictions);
             auto base_pk = partition_key::from_optional_exploded(*_schema, single_pk_restrictions->values(options));
             bytes token_value = compute_idx_token(base_pk);
@@ -1155,6 +1170,10 @@ query::partition_slice indexed_table_select_statement::get_partition_slice_for_g
             }
 
             partition_slice_builder.with_ranges(clustering_restrictions->bounds_ranges(options));
+        } else if (_global_index_prepare_data->_is_token_partition_slice) {
+            // TOKEN restriction on base partition key
+            auto bounds = _global_index_prepare_data->_posting_list_clustering_restrictions->bounds_ranges(options);
+            partition_slice_builder.with_ranges(std::move(bounds));
         }
     }
 
--- a/cql3/statements/select_statement.cc
+++ b/cql3/statements/select_statement.cc
@@ -831,6 +831,16 @@ indexed_table_select_statement::prepare(database& db,
     const auto& im = index_opt->metadata();
     sstring index_table_name = im.name() + "_index";
     schema_ptr view_schema = db.find_schema(schema->ks_name(), index_table_name);
+    std::optional<global_index_prepare_data> global_index_data;
+
+    if (!im.local()) {
+        // Fill in the global_index_prepare_data structure
+        const column_definition& token_cdef = *view_schema->clustering_key_columns().begin();
+
+        global_index_data.emplace(global_index_prepare_data{
+            token_cdef
+        });
+    }
 
     return ::make_shared<cql3::statements::indexed_table_select_statement>(
             schema,
@@ -846,7 +856,8 @@ indexed_table_select_statement::prepare(database& db,
             stats,
             *index_opt,
             std::move(used_index_restrictions),
-            view_schema);
+            view_schema,
+            std::move(global_index_data));
 
 }
 
@@ -862,11 +873,13 @@ indexed_table_select_statement::indexed_table_select_statement(schema_ptr schema
                                                            cql_stats &stats,
                                                            const secondary_index::index& index,
                                                            ::shared_ptr<restrictions::restrictions> used_index_restrictions,
-                                                           schema_ptr view_schema)
+                                                           schema_ptr view_schema,
+                                                           std::optional<global_index_prepare_data> global_index_prepare_data)
     : select_statement{schema, bound_terms, parameters, selection, restrictions, group_by_cell_indices, is_reversed, ordering_comparator, limit, per_partition_limit, stats}
     , _index{index}
     , _used_index_restrictions(used_index_restrictions)
     , _view_schema(view_schema)
+    , _global_index_prepare_data(std::move(global_index_prepare_data))
 {
     if (_index.metadata().local()) {
         _get_partition_ranges_for_posting_list = [this] (const query_options& options) { return get_partition_ranges_for_local_index_posting_list(options); };
@@ -1122,13 +1135,11 @@ query::partition_slice indexed_table_select_statement::get_partition_slice_for_g
         // Only EQ restrictions on base partition key can be used in an index view query
         if (single_pk_restrictions && single_pk_restrictions->is_all_eq()) {
             auto clustering_restrictions = ::make_shared<restrictions::single_column_clustering_key_restrictions>(_view_schema, *single_pk_restrictions);
-            // Computed token column needs to be added to index view restrictions
-            const column_definition& token_cdef = *_view_schema->clustering_key_columns().begin();
             auto base_pk = partition_key::from_optional_exploded(*_schema, single_pk_restrictions->values(options));
             bytes token_value = compute_idx_token(base_pk);
-            auto token_restriction = ::make_shared<restrictions::single_column_restriction>(token_cdef);
+            auto token_restriction = ::make_shared<restrictions::single_column_restriction>(_global_index_prepare_data->_token_cdef);
             token_restriction->expression = expr::binary_operator{
-                    &token_cdef, expr::oper_t::EQ,
+                    &_global_index_prepare_data->_token_cdef, expr::oper_t::EQ,
                     ::make_shared<cql3::constants::value>(cql3::raw_value::make_value(token_value))};
             clustering_restrictions->merge_with(token_restriction);
 
--- a/cql3/statements/select_statement.hh
+++ b/cql3/statements/select_statement.hh
@@ -175,9 +175,19 @@ public:
 };
 
 class indexed_table_select_statement : public select_statement {
+public:
+    // Struct which stores data related to global index,
+    // computed on the prepare stage
+    struct global_index_prepare_data {
+        // Column definition of computed token column in global index view
+        const column_definition& _token_cdef;
+    };
+
+private:
     secondary_index::index _index;
     ::shared_ptr<restrictions::restrictions> _used_index_restrictions;
     schema_ptr _view_schema;
+    std::optional<global_index_prepare_data> _global_index_prepare_data;
     noncopyable_function<dht::partition_range_vector(const query_options&)> _get_partition_ranges_for_posting_list;
     noncopyable_function<query::partition_slice(const query_options&)> _get_partition_slice_for_posting_list;
 public:
@@ -207,7 +217,8 @@ public:
                                    cql_stats &stats,
                                    const secondary_index::index& index,
                                    ::shared_ptr<restrictions::restrictions> used_index_restrictions,
-                                   schema_ptr view_schema);
+                                   schema_ptr view_schema,
+                                   std::optional<global_index_prepare_data> global_index_prepare_data);
 
 private:
     virtual future<::shared_ptr<cql_transport::messages::result_message>> do_execute(service::storage_proxy& proxy,
--- a/cql3/statements/select_statement.hh
+++ b/cql3/statements/select_statement.hh
@@ -54,6 +54,7 @@
 #include <seastar/core/distributed.hh>
 #include "validation.hh"
 #include "transport/messages/result_message.hh"
+#include "cql3/restrictions/single_column_primary_key_restrictions.hh"
 
 namespace cql3 {
 
@@ -181,6 +182,12 @@ public:
     struct global_index_prepare_data {
         // Column definition of computed token column in global index view
         const column_definition& _token_cdef;
+
+        // Will global posting list partition be sliced according to base query TOKEN restriction
+        bool _is_token_partition_slice;
+
+        // Clustering restrictions of global posting list precomputed at prepare stage
+        shared_ptr<restrictions::single_column_clustering_key_restrictions> _posting_list_clustering_restrictions;
     };
 
 private:
--- a/test/boost/secondary_index_test.cc
+++ b/test/boost/secondary_index_test.cc
@@ -1560,6 +1560,228 @@ SEASTAR_TEST_CASE(test_token_order) {
     });
 }
 
+SEASTAR_TEST_CASE(test_select_with_token_range_cases) {
+    return do_with_cql_env_thread([] (auto& e) {
+        cquery_nofail(e, "CREATE TABLE t (pk int, v int, PRIMARY KEY(pk))");
+
+        for (int i = 0; i < 7; i++) {
+            // v=1 in each row, so WHERE v = 1 will select them all
+            cquery_nofail(e, format("INSERT INTO t (pk, v) VALUES ({}, 1)", i).c_str());
+        }
+
+        auto get_result_rows = [&](int start, int end) {
+            std::vector<std::vector<bytes_opt>> result;
+            for (int i = start; i <= end; i++) {
+                result.push_back({ testset_pks[i][0], int32_type->decompose(1) });
+            }
+            return result;
+        };
+
+        auto get_query = [&](sstring token_restriction) {
+            return format("SELECT * FROM t WHERE v = 1 AND {} ALLOW FILTERING", token_restriction);
+        };
+
+        auto q = [&](sstring token_restriction) {
+            return cquery_nofail(e, get_query(token_restriction).c_str());
+        };
+
+        auto inclusive_inclusive_range = [](int64_t start, int64_t end) { return format("token(pk) >= {} AND token(pk) <= {}", start, end); };
+        auto exclusive_inclusive_range = [](int64_t start, int64_t end) { return format("token(pk) > {} AND token(pk) <= {}", start, end); };
+        auto inclusive_exclusive_range = [](int64_t start, int64_t end) { return format("token(pk) >= {} AND token(pk) < {}", start, end); };
+        auto exclusive_exclusive_range = [](int64_t start, int64_t end) { return format("token(pk) > {} AND token(pk) < {}", start, end); };
+
+        auto inclusive_infinity_range = [](int64_t start) { return format("token(pk) >= {}", start); };
+        auto exclusive_infinity_range = [](int64_t start) { return format("token(pk) > {}", start); };
+
+        auto infinity_inclusive_range = [](int64_t end) { return format("token(pk) <= {}", end); };
+        auto infinity_exclusive_range = [](int64_t end) { return format("token(pk) < {}", end); };
+
+        auto equal_range = [](int64_t value) { return format("token(pk) = {}", value); };
+
+        auto do_tests = [&] {
+            assert_that(q(inclusive_inclusive_range(testset_tokens[1], testset_tokens[5]))).is_rows().with_rows(get_result_rows(1, 5));
+            assert_that(q(inclusive_inclusive_range(testset_tokens[1], testset_tokens[5] - 1))).is_rows().with_rows(get_result_rows(1, 4));
+            assert_that(q(inclusive_inclusive_range(testset_tokens[1], testset_tokens[5] + 1))).is_rows().with_rows(get_result_rows(1, 5));
+            assert_that(q(inclusive_inclusive_range(testset_tokens[1] + 1, testset_tokens[5]))).is_rows().with_rows(get_result_rows(2, 5));
+
+            assert_that(q(exclusive_inclusive_range(testset_tokens[1], testset_tokens[4]))).is_rows().with_rows(get_result_rows(2, 4));
+
+            assert_that(q(inclusive_exclusive_range(testset_tokens[1], testset_tokens[4]))).is_rows().with_rows(get_result_rows(1, 3));
+
+            assert_that(q(exclusive_exclusive_range(testset_tokens[1], testset_tokens[4]))).is_rows().with_rows(get_result_rows(2, 3));
+
+            assert_that(q(inclusive_infinity_range(testset_tokens[3]))).is_rows().with_rows(get_result_rows(3, testset_pks.size() - 1));
+
+            assert_that(q(exclusive_infinity_range(testset_tokens[3]))).is_rows().with_rows(get_result_rows(4, testset_pks.size() - 1));
+
+            assert_that(q(infinity_inclusive_range(testset_tokens[3]))).is_rows().with_rows(get_result_rows(0, 3));
+
+            assert_that(q(infinity_exclusive_range(testset_tokens[3]))).is_rows().with_rows(get_result_rows(0, 2));
+            
+            assert_that(q("token(pk) < 0")).is_rows().with_rows(get_result_rows(0, 4));
+
+            assert_that(q("token(pk) > 0")).is_rows().with_rows(get_result_rows(5, testset_pks.size() - 1));
+
+            assert_that(q(equal_range(testset_tokens[3]))).is_rows().with_rows(get_result_rows(3, 3));
+
+            // empty range
+            assert_that(q("token(pk) < 5 AND token(pk) > 100")).is_rows().with_size(0);
+
+            // prepared statement
+            auto prepared_id = e.prepare(get_query("token(pk) >= ? AND token(pk) <= ?")).get0();
+            auto msg = e.execute_prepared(prepared_id, {
+                cql3::raw_value::make_value(long_type->decompose(testset_tokens[1])), cql3::raw_value::make_value(long_type->decompose(testset_tokens[5]))
+            }).get0();
+            assert_that(msg).is_rows().with_rows(get_result_rows(1, 5));
+
+            msg = e.execute_prepared(prepared_id, {
+                cql3::raw_value::make_value(long_type->decompose(testset_tokens[2])), cql3::raw_value::make_value(long_type->decompose(testset_tokens[6]))
+            }).get0();
+            assert_that(msg).is_rows().with_rows(get_result_rows(2, 6));
+        };
+
+        eventually(do_tests);
+
+        cquery_nofail(e, "CREATE INDEX t_global ON t(v)");
+
+        eventually(do_tests);
+
+        cquery_nofail(e, "DROP INDEX t_global");
+        cquery_nofail(e, "CREATE INDEX t_local ON t((pk), v)");
+
+        eventually(do_tests);
+    });
+}
+
+struct testset_row {
+    int pk1, pk2;
+    int64_t token;
+    int ck1, ck2;
+    int v, v2;
+};
+
+SEASTAR_TEST_CASE(test_select_with_token_range_filtering) {
+    return do_with_cql_env_thread([] (auto& e) {
+        // Do tests for each column (excluding partition key columns - cannot mix partition column 
+        // restrictions and TOKEN restrictions). First run the query without index, then with global index,
+        // and finally with local index.
+
+        cquery_nofail(e, "CREATE TABLE t (pk1 int, pk2 int, ck1 int, ck2 int, v int, v2 int, PRIMARY KEY((pk1, pk2), ck1, ck2))");
+
+        std::map<std::pair<int, int>, int64_t> pk_to_token = {
+            {{3, 3}, -5763496297744201138},
+            {{2, 2}, -3974863545882308264},
+            {{1, 3},   793791837967961899},
+            {{2, 1},  1222388547083740924},
+            {{2, 3},  3312996362008120679},
+            {{1, 2},  4881097376275569167},
+            {{1, 1},  5765203080415074583},
+            {{3, 2},  6375086356864122089},
+            {{3, 1},  8740098777817515610}
+        };
+
+        std::vector<testset_row> rows;
+        auto insert_row = [&](testset_row row) {
+            cquery_nofail(e, format("INSERT INTO t(pk1, pk2, ck1, ck2, v, v2) VALUES ({}, {}, {}, {}, {}, {})", row.pk1, row.pk2, row.ck1, row.ck2, row.v, row.v2).c_str());
+            rows.push_back(row);
+        };
+
+        for (int pk1 = 1; pk1 <= 3; pk1++) {
+            for (int pk2 = 1; pk2 <= 3; pk2++) {
+                insert_row(testset_row{pk1, pk2, pk_to_token[{pk1, pk2}], 1, 1, 1, 1});
+                insert_row(testset_row{pk1, pk2, pk_to_token[{pk1, pk2}], 1, 3, 2, 1});
+                insert_row(testset_row{pk1, pk2, pk_to_token[{pk1, pk2}], 2, 1, 3, 1});
+            }
+        }
+
+        auto do_test = [&](sstring token_restriction, sstring column_restrictions, std::function<bool(const testset_row&)> matches_row) {
+            auto expected_rows = boost::copy_range<std::vector<std::vector<bytes_opt>>>(rows |
+                boost::adaptors::filtered(std::move(matches_row)) | boost::adaptors::transformed([] (const testset_row& row) {
+                return std::vector<bytes_opt> { 
+                    int32_type->decompose(row.pk1), int32_type->decompose(row.pk2), 
+                    int32_type->decompose(row.ck1), int32_type->decompose(row.ck2), 
+                    int32_type->decompose(row.v), int32_type->decompose(row.v2) 
+                };
+            }));
+            auto msg = cquery_nofail(e, format("SELECT pk1, pk2, ck1, ck2, v, v2 FROM t WHERE {} AND {} ALLOW FILTERING", token_restriction, column_restrictions).c_str());
+            assert_that(msg).is_rows().with_rows_ignore_order(expected_rows);
+        };
+
+        auto do_tests_ck1 = [&] {
+            do_test("token(pk1, pk2) >= token(1, 2)", "ck1 = 1", [&](const testset_row& row) { return row.ck1 == 1 && row.token >= pk_to_token[{1, 2}]; });
+            do_test("token(pk1, pk2) >= token(1, 2)", "ck1 = 1 AND v2 = 1", [&](const testset_row& row) { return row.ck1 == 1 && row.v2 == 1 && row.token >= pk_to_token[{1, 2}]; });
+            do_test("token(pk1, pk2) >= token(1, 2)", "ck1 = 1 AND ck2 = 3", [&](const testset_row& row) { return row.ck1 == 1 && row.ck2 == 3 && row.token >= pk_to_token[{1, 2}]; });
+            do_test("token(pk1, pk2) >= token(1, 2)", "ck1 = 1 AND ck2 = 3 AND v = 2", [&](const testset_row& row) { return row.ck1 == 1 && row.ck2 == 3 && row.v == 2 && row.token >= pk_to_token[{1, 2}]; });
+            do_test("token(pk1, pk2) >= token(1, 2)", "ck1 = 1 AND v = 3", [&](const testset_row& row) { return row.ck1 == 1 && row.v == 3 && row.token >= pk_to_token[{1, 2}]; });
+        };
+
+        auto do_tests_ck2 = [&] {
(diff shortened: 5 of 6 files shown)


avikivity:
> This PR does not change when ALLOW FILTERING clause is required, because there are multiple opinions on when it should be required.

very wise

cql3/statements/select_statement.cc
```diff
@@ -1118,12 +1118,15 @@ query::partition_slice indexed_table_select_statement::get_partition_slice_for_g
     partition_slice_builder partition_slice_builder{*_view_schema};
 
     if (!_restrictions->has_partition_key_unrestricted_components()) {
+        // Computed token column to be added to index view restrictions
+        const column_definition& token_cdef = *_view_schema->clustering_key_columns().begin();
+        // Is secondary index with new token_column_computation type (computation name: token_v2)
+        const bool is_token_v2_index = token_cdef.type == long_type;
```
avikivity, line 1124: In theory, we could support the old style too (and convert the blob token to a number and filter that). But I agree it's sufficient to support the new indexes.

avikivity: These calculations belong in the prepare state, not execution stage. Please move whatever parts are possible (those that don't consult `options` to prepare stage and store the results in members of indexed_table_select_statement. I don't expect any huge benefit from it, rather it reduces the amount of work we'll have to do later when we move even more stuff to the prepare stage.

Hm, I see you moved token_cdef from below, it's not a new variable. So no need to do anything, but keep in mind the preference to do stuff that doesn't depend on query parameters during preparation.

cql3/statements/select_statement.cc
```diff
@@ -1144,6 +1147,38 @@ query::partition_slice indexed_table_select_statement::get_partition_slice_for_g
             }
 
             partition_slice_builder.with_ranges(clustering_restrictions->bounds_ranges(options));
+        } else if (has_token(_restrictions->get_partition_key_restrictions()->expression) && is_token_v2_index) {
```
avikivity, line 1150: Just to continue that thought (again no action needed), at prepare time we could choose which path to take by selecting a sub-class of indexed_table_select_statement, rather than this run-time choice.

cql3/statements/select_statement.cc
```diff
@@ -1144,6 +1147,38 @@ query::partition_slice indexed_table_select_statement::get_partition_slice_for_g
...
+            auto token_ranges = _restrictions->get_partition_key_restrictions()->bounds_ranges(options);
+            std::vector<query::clustering_range> all_bounds;
+
+            for (auto& token_range : token_ranges) {
+                auto& start_ring_postition = token_range.start()->value();
+                auto& end_ring_position = token_range.end()->value();
```
avikivity, line 1156: Is it guaranteed that the interval is bounded? why?

cql3/statements/select_statement.cc
```diff
@@ -1144,6 +1147,38 @@ query::partition_slice indexed_table_select_statement::get_partition_slice_for_g
...
+                auto min_token_expr = expr::binary_operator{
+                    &token_cdef, include_start ? expr::oper_t::GTE : expr::oper_t::GT,
+                    ::make_shared<cql3::constants::value>(cql3::raw_value::make_value(long_type->decompose(start_position)))};
+                auto max_token_expr = expr::binary_operator{
+                    &token_cdef, include_end ? expr::oper_t::LTE : expr::oper_t::LT,
+                    ::make_shared<cql3::constants::value>(cql3::raw_value::make_value(long_type->decompose(end_position)))};
+
+                auto token_restriction = ::make_shared<restrictions::single_column_restriction>(token_cdef);
+                token_restriction->expression = expr::make_conjunction(std::move(min_token_expr), std::move(max_token_expr));
+                clustering_restrictions->merge_with(token_restriction);
```
avikivity, line 1175: Can these be built during the prepare phase?

avikivity: (using an unbound term for the token values)

avikivity: /cc @dekimir for advice

cql3/statements/select_statement.cc
avelanarius: In this particular place we know that we are using `token_restriction` (`has_token` check is essentially a `token_restriction` check - it is the only class handling those types of restrictions). 

`bounds_ranges` returned by `token_restriction` always have `start()` and `end()` present:

https://github.com/scylladb/scylla/blob/f8c7d899b40a01d238b8adb35b93e4756f43837c/cql3/restrictions/token_restriction.hh#L100-L102

avikivity: Ah! But I want to have dht::{min,max}imum_token killed (I don't want to be the one doing the killing, though) because they are redundant with interval open bounds.

So please add checks. It's dangerous to rely on the caller anyway.

dekimir:
> Adds support for `TOKEN(...)` restrictions in `SELECTs` on tables with global secondary indexes

This was supported before (ie, such queries were not rejected as invalid).  I think you mean "fixes TOKEN evaluation when combined with index-supported restriction".

Also clarify in the PR title, eg "Fix TOKEN evaluation..."

cql3/restrictions/statement_restrictions.hh
```diff
@@ -414,7 +414,11 @@ public:
     }
 
     bool ck_restrictions_need_filtering() const {
-        return _partition_key_restrictions->has_unrestricted_components(*_schema) || _clustering_columns_restrictions->needs_filtering(*_schema);
+        return _partition_key_restrictions->has_unrestricted_components(*_schema) 
+            // In TOKEN restrictions on tables with secondary indexes we only restrict idx_token column in view,
+            // leaving all partition key columns unrestricted.
```
dekimir, line 419: I'm having trouble understanding this.  Are we fetching too many rows from the index table?  That seems odd, given that TOKEN is the first clustering column there.

cql3/statements/select_statement.cc
```diff
@@ -1118,12 +1118,15 @@ query::partition_slice indexed_table_select_statement::get_partition_slice_for_g
...
 
     if (!_restrictions->has_partition_key_unrestricted_components()) {
+        // Computed token column to be added to index view restrictions
+        const column_definition& token_cdef = *_view_schema->clustering_key_columns().begin();
+        // Is secondary index with new token_column_computation type (computation name: token_v2)
+        const bool is_token_v2_index = token_cdef.type == long_type;
+
         auto single_pk_restrictions = dynamic_pointer_cast<restrictions::single_column_partition_key_restrictions>(_restrictions->get_partition_key_restrictions());
-        // Only EQ restrictions on base partition key can be used in an index view query
+        // Only EQ restrictions or TOKEN restriction on base partition key can be used in an index view query
```
dekimir, line 1127: Confusing: TOKEN will make single_pk_restrictions null.

cql3/statements/select_statement.cc
dekimir: Yes, this can theoretically be done in a `prepare` method, with abstract_marker for rhs in min/max_token_expr.

test/boost/secondary_index_test.cc
```diff
@@ -1529,26 +1550,301 @@ SEASTAR_TEST_CASE(test_token_order) {
...
+        };
+
+        eventually(do_tests);
+
+        cquery_nofail(e, "CREATE INDEX t_global ON t(v)");
+
+        eventually(do_tests);
+
+        cquery_nofail(e, "DROP INDEX t_global");
+        cquery_nofail(e, "CREATE INDEX t_local ON t((pk), v)");
```
dekimir, line 1654: Would it also make sense to test with multiple indices or partition-column indices?

dekimir (commented):
I'm having trouble reading the changes, so I'll start with some basic comments that will help me with reading the details.

cql3/restrictions/statement_restrictions.hh
avelanarius: Example that shows this case:
```
CREATE TABLE ks.t(pk1 int, ck1 int, ck2 int, v int, PRIMARY KEY(pk1, ck1, ck2));
CREATE INDEX ON ks.t(v);
SELECT * FROM ks.t WHERE v = 5 AND token(pk1) < 0 AND ck1 = 1 AND ck2 = 2;
```

Created global index has primary key: `((v), idx_token, pk1, ck1, ck2)`.

When doing the index query, we have the following partition slice: `idx_token < 0` (partition: `v = 5`). We cannot "add" `ck1` and `ck2` to this slice, because there is `pk1` in between `idx_token` and `ck1`, and we don't know the value of `pk1`. 

Therefore, we have to do some filtering on `ck1` and `ck2`. But previously `ck_restrictions_need_filtering()` returned `false`, because: `_partition_key_restrictions->has_unrestricted_components(*_schema)` was `false` - `TOKEN(pk1)` restriction and `_clustering_columns_restrictions->needs_filtering(*_schema)` was `false` - whole clustering key is restricted.

test/boost/secondary_index_test.cc
avelanarius: I don't think it would make sense to create index on partition column, as no queries could then use this index - cannot mix `TOKEN` restrictions and non-token restrictions (which would be required to cause indexed select).

In the amended commit, I added a test with multiple (2) indexes.

cql3/statements/select_statement.cc
avelanarius: Fixed it in the amended commit.

avelanarius:
@dekimir Fixed the PR description, commit messages according to the Add/Fix TOKEN distinction.

@avikivity Did not move stuff to prepare stage (as per "again no action needed" comments). Should I also do the `abstract_marker` stuff?

cql3/statements/select_statement.cc
avelanarius: `single_pk_restrictions` is only used in `EQ` case. In the amended commit, I split this comment into two, so hopefully it is not as confusing.

test/boost/secondary_index_test.cc
avelanarius: (I am speaking of `test_select_with_token_range_filtering`; `test_select_with_token_range_cases` is intended just for validating how different types of ranges are supported).

avikivity:
> @dekimir Fixed the PR description, commit messages according to the Add/Fix TOKEN distinction.
> 
> @avikivity Did not move stuff to prepare stage (as per "again no action needed" comments). Should I also do the `abstract_marker` stuff?

The stuff that was preexisting, do not move (although it would be nice in a follow-up).

But the newly introduced stuff, let's put that into the prepared statement now, so as not to introduce new technical debt.

cql3/restrictions/statement_restrictions.hh
```diff
@@ -414,7 +414,11 @@ public:
...
+            || (_uses_secondary_indexing && has_token(_partition_key_restrictions->expression))
```
dekimir, line 420: Technically, you also need `&& !_clustering_columns_restrictions->empty()` here; otherwise we may have an absurd result of no CK restrictions but ck_restrictions_need_filtering() returns true.  The current caller happens not to care, but it's better to be future-proof.

test/boost/secondary_index_test.cc
```diff
@@ -1529,26 +1550,315 @@ SEASTAR_TEST_CASE(test_token_order) {
...
+            assert_that(index_order).is_rows().with_rows(testset_pks);
         });
     });
 }
 
+SEASTAR_TEST_CASE(test_select_with_token_range_cases) {
+    return do_with_cql_env_thread([] (auto& e) {
+        cquery_nofail(e, "CREATE TABLE t (pk int, v int, PRIMARY KEY(pk))");
+
+        for (int i = 0; i < 7; i++) {
```
dekimir, line 1567: Add a comment: "v=1 in each row, so `WHERE v=1` selects them all"

test/boost/secondary_index_test.cc
```diff
@@ -1529,26 +1550,315 @@ SEASTAR_TEST_CASE(test_token_order) {
...
+            std::vector<std::vector<bytes_opt>> result;
+            for (int i = start; i <= end; i++) {
+                result.push_back({ testset_pks[i][0], int32_type->decompose(1) });
+            }
+            return result;
+        };
+
+        auto do_tests = [&] {
+            // inclusive-inclusive range
+            auto msg = cquery_nofail(e, format("SELECT * FROM t WHERE v = 1 AND token(pk) >= {} AND token(pk) <= {} ALLOW FILTERING", testset_tokens[1], testset_tokens[4]).c_str());
```
dekimir, line 1581: This pattern repeats so much that I think it should be DRYd.  Something like:
```c++
auto q = [&] (sstring_view token_restriction) { return cquery_nofail(e, format("SELECT * FROM t WHERE v = 1 AND {} ALLOW FILTERING", token_restriction); }

auto closed_range = [] (int a, int b) { return format("token(pk) <= {} AND token(pk) <= {}", a, b); }
auto open_range = [] (int a, int b) { return format("token(pk) < {} AND token(pk) < {}"; }
auto left_half_open_range = [] (int a, int b) { return format("token(pk) < {} AND token(pk) <= {}"; }
auto left_half_open_range = [] (int a, int b) { return format("token(pk) <= {} AND token(pk) < {}"; }

assert_that(q(closed_range(testset_tokens[1], testset_tokens[4])))...
```
This makes it more readable (to me, at least), and you don't need the comments because the function names tell us what's being tested.

test/boost/secondary_index_test.cc
dekimir: OK

test/boost/secondary_index_test.cc
```diff
@@ -1529,26 +1550,315 @@ SEASTAR_TEST_CASE(test_token_order) {
...
+                { int32_type->decompose(1), int32_type->decompose(2), int32_type->decompose(1), int32_type->decompose(3) },
+                { int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(1) },
+                { int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(3) },
+                { int32_type->decompose(3), int32_type->decompose(2), int32_type->decompose(1), int32_type->decompose(1) },
+                { int32_type->decompose(3), int32_type->decompose(2), int32_type->decompose(1), int32_type->decompose(3) },
+                { int32_type->decompose(3), int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(1) },
+                { int32_type->decompose(3), int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(3) }
+            });
+
+            msg = cquery_nofail(e, "SELECT pk1, pk2, ck1, ck2, v2 FROM t WHERE token(pk1, pk2) >= token(1, 2) AND ck1 = 1 AND v2 = 1 ALLOW FILTERING");
```
dekimir, line 1696: Again, I'd find this much more readable without repetitive copy-paste.

dekimir (changes requested):
I understand it now, but I do have suggestions for improved readability.

test/boost/secondary_index_test.cc
```diff
@@ -1529,26 +1550,315 @@ SEASTAR_TEST_CASE(test_token_order) {
...
+                { int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(2), int32_type->decompose(1), int32_type->decompose(1) },
+                { int32_type->decompose(3), int32_type->decompose(2), int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(1) },
+                { int32_type->decompose(3), int32_type->decompose(2), int32_type->decompose(2), int32_type->decompose(1), int32_type->decompose(1) }
+            });
+        };
+
+        // Do tests for each column (excluding partition key columns - cannot mix partition column 
+        // restrictions and TOKEN restrictions). First run the query without index, then with global index,
+        // and finally with local index.
+        eventually(do_tests_ck1);
```
dekimir, line 1814: Out of curiosity, what does `eventually` do here?  If we need those queries to happen right now, before the index is created below, isn't `eventually` foiling that?

dekimir: Sorry, I confused `eventually` with `defer`.  But I'm still curious why exceptions are expected from do_tests_ck1.

cql3/restrictions/statement_restrictions.hh
avelanarius: Maybe it would make more sense to just rename this method to `should_skip_ck_restrictions()`?

It is used only in those 4 places:

Zrzut ekranu z 2020-11-16 12-14-10

For me it makes more sense the way it works now, rather than making it more "general" by adding `&& !_clustering_columns_restrictions->empty()`.

avikivity: @avelanarius please use text quoting, not gifs, they are searchable and don't trigger allergies in old people.

dekimir: > Maybe it would make more sense to just rename this method to `should_skip_ck_restrictions()` (and negate it)?

I'd be OK with that.  Especially if the method's declaration included a good description.

cql3/restrictions/statement_restrictions.hh
```diff
@@ -414,7 +414,11 @@ public:
     }
 
     bool ck_restrictions_need_filtering() const {
-        return _partition_key_restrictions->has_unrestricted_components(*_schema) || _clustering_columns_restrictions->needs_filtering(*_schema);
+        return _partition_key_restrictions->has_unrestricted_components(*_schema)
```
nyh, line 417: I can't write a comment on the commit message, so I'll write it here.
When fixing a problem, please don't just say you "handle it properly" or "fix it". Please also explain what was the bug earlier - what was actually wrong? Did such query print an error? Fail? Produce wrong results (if so, which sort of wrong results)?

cql3/statements/select_statement.cc
```diff
@@ -1118,12 +1118,15 @@ query::partition_slice indexed_table_select_statement::get_partition_slice_for_g
     partition_slice_builder partition_slice_builder{*_view_schema};
 
     if (!_restrictions->has_partition_key_unrestricted_components()) {
+        // Computed token column to be added to index view restrictions
+        const column_definition& token_cdef = *_view_schema->clustering_key_columns().begin();
+        // Is secondary index with new token_column_computation type (computation name: token_v2)
+        const bool is_token_v2_index = token_cdef.type == long_type;
+
         auto single_pk_restrictions = dynamic_pointer_cast<restrictions::single_column_partition_key_restrictions>(_restrictions->get_partition_key_restrictions());
-        // Only EQ restrictions on base partition key can be used in an index view query
```
nyh, line 1122: I'm curious why you dropped / moved this comment?

cql3/statements/select_statement.cc
```diff
@@ -1118,12 +1118,15 @@ query::partition_slice indexed_table_select_statement::get_partition_slice_for_g
...
         if (single_pk_restrictions && single_pk_restrictions->is_all_eq()) {
+            // EQ restrictions on base partition key
             auto clustering_restrictions = ::make_shared<restrictions::single_column_clustering_key_restrictions>(_view_schema, *single_pk_restrictions);
-            // Computed token column needs to be added to index view restrictions
-            const column_definition& token_cdef = *_view_schema->clustering_key_columns().begin();
```
nyh, line 1126: I don't understand how you can drop this line, you need token_cdef a few lines down... What am I missing?

test/boost/secondary_index_test.cc
```diff
@@ -1560,6 +1560,305 @@ SEASTAR_TEST_CASE(test_token_order) {
...
+
+            // infinity-exclusive range
+            msg = cquery_nofail(e, format("SELECT * FROM t WHERE v = 1 AND token(pk) < {} ALLOW FILTERING", testset_tokens[3]).c_str());
+            assert_that(msg).is_rows().with_rows(get_result_rows(0, 2));
+
+            // negative-positive range
+            auto start_index =  3;
+            auto end_index = 5;
+            assert(testset_tokens[start_index] < 0);
+            assert(testset_tokens[end_index] > 0);
```
nyh, line 1625: Nitpick: why does it even matter if the token is < 0 or > 0? are you trying to reproduce some potential bug involving encoding of negative tokens or something? Did we have such bug?

test/boost/secondary_index_test.cc
nyh: @avelanarius I think you can create an index on one of *several* parition key columns (when the partition key is compound).

test/boost/secondary_index_test.cc
```diff
@@ -1560,6 +1560,305 @@ SEASTAR_TEST_CASE(test_token_order) {
...
+            // token equality
+            msg = cquery_nofail(e, format("SELECT * FROM t WHERE v = 1 AND token(pk) = {} ALLOW FILTERING", testset_tokens[3]).c_str());
+            assert_that(msg).is_rows().with_rows(get_result_rows(3, 3));
+
+            // empty range
+            msg = cquery_nofail(e, "SELECT * FROM t WHERE v = 1 AND token(pk) < 5 AND token(pk) > 100 ALLOW FILTERING");
+            assert_that(msg).is_rows().with_size(0);
+        };
+
+        eventually(do_tests);
```
nyh, line 1647: a comment not really related to this patch but maybe will interest you and @dekimir  - I don't like to write tests with eventually(). Every test failure (which are common during development) takes ages, while it retries the same test over and over - long after the index is built. In my new cql-pytest code I used a different approach: I build an index and then wait for the data to be available in this index. Only then I try the actual test I want to do - without any eventually(). If that test fails, it will fail immediately.

test/boost/secondary_index_test.cc
```diff
@@ -1560,6 +1560,305 @@ SEASTAR_TEST_CASE(test_token_order) {
...
+            msg = cquery_nofail(e, "SELECT pk1, pk2, ck1, ck2, v2 FROM t WHERE token(pk1, pk2) >= token(1, 1) AND token(pk1, pk2) <= token(3, 2) AND v2 = 1 AND ck2 = 1 ALLOW FILTERING");
+            assert_that(msg).is_rows().with_rows({
+                { int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(1) },
+                { int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(2), int32_type->decompose(1), int32_type->decompose(1) },
+                { int32_type->decompose(3), int32_type->decompose(2), int32_type->decompose(1), int32_type->decompose(1), int32_type->decompose(1) },
+                { int32_type->decompose(3), int32_type->decompose(2), int32_type->decompose(2), int32_type->decompose(1), int32_type->decompose(1) }
+            });
+        };
+
+        // Do tests for each column (excluding partition key columns - cannot mix partition column
```
nyh, line 1811: I would have moved this comment to the *top* or *above* the test. There is no way to guess that this is what this test is actually doing, before going 100 lines into its code.

test/boost/secondary_index_test.cc
nyh: No, but really, the first run doesn't have an index, so why does it need eventually() at all?

test/boost/secondary_index_test.cc
```diff
@@ -1560,6 +1560,305 @@ SEASTAR_TEST_CASE(test_token_order) {
...
+                { int32_type->decompose(3), int32_type->decompose(2), int32_type->decompose(2), int32_type->decompose(1), int32_type->decompose(1) }
+            });
+        };
+
+        // Do tests for each column (excluding partition key columns - cannot mix partition column 
+        // restrictions and TOKEN restrictions). First run the query without index, then with global index,
+        // and finally with local index.
+        eventually(do_tests_ck1);
+        cquery_nofail(e, "CREATE INDEX t_global ON t(ck1)");
+        eventually(do_tests_ck1);
```
nyh, line 1816: Houston, I think we have a problem...
Earlier I noted that I don't like to use eventually() for these tests. But in this case, I think using eventually isn't just annoying for development - it may even be wrong... The thing is (but please correct me if I'm wrong!), that this test will succeed even **without** an index - as you checked above. So if the index was not yet done creating, the test may simply pass without the index, and you check won't check anything!
So I think you really, as I suggested above, need to wait until the index is fully available (e.g., fetch all the different index-column values and see you can read the entire table using the index), and only then run the test.

nyh (changes requested):
Looks mostly good, and I like the extensive test!
However, I did leave a few comments and requests, please have a look. Thanks.

test/boost/secondary_index_test.cc
psarna: Nice approach, but it doesn't cover all paths - view building and updating an already existing view exercise different code paths, so we need tests for both. If a view already exists and gets updated, I'm not aware of any other mechanisms than `eventually` to make sure that the updates already arrived. One can also poll the materialized view directly, but it's much less convenient to write and depends on implementation details.

nyh: First, my suggestion doesn't need to cover **all** uses in all tests we ever wrote - it is aimed at tests like this one that I'm reviewing now - where you want to set up an index (or a view), and then run various tests on this index (or view).

Second, my suggestion **does** work even when you update a view during the test, provided the table is small enough for the approach to make sense. Quite simply, after updating the index/view, you need to again call the waiting function, which reads the whole table and then reads the whole table through the index/view - and compares the two and wait until they finally compare equal. Only when they finally compare equal, you can run the  actual check that you wanted to check, and you know that the index is already up to date and don't need an "eventually". Obviously, there is one case where this approach doesn't make much sense - if the only purpose of the eventually() was to verify that something written to the table is readable in the view, it doesn't make too much sense to read everything instead of just the one item you want to verify.

test/boost/secondary_index_test.cc
avelanarius: @nyh Yes, but if I added a test with index on one of partition key columns, I still wouldn't be able to run a query that has both `TOKEN` restriction and restriction on indexed partition column (you cannot mix those restrictions). And this is exactly the type of query I'm running in this test (it wouldn't make sense to just test `TOKEN` restriction without index or just index query without `TOKEN`).

cql3/statements/select_statement.cc
avelanarius: I moved it a few lines up, outside `if`, because it is now also used in `else` case.

test/boost/secondary_index_test.cc
avelanarius: It's leftover from the time when I implemented this PR for older indexes (those with `bytes` `idx_token` column) and this triggered different code path (due to `bytes` being unsigned).

test/boost/secondary_index_test.cc
avelanarius: From what I discussed with @psarna, my understanding is that after `CREATE INDEX`, the schema is updated (before `do_tests_ck1` is performed). Therefore next `do_tests_ck1` will read from the index (not without) - a few initial tries will fail as the index is not fully built (therefore `eventually`). Experimentally, if I did `do_tests_ck1` without `eventually` after `CREATE INDEX` it failed (not observed the query being done without index).

@psarna Is my understanding correct?

cql3/statements/select_statement.cc
avelanarius: At first I updated it to "Only EQ and TOKEN restrictions ...", but for Dejan it seemed that this comment was related to `single_pk_restrictions` which would be null:

https://github.com/scylladb/scylla/pull/7582#discussion_r520793882

Therefore I decided to split this comment into two (inside `if` and `else`), so it's more clear which part of code it comments.

cql3/restrictions/statement_restrictions.hh
avelanarius: Updated the comment in the amended commit.

cql3/restrictions/statement_restrictions.hh
avelanarius: Having fixed it according to the `should_skip_*` idea, I ultimately decided to do the `&& !_clustering_columns_restrictions->empty()` fix, as it proved to be more elegant. (changed to it in amended commit).

test/boost/secondary_index_test.cc
avelanarius: Fixed it in the amended commit.

test/boost/secondary_index_test.cc
avelanarius: I agree, `eventually()` is not needed here (for queries without an index). Removed `eventually` from those queries in amended commit.

test/boost/secondary_index_test.cc
avelanarius: In the amended commit I deleted this query (but slightly modified some other query to also be over negative-positive range, but without `assert` check).

avelanarius:
I believe I have fixed (almost) all raised suggestions in the amended commits. The only issues I think are outstanding:

1. Whether doing `eventually(some SELECTs)` immediately after `CREATE INDEX` could cause those `SELECT`s to be executed not using newly created index - @psarna ?

2. As to this comment:

> The stuff that was preexisting, do not move (although it would be nice in a follow-up).
> 
> But the newly introduced stuff, let's put that into the prepared statement now, so as not to introduce new technical debt.

Could you be more specific about "newly introduced stuff"? I can only see those (not dependant on `options`):

- `has_token(_restrictions->get_partition_key_restrictions()->expression)` (`bool`)
- `is_token_v2_index` (`bool`)
- maybe `token_cdef`? But it was present before, I just moved it a few lines above.

Other "newly introduced stuff" is `start_expr` and `end_expr`, but adding `abstract_marker` to it and extracting it to `prepare` would in my opinion add too much complexity without clear benefit. 

As I'm not that familiar with `abstract_marker`, am I correct that apart from adding `abstract_marker` to those expressions, I would have to copy options, append start, end tokens value to (new) options and then call `bind_and_get` with new options on those `start_expr`, `end_expr` here?
https://github.com/scylladb/scylla/blob/0bde9f70544b3bada83ac664d3685a3011680a4e/cql3/statements/select_statement.cc#L1183-L1191

Please provide me with guidance what to do regarding the point 2.

avikivity:
I would like to make "the number of values that are computed at execution time, but could be computed at prepare time" a monotonically decreasing function, so we don't add more work to the cleanup we have to do later.

This isn't at all costs, and if you have to invest a week of work for that then it's not worth it. But I think that building an expression with a the marker (it should be the term you got from the parser, not a new term) should be doable without a lot of effort. Please try it and if it turns out to be difficult we'll rethink it.

> As I'm not that familiar with abstract_marker, am I correct that apart from adding abstract_marker to those expressions, I would have to copy options, append start, end tokens value to (new) options and then call bind_and_get with new options on those start_expr, end_expr here?

No, creating new options defeats the purpose.

From a high level, the WHERE clause is an expression, and you are translating it to a `expression` object. Let's look at your example:

```
CREATE TABLE t(pk int, ck int, v int, PRIMARY KEY(pk, ck));
CREATE INDEX ON t(v);
SELECT * FROM t WHERE v = 5 AND token(pk) < 0 ALLOW FILTERING;
```

Here, 0 would be translated to a cql3::constants::literal object (by the parser), which you get as a term and put into the expr object. If instead the query was

```
SELECT * FROM t WHERE v = 5 AND token(pk) < :tok ALLOW FILTERING;
```

then the parser creates a cql3::constants::marker object instead. For your point of view, it's exactly the same. The options object contains the values of the bound variables and comes from the query, not you.

test/boost/secondary_index_test.cc
psarna: It is. Here's the relevant code:
https://github.com/scylladb/scylla/blob/f55b522c1bdc25dbae56c1e1ff6fe1c14c9ecc60/service/migration_manager.cc#L958-L959

avelanarius:
In the amended commits, I introduced a new `global_index_prepare_data` structure, which stores the data calculated in `prepare` stage (related to global index - for example to be used later in `get_partition_slice_for_global_index_posting_list`).

I calculate the clustering restrictions (for posting list in the `TOKEN` case) in the prepare stage. It is generated by getting the partition key restriction expression and replacing all `TOKEN` occurances with `idx_token` column (column in global index view). For example:

Original restriction (base query): `TOKEN(pk) < :tok`
Modified restriction (to be used in index query): `idx_token < :tok`

I added one new test case to unit tests - preparing the prepared statement with bind markers on `TOKEN` restrictions and executing it twice with different values.

> I would like to make "the number of values that are computed at execution time, but could be computed at prepare time" a monotonically decreasing function, so we don't add more work to the cleanup we have to do later.

Now no new values are computed at execution time (that could be moved to prepare) and I additionally moved `token_cdef` to `prepare`.

cql3/statements/select_statement.cc
avelanarius: Now my previous comment is out-of-date, as `token_cdef` is moved into `_global_index_prepare_data`.

cql3/expr/expression.cc
```diff
@@ -863,6 +863,26 @@ expression replace_column_def(const expression& expr, const column_definition* n
...
+                        [&] (const column_value& col) { return expr; },
+                        [&] (const std::vector<column_value>& cvs) { return expr; },
+                        [&] (const token&) { 
+                            return expression(binary_operator{column_value{new_cdef}, oper.op, oper.rhs});
+                        },
+                    }, oper.lhs);
+            },
+        }, expr);
+}
+
```
avikivity, line 885: I see that replace_xxx() is a popular function. I don't suggest to do it now, but if we collect a few more we can introduce a generic replace_term<Term> that will look for a class type down the tree and replace it.

cql3/statements/select_statement.cc
```diff
@@ -831,6 +831,16 @@ indexed_table_select_statement::prepare(database& db,
     const auto& im = index_opt->metadata();
     sstring index_table_name = im.name() + "_index";
     schema_ptr view_schema = db.find_schema(schema->ks_name(), index_table_name);
+    std::optional<global_index_prepare_data> global_index_data;
+
+    if (!im.local()) {
```
avikivity, line 835: We might have separate `*_indexed_table_select_statement` classes for local and global indexes, if their execution is different enough to require different data structures. Not proposing to do it now.

avikivity:
Looks good. @dekimir please review the expression usage, I think it is working out well.

avikivity:
(@avelanarius the expression stuff is relatively new and is proposed to replace the older restrictions/bounds/selector core).

cql3/expr/expression.cc
```diff
@@ -863,6 +863,26 @@ expression replace_column_def(const expression& expr, const column_definition* n
...
+            [] (bool b){ return expression(b); },
+            [&] (const conjunction& conj) {
+                const auto applied = conj.children | transformed(
+                        std::bind(replace_token, std::placeholders::_1, new_cdef));
+                return expression(conjunction{std::vector(applied.begin(), applied.end())});
+            },
+            [&] (const binary_operator& oper) {
+                return std::visit(overloaded_functor{
+                        [&] (const column_value& col) { return expr; },
+                        [&] (const std::vector<column_value>& cvs) { return expr; },
```
dekimir, line 877: Nitpick: it's better to replace the above two lines with a default case, for future-proofing:
```c=+
[&] (const auto& default_case) { return expr; }
```

cql3/statements/select_statement.cc
```diff
@@ -831,6 +830,32 @@ indexed_table_select_statement::prepare(database& db,
...
+        if (is_token_partition_slice) {
+            posting_list_clustering_restrictions = ::make_shared<restrictions::single_column_clustering_key_restrictions>(view_schema, false);
+
+            auto token_restriction = ::make_shared<restrictions::single_column_restriction>(token_cdef);
+            token_restriction->expression = expr::replace_token(partition_key_restriction_expression, &token_cdef);
+
+            posting_list_clustering_restrictions->merge_with(token_restriction);
+        }
+
+        global_index_data.emplace(global_index_prepare_data{
```
dekimir, line 855: Nitpick: with `.emplace`, why invoke the constructor explicitly?

cql3/statements/select_statement.hh
```diff
@@ -175,9 +176,25 @@ public:
...
 class indexed_table_select_statement : public select_statement {
+public:
+    // Struct which stores data related to global index,
+    // computed on the prepare stage
+    struct global_index_prepare_data {
+        // Column definition of computed token column in global index view
+        const column_definition& _token_cdef;
+
+        // Will global posting list partition be sliced according to base query TOKEN restriction
+        bool _is_token_partition_slice;
```
dekimir, line 187: If this class's invariant is `_is_token_partition_slice == bool(_posting_list_clustering_restrictions)`, maybe we should get rid of _is_token_partition_slice.  If the two can diverge, please explain in the blurb what that means.

dekimir (approved):
LGTM, just a few nitpicks.

cql3/expr/expression.cc
avelanarius: When I was writing this code I had a similar idea but decided against it in order for the code to be more strict about types.

Consider the following (maybe artificial) scenario:

```cpp
struct token {};

using var_t = std::variant<token, int, bool>;

void version1(var_t v) {
    std::visit(overloaded {
        [](token t) { puts("Processing token"); },
        [](int i) { /* Do nothing */ },
        [](bool b) { /* Do nothing */ },
    }, v);
}

void version2(var_t v) {
    std::visit(overloaded {
        [](token t) { puts("Processing token"); },
        [](auto i) { /* Do nothing */ }
    }, v);
}
```

If someone now decided to introduce a new `shiny_new_token` struct replacing `token` in variant:
```cpp
struct token {};

struct shiny_new_token {
    token t;
    int new_field;
};

using var_t = std::variant<shiny_new_token, int, bool>;
```

`version1` rightfully would fail to compile (requiring the programmer to fix it), but `version2` would compile fine, but doing the wrong thing (`shiny_new_token` going into `auto` case and doing nothing).

Of course in the actual code (`replace_token`) this issue would have been caught by unit tests. One disadvantage of this approach is that it might add boilerplate code and make changes harder to do (having to fix all `std::visit`s). And maybe in this case `token` will never be replaced by other type and no new types will require the same action as `token` (so safe to do `auto`).

cql3/statements/select_statement.hh
avelanarius: The idea behind separating those two things is that in the future more things could be moved to `prepare`. In such a case, `_posting_list_clustering_restrictions` could be used by other "cases" (not only `_is_token_partition_slice`) and `_is_token_partition_slice` would be one "case" only (`TOKEN`).

cql3/expr/expression.cc
avikivity: Agree with @avelanarius. Better to annoy someone who has to add a new do-nothing case, than to miss a case where we need to add something.

cql3/statements/select_statement.cc
avikivity: I had issues with clang not supporting it well.

avelanarius: Tried those two options:

```cpp
global_index_data.emplace(
    token_cdef, is_token_partition_slice, std::move(posting_list_clustering_restrictions)
);
```

```cpp
global_index_data.emplace({
    token_cdef, is_token_partition_slice, std::move(posting_list_clustering_restrictions)
});
```

And they don't compile under clang 10.0.0. (Didn't debug this extensively, but it seems to be caused by `token_cdef` being reference and `emplace` treating it as not a reference? `candidate template ignored: could not match 'initializer_list<type-parameter-0-0>' against 'column_definition'`)

avikivity: I think it's p0960r3, parenthesized initialization of aggregates. emplace() calls a parenthesized constructor, which gcc 10 accepts due to p0960r3, but clang rejects since it doesn't have it.

cql3/expr/expression.cc
dekimir: Annoyance and code bloat are rarely a good thing.  I'll bet you $10 that out of the next 50 changes to this function, zero will be shiny_new_token.  Almost every future evolution of `expression` will require touching a dozen functions implemented like this, wasting the coders' and reviewers' time.  I thought we invented computers to prevent just that. :)

"Miss a case" scenario is certainly possible (despite likelihood diminished by good testing coverage), but I judge it overall cheaper than the ongoing cost of maintaining implementations like this one everywhere.

cql3/statements/select_statement.cc
dekimir: OK

cql3/statements/select_statement.hh
dekimir: Either leave that separation for that future or explain the current contract.

cql3/statements/select_statement.cc
dekimir: What happens if you just say `global_index_data = global_index_prepare_data{...}`?

avelanarius: Then the copy constructor of `global_index_prepare_data` is called and it fails to compile (`object of type 'std::optional<global_index_prepare_data>' cannot be assigned because its copy assignment operator is implicitly deleted
` - probably related to reference inside it?).

test/boost/secondary_index_test.cc
nyh: I see. Ok.

avelanarius:
Is there anything left for me to fix (I think not)? Should we merge this?

/cc @psarna @avikivity @nyh

avikivity:
Queued.

avikivity:
Dequeued, breaks secondary_index_test.

avelanarius:
On my branch the tests ran correctly, however this branch was ~2 weeks behind next (as of today). 

In the meantime, this commit was merged: https://github.com/scylladb/scylla/commit/d5a6aa453335ad5594dc957556981415094a135b ("cql3: Rewrite the need_filtering logic"), but it incorrectly returns that some queries with `TOKEN` don't need filtering (even though they require it), so my tests fail (no filtering done).

psarna:
@avelanarius could you create a separate issue with examples of queries that now fail? (And if you have a clear idea how to fix them, do that?). And then we can go back to this pull request.

avelanarius:
@psarna Refs #7708

avelanarius:
Closing this old PR, as the fixed version of it: https://github.com/scylladb/scylla/pull/9067 was merged (a long time ago...)

PR #7592, opened 2020-11-11
scylla_util.py: properly parse GCP instances without size

### LINKED ISSUE #7577: offline installation: scylla_io_setup failed
*Installation details*
Scylla version (or git commit hash): scylla-unified-package-4.4.dev.0.20201110.78c598e08e.tar.gz
Cluster size: 1
OS (RHEL/CentOS/Ubuntu/AWS AMI): CentOS8 instance on GCE

### Description

In nonroot offline installation, the scylla_io_setup failed for `IndexError: list index out of range`

GCE Machine type:  e2-medium (2 vCPUs, 4 GB memory)

````
[scylla-test@amos-offlineinstall-centos7 ~]$ ./scylladb/sbin/scylla_setup 
node exporter setup is disabled by default, since the installation is offline mode.

Skip any of the following steps by answering 'no'
Do you want the Scylla server service to automatically start when the Scylla node boots?
Yes - Scylla server service automatically starts on Scylla node boot. No - skips this step. Note you will have to start the Scylla Server service manually.
[YES/no]
Created symlink /home/scylla-test/.config/systemd/user/multi-user.target.wants/scylla-server.service → /home/scylla-test/.config/systemd/user/scylla-server.service.
Do you want IOTune to study your disks IO profile and adapt Scylla to it? (*WARNING* Saying NO here means the node will not boot in production mode unless you configure the I/O Subsystem manually!)
Yes - let iotune study my disk(s). Note that this action will take a few minutes. No - skip this step.
[YES/no]
  File "/home/scylla-test/scylladb/scripts/libexec/scylla_io_setup", line 239, in <module>
    if idata.is_recommended_instance():
  File "/home/scylla-test/scylladb/scripts/scylla_util.py", line 299, in is_recommended_instance
    if self.is_recommended_instance_size() and not self.is_unsupported_instance_class() and self.is_supported_instance_class():
  File "/home/scylla-test/scylladb/scripts/scylla_util.py", line 267, in is_recommended_instance_size
    if int(self.instance_size()) > 1:
  File "/home/scylla-test/scylladb/scripts/scylla_util.py", line 237, in instance_size
    return self.instancetype.split("-")[2]
IndexError: list index out of range
IO configuration setup failed. Press any key to continue...

Do you want to install node exporter to export Prometheus data from the node? Note that the Scylla monitoring stack uses this data
Yes - install node exporter. No - skip this  step.
[yes/NO]
ScyllaDB setup finished.
[scylla-test@amos-offlineinstall-centos7 ~]$ echo $?
0

````

Failed to start scylla for lack of io setup:

- [shard 0] init - Startup failed: std::runtime_error (Bad I/O Scheduler configuration)

```
[scylla-test@amos-offlineinstall-centos7 ~]$ cat scylladb/etc/scylla.d/io.conf 
# DO NO EDIT
# This file should be automatically configure by scylla_io_setup
#
# SEASTAR_IO="--max-io-requests=1 --num-io-queues=1"
```

CC @penberg @roydahan

tarzanek:
fixes #7577

### DIFF
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -234,7 +234,8 @@ class gcp_instance:
 
     def instance_size(self):
         """Returns the size of the instance we are running in. i.e.: 2"""
-        return self.instancetype.split("-")[2]
+        instancetypesplit = self.instancetype.split("-")
+        return instancetypesplit[2] if len(instancetypesplit)>2 else 0
 
     def instance_class(self):
         """Returns the class of the instance we are running in. i.e.: n2"""
@@ -296,7 +297,7 @@ class gcp_instance:
         return self.__firstNvmeSize
 
     def is_recommended_instance(self):
-        if self.is_recommended_instance_size() and not self.is_unsupported_instance_class() and self.is_supported_instance_class():
+        if not self.is_unsupported_instance_class() and self.is_supported_instance_class() and self.is_recommended_instance_size():
             # at least 1:2GB cpu:ram ratio , GCP is at 1:4, so this should be fine
             if self.cpu/self.memoryGB < 0.5:
               # 30:1 Disk/RAM ratio must be kept at least(AWS), we relax this a little bit


fruch:
Isn't there a more portable way to get the number of CPUs ? why parsing the other instance name ?

bentsi (approved):

bentsi:
@penberg we will need to backport this to 4.3

tarzanek:
@fruch this is only about the "size" of VM, cpus are counted elsewhere
but I guess you are right
we could deduplicate this code and remove it in favour of cpu count
now the approach is more like AWS where we decide if this is or isn't a supported "instance type" and this is just one of numbers which when present in name can give some early guidance (or early refuse)
later we count cpu to ram ratio so number of cpus is definitely tested later from proper python calls and not from metadata

Issue #7057, opened 2020-08-18
Unable to ALTER system_auth tables


vladzcloudius:
*Installation details*
Scylla version: 4.0.3 (I suspect it's broken in the `master` too)
Cluster size: 1

*Description*
When I try to alter most (not all!) system_auth tables as follows I'm getting an error:

```
cqlsh> ALTER TABLE system_auth.roles with speculative_retry = 'NONE';
Unauthorized: Error from server: code=2100 [Unauthorized] message="<table system_auth.roles> is protected"
```

The error above has been received both when authorization was disabled and enabled.

This seems to be a regression from https://github.com/scylladb/scylla/commit/a855e82eff5dfe7f2b2a4f776e302d3b534dd018

vladzcloudius:
The problem comes from the `standard_role_manager::protected_resources()`.
There is also seems to be no way to disable a role manager or choose some kind of "allow all role manager" similarly to authenticator.

eliransin:
@vladzcloudius Why is this a bug? This table is protected for a reason. The system relies on a some of it's properties like schema for functionality. Lets assume for the sake of argument this is a virtual table, what meaning would be to this alter 
query now.
Another question is what is the purpose of the query?
If anything, I would expect something like a configuration option like: `internal_tables_speculative_retry`
Or another possibility is to be more fine grained about what are we protecting, for example some options like speculative retry,
to be unprotected.

vladzcloudius:
> @vladzcloudius Why is this a bug? This table is protected for a reason. The system relies on a some of it's properties like schema for functionality. Lets assume for the sake of argument this is a virtual table, what meaning would be to this alter
> query now.
> Another question is what is the purpose of the query?
> If anything, I would expect something like a configuration option like: `internal_tables_speculative_retry`
> Or another possibility is to be more fine grained about what are we protecting, for example some options like speculative retry,
> to be unprotected.

@eliransin I agree that this whole KS (together with other `system_xxx` KSes) is special - no doubt.
That's why only a privileged users should be allowed to alter them - which I think IS the case.

As to being able to alter both KS and tables - this is a fact of life and we need to do it from time to time, e.g. when we want to change its replication configuration, or like here - when we want to change parameters like `speculative_retry`.

While I do understand your concern, however we were able to do such alterations in the past and the intent of the code seems to allow this when authorization is disabled (at least).

It simply looks like a bug that we can't do it now.

[...]

PR #7593, opened 2020-11-11
Allow ALTERing the properties of system_auth tables

dekimir:
As requested in #7057, allow certain alterations of system_auth tables.  Potentially destructive alterations are still rejected.

Tests: unit (dev)

### DIFF
--- a/auth/authenticator.hh
+++ b/auth/authenticator.hh
@@ -64,6 +64,8 @@ namespace db {
     class config;
 }
 
+namespace cql3::statements { class schema_altering_statement; }
+
 namespace auth {
 
 class authenticated_user;
@@ -149,6 +151,9 @@ public:
     /// System resources used internally as part of the implementation. These are made inaccessible to users.
     ///
     virtual const resource_set& protected_resources() const = 0;
+
+    /// True iff alteration is safe for continuous functioning of this authenticator.
+    virtual bool safe(const cql3::statements::schema_altering_statement&) const { return true; }
 
     virtual ::shared_ptr<sasl_challenge> new_sasl_challenge() const = 0;
 };
--- a/auth/default_authorizer.cc
+++ b/auth/default_authorizer.cc
@@ -58,6 +58,7 @@ extern "C" {
 #include "auth/permission.hh"
 #include "auth/role_or_anonymous.hh"
 #include "cql3/query_processor.hh"
+#include "cql3/statements/schema_altering_statement.hh"
 #include "cql3/untyped_result_set.hh"
 #include "exceptions/exceptions.hh"
 #include "log.hh"
@@ -336,6 +337,10 @@ future<> default_authorizer::revoke_all(const resource& resource) const {
 const resource_set& default_authorizer::protected_resources() const {
     static const resource_set resources({ make_data_resource(meta::AUTH_KS, PERMISSIONS_CF) });
     return resources;
+}
+
+bool default_authorizer::safe(const cql3::statements::schema_altering_statement& stmt) const {
+    return !protected_resources().contains(make_data_resource(stmt.keyspace(), stmt.column_family()));
 }
 
 }
--- a/auth/default_authorizer.cc
+++ b/auth/default_authorizer.cc
@@ -58,7 +58,7 @@ extern "C" {
 #include "auth/permission.hh"
 #include "auth/role_or_anonymous.hh"
 #include "cql3/query_processor.hh"
-#include "cql3/statements/schema_altering_statement.hh"
+#include "cql3/statements/alter_table_statement.hh"
 #include "cql3/untyped_result_set.hh"
 #include "exceptions/exceptions.hh"
 #include "log.hh"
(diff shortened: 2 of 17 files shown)


auth/authenticator.hh
```diff
@@ -150,6 +152,9 @@ public:
     ///
     virtual const resource_set& protected_resources() const = 0;
 
+    /// True iff alteration is safe for continuous functioning of this authenticator.
+    virtual bool safe(const cql3::statements::schema_altering_statement&) const { return true; }
+
```
avikivity, line 157: This is entangling the cql3 hierarchy into the auth hierarchy (it is already entangled, but in an isolated way, with the table based auth* providers).

It would be better to have an intermediate representation that avoids the entanglement. I'll see later if I can suggest one.

cql3/statements/alter_table_statement.cc
```diff
@@ -70,7 +70,15 @@ alter_table_statement::alter_table_statement(shared_ptr<cf_name> name,
 }
 
 future<> alter_table_statement::check_access(service::storage_proxy& proxy, const service::client_state& state) const {
-    return state.has_column_family_access(keyspace(), column_family(), auth::permission::ALTER);
+    return state.has_column_family_access(keyspace(), column_family(), auth::permission::ALTER).then([this, &state] {
+        const auto table = column_family();
+        if (!state.get_auth_service()->underlying_role_manager().safe(*this) ||
+            !state.get_auth_service()->underlying_authenticator().safe(*this) ||
```
avikivity, line 76: Why are role manager/authenticator consulted? Is it to check that you don't mess with the tables implementing them?

I think a better approach is to register their concerns with authorizer, so we don't have to hook every internal service that happens to have protected resources.

cql3/statements/drop_table_statement.cc
```diff
@@ -58,7 +58,15 @@ future<> drop_table_statement::check_access(service::storage_proxy& proxy, const
 {
     // invalid_request_exception is only thrown synchronously.
     try {
-        return state.has_column_family_access(keyspace(), column_family(), auth::permission::DROP);
+        return state.has_column_family_access(keyspace(), column_family(), auth::permission::DROP).then([this, &state] {
+            const auto table = column_family();
+            if (!state.get_auth_service()->underlying_role_manager().safe(*this) ||
+                !state.get_auth_service()->underlying_authenticator().safe(*this) ||
+                !state.get_auth_service()->underlying_authorizer().safe(*this)) {
```
avikivity, line 65: See? it's already repeated.

auth/default_authorizer.cc
```diff
@@ -340,7 +340,20 @@ const resource_set& default_authorizer::protected_resources() const {
...
+        using alter_type = cql3::statements::alter_table_statement::type;
+        switch (alter->get_type()) {
+        case alter_type::add:
+        case alter_type::opts:
+            return true;
+        default:
+            return false;
+        }
+    }
+    return false;
```
avikivity, line 356: Ok, so I suggest adding a type `alteration_properties` exposed by authorizer and imported by cql3 (which already knows about authorizer) where alter_statement descripts what it is going to do, and authorizer  examines. I don't think it should mirror everything alter_table_statement can do, just the bits that authorizer needs, and we can expand it on demand (hopefully there won't be a huge flood of things to protect).

auth/standard_role_manager.cc
```diff
@@ -134,7 +134,20 @@ const resource_set& standard_role_manager::protected_resources() const {
 }
 
 bool standard_role_manager::safe(const cql3::statements::schema_altering_statement& stmt) const {
-    return !protected_resources().contains(make_data_resource(stmt.keyspace(), stmt.column_family()));
+    if (!protected_resources().contains(make_data_resource(stmt.keyspace(), stmt.column_family()))) {
+        return true;
+    }
+    if (auto alter = dynamic_cast<const cql3::statements::alter_table_statement*>(&stmt)) {
```
avikivity, line 140: this dynamic_cast is certainly reaching in too deeply into the smelly guts of cql3.

cql3/statements/alter_table_statement.cc
dekimir: > Why are role manager/authenticator consulted? Is it to check that you don't mess with the tables implementing them?

Yes.  This is equivalent to the old code here:
https://github.com/scylladb/scylla/blob/052a8d036d8c5ef73cb2ace333def29accde3634/auth/service.cc#L375-L377


> I think a better approach is to register their concerns with authorizer, so we don't have to hook every internal service that happens to have protected resources.

Why not hook them up to auth::service?  It's more generic and already plumbed through to permission checks.

auth/default_authorizer.cc
dekimir: Let me try.  I'm not sure I'll like having a schema_altering_statement doppelgänger just for auth to consume; these things are expensive to maintain over time and confusing when out-of-sync.  Also, the best way for authorizer/authenticator/role_manager to express what they (dis)allow is via executable code, rather than a registered data structure.

dekimir:
Based on review comments, I decided to go in a different direction.  Please see #7606.  If that gets merged, I'll close this PR.

auth/default_authorizer.cc
avikivity: The idea is for alter_table_statement/drop_stable_statement to express what they are changing, rather than for the authorizer to express what it allows. The authorizer is still code, and alter_table_statement is already data driven.

Note the intermediary does not have to be as expressive as alter_table_statement.

dekimir:
Superseded by #7606, closing.

PR #7599, opened 2020-11-12
main.cc: wait for hints manager to start

### LINKED ISSUE #7598: The future that starts hints manager is not waited on
In main.cc, we spawn a future which starts the hints manager, but we don't wait for it to complete:

https://github.com/scylladb/scylla/blob/d9810ec8ebcdce41960889ad6e522e6674661680/main.cc#L1028-L1034

This can have the following consequences:
- The hints manager does some asynchronous operations during startup, so it can take some time to start. If it is started after we start handling requests, and we admit some requests which would result in hints being generated, those hints will be dropped instead because we check if hints manager is started before writing them.
- Initialization of hints manager may fail, and Scylla won't be stopped because of it (e.g. we don't have permissions to create hints directories). The consequence of this is that hints manager won't be started, and hints will be dropped instead of being written. This may affect both regular hints manager, and the view hints manager.

As it doesn't seem intentional, we should wait until hints manager is started and check if any errors occured.

piodul:
In main.cc, we spawn a future which starts the hints manager, but we
don't wait for it to complete. This can have the following consequences:

- The hints manager does some asynchronous operations during startup,
  so it can take some time to start. If it is started after we start
  handling requests, and we admit some requests which would result in
  hints being generated, those hints will be dropped instead because we
  check if hints manager is started before writing them.
- Initialization of hints manager may fail, and Scylla won't be stopped
  because of it (e.g. we don't have permissions to create hints
  directories). The consequence of this is that hints manager won't be
  started, and hints will be dropped instead of being written. This may
  affect both regular hints manager, and the view hints manager.

This commit causes us to wait until hints manager start and see if there
were any errors during initialization.

Fixes #7598

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -1029,8 +1029,7 @@ int main(int ac, char** av) {
             proxy.invoke_on_all([] (service::storage_proxy& local_proxy) {
                 auto& ss = service::get_local_storage_service();
                 ss.register_subscriber(&local_proxy);
-                //FIXME: discarded future
-                (void)local_proxy.start_hints_manager(gms::get_local_gossiper().shared_from_this(), ss.shared_from_this());
+                return local_proxy.start_hints_manager(gms::get_local_gossiper().shared_from_this(), ss.shared_from_this());
             }).get();
 
             supervisor::notify("starting messaging service");


haaawk (approved):
LGTM. It could make sense to start the process and wait for the future later to allow other code to initialize in parallel.

psarna (approved):
Looks good, but let's wait for @vladzcloudius to ack, in case it was done on purpose

vladzcloudius (approved):
Looks good to me.
I see that I "lost" the return between the v1 and v2 of the initial HH commit when I moved the hints_manager instance into the storage_proxy.

I can't remember any good reason for this. Nor I see any comment of the series that would have explained this.
So, I suspect that this was either some debug leftover of a copy-paste bug.

Therefore I don't see why not to fix it.

PR #7600, opened 2020-11-12
scylla_util.py: increase disk to ram ratio for GCP

### LINKED ISSUE #7587: Disk to RAM ratio for GCE is not set correctly
*Installation details*
Scylla version (or git commit hash): 4.4.dev-0.20201109.8bcdf39a1
OS (RHEL/CentOS/Ubuntu/AWS AMI): CentOS 8

According to [Google](https://cloud.google.com/compute/docs/machine-types#n1_standard_machine_types) n1-standard-2 has 7.5Gb RAM 
But actually the number is ~7.3 GB:
```
>>> psutil.virtual_memory().total/1024/1024/1024
7.280799865722656
```

and when adding one 375Gb NVMe device `dist.common.scripts.scylla_util.gcp_instance.is_recommended_instance` doesn't 
return True
hence `dist/common/scripts/scylla_io_setup` runs iotune instead of using the predefined values

```
Nov 09 23:59:33 artifacts-gce-image-jenkins-db-node-38e42086-0-1 systemd[1]: Mounted Scylla data directory.
Nov 09 23:59:33 artifacts-gce-image-jenkins-db-node-38e42086-0-1 scylla_image_setup[740]: Created symlink /etc/systemd/system/multi-user.target.wants/var-lib-scylla.mount → /etc/systemd/system/var-lib-scylla.mount.
Nov 09 23:59:33 artifacts-gce-image-jenkins-db-node-38e42086-0-1 systemd[1]: Reloading.
Nov 09 23:59:33 artifacts-gce-image-jenkins-db-node-38e42086-0-1 scylla_image_setup[740]: Creating XFS volume for scylla using 1 disk(s): /dev/nvme0n1
Nov 09 23:59:34 artifacts-gce-image-jenkins-db-node-38e42086-0-1 scylla_image_setup[740]: ERROR:root:This is not a recommended Google Cloud instance setup for auto tuning, running manual iotune.
Nov 09 23:59:35 artifacts-gce-image-jenkins-db-node-38e42086-0-1 scylla_image_setup[740]: INFO  2020-11-09 23:59:35,253 [shard 0] iotune - /var/lib/scylla/saved_caches passed sanity checks
Nov 09 23:59:35 artifacts-gce-image-jenkins-db-node-38e42086-0-1 scylla_image_setup[740]: Starting Evaluation. This may take a while...
```

We should probably increase the disk to RAM ratio to 53.
@tarzanek WUT?

tarzanek:
to 60 to accomodate even 7.5GB of RAM for one nvme

fixes #7587

### DIFF
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -17,6 +17,7 @@
 
 import configparser
 import io
+import logging
 import os
 import re
 import shlex
@@ -301,19 +302,27 @@ class gcp_instance:
         if not self.is_unsupported_instance_class() and self.is_supported_instance_class() and self.is_recommended_instance_size():
             # at least 1:2GB cpu:ram ratio , GCP is at 1:4, so this should be fine
             if self.cpu/self.memoryGB < 0.5:
-              # 30:1 Disk/RAM ratio must be kept at least(AWS), we relax this a little bit
-              # on GCP we are OK with 50:1 , n1-standard-2 can cope with 1 disk, not more
-              diskCount = self.nvmeDiskCount
-              # to reach max performance for > 16 disks we mandate 32 or more vcpus
-              # https://cloud.google.com/compute/docs/disks/local-ssd#performance
-              if diskCount >= 16 and self.cpu < 32:
-                  return False
-              diskSize= self.firstNvmeSize
-              if diskCount < 1:
-                  return False
-              disktoramratio = (diskCount*diskSize)/self.memoryGB
-              if (disktoramratio <= 50) and (disktoramratio > 0):
-                  return True
+                diskCount = self.nvmeDiskCount
+                # to reach max performance for > 16 disks we mandate 32 or more vcpus
+                # https://cloud.google.com/compute/docs/disks/local-ssd#performance
+                if diskCount >= 16 and self.cpu < 32:
+                    logging.warning(
+                        "This machine doesn't have enough CPUs for allocated number of NVMEs (at least 32 cpus for >=16 disks). Performance will suffer.")
+                    return False
+                diskSize = self.firstNvmeSize
+                if diskCount < 1:
+                    return False
+                max_disktoramratio = 105
+                # 30:1 Disk/RAM ratio must be kept at least(AWS), we relax this a little bit
+                # on GCP we are OK with {max_disktoramratio}:1 , n1-standard-2 can cope with 1 disk, not more
+                disktoramratio = (diskCount * diskSize) / self.memoryGB
+                if (disktoramratio > max_disktoramratio):
+                    logging.warning(
+                        f"Instance disk-to-RAM ratio is {disktoramratio}, which is higher than the recommended ratio {max_disktoramratio}. Performance may suffer.")
+                    return False
+                return True
+            else:
+                logging.warning("At least 2G of RAM per CPU is needed. Performance will suffer.")
         return False
 
     def private_ipv4(self):


penberg:
@avikivity You had some objections to switching to another (random) number. Is there some alternative way to fix this? Make it a warning instead?

tarzanek:
@penberg it's not really a random number
we have a huge amount of clusters on AWS, where this ratio is 30:1

from my testing I relaxed it to 50:1 on GCP (to cover n2-standard-2 with one nvme, but 53 would have been better, since it would cover n1-standard-2 too) and it fits the Googles recommendations ( https://cloud.google.com/compute/docs/disks/local-ssd#performance )

Avi is right that Scylla should theoretically support 100:1 , but real clusters usage and scaling shows something else

this number basically grants that the settings for io which we blessed will work (which is what Google promises in their docs and I measured with fio, diskxplorer and iotune in my tests for various nvme disk counts and cpu counts)
if we will go for 100:1 we will bless other combinations with maximum io tuning, e.g. 2 nvme disks with n1-standard-2 - and this definitely won't be able to run on max io, simply because the cpus won't be able to handle such speeds
so with 60 we will just keep our promise on blessed io settings and in other cases depend on iotune to measure proper value (so 2 nvmes on n1-standard-2 will definitely have different speed than 2 nvmes on n1-standard-4 which will get the blessed speed from scylla_io_setup)

penberg:
@tarzanek Let me see if I understood this correctly. In your prior tests, you determined that maximum disk-to-RAM ratio is 53 (to accommodate all the instance types), but this ratio check was set to 50, which is too low? Now you increase the ratio to 60 to accommodate the instance types where that ratio is 53?

I think what Avi was asking was why is 60 better number than 53.1, for example? What I am asking is why should we fail I/O setup if we exceed the disk-to-RAM ratio at all? IOW, why not do something like this:

```patch
diff --git a/dist/common/scripts/scylla_util.py b/dist/common/scripts/scylla_util.py
index 453cae0b36..31bfe891fb 100644
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -16,6 +16,7 @@
 # along with Scylla.  If not, see <http://www.gnu.org/licenses/>.

 import configparser
+import logging
 import io
 import os
 import re
@@ -301,7 +302,7 @@ class gcp_instance:
             # at least 1:2GB cpu:ram ratio , GCP is at 1:4, so this should be fine
             if self.cpu/self.memoryGB < 0.5:
               # 30:1 Disk/RAM ratio must be kept at least(AWS), we relax this a little bit
-              # on GCP we are OK with 50:1 , n1-standard-2 can cope with 1 disk, not more
+              # on GCP we are OK with 53:1 , n1-standard-2 can cope with 1 disk, not more
               diskCount = self.nvmeDiskCount
               # to reach max performance for > 16 disks we mandate 32 or more vcpus
               # https://cloud.google.com/compute/docs/disks/local-ssd#performance
@@ -310,9 +311,11 @@ class gcp_instance:
               diskSize= self.firstNvmeSize
               if diskCount < 1:
                   return False
+              max_disktoramratio = 53
               disktoramratio = (diskCount*diskSize)/self.memoryGB
-              if (disktoramratio <= 50) and (disktoramratio > 0):
-                  return True
+              if (disktoramratio > max_disktoramratio):
+                  logging.warning(f"Instance disk-to-RAM ratio is {disktoramratio}, which is higher than the recommended ratio {max_disktoramratio}. Performance may suffer.")
+              return True
         return False

     def private_ipv4(self):
```

tarzanek:
so original number 50 was set so you could support at least one disk in n2-standard-2 (which is smallest supported nvme instance and works well with one disk) - this should have been 53 (since I missed n1-standard-2 doesn't have 8G but 7.5G of RAM, so it's my fault I didn't check all instances in detail)

new number 60 was set just to align for any exotic numbers that are not 8G of RAM but close to it

Why we want to fail?
if we fail this, iotune will run instead and we will basically measure adhoc the performance (which will be definitely worse than https://cloud.google.com/compute/docs/disks/local-ssd#performance - which is what we hardcode for properly detected recommended type)

so from my side this seems legit and it's better to figure out bugs in iotune (which Pavel did) than hardcode io numbers too high and make the scheduler overload queries by default

tarzanek:
above said I am OK with logging the problem
we can actually log the reason from all 3 if branches in the method
other than that the only log is:
"This is not a recommended Google Cloud instance setup for auto tuning, running manual iotune." from
https://github.com/scylladb/scylla/blob/master/dist/common/scripts/scylla_io_setup#L280

dist/common/scripts/scylla_util.py
```diff
@@ -299,19 +300,24 @@ def is_recommended_instance(self):
...
               diskSize= self.firstNvmeSize
               if diskCount < 1:
                   return False
+              max_disktoramratio = 60
+              # 30:1 Disk/RAM ratio must be kept at least(AWS), we relax this a little bit
+              # on GCP we are OK with {max_disktoramratio}:1 , n1-standard-2 can cope with 1 disk, not more
               disktoramratio = (diskCount*diskSize)/self.memoryGB
-              if (disktoramratio <= 50) and (disktoramratio > 0):
+              if (disktoramratio <= max_disktoramratio) and (disktoramratio > 0):
+                  logging.warning(f"Instance disk-to-RAM ratio is %s, which is higher than the recommended ratio %s. Performance may suffer.",disktoramratio,max_disktoramratio)
```
penberg, line 317: The warning should not be here, right?

penberg: Please use f-strings.

tarzanek: it should
it warns when this setup isn't optimal

tarzanek: (note I didn't use your logic from above
I think this should still fail and force iotune to be run)

tarzanek: (and these two patches should be squashed to one)

penberg: I don't understand.

Let's assume that `disktoramratio` is 49, for example, which is fine. The `if` statement expression will evaluate to `true`:

```python
if (disktoramratio <= max_disktoramratio) and (disktoramratio > 0):
# =>
if (49 <= 60) and (49 > 0):
```

We will then log a warning that says

```
Instance disk-to-RAM ratio is 49, which is higher than the recommended ratio 60. Performance may suffer.
```

Why is this what we want?

(Also, please use f-strings: https://realpython.com/python-f-strings/)

penberg: Yeah, please squash the commits, it's easier to review it that way.

tarzanek: ah, yes, sorry, will rephrase the log and fix it appropriately

tarzanek:
squashed, logic fixed
f-strings used (though here I'd love to doubt it, since my logic is that if the parameter gets evaluated at runtime, it means it will be evaluated before handing over to logging (unless python does some real magic here)
while the rule of the thumb is that logging should be lazy, so the logger starts parsing/formatting ONLY if it is enabled )

dist/common/scripts/scylla_util.py
```diff
@@ -300,19 +301,25 @@ def is_recommended_instance(self):
...
                   return False
+              max_disktoramratio = 60
+              # 30:1 Disk/RAM ratio must be kept at least(AWS), we relax this a little bit
+              # on GCP we are OK with {max_disktoramratio}:1 , n1-standard-2 can cope with 1 disk, not more
               disktoramratio = (diskCount*diskSize)/self.memoryGB
-              if (disktoramratio <= 50) and (disktoramratio > 0):
-                  return True
+              if (disktoramratio > max_disktoramratio):
+                  logging.warning(f"Instance disk-to-RAM ratio is {disktoramratio}, which is higher than the recommended ratio {max_disktoramratio}. Performance may suffer.")
+                  return False
```
penberg, line 319: So do we really want to return `False` here? You moved the ratio limit to 60, but that's arbitrary (61 could be fine just as well). So my thinking was that we turn this error into a warning instead (so always return `True`).

If you really insist returning `False` here, please argue your case, and let's move forward with one of the two solutions.

tarzanek: up to 100 should be fine so I do agree with Avi

but my argument is that I'd rather let iotune decide on actual parameters than hardcoding something that might not work

avikivity:
AWS I3en has a disk:memory ratio of 73:1.

Is there a reason GCP can only go to 60:1?

If not, let's use our standard ratio of 100:1.

tarzanek:
ok, let me then at least test n1-standard-2 with 2 (3 or 4) nvme disks
to see how bad will be the impact of allowing this instead of letting iotune measure the io
will update in few mins with results

tarzanek:
n1-standard-2 with 2 nvmes from scylla 4.3 iotune:
```
[root@gcpdisktoramtest ~]# cat /etc/scylla.d/io_properties.yaml
disks:
  - mountpoint: /var/lib/scylla
    read_iops: 351929
    read_bandwidth: 1474700800
    write_iops: 199758
    write_bandwidth: 819160320
```

our counted data would be:
read_iops = 340000
read_bandwidth = 1384120320
write_iops = 180000
write_bandwidth = 734003200

:-D

let's try with 4 disks and I will adjust to 100 then

tarzanek:
4 disks  (ratio 375*4 : 7.5 so 200:1  ):
```
[root@gcpdisktoramtest ~]# cat /etc/scylla.d/io_properties.yaml
disks:
  - mountpoint: /var/lib/scylla
    read_iops: 342238
    read_bandwidth: 2949559040
    write_iops: 252356
    write_bandwidth: 1346182144
```
io setup counted data would be (so it is wrong in this case, which is correct):
read_iops = 680000
read_bandwidth = 2768240640
write_iops = 360000
write_bandwidth = 1468006400

so above 100 our hardcoded setup definitely doesn't work

tarzanek:
I will run one more test, will update once I am done, then it can be merged

penberg:
@tarzanek How is https://github.com/scylladb/scylla/pull/7600/commits/0bfcc8d961ffe6283a170e74395bcac6f95e0ed4 related to fixing the disk/RAM ratio issue? I am having a though time understanding what it's about because there's no information in the commit description. Also, if we need to backport it, it needs to have a reference to some issue...

tarzanek:
@penberg I just checked and 4.2.1 doesn't need this fix - if we merge this will it get only to master or to 4.3 too?
I can split the patch, but it's so tiny, that the bureaucracy around it is worth it only in case we need to backport to 4.3
and it's a regression of method signature to fetch gcp metadata, before it was recursive REST for all, now it's recursive only if it's set to True and disk listing depends on recursive print of json so we can then filter only nvmes out of the list (without recursive it will only print 0,1,2,... and won't print what are those objects and some are nvmes, some are OS or other disks)

tarzanek:
ok, I see I asked a bad question
4.3 branch is cut, so I need to open a new PR for that patch, since it definitely needs to go to 4.3
and the disk to ram ratio is non critical, will split and open a new PR for the metadata recursive "True" patch

tarzanek:
@penberg is this PR OK now?

the other patch is now part of 
https://github.com/scylladb/scylla/pull/7685
and should be taken to 4.3 too

slivne:
@penberg / @avikivity ping

bentsi (changes requested):
you have bad indentation in `if self.cpu/self.memoryGB < 0.5:` block (should be 4 spaces)

image

tarzanek:
thnx Bentsi, fixed formatting - fwiw I am using the default formatter inside VS Code, is there a better way to autoformat python code in scylla?

penberg:
Please don't mix code cleanups and bug fixes in the same commit. It makes reviewing and backporting harder.

But, since this has already dragged on forever, I merged this anyway.

PR #7604, opened 2020-11-13
cql_metrics: Add metrics for CQL errors

### LINKED ISSUE #5859: add metrics for CQL errors
I think we could use a metric for when scylla replies with an error for a CQL request

right now we have various storage proxy timeouts, etc. but no (single) metric that would clearly say - Scylla replied with X amount of errors or fails per second that client had to handle

StarostaGit:
This change adds tracking of all the CQL errors that can be
raised in response to a CQL message from a client, as described
in the CQL v4 protocol and with Scylla's CDC_WRITE_FAILUREs
included.

Fixes #5859

### DIFF
--- a/transport/server.cc
+++ b/transport/server.cc
@@ -159,7 +159,7 @@ cql_server::cql_server(distributed<cql3::query_processor>& qp, auth::service& au
 {
     namespace sm = seastar::metrics;
 
-    _metrics.add_group("transport", {
+    auto ls = {
         sm::make_derive("startups", _stats.startups,
                         sm::description("Counts the total number of received CQL STARTUP messages.")),
 
@@ -207,12 +207,30 @@ cql_server::cql_server(distributed<cql3::query_processor>& qp, auth::service& au
         sm::make_derive("requests_shed", _stats.requests_shed,
                         sm::description("Holds an incrementing counter with the requests that were shed due to overload (threshold configured via max_concurrent_requests_per_shard). "
                                             "The first derivative of this value shows how often we shed requests due to overload in the \"CQL transport\" component.")),
-       sm::make_gauge("requests_memory_available", [this] { return _memory_available.current(); },
+        sm::make_gauge("requests_memory_available", [this] { return _memory_available.current(); },
                         sm::description(
                             seastar::format("Holds the amount of available memory for admitting new requests (max is {}B)."
-                                            "Zero value indicates that our bottleneck is memory and more specifically - the memory quota allocated for the \"CQL transport\" component.", _max_request_size))),
+                                            "Zero value indicates that our bottleneck is memory and more specifically - the memory quota allocated for the \"CQL transport\" component.", _max_request_size)))
+    };
 
-    });
+    std::vector<sm::metric_definition> transport_metrics;
+    for (auto& m : ls) {
+        transport_metrics.emplace_back(std::move(m));
+    }
+
+    sm::label cql_error_label("type");
+    for (const auto& e : exceptions::exception_map()) {
+        _stats.errors.insert({e.first, 0});
+        auto label_instance = cql_error_label(e.second);
+
+        transport_metrics.emplace_back(
+            sm::make_derive("cql_errors_total", sm::description("Counts the total number of returned CQL errors."),
+                        {label_instance},
+                        [this, code = e.first] { auto it = _stats.errors.find(code); return it != _stats.errors.end() ? it->second : 0; })
+        );
+    }
+
+    _metrics.add_group("transport", std::move(transport_metrics));
 }
 
 future<> cql_server::stop() {
@@ -534,24 +552,34 @@ future<foreign_ptr<std::unique_ptr<cql_server::response>>>
             tracing::set_response_size(trace_state, response->size());
             return response;
         } catch (const exceptions::unavailable_exception& ex) {
+            try { ++_server._stats.errors[ex.code()]; } catch(...) {};
             return make_unavailable_error(stream, ex.code(), ex.what(), ex.consistency, ex.required, ex.alive, trace_state);
         } catch (const exceptions::read_timeout_exception& ex) {
+            try { ++_server._stats.errors[ex.code()]; } catch(...) {};
             return make_read_timeout_error(stream, ex.code(), ex.what(), ex.consistency, ex.received, ex.block_for, ex.data_present, trace_state);
         } catch (const exceptions::read_failure_exception& ex) {
+            try { ++_server._stats.errors[ex.code()]; } catch(...) {};
             return make_read_failure_error(stream, ex.code(), ex.what(), ex.consistency, ex.received, ex.failures, ex.block_for, ex.data_present, trace_state);
         } catch (const exceptions::mutation_write_timeout_exception& ex) {
+            try { ++_server._stats.errors[ex.code()]; } catch(...) {};
             return make_mutation_write_timeout_error(stream, ex.code(), ex.what(), ex.consistency, ex.received, ex.block_for, ex.type, trace_state);
         } catch (const exceptions::mutation_write_failure_exception& ex) {
+            try { ++_server._stats.errors[ex.code()]; } catch(...) {};
             return make_mutation_write_failure_error(stream, ex.code(), ex.what(), ex.consistency, ex.received, ex.failures, ex.block_for, ex.type, trace_state);
         } catch (const exceptions::already_exists_exception& ex) {
+            try { ++_server._stats.errors[ex.code()]; } catch(...) {};
             return make_already_exists_error(stream, ex.code(), ex.what(), ex.ks_name, ex.cf_name, trace_state);
         } catch (const exceptions::prepared_query_not_found_exception& ex) {
+            try { ++_server._stats.errors[ex.code()]; } catch(...) {};
             return make_unprepared_error(stream, ex.code(), ex.what(), ex.id, trace_state);
         } catch (const exceptions::cassandra_exception& ex) {
+            try { ++_server._stats.errors[ex.code()]; } catch(...) {};
             return make_error(stream, ex.code(), ex.what(), trace_state);
         } catch (std::exception& ex) {
+            try { ++_server._stats.errors[exceptions::exception_code::SERVER_ERROR]; } catch(...) {};
             return make_error(stream, exceptions::exception_code::SERVER_ERROR, ex.what(), trace_state);
         } catch (...) {
+            try { ++_server._stats.errors[exceptions::exception_code::SERVER_ERROR]; } catch(...) {};
             return make_error(stream, exceptions::exception_code::SERVER_ERROR, "unknown error", trace_state);
         }
     });
@@ -587,10 +615,13 @@ future<> cql_server::connection::process()
             try {
                 f.get();
             } catch (const exceptions::cassandra_exception& ex) {
+                try { ++_server._stats.errors[ex.code()]; } catch(...) {};
                 write_response(make_error(0, ex.code(), ex.what(), tracing::trace_state_ptr()));
             } catch (std::exception& ex) {
+                try { ++_server._stats.errors[exceptions::exception_code::SERVER_ERROR]; } catch(...) {};
                 write_response(make_error(0, exceptions::exception_code::SERVER_ERROR, ex.what(), tracing::trace_state_ptr()));
             } catch (...) {
+                try { ++_server._stats.errors[exceptions::exception_code::SERVER_ERROR]; } catch(...) {};
                 write_response(make_error(0, exceptions::exception_code::SERVER_ERROR, "unknown error", tracing::trace_state_ptr()));
             }
         });
--- a/transport/server.hh
+++ b/transport/server.hh
@@ -136,6 +136,8 @@ private:
         uint64_t execute_requests;
         uint64_t batch_requests;
         uint64_t register_requests;
+
+        std::unordered_map<exceptions::exception_code, uint64_t> errors;
     };
 private:
     class event_notifier;
--- a/exceptions/exceptions.cc
+++ b/exceptions/exceptions.cc
@@ -44,6 +44,35 @@
 #include "exceptions.hh"
 #include "log.hh"
 
-exceptions::truncate_exception::truncate_exception(std::exception_ptr ep)
+namespace exceptions {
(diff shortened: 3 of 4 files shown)


transport/server.hh
```diff
@@ -136,6 +136,9 @@ private:
         uint64_t execute_requests;
         uint64_t batch_requests;
         uint64_t register_requests;
+
+        // errors
```
dekimir, line 140: This comment is pointless; it contains no information beyond the code.

dekimir (approved):
LGTM, one nitpick.

transport/server.cc
```diff
@@ -207,11 +207,68 @@ cql_server::cql_server(distributed<cql3::query_processor>& qp, auth::service& au
         sm::make_derive("requests_shed", _stats.requests_shed,
                         sm::description("Holds an incrementing counter with the requests that were shed due to overload (threshold configured via max_concurrent_requests_per_shard). "
                                             "The first derivative of this value shows how often we shed requests due to overload in the \"CQL transport\" component.")),
-       sm::make_gauge("requests_memory_available", [this] { return _memory_available.current(); },
+        sm::make_gauge("requests_memory_available", [this] { return _memory_available.current(); },
                         sm::description(
                             seastar::format("Holds the amount of available memory for admitting new requests (max is {}B)."
                                             "Zero value indicates that our bottleneck is memory and more specifically - the memory quota allocated for the \"CQL transport\" component.", _max_request_size))),
+        
+        sm::make_derive("server_errors", _stats.errors[exceptions::exception_code::SERVER_ERROR],
```
nyh, line 215: I hope that someone who's an expert in the overall monitoring "experience" (@amnonh ?) can review the naming of these counters. Something that bothers me is that you're adding about 20 new metrics which have a common theme (types of CQL errors) and yet there is no common naming that could make that clear. Maybe we should name all of them "cql_error_*"? Or maybe they should all be one metric with different labels - something which will also allow the user to request the sum of all errors?
I'm just raising a question here. I don't know the answer.

transport/server.cc
```diff
@@ -207,11 +207,68 @@ cql_server::cql_server(distributed<cql3::query_processor>& qp, auth::service& au
...
+                        sm::description("Counts the total number of raised CQL SERVER errors.")),
```
nyh, line 216: nitpick: I don't know if the word "raised" is the right term. It's errors "returned" to the client, no? Although I guess "raised" will be understood as well, so maybe it's ok.

transport/server.cc
```diff
@@ -207,11 +207,68 @@ cql_server::cql_server(distributed<cql3::query_processor>& qp, auth::service& au
...
+
+        sm::make_derive("protocol_errors", _stats.errors[exceptions::exception_code::PROTOCOL_ERROR],
+                        sm::description("Counts the total number of raised CQL PROTOCOL errors.")),
+
+        sm::make_derive("authentication_errors", _stats.errors[exceptions::exception_code::BAD_CREDENTIALS],
+                        sm::description("Counts the total number of raised CQL AUTHENTICATION errors.")),
```
nyh, line 222: Nitpick: this is the only case in this patch where the message differs from the name of the error code. I think here too you should write "BAD_CREDENTIALS" in te description, not "AUTHENTICATION". Unless the name BAD_CREDENTIAL itself is a mistake?

transport/server.hh
nyh: I agree. It would have been useful if there were a bunch of separate error counters, but if they are just one variable - a map - and it's called "errors", then it's rather obvious.

nyh (changes requested):

transport/server.cc
```diff
@@ -207,11 +207,68 @@ cql_server::cql_server(distributed<cql3::query_processor>& qp, auth::service& au
         sm::make_derive("requests_shed", _stats.requests_shed,
                         sm::description("Holds an incrementing counter with the requests that were shed due to overload (threshold configured via max_concurrent_requests_per_shard). "
                                             "The first derivative of this value shows how often we shed requests due to overload in the \"CQL transport\" component.")),
-       sm::make_gauge("requests_memory_available", [this] { return _memory_available.current(); },
+        sm::make_gauge("requests_memory_available", [this] { return _memory_available.current(); },
```
avikivity, line 210: Why this change?

transport/server.cc
```diff
@@ -207,11 +207,68 @@ cql_server::cql_server(distributed<cql3::query_processor>& qp, auth::service& au
...
+
+        sm::make_derive("config_errors", _stats.errors[exceptions::exception_code::CONFIG_ERROR],
+                        sm::description("Counts the total number of raised CQL CONFIG errors.")),
+
+        sm::make_derive("already_exists_errors", _stats.errors[exceptions::exception_code::ALREADY_EXISTS],
+                        sm::description("Counts the total number of raised CQL ALREADY_EXISTS errors.")),
+
+        sm::make_derive("unprepared_errors", _stats.errors[exceptions::exception_code::UNPREPARED],
+                        sm::description("Counts the total number of raised CQL UNPREPARED errors.")),
+
```
avikivity, line 271: Maybe there should be just one metric, errors, with a label for the exception code.

Adding so many metrics fills me with dread.

transport/server.cc
```diff
@@ -207,11 +207,68 @@ cql_server::cql_server(distributed<cql3::query_processor>& qp, auth::service& au
...
+        sm::make_derive("unathorized_errors", _stats.errors[exceptions::exception_code::UNAUTHORIZED],
+                        sm::description("Counts the total number of raised CQL UNATHORIZED errors.")),
+
+        sm::make_derive("invalid_errors", _stats.errors[exceptions::exception_code::INVALID],
+                        sm::description("Counts the total number of raised CQL INVALID errors.")),
+
+        sm::make_derive("config_errors", _stats.errors[exceptions::exception_code::CONFIG_ERROR],
+                        sm::description("Counts the total number of raised CQL CONFIG errors.")),
+
+        sm::make_derive("already_exists_errors", _stats.errors[exceptions::exception_code::ALREADY_EXISTS],
```
avikivity, line 266: [] can throw. Will probably wreak havoc on the metric collector.

transport/server.cc
amnonh: Adding to @avikivity and @nyh it's best to add only one metrics: errors_total (using a suffix that indicates the unit is  good practice)
with a label per error type.
Things to note:
1. It's best to have metrics initialized to zero at an early stage.
2. Labels are mutually exclusive, each error should be counted once under one of the labels, this way summing over all the labels values acts as having one global counter for all the combine errorss.
3. when you add the different metrics with their specific labels, you need to use the same description.

transport/server.hh
StarostaGit: Yes, it's a leftover from before the `map` was used, my mistake

transport/server.cc
StarostaGit: Not very adamant on the choice of that word. In fact, you might be right that 'returned' is more straightforward in its meaning

transport/server.cc
StarostaGit: I used the name of the error code `0x0100` specified in the CQL v4 Protocol (https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec - section 9). Don't know whether there's a reason our name for it differs.
Maybe somebody knows some more context?

transport/server.cc
StarostaGit: Shouldn't it be aligned with the rest of the metrics?
It's such a detrimental change that I included it in this commit since I'm editing the metrics anyway

transport/server.cc
StarostaGit: Looking at the labels then

transport/server.cc
StarostaGit: It's a small map, how likely are we to actually get something thrown out of there?
I'm now looking at labels to see if that changes what I'm using to keep track of the errors, but if not, what would you suggest I do instead? List each stat as separate member?

transport/server.cc
avikivity: We usually make changes minimal.

StarostaGit:
Addressed feedback and put everything into labels

transport/server.cc
```diff
@@ -212,6 +218,101 @@ cql_server::cql_server(distributed<cql3::query_processor>& qp, auth::service& au
                             seastar::format("Holds the amount of available memory for admitting new requests (max is {}B)."
                                             "Zero value indicates that our bottleneck is memory and more specifically - the memory quota allocated for the \"CQL transport\" component.", _max_request_size))),
 
+        sm::make_derive("cql_errors_total", sm::description("Counts the total number of returned CQL errors."),
+                        error_label_instances.contains(exceptions::exception_code::SERVER_ERROR)
+                        ? std::vector<sm::label_instance>({error_label_instances.at(exceptions::exception_code::SERVER_ERROR)}) : std::vector<sm::label_instance>{},
+                        [this] { return _stats.errors.contains(exceptions::exception_code::SERVER_ERROR) ? _stats.errors.at(exceptions::exception_code::SERVER_ERROR) : 0; }),
```
amnonh, line 224: I don't think you need the contains in the map, it will be initialized to zero

avikivity: But the initialization can throw.

avikivity: This is a double lookup, better to use find() and examine the returned iterator.

transport/server.cc
```diff
@@ -159,6 +159,12 @@ cql_server::cql_server(distributed<cql3::query_processor>& qp, auth::service& au
 {
     namespace sm = seastar::metrics;
 
+    sm::label cql_error_label("type");
+    std::unordered_map<exceptions::exception_code, sm::label_instance> error_label_instances;
```
amnonh, line 163: no need for that, just add everything.

transport/server.cc
```diff
@@ -212,6 +218,101 @@ cql_server::cql_server(distributed<cql3::query_processor>& qp, auth::service& au
                             seastar::format("Holds the amount of available memory for admitting new requests (max is {}B)."
                                             "Zero value indicates that our bottleneck is memory and more specifically - the memory quota allocated for the \"CQL transport\" component.", _max_request_size))),
 
+        sm::make_derive("cql_errors_total", sm::description("Counts the total number of returned CQL errors."),
```
amnonh, line 221: It's not going to work this way
1. always register the metrics:
a.  you will not get a second chance to do it
b. you can not register the same metrics twice (two metrics with no label)
c. it's better to start a metrics early

transport/server.cc
StarostaGit: you mean get rid of the map and put there 19 label instances? It seems rather ugly imo
And also, if anything changes in the `exceptions` namespace, creating the instances like this is just one less place to be affected

transport/server.cc
StarostaGit: Alright, I understand when I return the empty label vector it's gonna create problems. And that I should initialize all entries early.
But is there a metric that is not registered? I don't get what you're getting at with 1 and a.

amnonh: in the original code, if a metric was not used yet (i.e. it will be used in the future) you will never register it.
The metrics registration code is called once, if it was not register then, it will never ber register.

I hope at least b,c, convinced you :)

transport/server.cc
amnonh: but in your code later you manually call them, so it does not save anything, if your code would have use a loop, I can understand why you do it.

transport/server.cc
StarostaGit: I'd say it's a negligible difference when it comes to hashing ints. And this is somewhat more readable. But I can live with changing that.
Although, do we actually need all that stuff if the entries were initialized beforehand? It's true for the `error_label_instances` and can be true for the `_stats.errors` (I just set all of them to 0 in the same place I'm creating the label instances hashmap)

amnonh: @avikivity I probably missing something, if you call `_stats.errors[key]`, wouldn't it return zero if `key` is not in errors?

transport/server.cc
StarostaGit: This just makes me more confused. What do you mean by "use" the metric? How can I use it before registering it?
I feel like I'm misunderstanding the terms here - isn't what I'm doing here "registering"? I thought when I call `add_group` and pass a metric created with `sm::make_derive` for example, it's what leads to registration of said metric.

transport/server.cc
StarostaGit: I'm more keen on putting creating the metric list for the `cql` group in the loop as well then. Will do that

transport/server.cc
StarostaGit: It can resize the table underneath and throw if it can't do it

StarostaGit: At least that's the reason I'm aware of

transport/server.cc
amnonh: yes, the registration part (this code) is called once in this case during the contractor.
if you condition your metrics registration, i.g. you only register a metric if it was used, it will never be registered.

let me try to explain it differently
you have a map of errors and your map is empty and an entry is only added on the first time it is used (on the first error) you have a problem with the registration.
If you condition a registration on the existing of an entry, any entry that will not be available during the registration will never be registered even when it will be added at a future time.

StarostaGit: ok, that I understand
but I fail to see how this is conditioned on anything (apart from the situation where I return a separate label instance vector, but you mentioned that in a different point)

StarostaGit: Because in case of the error counting map, we attach the lambda to the metric and only that lambda checks the entry's existence. Is this where I'm mistaken?

amnonh: I gave you three reasons why not to condition your metrics registration on an entry existence, I think we've passed that and we agree you're not going to do it

StarostaGit:
Refactored the code a bit, in response to the above discussions

transport/server.cc
avikivity: Yes

avikivity:
What were the resolutions? Are you always registering the metrics or not?

StarostaGit:
> 
> 
> What were the resolutions? Are you always registering the metrics or not?

Yes, add the metrics in a loop now, so no lookup for the labels is needed. And changed the lambdas to use the found iterator instead of a contains/at pair.

PR #7606, opened 2020-11-14
Allow ALTERing the properties of system_auth tables

dekimir:
As requested in #7057, allow certain alterations of system_auth tables. Potentially destructive alterations are still rejected.

Tests: unit (dev)

### DIFF
--- a/auth/service.hh
+++ b/auth/service.hh
@@ -181,10 +181,17 @@ future<permission_set> get_permissions(const service&, const authenticated_user&
 ///
 bool is_enforcing(const service&);
 
+/// A description of a CQL command from which auth::service can tell whether or not this command could endanger
+/// internal data on which auth::service depends.
+struct command_desc {
+    auth::permission permission; ///< Nature of the command's alteration.
+    const resource& resource; ///< Resource impacted by this command.
+};
+
 ///
 /// Protected resources cannot be modified even if the performer has permissions to do so.
 ///
-bool is_protected(const service&, const resource&) noexcept;
+bool is_protected(const service&, command_desc) noexcept;
 
 ///
 /// Create a role with optional authentication information.
--- a/auth/service.hh
+++ b/auth/service.hh
@@ -186,6 +186,10 @@ bool is_enforcing(const service&);
 struct command_desc {
     auth::permission permission; ///< Nature of the command's alteration.
     const resource& resource; ///< Resource impacted by this command.
+    enum class type {
+        ALTER_WITH_OPTS, ///< Command is ALTER ... WITH ...
+        OTHER
+    } type_ = type::OTHER;
 };
 
 ///
--- a/service/client_state.cc
+++ b/service/client_state.cc
@@ -118,11 +118,11 @@ future<> service::client_state::has_keyspace_access(const sstring& ks,
 }
 
 future<> service::client_state::has_column_family_access(const sstring& ks,
-                const sstring& cf, auth::permission p) const {
+                const sstring& cf, auth::permission p, auth::command_desc::type t) const {
     validation::validate_column_family(ks, cf);
 
-    return do_with(ks, auth::make_data_resource(ks, cf), [this, p](const auto& ks, const auto& r) {
-        return has_access(ks, {p, r});
+    return do_with(ks, auth::make_data_resource(ks, cf), [this, p, t](const auto& ks, const auto& r) {
+        return has_access(ks, {p, r, t});
     });
 }
 
--- a/service/client_state.cc
+++ b/service/client_state.cc
@@ -106,14 +106,14 @@ future<> service::client_state::has_all_keyspaces_access(
     validate_login();
 
     return do_with(auth::resource(auth::resource_kind::data), [this, p](const auto& r) {
-        return ensure_has_permission(p, r);
+        return ensure_has_permission({p, r});
     });
 }
 
 future<> service::client_state::has_keyspace_access(const sstring& ks,
                 auth::permission p) const {
     return do_with(ks, auth::make_data_resource(ks), [this, p](auto const& ks, auto const& r) {
-        return has_access(ks, p, r);
+        return has_access(ks, {p, r});
     });
 }
 
@@ -122,7 +122,7 @@ future<> service::client_state::has_column_family_access(const sstring& ks,
     validation::validate_column_family(ks, cf);
 
     return do_with(ks, auth::make_data_resource(ks, cf), [this, p](const auto& ks, const auto& r) {
-        return has_access(ks, p, r);
+        return has_access(ks, {p, r});
     });
 }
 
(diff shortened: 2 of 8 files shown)


auth/allow_all_authenticator.hh
```diff
@@ -44,7 +44,7 @@ public:
     allow_all_authenticator(cql3::query_processor&, ::service::migration_manager&) {
     }
 
-    virtual future<> start() override {
+    virtual future<> start(service&) override {
```
avikivity, line 47: Isn't this a circular dependency? auth::service contains the authenticator, and the authenticator has a reference to service?

Better to pass the auth::service components that authenticator needs.

I don't see why authenticator needs anything from the authotizer, though.

avikivity: Ah, it's for protecting the authenticator's private property, not to use for its main work.

auth/password_authenticator.cc
```diff
@@ -149,13 +149,15 @@ future<> password_authenticator::create_default_if_missing() const {
...
-     return once_among_shards([this] {
+future<> password_authenticator::start(service& svc) {
+     return once_among_shards([this, &svc] {
          auto f = create_metadata_table_if_missing(
                  meta::roles_table::name,
                  _qp,
                  meta::roles_table::creation_query(),
-                 _migration_manager);
+                 _migration_manager).then([&svc] {
+                     svc.add_protector([] (command_desc) { return true; });
```
avikivity: Maybe add_protector should belong to authorizer, not auth::service.

auth/service.hh
```diff
@@ -69,6 +70,13 @@ public:
     using std::invalid_argument::invalid_argument;
 };
 
+/// A description of a CQL command from which auth::service can tell whether or not this command could endanger
+/// internal data on which auth::service depends.
+struct command_desc {};
+
+/// A predicate on command_desc that decides whether it's safe to execute.
+using protector = bool(*)(command_desc);
```
avikivity: Shouldn't that be std::function (or noncopyable_function)? So it can contain state?

In the bikeshed department, "protector" doesn't indicate what the bool signifies. Is it protected or allowed? Either return a bool_class, or change the name to is_protected or is_allowed.

auth/service.hh
```diff
@@ -72,10 +73,12 @@ public:
...
 /// A description of a CQL command from which auth::service can tell whether or not this command could endanger
 /// internal data on which auth::service depends.
-struct command_desc {};
+struct command_desc {
+    const resource& resource; ///< Resource impacted by this command.
+};
 
 /// A predicate on command_desc that decides whether it's safe to execute.
-using protector = bool(*)(command_desc);
+using protector = std::function<bool(command_desc)>;
```
avikivity, line 82: I knew it

auth/allow_all_authenticator.hh
```diff
@@ -88,11 +88,6 @@ public:
         return make_ready_future<custom_options>();
     }
 
-    virtual const resource_set& protected_resources() const override {
-        static const resource_set resources;
-        return resources;
-    }
-
     virtual ::shared_ptr<sasl_challenge> new_sasl_challenge() const override {
```
avikivity, line 91: We could have retained it, by making it return a protector (or vector<protector>) instead of a resource_set.

But registering a protector is nicer in a way. Btw, I think you forgot to unregister it on stop.

avikivity: (it's nicer in that the enforcer doesn't have to name the services it protects, instead the services that wish protection have to enlist)

auth/service.hh
```diff
@@ -75,6 +75,7 @@ public:
 /// internal data on which auth::service depends.
 struct command_desc {
     const resource& resource; ///< Resource impacted by this command.
+    bool alter_with_opts; ///< True iff command is ALTER ... WITH ...
```
avikivity, line 78: This should be an enum (and alter_with_opts one of the choices, and `other` as another). If we add `drop` as the next option, only one of `drop` or `alter_with_opts` can hold.

Perhaps it should be a variant, so each choice can carry some extra data if needed (that can be done if/when needed, but I think it's the logical data structure).

service/client_state.cc
```diff
@@ -118,11 +118,11 @@ future<> service::client_state::has_keyspace_access(const sstring& ks,
 }
 
 future<> service::client_state::has_column_family_access(const sstring& ks,
-                const sstring& cf, auth::permission p) const {
+                const sstring& cf, auth::permission p, bool alter_with_opts) const {
```
avikivity, line 121: I think `auth::permission p` should be replaced with `command_desc`. p already talks about ALTER, and if we add more and more variations, the parameter list will grow.

In fact the entire parameter list can be replaced with command_desc now.

avikivity:
Please make the cover letter self contained, so it can be understood without reading the issue or another pull request.

auth/password_authenticator.cc
dekimir: I don't see it that way.  `client_state` relies on a `service` member to determine access rights:
https://github.com/scylladb/scylla/blob/fcec68b1028672cffe0491e2672c1b927e79cc57/service/client_state.cc#L191

And `service`, in turn treats authorizer as just one of its clients declaring protected assets:
https://github.com/scylladb/scylla/blob/fcec68b1028672cffe0491e2672c1b927e79cc57/auth/service.cc#L374-L378

auth/service.hh
dekimir: I think it's fine to use function pointer until something more complex is needed, just like it's fine to use a simple `command_desc` until a complex one becomes necessary.

service/client_state.cc
dekimir: I like that, will do.

auth/password_authenticator.cc
avikivity: Well, it's already this way so we can keep it. But in general we should disaggregate large blobs.

authenticator is unneeded by cql and needed by transport
authorizer is needed by cql and unneeded by transport

If each gets just what it needs, recompiles are fewer.

auth/allow_all_authenticator.hh
dekimir: On second thought, registering and unregistering protectors correctly is a lot of extra code, and I'm not convinced we quite need it yet.  I decided to adapt existing is_protected to work with command_desc.

auth/service.hh
dekimir: Made it enum.

service/client_state.cc
dekimir: Done.

auth/service.hh
dekimir: Abandoned protectors.

dekimir:
Pushed v3, PTAL.  Changelog
- made cover letter self-contained
- abandoned protectors
- extended is_protected to take command_desc

PR #7616, opened 2020-11-16
Commitlog multi-entry write

### LINKED ISSUE #7615: Make commitlog accept N mutations in bulk
To enable "atomically" adding N mutations to CL (all fail or all succeed), with shared persistence cycle (i.e. part of same write cycle). 

Requested by lwt/@tgrabiec

elcallio:
Fixes #7615

Makes the CL writer interface N-valued (though still 1 for the "old" paths). Adds a new write path to input N mutations -> N rp_handles. 
Guarantees that all entries are written or none are, and that they will be flushed to disk together. 

Small test included. 

v2: 
* Improved comments
* add_entries uses set for schema inclusion check
* Test now verifies written data

v3:
* Changed file format to allow entries within entry
* Now CRC multi-entry and verify full multi-entry set before replay

v4: 
* Fixed "produce" call not waited for in replay

### DIFF
--- a/db/commitlog/commitlog.cc
+++ b/db/commitlog/commitlog.cc
@@ -62,6 +62,7 @@
 #include <seastar/core/chunked_fifo.hh>
 #include <seastar/core/queue.hh>
 #include <seastar/core/sleep.hh>
+#include <seastar/core/coroutine.hh>
 #include <seastar/net/byteorder.hh>
 
 #include "seastarx.hh"
@@ -188,14 +189,54 @@ db::commitlog::descriptor::operator db::replay_position() const {
     return replay_position(id);
 }
 
+/**
+ * virtual dispatch for actually inputting data.
+ * purposely de/un-templated
+ *
+ * Writes N entries to a single segment,
+ * where each entry has its own header+crc,
+ * i.e. will be deserialized separately.
+ */
 struct db::commitlog::entry_writer {
     force_sync sync;
-    explicit entry_writer(force_sync sync_) : sync(sync_) {}
+    size_t num_entries;
+
+    explicit entry_writer(force_sync fs, size_t ne = 1)
+        : sync(fs)
+        , num_entries(ne)
+    {}
+    virtual ~entry_writer() = default;
+
+    /** return the CF id for n:th entry */
+    virtual const cf_id_type& id(size_t) const = 0;
+
+    /**
+     * Returns segment-independent size of all entries combined. Must be >= than segment-dependant total size.
+     * This is always called first, and should return "worst-case"
+     * for the complete set of entries
+     */
+    virtual size_t size() const = 0;
+    /**
+     * Return the total size of all entries in this given segment
+     * Called after size(void), once a segment has been chosen.
+     * Should return the total, exact, size for all entries + overhead (i.e. schema)
+     * for this segment.
+     *
+     * Can be called more than once, if segment switch is neccesary (because race)
+     */
     virtual size_t size(segment&) = 0;
-    // Returns segment-independent size of the entry. Must be <= than segment-dependant size.
-    virtual size_t size() = 0;
-    virtual void write(segment&, output&) = 0;
-    virtual ~entry_writer() {};
+    /**
+     * return the size of the n:th entry in this given segment
+     * Only called IFF num_entries > 1, and if so, after size(void)/size(segment&)
+     * and before write(...)
+     */
+    virtual size_t size(segment&, size_t) = 0;
+
+    /* write nth entry */
+    virtual void write(segment&, output&, size_t) const = 0;
+
+    /** the resulting rp_handle for writing a given entry */
+    virtual void result(size_t, rp_handle) = 0;
 };
 
 const std::string db::commitlog::descriptor::SEPARATOR("-");
@@ -256,8 +297,8 @@ public:
         _request_controller.signal(size);
     }
 
-    future<rp_handle>
-    allocate_when_possible(const cf_id_type& id, shared_ptr<entry_writer> writer, db::timeout_clock::time_point timeout);
+    future<>
+    allocate_when_possible(shared_ptr<entry_writer> writer, db::timeout_clock::time_point timeout);
 
     struct stats {
         uint64_t cycle_count = 0;
@@ -507,9 +548,11 @@ public:
 
     // The commit log entry overhead in bytes (int: length + int: head checksum + int: tail checksum)
     static constexpr size_t entry_overhead_size = 3 * sizeof(uint32_t);
+    static constexpr size_t multi_entry_overhead_size = entry_overhead_size + sizeof(uint32_t);
     static constexpr size_t segment_overhead_size = 2 * sizeof(uint32_t);
     static constexpr size_t descriptor_header_size = 5 * sizeof(uint32_t);
     static constexpr uint32_t segment_magic = ('S'<<24) |('C'<< 16) | ('L' << 8) | 'C';
+    static constexpr uint32_t multi_entry_size_magic = 0xffffffff;
 
     // The commit log (chained) sync marker/header size in bytes (int: length + int: checksum [segmentId, position])
     static constexpr size_t sync_marker_size = 2 * sizeof(uint32_t);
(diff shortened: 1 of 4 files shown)


db/commitlog/commitlog.cc
```diff
@@ -1841,6 +1841,64 @@ future<db::rp_handle> db::commitlog::add_entry(const cf_id_type& id, const commi
...
+                        }
+                    }
+                }
+                i->set_with_schema(!known);
+                res += i->size();
+            }
+            return res;
+        }
+        size_t size(segment& seg, size_t i) override {
+            return _writers.at(i).size(); // we have already set schema known/unknown
```
gleb-cloudius, line 1877: `size(segment&)` has to be called before `size(segment&, size_t)` can be called? Seams kind of fragile.

db/commitlog/commitlog.cc
```diff
@@ -1841,6 +1841,64 @@ future<db::rp_handle> db::commitlog::add_entry(const cf_id_type& id, const commi
...
+        }
+        const cf_id_type& id(size_t i) const override {
+            return _writers.at(i).schema()->id();
+        }
+        size_t size(segment& seg) override {
+            size_t res = 0;
+            for (auto i = _writers.begin(), e = _writers.end(); i != e; ++i) {
+                auto known = seg.is_schema_version_known(i->schema());
+                if (!known) {
+                    for (auto j = _writers.begin(); j != i; ++j) {
```
tgrabiec, line 1864: This can become quadratic, maybe use unordered_set for "known only locally".

db/commitlog/commitlog.cc
elcallio: Given that it is 100% internal dispatching to known code I don't know if I think it really matters. Can of course set a flag just because, but it seems more reasonable to just add documentation to the internal abstract type that `size(segment&)` always is called first (total estimate), and `size(segment&, size_t)` _only_ is called if num_elem > 1.

test/boost/commitlog_test.cc
```diff
@@ -650,3 +651,37 @@ SEASTAR_TEST_CASE(test_commitlog_replay_invalid_key){
...
+                    writers.emplace_back(gen.schema(), mutations.back(), fs);
+                }
+
+                auto res = log.add_entries(std::move(writers), db::timeout_clock::now() + 60s).get0();
+
+                std::set<segment_id_type> ids;
+                for (auto& h : res) {
+                    ids.emplace(h.rp().id);
+                }
+                BOOST_CHECK_EQUAL(ids.size(), 1);
```
tgrabiec, line 682: Maybe also test that we read all mutations back

db/commitlog/commitlog.cc
elcallio: I opted against it assuming that the number of mutations in would still be relatively few, and the burden of a bunch or allocations in yet another collection was bigger than a few iterations + compares... Maybe I am wrong?

test/boost/commitlog_test.cc
elcallio: Ugh. Fine.

db/commitlog/commitlog.cc
gleb-cloudius: The interface is indeed tightly coupled with its only user. May be documentation suffice. To be honest three different size() functions confuse me. They _are_ documented, but documentation assumes knowledge of the code internals.

db/commitlog/commitlog.cc
tgrabiec: The allocations would be the only in the !known case, so once per segment.

I will use only a few mutations in my use, but who knows what happens later?

gleb-cloudius:
Question: what happens if sum of all mutations are larger than a segment size?

db/commitlog/commitlog.cc
```diff
@@ -904,41 +918,47 @@ class db::commitlog::segment : public enable_shared_from_this<segment>, public c
...
+
+            // actual data
+            auto entry_out = out.write_substream(entry_size);
+            auto entry_data = entry_out.to_input_stream();
+            writer->write(*this, entry_out, entry);
+            entry_data.with_stream([&] (auto data_str) {
+                crc.process_fragmented(ser::buffer_view<typename std::vector<temporary_buffer<char>>::iterator>(data_str));
+            });
+
+            write<uint32_t>(out, crc.checksum());
```
gleb-cloudius, line 952: Looks like we checksum each mutation separately. How can we guaranty that the write is atomic?

elcallio: The CRC is still inside the segment buffer, and in the uninterruptible write loop. The guarantee is that all the data is committed (i.e. flushed to disk) together. Since we reserve the segment space and fill all the entries inside this buffer in a single call, we either do them all, or none. 
_Reading_ will not be atomic, in the sense that we will read each entry separately, but @tgrabiec  said this was fine.

elcallio:
> 
> On Wed, Nov 18, 2020 at 02:19:09AM -0800, Calle Wilund wrote: We fail, similarly to if it was a single super-huge mutation.
> This basically means that we will not be able to process mutations of "max_mutation_size".
> […](#)
> -- Gleb.

No, not if they are submitted as a set. So the constant should be named "max single entries write payload" or something. 
Or we adjust the size checks to either go through all sizes, or just take a mean of total_size/num_elems, and then also check total_size against max segment size.

db/commitlog/commitlog.cc
gleb-cloudius: I am not sure about the guaranty. If a buffer is large and the crash happens while it is flushed to a disk why can't it be written partially? We have crc exactly to check for such cases, no?

elcallio: Well, any disk corruption, but yes. You can only really trust a multi-write if you use it in direct flush (or batch) mode, so you know things are actually on disk. 
However, as stated above, pushing all the entries into a single segment entry (crc block) does not work well either, since then you would have a single RP (handle) representing this, and can't properly signal from respective table when we no longer need this segment. Can be dealt with, but I am doubtful it would look nicer...

gleb-cloudius: The system may crash before flush completes. @tgrabiec what is the guaranty you want from that? Because as it is implemented I do not see a difference between this and writing mutations one by one.

elcallio: The only difference is that you are guaranteed to be written in a single flush call, not sequential ones. 
If you want additional crc-guarantees I would probably have to evolve the disk format to allow for nested entries or something.

tgrabiec: We need atomicity in case of a crash, so I think the CRC problem mentioned by Gleb should be fixed.

elcallio: Very well. This requires file format change, but that is what version numbers are for. Now we explicitly wrap the multi-entry in a header + crc and verify it on read.

db/commitlog/commitlog.cc
```diff
@@ -188,14 +189,54 @@ db::commitlog::descriptor::operator db::replay_position() const {
...
+     * Can be called more than once, if segment switch is neccesary (because race)
+     */
     virtual size_t size(segment&) = 0;
-    // Returns segment-independent size of the entry. Must be <= than segment-dependant size.
-    virtual size_t size() = 0;
-    virtual void write(segment&, output&) = 0;
-    virtual ~entry_writer() {};
+    /**
+     * return the size of the n:th entry in this given segment
+     * Only called IFF num_entries > 1, and if so, after size(size_t)
```
gleb-cloudius, line 230: What is `size(size_t)`? Did you mean `size()`?

db/commitlog/commitlog.cc
```diff
@@ -2046,59 +2166,118 @@ db::commitlog::read_log_file(const sstring& filename, const sstring& pfx, seasta
...
+            crc.process(size);
+
+            // check for multi-entry
+            if (d.ver >= descriptor::segment_version_2 && size == segment::multi_entry_size_magic) {
+                auto actual_size = checksum;
+                auto end = pos + actual_size - entry_header_size - sizeof(uint32_t);
+
+                assert(end <= next);
+                // really small read...
+                buf = co_await frag_reader.read_exactly(fin, sizeof(uint32_t));
```
gleb-cloudius, line 2252: croutings! nice.

db/commitlog/commitlog.cc
```diff
@@ -904,41 +946,77 @@ class db::commitlog::segment : public enable_shared_from_this<segment>, public c
...
-        replay_position rp(_desc.id, position());
-        _cf_dirty[id]++; // increase use count for cf.
+        auto& out = _buffer_ostream;
+
+        std::optional<crc32_nbo> mecrc;
+
+        // if this is multi-entry write, we need to add an extra header + crc
+        // the header and crc formula is:
+        // header:
+        //      magic : uint32_t
```
gleb-cloudius, line 962: May be worth adding `uint32_t flags` field for future extensions and check that it is zero while reading.

db/commitlog/commitlog.cc
elcallio: I did.

db/commitlog/commitlog.cc
elcallio: Can just bump the version again if any more info is needed, no?

gleb-cloudius: Yeah. The code that parses it will start to be more and more cumbersome though. But it is possible. Flags make it easier to backport features selectively, not sure it is needed here.

db/commitlog/commitlog.cc
```diff
@@ -904,41 +946,77 @@ class db::commitlog::segment : public enable_shared_from_this<segment>, public c
...
 
-        replay_position rp(_desc.id, position());
-        _cf_dirty[id]++; // increase use count for cf.
+        auto& out = _buffer_ostream;
+
+        std::optional<crc32_nbo> mecrc;
+
+        // if this is multi-entry write, we need to add an extra header + crc
+        // the header and crc formula is:
+        // header:
```
gleb-cloudius, line 961: A little bird told me to pass to you that this should be in the docs/ as well, not only comments.

avikivity: You are hearing sounds?!

gleb-cloudius: Constantly! From Google (TM) Meet.

db/commitlog/commitlog.cc
elcallio: I am opting against. If the format needs changing, I'll bump again. Saving space.

db/commitlog/commitlog.cc
elcallio: So, terrible question: Docs where? There is no commit log entry in the MD docs in the source tree. There is a very old and not very informative (nothing on file format) wiki page, but that seems lame. 
I would be happy to add a secondary issue to actually create a commit log format and mechanism(?) document later, outside this PR.

db/commitlog/commitlog.cc
elcallio: fixed.

db/commitlog/commitlog.cc
avikivity: In docs/, ok for adding it later.

db/commitlog/commitlog.cc
gleb-cloudius: I will not insist, though feature flags are superior to versioning  IMO.

elcallio:
ping?

avikivity:
Dequeued, doesn't build.

avikivity:
(discarded future - we had a window in which discarded futures weren't detected)

elcallio:
That is what you get for leaving fruit out for months. I'll fix it up.

elcallio:
rebased + fixed missing co_await

PR #7618, opened 2020-11-16
dist/common/scripts/scylla_util.py: move DEBIAN_FRONTEND environment variable to apt_install()/apt_uninstall()

syuu1228:
DEBIAN_FRONTEND environment variable was added just for prevent opening
dialog when running 'apt-get install mdadm', no other program depends on it.
So we can move it inside of apt_install()/apt_uninstall() and drop scylla_env,
since we don't have any other environment variables.
To passing the variable, added env argument on run()/out().

### DIFF
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -489,21 +489,18 @@ class aws_instance:
         return curl(self.META_DATA_BASE_URL + "user-data")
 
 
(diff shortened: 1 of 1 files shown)


syuu1228:
After this change, I think we can drop run() / out() functions, because it becoming just an alias of subprocess.run().
We can use subprocess.run() directly.

avikivity:
Agree

Issue #7610, opened 2020-11-15
test_periodic_commitlog fails with Could not retrieve CDC streams with timestamp {2020/11/13 02:20:56} upon gossip event


bhalevy:
Seen in https://jenkins.scylladb.com/job/scylla-master/job/dtest-release/654/testReport/commitlog_test/TestCommitLog/test_periodic_commitlog/
Scylla version 5b123758421fa4d3611cb69915b9bf833a3c6527
(includes 78649c2322c63f3156440d86bdecb1c16ddb2efe)
```
Traceback (most recent call last):
  File "/usr/lib64/python3.7/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib64/python3.7/unittest/case.py", line 648, in run
    self.tearDown()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/commitlog_test.py", line 38, in tearDown
    super(TestCommitLog, self).tearDown()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/dtest.py", line 1118, in tearDown
    self.cleanUpCluster()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/dtest.py", line 1164, in cleanUpCluster
    raise AssertionError('Unexpected errors found: {}'.format(found_errors))
AssertionError: Unexpected errors found: [('node1', ['ERROR 2020-11-13 02:20:35,033 [shard 0] cdc - Could not retrieve CDC streams with timestamp {2020/11/13 02:20:56} upon gossip event. Reason: "std::runtime_error (Could not find CDC generation with timestamp 2020/11/13 02:20:56 in distributed system tables (current time: 2020/11/13 02:20:35), even though some node gossiped about it.)". Action: not retrying.'])]
```

bhalevy:
@kbr- on slack:
> hypothesis: the cdc generation write did not get persisted due to the periodic commit log mode and non-clean node shutdown.
> solution (if the hypothesis is true): we need to ensure that cdc generation writes are always in synchronous commitlog mode -- we cannot consider them successful if they are not persisted.

kbr-:
Managed to reproduce locally by inserting a tactical sleep in commitlog code. In `db::commitlog::segment::cycle()`:
```
        return _pending_ops.run_with_ordered_post_op(... { ...
            return do_with(... { ...
                return repeat(... { ...
                    return seastar::sleep(5s).then(... { ...
                        return _file.dma_write(...
[...]

PR #7619, opened 2020-11-16
cdc: ensure that CDC generation write is flushed to commitlog before ack

kbr-:
When a node bootstraps or upgrades from a pre-CDC version, it creates a
new CDC generation, writes it to a distributed table
(`system_distributed.cdc_generation_descriptions`), and starts gossiping
its timestamp. When other nodes see the timestamp being gossiped, they
retrieve the generation from the table.

The bootstrapping/upgrading node therefore assumes that the generation
is made durable and other nodes will be able to retrieve it from the
table. This assumption could be invalidated if periodic commitlog mode
was used: replicas would acknowledge the write and then immediately
crash, losing the write if they were unlucky (i.e. commitlog wasn't
synced to disk before the write was acknowledged).

This commit enforces all writes to the generations table to be
synced to commitlog immediately. It does not matter for performance as
these writes are very rare.

Fixes https://github.com/scylladb/scylla/issues/7610.

Tested by inserting a tactical sleep in commitlog code which allowed me to reproduce issue #7610 with each run of the `test_periodic_commitlog` test. With this commit the bug no longer occurred.

### DIFF
--- a/db/schema_tables.cc
+++ b/db/schema_tables.cc
@@ -58,6 +58,7 @@
 #include "schema_registry.hh"
 #include "mutation_query.hh"
 #include "system_keyspace.hh"
+#include "system_distributed_keyspace.hh"
 #include "cql3/cql3_type.hh"
 #include "cql3/functions/functions.hh"
 #include "cql3/util.hh"
@@ -103,6 +104,11 @@ using namespace std::chrono_literals;
 
 
 static logging::logger diff_logger("schema_diff");
+
+static bool is_extra_durable(const sstring& ks_name, const sstring& cf_name) {
+    return (is_system_keyspace(ks_name) && db::system_keyspace::is_extra_durable(cf_name))
+        || (ks_name == db::system_distributed_keyspace::NAME && db::system_distributed_keyspace::is_extra_durable(cf_name));
+}
 
 
 /** system.schema_* tables used to store keyspace/table/type attributes prior to C* 3.0 */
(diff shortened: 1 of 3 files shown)


haaawk (approved):
LGTM

haaawk:
Have you checked the dtest stops failing with this fix @kbr- ?

kbr-:
I added a paragraph at the end of the PR letter which answers that question (refresh github if it doesn't show it).

nyh (approved):
Looks good to me. I'll commit.
@gleb-cloudius I'm worried that this is a slippery slope - that no that "extra durability" is available, everyone would want it. Who will settle with "durability" when "extra durability" is available? And this is an example. But I understood during today's status call (I hope I'm not mistaken) that Gleb is aware of this new use and aproves of it.

gleb-cloudius:
Don't you need to add a call to `builder.set_wait_for_sync_to_commitlog(true)` to `db/system_distributed_keyspace.cc:cdc_generations()` for the case where it is created for the first time?

kbr-:
@gleb-cloudius it looks like it's not needed, the `schema` object that is used in the end when writing to the table is the one returned from `create_table_from_mutations` (the one defined in `cdc_generations()` is used only to announce the mutations, but this flag is not a part of the mutations, every node decides locally if the table should be extra durable).

avikivity:
Backported to 4.3.

kbr-:
@avikivity could we also backport this to 4.2?

PR #7621, opened 2020-11-16
Type listener interface for extensions/configurables

elcallio:
Loosly coupled listen-notify injection of types to extensions and whatnot, to avoid both tangled type knowledge in main (i.e. be able to have loosly coupled extensions), and yet avoid globals for certain high-demand services. 

This poc just publishes query_processor, migration_manger, storage_service and database, but should obviously be extended. 
Note: this combo of services allows de-coupling a certain enterprise feature from globals. 

Update 2021-03-08: 
* Rebased on master
* Now with RAII-result for object init, I.e. "destroys" listened objects when returned "handle" goes out of scope.
* Supports "qualifiers" to object listeners, making it possible to distinguish instances of same object type

Basically, object flow would be:

* Extension function registers static callback in config listeners. 
* Main init calls each of these object with config + extensions registry. 
* Extension creates an instance of itself, then registers listeners for all service and whatnot objects it require. 
* Main init calls "notify" for its service objects created, keeping the returned handles on stack. 
* Scylla runs.
* Main exists, RAII objects die, calls notifications to clear out dependent objects. 
* Extension objects die.

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -805,6 +805,10 @@ int main(int ac, char** av) {
             service::storage_service_config sscfg;
             sscfg.available_memory = memory::stats().total_memory();
             service::init_storage_service(stop_signal.as_sharded_abort_source(), db, gossiper, sys_dist_ks, view_update_generator, feature_service, sscfg, mm_notifier, token_metadata, messaging, cdc_generation_service).get();
+
+            // todo: localize storage service
+            auto ss_denotify = ext->notify_type(service::get_storage_service());
+
             supervisor::notify("starting per-shard database core");
 
             sst_dir_semaphore.start(cfg->initial_sstable_loading_concurrency()).get();
@@ -824,6 +828,9 @@ int main(int ac, char** av) {
                     });
                 }).get();
             });
+
+            auto db_denotify = ext->notify_type(db);
+
             api::set_server_config(ctx).get();
             verify_seastar_io_scheduler(opts.contains("max-io-requests"), opts.contains("io-properties") || opts.contains("io-properties-file"),
                                         cfg->developer_mode()).get();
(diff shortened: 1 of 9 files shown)


avikivity:
I see. Quite clever.

I prefer a more direct and explicit way of doing these initializations, but if we can't figure one out, we can go with this.

/cc @xemul

main.cc
```diff
@@ -897,6 +897,8 @@ int main(int ac, char** av) {
             qp.start(std::ref(proxy), std::ref(db), std::ref(mm_notifier), qp_mcfg, std::ref(cql_config)).get();
             // #293 - do not stop anything
             // engine().at_exit([&qp] { return qp.stop(); });
```
xemul, line 899: When some day we'll uncomment this line all the configurables that got their pointers on qp from next line, would end up pointing to freed local qp-s.

xemul (approved):
I'm OK with the PR, but we're probably preparing the ground for use-after-free.

main.cc
elcallio: That is equally true for all things explicitly initialized with said objects. 
With notifications we could at least tweak it to automatically un-notify/unregister the instances. And instead have NPE:s in code that did not expect this. :-)

xemul: Yes, but there's a difference. Consder we have classes A and B where B references (somehow) A's instance. Currently these objects would work like this:
````
sharded<A> a;
a.start()
defer(a.stop());

sharded<B> b;
b.start(a); // set up the A reference
defer(b.stop());
````
since `b.stop()` is called before `a.stop()` it can either unreference (whatever it means) a or destroy the b instance.

With notifiers things work the other way:
````
sharded B b;
b.start(); // this subscribes for notify(a) calls
defer(b.stop());

sharded A a;
a.start();
defer(a.stop());

notify(a);
````
since there's no such thing as `unnotify(a)` once the `a.stop()` is called `b` remains with dangling reference with no chances to handle it before `b.stop()`.

I don't consider this as a blocker, though.

elcallio: This is a Poc to demonstrate what I was talking about for @avikivity. Not for merging. 
Real impl must have unregister callbacks as well, and preferably tie this directly into subclasses of sharded (so use RAII for init/deinit). 
Today we don't really have any way to shut down neither extensions nor global instances. This is a similar issue, and yet another reason why we lack real shutdown.

avikivity: We can go with this, though I really do prefer a less clever way.

elcallio:
Updated with more services published, suitable for enterprise usage. Moving this from POC to full-blown feature PR.

mykaul:
Do we still wish to pursue this? @elcallio ?

mykaul:
ping @elcallio

elcallio:
No, superceeded by type query interface previously merged. Closing.

PR #7625, opened 2020-11-16
DRAFT: validation: Validate utf8 values when creating from CQL

StarostaGit:
The code for utf8 type validation exists, but is not invoked properly.
This change fixes that.

### DIFF
--- a/test/boost/types_test.cc
+++ b/test/boost/types_test.cc
@@ -107,6 +107,13 @@ BOOST_AUTO_TEST_CASE(test_ascii_type_string_conversions) {
     test_parsing_fails(ascii_type, "¡Hola!");
 }
 
+BOOST_AUTO_TEST_CASE(test_utf8_type_string_conversions) {
+    BOOST_REQUIRE(utf8_type->equal(utf8_type->from_string("utf8"), utf8_type->decompose("utf8")));
+    BOOST_REQUIRE_EQUAL(utf8_type->to_string(utf8_type->decompose("utf8")), "utf8");
+
+    // test_parsing_fails(ascii_type, ""); <- something to fail validation
+}
+
 BOOST_AUTO_TEST_CASE(test_short_type_string_conversions) {
     BOOST_REQUIRE(short_type->equal(short_type->from_string("12345"), short_type->decompose(int16_t(12345))));
     BOOST_REQUIRE_EQUAL(short_type->to_string(short_type->decompose(int16_t(12345))), "12345");
--- a/types.cc
+++ b/types.cc
@@ -2368,8 +2368,13 @@ struct from_string_visitor {
             throw marshal_exception(format("Value not compatible with type {}: '{}'", ascii_type_name, s));
         }
     }
-    bytes operator()(const string_type_impl&) {
-        return to_bytes(bytes_view(reinterpret_cast<const int8_t*>(s.begin()), s.size()));
+    bytes operator()(const utf8_type_impl&) {
+        auto bv = bytes_view(reinterpret_cast<const int8_t*>(s.begin()), s.size());
+        if (auto pos = utils::utf8::validate_with_error_position(bv); !pos.has_value()) {
+            return to_bytes(bv);
+        } else {
+            throw marshal_exception(format("Value not compatible with type {}: '{}' at position {}", utf8_type_name, s, pos));
+        }
     }
     bytes operator()(const bytes_type_impl&) { return from_hex(s); }
     bytes operator()(const boolean_type_impl& t) {


test/boost/types_test.cc
```diff
@@ -107,6 +107,13 @@ BOOST_AUTO_TEST_CASE(test_ascii_type_string_conversions) {
     test_parsing_fails(ascii_type, "¡Hola!");
 }
 
+BOOST_AUTO_TEST_CASE(test_utf8_type_string_conversions) {
+    BOOST_REQUIRE(utf8_type->equal(utf8_type->from_string("utf8"), utf8_type->decompose("utf8")));
+    BOOST_REQUIRE_EQUAL(utf8_type->to_string(utf8_type->decompose("utf8")), "utf8");
+
+    // test_parsing_fails(ascii_type, ""); <- something to fail validation
+}
```
nyh, line 115: I don't understand what this test is testing... And what is that commented part?
I would expect a good test to "prove" your patch: The test should fail before your patch and demonstrate the bug, and pass after your patch.

types.cc
```diff
@@ -2368,8 +2368,13 @@ struct from_string_visitor {
             throw marshal_exception(format("Value not compatible with type {}: '{}'", ascii_type_name, s));
         }
     }
-    bytes operator()(const string_type_impl&) {
-        return to_bytes(bytes_view(reinterpret_cast<const int8_t*>(s.begin()), s.size()));
+    bytes operator()(const utf8_type_impl&) {
```
nyh, line 2371: Is the original implementation for string_type_impl no longer needed?

nyh (changes requested):
Beyond the specific comments I left in the code, please open an issue about the bug that this fix is solving, and add the string "Fixes #nnnn" to the commit message. This association between issues is important for the backporting process - when an issue gets closed (by that "Fixes #nnnn" string), we automatically mark it as a backport candidate, and eventually backport the commit which is marked as fixing the bug.

test/boost/types_test.cc
nyh: Moreover, the title of the pull request suggests that this patch fixes validation for *CQL*. But this test doesn't check CQL, and it can't convince me CQL isn't still broken. Remember that we *thought* it wasn't broken also before, because the checking code existed!
To check it works in CQL you can write a test in `cql_query_test.cc` or if you prefer we also have the two newer CQL testing mechanisms - `test/cql` and `test/cql-pytest`.

types.cc
```diff
@@ -2368,8 +2368,13 @@ struct from_string_visitor {
...
+        auto bv = bytes_view(reinterpret_cast<const int8_t*>(s.begin()), s.size());
+        if (auto pos = utils::utf8::validate_with_error_position(bv); !pos.has_value()) {
+            return to_bytes(bv);
+        } else {
+            throw marshal_exception(format("Value not compatible with type {}: '{}' at position {}", utf8_type_name, s, pos));
```
avikivity, line 2376: Better: `*pos`, to remove fluff in std::optional formatting.

types.cc
haaawk: probably string_type_impl extends utf8_type_impl so it will pick utf8_type_impl overload

StarostaGit: yes, ascii and utf8 combined exhaust the string type

test/boost/types_test.cc
StarostaGit: The testing part is not yet there as this was just a quick draft I made to mark it as in progress in the discussion and will complete that today (as mentioned there as well)

nyh: Good. I didn't realize this was a "draft" pull request (now I see where it says it... It's not very conspicuous).
May I ask what's the point of a "draft pull request" of this sort? Who is supposed to use it or review it?

StarostaGit: As for the CQL part, I mirrored my previous commit fixing that for ascii types. The name should probably be a bit more suited to what it actually fixes, which is broader (validation for from string conversions)
If there is concern, I can also add a CQL test explicitly as well if we want.

haaawk: I think if the change is done and only tests are missing then it makes sense to show the code to the reviewers while working on tests.

denesb: Draft PR == RFC Patch

nyh: > As for the CQL part, I mirrored my previous commit fixing that for ascii types. The name should probably be a bit more suited to what it actually fixes, which is broader (validation for from string conversions)
> If there is concern, I can also add a CQL test explicitly as well if we want.

Well, if I understood correctly, this bug was first discovered in CQL, and after some digging you managed to hunt it down to this particular change to this internal data structure. So it's good to have a unit test that verifies this internal data structure, but I think it's also good (and perhaps even more important) to verify that you fixed the real user-visible CQL bug.

haaawk: Maybe we could test the change with CQL test?

nyh: > Maybe we could test the change with CQL test?

I hereby volunteer to write one, in test/cql-pytest (in Python). I hope I'll have a ready patch for your review very soon.

nyh: This is really weird - the test I wrote works without this patch - when I try to insert broken utf-8 to a 'text' type I get:
```
E   cassandra.InvalidRequest: Error from server: code=2200 [Invalid query] message="Invalid HEX constant (0xce80) for "t" of type text"
```
So when does this bug exist?

nyh:
I just wrote a test which I hope would be a reproducer for this bug, but it demonstrates that the bug doesn't exist, so I don't understand... I tried to INSERT some invalid utf-8 into a "text" column, and the validation worked well - it refused to insert such broken utf-8.  So what is this patch actually testing? Maybe it's not INSERT but something else?
I'll post my tests in a few minutes so you can try them too if you want.

test/boost/types_test.cc
StarostaGit: Hm, I can confirm that it's working for me as well
@avikivity you reported that it's broken, how can we reproduce that?

avikivity: What's the test?

I did not reproduce myself, but saw from the code that `INSERT INTO tab(foo) VALUES('broken utf8')` will succeed, but should fail.

nyh: @avikivity I just sent to the mailing list a test which checks exactly that - inserting broken utf8 into a "text" - and it already works correctly, i.e., fails as expected, with exactly the same error message as Cassandra (something like "Invalid HEX constant (...) for "t" of type text".). So maybe you looked at the wrong code.

In any case, let's not accept this patch before understand what exactly it fixes. In what code path is this function that you're changing actually used? if you comment out this code entirely, what doesn't compile? :-)

StarostaGit: For me this fails: `INSERT INTO ks.invalid (id, t) VALUES (2, 0xC0);`, where t is of type `text`
I wonder, if what you saw was indeed failing and this one is succeeding, maybe our validation is invoked properly, but doesn't cover some specific corner cases or smth

Are you able to find what was the broken utf8 exactly?

haaawk: `saw from the code ...` xD

nyh: > For me this fails: `INSERT INTO ks.invalid (id, t) VALUES (2, 0xC0);`, where t is of type `text`
> I wonder, if what you saw was indeed failing and this one is succeeding, maybe our validation is invoked properly, but doesn't cover some specific corner cases or smth
> 
> Are you able to find what was the broken utf8 exactly?

@StarostaGit how did you run this command? Is this the command verbatim - with 0xC0 in the string and not the character 0xC0?

StarostaGit: As I'm thinking about it now, maybe the important question is whether a text value put in ' ' should be a raw string or something that accounts for special chars (after `\` for example, like `\x60`)

nyh: @StarostaGit when you say "for me this fails", I guess you mean the insert fails, i.e., Scylla succeeds, behaves as expected - right?

nyh: > @StarostaGit how did you run this command? Is this the command verbatim - with 0xC0 in the string and not the character 0xC0?

I checked with "0xC0" (I didn't know CQL had this syntax, nice, I can update my test to also check that) and the insert fails as expected. I guess I misunderstood you when you said "for me this fails" - the insert fails, so Scylla works correctly. It seems there is no bug here.

avikivity: > For me this fails: `INSERT INTO ks.invalid (id, t) VALUES (2, 0xC0);`, where t is of type `text`
> I wonder, if what you saw was indeed failing and this one is succeeding, maybe our validation is invoked properly, but doesn't cover some specific corner cases or smth
> 

This is invoking a different path. `0xC0` isn't a string literal, it's a blob literal, so it's probably rejected earlier.

StarostaGit: Yes, I meant that it "fails as expected"
Sorry for the confusion

nyh: @avikivity can you suggest a way to pass an invalid string literal rather than a blob literal?
Also, why would a blob literal be "rejected earlier"? 0xC0 is perfectly fine as a blog... Only when it is handled as a string should it be rejected.

avikivity: 0xC0 is a fine blob, but a blob is not a string. It should be rejected in the same way as `SET text_column = 3` or `SET numeric_column = 'nadav'` is rejected.

avikivity: Probably via an `assignment_testable::is_assignable` failure.

avikivity:
```c++
bytes
constants::literal::parsed_value(data_type validator) const
{
    try {
        if (_type == type::HEX && validator == bytes_type) {
            auto v = static_cast<sstring_view>(_text);
            v.remove_prefix(2);
            return validator->from_string(v);
        }
        if (validator->is_counter()) {
            return long_type->from_string(_text);
        }
        return validator->from_string(_text);
    } catch (const marshal_exception& e) {
        throw exceptions::invalid_request_exception(e.what());
    }
}
```

avikivity: Actually see `constants::literal::test_assignment`.

StarostaGit: I did also blobastext(0xC0), but this unfortunately fails to convert too soon to have any meaning here (as expected tho)
However, at least the path that was wrongly handled in the ascii case seems to be correct here - I added a test that is trying to build a `utf8_type` from a given string (in `boost/types_test.cc`):

```
void test_parsing_fails(const shared_ptr<const abstract_type>& type, sstring str)
{
    try {
        type->from_string(str);
        BOOST_FAIL(format("Parsing of '{}' should have failed", str));
    } catch (const marshal_exception& e) {
        // expected
    }
}

// other stuff, tests, all that jazz and the 

BOOST_AUTO_TEST_CASE(test_utf8_type_string_conversions) {
    test_parsing_fails(utf8_type, "\xC0\xC1");
}
```

nyh: > 0xC0 is a fine blob, but a blob is not a string. It should be rejected in the same way as `SET text_column = 3` or `SET numeric_column = 'nadav'` is rejected.

It is in fact not rejected, neither in Cassandra nor Scylla. But maybe indeed it goes through a different path. The question is how can you NOT go through that different path in a test. I'm still trying to come up with a test in CQL, not a unit tests involving our internal types. I'll let you know if I find something.

StarostaGit: Without the code snippet that you mentioned should've been failing but wasn't I'm afraid we might be wandering in the dark here

nyh: I'm sorry, I had a few bugs in my original test which made it *seem* like I found a bunch of bugs in CQL's UTF-8 validation, when in fact there are none. The new version of my tests are in this PR: https://github.com/scylladb/scylla/pull/7719
All the tests now pass on Scylla, meaning there is no visible (for these tests) bug in CQL's UTF-8 validation. Which of course we doesn't mean we don't have a bug in some internal code path that isn't reached by any of the CQL commands I tested (I actually tested 4 different CQL requests that need to part UTF-8 strings).
So now that I have a much better understanding of this issue, I'll re-review your patches.

avikivity:
@StarostaGit what's the status? Is this patch invalid, or correct? I got lost.

StarostaGit:
In light of @nyh 's recently updated tests and my own testing I think this PR can be safely closed, so will do so

nyh:
But @StarostaGit, you and @avikivity mentioned in the past that it's possible that even though the various cases I tried in my CQL tests didn't reach this low-level code which you fixed or reached it only after some higher layers verified UTF-8, there may still be cases it is used. Moreover, we shouldn't have broken low-level code - it might get used in the future and create new bugs.
So maybe this PR should not have been closed?

StarostaGit:
I guess it's a matter of whether the checks for all of them should be in that one place. The code for that is indeed invoked in the place I changed on INSERTs and UPDATEs, but it seems the errors get intercepted before they reach there. And so far, we cannot find a way to force it to behave otherwise.
When needed, I can always reopen the PR or just re-push the commit. Unless you feel like committing this would be a good prophylactic.

nyh:
What I'm worried about is that maybe today all the code which calls this function tests for UTF-8 validity earlier, but maybe next month somebody will call this function *without* testing UTF-8 validity.
The least ideal solution is to add some comments saying you shouldn't call this function without validating UTF-8 first.
Another not-very-ideal solution is to add validation (like you did), and strings will always be validated twice, in two layers.
Another possible solution is to have the validation *only* in the lower level (like you added here), and remove the validation call from the higher layer. I'm afraid I am not really familiar with the code and which layers do these validation calls, so I can't give specific examples.

I don't know what I prefer that you do. I thought that @avikivity expressed somewhere in this discussion his preference that the low-level function be fixed if we believe it is incorrect.

StarostaGit:
I wouldn't remove the checks from the layers above as from what I understand they serve a different purpose (e.g. check correctness of the whole query string). However, I don't think it would hurt as to put the check there and be more secure about it, I agree.

nyh:
@StarostaGit think of simpler cases (some of which are exercised in my test). Perhaps the simplest example is this:
We have the `blobToText(0x12345)` CQL function which takes a blob and converts it to text. Apparently, this function validates utf-8, and then calls the text object's constructor, which would (if I understand your patch correctly) validate utf-8 again. The higher-level validation in this case is completely unnecessary if it knows the lower level will validate again.

Even in the case of the whole-query-string validation, if it even exists (I'm not sure if it does, my test cannot yet exercise this case), the value of a separate whole-string validation is dubious (non-UTF8 garbage will either be rejected as a syntax error, or be part of a string and can be validated as part of that string). It perhaps has some minor value for producing a better error in this case - because the CQL specification specifies that the entire query string needs to be encoded in UTF-8.

StarostaGit:
So I've found the function below:
```cpp
inline
shared_ptr<function>
make_from_blob_function(data_type to_type) {
    sstring name = sstring("blobas") + to_type->as_cql3_type().to_string();
    return make_native_scalar_function<true>(name, to_type, { bytes_type },
            [name, to_type] (cql_serialization_format sf, const std::vector<bytes_opt>& parameters) -> bytes_opt {
        auto&& val = parameters[0];
        if (!val) {
            return val;
        }
        try {
            to_type->validate(*val, sf);
            return val;
        } catch (marshal_exception& e) {
            using namespace exceptions;
            throw invalid_request_exception(format("In call to function {}, value 0x{} is not a valid binary representation for type {}",
                    name, to_hex(val), to_type->as_cql3_type().to_string()));
        }
    });
}
```
From my understanding this is the implementation of the `blobAsText()` from CQL. It's located in the `cql3/functions/bytes_conversions_fcts.hh` file.
Notice that it doesn't return the actual type we're converting to, but instead validates and returns `std::optional<bytes>`. That means if we remove validation from here, the caller might never actually create the utf8 type from this and it will get fed forward without any validation.

PR #7626, opened 2020-11-17
repair: Make removenode safe by default

### LINKED ISSUE #7359: Repair-based streaming is "best effort" which leads to linearizability violation during nodetool removenode

asias:
Currently removenode works like below:

- The coordinator node advertises the node to be removed in
  REMOVING_TOKEN status in gossip

- Existing nodes learn the node in REMOVING_TOKEN status

- Existing nodes sync data for the range it owns

- Existing nodes send notification to the coordinator

- The coordinator node waits for notification and announce the node in
  REMOVED_TOKEN

Current problems:

- Existing nodes do not tell the coordinator if the data sync is ok or failed.

- The coordinator can not abort the removenode operation in case of error

- Failed removenode operation will make the node to be removed in
  REMOVING_TOKEN forever.

- The removenode runs in best effort mode which may cause data
  consistency issues.

  It means if a node that owns the range after the removenode
  operation is down during the operation, the removenode node operation
  will continue to succeed without requiring that node to perform data
  syncing. This can cause data consistency issues.

  For example, Five nodes in the cluster, RF = 3, for a range, n1, n2,
  n3 is the old replicas, n2 is being removed, after the removenode
  operation, the new replicas are n1, n5, n3. If n3 is down during the
  removenode operation, only n1 will be used to sync data with the new
  owner n5. This will break QUORUM read consistency if n1 happens to
  miss some writes.

Improvements in this patch:

- This patch makes the removenode safe by default.

We require all nodes in the cluster to participate in the removenode operation and
sync data if needed. We fail the removenode operation if any of them is down or
fails.

If the user want the removenode operation to succeed even if some of the nodes
are not available, the user has to explicitly pass a list of nodes that can be
skipped for the operation.

$ nodetool removenode --ignore-dead-nodes <list_of_dead_nodes_to_ignore> <host_id>

Example restful api:

$ curl -X POST "http://127.0.0.1:10000/storage_service/remove_node/?host_id=7bd303e9-4c7b-4915-84f6-343d0dbd9a49&ignore_nodes=127.0.0.3,127.0.0.5"

- The coordinator can abort data sync on existing nodes

For example, if one of the nodes fails to sync data. It makes no sense for
other nodes to continue to sync data because the whole operation will
fail anyway.

- The coordinator can decide which nodes to ignore and pass the decision
  to other nodes

Previously, there is no way for the coordinator to tell existing nodes
to run in strict mode or best effort mode. Users will have to modify
config file or run a restful api cmd on all the nodes to select strict
or best effort mode. With this patch, the cluster wide configuration is
eliminated.

Fixes #7359

### DIFF
--- a/api/api-doc/storage_service.json
+++ b/api/api-doc/storage_service.json
@@ -1105,6 +1105,14 @@
                      "allowMultiple":false,
                      "type":"string",
                      "paramType":"query"
+                  },
+                  {
+                     "name":"ignore_nodes",
+                     "description":"List of dead nodes to ingore in removenode operation",
+                     "required":false,
+                     "allowMultiple":false,
+                     "type":"string",
+                     "paramType":"query"
                   }
                ]
             }
--- a/api/storage_service.cc
+++ b/api/storage_service.cc
@@ -27,6 +27,7 @@
 #include <time.h>
 #include <boost/range/adaptor/map.hpp>
 #include <boost/range/adaptor/filtered.hpp>
+#include <boost/algorithm/string/trim_all.hpp>
 #include "service/storage_service.hh"
 #include "service/load_meter.hh"
 #include "db/commitlog/commitlog.hh"
@@ -496,7 +497,22 @@ void set_storage_service(http_context& ctx, routes& r) {
 
     ss::remove_node.set(r, [](std::unique_ptr<request> req) {
         auto host_id = req->get_query_param("host_id");
-        return service::get_local_storage_service().removenode(host_id).then([] {
+        std::vector<sstring> ignore_nodes_strs= split(req->get_query_param("ignore_nodes"), ",");
+        auto ignore_nodes = std::list<gms::inet_address>();
+        for (std::string n : ignore_nodes_strs) {
+            try {
+                std::replace(n.begin(), n.end(), '\"', ' ');
+                std::replace(n.begin(), n.end(), '\'', ' ');
+                boost::trim_all(n);
+                if (!n.empty()) {
+                    auto node = gms::inet_address(n);
+                    ignore_nodes.push_back(node);
+                }
+            } catch (...) {
+                throw std::runtime_error(format("Failed to parse ignore_nodes parameter: ignore_nodes={}, node={}", ignore_nodes_strs, n));
+            }
+        }
+        return service::get_local_storage_service().removenode(host_id, std::move(ignore_nodes)).then([] {
             return make_ready_future<json::json_return_type>(json_void());
         });
     });
--- a/idl/partition_checksum.idl.hh
+++ b/idl/partition_checksum.idl.hh
@@ -103,3 +103,22 @@ enum class repair_row_level_start_status: uint8_t {
 struct repair_row_level_start_response {
     repair_row_level_start_status status;
 };
+
+enum class node_ops_cmd : uint32_t {
+     removenode_prepare,
+     removenode_heartbeat,
+     removenode_sync_data,
+     removenode_abort,
+     removenode_done,
+};
+
+struct node_ops_cmd_request {
+    node_ops_cmd cmd;
+    utils::UUID ops_uuid;
+    std::list<gms::inet_address> ignore_nodes;
+    std::list<gms::inet_address> leaving_nodes;
+};
+
+struct node_ops_cmd_response {
+    bool ok;
+};
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -53,6 +53,14 @@ logging::logger rlogger("repair");
 
 static sharded<netw::messaging_service>* _messaging;
 
+void node_ops_info::check_abort() {
+    if (abort) {
+        auto msg = format("Node operation with ops_uuid={} is aborted", ops_uuid);
+        rlogger.warn("{}", msg);
+        throw std::runtime_error(msg);
+    }
+}
+
 class node_ops_metrics {
 public:
     node_ops_metrics() {
@@ -434,6 +442,16 @@ void tracker::abort_all_repairs() {
     rlogger.info0("Aborted {} repair job(s)", count);
 }
 
+void tracker::abort_repair_node_ops(utils::UUID ops_uuid) {
+    for (auto& x : _repairs[this_shard_id()]) {
+        auto& ri = x.second;
+        if (ri->ops_uuid() && ri->ops_uuid().value() == ops_uuid) {
+            rlogger.info0("Aborted repair jobs for ops_uuid={}", ops_uuid);
+            ri->abort();
+        }
+    }
+}
+
 float tracker::report_progress(streaming::stream_reason reason) {
     uint64_t nr_ranges_finished = 0;
     uint64_t nr_ranges_total = 0;
@@ -792,7 +810,8 @@ repair_info::repair_info(seastar::sharded<database>& db_,
     repair_uniq_id id_,
     const std::vector<sstring>& data_centers_,
     const std::vector<sstring>& hosts_,
-    streaming::stream_reason reason_)
+    streaming::stream_reason reason_,
+    std::optional<utils::UUID> ops_uuid)
     : db(db_)
     , messaging(ms_)
     , sharder(get_sharder_for_tables(db_, keyspace_, table_ids_))
@@ -806,7 +825,8 @@ repair_info::repair_info(seastar::sharded<database>& db_,
     , hosts(hosts_)
     , reason(reason_)
     , nr_ranges_total(ranges.size())
-    , _row_level_repair(db.local().features().cluster_supports_row_level_repair()) {
+    , _row_level_repair(db.local().features().cluster_supports_row_level_repair())
+    , _ops_uuid(std::move(ops_uuid)) {
 }
 
 future<> repair_info::do_streaming() {
@@ -1625,7 +1645,7 @@ static int do_repair_start(seastar::sharded<database>& db, seastar::sharded<netw
                 _node_ops_metrics.repair_total_ranges_sum += ranges.size();
                 auto ri = make_lw_shared<repair_info>(db, ms,
                         std::move(keyspace), std::move(ranges), std::move(table_ids),
-                        id, std::move(data_centers), std::move(hosts), streaming::stream_reason::repair);
+                        id, std::move(data_centers), std::move(hosts), streaming::stream_reason::repair, id.uuid);
                 return repair_ranges(ri);
             });
             repair_results.push_back(std::move(f));
@@ -1695,14 +1715,15 @@ static future<> sync_data_using_repair(seastar::sharded<database>& db,
         sstring keyspace,
         dht::token_range_vector ranges,
         std::unordered_map<dht::token_range, repair_neighbors> neighbors,
-        streaming::stream_reason reason) {
+        streaming::stream_reason reason,
+        std::optional<utils::UUID> ops_uuid) {
     if (ranges.empty()) {
         return make_ready_future<>();
     }
-    return smp::submit_to(0, [&db, &ms, keyspace = std::move(keyspace), ranges = std::move(ranges), neighbors = std::move(neighbors), reason] () mutable {
+    return smp::submit_to(0, [&db, &ms, keyspace = std::move(keyspace), ranges = std::move(ranges), neighbors = std::move(neighbors), reason, ops_uuid] () mutable {
         repair_uniq_id id = repair_tracker().next_repair_command();
         rlogger.info("repair id {} to sync data for keyspace={}, status=started", id, keyspace);
-        return repair_tracker().run(id, [id, &db, &ms, keyspace, ranges = std::move(ranges), neighbors = std::move(neighbors), reason] () mutable {
+        return repair_tracker().run(id, [id, &db, &ms, keyspace, ranges = std::move(ranges), neighbors = std::move(neighbors), reason, ops_uuid] () mutable {
             auto cfs = list_column_families(db.local(), keyspace);
             if (cfs.empty()) {
                 rlogger.warn("repair id {} to sync data for keyspace={}, no table in this keyspace", id, keyspace);
@@ -1712,12 +1733,12 @@ static future<> sync_data_using_repair(seastar::sharded<database>& db,
             std::vector<future<>> repair_results;
             repair_results.reserve(smp::count);
             for (auto shard : boost::irange(unsigned(0), smp::count)) {
-                auto f = db.invoke_on(shard, [&db, &ms, keyspace, table_ids, id, ranges, neighbors, reason] (database& localdb) mutable {
+                auto f = db.invoke_on(shard, [&db, &ms, keyspace, table_ids, id, ranges, neighbors, reason, ops_uuid] (database& localdb) mutable {
                     auto data_centers = std::vector<sstring>();
                     auto hosts = std::vector<sstring>();
                     auto ri = make_lw_shared<repair_info>(db, ms,
                             std::move(keyspace), std::move(ranges), std::move(table_ids),
-                            id, std::move(data_centers), std::move(hosts), reason);
+                            id, std::move(data_centers), std::move(hosts), reason, ops_uuid);
                     ri->neighbors = std::move(neighbors);
                     return repair_ranges(ri);
                 });
@@ -1910,16 +1931,16 @@ future<> bootstrap_with_repair(seastar::sharded<database>& db, seastar::sharded<
                 }
             }
             auto nr_ranges = desired_ranges.size();
-            sync_data_using_repair(db, ms, keyspace_name, std::move(desired_ranges), std::move(range_sources), reason).get();
+            sync_data_using_repair(db, ms, keyspace_name, std::move(desired_ranges), std::move(range_sources), reason, {}).get();
             rlogger.info("bootstrap_with_repair: finished with keyspace={}, nr_ranges={}", keyspace_name, nr_ranges);
         }
         rlogger.info("bootstrap_with_repair: finished with keyspaces={}", keyspaces);
     });
 }
 
-static future<> do_decommission_removenode_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr, gms::inet_address leaving_node) {
+static future<> do_decommission_removenode_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr, gms::inet_address leaving_node, shared_ptr<node_ops_info> ops) {
     using inet_address = gms::inet_address;
-    return seastar::async([&db, &ms, tmptr = std::move(tmptr), leaving_node = std::move(leaving_node)] () mutable {
+    return seastar::async([&db, &ms, tmptr = std::move(tmptr), leaving_node = std::move(leaving_node), ops] () mutable {
         auto myip = utils::fb_utilities::get_broadcast_address();
         auto keyspaces = db.local().get_non_system_keyspaces();
         bool is_removenode = myip != leaving_node;
@@ -1978,6 +1999,9 @@ static future<> do_decommission_removenode_with_repair(seastar::sharded<database
             auto local_dc = get_local_dc();
             bool find_node_in_local_dc_only = strat.get_type() == locator::replication_strategy_type::network_topology;
             for (auto&r : ranges) {
+                if (ops) {
+                    ops->check_abort();
+                }
                 auto end_token = r.end() ? r.end()->value() : dht::maximum_token();
                 const std::vector<inet_address> new_eps = ks.get_replication_strategy().calculate_natural_endpoints(end_token, temp, locator::can_yield::yes);
                 const std::vector<inet_address>& current_eps = current_replica_endpoints[r];
@@ -2059,6 +2083,12 @@ static future<> do_decommission_removenode_with_repair(seastar::sharded<database
                 }
                 neighbors_set.erase(myip);
                 neighbors_set.erase(leaving_node);
+                // Remove nodes in ignore_nodes
+                if (ops) {
+                    for (const auto& node : ops->ignore_nodes) {
+                        neighbors_set.erase(node);
+                    }
+                }
                 auto neighbors = boost::copy_range<std::vector<gms::inet_address>>(neighbors_set |
                     boost::adaptors::filtered([&local_dc, &snitch_ptr] (const gms::inet_address& node) {
                         return snitch_ptr->get_datacenter(node) == local_dc;
@@ -2070,9 +2100,10 @@ static future<> do_decommission_removenode_with_repair(seastar::sharded<database
                     rlogger.debug("{}: keyspace={}, range={}, current_replica_endpoints={}, new_replica_endpoints={}, neighbors={}, skipped",
                         op, keyspace_name, r, current_eps, new_eps, neighbors);
                 } else {
-                    rlogger.debug("{}: keyspace={}, range={}, current_replica_endpoints={}, new_replica_endpoints={}, neighbors={}",
-                        op, keyspace_name, r, current_eps, new_eps, neighbors);
-                    range_sources[r] = repair_neighbors(std::move(neighbors));
+                    std::vector<gms::inet_address> mandatory_neighbors = is_removenode ? neighbors : std::vector<gms::inet_address>{};
+                    rlogger.info("{}: keyspace={}, range={}, current_replica_endpoints={}, new_replica_endpoints={}, neighbors={}, mandatory_neighbor={}",
+                            op, keyspace_name, r, current_eps, new_eps, neighbors, mandatory_neighbors);
+                    range_sources[r] = repair_neighbors(std::move(neighbors), std::move(mandatory_neighbors));
                     if (is_removenode) {
                         ranges_for_removenode.push_back(r);
                     }
@@ -2091,7 +2122,7 @@ static future<> do_decommission_removenode_with_repair(seastar::sharded<database
                 ranges.swap(ranges_for_removenode);
             }
             auto nr_ranges_synced = ranges.size();
-            sync_data_using_repair(db, ms, keyspace_name, std::move(ranges), std::move(range_sources), reason).get();
+            sync_data_using_repair(db, ms, keyspace_name, std::move(ranges), std::move(range_sources), reason, ops->ops_uuid).get();
             rlogger.info("{}: finished with keyspace={}, leaving_node={}, nr_ranges={}, nr_ranges_synced={}, nr_ranges_skipped={}",
                 op, keyspace_name, leaving_node, nr_ranges_total, nr_ranges_synced, nr_ranges_skipped);
         }
@@ -2100,11 +2131,17 @@ static future<> do_decommission_removenode_with_repair(seastar::sharded<database
 }
 
 future<> decommission_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr) {
-    return do_decommission_removenode_with_repair(db, ms, std::move(tmptr), utils::fb_utilities::get_broadcast_address());
+    return do_decommission_removenode_with_repair(db, ms, std::move(tmptr), utils::fb_utilities::get_broadcast_address(), {});
 }
 
-future<> removenode_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr, gms::inet_address leaving_node) {
-    return do_decommission_removenode_with_repair(db, ms, std::move(tmptr), std::move(leaving_node));
+future<> removenode_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr, gms::inet_address leaving_node, shared_ptr<node_ops_info> ops) {
+    return do_decommission_removenode_with_repair(db, ms, std::move(tmptr), std::move(leaving_node), std::move(ops));
+}
+
+future<> abort_repair_node_ops(utils::UUID ops_uuid) {
+    return smp::invoke_on_all([ops_uuid] {
+        return repair_tracker().abort_repair_node_ops(ops_uuid);
+    });
 }
 
 static future<> do_rebuild_replace_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr, sstring op, sstring source_dc, streaming::stream_reason reason) {
@@ -2179,7 +2216,7 @@ static future<> do_rebuild_replace_with_repair(seastar::sharded<database>& db, s
                 }).get();
             }
             auto nr_ranges = ranges.size();
-            sync_data_using_repair(db, ms, keyspace_name, std::move(ranges), std::move(range_sources), reason).get();
+            sync_data_using_repair(db, ms, keyspace_name, std::move(ranges), std::move(range_sources), reason, {}).get();
             rlogger.info("{}: finished with keyspace={}, source_dc={}, nr_ranges={}", op, keyspace_name, source_dc, nr_ranges);
         }
         rlogger.info("{}: finished with keyspaces={}, source_dc={}", op, keyspaces, source_dc);
@@ -2218,12 +2255,19 @@ static future<> init_messaging_service_handler(sharded<database>& db, sharded<ne
                 return checksum_range(db, keyspace, cf, range, hv);
             });
         });
+        ms.register_node_ops_cmd([] (const rpc::client_info& cinfo, node_ops_cmd_request req) {
+            auto src_cpu_id = cinfo.retrieve_auxiliary<uint32_t>("src_cpu_id");
+            auto coordinator = cinfo.retrieve_auxiliary<gms::inet_address>("baddr");
+            return smp::submit_to(src_cpu_id % smp::count, [coordinator, req = std::move(req)] () mutable {
+                return service::get_local_storage_service().node_ops_cmd_handler(coordinator, std::move(req));
+            });
+        });
     });
 }
 
 static future<> uninit_messaging_service_handler() {
     return _messaging->invoke_on_all([] (auto& ms) {
-        return ms.unregister_repair_checksum_range();
+        return when_all_succeed(ms.unregister_repair_checksum_range(), ms.unregister_node_ops_cmd()).discard_result();
     });
 }
 
--- a/repair/repair.hh
+++ b/repair/repair.hh
@@ -76,12 +76,21 @@ struct repair_uniq_id {
 };
 std::ostream& operator<<(std::ostream& os, const repair_uniq_id& x);
 
+struct node_ops_info {
+    utils::UUID ops_uuid;
+    bool abort = false;
+    std::list<gms::inet_address> ignore_nodes;
+    void check_abort();
+};
+
 // The tokens are the tokens assigned to the bootstrap node.
 future<> bootstrap_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr, std::unordered_set<dht::token> bootstrap_tokens);
 future<> decommission_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr);
-future<> removenode_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr, gms::inet_address leaving_node);
+future<> removenode_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr, gms::inet_address leaving_node, shared_ptr<node_ops_info> ops);
 future<> rebuild_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr, sstring source_dc);
 future<> replace_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr, std::unordered_set<dht::token> replacing_tokens);
+
+future<> abort_repair_node_ops(utils::UUID ops_uuid);
 
 // NOTE: repair_start() can be run on any node, but starts a node-global
 // operation.
@@ -244,6 +253,7 @@ public:
     bool _row_level_repair;
     uint64_t _sub_ranges_nr = 0;
     std::unordered_set<sstring> dropped_tables;
+    std::optional<utils::UUID> _ops_uuid;
 public:
     repair_info(seastar::sharded<database>& db_,
             seastar::sharded<netw::messaging_service>& ms_,
@@ -253,7 +263,8 @@ public:
             repair_uniq_id id_,
             const std::vector<sstring>& data_centers_,
             const std::vector<sstring>& hosts_,
-            streaming::stream_reason reason_);
+            streaming::stream_reason reason_,
+            std::optional<utils::UUID> ops_uuid);
     future<> do_streaming();
     void check_failed_ranges();
     future<> request_transfer_ranges(const sstring& cf,
@@ -272,6 +283,9 @@ public:
     const std::vector<sstring>& table_names() {
         return cfs;
     }
+    const std::optional<utils::UUID>& ops_uuid() const {
+        return _ops_uuid;
+    };
 };
 
 // The repair_tracker tracks ongoing repair operations and their progress.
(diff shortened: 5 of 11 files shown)


api/api-doc/storage_service.json
```diff
@@ -1105,6 +1105,14 @@
                      "allowMultiple":false,
                      "type":"string",
                      "paramType":"query"
+                  },
+                  {
+                     "name":"ignore_nodes",
+                     "description":"List of dead nodes to ingore in removenode operation",
+                     "required":false,
+                     "allowMultiple":false,
+                     "type":"string",
```
avikivity, line 1114: Shouldn't the type be array?

api/storage_service.cc
```diff
@@ -495,7 +496,22 @@ void set_storage_service(http_context& ctx, routes& r) {
 
     ss::remove_node.set(r, [](std::unique_ptr<request> req) {
         auto host_id = req->get_query_param("host_id");
-        return service::get_local_storage_service().removenode(host_id).then([] {
+        std::vector<sstring> ignore_nodes = split(req->get_query_param("ignore_nodes"), ",");
```
avikivity, line 499: inet_address?

api/api-doc/storage_service.json
avikivity: @amnonh please advise

idl/partition_checksum.idl.hh
```diff
@@ -103,3 +103,22 @@ enum class repair_row_level_start_status: uint8_t {
 struct repair_row_level_start_response {
     repair_row_level_start_status status;
 };
+
+enum class removenode_cmd : uint32_t {
+     prepare,
+     sync_data,
+     abort,
+     error,
+     ok,
```
avikivity, line 112: What is an error command or an ok command?

idl/partition_checksum.idl.hh
```diff
@@ -103,3 +103,22 @@ enum class repair_row_level_start_status: uint8_t {
...
+struct removenode_request {
+    removenode_cmd cmd;
+    utils::UUID req_id;
+    std::list<gms::inet_address> leaving_nodes;
+    std::list<gms::inet_address> skip_sync_on_nodes;
+};
+
+struct removenode_response {
+    removenode_cmd cmd;
+};
```
avikivity, line 124: Ah, I see you use it as a response. Better to have its own enum for that.

service/storage_service.cc
```diff
@@ -2257,6 +2257,143 @@ future<> storage_service::removenode(sstring host_id_string) {
...
+                        slogger.warn("removenode[{}]: Node {} does not support removenode verb", uuid, node);
+                        nodes_unknown_verb.emplace(node);
+                    }).handle_exception_type([&nodes_down, node, uuid] (seastar::rpc::closed_error&) {
+                        slogger.warn("removenode[{}]: Node {} is down for removenode_cmd verb", uuid, node);
+                        nodes_down.emplace(node);
+                    });
+                }).get();
+                if (!nodes_unknown_verb.empty()) {
+                    auto msg = format("removenode[{}]: Nodes={} do not support removenode verb. Please upgrade your cluster and run removenode again.", uuid, nodes_unknown_verb);
+                    slogger.warn("{}", msg);
```
avikivity, line 2313: Better to have a cluster feature for it.

service/storage_service.cc
```diff
@@ -2257,6 +2257,143 @@ future<> storage_service::removenode(sstring host_id_string) {
...
+            }
+            if (req.cmd == removenode_cmd::prepare) {
+                auto tmlock = std::make_unique<token_metadata_lock>(ss.get_token_metadata_lock().get0());
+                auto tmptr = ss.get_mutable_token_metadata_ptr().get0();
+                for (auto& node : req.leaving_nodes) {
+                    slogger.info("removenode[{}]: Added node={} as leaving node, coordinator={}", req.req_id, node, coordinator);
+                    tmptr->add_leaving_endpoint(node);
+                }
+                ss.update_pending_ranges(tmptr, format("removenode {}", req.leaving_nodes)).get();
+                ss.replicate_to_all_cores(std::move(tmptr)).get();
```
avikivity, line 2374: better to use mutate_token_metadata(), less things that can go wrong.

service/storage_service.cc
```diff
@@ -2257,6 +2257,143 @@ future<> storage_service::removenode(sstring host_id_string) {
...
+                }
+            } else if (req.cmd == removenode_cmd::abort) {
+                auto tmlock = std::make_unique<token_metadata_lock>(ss.get_token_metadata_lock().get0());
+                auto tmptr = ss.get_mutable_token_metadata_ptr().get0();
+                for (auto& node : req.leaving_nodes) {
+                    slogger.info("removenode[{}]: Removed node={} as leaving node, coordinator={}", req.req_id, node, coordinator);
+                    tmptr->del_leaving_endpoint(node);
+                }
+                ss.update_pending_ranges(tmptr, format("removenode {}", req.leaving_nodes)).get();
+                ss.replicate_to_all_cores(std::move(tmptr)).get();
```
avikivity, line 2388: Here too.

api/storage_service.cc
asias: It is converted to inet_address below. We can only get string from the query parameter.

idl/partition_checksum.idl.hh
asias: It is used by the response.

idl/partition_checksum.idl.hh
asias: I prefer a single cmd for both request and response to make the protocol simple. We have similar usage in row level repair.

service/storage_service.cc
asias: ok. I will switch to the new helper.

asias:
Changes in v2 e04a66a0425b253c75949fbb20b447f0fa5ff9d0:
- Use mutate_token_metadata
- Drop the unused removenode_cmd::error
- Rename skip_sync_on_nodes to ignore_nodes
- Rename req_id to ops_uuid in removenode_request

service/storage_service.cc
asias: I like to avoid using one if it is possible. The features are crowded.

api/storage_service.cc
avikivity: Well, at least we can start with an array.

idl/partition_checksum.idl.hh
avikivity: It is less to type, but confusing to readers. The request and response are not the same thing.

service/storage_service.cc
avikivity: I don't see a problem with crowding, and I'd like to have one method of determining whether to enable a feature during upgrades, not two.

asias:
On Wed, Nov 18, 2020 at 4:26 PM Avi Kivity <notifications@github.com> wrote:

>
> In service/storage_service.cc
> <https://github.com/scylladb/scylla/pull/7626#discussion_r525895834>:
>
> > +            auto req = removenode_request{removenode_cmd::prepare, uuid, leaving_nodes, skip_sync_on_nodes};
> +            try {
> +                parallel_for_each(nodes, [&ss, &req, &nodes_unknown_verb, &nodes_down, uuid] (const gms::inet_address& node) {
> +                    return ss._messaging.local().send_removenode(netw::msg_addr(node), req).then([uuid, node] (removenode_response resp) {
> +                        slogger.info("removenode[{}]: Got prepare response from node={}", uuid, node);
> +                    }).handle_exception_type([&nodes_unknown_verb, node, uuid] (seastar::rpc::unknown_verb_error&) {
> +                        slogger.warn("removenode[{}]: Node {} does not support removenode verb", uuid, node);
> +                        nodes_unknown_verb.emplace(node);
> +                    }).handle_exception_type([&nodes_down, node, uuid] (seastar::rpc::closed_error&) {
> +                        slogger.warn("removenode[{}]: Node {} is down for removenode_cmd verb", uuid, node);
> +                        nodes_down.emplace(node);
> +                    });
> +                }).get();
> +                if (!nodes_unknown_verb.empty()) {
> +                    auto msg = format("removenode[{}]: Nodes={} do not support removenode verb. Please upgrade your cluster and run removenode again.", uuid, nodes_unknown_verb);
> +                    slogger.warn("{}", msg);
>
> I don't see a problem with crowding, and I'd like to have one method of
> determining whether to enable a feature during upgrades, not two.
>
How do you not enable a feature during upgrades? What do you want to
achieve with features here exactly?

-- 
Asias

api/api-doc/storage_service.json
amnonh: @avikivity It's a query parameter, we don't support arrays in query parameters

avikivity:
(please reply on github so I don't have to fix the formatting)

>> I don't see a problem with crowding, and I'd like to have one method of determining whether to enable a feature during upgrades, not two.

> How do you not enable a feature during upgrades?

The feature is still enabled, but it uses the old mechanism.

> What do you want to achieve with features here exactly?

Orthogonality between upgrades and other operations. You don't have special limitations during upgrade, everything continues to work. It's not a special state.

You remember all the limitations and dependencies, but operators (human and machine) don't. A human operator may see the message and recover, but a machine operator won't. And in any case it will take some back-and-forth to understand the problem.

asias:
> (please reply on github so I don't have to fix the formatting)

OK.

> > > I don't see a problem with crowding, and I'd like to have one method of determining whether to enable a feature during upgrades, not two.
> 
> > How do you not enable a feature during upgrades?
> 
> The feature is still enabled, but it uses the old mechanism.

Feature will not help in terms of using old mechanism while the feature is enabled. 

> 
> > What do you want to achieve with features here exactly?
> 
> Orthogonality between upgrades and other operations. You don't have special limitations during upgrade, everything continues to work. It's not a special state.

During upgrade, operations like add / remove node is not supported. 

> 
> You remember all the limitations and dependencies, but operators (human and machine) don't. A human operator may see the message and recover, but a machine operator won't. And in any case it will take some back-and-forth to understand the problem.

There is no need to remember anything. The message says clearly upgrade then run the command. The machine operator is doing wrong if it removes a node during upgrade. With this patch, the operation error is prevented. Also, it makes no sense to use the old unsafe removenode if the upgrade is happening. 

Another point using cluster feature is bad:

- old version
- all nodes upgrade to new version, feature is enabled
- user finds the new method does not work well but there is no way to downgrade to old version

Without a feature, user can downgrade nicely.

avikivity:
> > (please reply on github so I don't have to fix the formatting)
> 
> OK.
> 
> > > > I don't see a problem with crowding, and I'd like to have one method of determining whether to enable a feature during upgrades, not two.
> > 
> > 
> > > How do you not enable a feature during upgrades?
> > 
> > 
> > The feature is still enabled, but it uses the old mechanism.
> 
> Feature will not help in terms of using old mechanism while the feature is enabled.

Can you explain? Is the new mechanism not a superset?

> 
> > > What do you want to achieve with features here exactly?
> > 
> > 
> > Orthogonality between upgrades and other operations. You don't have special limitations during upgrade, everything continues to work. It's not a special state.
> 
> During upgrade, operations like add / remove node is not supported.

That's the theory, but I don't believe in it. What if you have a disaster during upgrade and must remove a node? What if upgrade cannot proceed, and you need to scale up? do you just wait until the upgrade problems are resolved?

These operations must be orthogonal to each other, especially when automation is responsible for issuing these commands.

> 
> > You remember all the limitations and dependencies, but operators (human and machine) don't. A human operator may see the message and recover, but a machine operator won't. And in any case it will take some back-and-forth to understand the problem.
> 
> There is no need to remember anything. The message says clearly upgrade then run the command. The machine operator is doing wrong if it removes a node during upgrade. With this patch, the operation error is prevented. Also, it makes no sense to use the old unsafe removenode if the upgrade is happening.

No. Issuing a command and then having to dig through logs to find why it failed is bad experience. Things should just work if we can make it so, not fail due to unrelated reasons.

> 
> Another point using cluster feature is bad:
> 
> * old version
> * all nodes upgrade to new version, feature is enabled
> * user finds the new method does not work well but there is no way to downgrade to old version
> 
> Without a feature, user can downgrade nicely.

No, new versions can change the storage format.

We can add an extra step during upgrade, between completion of the upgrade and actually using the new features, to let the operator decide whether they want to commit to the upgrade or not (at which point we will start using the new features).

Without features, the cluster starts using the new features one node at a time, leading to strange failures.

asias:
> No. Issuing a command and then having to dig through logs to find why it failed is bad experience. Things should just work if we can make it so, not fail due to unrelated reasons.

No. The reason why it fails is returned back to the command. No need to dig logs.

asias:
> Can you explain? Is the new mechanism not a superset?

No. This patch drops the old removenode operation because it is too dangerous and fragile to use.

asias:
> That's the theory, but I don't believe in it. What if you have a disaster during upgrade and must remove a node? What if upgrade cannot proceed, and you need to scale up? do you just wait until the upgrade problems are resolved?

In what case, one MUST remove a node during upgrade?

If there is problem upgrading, downgrading and continue.

> 
> These operations must be orthogonal to each other, especially when automation is responsible for issuing these commands.

Operations can not be orthogonal to each other. We can not add or remove a node while another node is joining or leaving.

asias:
> No, new versions can change the storage format.

Why storage format is relevant here? We are talking about the feature to allow use the new removenode operation. If we do not introduce a feature for that. There is no such feature. User can downgrade to a previous version after all nodes have used the new version. 

> We can add an extra step during upgrade, between completion of the upgrade and actually using the new features, to let the operator decide whether they want to commit to the upgrade or not (at which point we will start using the new features).

If you want admin to decide use the old or new method, isn't an option to enable is more robust and easier. 
> 
> Without features, the cluster starts using the new features one node at a time, leading to strange failures.

I do not see why the failure is strange to you.

```
removenode[{}]: Nodes={} do not support removenode verb. Please upgrade your cluster and run removenode again.
```

asias:
@avikivity If your intention is to fallback to the old removenode method in case the cluster does not support it instead of returning an error. 

We still do not need to a gossip feature for that.  

if (the new verb is not supported) {
   fall back
}

avikivity:
> > No, new versions can change the storage format.
> 
> Why storage format is relevant here? We are talking about the feature to allow use the new removenode operation. If we do not introduce a feature for that. There is no such feature. User can downgrade to a previous version after all nodes have used the new version.


You made a comment about a disadvantage of features in general, that they prevent downgrades. I explained that features sometimes correctly block downgrades.

> 
> > We can add an extra step during upgrade, between completion of the upgrade and actually using the new features, to let the operator decide whether they want to commit to the upgrade or not (at which point we will start using the new features).
> 
> If you want admin to decide use the old or new method, isn't an option to enable is more robust and easier.
> 

Features are there to provide a general mechanism for upgrades. I don't want each individual caller to use its own method to handle mismatched versions.

You explained that features prevent downgrades, and I explained how we can enhance features to support downgrades after the cluster is fully upgraded (at the cost of not using the new capaibilities immediately).


> > Without features, the cluster starts using the new features one node at a time, leading to strange failures.
> 
> I do not see why the failure is strange to you.
> 
> ```
> removenode[{}]: Nodes={} do not support removenode verb. Please upgrade your cluster and run removenode again.
> ```

That is not bad. But still, why not use our standard method of coordinating features?

btw, you can get this message even without upgrades, if you hit a restarting node before it registers the verb.

avikivity:
> > That's the theory, but I don't believe in it. What if you have a disaster during upgrade and must remove a node? What if upgrade cannot proceed, and you need to scale up? do you just wait until the upgrade problems are resolved?
> 
> In what case, one MUST remove a node during upgrade?
> 

I don't know. I do know that the more constraints there are, the easier it is to get trapped into a scenario with no way out.

> If there is problem upgrading, downgrading and continue.
> 
> > These operations must be orthogonal to each other, especially when automation is responsible for issuing these commands.
> 
> Operations can not be orthogonal to each other. We can not add or remove a node while another node is joining or leaving.

We will work to remove constraints, maybe it will take more than half an hour.

asias:
> You made a comment about a disadvantage of features in general, that they prevent downgrades. I explained that features sometimes correctly block downgrades.

I am only commenting that removenode feature is unnecessary blocking of downgrades. I agree blocking downgrade is useful in some cases that why I added the gossip feature support in the first place.

asias:
> That is not bad. But still, why not use our standard method of coordinating features?
> 
> btw, you can get this message even without upgrades, if you hit a restarting node before it registers the verb.

Still, it is correct to fail the operation because one node is not ready. We can probably get "node is down" error too.

asias:
> I don't know. I do know that the more constraints there are, the easier it is to get trapped into a scenario with no way out.

Feature does not help here. In this case, it adds unnecessary downgrade block which may get user trapped too.  If we want user to be able to select old or new method, use config option or restful api. IMHO, depending on gossip features is wrong to do the job.

avikivity:
Well, I don't know what I can say without repeating myself. I'd like to have one method of checking for things, so that if we ever improve it, everything gains.

asias:
> Well, I don't know what I can say without repeating myself. I'd like to have one method of checking for things, so that if we ever improve it, everything gains.

Avi, nothing will be improved by introducing a gossip feature to use the new removenode operation method. 

Assume we have a SAFE_REMOVENODE feature. We will  do:

if (SAFE_REMOVENODE enabled) {
   use new removenode method
} else {
   reject the operation and tell user to upgrade the cluster and retry
}

We can do this without the fetaure by detecting if a node supports the new method. 

if (all nodes involved support the new removenode verb) {
   use new removenode method
} else {
   reject the operation and tell user to upgrade the cluster and retry
}

Assume two nodes are dead. We want to removenode the dead nodes. If we require a feature, we can not use the new method to remove the node after upgrade all remaining live node to the new version, because the feature won't be enabled since not all nodes enable it (the dead node can not enable it).

With this patch, it can work nicely.

avikivity:
I see, so indeed the smooth transition aspect is not gained by a cluster feature, since there is no smooth transition.

There remains the "one way to detect changes" rather than ad-hoc with every feature deciding itself on how to detect changes in how the cluster should behave.

asias:
> I see, so indeed the smooth transition aspect is not gained by a cluster feature, since there is no smooth transition.
> 
> There remains the "one way to detect changes" rather than ad-hoc with every feature deciding itself on how to detect changes in how the cluster should behave.

Using the new removenode feature or not is not cluster wide decision.

Let me repeat my example above:

Assume two nodes are dead. We want to removenode the dead nodes. If we require a feature, we can not use the new method to remove the node after upgrade all remaining live node to the new version, because the feature won't be enabled since not all nodes enable it (the dead node can not enable it).

We do not have a lot of ad-hoc methods to detect features, we currently only have two 1) Gossip features 2) RPC verb detection.
We need to use 1) if we can only use the features if all nodes in the cluster must agree to use it. 

BTW, I am planing to extend this verb to be more generic to use it for replace operation too. Replacing is during boot up process, it may not get a stable gossip propagation yet. 

Do you have other comments other than "feature detection"?

avikivity:
I don't think we use RPC verb detection now. It has a downside that users see RPC errors on the server side and report them.

removenode is a rare operation, and must be used in mid-upgrade in order to cause any problems, making the error very rare.

api/api-doc/storage_service.json
avikivity: What about allowMultiple?

repair/repair.cc
```diff
@@ -2090,20 +2121,45 @@ static future<> do_decommission_removenode_with_repair(seastar::sharded<database
...
             rlogger.info("{}: finished with keyspace={}, leaving_node={}, nr_ranges={}, nr_ranges_synced={}, nr_ranges_skipped={}",
                 op, keyspace_name, leaving_node, nr_ranges_total, nr_ranges_synced, nr_ranges_skipped);
         }
         rlogger.info("{}: finished with keyspaces={}, leaving_node={}", op, keyspaces, leaving_node);
     });
 }
 
+
+static thread_local std::unordered_map<utils::UUID, lw_shared_ptr<node_ops>> _node_ops;
+
```
avikivity, line 2134: Is there no better place for this? It's better to put things inside sharded<> instances rather than globals that then have to be disentangled by @xemul .

idl/partition_checksum.idl.hh
avikivity: I still think sending a command as a response is confusing.

repair/repair.cc
```diff
@@ -2058,6 +2082,12 @@ static future<> do_decommission_removenode_with_repair(seastar::sharded<database
                 }
                 neighbors_set.erase(myip);
                 neighbors_set.erase(leaving_node);
+                // Remove nodes in ignore_nodes
+                if (ops) {
+                    for (const auto& node : ops->ignore_nodes) {
+                        neighbors_set.erase(node);
+                    }
+                }
```
avikivity, line 2090: Maybe we should abort the repair. Aren't our quorum calculations invalidated by the node removal?

What happens if the removal happens after we pass this point?

avikivity: Ah, I see this is for the removenode itself.

Maybe check the parameters for consistency? If doing a decommission, ignore_nodes must be empty, and vice versa.

service/storage_service.cc
```diff
@@ -2158,102 +2158,134 @@ future<> storage_service::decommission() {
...
+                    slogger.warn("{}", msg);
+                    throw std::runtime_error(msg);
+                }
 
-            // A leaving endpoint that is dead is already being removed.
-            if (tmptr->is_leaving(endpoint)) {
-                slogger.warn("Node {} is already being removed, continuing removal anyway", endpoint);
-            }
+                // Step 3: Start to sync data
+                req.cmd = removenode_cmd::sync_data;
```
avikivity, line 2221: What happens if a node that has prepared restarts? What happens if the node that initiates removenode restarts? Does the prepared node stay prepared forever?

service/storage_service.cc
```diff
@@ -2158,102 +2158,134 @@ future<> storage_service::decommission() {
...
 
-            if (!ss._replicating_nodes.empty()) {
-                throw std::runtime_error("This node is already processing a removal. Wait for it to complete, or use 'removenode force' if this has failed.");
+                // Step 4: Announce the node has left
+                std::unordered_set<token> tmp(tokens.begin(), tokens.end());
+                ss.excise(std::move(tmp), endpoint);
+                ss._gossiper.advertise_token_removed(endpoint, host_id).get();
+                slogger.info("removenode[{}]: Finished removenode operation, removing node={}, sync_nodes={}, ingore_nodes={}", uuid, endpoint, nodes, ignore_nodes);
+            } catch (...) {
+                // we need to revert the effect of prepare verb the removenode ops is failed
```
avikivity, line 2235: What if we die before we reach here?

asias: We need an api to revert the effect. This  node can do nothing if it dies.

service/storage_service.cc
asias: Yes, it stays forever until it is cancelled. The node initiates removenode is stateless.

avikivity: How can it be cancelled?

We need a better solution. Each thing creates a new set of decisions to be taken by the operator. This is not documented anywhere, and even if it is, we have to teach scylla-operator about it.

asias:
> I don't think we use RPC verb detection now. It has a downside that users see RPC errors on the server side and report them.

We do. For example, we use it to decide if the new shadow round verb method should be used. 

> removenode is a rare operation, and must be used in mid-upgrade in order to cause any problems, making the error very rare.

To quote your comments somewhere else, it is theory, we'd better not to add unnecessary constrains. 

The example above is a valid use case in my view. Some nodes are down, user wants to remove it in a safe way. So they upgrade the cluster to use new version to get the new removenode feature. If we use gossip feature, it is not possible. What gossip feature gives us here, nothing. Instead it imposes an unnecessary constrain.

service/storage_service.cc
avikivity: Well, with Raft the plan is to commit the topology change intent somewhere, and so, as long as the cluster is alive, someone else takes over and pushes the change forwards.

I don't know what we can do here without making it even more complicated, but leaving it open like that is also bad.

service/storage_service.cc
asias: For example, 

nodetool removenode abort $ops_uuid.

I am thinking about revert-automatically-if-removenode-coordinator-does-not-ping-in-some-time.

service/storage_service.cc
asias: Let's not depend it on raft here. See my idea here: https://github.com/scylladb/scylla/pull/7626#discussion_r532496787

service/storage_service.cc
avikivity: This has to be run on all nodes? Or just one node and it propagates it to the cluster?

Using a timeout is possible, but also adds complications.

service/storage_service.cc
avikivity: Of course, just citing it as a strategy that isn't prone to cascading complexity.

service/storage_service.cc
asias: Either admin or the nodes that have responded removenode_cmd::prepare have to revert it.

All nodes that have marked the node to be removed in the pending endpoints have to revert it. So the idea is that once a node has responded removenode_cmd::prepare, it starts a watchdog timer. The removenode coordinator has to feed the watdchdog otherwise the effect of removenode_cmd::prepare is reverted. 

The watchdog logic does add extra complexity. We need to store a handler on the nodes (in current patch, we do not need to, it is almost stateless). The reverted pending endpoints will cause the whole cluster to have extra writes. It is best we can revert it automatically. 

Please note, this is not a regression. Current situation is 10x worse.

repair/repair.cc
asias: We have:
```
2135 future<> decommission_with_repair(seastar::sharded<database>& db, seastar::sharded<netw::messaging_service>& ms, locator::token_metadata_ptr tmptr) {
2136     return do_decommission_removenode_with_repair(db, ms, std::move(tmptr), utils::fb_utilities::get_broadcast_address(), {});
2137 }
```

It is guaranteed the node_ops is empty, thus node_ops->ignore_nodes is empty, for decommission.

repair/repair.cc
asias: I can put it inside the repair_tracker.

asias: Removed the global in new version.

asias:
Changes in v2 1ccb4c9ca459ba6a289f13e4d65c882172224b07:

- Make the  removenode verb generic  so we can use it for replace later
- Add auto revert support based on hearbeat update from the coordinator
- Get rid of the global _node_ops object

asias:
@avikivity FYI, I managed to use this new verb to improve replace operation too, so that replace ops can auto revert if the replacing node is gone. Looks promising. I will send patch for repalce operation after this one is merged.

db/config.cc
```diff
@@ -718,7 +718,7 @@ db::config::config(std::shared_ptr<db::extensions> exts)
     , replace_address(this, "replace_address", value_status::Used, "", "The listen_address or broadcast_address of the dead node to replace. Same as -Dcassandra.replace_address.")
     , replace_address_first_boot(this, "replace_address_first_boot", value_status::Used, "", "Like replace_address option, but if the node has been bootstrapped successfully it will be ignored. Same as -Dcassandra.replace_address_first_boot.")
     , override_decommission(this, "override_decommission", value_status::Used, false, "Set true to force a decommissioned node to join the cluster")
-    , enable_repair_based_node_ops(this, "enable_repair_based_node_ops", liveness::LiveUpdate, value_status::Used, false, "Set true to use enable repair based node operations instead of streaming based")
+    , enable_repair_based_node_ops(this, "enable_repair_based_node_ops", liveness::LiveUpdate, value_status::Used, true, "Set true to use enable repair based node operations instead of streaming based")
     , ring_delay_ms(this, "ring_delay_ms", value_status::Used, 30 * 1000, "Time a node waits to hear from other nodes before joining the ring in milliseconds. Same as -Dcassandra.ring_delay_ms in cassandra.")
```
avikivity, line 722: I expect this is accidental? It should only be re-enabled when we are sure it is safe, and the changelog should explain why it is safe (test details).

api/api-doc/storage_service.json
avikivity: @amnonh ping here

idl/partition_checksum.idl.hh
```diff
@@ -103,3 +103,22 @@ enum class repair_row_level_start_status: uint8_t {
...
+     removenode_sync_data,
+     removenode_abort,
+     removenode_done,
+};
+
+struct node_ops_cmd_request {
+    node_ops_cmd cmd;
+    utils::UUID ops_uuid;
+    std::list<gms::inet_address> leaving_nodes;
+    std::list<gms::inet_address> ignore_nodes;
```
avikivity, line 119: Are those members applicable for all commands?

At a minimum, there should be documentation about when to expect which members.

repair/repair.cc
```diff
@@ -1695,14 +1715,15 @@ static future<> sync_data_using_repair(seastar::sharded<database>& db,
...
+        std::optional<utils::UUID> ops_uuid) {
     if (ranges.empty()) {
         return make_ready_future<>();
     }
-    return smp::submit_to(0, [&db, &ms, keyspace = std::move(keyspace), ranges = std::move(ranges), neighbors = std::move(neighbors), reason] () mutable {
+    return smp::submit_to(0, [&db, &ms, keyspace = std::move(keyspace), ranges = std::move(ranges), neighbors = std::move(neighbors), reason, ops_uuid] () mutable {
         repair_uniq_id id = repair_tracker().next_repair_command();
         rlogger.info("repair id {} to sync data for keyspace={}, status=started", id, keyspace);
-        return repair_tracker().run(id, [id, &db, &ms, keyspace, ranges = std::move(ranges), neighbors = std::move(neighbors), reason] () mutable {
+        return repair_tracker().run(id, [id, &db, &ms, keyspace, ranges = std::move(ranges), neighbors = std::move(neighbors), reason, ops_uuid] () mutable {
```
avikivity, line 1726: You hit the new global in an existing global. That is fine, but let's get rid of the existing global too, later.

avikivity: (it can be a member of storge_service, or perhaps created by main)

repair/repair.cc
```diff
@@ -2218,12 +2255,19 @@ static future<> init_messaging_service_handler(sharded<database>& db, sharded<ne
                 return checksum_range(db, keyspace, cf, range, hv);
             });
         });
+        ms.register_node_ops_cmd([] (const rpc::client_info& cinfo, node_ops_cmd_request req) {
```
avikivity, line 2258: and here, we'd capture the tracker so we don't need to pick it up from a global (again, later).

service/storage_service.cc
```diff
@@ -2158,102 +2158,174 @@ future<> storage_service::decommission() {
...
+                }
 
-            // A leaving endpoint that is dead is already being removed.
-            if (tmptr->is_leaving(endpoint)) {
-                slogger.warn("Node {} is already being removed, continuing removal anyway", endpoint);
-            }
+                // Step 3: Start heartbeat updater
+                seastar::timer<std::chrono::steady_clock> heartbeat_timer([&ss, &nodes, uuid] {
+                    auto req = node_ops_cmd_request{node_ops_cmd::removenode_heartbeat, uuid, {}, {}};
+                    parallel_for_each(nodes, [&ss, &req, uuid] (const gms::inet_address& node) {
```
avikivity, line 2223: Is this not a discarded future? If so it can be a use-after-free:

1. timer fires, continuation chain started
2. removenode completes, deallocates state
3. continuation chain continues, hits deallocated state.

The only relevant state I see is `req`. It's better for timers just to notify existing loops (like in repeat()), not to launch continuation chains that no one waits for.

avikivity: btw, this should not have compiled, we have [[nodiscard]] on future<>.

service/storage_service.cc
```diff
@@ -3209,5 +3283,68 @@ bool storage_service::is_repair_based_node_ops_enabled() {
...
+}
+
+void node_ops_meta_data::cancel_watchdog() {
+    _watchdog.cancel();
+}
+
+void node_ops_meta_data::exec_watchdog() {
+    abort().handle_exception([this] (std::exception_ptr ep) {
+        slogger.warn("Failed to run node ops abort for coordinator={}: {}", _coordinator, ep);
+    });
```
avikivity, line 3314: This can also cause a use after free, if this is deallocated before handle_exception gets a chance to execute.

avikivity:
Looks much better. It replaces the "get into strange state on crash" problem with "may not make forward progress due to timeouts" problem, but this is expected in a distributed system.

db/config.cc
asias: Yes, it is accidental. It belongs to a local debug commit. Sorry I will remove it.

idl/partition_checksum.idl.hh
asias: cmd and ops_uuid are mandatory.  leaving_nodes and ignore_nodes are optionally. I will add some docs.

repair/repair.cc
asias: It is used by repair.cc too. If we put it inside storge_service, repair.cc would require the global. It is better to put inside main.cc. 

Yes, let's do it later.

repair/repair.cc
asias: ok

service/storage_service.cc
asias: I think our time api is pretty problematic when a continuation is needed in the callback. Can we make the timer object waits for the callbacks to finish so that we can wait for the timer to finish the callbacks if any pending. 

Before we destroy a timer:

timer.cancel()
timer.wait_for_pending_callbacks();

asias:
Changes in v3 db8371c2be2cc61c40f4a1ef153dbc7f959d2722:
- Add comments for node_ops_cmd_request
- Fix user after free for timer usage

service/storage_service.cc
avikivity: It will really be a new API (perhaps on top of existing timers), since the callback now has to return a future to indicate when it is done.

idl/partition_checksum.idl.hh
```diff
@@ -103,3 +103,22 @@ enum class repair_row_level_start_status: uint8_t {
...
+     removenode_abort,
+     removenode_done,
+};
+
+struct node_ops_cmd_request {
+    node_ops_cmd cmd;
+    utils::UUID ops_uuid;
+    std::list<gms::inet_address> ignore_nodes;
+    std::list<gms::inet_address> leaving_nodes;
+};
```
avikivity, line 120: You promised some docs for this.

repair/repair.hh
```diff
@@ -464,6 +479,27 @@ enum class row_level_diff_detect_algorithm : uint8_t {
...
+enum class node_ops_cmd : uint32_t {
+     removenode_prepare,
+     removenode_heartbeat,
+     removenode_sync_data,
+     removenode_abort,
+     removenode_done,
+};
+
+// The cmd and ops_uuid are mandatory for each request.
+// The ignore_nodes and leaving_node are optional.
```
avikivity, line 491: Ah I see you added the docs here.

asias: Yes.

idl/partition_checksum.idl.hh
asias: This is idll. The doc is is in the .hh.

service/storage_service.cc
```diff
@@ -2158,102 +2162,191 @@ future<> storage_service::decommission() {
...
-                throw std::runtime_error("Cannot remove self");
-            }
+                // Step 3: Start heartbeat updater
+                heartbeat_updater = seastar::async([&ss, &nodes, uuid, heartbeat_updater_done] {
+                    slogger.info("removenode[{}]: Started heartbeat_updater", uuid);
+                    while (!(*heartbeat_updater_done)) {
+                        auto req = node_ops_cmd_request{node_ops_cmd::removenode_heartbeat, uuid, {}, {}};
+                        parallel_for_each(nodes, [&ss, &req, uuid] (const gms::inet_address& node) {
+                            return ss._messaging.local().send_node_ops_cmd(netw::msg_addr(node), req).then([uuid, node] (node_ops_cmd_response resp) {
+                                slogger.info("removenode[{}]: Got heartbeat response from node={}", uuid, node);
```
avikivity, line 2234: This is too noisy, no?

Shouldn't you send the heartbeat with a timeout? If one node doesn't respond, it will delay the heartbeat for the other nodes (maybe you don't care about that).

service/storage_service.cc
```diff
@@ -2158,102 +2162,191 @@ future<> storage_service::decommission() {
...
+                                return make_ready_future<>();
+                            });
+                        }).handle_exception([uuid] (std::exception_ptr ep) {
+                            slogger.warn("removenode[{}]: Failed to send heartbeat", uuid);
+                        }).get();
+                        int nr_seconds = 10;
+                        while (!(*heartbeat_updater_done) && nr_seconds--) {
+                            sleep_abortable(std::chrono::seconds(1), ss._abort_source).get();
```
avikivity, line 2242: Why not sleep for nr_seconds?

service/storage_service.cc
```diff
@@ -2158,102 +2162,191 @@ future<> storage_service::decommission() {
...
+                        slogger.info("removenode[{}]: Got done response from node={}", uuid, node);
+                        return make_ready_future<>();
+                    });
+                }).get();
+                *heartbeat_updater_done = true;
+                heartbeat_updater.get();
+                slogger.info("removenode[{}]: Finished removenode operation, removing node={}, sync_nodes={}, ingore_nodes={}", uuid, endpoint, nodes, ignore_nodes);
+            } catch (...) {
+                *heartbeat_updater_done = true;
+                heartbeat_updater.get();
```
avikivity, line 2276: A safer way to do this is to put these two lines in a defer() block as soon as you create the updater.

service/storage_service.cc
asias: It should be debug. I need to change it back.

Yes, I do not care.  One bad node is enough to fail the whole operation.

service/storage_service.cc
asias: When the while operation is done, we do not need to wait for nr_seconds to wait for the update to wake up.

service/storage_service.cc
asias: but defer does not wait for the fiber in the callback to finish

service/storage_service.cc
avikivity: Yes, but it's abortable. I see you don't use the abort_source for stopping the repair though.

Not pretty, but ok.

service/storage_service.cc
avikivity: Why not?

```c++
auto stop_heartbeat = defer([&] {
    *heartbeat_updater_done = true;
    heartbeat_updater.get();
};
```

service/storage_service.cc
avikivity: I meant that the other nodes will not see us sending the heartbeat. But in any case if we abort, it's just a different way to abort, so ok.

service/storage_service.cc
asias: I was think this:

class deferred_action {
~deferred_action() { if (!_cancelled) { _func(); }; }
}

But the whole thing is inside a thread, so it is ok to call .get() inside the defer.

asias:
Changes in v4 8e2ced8:
- Use lowres_clock
- Fix typo ignore_nodes
- Move more logs to debug from info
- Use defer to stop heartbeat_updater
- Do not capture this in timer callback
- Fix check of uuid_opt in storage_service::node_ops_abort_thread
- Add dtests to test abort the removenode coordinator node or peer node to check auto revert works

diff between v3 and v4
git diff db8371c..8e2ced8

PR #7630, opened 2020-11-17
gossip: Added CLUSTER_INFO to `application_state`

### LINKED ISSUE #6832: Do not allow a node to join with a different snitch
Scylla version (or git commit hash): 4.x

Having nodes in the same cluster using different snitch does not make sense.
If a user forgot to set the endpoint_snitch before starting the service, the node might join in the wrong DC/Rack, which take time to fix later.
Instead, Scylla can test the snitch, similar to the cluster_name snitch.
Note that [rolling snitch switch is not supported](snitch) so this does not add a risk.

jul-stas:
CLUSTER_INFO is a map of unchangeable parameters that need to be
exchanged within cluster once, on shadow round.

Here it's mainly used to exchange snitch name, so joining nodes
cannot use wrong snitch. I also dropped cluster name into the map.

Fixes #6832

### DIFF
--- a/gms/application_state.hh
+++ b/gms/application_state.hh
@@ -65,8 +65,8 @@ enum class application_state {
     SHARD_COUNT,
     IGNORE_MSB_BITS,
     CDC_STREAMS_TIMESTAMP,
+    CLUSTER_INFO,
     // pad to allow adding new states to existing cluster
-    X9,
     X10,
 };
 
--- a/gms/application_state.cc
+++ b/gms/application_state.cc
@@ -68,6 +68,7 @@ static const std::map<application_state, sstring> application_state_names = {
     {application_state::SHARD_COUNT,            "SHARD_COUNT"},
     {application_state::IGNORE_MSB_BITS,        "IGNOR_MSB_BITS"},
     {application_state::CDC_STREAMS_TIMESTAMP,  "CDC_STREAMS_TIMESTAMP"},
+    {application_state::CLUSTER_INFO,           "CLUSTER_INFO"},
 };
 
 std::ostream& operator<<(std::ostream& os, const application_state& m) {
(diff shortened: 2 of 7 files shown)


haaawk:
I think @asias suggested in the email to name the new application state "NODE_INFO".

jul-stas:
Meh, I just noticed that mail from Asias. Converting to draft until I add the check for older versions.

haaawk:
> Meh, I just noticed that mail from Asias. Converting to draft until I add the check for older versions.

I don't think you need to add any check @jul-stas. @asias just wanted you to check that old nodes will gracefully ignore the new application state. There shouldn't be any change needed.

jul-stas:
> > Meh, I just noticed that mail from Asias. Converting to draft until I add the check for older versions.
> 
> I don't think you need to add any check @jul-stas. @asias just wanted you to check that old nodes will gracefully ignore the new application state. There shouldn't be any change needed.

I hope so, but after testing this I'll have to add columns to `system.peers` anyway, which I didn't know about, so it's missing in this PR.

gms/application_state.hh
```diff
@@ -65,8 +65,8 @@ enum class application_state {
     SHARD_COUNT,
     IGNORE_MSB_BITS,
     CDC_STREAMS_TIMESTAMP,
+    CLUSTER_INFO,
```
tgrabiec, line 68: Why not put SNITCH_NAME and CLUSTER_NAME as separate states?

jul-stas: The argument against it (not mine) was that the `application_state` should not grow forever.

tgrabiec: But you grow it anyway now. What do we gain by packing things into one state instead of separate states?

jul-stas:
v2 is coming, I'm closing this

PR #7634, opened 2020-11-17
dist/common/scripts/scylla_setup: Optionally config rsyslog destination

### LINKED ISSUE #7589: Scylla-setup should have an option to configure rsyslog
We want to support remote logging, setting rsyslog in scylla-setup will be helpful

amnonh:
This patch adds an option to scylla_setup to configure an rsyslog destination.

The monitoring stack has an option to get information from rsyslog it
requires that rsyslog on the scylla machines will send the trace line to
it.

The configuration will be in a Scylla configuration file, so it is safe to run it multiple times.

Fixes #7589

### DIFF
--- a/dist/common/scripts/scylla_setup
+++ b/dist/common/scripts/scylla_setup
@@ -89,6 +89,10 @@ def interactive_choose_nic():
             if is_valid_nic(n):
                 return n
 
+def interactive_ask_rsyslog_server():
+    print("Please input rsyslog remote server, use ip:port format.")
+    return input('ip:port> ').strip()
+
 def interactive_choose_swap_directory():
     print('Please input swapfile directory:')
     while True:
@@ -278,6 +282,10 @@ if __name__ == '__main__':
                         help='skip memory setup')
     parser.add_argument('--no-swap-setup', action='store_true', default=False,
                         help='skip swap setup')
+    parser.add_argument('--no-rsyslog-setup', action='store_true', default=False,
+                        help='skip rsyslog setup')
+    parser.add_argument('--rsyslog-server',
+                        help='specify a remote rsyslog server to send Scylla logs to. Use ip:port format, if no port is given, Scylla-Monitoring rsyslog will be used')
     if len(sys.argv) == 1:
         interactive = True
         group.required = False # interactive mode: no option is required.
@@ -320,6 +328,8 @@ if __name__ == '__main__':
     fstrim_setup = not args.no_fstrim_setup
     memory_setup = not args.no_memory_setup
     swap_setup = not args.no_swap_setup
+    rsyslog_setup = not args.no_rsyslog_setup
+    rsyslog_server = args.rsyslog_server
     selinux_reboot_required = False
     set_clocksource = False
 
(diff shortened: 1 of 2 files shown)


dist/common/scripts/scylla_setup
```diff
@@ -176,6 +181,20 @@ def current_umask():
...
+def update_rsysconf(rsyslog_server):
+    if ':' not in rsyslog_server:
+        rsyslog_server = rsyslog_server + ':1514'
+
+    with open('/etc/rsyslog.conf', 'r') as f:
+        for l in f:
+            if l.strip().startswith('#'):
+                continue
+            if rsyslog_server in l:
+                return
```
avikivity, line 193: Can we use /etc/rsyslog.d instead of editing files?

amnonh: It's a good idea, I'll switch to that, it removes the corner cases

amnonh:
@avikivity I've forced-pushed a newer version and use scylla.conf file instead of editing the global one.

avikivity:
Ok, I'm still waiting for @syuu1228 review.

syuu1228 (approved):
I think it should be implemented in same way of existing setups:
1. ask to setup or skip in interactive_ask_service()
2. make setup code to individual script, ex: scylla_rsyslog_setup
But I can send a fix to do that, so I approve this now.

slivne:
@amnonh please send a new version although takuya can help - the priority will be set according to his queue

amnonh:
@syuu1228 I've just forced-push a newer version following your review comment, it uses a dedicated script (that we need to pack with Scylla now, but I think it's a different repository).

It also asks first if the user wants that configuration before asking for the IP.

syuu1228 (approved):
Sorry for late reviewing, LGTM

dist/common/scripts/scylla_setup
```diff
@@ -550,6 +560,14 @@ if __name__ == '__main__':
 
                 run_setup_script('swap', 'scylla_swap_setup' + options)
 
+        if os.path.isdir('/etc/rsyslog.d'):
+            rsyslog_setup = interactive_ask_service('Do you want to configure rsyslog to send log to a remote repository?', 'Answer yes to setup rsyslog to a remote server, Answer no to do nothing.', rsyslog_setup)
+            args.no_rsyslog_setup = not rsyslog_setup
+            if rsyslog_setup:
+                if interactive:
+                    rsyslog_server = interactive_ask_rsyslog_server()
+                run_setup_script('rsyslog', 'scylla_rsyslog_setup --remote-server ' + rsyslog_server)
```
avikivity, line 569: This will crash if someone uses the non-interactive setup and forgets to use --remote-server (like using an existing script). We should skip if --remote-server was not specified.

amnonh: I found it confusing but I followed what happens in swap configuration.

Originally, I only looked at remove-server and if it was empty, skip, but following @syuu1228 comment changed to the current behavior

avikivity: Looks at both, or we make random ansible scripts breaks.

amnonh:
@avikivity I've forced push a version that skips the rsyslog setup if no remote server is passed as a parameter.
@syuu1228 I think that the same issue exists with the other commands that require parameters like swap.

PR #7638, opened 2020-11-18
repair: Use label for node ops metrics

### LINKED ISSUE #7270: Make repair progress metrics shorter and use label
Commit 7c8728dd735eecc5d2dae9350c606a28a3ec5876 added repair-based operation metrics.
There are two issues with those metrics:
1. The name is long for example: scylla_node_maintenance_operations_rebuild_finished_percentage
2. The self reporting metrics (i.e. metrics that a node report about itself) would be better off using a single metric differ by label.
this way a user of that metric can check for any repair-based operation.

asias:
Make it easier to be consumed by the scylla-monitor.

Fixes #7270

### DIFF
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -57,19 +57,20 @@ class node_ops_metrics {
 public:
     node_ops_metrics() {
         namespace sm = seastar::metrics;
-        _metrics.add_group("node_maintenance_operations", {
-            sm::make_gauge("bootstrap_finished_percentage", sm::description("Number of finished percentage for bootstrap operation on this shard."),
-                [this] { return bootstrap_finished_percentage(); }),
-            sm::make_gauge("replace_finished_percentage", sm::description("Number of finished percentage for replace operation on this shard."),
-                [this] { return replace_finished_percentage(); }),
-            sm::make_gauge("rebuild_finished_percentage", sm::description("Number of finished percentage for rebuild operation on this shard."),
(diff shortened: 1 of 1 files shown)


repair/repair.cc
```diff
@@ -57,19 +57,20 @@ class node_ops_metrics {
...
+                    sm::description("Number of finished percentage for replace operation on this shard."), {ops_label_type("replace")}),
+            sm::make_gauge("finished_percentage", [this] { return rebuild_finished_percentage(); },
+                    sm::description("Number of finished percentage for rebuild operation on this shard."), {ops_label_type("rebuild")}),
+            sm::make_gauge("finished_percentage", [this] { return decommission_finished_percentage(); },
+                    sm::description("Number of finished percentage for decommission operation on this shard."), {ops_label_type("decommission")}),
+            sm::make_gauge("finished_percentage", [this] { return removenode_finished_percentage(); },
+                    sm::description("Number of finished percentage for removenode operation on this shard."), {ops_label_type("removenode")}),
+            sm::make_gauge("finished_percentage", [this] { return repair_finished_percentage(); },
+                    sm::description("Number of finished percentage for repair operation on this shard."), {ops_label_type("repair")}),
```
avikivity, line 73: I think you can go further and make the variables arrays indexed by stream_reason. It will probably remove a lot of ifs.

avikivity: (later)

PR #7437, opened 2020-10-14
Avoid scanning sstables in parallel for TWCS single-partition queries

kbr-:
We introduce a new single-key sstable reader for sstables created by `TimeWindowCompactionStrategy`.

The reader uses the fact that sstables created by TWCS are mostly disjoint with respect to the contained `position_in_partition`s in order to avoid having multiple sstable readers opened at the same time unnecessarily. In case there are overlapping ranges (for example, in the current time-window), it performs the necessary merging (it uses `clustering_order_reader_merger`, introduced recently).

The reader uses min/max clustering key metadata present in `md` sstables in order to decide when to open or close a sstable reader.

The following experiment was performed:
1. create a TWCS table with 1 minute windows
2. fill the table with 8 equal windows of data
   (each window flushed to a separate sstable)
3. perform `select * from ks.t where pk = 0 limit 1` query
   with and without the change

The expectation is that with the commit, only one sstable will be opened
to fetch that one row; without the commit all 8 sstables would be opened at once.
The difference in the value of `scylla_reactor_aio_bytes_read` was measured
(value after the query minus value before the query), both with and without the commit.

With the commit, the difference was 67584.
Without the commit, the difference was 528384.
528384 / 67584 ~= 7.8.

Fixes https://github.com/scylladb/scylla/issues/6418.

PR #7646, opened 2020-11-18
sstables: introduce HasPartitionTombstones sstable metadata

kbr-:
This is a type with three possible values:
- `Yes` - there are partition tombstones in the sstable
- `No` - there are no partition tombstones
- `Unknown` - we don't know

The metadata is calculated and written for sstable versions greater
or equal to `mc`. For lower versions we don't write it.
When reading, if the metadata is not present, we interpret it as
`Unknown`.

Split from https://github.com/scylladb/scylla/pull/7437.

### DIFF
--- a/sstables/kl/writer.cc
+++ b/sstables/kl/writer.cc
@@ -663,7 +663,7 @@ void sstable_writer_k_l::consume_end_of_stream()
         features.disable(sstable_feature::NonCompoundRangeTombstones);
     }
     run_identifier identifier{_run_identifier};
-    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier));
+    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier), has_partition_tombstones::Unknown);
 
     if (!_leave_unsealed) {
         _sst.seal_sstable(_backup).get();
--- a/sstables/mx/writer.cc
+++ b/sstables/mx/writer.cc
@@ -608,6 +608,7 @@ private:
     } _pi_write_m;
     utils::UUID _run_identifier;
     bool _write_regular_as_static; // See #4139
+    has_partition_tombstones _has_partition_tombstones = has_partition_tombstones::No;
 
     void init_file_writers();
 
@@ -1003,6 +1004,7 @@ void writer::consume(tombstone t) {
 
     if (t) {
         _collector.update_min_max_components(clustering_key_prefix::make_empty(_schema));
+        _has_partition_tombstones = has_partition_tombstones::Yes;
     }
 }
 
@@ -1436,7 +1438,7 @@ void writer::consume_end_of_stream() {
         features.disable(sstable_feature::CorrectStaticCompact);
     }
     run_identifier identifier{_run_identifier};
-    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier));
+    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier), _has_partition_tombstones);
     if (!_cfg.leave_unsealed) {
         _sst.seal_sstable(_cfg.backup).get();
     }
--- a/sstables/sstables.cc
+++ b/sstables/sstables.cc
@@ -1600,7 +1600,7 @@ sstable::read_scylla_metadata(const io_priority_class& pc) noexcept {
 }
 
 void
-sstable::write_scylla_metadata(const io_priority_class& pc, shard_id shard, sstable_enabled_features features, struct run_identifier identifier) {
+sstable::write_scylla_metadata(const io_priority_class& pc, shard_id shard, sstable_enabled_features features, struct run_identifier identifier, has_partition_tombstones hpt) {
     auto&& first_key = get_first_decorated_key();
     auto&& last_key = get_last_decorated_key();
     auto sm = create_sharding_metadata(_schema, first_key, last_key, shard);
@@ -1618,6 +1618,9 @@ sstable::write_scylla_metadata(const io_priority_class& pc, shard_id shard, ssta
     _components->scylla_metadata->data.set<scylla_metadata_type::Sharding>(std::move(sm));
     _components->scylla_metadata->data.set<scylla_metadata_type::Features>(std::move(features));
     _components->scylla_metadata->data.set<scylla_metadata_type::RunIdentifier>(std::move(identifier));
+    if (hpt != has_partition_tombstones::Unknown) {
+        _components->scylla_metadata->data.set<scylla_metadata_type::HasPartitionTombstones>(std::move(hpt));
+    }
 
     write_simple<component_type::Scylla>(*_components->scylla_metadata, pc);
 }
--- a/sstables/types.hh
+++ b/sstables/types.hh
@@ -498,6 +498,7 @@ enum class scylla_metadata_type : uint32_t {
     Features = 2,
     ExtensionAttributes = 3,
     RunIdentifier = 4,
+    HasPartitionTombstones = 5,
 };
 
 struct run_identifier {
@@ -509,6 +510,12 @@ struct run_identifier {
     auto describe_type(sstable_version_types v, Describer f) { return f(id); }
 };
 
+enum class has_partition_tombstones : uint8_t {
+    Unknown = 0,
+    Yes = 1,
+    No = 2,
+};
+
 struct scylla_metadata {
     using extension_attributes = disk_hash<uint32_t, disk_string<uint32_t>, disk_string<uint32_t>>;
 
@@ -516,7 +523,8 @@ struct scylla_metadata {
             disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::Sharding, sharding_metadata>,
             disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::Features, sstable_enabled_features>,
             disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::ExtensionAttributes, extension_attributes>,
-            disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::RunIdentifier, run_identifier>
+            disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::RunIdentifier, run_identifier>,
+            disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::HasPartitionTombstones, has_partition_tombstones>
             > data;
 
     sstable_enabled_features get_features() const {
@@ -543,6 +551,11 @@ struct scylla_metadata {
     std::optional<utils::UUID> get_optional_run_identifier() const {
         auto* m = data.get<scylla_metadata_type::RunIdentifier, run_identifier>();
         return m ? std::make_optional(m->id) : std::nullopt;
+    }
+
+    has_partition_tombstones has_partition_tombstones() const {
+        auto v = data.get<scylla_metadata_type::HasPartitionTombstones, sstables::has_partition_tombstones>();
+        return v ? *v : has_partition_tombstones::Unknown;
     }
 
     template <typename Describer>
--- a/test/boost/sstable_datafile_test.cc
+++ b/test/boost/sstable_datafile_test.cc
@@ -6685,3 +6685,43 @@ SEASTAR_TEST_CASE(test_zero_estimated_partitions) {
         return make_ready_future<>();
     });
 }
+
+SEASTAR_TEST_CASE(test_partition_tombstones_metadata) {
+    return test_env::do_with_async([] (test_env& env) {
+        storage_service_for_tests ssft;
+        simple_schema ss;
+        auto s = ss.schema();
+        auto pks = ss.make_pkeys(2);
+
+        auto tmp = tmpdir();
+        unsigned gen = 0;
+        for (auto version : all_sstable_versions) {
+            if (version < sstable_version_types::mc) {
+                continue;
+            }
+
+            auto mut1 = mutation(s, pks[0]);
+            auto mut2 = mutation(s, pks[1]);
+            mut1.partition().apply_insert(*s, ss.make_ckey(0), ss.new_timestamp());
+            mut1.partition().apply_delete(*s, ss.make_ckey(1), ss.new_tombstone());
+            ss.add_row(mut1, ss.make_ckey(2), "val");
+            ss.delete_range(mut1, query::clustering_range::make({ss.make_ckey(3)}, {ss.make_ckey(5)}));
+            ss.add_row(mut2, ss.make_ckey(6), "val");
+
+            auto sst_gen = [&env, s, &tmp, &gen, version] () {
+                return env.make_sstable(s, tmp.path().string(), ++gen, version, big);
+            };
+
+            {
+                auto sst = make_sstable_containing(sst_gen, {mut1, mut2});
+                sst->load().get();
+                BOOST_REQUIRE(!sst->may_have_partition_tombstones());
+            }
+
+            mut2.partition().apply(ss.new_tombstone());
+            auto sst = make_sstable_containing(sst_gen, {mut1, mut2});
+            sst->load().get();
+            BOOST_REQUIRE(sst->may_have_partition_tombstones());
+        }
+    });
+}
--- a/sstables/sstables.hh
+++ b/sstables/sstables.hh
@@ -553,7 +553,7 @@ private:
     void write_compression(const io_priority_class& pc);
 
     future<> read_scylla_metadata(const io_priority_class& pc) noexcept;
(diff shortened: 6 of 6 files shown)


test/boost/sstable_datafile_test.cc
```diff
@@ -6685,3 +6685,33 @@ SEASTAR_TEST_CASE(test_zero_estimated_partitions) {
         return make_ready_future<>();
     });
 }
+
+SEASTAR_TEST_CASE(test_cannot_have_partition_tombstones_metadata) {
```
haaawk, line 6689: The test is a bit simplistic but better than nothing. An extension would be to have more than one partition and only one of them having a partition tombstone.

kbr-: done

sstables/sstables.cc
```diff
@@ -1618,6 +1618,7 @@ sstable::write_scylla_metadata(const io_priority_class& pc, shard_id shard, ssta
     _components->scylla_metadata->data.set<scylla_metadata_type::Sharding>(std::move(sm));
     _components->scylla_metadata->data.set<scylla_metadata_type::Features>(std::move(features));
     _components->scylla_metadata->data.set<scylla_metadata_type::RunIdentifier>(std::move(identifier));
+    _components->scylla_metadata->data.set<scylla_metadata_type::CannotHavePartitionTombstones>(std::move(cannot_have_partition_tombstones));
```
haaawk, line 1621: std::move for a bool seems like overkill. I would remove std::move as it's useless noise.

sstables/kl/writer.cc
```diff
@@ -663,7 +663,7 @@ void sstable_writer_k_l::consume_end_of_stream()
         features.disable(sstable_feature::NonCompoundRangeTombstones);
     }
     run_identifier identifier{_run_identifier};
-    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier));
+    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier), false);
```
haaawk, line 666: Why K/L sstables always have the new metadata set to false? Out of laziness or is there a reason?

haaawk (approved):
LGTM

sstables/sstables.cc
kbr-: without it the code doesn't compile ;)

sstables/kl/writer.cc
kbr-: Laziness and I only really needed the metadata for >= md

test/boost/sstable_datafile_test.cc
```diff
@@ -6685,3 +6685,41 @@ SEASTAR_TEST_CASE(test_zero_estimated_partitions) {
...
+            ss.add_row(mut1, ss.make_ckey(2), "val");
+            ss.delete_range(mut1, query::clustering_range::make({ss.make_ckey(3)}, {ss.make_ckey(5)}));
+            ss.add_row(mut2, ss.make_ckey(6), "val");
+
+            auto sst_gen = [&env, s, &tmp, &gen, version] () {
+                return env.make_sstable(s, tmp.path().string(), ++gen, version, big);
+            };
+
+            {
+                auto sst = make_sstable_containing(sst_gen, {mut1, mut2});
```
raphaelsc, line 6716: I suggest you reload the sstable to force the metadata to be read from disk rather than served from memory as leftover by the writer.

sstables/sstables.cc
haaawk: wh00t!!! What's the error?

kbr-: It was something with boost's variant's `operator=` being defined for rvalue-reference only

sstables/mx/writer.cc
```diff
@@ -608,6 +608,7 @@ class writer : public sstable_writer::writer_impl {
     } _pi_write_m;
     utils::UUID _run_identifier;
     bool _write_regular_as_static; // See #4139
+    bool _cannot_have_partition_tombstones = true;
```
raphaelsc, line 611: cannot_have_partition_tombstones is hard to parse to me. what about renaming it to contains_partition_tombstones and setting it to false by default?

sstables/kl/writer.cc
raphaelsc: I'd like to understand this as well. if we set it to false by default, the metadata will be made available, and if it's set incorrectly, that could lead to a wrong decision down the road.

raphaelsc (changes requested):

sstables/kl/writer.cc
haaawk: I think it's safe because the metadata either says there's no partition tombstone or we don't know. Like a bloom filter.

haaawk: false means we know nothing partition tombstones are there or not.

sstables/sstables.cc
raphaelsc: that's because disk_set_... defines set as follow:     template <TagType Tag, typename T>
    void set(T&& value) {
        data[Tag] = disk_tagged_union_member<TagType, Tag, T>{std::forward<T>(value)};
    }

alternatively, we could perfect forward cannot_... to it rather than moving

haaawk: nah. std::forward is even more noise than std::move.

sstables/kl/writer.cc
raphaelsc: so perhaps we shouldn't even bother writing it for older versions

raphaelsc: we do have the feature bit set in the scylla file however so that should be fine backward compatibility wise

sstables/types.hh
```diff
@@ -498,6 +498,7 @@ enum class scylla_metadata_type : uint32_t {
     Features = 2,
     ExtensionAttributes = 3,
     RunIdentifier = 4,
+    CannotHavePartitionTombstones = 5, // true => no partition tombstones, false => we don't know
```
raphaelsc, line 501: this doc would be be located at the top of cannot_have_partition_tombstones()

sstables/mx/writer.cc
```diff
@@ -1003,6 +1004,7 @@ void writer::consume(tombstone t) {
 
     if (t) {
         _collector.update_min_max_components(clustering_key_prefix::make_empty(_schema));
+        _cannot_have_partition_tombstones = false;
```
raphaelsc, line 1007: @kbr- could u pls explain to me why we don't need to track range tombstones as well? as part of https://github.com/scylladb/scylla/pull/7437 work. would we have a problem in 7437 if we have a range tombstone which is essentially a partition tombstone?

raphaelsc: In this commit https://github.com/scylladb/scylla/pull/7437/commits/34bb4d6da434d8cf55e8932c137da3466dc44abb, I found this comment:
    // 3. The sstables cannot have partition tombstones for the same reason as above. There
    //    is no way to check exactly this, but there is a way to check if the sstable has no tombstones
    //    of any type, so we use that. TWCS sstables will usually pass this condition.

But we aren't checking for range tombstones here, and I think they're important when doing the single-partition query filtering optimization.

kbr-: The comment is outdated. It turned out that we don't have a way that "sstable has no tombstones of any type", hence this PR.

Range tombstones are OK. Only partition tombstones are a problem since the new reader merger could encounter a partition tombstone in the middle of a partition query long after it returned a `partition_start` fragment.

kbr-: Range tombstones have a position in partition and are returned in order of their positions.

Partition tombstones are always returned as the first thing by mutation readers (in the `partition_start` fragment). However, the reader merger from #7437 is opening new mutation readers in the middle of a partition, after it already merged fragments returned from other readers. So imagine the following scenario: we're merging readers `r1` and `r2`:
1. the merger opens `r1`
2. `r1` returns `partition_start` with no tombstone, the merger returns it
3. `r1` returns a clustering row, the merger returns it
4. the merger opens `r2`
5. `r2` returns `partition_start` with a partition tombstone
6. ??? The merger cannot return it since it's illegal to return a `partition_start` twice. And it cannot merge it with the first `partition_start` since it already returned it.

The simplest way to deal with this problem is to ensure that there are no partition tombstones.

kbr-:
Small fixes, see diff: https://github.com/scylladb/scylla/compare/4489df8441e2266c7f0ed74b650bcc90be666c31..4e90cac08ce3e80dad39eb579e9266fba02933f2

sstables/types.hh
kbr-: done

test/boost/sstable_datafile_test.cc
kbr-: done

sstables/kl/writer.cc
avikivity, line 666: Better to use a bool_class, or even better a struct with a bool member. That naked `false` is less readable.

sstables/sstables.cc
avikivity, line 1621: I find the name "cannot have" a little awkward. It's describing something already done, and immutable. More natural to say "has no".

avikivity: Or even, "has partition tombstones", to avoid the double negative.

No entry -> unknown
true -> has partition tombstones
false -> does not have partition tombstones

kbr-: but the negation to "has no" is "has", which would suggest that the flag being false means that the sstable has partition tombstones. But it merely means that it may have - we don't know if it does.

kbr-: https://github.com/scylladb/scylla/pull/7437#discussion_r521200478

sstables/mx/writer.cc
```diff
@@ -608,6 +608,7 @@ class writer : public sstable_writer::writer_impl {
     } _pi_write_m;
     utils::UUID _run_identifier;
     bool _write_regular_as_static; // See #4139
+    bool _cannot_have_partition_tombstones = true;
```
avikivity, line 612: Maybe we should collect the lowest partition tombstone timestamp instead (and if there are none, it's the maximum timestamp). Maybe this can help with something, and has exactly the same collection cost.

@tgrabiec @raphaelsc maybe you can find a use for it.

sstables/sstables.cc
kbr-: Ok, maybe I'll just use an enum class `HasPartitionTombstones` with three options: `Yes`, `No`, `Unknown`

avikivity: There is a payload, a bool, no?

So if the section is missing, it's unknown, and if it exists, it's known, and the bool indicates what.

kbr-: I was worried about the API. If we define a function `has_partition_tombstones()`, it shouldn't return a `bool`, it should return a type with 3 possible values. On the other hand, if we define a function like `can_have_partition_tombstones()` or `cannot_have_partition_tombstones()` or `may_have_partition_tombstones()` or whatever, then it can return a `bool` and the semantics are consistent with the naming.

sstables/kl/writer.cc
kbr-: Using `enum class` now.

kbr-:
new version:
- `CannotHavePartitionTombstones` was replaced with `HasPartitionTombstones`. It is an enum class with three possible values: `Yes`, `No`, and `Unknown`.
- in `write_scylla_metadata` if the value of `has_partition_tombstones` is `Unknown`, we don't write it at all. This is used in the `kl` writer.
- `sstable::cannot_have_partition_tombstones` was renamed to `sstable::may_have_partition_tombstones` to avoid double negatives.

sstables/sstables.cc
kbr-: check out the new version, now using enum class with 3 possible values.

sstables/mx/writer.cc
avikivity: btw, the reason I am expanding the scope is that I'd like to make format changes as rare as we can.

raphaelsc: in sst 3 format, there's min deletion time metadata to track if a sstable has tombstones of any sort, so it makes sense to have min partition tombstone deletion time to track for existence of partition tombstone.

kbr-: > min deletion time metadata to track if a sstable has tombstones of any sort

Note that it's also updated on TTLs, hence it cannot be really used to track "real" tombstones. That's why I needed this new metadata.

kbr-: > it makes sense to have min partition tombstone deletion time to track for existence of partition tombstone

But would it have an actual use? I could modify the PR to collect a timestamp instead of a flag but honestly I like it the way it is now :)

kbr-: Btw. if we want to avoid format changes, maybe we could somehow make an extensible "statistics" metadata entry. Like `stats_metadata` but Scylla specific. However, `stats_metadata` is not extensible as far as I understand, and I don't know how we could make a nice extensible field (which can be appended to over time). I guess we could use a `disk_array` but that wouldn't be very elegant since it doesn't have type information. It would be best to represent the field as a `struct`. It probably could be done with a sufficient amount of template magic but I don't know if doing this as part of this PR wouldn't miss the point (I just want to finish #7437 :( ).

raphaelsc: with the boolean, we gain simplicity and I like it more too. with the min timestamp, we have extra information that may or may not be required later on. however, extra information is useless if it will never be used. so we'd only go with the latter if anybody can think of where it could possibly be applied in the future.

kbr-: > with the boolean, we gain simplicity and I like it more too. with the min timestamp, we have extra information that may or may not be required later on. however, extra information is useless if it will never be used. so we'd only go with the latter if anybody can think of where it could possibly be applied in the future.

also it's not the end of the world if we eventually end up adding a new field. It's inside Scylla-specific metadata anyway which works in a backwards-compatible way as far as I understand (does it?).

raphaelsc: > > with the boolean, we gain simplicity and I like it more too. with the min timestamp, we have extra information that may or may not be required later on. however, extra information is useless if it will never be used. so we'd only go with the latter if anybody can think of where it could possibly be applied in the future.
> 
> also it's not the end of the world if we eventually end up adding a new field. It's inside Scylla-specific metadata anyway which works in a backwards-compatible way as far as I understand (does it?).

it does. scylla_metadata is capable of determining whether a metadata is available or not, so it's important to have a safe default value, like you did.

raphaelsc: but please keep in mind that it's important to anticipate future usage of metadata to prevent scylla component from being eventually bloated. if we know that adding a bit more info to a metadata field can *very likely* be useful down the road, then go for it.

sstables/mx/writer.cc
raphaelsc: thanks for the explanation!

raphaelsc (approved):
LGTM. loved the new version with enum for has_partition_tombstones

sstables/mx/writer.cc
```diff
@@ -1003,6 +1004,7 @@ void writer::consume(tombstone t) {
 
     if (t) {
         _collector.update_min_max_components(clustering_key_prefix::make_empty(_schema));
+        _has_partition_tombstones = has_partition_tombstones::Yes;
```
avikivity, line 1007: Wait. We wanted this new metadata to be sure to include an sstable that has a tombstone early. But the previous line (update_min_max_components) will already do that! Filtering based on clustering key will include it early, since the clustering range for this sstable will be empty.

kbr-: On the contrary, the clustering range will be the full range (`before_all_clustered`, `after_all_clustered`). This is what empty clustering key prefix denotes.

If we would set the sstable's clustering range to be empty on a partition tombstone, the current sstable reading functions would give incorrect results because they would not include sstables which have partition tombstones (the ck filter would filter them out).

If the range contains at least one mutation fragment in the partition, the range must include this fragment's position. Since partition tombstones are "everywhere" - they affect all other positions - if the sstable contains a partition tombstone, its range must be the full range.

kbr-: > We wanted this new metadata to be sure to include an sstable that has a tombstone early

This metadata is used to determine whether or not we should use the new TWCS-specific sstable-reader-merger. The merger only works correctly when there are no partition tombstones in the merged sstables. If there may be (which we check using this metadata), we use the old merger.

kbr-: I guess we could use the following method:
- use the new merger only if the clustering range of the sstable is not full.

Reason: it is safe to use the new merger if the clustering range is not full, because that means that it doesn't have partition tombstones.

However I'm not sure if this works for all sstables -- created by Cassandra or an older version of Scylla, i.e. if they enforce this condition -- so the safer condition would be:
- use the new merger only if the clustering range of the sstable is not full, the version is `>= md`, and the sstable has a Scylla component.

PR #7649, opened 2020-11-19
tools: Rename centos:7.sh to centos-7.sh

asias:
Fix scylla.git checout issue on NTFS.

$ git reset --hard origin/master
error: invalid path 'tools/testing/dist-check/docker.io/centos:7.sh'
fatal: Could not reset index file to revision 'origin/master'.


penberg:
@asias We also need to adjust one of the scripts, please see: https://github.com/scylladb/scylla/pull/7648

asias:
It is amazing Pekka and I started to work on this at the same time 3 days later after my report of this issue ;-)

PR #7654, opened 2020-11-19
Don't add delay to the timestamp of the first CDC generation

### LINKED ISSUE #7645: Creating a table with CDC failed: get_stream: could not find any CDC stream
Scylla version (or git commit hash): 4.3rc1
Cluster size: 1

```
docker run --name some-scylla-43 -d scylladb/scylla:4.3.rc1  --smp 1 --memory 1G --experimental 1

cqlsh> CREATE KEYSPACE ks WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;
cqlsh> CREATE TABLE ks.t (pk int, ck int, v int, PRIMARY KEY (pk, ck, v)) WITH cdc = {'enabled':true};
cqlsh> INSERT INTO ks.t (pk,ck,v) VALUES (1,2,3);
ServerError: cdc::metadata::get_stream: could not find any CDC stream (current time: 2020/11/18 15:01:46). Are we in the middle of a cluster upgrade?
```

haaawk:
After the concept of the seed nodes was removed we can distinguish
whether the node is the first node in the cluster or not.

Thanks to this we can avoid adding delay to the timestamp of the first
CDC generation.

The delay is added to the timestamp to make sure that all the nodes
in the cluster manage to learn about it before the timestamp becomes in the past.
It is safe to not add the delay for the first node because we know it's the only node
in the cluster and no one else has to learn about the timestamp.

Fixes #7645

Tests: unit(dev)

### DIFF
--- a/cdc/generation.cc
+++ b/cdc/generation.cc
@@ -302,13 +302,13 @@ db_clock::time_point make_new_cdc_generation(
         const gms::gossiper& g,
         db::system_distributed_keyspace& sys_dist_ks,
         std::chrono::milliseconds ring_delay,
-        bool for_testing) {
+        bool add_delay) {
     using namespace std::chrono;
     auto gen = topology_description_generator(cfg, bootstrap_tokens, tmptr, g).generate();
 
     // Begin the race.
     auto ts = db_clock::now() + (
-            (for_testing || ring_delay == milliseconds(0)) ? milliseconds(0) : (
+            (!add_delay || ring_delay == milliseconds(0)) ? milliseconds(0) : (
                 2 * ring_delay + duration_cast<milliseconds>(generation_leeway)));
     sys_dist_ks.insert_cdc_topology_description(ts, std::move(gen), { tmptr->count_normal_token_owners() }).get();
 
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -570,7 +570,7 @@ void storage_service::join_token_ring(int delay) {
             try {
                 _cdc_streams_ts = cdc::make_new_cdc_generation(db().local().get_config(),
                         _bootstrap_tokens, get_token_metadata_ptr(), _gossiper,
-                        _sys_dist_ks.local(), get_ring_delay(), _for_testing);
+                        _sys_dist_ks.local(), get_ring_delay(), !_for_testing);
             } catch (...) {
                 cdc_log.warn(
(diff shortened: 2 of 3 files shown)


service/storage_service.cc
```diff
@@ -863,7 +863,7 @@ future<> storage_service::check_and_repair_cdc_streams() {
         }
         const auto new_streams_ts = cdc::make_new_cdc_generation(db().local().get_config(),
                 {}, std::move(tmptr), _gossiper,
-                _sys_dist_ks.local(), get_ring_delay(), false /* for_testing */);
+                _sys_dist_ks.local(), get_ring_delay(), !is_first_node() /* add delay */);
```
kbr-, line 866: When we create a new generation in an existing cluster it must have a delay! At this point it doesn't matter which node it is.

The trick of not adding a delay on the first node works because it's the **only** node, so the generation doesn't need to be gossiped around the cluster.

kbr- (changes requested):

service/storage_service.cc
kbr-: The PR letter should also explain why the delay was added in the first place and why it's safe not to add the delay for the first node.

haaawk: Right. I confussed is_first with is_only

haaawk: Fixed.

kbr- (approved):

cdc/generation.cc
```diff
@@ -302,13 +302,13 @@ db_clock::time_point make_new_cdc_generation(
         const gms::gossiper& g,
         db::system_distributed_keyspace& sys_dist_ks,
         std::chrono::milliseconds ring_delay,
-        bool for_testing) {
+        bool add_delay) {
```
nyh, line 305: nitpick - the name is somewhat misleading, because even if add_delay=true we *don't* add the delay when ring_delay=0 (there is a "||" below). But I guess we can live with that.

nyh (approved):

cdc/generation.cc
haaawk: yeah. I thought about it when I was changing the name and couldn't have figured a better way.
One solution would be to move the ring_delay == 0 check out of the method.

haaawk:
@nyh @avikivity could we backport this to 4.3 please?

PR #7655, opened 2020-11-19
sstables: a bunch of refactors

kbr-:
1. sstables: move `sstable_set` implementations to a separate module

    All the implementations were kept in sstables/compaction_strategy.cc
    which is quite large even without them. `sstable_set` already had its
    own header file, now it gets its own implementation file.

    The declarations of implementation classes and interfaces (`sstable_set_impl`,
    `bag_sstable_set`, and so on) were also exposed in a header file,
    sstable_set_impl.hh, for the purposes of potential unit testing.

2. mutation_reader: move `mutation_reader::forwarding` to flat_mutation_reader.hh

    Files which need this definition won't have to include
    mutation_reader.hh, only flat_mutation_reader.hh (so the inclusions are
    in total smaller; mutation_reader.hh includes flat_mutation_reader.hh).

3. sstables: move sstable reader creation functions to `sstable_set`

    Lower level functions such as `create_single_key_sstable_reader`
    were made methods of `sstable_set`.

    The motivation is that each concrete sstable_set
    may decide to use a better sstable reading algorithm specific to the
    data structures used by this sstable_set. For this it needs to access
    the set's internals.

    A nice side effect is that we moved some code out of table.cc
    and database.hh which are huge files.

4. sstables: pass `ring_position` to `create_single_key_sstable_reader`

    instead of `partition_range`.

    It would be best to pass `partition_key` or `decorated_key` here.
    However, the implementation of this function needs a `partition_range`
    to pass into `sstable_set::select`, and `partition_range` must be
    constructed from `ring_position`s. We could create the `ring_position`
    internally from the key but that would involve a copy which we want to
    avoid.

5. sstable_set: refactor `filter_sstable_for_reader_by_pk`

    Introduce a `make_pk_filter` function, which given a ring position,
    returns a boolean function (a filter) that given a sstable, tells
    whether the sstable may contain rows with the given position.

    The logic has been extracted from `filter_sstable_for_reader_by_pk`.

Split from #7437.


sstables/sstable_set.hh
```diff
@@ -111,7 +111,7 @@ public:
         schema_ptr,
         reader_permit,
         utils::estimated_histogram&,
-        const dht::partition_range&, // must be singular
+        const dht::ring_position&,
```
avikivity, line 114: I think that partition_key or decorated_key is more accurate.

ring_position doesn't necessarily denote a partition. It can denote a token.

avikivity: (ring_position denotes a boundary, not a point)

kbr-: We discussed this with @denesb on https://github.com/scylladb/scylla/pull/7437#discussion_r506067353:

the problem is that `sstable_set::select` (called in this function) takes a `partition_range`, which is constructed from `ring_position`s. Therefore I need a `ring_position` inside this function. I could construct it from passed-in `partition_key` or `decorated_key`, but that would involve a copy, and I want to avoid unnecessary copies here (the caller of the function would first extract the key from `ring_position`, then we would again copy it into a new `ring_position` --- lots of unnecessary work).

Passing `ring_position` instead of `partition_range` is still an improvement.

However, if you think that the copy is not so bad I could do it (in the context it's not really much, we're reading from disk here after all).

The right solution would probably be to modify `select` to take some kind of new type of `partition_range` that uses `ring_position_view`s. Constructing a `ring_position_view` from a `decorated_key` is cheap. However, we don't have that kind of `partition_range` class, so I'd have to write it.

denesb: I think we can leave it as a `dht::ring_position`. The best solution would be to squash this with range version and decide on the reader to create internally, based on whether the range is singular or not.

In fact we should have a single factory function, which either leaves the wrapping with `restricting_reader` to the caller, or we add a flag `wait_for_admission` and do it if this is `wait_for_admission::yes`. We can do this in a separate PR.

kbr-:
v2: added some forgotten includes in sstable_set_impl.hh

sstables/sstable_set.hh
avikivity: ok.

kbr-:
v3:
- added a comment at `create_single_key_sstable_reader` declaration that the `ring_position` must contain a key.
- Improved the commit message for the commit which changes `partition_range` to `ring_position`.

See diff: https://github.com/scylladb/scylla/compare/f7943b1c8daf8ae1f2d5f0362f3ef78cf0b07238..6c8b0af5055cbd528bd8d7d6af574c894ec99b41

sstables/sstable_set.hh
kbr-: For now I just added a comment:
```
+        const dht::ring_position&, // must contain a key
```

avikivity:
@denesb I rely on your review here. I am okay with the principles from previous reviews (including in the original series).

denesb (approved):

PR #7660, opened 2020-11-19
Changing hints configuration followup

piodul:
Follow-up to https://github.com/scylladb/scylla/pull/6916.

- Fixes wrong usage of `resource_manager::prepare_per_device_limits`,
- Improves locking in `resource_manager` so that it is more safe to call its methods concurrently,
- Adds comments around `resource_manager::register_manager` so that it's more clear what this method does and why.


avikivity:
@piodul there is no need to change the merge target to next.

PR #7665, opened 2020-11-20
index: Ensure restriction is supported in find_idx

### LINKED ISSUE #7659: Query with multiple indexes, one date one boolean, fail with ServerError
*Installation details*
Scylla version (or git commit hash): 4.4.dev-0.20201116.df197e36f
Cluster size: 1
OS (RHEL/CentOS/Ubuntu/AWS AMI): docker

While looking at at issue raised by user, I've run into this case
When we have multiple indexes, one on date, on on boolean
query on both with ALLOW_FILTERING is failing, with unclear error.

```python
CREATE KEYSPACE ks WITH replication = {'class':'SimpleStrategy', 'replication_factor':1};
 
CREATE TABLE ks.index_with_dates (id uuid PRIMARY KEY,
                                  deleted boolean,     
                                  name text,     
                                  text text,  
                                  start_date date , 
                                  end_date date);

CREATE INDEX index_with_dates_deleted_idx ON ks.index_with_dates (deleted);
CREATE INDEX index_with_dates_start_date_idx ON ks.index_with_dates (start_date);
CREATE INDEX index_with_dates_end_date_idx ON ks.index_with_dates (end_date);

USE ks; 

# query when no data
SELECT * FROM index_with_dates WHERE 
      start_date <= toDate(now()) AND 
      end_date >= toDate(now()) AND 
     deleted = false ALLOW FILTERING ;
ServerError: std::get: wrong index for variant

# query when there is partial data
INSERT INTO index_with_dates (id, deleted) VALUES ( UUID(), true );
INSERT INTO index_with_dates (id, deleted) VALUES ( UUID(), false );

SELECT * FROM index_with_dates WHERE 
      start_date <= toDate(now()) AND 
      end_date >= toDate(now()) AND 
     deleted = false ALLOW FILTERING ;
ServerError: std::get: wrong index for variant

# query with all indexes has data, also fails
TRUNCATE index_with_dates;
INSERT INTO index_with_dates (id, deleted, start_date, end_date) VALUES 
                             (UUID(), false, toDate(now()), toDate(now()) ); 
INSERT INTO index_with_dates (id, deleted, start_date, end_date) VALUES 
                             (UUID(), true, toDate(now()), toDate(now()) );

SELECT * FROM index_with_dates WHERE 
      start_date <= toDate(now()) AND 
      end_date >= toDate(now()) AND 
     deleted = false ALLOW FILTERING ;
ServerError: std::get: wrong index for variant
```

dekimir:
Previously, statement_restrictions::find_idx() would happily return an
index for a non-EQ restriction (because it checked only the column
name, not the operator).  This is incorrect: when the selected index
is for a non-EQ restriction, it is impossible to query that index
table.

Fixes #7659.

Tests: unit (dev)

### DIFF
--- a/cql3/restrictions/statement_restrictions.cc
+++ b/cql3/restrictions/statement_restrictions.cc
@@ -330,20 +330,39 @@ int statement_restrictions::score(const secondary_index::index& index) const {
     return 1;
 }
 
+namespace {
+
+using namespace cql3::restrictions;
+
+/// If rs contains a restrictions_map of individual columns to their restrictions, returns it.  Otherwise, returns null.
+const single_column_restrictions::restrictions_map* get_individual_restrictions_map(const restrictions* rs) {
+    if (auto regular = dynamic_cast<const single_column_restrictions*>(rs)) {
+        return &regular->restrictions();
+    } else if (auto partition = dynamic_cast<const single_column_partition_key_restrictions*>(rs)) {
+        return &partition->restrictions();
+    } else if (auto clustering = dynamic_cast<const single_column_clustering_key_restrictions*>(rs)) {
+        return &clustering->restrictions();
+    }
+    return nullptr;
+}
+
+} // anonymous namespace
+
 std::pair<std::optional<secondary_index::index>, ::shared_ptr<cql3::restrictions::restrictions>> statement_restrictions::find_idx(secondary_index::secondary_index_manager& sim) const {
     std::optional<secondary_index::index> chosen_index;
     int chosen_index_score = 0;
     ::shared_ptr<cql3::restrictions::restrictions> chosen_index_restrictions;
 
     for (const auto& index : sim.list_indexes()) {
+        auto cdef = _schema->get_column_definition(to_bytes(index.target_column()));
         for (::shared_ptr<cql3::restrictions::restrictions> restriction : index_restrictions()) {
-            for (const auto& cdef : restriction->get_column_defs()) {
-                if (index.depends_on(*cdef)) {
-                    if (score(index) > chosen_index_score) {
-                        chosen_index = index;
-                        chosen_index_score = score(index);
-                        chosen_index_restrictions = restriction;
-                    }
+            if (auto rmap = get_individual_restrictions_map(restriction.get())) {
+                const auto found = rmap->find(cdef);
+                if (found != rmap->end() && is_supported_by(found->second->expression, index)
+                    && score(index) > chosen_index_score) {
+                    chosen_index = index;
+                    chosen_index_score = score(index);
+                    chosen_index_restrictions = restriction;
                 }
             }
         }
--- a/test/boost/secondary_index_test.cc
+++ b/test/boost/secondary_index_test.cc
@@ -323,6 +323,13 @@ SEASTAR_TEST_CASE(test_many_columns) {
                 {{int32_type->decompose(1)}, {int32_type->decompose(2)}, {int32_type->decompose(3)}, {int32_type->decompose(4)}, {int32_type->decompose(5)}, {int32_type->decompose(6)}},
             });
         });
+        BOOST_TEST_PASSPOINT();
+        eventually([&] {
+            // #7659
+            cquery_nofail(e, "SELECT * FROM tab WHERE d=0 AND f>0 ALLOW FILTERING");
+            cquery_nofail(e, "SELECT * FROM tab WHERE f=0 AND d>0 ALLOW FILTERING");
+            cquery_nofail(e, "SELECT * FROM tab WHERE f=0 AND f>0 ALLOW FILTERING");
+        });
     });
 }
 
--- a/cql3/restrictions/single_column_restrictions.hh
+++ b/cql3/restrictions/single_column_restrictions.hh
@@ -119,7 +119,7 @@ public:
      * @param column_def the column definition
      * @return the restriction associated to the specified column
      */
-    ::shared_ptr<restriction> get_restriction(const column_definition& column_def) const {
+    ::shared_ptr<single_column_restriction> get_restriction(const column_definition& column_def) const {
         auto i = _restrictions.find(&column_def);
         if (i == _restrictions.end()) {
             return {};


psarna (approved):
Ouch, a sneaky bug. Looks good.

cql3/restrictions/statement_restrictions.cc
```diff
@@ -329,20 +329,38 @@ int statement_restrictions::score(const secondary_index::index& index) const {
...
+
+/// If rs contains a restrictions_map of individual columns to their restrictions, returns it.
+std::optional<single_column_restrictions::restrictions_map> get_individual_restrictions_map(const restrictions* rs) {
+    if (auto regular = dynamic_cast<const single_column_restrictions*>(rs)) {
+        return regular->restrictions();
+    } else if (auto partition = dynamic_cast<const single_column_partition_key_restrictions*>(rs)) {
+        return partition->restrictions();
+    } else if (auto clustering = dynamic_cast<const single_column_clustering_key_restrictions*>(rs)) {
+        return clustering->restrictions();
+    }
```
avikivity, line 344: This code is saying, please convert me to expr, I'll be so much nicer.

avikivity: (of course this quick fix should come first)

avikivity:
I have a question about `is_supported_by()`:

```c++
bool is_supported_by(const expression& expr, const secondary_index::index& idx) {
    using std::placeholders::_1;
    return std::visit(overloaded_functor{
            [&] (const conjunction& conj) {
                return boost::algorithm::all_of(conj.children, std::bind(is_supported_by, _1, idx));
            },
            [&] (const binary_operator& oper) {
                return std::visit(overloaded_functor{
                        [&] (const column_value& col) {
                            return idx.supports_expression(*col.col, oper.op);
                        },
                        [&] (const std::vector<column_value>& cvs) {
                            return boost::algorithm::any_of(cvs, [&] (const column_value& c) {
                                return idx.supports_expression(*c.col, oper.op);
                            });
                        },
                        [&] (const token&) { return false; },
                    }, oper.lhs);
            },
            [] (const auto& default_case) { return false; }
        }, expr);
}
```

We use all_of() for a conjunction, but any_of() of vector<column_value>. I think any_of is correct for both; if an index supports one restriction, the other can be filtered. That's better than filtering everything.

`WHERE a = :a AND b = :b` -> should match an index on a, but doesn't now.

`WHERE (a, b) = :a_b` -> matches an index on a, as it should

If I'm correct, then this change introduces a regression, and is_supported_by() should be fixed first.




(looks like it wasn't a question but a statement)

dekimir:
> `WHERE a = :a AND b = :b` -> should match an index on a, but doesn't now.

It does, as each atom goes into a separate map entry.

> WHERE (a, b) = :a_b -> matches an index on a, as it should

This is independent of the any/all_of behaviour of conjunction.

I tried switching conjunction to any_of in is_supported_by(), but then the following case bombs: `WHERE a>10 AND a=11`.  The rest of the code isn't ready to deal with this yet.

> If I'm correct, then this change introduces a regression, and is_supported_by() should be fixed first.

Why precisely do you think a regression is introduced?  I'm not seeing it.

test/boost/secondary_index_test.cc
```diff
@@ -255,6 +255,9 @@ SEASTAR_TEST_CASE(test_many_columns) {
         e.execute_cql("CREATE INDEX ON tab (d)").get();
         e.execute_cql("CREATE INDEX ON tab (e)").get();
         e.execute_cql("CREATE INDEX ON tab (f)").get();
+        // #7659
+        require_rows(e, "SELECT * FROM tab WHERE d=0 AND f>0 ALLOW FILTERING", {});
```
nyh, line 259: Don't you need eventually() here because you didn't wait for the index creation to finish?
Did this test fail before this patch?

nyh (commented):
Looks almost good to me, but isn't the test missing an eventually()?

dekimir:
> > WHERE (a, b) = :a_b -> matches an index on a, as it should
> 
> This is independent of the any/all_of behaviour of conjunction.

Please ignore this remark; I just realized why you gave that example...

The rest of my comment still stands.  I'm not sure I see any regressions?

test/boost/secondary_index_test.cc
dekimir: > Don't you need eventually() here because you didn't wait for the index creation to finish?

Apparently, _I_ don't, but that may be just a quirk of my dev environment.  I think I'll add it anyway.

> Did this test fail before this patch?

Yes, it failed in the same way as #7659.

avikivity:
> > `WHERE a = :a AND b = :b` -> should match an index on a, but doesn't now.
> 
> It does, as each atom goes into a separate map entry.

What do you mean? that these are two expressions? What's the map?

If indeed it turns out that these are two expressions, there's still a bug here, it's just latent.

> 
> > WHERE (a, b) = :a_b -> matches an index on a, as it should
> 
> This is independent of the any/all_of behaviour of conjunction.
> 
> I tried switching conjunction to any_of in is_supported_by(), but then the following case bombs: `WHERE a>10 AND a=11`. The rest of the code isn't ready to deal with this yet.

Why does it bomb? just curious.

> 
> > If I'm correct, then this change introduces a regression, and is_supported_by() should be fixed first.
> 
> Why precisely do you think a regression is introduced? I'm not seeing it.

If `WHERE a = :a AND b = :b` is a single expression, then it wouldn't match an index on a due to all_of.

If it's not a single expression, then it's not a regression, just a latent bug that is not expressed.

dekimir:
> > > `WHERE a = :a AND b = :b` -> should match an index on a, but doesn't now.
> > 
> > 
> > It does, as each atom goes into a separate map entry.
> 
> What do you mean? that these are two expressions? What's the map?

See `rmap` in the changed code.

> > I tried switching conjunction to any_of in is_supported_by(), but then the following case bombs: `WHERE a>10 AND a=11`. The rest of the code isn't ready to deal with this yet.
> 
> Why does it bomb? just curious.

IIRC, `used_index_restrictions->value_for()` doesn't find a single value and asserts out.  I'll recheck when convenient, right now I'm in the middle of something else.

> If `WHERE a = :a AND b = :b` is a single expression, then it wouldn't match an index on a due to all_of.
> 
> If it's not a single expression, then it's not a regression, just a latent bug that is not expressed.

It's not a single expression.  There may be a bug, but it cannot be exposed by any query until we take the front-end rewrite a step further.

PS1: Please don't merge yet, I discovered a bug in how multi-column restrictions are treated.

PS2: The 4.2 fix will likely diverge further from this patch.  Do you think I should submit a separate patch for 4.2?

dekimir:
OK, the [bug I discovered](#7680) is orthogonal.  I addressed feedback and added more tests.  [Changelog](https://github.com/dekimir/scylla/commits/fix-find_idx-ineq.v2):
- 3435e3be8 Handle multi-column restrictions in find_idx()
- 6735b0e8c Use eventually() in new tests
- 809d10870 Add testcase for multiple atoms on same column

avikivity:
> 
> PS2: The 4.2 fix will likely diverge further from this patch. Do you think I should submit a separate patch for 4.2?

Focus on fixing on master. My approach is to prepare a backportable fix if it is natural - if the code did not diverge too much. If it did, then priority is to keep master clean and moving in the correct direction, and if it means more backport work, then that's what I do.

avikivity:
> > > > `WHERE a = :a AND b = :b` -> should match an index on a, but doesn't now.
> > > 
> > > 
> > > It does, as each atom goes into a separate map entry.
> > 
> > 
> > What do you mean? that these are two expressions? What's the map?
> 
> See `rmap` in the changed code.
> 

I see. And it's already stored split.

> > > I tried switching conjunction to any_of in is_supported_by(), but then the following case bombs: `WHERE a>10 AND a=11`. The rest of the code isn't ready to deal with this yet.
> > 
> > 
> > Why does it bomb? just curious.
> 
> IIRC, `used_index_restrictions->value_for()` doesn't find a single value and asserts out. I'll recheck when convenient, right now I'm in the middle of something else.
> 
> > If `WHERE a = :a AND b = :b` is a single expression, then it wouldn't match an index on a due to all_of.
> > If it's not a single expression, then it's not a regression, just a latent bug that is not expressed.
> 
> It's not a single expression. There may be a bug, but it cannot be exposed by any query until we take the front-end rewrite a step further.

Ok, so at  least it wants a comment.

dekimir:
Pausing this until #7699 is merged.  That will "fix" is_supported_by and allow me to simplify this patch.

dekimir:
Redone after #7699, PTAL.  [Changelog](https://github.com/dekimir/scylla/commits/fix-find_idx-ineq.v3):
* 8a25356fe Don't copy restrictions_map in get_individual_restrictions_map
* d4c4ef94a Revert 3435e3b Handle multi-column restrictions in find_idx()
* rebased on master

PR #7669, opened 2020-11-22
Use large_data_counters in scylla_metadata to decide when to delete large_data records

### LINKED ISSUE #7668: Deletion of large data records is overzealous
As @raphaelsc noted in https://github.com/scylladb/scylla/issues/7562#issuecomment-731739849,
the way we decide whether to delete a large data record is by comparing the sstable data_size to each large data (partitions / rows / cells) threshold.  This will create unneeded tombstones, especially for the low order records - for large cells and large rows, as sstables bigger than 1MB or 10MB, respectively, are very common.

bhalevy:
This series introduces a `large_data_counters` element to `scylla_metadata` component to explicitly count the number of `large_{partitions,rows,cells}` and `too_many_rows` in the sstable.  These are accounted for in the sstable writer whenever the respective large data entry is encountered.

It is taken into account in `large_data_handler::maybe_delete_large_data_entries`, when engaged.
Otherwise, if deleting a legacy sstable that has no such entry in `scylla_metadata`, just revert to using the current method of comparing the sstable's `data_size` to the various thresholds.

Fixes #7668

Test: unit(dev)
Dtest: wide_rows_test.py (in progress)

### DIFF
--- a/db/large_data_handler.cc
+++ b/db/large_data_handler.cc
@@ -25,6 +25,8 @@
 #include "db/large_data_handler.hh"
 #include "sstables/sstables.hh"
 
+static logging::logger large_data_logger("large_data");
+
 namespace db {
 
 nop_large_data_handler::nop_large_data_handler()
@@ -32,6 +34,16 @@ nop_large_data_handler::nop_large_data_handler()
           std::numeric_limits<uint64_t>::max(), std::numeric_limits<uint64_t>::max()) {
     // Don't require start() to be called on nop large_data_handler.
     start();
+}
+
+large_data_handler::large_data_handler(uint64_t partition_threshold_bytes, uint64_t row_threshold_bytes, uint64_t cell_threshold_bytes, uint64_t rows_count_threshold)
+        : _partition_threshold_bytes(partition_threshold_bytes)
+        , _row_threshold_bytes(row_threshold_bytes)
+        , _cell_threshold_bytes(cell_threshold_bytes)
+        , _rows_count_threshold(rows_count_threshold)
+{
+    large_data_logger.debug("partition_threshold_bytes={} row_threshold_bytes={} cell_threshold_bytes={} rows_count_threshold={}",
+        partition_threshold_bytes, row_threshold_bytes, cell_threshold_bytes, rows_count_threshold);
 }
 
 future<> large_data_handler::maybe_record_large_partitions(const sstables::sstable& sst, const sstables::key& key, uint64_t partition_size) {
@@ -56,8 +68,6 @@ future<> large_data_handler::stop() {
     _running = false;
     return _sem.wait(max_concurrency);
 }
-
-static logging::logger large_data_logger("large_data");
 
 template <typename T> static std::string key_to_str(const T& key, const schema& s) {
     std::ostringstream oss;
--- a/db/large_data_handler.cc
+++ b/db/large_data_handler.cc
@@ -84,19 +84,22 @@ future<> large_data_handler::maybe_delete_large_data_entries(sstables::shared_ss
     auto data_size = sst->data_size();
 
     future<> large_partitions = make_ready_future<>();
-    if (__builtin_expect(data_size > _partition_threshold_bytes, false)) {
+    auto entry = sst->get_large_data_stat(sstables::large_data_type::partition_size);
+    if (entry ? entry->above_threshold : (data_size > _partition_threshold_bytes)) {
         large_partitions = with_sem([schema, filename, this] () mutable {
             return delete_large_data_entries(*schema, std::move(filename), db::system_keyspace::LARGE_PARTITIONS);
         });
     }
     future<> large_rows = make_ready_future<>();
-    if (__builtin_expect(data_size > _row_threshold_bytes, false)) {
+    entry = sst->get_large_data_stat(sstables::large_data_type::row_size);
+    if (entry ? entry->above_threshold : (data_size > _row_threshold_bytes)) {
         large_rows = with_sem([schema, filename, this] () mutable {
             return delete_large_data_entries(*schema, std::move(filename), db::system_keyspace::LARGE_ROWS);
         });
     }
     future<> large_cells = make_ready_future<>();
-    if (__builtin_expect(data_size > _cell_threshold_bytes, false)) {
+    entry = sst->get_large_data_stat(sstables::large_data_type::cell_size);
+    if (entry ? entry->above_threshold : (data_size > _cell_threshold_bytes)) {
         large_cells = with_sem([schema, filename, this] () mutable {
             return delete_large_data_entries(*schema, std::move(filename), db::system_keyspace::LARGE_CELLS);
         });
--- a/db/large_data_handler.cc
+++ b/db/large_data_handler.cc
@@ -46,15 +46,17 @@ large_data_handler::large_data_handler(uint64_t partition_threshold_bytes, uint6
         partition_threshold_bytes, row_threshold_bytes, cell_threshold_bytes, rows_count_threshold);
 }
 
-future<> large_data_handler::maybe_record_large_partitions(const sstables::sstable& sst, const sstables::key& key, uint64_t partition_size) {
+future<bool> large_data_handler::maybe_record_large_partitions(const sstables::sstable& sst, const sstables::key& key, uint64_t partition_size) {
     assert(running());
     if (partition_size > _partition_threshold_bytes) {
         ++_stats.partitions_bigger_than_threshold;
         return with_sem([&sst, &key, partition_size, this] {
             return record_large_partitions(sst, key, partition_size);
+        }).then([] {
+            return true;
         });
     }
-    return make_ready_future<>();
+    return make_ready_future<bool>(false);
 }
 
 void large_data_handler::start() {
--- a/db/large_data_handler.cc
+++ b/db/large_data_handler.cc
@@ -77,24 +77,28 @@ template <typename T> static std::string key_to_str(const T& key, const schema& 
     return oss.str();
 }
 
-future<> large_data_handler::maybe_delete_large_data_entries(const schema& s, sstring filename, uint64_t data_size) {
+future<> large_data_handler::maybe_delete_large_data_entries(sstables::shared_sstable sst) {
     assert(running());
+    auto schema = sst->get_schema();
+    auto filename = sst->get_filename();
+    auto data_size = sst->data_size();
+
     future<> large_partitions = make_ready_future<>();
     if (__builtin_expect(data_size > _partition_threshold_bytes, false)) {
-        large_partitions = with_sem([&s, filename, this] () mutable {
-            return delete_large_data_entries(s, std::move(filename), db::system_keyspace::LARGE_PARTITIONS);
+        large_partitions = with_sem([schema, filename, this] () mutable {
+            return delete_large_data_entries(*schema, std::move(filename), db::system_keyspace::LARGE_PARTITIONS);
         });
     }
     future<> large_rows = make_ready_future<>();
     if (__builtin_expect(data_size > _row_threshold_bytes, false)) {
-        large_rows = with_sem([&s, filename, this] () mutable {
-            return delete_large_data_entries(s, std::move(filename), db::system_keyspace::LARGE_ROWS);
+        large_rows = with_sem([schema, filename, this] () mutable {
+            return delete_large_data_entries(*schema, std::move(filename), db::system_keyspace::LARGE_ROWS);
         });
     }
     future<> large_cells = make_ready_future<>();
     if (__builtin_expect(data_size > _cell_threshold_bytes, false)) {
-        large_cells = with_sem([&s, filename, this] () mutable {
-            return delete_large_data_entries(s, std::move(filename), db::system_keyspace::LARGE_CELLS);
+        large_cells = with_sem([schema, filename, this] () mutable {
+            return delete_large_data_entries(*schema, std::move(filename), db::system_keyspace::LARGE_CELLS);
         });
     }
     return when_all(std::move(large_partitions), std::move(large_rows), std::move(large_cells)).discard_result();
--- a/db/large_data_handler.cc
+++ b/db/large_data_handler.cc
@@ -77,6 +77,29 @@ template <typename T> static std::string key_to_str(const T& key, const schema& 
     return oss.str();
 }
 
+future<> large_data_handler::maybe_delete_large_data_entries(const schema& s, sstring filename, uint64_t data_size) {
+    assert(running());
+    future<> large_partitions = make_ready_future<>();
+    if (__builtin_expect(data_size > _partition_threshold_bytes, false)) {
+        large_partitions = with_sem([&s, filename, this] () mutable {
+            return delete_large_data_entries(s, std::move(filename), db::system_keyspace::LARGE_PARTITIONS);
+        });
+    }
+    future<> large_rows = make_ready_future<>();
+    if (__builtin_expect(data_size > _row_threshold_bytes, false)) {
+        large_rows = with_sem([&s, filename, this] () mutable {
+            return delete_large_data_entries(s, std::move(filename), db::system_keyspace::LARGE_ROWS);
+        });
+    }
+    future<> large_cells = make_ready_future<>();
+    if (__builtin_expect(data_size > _cell_threshold_bytes, false)) {
+        large_cells = with_sem([&s, filename, this] () mutable {
+            return delete_large_data_entries(s, std::move(filename), db::system_keyspace::LARGE_CELLS);
+        });
+    }
+    return when_all(std::move(large_partitions), std::move(large_rows), std::move(large_cells)).discard_result();
+}
+
 template <typename... Args>
 static future<> try_record(std::string_view large_table, const sstables::sstable& sst,  const sstables::key& partition_key, int64_t size,
         std::string_view desc, std::string_view extra_path, const std::vector<sstring> &extra_fields, Args&&... args) {
--- a/docs/sstable-scylla-format.md
+++ b/docs/sstable-scylla-format.md
@@ -26,6 +26,7 @@ in individual sections
         | features
         | extension_attributes
         | run_identifier
+        | large_data_stats
 
 `sharding_metadata` (tag 1): describes what token sub-ranges are included in this
 sstable. This is used, when loading the sstable, to determine which shard(s)
@@ -37,6 +38,9 @@ it occupies.
 
 `run_identifier` (tag 4): a uuid that is the same for all sstables in the same run
 (and different for sstables in different runs).
+
+`large_data_stats`: a map<large_data_type, large_data_stats_entry> with statistics
+about large data entities in the sstable.
 
 ## sharding_metadata subcomponent
 
@@ -101,3 +105,20 @@ There are currently no defined attributes.
 If the run_identifier subcomponent is present, the sstable is part of a run.
 All sstables with the same run_identifier belong to the same run. They are
 guaranteed to be disjoint (non-overlapping) in their partition keys.
+
+## large_data_stats subcomponent
+
+    large_data_stats = large_data_count large_data_pair*
+    large_data_count = be32
+    large_data_pair = large_data_type large_data_stats_entry
+    large_data_type = be32
+    large_data_stats_entry = max_value threshold above_threshold
+    max_value = be64
+    threshold = be64
+    above_threshold = be32
+
+The large_data_stats component holds statistics about partition,
+row, and cell sizes and about number of rows in partition.
+For each entry, it keeps the largest value for the entry type,
+the respective large_data threshold and the number of entities
+that are above the threshold.
--- a/sstables/kl/writer.cc
+++ b/sstables/kl/writer.cc
@@ -613,7 +613,6 @@ stop_iteration sstable_writer_k_l::consume_end_of_partition() {
     _c_stats.partition_size = _writer->offset() - _c_stats.start_offset;
 
     _sst.get_large_data_handler().maybe_record_large_partitions(_sst, *_partition_key, _c_stats.partition_size).get();
-    _sst.get_large_data_handler().maybe_log_too_many_rows(_sst, *_partition_key, _c_stats.rows_count);
 
     // update is about merging column_stats with the data being stored by collector.
     _collector.update(std::move(_c_stats));
--- a/sstables/kl/writer.cc
+++ b/sstables/kl/writer.cc
@@ -662,7 +662,7 @@ void sstable_writer_k_l::consume_end_of_stream()
         features.disable(sstable_feature::NonCompoundRangeTombstones);
     }
     run_identifier identifier{_run_identifier};
-    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier));
+    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier), {});
 
     if (!_leave_unsealed) {
         _sst.seal_sstable(_backup).get();
--- a/sstables/mx/writer.cc
+++ b/sstables/mx/writer.cc
@@ -608,6 +608,7 @@ private:
     } _pi_write_m;
     utils::UUID _run_identifier;
     bool _write_regular_as_static; // See #4139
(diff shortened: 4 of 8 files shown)


bhalevy:
Mark as draft until the tests are complete.
The code is ready for review though...

raphaelsc:
> This series introduces a `large_data_counters` element to `scylla_metadata` component to explicitly count the number of `large_{partitions,rows,cells}` and `too_many_rows` in the sstable. These are accounted for in the sstable writer whenever the respective large data entry is encountered.
> 
> It is taken into account in `large_data_handler::maybe_delete_large_data_entries`, when engaged.
> Otherwise, if deleting a legacy sstable that has no such entry in `scylla_metadata`, just revert to using the current method of comparing the sstable's `data_size` to the various thresholds.
> 
> Fixes #7668
> 
> Test: unit(dev)
> Dtest: wide_rows_test.py (in progress)

what's too_many_rows?

raphaelsc:
the direction looks good to me. the advantage of the approach taken here over read-modify-write approach is that an internal read is avoided, also we can potentially find useful to have these counters in each sstable, to help troubleshooting issues which come from large data. it will allow us to pinpoint the sstable(s) storing large data

raphaelsc:
also reviewed the code, and at first glance, it looks good to me.

bhalevy:
Note: `wide_rows_test.py:TestWideRows_with_TimeWindowCompactionStrategy.test_large_partition_detector_with_node_stop` times out for me locally, but also against the scylla baseline at 2c8dcbe5c5a85bbf9cb591520e959a222a292eab

sstables/sstables.cc
```diff
@@ -1327,6 +1327,11 @@ future<> sstable::open_data() noexcept {
             _open = true;
             return make_ready_future<>();
         });
+    }).then([this] {
+        auto* ldcs = _components->scylla_metadata->data.get<scylla_metadata_type::LargeDataCounters, scylla_metadata::large_data_counters>();
+        if (ldcs) {
+            _large_data_counters.emplace(*ldcs);
+        }
```
raphaelsc, line 1334: if we're opening a sstable just written, _large_data_counters already have the correct data. so we can just skip this part, right?

db/large_data_handler.cc
```diff
@@ -84,19 +84,19 @@ future<> large_data_handler::maybe_delete_large_data_entries(sstables::shared_ss
     auto data_size = sst->data_size();
 
     future<> large_partitions = make_ready_future<>();
-    if (__builtin_expect(data_size > _partition_threshold_bytes, false)) {
+    if (sst->get_large_data_count(sstables::large_data_type::large_partitions).value_or(data_size > _partition_threshold_bytes)) {
         large_partitions = with_sem([schema, filename, this] () mutable {
```
raphaelsc, line 88: nice :-)

raphaelsc (approved):
LGTM

sstables/sstables.cc
raphaelsc: disregard my comment above. _large_data_counters used during write is the one in sstable_writer, not the one in sstable class.

bhalevy: Right. In the first version I initialized the one in the sstable and used, but I modified the flow to follow the existing design that collects the metadata in the writer and applies it to the sstable in `write_scylla_metadata`, to be reloaded by a following `open_data` that is called in any case.

bhalevy:
> I can see advantages and disadvantages to the approach:
> 
> Advantages:
> 
> * avoids read-modify-write (minor, since sstable creation/deletion is much more expensive than a local rmw)
> * this metadata might be useful outside this use case
> 
> Disadvantages:
> 
> * adding state, for something that is transient and internal
> 
> Why not do the RMW?

@avikivity I think this approach is simpler than RMW and does not degrade performance.
If we do RMW, I'd separate the read part, and do it once on startup after loading all sstables.
Then read the `system.large_*` tables and lookup the respective sstables.
If they do not exist anymore, just delete the entry (this may happen if we shutdown scylla abruptly since we delete the sstables without waiting for the deletion of the associated rows in the system.large_* tables. See https://github.com/scylladb/scylla/issues/4264)
Otherwise, mark the sstable in-core as having large partitions/rows/cells.
Then, during normal operations when new sstables with large data are written, mark them in the same way (similar to this patchset), and use that in a similar way to this series when sstables are deleted.

The difference from this patchset is that we would know definitely if large data exists in the sstable or not for any sstable in the system.

raphaelsc:
On Mon, Nov 23, 2020 at 12:21 PM Avi Kivity <notifications@github.com>
wrote:

> I can see advantages and disadvantages to the approach:
>
> Advantages:
>
>    - avoids read-modify-write (minor, since sstable creation/deletion is
>    much more expensive than a local rmw)
>    - this metadata might be useful outside this use case
>
> Disadvantages:
>
>    - adding state, for something that is transient and internal
>
> Why not do the RMW?
>
I am worried that this could add unnecessary internal overhead which may be
significant with LCS. With LCS, you have ~10 unlinks in average for each
1.6G of compacted data. Also this metadata will be helpful in the field to
determine the content of any given sstable in terms of large data. I would
dislike if the metadata were a boolean saying whether or not this sstable
is inserted into an internal system table, but that's not the case, it was
made generic by being a counter of large cell/row/partition.

I am leaning towards the approach taken by this patchset because of the
advantages you mentioned, which weights more than the disadvantage, which
is not really a disadvantage if we consider that the counters will be very
useful in the field.

avikivity:
Adding persistent data should be done with care, it's hard to get rid of it later. Reading the data during startup could work, sstable_manager already has an interface to large_data_handler.

raphaelsc:
> Adding persistent data should be done with care, it's hard to get rid of it later. Reading the data during startup could work, sstable_manager already has an interface to large_data_handler.

we'd need to query all large_* tables for each sstable on boot.  in systems with thousands and thousands of sstables, I am worried about it adding some significant overhead to boot (increasing the overall time which is valuable to users)

we can avoid this overhead, that can catch user by surprise, with a couple of uint32_t counters that will go in scylla metadata (and these counters are potentially useful for troubleshooting). so it seems like a fair trade-off. if we decide to deprecate this metadata later, we can retire its id in scylla_metadata_type.

avikivity:
> > Adding persistent data should be done with care, it's hard to get rid of it later. Reading the data during startup could work, sstable_manager already has an interface to large_data_handler.
> 
> we'd need to query all large_* tables for each sstable on boot. in systems with thousands and thousands of sstables, I am worried about it adding some significant overhead to boot (increasing the overall time which is valuable to users)


I doubt it would be noticable. Such a query is similar to opening a file - you read some data from disk.

> 
> we can avoid this overhead, that can catch user by surprise, with a couple of uint32_t counters that will go in scylla metadata (and these counters are potentially useful for troubleshooting). so it seems like a fair trade-off. if we decide to deprecate this metadata later, we can retire its id in scylla_metadata_type.

I agree that it's small, and can be retired (after all, you can just not write it any more). But it seems somewhat arbitrary.

raphaelsc:
On Tue, Nov 24, 2020 at 2:58 PM Avi Kivity <notifications@github.com> wrote:

> Adding persistent data should be done with care, it's hard to get rid of
> it later. Reading the data during startup could work, sstable_manager
> already has an interface to large_data_handler.
>
> we'd need to query all large_* tables for each sstable on boot. in systems
> with thousands and thousands of sstables, I am worried about it adding some
> significant overhead to boot (increasing the overall time which is valuable
> to users)
>
> I doubt it would be noticable. Such a query is similar to opening a file -
> you read some data from disk.
>
hmm then we can proceed with this approach instead. however, it will not
work for the early boot phase, where sstables can be resharded and reshaped
and weren't feeded into the table's sstable set yet, so sstables unlinked
at this early stage would have a range tombstone created for them. if
resharding a LCS table, it means we'd potentially have thousands of range
tombstones in large_cells/rows for that particular table.

we can avoid this overhead, that can catch user by surprise, with a couple

raphaelsc:
On Tue, Nov 24, 2020 at 3:10 PM Raphael S. Carvalho <raphaelsc@scylladb.com>
wrote:

>
>
> On Tue, Nov 24, 2020 at 2:58 PM Avi Kivity <notifications@github.com>
> wrote:
>
>> Adding persistent data should be done with care, it's hard to get rid of
>> it later. Reading the data during startup could work, sstable_manager
>> already has an interface to large_data_handler.
>>
>> we'd need to query all large_* tables for each sstable on boot. in
>> systems with thousands and thousands of sstables, I am worried about it
>> adding some significant overhead to boot (increasing the overall time which
>> is valuable to users)
>>
>> I doubt it would be noticable. Such a query is similar to opening a file
>> - you read some data from disk.
>>
> hmm then we can proceed with this approach instead. however, it will not
> work for the early boot phase, where sstables can be resharded and reshaped
> and weren't feeded into the table's sstable set yet, so sstables unlinked
> at this early stage would have a range tombstone created for them. if
> resharding a LCS table, it means we'd potentially have thousands of range
> tombstones in large_cells/rows for that particular table.
>

wait. i think we populate system tables first and so we're able to query
them when processing non-system tables, so your suggested approach may work
at the early boot phase too.

slivne:
@avikivity there are other reason to have this metadata

- We need to analyze sstable files shared with us - its easier to know sstables have large partitions / large cells / large rows - from metadata then to go and analyze them and figure this out (we do not have a tool that does this today)

- We need to restore clusters - when we restore clusters, we download sstables from a backup and upload them into a cluster, if sharding information is incorrect - we will need to read and reshard the ssatbles and that can record the info of large partition/row/cell. However if the sharding information matches - there is no need to read the sstable it can be used - in that case we are starting from a point an sstable may have large partitions / large cells / large rows and we are not aware of that. If we had the metadata we could load the info that the sstable has large rows / large cells / large partitions (even without the specific information). This is misleading - the only way to figure this out is todo a compaction (is that really needed). Restoring from a backup / uploading sstables is not common - yet forcing sstable rewrite / reread just to figure they have large partitions / large rows / large cells is wasteful.

- In my view we should also record the limits the sstable was created with - so if an sstable has no cell greater than 50 KB we will know that all the cells are smaller than that.

- Last many times we ask users - do you have large rows / large cells / large partitions - they check the system tables - its not clear if they did it correctly or not - if we export this via nodetool cfstats we will have the info avail and handy.

avikivity:
Ok. Then let's make it useful, have it record the largest partition size, the largest row size, and the largest cell size. We can then use this information both to remove the sstable from the large records table, and to understand an sstable's contents.

bhalevy:
> Ok. Then let's make it useful, have it record the largest partition size, the largest row size, and the largest cell size. We can then use this information both to remove the sstable from the large records table, and to understand an sstable's contents.

@avikivity I like the idea of keeping the largest partition/row/cell size present in the sstable.
But if we only use that to decide whether to delete large data records from the system table at the time the sstable is deleted, without keeping the thresholds used when the sstable was written, we may either:
1. miss records if the configuration changed and the thresholds were raised, or
2. issue deletes for non-existing records if the thresholds were lowered after the sstable was written.

I don't think it's a disaster, but saving also the thresholds might not be elegant but it'd be useful for more accurate handling of the large data records.

What do you think?

avikivity:
> > Ok. Then let's make it useful, have it record the largest partition size, the largest row size, and the largest cell size. We can then use this information both to remove the sstable from the large records table, and to understand an sstable's contents.
> 
> @avikivity I like the idea of keeping the largest partition/row/cell size present in the sstable.
> But if we only use that to decide whether to delete large data records from the system table at the time the sstable is deleted, without keeping the thresholds used when the sstable was written, we may either:
> 
> 1. miss records if the configuration changed and the thresholds were raised, or
> 2. issue deletes for non-existing records if the thresholds were lowered after the sstable was written.
> 
> I don't think it's a disaster, but saving also the thresholds might not be elegant but it'd be useful for more accurate handling of the large data records.
> 


We can save both the maximum, the counts, and the thresholds used for the counts (so the sstable metadata can be understood without knowing the system thresholds).

The deletion can use just the counts.

raphaelsc:
On Thu, Nov 26, 2020 at 5:21 AM Avi Kivity <notifications@github.com> wrote:

> Ok. Then let's make it useful, have it record the largest partition size,
> the largest row size, and the largest cell size.
>

just remembered we have an estimated histogram for partition sizes, so we
can already figure out from metadata if sstable contains large partitions
based on configured threshold. we even have cf api retrieving it, possibly
for nodetool cfstats. perhaps we can add a size histogram for cells and
rows too?! which will give us even more insight into the size distribution.

bhalevy:
> On Thu, Nov 26, 2020 at 5:21 AM Avi Kivity ***@***.***> wrote: Ok. Then let's make it useful, have it record the largest partition size, the largest row size, and the largest cell size.
> just remembered we have an estimated histogram for partition sizes, so we can already figure out from metadata if sstable contains large partitions based on configured threshold. we even have cf api retrieving it, possibly for nodetool cfstats. perhaps we can add a size histogram for cells and rows too?! which will give us even more insight into the size distribution.

I don't think we need to over-complicate this. Let's keep it simple.
We can easily add the maximum values in addition to the threshold and the count of items above the threshold.
(In fact I'm testing my branch right now :))

bhalevy:
In v2:
* defined large_data_stats_entry { max_value, threshold, above_threshold }
* sstables: kl/writer: never record too many rows
* large_data_handler: expose methods to get threshold
* sstables: writer: keep track of large data stats
  - use new large_data_stats_entry
  - keep track of max value, threshold, and count of values above threshold
  - for partitions, rows, cells, and rows_count in partition
* large_data_handler: maybe_delete_large_data_entries: use sstable large data stats
  - adjust to use large_data_stats_entry.above_threshold, if available.

sstables/types.hh
```diff
@@ -509,14 +510,36 @@ struct run_identifier {
...
     using extension_attributes = disk_hash<uint32_t, disk_string<uint32_t>, disk_string<uint32_t>>;
+    using large_data_stats = disk_hash<uint32_t, large_data_type, large_data_stats_entry>;
 
     disk_set_of_tagged_union<scylla_metadata_type,
             disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::Sharding, sharding_metadata>,
             disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::Features, sstable_enabled_features>,
             disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::ExtensionAttributes, extension_attributes>,
-            disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::RunIdentifier, run_identifier>
+            disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::RunIdentifier, run_identifier>,
+            disk_tagged_union_member<scylla_metadata_type, scylla_metadata_type::LargeDataStats, large_data_stats>
```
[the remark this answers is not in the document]
bhalevy: Doesn't mater much IMO but can do...

sstables/mx/writer.cc
```diff
@@ -1006,6 +1013,80 @@ void writer::consume(tombstone t) {
...
+                .threshold = _sst.get_large_data_handler().get_cell_threshold_bytes(),
+            }
+        )).first;
+    } else if (it->second.max_value < cell_size) {
+        it->second.max_value = cell_size;
+    }
+    if (_sst.get_large_data_handler().maybe_record_large_cells(_sst, *_partition_key, clustering_key, cdef, cell_size).get0()) {
+        it->second.above_threshold++;
+    };
+}
```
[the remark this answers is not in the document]
bhalevy: looks good, except that I don't like always assigning `max_value`.  Both it and `above_threshold` change rarely and writing to memory isn't free.

sstables/kl/writer.cc
```diff
@@ -662,7 +662,7 @@ void sstable_writer_k_l::consume_end_of_stream()
         features.disable(sstable_feature::NonCompoundRangeTombstones);
     }
     run_identifier identifier{_run_identifier};
-    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier));
+    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier), std::move(_large_data_stats));
```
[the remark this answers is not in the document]
bhalevy: although we shouldn't write k/l sstables at all, you're correct.
I'll move `_large_data_stats` from `writer_impl` to the mx writer.

bhalevy:
In v3:
* moved hunk adding `large_data_stats` to `scylla_metadata` to the patch that stores it.
* add stats member only to mx writer and make it optional in `write_scylla_metadata` so it won't get written by k/l writer.
* initialize the mx writer stats in constructor to hold all thresholds and simplify `maybe_record_*` methods
  - still, set `max_value` only when changed.

sstables/mx/writer.cc
```diff
@@ -1006,6 +1040,48 @@ void writer::consume(tombstone t) {
     }
 }
 
+void writer::maybe_record_large_partitions(const sstables::sstable& sst, const sstables::key& partition_key, uint64_t partition_size) {
+    auto& entry = _large_data_stats.map.at(large_data_type::partition_size);
+    if (entry.max_value < partition_size) {
```
raphaelsc, line 1045: it's more intuitive to me if it were written if (size > max) max = size, but I am also fine with this :-)

sstables/mx/writer.cc
```diff
@@ -759,6 +767,32 @@ class writer : public sstable_writer::writer_impl {
         , _sst_schema(make_sstable_schema(s, _enc_stats, _cfg))
         , _run_identifier(cfg.run_identifier)
         , _write_regular_as_static(cfg.correctly_serialize_static_compact_in_mc && s.is_static_compact_table())
+        , _large_data_stats({{
+                {
+                    large_data_type::partition_size,
+                    large_data_stats_entry{
+                        .threshold = _sst.get_large_data_handler().get_partition_threshold_bytes(),
```
raphaelsc, line 774: with designated initializer, are the other fields guaranteed to be initialized or it's possibly UB?

sstables/kl/writer.cc
```diff
@@ -662,7 +662,7 @@ void sstable_writer_k_l::consume_end_of_stream()
         features.disable(sstable_feature::NonCompoundRangeTombstones);
     }
     run_identifier identifier{_run_identifier};
-    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier));
+    _sst.write_scylla_metadata(_pc, _shard, std::move(features), std::move(identifier), {});
```
raphaelsc, line 665: we could have prepended the type std::optional<scylla_metadata::large_data_stats> to {}, to make it easier for the reader, but this shouldn't block this series

raphaelsc (approved):
had one concern regarding designated initializer, but other than that, LGTM

sstables/mx/writer.cc
raphaelsc: looks like this is defined behavior, all members are set to 0 if not explicitly initialized

bhalevy: right. it's a valid concern but they are initialized to 0.

sstables/kl/writer.cc
bhalevy: yeah, but it's a bit of an overkill and reduce readability if done inline IMO.
Maybe on the previous statement we could prepare a disengaged optional ld_stats and move it here.

sstables/mx/writer.cc
raphaelsc: thanks. it looks nice

avikivity:
Please update docs/sstable-scylla-format.md

bhalevy:
@avikivity: doc updated

docs/sstable-scylla-format.md
```diff
@@ -101,3 +105,20 @@ There are currently no defined attributes.
 If the run_identifier subcomponent is present, the sstable is part of a run.
 All sstables with the same run_identifier belong to the same run. They are
 guaranteed to be disjoint (non-overlapping) in their partition keys.
+
+## large_data_stats subcomponent
+
+    large_data_stats = large_data_count large_data_pair*
+    large_data_count = be32
+    large_data_pair = large_data_type large_data_stats_entry
+    large_data_type = be32
```
avikivity, line 114: what is large_data_type?

sstables/types.hh
```diff
@@ -509,8 +509,29 @@ struct run_identifier {
...
+// Types of large data statistics.
+//
+// Note: For extensibility, never reuse an identifier,
+// only add new ones, since these are stored on stable storage.
+enum class large_data_type : uint32_t {
+    partition_size = 0,     // partition size, in bytes
+    row_size = 1,           // row size, in bytes
+    cell_size = 2,          // cell size, in bytes
+    rows_in_partition = 3,  // number of rows in a partition
+};
```
avikivity, line 522: This should be added to the docs.

Recommend starting from 1, so that there is always a non-zero value that stands out.

avikivity:
Looks good, the last patch needs a minor fix.

avikivity:
I think min_sstable_size compounded the problem we see.

Size tiered compaction treats anything below 50MB as a small sstable. Our eager system table sstbale flushing creates a new sstable on every schema change. So 5000 tables -> 5000 memtable flushes, and 5000 new compactions.

(although the 5000 memtable flushes are bad enough and cannot be avoided)

raphaelsc:
> I think min_sstable_size compounded the problem we see.
> 
> Size tiered compaction treats anything below 50MB as a small sstable. Our eager system table sstbale flushing creates a new sstable on every schema change. So 5000 tables -> 5000 memtable flushes, and 5000 new compactions.
> 
> (although the 5000 memtable flushes are bad enough and cannot be avoided)

min_threshold is still respected, so a single new sstable <= 50MB is not enough to trigger a new compaction. min_sstable_size also helps reducing the # of compactions in the long run, because min_sstable_size means a single tier for every sst smaller than the configured min_sstable_size. Without it, we'd have additional size tiers for small-sized ssts, meaning that data in them would have to travel through all those additional tiers.

avikivity:
> > I think min_sstable_size compounded the problem we see.
> > Size tiered compaction treats anything below 50MB as a small sstable. Our eager system table sstbale flushing creates a new sstable on every schema change. So 5000 tables -> 5000 memtable flushes, and 5000 new compactions.
> > (although the 5000 memtable flushes are bad enough and cannot be avoided)
> 
> min_threshold is still respected, so a single new sstable <= 50MB is not enough to trigger a new compaction. min_sstable_size also helps reducing the # of compactions in the long run, because min_sstable_size means a single tier for every sst smaller than the configured min_sstable_size. Without it, we'd have additional size tiers for small-sized ssts, meaning that data in them would have to travel through all those additional tiers.

In fact it is the same number of compactions. Each compaction has min_threshold input sstables and one output sstable, for a net reduction of (min_threshold - 1). If we have 5000 input sstables, we need 5000/(min_threshold - 1) to reach one (or another small number) of sstables.

raphaelsc:
On Tue, Dec 1, 2020 at 10:33 AM Avi Kivity <notifications@github.com> wrote:

> I think min_sstable_size compounded the problem we see.
> Size tiered compaction treats anything below 50MB as a small sstable. Our
> eager system table sstbale flushing creates a new sstable on every schema
> change. So 5000 tables -> 5000 memtable flushes, and 5000 new compactions.
> (although the 5000 memtable flushes are bad enough and cannot be avoided)
>
> min_threshold is still respected, so a single new sstable <= 50MB is not
> enough to trigger a new compaction. min_sstable_size also helps reducing
> the # of compactions in the long run, because min_sstable_size means a
> single tier for every sst smaller than the configured min_sstable_size.
> Without it, we'd have additional size tiers for small-sized ssts, meaning
> that data in them would have to travel through all those additional tiers.
>
> In fact it is the same number of compactions. Each compaction has
> min_threshold input sstables and one output sstable, for a net reduction of
> (min_threshold - 1). If we have 5000 input sstables, we need
> 5000/(min_threshold - 1) to reach one (or another small number) of sstables.
>
You're right. Assuming a steady state, min threshold is still applied on
newly flush ssts. min_sstable_size would only help reducing WA (# of tiers
new data has to travel through) if we compact those small-sized sstables
(<= min_sstable_size) using a higher threshold, but that costs RA.

sstables/types.hh
bhalevy: fixed in v5

bhalevy:
In v5 (ca5184052):
* large_data_type enum class starts from `1`
* added individual types to docs

PR #7670, opened 2020-11-22
build: compress unified package faster

avikivity:
The unified package is quite large (1GB compressed), and it
is the last step in the build so its build time cannot be
parallized with other tasks. Compress it with pigz to take
advantage of multiple cores and speed up the build a little.

### DIFF
--- a/unified/build_unified.sh
+++ b/unified/build_unified.sh
@@ -75,4 +75,4 @@ done
 ln -f unified/install.sh build/"$MODE"/unified/
 ln -f unified/uninstall.sh build/"$MODE"/unified/
 cd build/"$MODE"/unified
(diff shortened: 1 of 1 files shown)


nyh (approved):
Looks good. scripts/create-relocatable-package.py already uses pigz for a similar purpose (see https://github.com/scylladb/scylla/issues/4194), and it's on our dependency list, so indeed we should use it.
Note that another issue I mentioned in https://github.com/scylladb/scylla/issues/4194 is still relevant -  it is kind of silly that we compress the huge relocatable package and then open the compression for the unified package (or rpm, etc.) and then compress it back again. Yes, pigz is faster. But it would have been still better not to compress back and forth.

PR #7676, opened 2020-11-23
configure.py: Use variables for product, version, and release

penberg:
The configure.py script has variables for Scylla product, version, and
release. Use them instead of reading the
SCYLLA-{PRODUCT,VERSION,RELEASE} files.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1845,7 +1845,7 @@ with open(buildfile_tmp, 'w') as f:
         build dist-server: phony dist-server-tar dist-server-rpm dist-server-deb
 
         rule build-submodule-reloc
-          command = cd $reloc_dir && ./reloc/build_reloc.sh --version $$(<../../build/SCYLLA-PRODUCT-FILE)-$$(<../../build/SCYLLA-VERSION-FILE)-$$(<../../build/SCYLLA-RELEASE-FILE) --nodeps $args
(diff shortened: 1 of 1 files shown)


configure.py
```diff
@@ -1845,7 +1845,7 @@ def configure_abseil(build_dir, mode):
         build dist-server: phony dist-server-tar dist-server-rpm dist-server-deb
 
         rule build-submodule-reloc
-          command = cd $reloc_dir && ./reloc/build_reloc.sh --version $$(<../../build/SCYLLA-PRODUCT-FILE)-$$(<../../build/SCYLLA-VERSION-FILE)-$$(<../../build/SCYLLA-RELEASE-FILE) --nodeps $args
+          command = cd $reloc_dir && ./reloc/build_reloc.sh --version {scylla_product}-{scylla_version}-{scylla_release} --nodeps $args
         rule build-submodule-rpm
```
avikivity, line 1849: Aren't they updated at different schedules?

`scylla_release` is updated at ./configure time (read from SCYLLA-PRODUCT file), and here, it's updated at run time.

I don't think either is correct, and I think that the new code is better, but I'm not sure I understand 100% what's going on.

For formal builds, there is no problem, because formal builds start with a clean workspace. But we want to make sure that a developer can have their changes to SCYLLA-VERSION-GEN picked up and used.

avikivity: (without special precautions on their part)

penberg:
Closing this as inactive.

PR #7678, opened 2020-11-23
sstables: add `may_have_partition_tombstones` method

kbr-:
For sstable versions greater or equal than md, the `min_max_column_names`
sstable metadata gives a range of position-in-partitions such that all
clustering rows stored in this sstable have positions in this range.

Partition tombstones in this context are understood as covering the
entire range of clustering keys; thus, if the sstable contains at least
one partition tombstone, the sstable position range is set to be the
range of all clustered rows.

Therefore, by checking that the position range is *not* the range of all
clustered rows we know that the sstable cannot have any partition tombstones.

Manually tested that it works with Cassandra md sstables as well.

Refs: https://github.com/scylladb/scylla/pull/7646, https://github.com/scylladb/scylla/pull/7437

### DIFF
--- a/position_in_partition.hh
+++ b/position_in_partition.hh
@@ -593,6 +593,7 @@ public:
     position_in_partition&& end() && { return std::move(_end); }
     bool contains(const schema& s, position_in_partition_view pos) const;
     bool overlaps(const schema& s, position_in_partition_view start, position_in_partition_view end) const;
+    bool is_all_clustered_rows(const schema&) const;
 
     friend std::ostream& operator<<(std::ostream&, const position_range&);
 };
@@ -609,4 +610,9 @@ inline
 bool position_range::overlaps(const schema& s, position_in_partition_view start, position_in_partition_view end) const {
     position_in_partition::less_compare less(s);
     return !less(end, _start) && less(start, _end);
+}
+
+inline
+bool position_range::is_all_clustered_rows(const schema& s) const {
+    return _start.is_before_all_clustered_rows(s) && _end.is_after_all_clustered_rows(s);
 }
--- a/sstables/sstables.hh
+++ b/sstables/sstables.hh
@@ -797,6 +797,12 @@ public:
     // Return true if this sstable possibly stores clustering row(s) specified by ranges.
     bool may_contain_rows(const query::clustering_row_ranges& ranges) const;
 
+    // false => there are no partition tombstones, true => we don't know
+    bool may_have_partition_tombstones() const {
+        return !has_correct_min_max_column_names()
+            || _position_range.is_all_clustered_rows(*_schema);
+    }
+
     // Allow the test cases from sstable_test.cc to test private methods. We use
     // a placeholder to avoid cluttering this class too much. The sstable_test class
     // will then re-export as public every method it needs.
--- a/test/boost/sstable_datafile_test.cc
+++ b/test/boost/sstable_datafile_test.cc
@@ -6685,3 +6685,43 @@ SEASTAR_TEST_CASE(test_zero_estimated_partitions) {
         return make_ready_future<>();
     });
 }
+
+SEASTAR_TEST_CASE(test_may_have_partition_tombstones) {
+    return test_env::do_with_async([] (test_env& env) {
+        storage_service_for_tests ssft;
+        simple_schema ss;
+        auto s = ss.schema();
+        auto pks = ss.make_pkeys(2);
+
+        auto tmp = tmpdir();
+        unsigned gen = 0;
+        for (auto version : all_sstable_versions) {
+            if (version < sstable_version_types::md) {
+                continue;
+            }
+
+            auto mut1 = mutation(s, pks[0]);
+            auto mut2 = mutation(s, pks[1]);
+            mut1.partition().apply_insert(*s, ss.make_ckey(0), ss.new_timestamp());
+            mut1.partition().apply_delete(*s, ss.make_ckey(1), ss.new_tombstone());
+            ss.add_row(mut1, ss.make_ckey(2), "val");
+            ss.delete_range(mut1, query::clustering_range::make({ss.make_ckey(3)}, {ss.make_ckey(5)}));
+            ss.add_row(mut2, ss.make_ckey(6), "val");
+
+            auto sst_gen = [&env, s, &tmp, &gen, version] () {
+                return env.make_sstable(s, tmp.path().string(), ++gen, version, big);
+            };
+
+            {
+                auto sst = make_sstable_containing(sst_gen, {mut1, mut2});
+                sst->load().get();
+                BOOST_REQUIRE(!sst->may_have_partition_tombstones());
+            }
+
+            mut2.partition().apply(ss.new_tombstone());
+            auto sst = make_sstable_containing(sst_gen, {mut1, mut2});
+            sst->load().get();
+            BOOST_REQUIRE(sst->may_have_partition_tombstones());
+        }
+    });
+}


position_in_partition.hh
```diff
@@ -593,6 +593,7 @@ public:
     position_in_partition&& end() && { return std::move(_end); }
     bool contains(const schema& s, position_in_partition_view pos) const;
     bool overlaps(const schema& s, position_in_partition_view start, position_in_partition_view end) const;
+    bool is_full(const schema&) const;
```
bhalevy, line 596: The name is misleading since it means a different thing from the `full()` method that returns a range including the static row.  The method you want should correspond to `all_clustered_rows()`, so maybe call it `is_all_clustered_rows()`

bhalevy (changes requested):
Other than the naming issue LGTM

position_in_partition.hh
kbr-: Good point, fixed.

kbr-:
v2: renamed `is_full` into `is_all_clustered_rows`
see diff: https://github.com/scylladb/scylla/compare/07eb77aad4ef5efa092e84e153cb1393650dd3c1..21fb0a950baeef94953ba10c6fd097f36e10d1fd

avikivity:
Do we need it? I thought the combining reader will naturally work with min/max clustering keys if available.

The algorithm would be to add sstables as soon as their [min_clustering_key, max_clustering_key] intersects the cursor. For sstables with partition tombstones, that's (-∞, +∞) and always intersects the cursor, so such an sstable is added immediately.

position_in_partition.hh
kbr-: But now I wonder if the sstable's `_position_range` should not include static rows as well... e.g. if the schema has static columns we would not set it to `all_clustered_rows()` (by default) but to `full()`, and in `set_position_range()` the left end would be `for_static_row`.

Right now the semantics of `_position_range` is unclear. It doesn't cover all possible mutation fragments from the sstable but only those that have "positions" (so clustered rows, range tombstones) *and* it special-cases the partition tombstone...

kbr-:
@avikivity 
> Do we need it? I thought the combining reader will naturally work with min/max clustering keys if available.
> 
> The algorithm would be to add sstables as soon as their [min_clustering_key, max_clustering_key] intersects the cursor. For sstables with partition tombstones, that's (-∞, +∞) and always intersects the cursor, so such an sstable is added immediately.

Ahh, that's what you meant in the other comment (https://github.com/scylladb/scylla/pull/7646#discussion_r528328640). I misunderstood it at first.

The current implementation of the new twcs-merger does not work with partition tombstones, even if all readers with partition tombstones were included first. The reason is that I only take the `partition_start` fragment returned by the first opened reader and ignore the rest. It makes the implementation simpler and a bit more efficient.

I could modify it so it handles this case. Would have to add some additional state to the merger that remembers that the initial batch of partition starts was already returned so new partition starts (returned by readers opened in the middle) should be ignored. There's a chance it would introduce some overhead but it's probably insignificant.

However, I don't see much point in doing this. Normal TWCS usage assumes there are no partition tombstones, and the new merger is aimed at this use case. And if there are partition tombstones and we have to open many sstables in parallel because of that, the optimization loses its purpose and becomes a pessimization because the new merger is less efficient CPU-wise at least according to the microbenchmark I wrote (I'm trying to optimize it a bit, but I can't guarantee it will become better eventually). Its upside is reduced memory consumption but that works only when the ranges are mostly disjoint which is not the case in presence of partition tombstones, in which case it's better to use the old merger.

avikivity:
> @avikivity
> 
> > Do we need it? I thought the combining reader will naturally work with min/max clustering keys if available.
> > The algorithm would be to add sstables as soon as their [min_clustering_key, max_clustering_key] intersects the cursor. For sstables with partition tombstones, that's (-∞, +∞) and always intersects the cursor, so such an sstable is added immediately.
> 
> Ahh, that's what you meant in the other comment ([#7646 (comment)](https://github.com/scylladb/scylla/pull/7646#discussion_r528328640)). I misunderstood it at first.
> 
> The current implementation of the new twcs-merger does not work with partition tombstones, even if all readers with partition tombstones were included first. The reason is that I only take the `partition_start` fragment returned by the first opened reader and ignore the rest. It makes the implementation simpler and a bit more efficient.
> 
> I could modify it so it handles this case. Would have to add some additional state to the merger that remembers that the initial batch of partition starts was already returned so new partition starts (returned by readers opened in the middle) should be ignored. There's a chance it would introduce some overhead but it's probably insignificant.
> 
> However, I don't see much point in doing this. Normal TWCS usage assumes there are no partition tombstones, and the new merger is aimed at this use case. And if there are partition tombstones and we have to open many sstables in parallel because of that, the optimization loses its purpose and becomes a pessimization because the new merger is less efficient CPU-wise at least according to the microbenchmark I wrote (I'm trying to optimize it a bit, but I can't guarantee it will become better eventually). Its upside is reduced memory consumption but that works only when the ranges are mostly disjoint which is not the case in presence of partition tombstones, in which case it's better to use the old generic merger.

I think merging the partition_start from multiple sstables is more regular, and avoids an extra step that has to be explained (plus you have to explain why you ignore the partition tombstone).

I don't think it will have any effect on performance. partition_start is very rare in these workloads, because the partitions are large.

As to working when the partition tombstone actually exists, we don't have to optimize for it (same for range tombstones in this scenario).

kbr-:
> I think merging the partition_start from multiple sstables is more regular, and avoids an extra step that has to be explained

Which extra step? Note that I still have to skip partition_starts from sstables that are opened later.

> I don't think it will have any effect on performance. partition_start is very rare in these workloads, because the partitions are large.

The problem is that I would have to update state (e.g. set a flag) on the first **non**-partition-start fragment or the first returned batch (the state says that from now on, new partition starts should be ignored). Which means I would have to perform a check or unconditionally update state on every non-partition-start fragment/returned batch.

Consider the following scenario. We're merging r1, r2, and r3
1. r1 and r2 are opened
2. r1 returns partition start
3. r2 returns partition start
4. merger returns a batch of these two partition starts
5. r1 returns a clustered row
6. r3 is opened, r3 returns a partition start

at step 6 I have to know that the partition start returned from r3 should be ignored, therefore at step 5 I had to update some state that says "we're no longer gathering partition starts" so I could use this state at step 6. Therefore at every later fragment/batch, processing of which goes through the same code path, I would have to do something with this state as well.

But if I return only the first partition start, the only state updates/checks are performed on partition starts, which are rare as you said (the latest version of the merger which I haven't pushed yet currently only makes checks/state updates on `partition_start` fragments).

I presume this overhead would not really be significant. I'd rather just not do the additional work of adding more logic to the merger.

> As to working when the partition tombstone actually exists, we don't have to optimize for it (same for range tombstones in this scenario).

So the solution of using the old merger in presence of partition tombstones is as good as any, right? Hence we can just call `may_have_partition_tombstones` and only use the new merger when it returns false.

This PR is cheap to merge (it doesn't extend sstables with new metadata or anything).

avikivity:
> > I think merging the partition_start from multiple sstables is more regular, and avoids an extra step that has to be explained
> 
> Which extra step? Note that I still have to skip partition_starts from sstables that are opened later.

The extra step of calling may_have_partition_tombstones(). But I see the merger is already somewhat irregular due to having to skip the partition_start, so maybe we're not adding much irregularity.


> 
> > I don't think it will have any effect on performance. partition_start is very rare in these workloads, because the partitions are large.
> 
> The problem is that I would have to update state (e.g. set a flag) on the first **non**-partition-start fragment or the first returned batch (the state says that from now on, new partition starts should be ignored). Which means I would have to perform a check or unconditionally update state on every non-partition-start fragment/returned batch.
> 
> Consider the following scenario. We're merging r1, r2, and r3
> 
> 1. r1 and r2 are opened
> 2. r1 returns partition start
> 3. r2 returns partition start
> 4. merger returns a batch of these two partition starts
> 5. r1 returns a clustered row
> 6. r3 is opened, r3 returns a partition start
> 
> at step 6 I have to know that the partition start returned from r3 should be ignored, therefore at step 5 I had to update some state that says "we're no longer gathering partition starts" so I could use this state at step 6. Therefore at every later fragment/batch, processing of which goes through the same code path, I would have to do something with this state as well.
> 

Thanks, well explained.

> But if I return only the first partition start, the only state updates/checks are performed on partition starts, which are rare as you said (the latest version of the merger which I haven't pushed yet currently only makes checks/state updates on `partition_start` fragments).
> 
> I presume this overhead would not really be significant. I'd rather just not do the additional work of adding more logic to the merger.
> 
> > As to working when the partition tombstone actually exists, we don't have to optimize for it (same for range tombstones in this scenario).
> 
> So the solution of using the old merger in presence of partition tombstones is as good as any, right? Hence we can just call `may_have_partition_tombstones` and only use the new merger when it returns false.
> 
> This PR is cheap to merge (it doesn't extend sstables with new metadata or anything).

Yes, I was trying to make the other PR more regular, using interval arithmetic on the clustering ranges instead of special cases. But I see it is not possible.

PR #7679, opened 2020-11-23
flat_mutation_reader: document assumption about fast_forward_to

kbr-:
It is not legal to fast forward a reader before it enters a partition.
One must ensure that there even is a partition in the first place. For
this one must fetch a `partition_start` fragment.

This PR is a follow-up to a discussion on Slack.

### DIFF
--- a/flat_mutation_reader.hh
+++ b/flat_mutation_reader.hh
@@ -433,6 +433,11 @@ public:
     //
     // When forwarding mode is not enabled, fast_forward_to()
     // cannot be used.
+    //
+    // `fast_forward_to` can be called only when the reader is within a partition
+    // and it affects the set of fragments returned from that partition.
+    // In particular one must first enter a partition by fetching a `partition_start`
+    // fragment before calling `fast_forward_to`.
     future<> fast_forward_to(position_range cr, db::timeout_clock::time_point timeout) {
         return _impl->fast_forward_to(std::move(cr), timeout);
     }


flat_mutation_reader.hh
```diff
@@ -433,6 +433,10 @@ public:
     //
     // When forwarding mode is not enabled, fast_forward_to()
     // cannot be used.
+    //
+    // `partition_start` must be fetched *before* the first call to `fast_forward_to`.
```
tgrabiec, line 437: Also, it must not be called after the last `partition_end`. I would make this statement more general. It can be called only when the reader is within a partition, and it affects that partition's range. A reader is inside a partition between it produces `partition_start` and `partition_end`.

tgrabiec (changes requested):

flat_mutation_reader.hh
haaawk: That's not exactly true, is it? At some point partition_end was returned when the current ck range was exhausted and then it was ok to call f_f_t again. Unless I remember incorrectly?

haaawk: Or was it just the reader::end_of_stream() returning true?

haaawk: But I agree with the general statement that f_f_t can be called only when the reader is within a partition.

kbr-: But in forwarding mode readers don't return partition_ends

kbr-: @tgrabiec reworded it a bit, check it out

haaawk: Yes. So it's safer to just write within a partition.

tgrabiec (approved):

kbr-:
Thanks.

haaawk (approved):

PR #7682, opened 2020-11-24
repair: Use same description for the same metric

### LINKED ISSUE #7681: repair: Use same description for the same metric
In commit 9b28162f88121938219955f05cced490840764ed (repair: Use label for node ops metrics), we switched to use label for different node operations. We should use the same description for the same metric name.

asias:
In commit 9b28162f88121938219955f05cced490840764ed (repair: Use label
for node ops metrics), we switched to use label for different node
operations. We should use the same description for the same metric name.

Fixes #7681

### DIFF
--- a/repair/repair.cc
+++ b/repair/repair.cc
@@ -60,17 +60,17 @@ public:
         auto ops_label_type = sm::label("ops");
         _metrics.add_group("node_ops", {
             sm::make_gauge("finished_percentage", [this] { return bootstrap_finished_percentage(); },
-                    sm::description("Number of finished percentage for bootstrap operation on this shard."), {ops_label_type("bootstrap")}),
+                    sm::description("Finished percentage of node operation on this shard"), {ops_label_type("bootstrap")}),
             sm::make_gauge("finished_percentage", [this] { return replace_finished_percentage(); },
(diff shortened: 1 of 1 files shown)


repair/repair.cc
```diff
@@ -60,17 +60,17 @@ class node_ops_metrics {
...
             sm::make_gauge("finished_percentage", [this] { return decommission_finished_percentage(); },
-                    sm::description("Number of finished percentage for decommission operation on this shard."), {ops_label_type("decommission")}),
+                    sm::description("Finished percentage of node operation on this shard"), {ops_label_type("decommission")}),
             sm::make_gauge("finished_percentage", [this] { return removenode_finished_percentage(); },
-                    sm::description("Number of finished percentage for removenode operation on this shard."), {ops_label_type("removenode")}),
+                    sm::description("Finished percentage of node operation on this shard"), {ops_label_type("removenode")}),
             sm::make_gauge("finished_percentage", [this] { return repair_finished_percentage(); },
-                    sm::description("Number of finished percentage for repair operation on this shard."), {ops_label_type("repair")}),
+                    sm::description("Finished percentage of node operation on this shard"), {ops_label_type("repair")}),
```
avikivity, line 73: You should convert to a loop!

asias: Construct a map (lable -> function) and run the loop?

PR #7683, opened 2020-11-24
Update version.hh

fuliang-jlu:
add const reference for the version operator overloading

### DIFF
--- a/version.hh
+++ b/version.hh
@@ -43,24 +43,24 @@ public:
         return v;
     }
 
-    bool operator==(version v) const {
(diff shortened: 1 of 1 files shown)


avikivity:
Note you have to accept the contributor's agreement before we can accept patches. See https://www.scylladb.com/open-source-nosql-database/contributor-agreement/.

You didn't explain why the change is an improvement.

mykaul:
Closing - please re-open (after resolving conflict, explaining the motivation, etc.) if it's still relevant.

PR #7685, opened 2020-11-24
scylla_util.py: fix metadata gcp call for disks to get details

### LINKED ISSUE #7684: current master and 4.3 don't properly recognize disk count from metadata

*Installation details*
Scylla version (or git commit hash): 4.3 rc0 + rc1 and master
Cluster size:

Disks: (SSD/HDD, count)

when you run io setup you see an error
```
Problem when parsing disks from metadata:
Extra data: line 1 column 2 (char 1)
ERROR:root:This is not a recommended Google Cloud instance setup for auto tuning, running manual iotune.
```

fix is easy, https://github.com/scylladb/scylla/blob/master/dist/common/scripts/scylla_util.py#L190 should use "True" to get recursive list of objects for parsing for nvme disks

tarzanek:
fixes #7684

### DIFF
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -187,7 +187,7 @@ class gcp_instance:
         """get list of nvme disks from metadata server"""
         import json
         try:
-            disksREST=self.__instance_metadata("disks")
+            disksREST=self.__instance_metadata("disks", True)
             disksobj=json.loads(disksREST)
             nvmedisks=list(filter(self.isNVME, disksobj))
         except Exception as e:


tarzanek:
this is a backport candidate for 4.3 (where this bug is too)

avikivity:
What was the problem? And why doesn't the changelog explain?

tarzanek:
fixed the comment to explain the regression on the metadata REST call

tarzanek:
btw. without this patch the GCP disk parsing and detection of recommended VM is useless, so we should get it to master and 4.3 asap

Issue #3552, opened 2018-06-26
Partition can start to appear empty until restart with DateTiered or TimeWindow compaction strategies


tgrabiec:
*Installation details*
Scylla version (or git commit hash): 2.1+

There is a bad interaction between cache population on read and sstable read optimization which filters out sstables based on clustering key component ranges recorded in the sstables (only enabled for DateTiered and TimeWindow compaction strategies). When the clustering range of the query doesn't overlap with any sstable, the reader will return no partition markers for the read, which leads cache populator to assume there is no partition in sstables and it will cache an empty partition. This will cause later reads of that partition to miss prior writes to that partition until it is evicted from cache or node is restarted.

PR #7688, opened 2020-11-24
mutation_reader: introduce clustering_order_reader_merger

kbr-:
This abstraction is used to merge the output of multiple readers, each
opened for a single partition query, into a non-decreasing stream
of mutation_fragments.

It is similar to `mutation_reader_merger`,
but an important difference is that the new merger may select new readers
in the middle of a partition after it already returned some fragments
from that partition.  It uses the new `position_reader_queue` abstraction
to select new readers. It doesn't support multi-partition (ring range) queries.

The new merger will be later used when reading from sstable sets created
by TimeWindowCompactionStrategy. This strategy creates many sstables
that are mostly disjoint w.r.t the contained clustering keys, so we can
delay opening sstable readers when querying a partition until after we have
processed all mutation fragments with positions before the keys
contained by these sstables.

A microbenchmark was added that compares the existing combining reader
(which uses `mutation_reader_merger` underneath) with a new combining reader
built using the new `clustering_order_reader_merger` and a simple queue of readers
that returns readers from some supplied set. The used set of readers is built from the following
ranges of keys (each range corresponds to a single reader):
`[0, 31]`, `[30, 61]`, `[60, 91]`, `[90, 121]`, `[120, 151]`.
The microbenchmark runs the reader and divides the result by the number of mutation fragments.
The results on my laptop were:
```
$ build/release/test/perf/perf_mutation_readers -t clustering_combined.* -r 10
single run iterations:    0
single run duration:      1.000s
number of runs:           10

test                                      iterations      median         mad         min         max
clustering_combined.ranges_generic           2911678   117.598ns     0.685ns   116.175ns   119.482ns
clustering_combined.ranges_specialized       3005618   111.015ns     0.349ns   110.063ns   111.840ns
```
`ranges_generic` denotes the existing combining reader, `ranges_specialized` denotes the new reader.

Split from https://github.com/scylladb/scylla/pull/7437.

### DIFF
--- a/mutation_reader.cc
+++ b/mutation_reader.cc
@@ -2294,3 +2294,393 @@ flat_mutation_reader make_compacting_reader(flat_mutation_reader source, gc_cloc
         std::function<api::timestamp_type(const dht::decorated_key&)> get_max_purgeable) {
     return make_flat_mutation_reader<compacting_reader>(std::move(source), compaction_time, get_max_purgeable);
 }
+
+position_reader_queue::~position_reader_queue() {}
+
+// Merges output of readers opened for a single partition query into a non-decreasing stream of mutation fragments.
+//
+// Uses `position_reader_queue` to retrieve new readers lazily as the read progresses through the partition.
+// A reader is popped from the queue only if we find that it may contain fragments for the currently inspected positions.
+//
+// Readers are closed as soon as we find that they were exhausted for the given partition query.
+//
+// Implements the `FragmentProducer` concept. However, `next_partition` and `fast_forward_to(partition_range)`
+// are not implemented and throw an error; the reader is only used for single partition queries.
+//
+// Assumes that:
+// - the queue contains at least one reader,
+// - there are no static rows,
+// - the returned fragments do not contain partition tombstones.
+class clustering_order_reader_merger {
+    const schema_ptr _schema;
+    const reader_permit _permit;
+
+    // Compares positions using *_schema.
+    const position_in_partition::tri_compare _cmp;
+
+    // A queue of readers used to lazily retrieve new readers as we progress through the partition.
+    // Before the merger returns a batch for position `p`, it first ensures that all readers containing positions
+    // <= `p` are popped from the queue so it can take all of their fragments into account.
+    std::unique_ptr<position_reader_queue> _reader_queue;
+
+    // Owning container for the readers popped from _reader_queue.
+    // If we are sure that a reader is exhausted (all rows from the queried partition have been returned),
+    // we destroy and remove it from the container.
+    std::list<reader_and_upper_bound> _all_readers;
+    using reader_iterator = std::list<reader_and_upper_bound>::iterator;
+
+    // A min-heap of readers, sorted by the positions of their next fragments.
+    // The iterators point to _all_readers.
+    // Invariant: every reader in `_peeked_readers` satisfies `!is_buffer_empty()`,
+    // so it is safe to call `pop_mutation_fragment()` and `peek_buffer()` on it.
+    merger_vector<reader_iterator> _peeked_readers;
+
+    // Used to compare peeked_readers stored in the `_peeked_readers` min-heap.
+    struct peeked_reader_cmp {
+        const position_in_partition::less_compare _less;
+
+        explicit peeked_reader_cmp(const schema& s) : _less(s) {}
+
+        bool operator()(const reader_iterator& a, const reader_iterator& b) {
+            // Boost heaps are max-heaps, but we want a min-heap, so invert the comparison.
+            return _less(b->reader.peek_buffer().position(), a->reader.peek_buffer().position());
+        }
+    };
+
+    const peeked_reader_cmp _peeked_cmp;
+
+    // operator() returns a mutation_fragment_batch, which is a range (a pair of iterators);
+    // this is where the actual data is stored, i.e. the range points to _current_batch.
+    merger_vector<mutation_fragment> _current_batch;
+
+    // _unpeeked_readers stores readers for which we don't know the next fragment that they'll return.
+    // Before we return the next batch of fragments, we must peek all readers here (and move them to
+    // the _peeked_readers heap), since they might contain fragments with smaller positions than the
+    // currently peeked readers.
+    merger_vector<reader_iterator> _unpeeked_readers;
+
+    // In forwarding mode, after a reader returns end-of-stream, if we cannot determine that
+    // the reader won't return any more fragments in later position ranges, we save it in
+    // _halted_readers and restore it when we get fast-forwaded to a later range.
+    // See also comment in `peek_reader` when a reader returns end-of-stream.
+    // _halted_readers doesn't serve any purpose when not in forwarding mode, because then
+    // readers always return end-of-partition before end-of-stream, which is a signal that
+    // we can remove the reader immediately.
+    merger_vector<reader_iterator> _halted_readers;
+
+    // In forwarding mode, this is the right-end of the position range being currently queried;
+    // initially it's set to `before_all_clustered_rows` and updated on `fast_forward_to`.
+    // We use it when popping readers from _reader_queue so that we don't prematurely pop
+    // readers that only contain fragments from greater ranges.
+    // In non-forwarding mode _pr_end is always equal to `after_all_clustered_rows`.
+    position_in_partition_view _pr_end;
+    // In forwarding mode, _forwarded_to remembers the last range we were forwarded to.
+    // We need this because we're opening new readers in the middle of the partition query:
+    // after the new reader returns its initial partition-start, we immediately forward it
+    // to this range.
+    std::optional<position_range> _forwarded_to;
+
+    // Since we may open new readers when already inside the partition, i.e. after returning `partition_start`,
+    // we must ignore `partition_start`s returned by these new readers. The approach we take is to return
+    // the `partition_start` fetched from the first reader and ignore all the rest. This flag says whether
+    // or not we've already fetched the first `partition_start`.
+    bool _partition_start_fetched = false;
+
+    // In non-forwarding mode, remember if we've returned the last fragment, which is always partition-end.
+    // We construct the fragment ourselves instead of merging partition-ends returned from the merged readers,
+    // because we may close readers in the middle of the partition query.
+    // In forwarding mode this is always false.
+    bool _should_emit_partition_end;
+
+    // If a single reader wins with other readers (i.e. returns a smaller fragment) multiple times in a row,
+    // the reader becomes a ``galloping reader'' (and is pointed to by _galloping_reader).
+    // In this galloping mode we stop doing heap operations using the _peeked_readers heap;
+    // instead, we keep peeking the _galloping_reader and compare the returned fragment's position directly
+    // with the fragment of the reader stored at the heap front (if any), hoping that the galloping reader
+    // will keep winning. If he wins, we don't put the fragment on the heap, but immediately return it.
+    // If he loses, we go back to normal operation.
+    reader_iterator _galloping_reader;
+
+    // Counts how many times a potential galloping reader candidate has won with other readers.
+    int _gallop_mode_hits = 0;
+
+    // Determines how many times a fragment should be taken from the same
+    // reader in order to enter gallop mode. Must be greater than one.
+    static constexpr int _gallop_mode_entering_threshold = 3;
+
+    bool in_gallop_mode() const {
+        return _gallop_mode_hits >= _gallop_mode_entering_threshold;
+    }
+
+    // Retrieve the next fragment from the reader pointed to by `it`.
+    // The function assumes that we're not in galloping mode, `it` is in `_unpeeked_readers`,
+    // and all fragments previously returned from the reader have already been returned by operator().
+    //
+    // The peeked reader is pushed onto the _peeked_readers heap.
+    future<> peek_reader(reader_iterator it, db::timeout_clock::time_point timeout) {
+        return it->reader.peek(timeout).then([this, timeout, it] (mutation_fragment* mf) {
+            if (!mf) {
+                // The reader returned end-of-stream before returning end-of-partition
+                // (otherwise we would have removed it in a previous peek). This means that
+                // we are in forwarding mode and the reader won't return any more fragments in the current range.
+                // If the reader's upper bound is smaller then the end of the current range then it won't
+                // return any more fragments in later ranges as well (subsequent fast-forward-to ranges
+                // are non-overlapping and strictly increasing), so we can remove it now.
+                // Otherwise it may start returning fragments later, so we save it for the moment
+                // in _halted_readers and will bring it back when we get fast-forwarded.
+                if (_cmp(it->upper_bound, _pr_end) < 0) {
+                    _all_readers.erase(it);
+                } else {
+                    _halted_readers.push_back(it);
+                }
+                return make_ready_future<>();
+            }
+
+            if (mf->is_partition_start()) {
+                // We assume there are no partition tombstones.
+                // This should have been checked before opening the reader.
+                if (mf->as_partition_start().partition_tombstone()) {
+                    on_internal_error(mrlog, format(
+                            "clustering_order_reader_merger: partition tombstone encountered for partition {}."
+                            " This reader merger cannot be used for readers that return partition tombstones"
+                            " or it would give incorrect results.", mf->as_partition_start().key()));
+                }
+                if (!_partition_start_fetched) {
+                    _peeked_readers.emplace_back(it);
+                    boost::range::push_heap(_peeked_readers, _peeked_cmp);
+                    _partition_start_fetched = true;
+                    // there is no _forwarded_to range yet (see `fast_forward_to`)
+                    // so no need to forward this reader
+                    return make_ready_future<>();
+                }
+
+                it->reader.pop_mutation_fragment();
+                auto f = _forwarded_to ? it->reader.fast_forward_to(*_forwarded_to, timeout) : make_ready_future<>();
+                return f.then([this, timeout, it] { return peek_reader(it, timeout); });
+            }
+
+            // We assume that the schema does not have any static columns, so there cannot be any static rows.
+            if (mf->is_static_row()) {
+                on_internal_error(mrlog,
+                        "clustering_order_reader_merger: static row encountered."
+                        " This reader merger cannot be used for readers that return static rows"
+                        " or it would give incorrect results.");
+            }
+
+            if (mf->is_end_of_partition()) {
+                _all_readers.erase(it);
+            } else {
+                _peeked_readers.emplace_back(it);
+                boost::range::push_heap(_peeked_readers, _peeked_cmp);
+            }
+
+            return make_ready_future<>();
+        });
+    }
+
+    future<> peek_readers(db::timeout_clock::time_point timeout) {
+        return parallel_for_each(_unpeeked_readers, [this, timeout] (reader_iterator it) {
+            return peek_reader(it, timeout);
+        }).then([this] {
+            _unpeeked_readers.clear();
+        });
+    }
+
+    // Retrieve the next fragment from the galloping reader.
+    // The function assumes that we're in galloping mode and all fragments previously returned
+    // from the galloping reader have already been returned by operator().
+    //
+    // If the galloping reader wins with other readers again, the fragment is returned as the next batch.
+    // Otherwise, the reader is pushed onto _peeked_readers and we retry in non-galloping mode.
+    future<mutation_fragment_batch> peek_galloping_reader(db::timeout_clock::time_point timeout) {
+        return _galloping_reader->reader.peek(timeout).then([this, timeout] (mutation_fragment* mf) {
+            if (mf) {
+                if (mf->is_partition_start()) {
+                    on_internal_error(mrlog, format(
+                            "clustering_order_reader_merger: double `partition start' encountered"
+                            " in partition {} during read.", mf->as_partition_start().key()));
+                }
+
+                if (mf->is_static_row()) {
+                    on_internal_error(mrlog,
+                            "clustering_order_reader_merger: static row encountered."
+                            " This reader merger cannot be used for tables that have static columns"
+                            " or it would give incorrect results.");
+                }
+
+                if (mf->is_end_of_partition()) {
+                    _all_readers.erase(_galloping_reader);
+                } else {
+                    if (_reader_queue->empty(mf->position())
+                            && (_peeked_readers.empty()
+                                    || _cmp(mf->position(), _peeked_readers.front()->reader.peek_buffer().position()) < 0)) {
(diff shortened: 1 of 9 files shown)


avikivity:
I reviewed the cover letter and it already makes me want to commit, but I will let @denesb review the contents too.

denesb:
I'm going to review this today, but first can you explain how this relates to the original https://github.com/scylladb/scylla/pull/7437? Should I review that too? Or just this? What will happen to #7437?

kbr-:
It's a subset of #7437, I wanted to split it into smaller PRs. I'll rebase #7437 on top of master after this one gets merged.

kbr-:
There are some differences compared to the initial version you reviewed - the merger is now in mutation_reader.cc and uses a queue interface, so I think you'll like it better

denesb:
Ah, you are trying to sweet-talk me into reviewing, aren't you.

denesb:
> $ build/release/test/perf/perf_mutation_readers -t clustering_combined.* -r 10
> single run iterations:    0
> single run duration:      1.000s
> number of runs:           10
> 
> test                                      iterations      median         mad         min         max
> clustering_combined.ranges_generic             19572    17.126us    29.672ns    17.059us    17.191us
> clustering_combined.ranges_specialized         20352    15.988us    66.070ns    15.886us    16.086us

What is this `mad` column and why is the new reader's so much more than the old one's? What am I missing?

kbr-:
I have no idea. This column seems to be giving random results.
E.g. I just reexecuted the test:
```
test                                      iterations      median         mad         min         max
clustering_combined.ranges_generic             18068    18.623us   455.386ns    17.989us    23.670us
clustering_combined.ranges_specialized         17821    16.906us   126.091ns    16.780us    19.016us
```

denesb:
Maybe it has gone mad.

avikivity:
mad = mean absolute deviation. So it's both mean and mad.

mutation_reader.cc
```diff
@@ -2294,3 +2294,432 @@ flat_mutation_reader make_compacting_reader(flat_mutation_reader source, gc_cloc
...
+        , _halted_reader_cmp(*_schema)
+        , _pr_end(fwd_sm == streamed_mutation::forwarding::yes
+                        ? position_in_partition_view::before_all_clustered_rows()
+                        : position_in_partition_view::after_all_clustered_rows())
+        , _should_return_partition_end(fwd_sm == streamed_mutation::forwarding::no)
+    {
+        // The first call to `_reader_queue::pop` uses `after_all_clustered_rows`
+        // so we obtain at least one reader; we will return this reader's `partition_start`
+        // as the first fragment.
+        auto rs = _reader_queue->pop(position_in_partition_view::after_all_clustered_rows());
```
denesb, line 2593: This is confusing. A better interface would be to make the position parameter optional, when missing the next batch of readers are returned.

mutation_reader.cc
```diff
@@ -2294,3 +2294,432 @@ flat_mutation_reader make_compacting_reader(flat_mutation_reader source, gc_cloc
...
+        if (rs.empty()) {
+            on_internal_error(mrlog,
+                    "clustering_order_reader_merger: no readers available."
+                    " This merger must be given at least one reader to work correctly.");
+        }
```
denesb, line 2598: I don't think this warrants an `on_internal_error()`. Just set EOS and done.

denesb: Or `operator()()` can't handle no readers? This condition will also happen when reaching EOS genuinely. Callers are allowed (although it doesn't make sense) to call `fill_buffer()` on an EOS'd reader too.

test/lib/mutation_source_test.cc
```diff
@@ -2332,3 +2332,82 @@ void for_each_schema_change(std::function<void(schema_ptr, const std::vector<mut
...
+
+        trace.push_back(std::move(expected));
+    }
+
+    auto mf = tested(db::no_timeout).get();
+    if (mf) {
+        BOOST_FAIL(format("compare_readers: expected eos, but the reader returned:\n{}.\nTrace:{}",
+                    mfo_printer(s, mf), trace_printer(s, trace)));
+    }
+    trace.push_back(std::move(mf));
```
denesb, line 2396: I think this could be much easily done with `flat_mutation_reader_assertions()`:
```
auto rd_auth = make_authority();
auto rd_tested = assert_that(make_tested());
while (auto mf_opt = rd_auth(db::no_timeout).get0(); mf_opt) {
    rd_tested.produces(*mf_opt);
}
rd_tested.produces_end_of_stream();
```

denesb (commented):
LGTM - no blockers from my POV.
I have suggested some changes and improvements but I'm fine with them being follow-ups (if we agree on them).

kbr-:
v2?:
- simplify `_halted_readers`: it is no longer a heap but a simple bag of readers that should be brought back on the next `fast_forward_to` call. There is no more `collect_exhausted_readers`. We can decide whether the reader should be removed immediately after it returns `end-of-stream` by comparing its `upper_bound` with the end of the current forwarding range.
- rename variables: `_got_partition_start` -> `_partition_start_fetched`, `_should_return_partition_end` -> `_should_emit_partition_end`
- shorten error messages
- improve comments

mutation_reader.hh
```diff
@@ -609,3 +609,65 @@ std::pair<flat_mutation_reader, queue_reader_handle> make_queue_reader(schema_pt
...
+// 2. pop(for_key(10)) returns no readers => all readers from pref(for_key(10)) have been popped
+// 3. pop(for_key(20)) returns r2 => due to the previous step we know that r2 is not in pref(for_key(10))
+// 4. the first fragment (excluding partition_start) returned by r2 has position for_key(10)
+//        => illegal, because for_key(10) is not greater than for_key(10).
+//        The first position returned by r2 must be after_key(10) or higher.
+//
+// With each reader also comes an upper bound on the set of positions of fragments that the reader will return.
+class position_reader_queue {
+public:
+    virtual ~position_reader_queue() = 0;
```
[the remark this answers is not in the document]
kbr-: I guess it's not needed, but it makes it clearer that this is an abstract interface. It doesn't hurt.

mutation_reader.hh
```diff
@@ -609,3 +609,65 @@ std::pair<flat_mutation_reader, queue_reader_handle> make_queue_reader(schema_pt
...
+// that the reader will return. The upper bound does not need to be exact.
+struct reader_and_upper_bound {
+    flat_mutation_reader reader;
+    position_in_partition upper_bound;
+
+    reader_and_upper_bound(flat_mutation_reader r, position_in_partition bound)
+        : reader(std::move(r)), upper_bound(std::move(bound)) {}
+};
+
+// A queue of mutation readers returning fragments with the same schema from a single partition.
```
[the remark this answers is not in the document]
kbr-: Done

mutation_reader.hh
```diff
@@ -609,3 +609,65 @@ std::pair<flat_mutation_reader, queue_reader_handle> make_queue_reader(schema_pt
...
+// to return some readers earlier, but they must not break property 6.
+// For example, the following scenario is illegal:
+// 1. pop(for_key(10)) returns r1
+// 2. pop(for_key(10)) returns no readers => all readers from pref(for_key(10)) have been popped
+// 3. pop(for_key(20)) returns r2 => due to the previous step we know that r2 is not in pref(for_key(10))
+// 4. the first fragment (excluding partition_start) returned by r2 has position for_key(10)
+//        => illegal, because for_key(10) is not greater than for_key(10).
+//        The first position returned by r2 must be after_key(10) or higher.
+//
+// With each reader also comes an upper bound on the set of positions of fragments that the reader will return.
```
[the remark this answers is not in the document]
kbr-: > Or is it to circumvent the false-EOS in forwarding-mode?

Precisely. I'm only using it to decide if the reader should be removed in forwarding mode.

mutation_reader.hh
```diff
@@ -609,3 +609,65 @@ std::pair<flat_mutation_reader, queue_reader_handle> make_queue_reader(schema_pt
...
+class position_reader_queue {
+public:
+    virtual ~position_reader_queue() = 0;
+
+    // `empty(b)` <=>
+    //      we have popped all readers from `pref(b)` so `pop(b)`
+    //      will not return any more readers.
+    virtual bool empty(position_in_partition_view bound) const = 0;
```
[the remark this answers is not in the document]
kbr-: Yes, I even checked whether it would improve performance, but didn't observe an improvement (or it was insignificant) in the microbenchmark, so I left it pure virtual for simplicity. But maybe I'll devirtualize in a follow-up, after the final PR is merged.

mutation_reader.cc
```diff
@@ -2294,3 +2294,432 @@ flat_mutation_reader make_compacting_reader(flat_mutation_reader source, gc_cloc
...
+    // Before we return the next batch of fragments, we must peek all readers here (and move them to
+    // the _peeked_readers heap), since they might contain fragments with smaller positions than the
+    // currently peeked readers.
+    merger_vector<reader_iterator> _unpeeked_readers;
+
+    // In forwarding mode, after a reader returns end-of-stream, we save it in _halted_readers,
+    // a min-heap sorted by the upper bounds of possible positions of fragments returned by the readers.
+    // After each returned batch of fragments we compare the position of this batch to the
+    // upper bound of the heap's front; if the bound is smaller, it means that
+    // this reader won't return any more fragments, so we can remove it.
```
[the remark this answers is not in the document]
kbr-: Great idea. Not only it makes things more efficient, it also simplifies the merger. Done.

mutation_reader.cc
```diff
@@ -2294,3 +2294,432 @@ flat_mutation_reader make_compacting_reader(flat_mutation_reader source, gc_cloc
...
+    // We need this because we're opening new readers in the middle of the partition query:
+    // after the new reader returns its initial partition-start, we immediately forward it
+    // to this range.
+    std::optional<position_range> _forwarded_to;
+
+    // Since we may open new readers when already inside the partition, i.e. after returning `partition_start`,
+    // we must ignore `partition_start`s returned by these new readers. The approach we take is to return
+    // the `partition_start` fetched from the first reader and ignore all the rest. This flag says whether
+    // or not we've already fetched the first `partition_start`.
+    bool _got_partition_start = false;
```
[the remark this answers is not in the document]
kbr-: for me `emitted` suggests that **we** emitted it (returned it from `operator()`), but that's not what the flag says; it says whether or not one of the internal readers emitted this fragment. I renamed it nevertheless, now it's called `_partition_start_fetched`, I feel it's a bit better.

mutation_reader.cc
```diff
@@ -2294,3 +2294,432 @@ flat_mutation_reader make_compacting_reader(flat_mutation_reader source, gc_cloc
...
+
+    // In non-forwarding mode, remember if we've returned the last fragment, which is always partition-end.
+    // We construct the fragment ourselves instead of merging partition-ends returned from the merged readers,
+    // because we may close readers in the middle of the partition query.
+    // In forwarding mode this is always false.
+    bool _should_return_partition_end;
```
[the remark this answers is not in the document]
kbr-: done

mutation_reader.cc
```diff
@@ -2294,3 +2294,432 @@ flat_mutation_reader make_compacting_reader(flat_mutation_reader source, gc_cloc
...
+
+            if (mf->is_partition_start()) {
+                // We assume there are no partition tombstones.
+                // This should have been checked before opening the reader.
+                if (mf->as_partition_start().partition_tombstone()) {
+                    on_internal_error(mrlog, format(
+                            "clustering_order_reader_merger: partition tombstone encountered for partition {}."
+                            " This reader merger cannot be used for readers that return partition tombstones"
+                            " or it would give incorrect results. The no-tombstone condition should have been checked"
+                            " before creating the merger.", mf->as_partition_start().key()));
```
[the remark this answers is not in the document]
kbr-: done

mutation_reader.cc
kbr-: But it would make the interface more complex, and `pop(std::nullopt)` would be the same as `pop({after_all_clustered_rows})`.

mutation_reader.cc
kbr-: I wanted to guarantee the following:
_all merged readers return `partition_start` => the merger returns `partition_start`_.
But this can only be guaranteed if we have at least one reader. The merger could also construct a `partition_start` on its own, but it would need to know the partition key (initially that's what it did, but I wanted to simplify the reader and make it more efficient so now it obtains the `partition_start` from the readers).

kbr-:
v3: simplify `compare_readers`, see diff: https://github.com/scylladb/scylla/compare/3276bb49d292b5ae57e46aad9cb605d993cad1c3..1fba265de3c834116d30ad503ceebbcdfcfcf137

test/lib/mutation_source_test.cc
kbr-: Done

kbr-:
v4: normalize the microbenchmark's results per 10 mutation fragments
See diff: https://github.com/scylladb/scylla/compare/1fba265de3c834116d30ad503ceebbcdfcfcf137..a284c90bc9e723e3934b03bb8dcf1d890a7c610c

normalized results:
```
test                                      iterations      median         mad         min         max
clustering_combined.ranges_generic            292995     1.162us     1.869ns     1.159us     1.195us
clustering_combined.ranges_specialized        305520     1.069us     1.276ns     1.066us     1.077us
```

kbr-:
@avikivity I had a hunch that I should normalize per 10 fragments instead of per 1 because the times would go under `1us`, which felt a bit too small for me (and `mad` is now at the order of `1ns`).

avikivity:
What kind of normalization is that? And what's the problem with 0.1 usec per fragment?

kbr-:
> What kind of normalization is that?

A kind that makes the result invariant of the number of fragments.
Any "normalized" value multiplied by a constant is still a "normalized" value (and I don't know if the word "normalized" can even be precisely defined; it can have many meanings).

> And what's the problem with 0.1 usec per fragment?

Probably none, I'm just afraid to go closer to the `0.001ns` boundary (what will happen then? will the perf framework go into picoseconds?); with normalization per fragment the `mad` value is slowly getting there:
```
test                                      iterations      median         mad         min         max
clustering_combined.ranges_generic           2911678   117.598ns     0.685ns   116.175ns   119.482ns
clustering_combined.ranges_specialized       3005618   111.015ns     0.349ns   110.063ns   111.840ns
```
But since `median` is nowhere close, I guess my fear is unjustified. And maybe it's unjustified regardless.

I'll normalize per fragment then.

kbr-:
v5: normalize the microbenchmark's result per mutation fragment, diff: https://github.com/scylladb/scylla/compare/a284c90bc9e723e3934b03bb8dcf1d890a7c610c..a0a0a919c87741e40d52b8b73d3f99e48231a361

avikivity:
> > What kind of normalization is that?
> 
> A kind that makes the result invariant of the number of fragments.
> Any "normalized" value multiplied by a constant is still a "normalized" value (and I don't know if the word "normalized" can even be precisely defined; it can have many meanings).
> 
> > And what's the problem with 0.1 usec per fragment?
> 
> Probably none, I'm just afraid to go closer to the `0.001ns` boundary (what will happen then? will the perf framework go into picoseconds?); with normalization per fragment the `mad` value is slowly getting there:


This is where I get on a soapbox and complain that computer science people need some physics.

A computer can on a very windy day achieve 8 instructions per nanosecond per core. In a normal day 2-3 instructions per nanosecond per core. You can't process a fragment in 8 instructions, so there is no risk of getting even close to 1ns.


0.1us per fragment is 200-300 instructions, quite nice. Maybe 400-600 instructions if running on a nice desktop (servers are slower).

> 
> ```
> test                                      iterations      median         mad         min         max
> clustering_combined.ranges_generic           2911678   117.598ns     0.685ns   116.175ns   119.482ns
> clustering_combined.ranges_specialized       3005618   111.015ns     0.349ns   110.063ns   111.840ns
> ```
> 
> But since `median` is nowhere close, I guess my fear is unjustified. And maybe it's unjustified regardless.
> 
> I'll normalize per fragment then.

mutation_reader.cc
denesb: Won't removing this check just work and the reader will be in instant EOS?

mutation_reader.cc
```diff
@@ -2294,3 +2294,401 @@ flat_mutation_reader make_compacting_reader(flat_mutation_reader source, gc_cloc
...
+            // We assume that the schema does not have any static columns, so there cannot be any static rows.
+            if (mf->is_static_row()) {
+                on_internal_error(mrlog,
+                        "clustering_order_reader_merger: static row encountered."
+                        " This reader merger cannot be used for readers that return static rows"
+                        " or it would give incorrect results.");
+            }
+
+            if (mf->is_end_of_partition()) {
+                it->reader.pop_mutation_fragment();
```
denesb, line 2471: This seems unnecessary.

test/lib/mutation_source_test.cc
```diff
@@ -2332,3 +2332,43 @@ void for_each_schema_change(std::function<void(schema_ptr, const std::vector<mut
...
+}
+
+void compare_readers(const schema& s, flat_mutation_reader authority, flat_mutation_reader tested, const std::vector<position_range>& fwd_ranges) {
+    auto assertions = assert_that(std::move(tested));
+    compare_readers(s, authority, assertions);
+    for (auto& r: fwd_ranges) {
+        authority.fast_forward_to(r, db::no_timeout);
+        assertions.fast_forward_to(r);
+        compare_readers(s, authority, assertions);
+    }
```
denesb, line 2373: I don't understand this test. Once you drive the readers to EOS, the following fast-forwarding-loop is essentially NOOP.

test/boost/mutation_reader_test.cc
```diff
@@ -3520,3 +3521,139 @@ SEASTAR_THREAD_TEST_CASE(test_evictable_reader_self_validation) {
...
+        {
+            auto fwd = streamed_mutation::forwarding::no;
+            compare_readers(*g._s, make_authority(merged, fwd), make_tested(scenario.readers_data, fwd));
+        }
+
+        auto fwd = streamed_mutation::forwarding::yes;
+        compare_readers(*g._s, make_authority(std::move(merged), fwd),
+                make_tested(std::move(scenario.readers_data), fwd), scenario.fwd_ranges);
+    }
+}
```
denesb, line 3659: I would also like to see a test with `run_mutation_source_tests()`. To work-around the single partition limitation, you can write a meta reader which constructs a new merging reader on every new partition.
For any mutation that has a static row or partition tombstone you can just bail-out and create `flat_mutation_reader_from_mutations()` instead. For increased effects you can de-compose mutations having many rows into many mutations containing non-overlapping clustering ranges.

I cannot count the amount of times the fast-forwarding stress tests in this suite shaken out hidden bugs from my readers that the often simplistic tests I wrote didn't catch.

denesb (changes requested):

test/lib/mutation_source_test.cc
kbr-: It checks that `tested` returns the same thing as `authority`. After they both reach EOS, it checks that `tested` keeps returning EOS (by comparing it to `authority`, which I presume in this case will keep returning EOS if it's correct).

mutation_reader.cc
kbr-: Yes, it will.

I actually added this check because my test catched that this reader when merging 0 readers will return a different sequence of fragments than one created by `flat_mutation_reader_from_mutations` for empty mutation: it will instantly return EOS, while the reader_from_mutations will return `partition_start` followed by `partition_end`.

Hence I modified the test so it doesn't check the 0 readers/empty mutation case, and added this check here.

Isn't returning EOS before `partition_start` wrong? Wouldn't it confuse cache populator like in https://github.com/scylladb/scylla/issues/3552?

denesb: The reason is that an empty mutation != no data. En empty mutation contains exactly what you got: a partition start and end fragments.

denesb: > Isn't returning EOS before partition_start wrong? Wouldn't it confuse cache populator like in #3552?

This is the problem of the higher layer, the combined reader also doesn't concern itself of how one of the many users react to no data. They were the one passing in 0 readers in the first place, they should deal with the result.

kbr-:
v6:
- when `peek_reader` sees that the reader has `partition_end` as its next fragment, don't pop it - it's unnecessary, we're removing the reader anyway
- remove the check that there is at least one reader; in that case the merger will just return eos
- remove leftover `mfo_printer`
- add a mutation_source test for the new merger

See diff here: https://github.com/scylladb/scylla/compare/a0a0a919c87741e40d52b8b73d3f99e48231a361..cfc12fa4c14582a67cacceb1635d4aa4f7ae2114

kbr-:
v7: don't pop `partition_end` for galloping reader too (forgot to do that in the previous version)

denesb (approved):
LGTM.

I have one minor nit and a question, neither should block this series from going in.

test/boost/mutation_reader_test.cc
```diff
@@ -3658,3 +3658,214 @@ SEASTAR_THREAD_TEST_CASE(test_clustering_order_merger_in_memory) {
...
+            : impl(std::move(s), std::move(permit))
+            , _readers(std::move(readers)), _range(range)
+        {
+            assert(!_readers.empty());
+
+            _it = _readers.begin();
+            // restore invariant 2
+            while (_it != _readers.end() && _range.get().before(_it->first, dht::ring_position_comparator(*_schema))) {
+                ++_it;
+            }
```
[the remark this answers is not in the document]
kbr-: I needed the more generic `std::partition_point`, but anyway. Here you go.
See diff: https://github.com/scylladb/scylla/compare/7d16e56dd01e2f6410aa93149248c7e21ae6845f..502ed2e9f7fb9cf00d257137022f585ec933495b

test/boost/mutation_reader_test.cc
```diff
@@ -3658,3 +3658,214 @@ SEASTAR_THREAD_TEST_CASE(test_clustering_order_merger_in_memory) {
...
+        virtual future<> fill_buffer(db::timeout_clock::time_point timeout) override {
+            while (!is_buffer_full()) {
+                auto mfo = co_await next_fragment(timeout);
+                if (!mfo) {
+                    _end_of_stream = true;
+                    break;
+                }
+                push_mutation_fragment(std::move(*mfo));
+            }
+        }
```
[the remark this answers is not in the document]
avikivity: coroutines treat falling of the end as `co_return;`. `co_return x` translates to `return make_ready_future<T>(std::move(x))`, and `co_return;` translates to `return make_ready_future<>()`.

Note: this is for professionals only. Do not try falling off the end at home.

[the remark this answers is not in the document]
kbr-: It looks like for coroutines there's an implicit `co_return` inserted at the end, just like `return` for functions.
In case of the `future` coroutine type it looks like it returns `make_ready_future<>()` (not sure - but looks like it works)

Issue #6138, opened 2020-04-06
types: de/serialization: work in terms of fragmented buffers


denesb:
In the past few years we have converted most of our codebase to work in terms of fragmented buffers, instead of linearised ones, to help avoid large allocations that put large pressure on the memory allocator.
One prominent component that still works exclusively in terms of linearised buffers is the types hierarchy, more specifically the de/serialization code to/from CQL format. Note that for most types, this is the same as our internal format, notable exceptions are non-frozen collections and user types.

Most types are expected to contain reasonably small values, but texts, blobs and especially collections can get very large. Since the entire hierarchy shares a common interface we can either transition all or none to work with fragmented buffers.

We already have the tools to work with fragmented buffers: [bytes_ostream](https://github.com/scylladb/scylla/blob/master/bytes_ostream.hh) for writing fragmented buffers and [utils::linearizing_input_stream](https://github.com/scylladb/scylla/blob/master/utils/linearizing_input_stream.hh) for reading them.

Once we converted the types hierarchy we could convert the keys hierarchy as well ([keys.hh](https://github.com/scylladb/scylla/blob/master/keys.hh), [compound.hh](https://github.com/scylladb/scylla/blob/master/compound.hh) and [compound_compat.hh](https://github.com/scylladb/scylla/blob/master/compound_compat.hh)) and maybe we could kill `with_linearized_managed_bytes()` and friends for good.

PR #7692, opened 2020-11-25
types: get rid of linearization in deserialize()

michoecho:
Citing #6138:
> In the past few years we have converted most of our codebase to work in terms of fragmented buffers, instead of linearised ones, to help avoid large allocations that put large pressure on the memory allocator.
> One prominent component that still works exclusively in terms of linearised buffers is the types hierarchy, more specifically the de/serialization code to/from CQL format. Note that for most types, this is the same as our internal format, notable exceptions are non-frozen collections and user types.
> 
> Most types are expected to contain reasonably small values, but texts, blobs and especially collections can get very large. Since the entire hierarchy shares a common interface we can either transition all or none to work with fragmented buffers.

This series gets rid of intermediate linearizations in deserialization. The next steps are removing linearizations from serialization, validation and comparison code.

Series summary:
- Fix a bug in `fragmented_temporary_buffer::view::remove_prefix`. (Discovered while testing. Since it wasn't discovered earlier, I guess it doesn't occur in any code path in master.)
- Add a `FragmentedView` concept to allow uniform handling of various types of fragmented buffers (`bytes_view`, `temporary_fragmented_buffer::view`, `ser::buffer_view` and likely `managed_bytes_view` in the future).
- Implement `FragmentedView` for relevant fragmented buffer types.
- Add helper functions for reading from `FragmentedView`.
- Switch `deserialize()` and all its helpers from `bytes_view` to `FragmentedView`.
- Remove `with_linearized()` calls which just became unnecessary.
- Add an optimization for single-fragment cases.

The addition of `FragmentedView` might be controversial, because another concept meant for the same purpose - `FragmentRange` - is already used. Unfortunately, it lacks the functionality we need. The main (only?) thing we want to do with a fragmented buffer is to extract a prefix from it and `FragmentRange` gives us no way to do that, because it's immutable by design. We can work around that by wrapping it into a mutable view which will track the offset into the immutable `FragmentRange`, and that's exactly what `linearizing_input_stream` is. But it's wasteful. `linearizing_input_stream` is a heavy type, unsuitable for passing around as a view - it stores a pair of fragment iterators, a fragment view and a size (11 words) to conform to the iterator-based design of `FragmentRange`, when one fragment iterator (4 words) already contains all needed state, just hidden.
I suggest we replace `FragmentRange` with `FragmentedView` (or something similar) altogether.

Refs: #6138

### DIFF
--- a/cql3/maps.cc
+++ b/cql3/maps.cc
@@ -158,15 +158,13 @@ maps::value::from_serialized(const fragmented_temporary_buffer::view& fragmented
         // Collections have this small hack that validate cannot be called on a serialized object,
         // but compose does the validation (so we're fine).
         // FIXME: deserialize_for_native_protocol?!
-      return with_linearized(fragmented_value, [&] (bytes_view value) {
-        auto m = value_cast<map_type_impl::native_type>(type.deserialize(value, sf));
+        auto m = value_cast<map_type_impl::native_type>(type.deserialize(fragmented_value, sf));
         std::map<bytes, bytes, serialized_compare> map(type.get_keys_type()->as_less_comparator());
         for (auto&& e : m) {
             map.emplace(type.get_keys_type()->decompose(e.first),
                         type.get_values_type()->decompose(e.second));
         }
         return maps::value { std::move(map) };
-      });
     } catch (marshal_exception& e) {
         throw exceptions::invalid_request_exception(e.what());
     }
--- a/types.cc
+++ b/types.cc
@@ -1987,11 +1987,12 @@ decltype(auto) deserialize_value(const T& t, bytes_view v) {
 }
 
 namespace {
+template <FragmentedView View>
 struct deserialize_visitor {
-    bytes_view v;
+    View v;
     data_value operator()(const reversed_type_impl& t) { return t.underlying_type()->deserialize(v); }
     template <typename T> data_value operator()(const T& t) {
-        if (v.empty()) {
+        if (!v.size_bytes()) {
             return t.make_empty();
         }
         return t.make_value(deserialize_value(t, v));
@@ -2003,7 +2004,7 @@ struct deserialize_visitor {
          return t.make_value(deserialize_value(t, v));
     }
     data_value operator()(const bytes_type_impl& t) {
-        return t.make_value(std::make_unique<bytes_type_impl::native_type>(v.begin(), v.end()));
+        return t.make_value(std::make_unique<bytes_type_impl::native_type>(linearized(v)));
     }
     data_value operator()(const counter_type_impl& t) {
         return static_cast<const long_type_impl&>(*long_type).make_value(read_simple_exactly<int64_t>(v));
@@ -2023,9 +2024,15 @@ struct deserialize_visitor {
 };
 }
 
-data_value abstract_type::deserialize(bytes_view v) const {
-    return visit(*this, deserialize_visitor{v});
+template <FragmentedView View>
+data_value abstract_type::deserialize(View v) const {
+    return visit(*this, deserialize_visitor<View>{v});
 }
+// Explicit instantiation.
+// This should be repeated for every type passed to deserialize().
+template data_value abstract_type::deserialize<>(fragmented_temporary_buffer::view) const;
+template data_value abstract_type::deserialize<>(single_fragmented_view) const;
+template data_value abstract_type::deserialize<>(ser::buffer_view<bytes_ostream::fragment_iterator>) const;
 
 int32_t compare_aux(const tuple_type_impl& t, bytes_view v1, bytes_view v2) {
     // This is a slight modification of lexicographical_tri_compare:
--- a/types.cc
+++ b/types.cc
@@ -1849,15 +1849,18 @@ static cql_duration deserialize_value(const duration_type_impl& t, bytes_view v)
     return cql_duration(months_counter(months), days_counter(days), nanoseconds_counter(nanoseconds));
 }
 
-static inet_address deserialize_value(const inet_addr_type_impl&, bytes_view v) {
-    switch (v.size()) {
+template<FragmentedView View>
+inet_address deserialize_value(const inet_addr_type_impl&, View v) {
+    switch (v.size_bytes()) {
     case 4:
         // gah. read_simple_be, please...
         return inet_address(::in_addr{net::hton(read_simple<uint32_t>(v))});
-    case 16:
-        return inet_address(*reinterpret_cast<const ::in6_addr*>(v.data()));
+    case 16:;
+        ::in6_addr buf;
+        read_fragmented(v, sizeof(buf), reinterpret_cast<bytes::value_type*>(&buf));
+        return inet_address(buf);
     default:
-        throw marshal_exception(format("cannot deserialize inet_address, unsupported size {:d} bytes", v.size()));
+        throw marshal_exception(format("cannot deserialize inet_address, unsupported size {:d} bytes", v.size_bytes()));
     }
 }
 
--- a/types.cc
+++ b/types.cc
@@ -1091,11 +1091,6 @@ map_type_impl::deserialize(View in, cql_serialization_format sf) const {
     return make_value(std::move(m));
 }
 
-data_value
-map_type_impl::deserialize(bytes_view in, cql_serialization_format sf) const {
-    return deserialize(single_fragmented_view(in), sf);
-}
-
 static void validate_aux(const map_type_impl& t, bytes_view v, cql_serialization_format sf) {
     auto size = read_collection_size(v, sf);
     for (int i = 0; i < size; ++i) {
@@ -1266,11 +1261,6 @@ set_type_impl::deserialize(View in, cql_serialization_format sf) const {
     return make_value(std::move(s));
 }
 
-data_value
-set_type_impl::deserialize(bytes_view in, cql_serialization_format sf) const {
-    return deserialize(single_fragmented_view(in), sf);
-}
-
 bytes
 set_type_impl::serialize_partially_deserialized_form(
         const std::vector<bytes_view>& v, cql_serialization_format sf) {
@@ -1371,11 +1361,6 @@ list_type_impl::deserialize(View in, cql_serialization_format sf) const {
         s.push_back(std::move(e));
     }
     return make_value(std::move(s));
-}
-
-data_value
-list_type_impl::deserialize(bytes_view in, cql_serialization_format sf) const {
-    return deserialize(single_fragmented_view(in), sf);
 }
 
 static sstring vector_to_string(const std::vector<data_value>& v, std::string_view sep) {
--- a/types.cc
+++ b/types.cc
@@ -1808,7 +1808,7 @@ static void serialize(const abstract_type& t, const void* value, bytes::iterator
 }
 
 template <FragmentedView View>
-data_value collection_type_impl::deserialize(View v, cql_serialization_format sf) const {
+data_value collection_type_impl::deserialize_impl(View v, cql_serialization_format sf) const {
     struct visitor {
         View v;
         cql_serialization_format sf;
@@ -1829,9 +1829,9 @@ data_value collection_type_impl::deserialize(View v, cql_serialization_format sf
 }
 // Explicit instantiation.
 // This should be repeated for every View type passed to collection_type_impl::deserialize.
-template data_value collection_type_impl::deserialize<>(ser::buffer_view<bytes_ostream::fragment_iterator>, cql_serialization_format) const;
-template data_value collection_type_impl::deserialize<>(fragmented_temporary_buffer::view, cql_serialization_format) const;
-template data_value collection_type_impl::deserialize<>(single_fragmented_view, cql_serialization_format) const;
+template data_value collection_type_impl::deserialize_impl<>(ser::buffer_view<bytes_ostream::fragment_iterator>, cql_serialization_format) const;
+template data_value collection_type_impl::deserialize_impl<>(fragmented_temporary_buffer::view, cql_serialization_format) const;
+template data_value collection_type_impl::deserialize_impl<>(single_fragmented_view, cql_serialization_format) const;
 
 template <FragmentedView View>
 data_value deserialize_aux(const tuple_type_impl& t, View v) {
--- a/types.cc
+++ b/types.cc
@@ -1347,19 +1347,28 @@ serialize_list(const list_type_impl& t, const void* value, bytes::iterator& out,
     }
 }
 
+template <FragmentedView View>
 data_value
-list_type_impl::deserialize(bytes_view in, cql_serialization_format sf) const {
+list_type_impl::deserialize(View in, cql_serialization_format sf) const {
     auto nr = read_collection_size(in, sf);
     native_type s;
     s.reserve(nr);
     for (int i = 0; i != nr; ++i) {
-        auto e = _elements->deserialize(read_collection_value(in, sf));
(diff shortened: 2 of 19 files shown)


utils/fragment_range.hh
```diff
@@ -173,3 +173,28 @@ concept FragmentedMutableView = requires (T view) {
...
+        out = std::copy(v.current_fragment().begin(), v.current_fragment().end(), out);
+        v.remove_current();
+    }
+    return b;
+}
+
+template<FragmentedView View, typename Function>
+requires (!FragmentRange<View>) && requires (Function fn, bytes_view bv) {
+    fn(bv);
+}
```
avikivity, line 192: You can write the latter constraint as std::invocable<Function, bytes_view>

types.cc
```diff
@@ -1883,7 +1883,8 @@ static uint32_t deserialize_value(const simple_date_type_impl&, bytes_view v) {
     return read_simple_exactly<uint32_t>(v);
 }
 
-static int64_t deserialize_value(const time_type_impl&, bytes_view v) {
+template<FragmentedView View>
+int64_t deserialize_value(const time_type_impl&, View v) {
```
avikivity, line 1942: You can use the abbreviated syntax: `int64_t deserialize_value(const time_type_impl&, FragmentedView auto v)`, if you like.

types.cc
```diff
@@ -1811,16 +1811,20 @@ static data_value deserialize_aux(const tuple_type_impl& t, bytes_view v) {
     return data_value::make(t.shared_from_this(), std::make_unique<tuple_type_impl::native_type>(std::move(ret)));
 }
 
-static utils::multiprecision_int deserialize_value(const varint_type_impl&, bytes_view v) {
-    auto negative = v.front() < 0;
+template<FragmentedView View>
+utils::multiprecision_int deserialize_value(const varint_type_impl&, View v) {
+    bool negative = v.current_fragment().front() < 0;
```
avikivity, line 1855: What if the view is empty? The existing code will survive it.

I don't think we can actually get called here with an empty view, but better to be prepared.

cql3/maps.cc
```diff
@@ -158,15 +158,13 @@ maps::value::from_serialized(const fragmented_temporary_buffer::view& fragmented
         // Collections have this small hack that validate cannot be called on a serialized object,
         // but compose does the validation (so we're fine).
         // FIXME: deserialize_for_native_protocol?!
-      return with_linearized(fragmented_value, [&] (bytes_view value) {
-        auto m = value_cast<map_type_impl::native_type>(type.deserialize(value, sf));
+        auto m = value_cast<map_type_impl::native_type>(type.deserialize(fragmented_value, sf));
```
avikivity, line 161: In many cases, the value will come in a single fragment (e.g. someone passed us an int32, the odds of it being fragmented in our 8k buffers are 1:2000). But we will pay the cost of fragmented deserialization nevertheless (endless checking).

So abstract_type::deserialize() can check whether we have 0 or 1 fragments, and call the bytes_view variant in those cases.

Or maybe it can be a helper function that takes in a fragmented view and a function, checks if it has more than one fragment, and calls the inner function with a single_fragment_view if possible, and with the original input if not. This may replace many checks during inner processing with just one check in the outer processing.

This can come post merge.

cql3/maps.cc
```diff
@@ -158,15 +158,13 @@ maps::value::from_serialized(const fragmented_temporary_buffer::view& fragmented
         // Collections have this small hack that validate cannot be called on a serialized object,
         // but compose does the validation (so we're fine).
         // FIXME: deserialize_for_native_protocol?!
-      return with_linearized(fragmented_value, [&] (bytes_view value) {
-        auto m = value_cast<map_type_impl::native_type>(type.deserialize(value, sf));
```
avikivity, line 162: It pays to be lazy - we did not fix the half-indent, and now we don't need to.

types.cc
```diff
@@ -1822,6 +1822,29 @@ static void serialize(const abstract_type& t, const void* value, bytes::iterator
     return ::serialize(t, value, out, cql_serialization_format::internal());
 }
 
+template <FragmentedView View>
+data_value collection_type_impl::deserialize(View v, cql_serialization_format sf) const {
+    struct visitor {
+        View v;
+        cql_serialization_format sf;
+        data_value operator()(const abstract_type&) { abort(); }
```
avikivity, line 1815: Use on_internal_error() so as not to upgrade a but to total loss of availability.

types.cc
```diff
@@ -1822,6 +1822,29 @@ static void serialize(const abstract_type& t, const void* value, bytes::iterator
...
+        data_value operator()(const list_type_impl& t) {
+            return t.deserialize(v, sf);
+        }
+        data_value operator()(const map_type_impl& t) {
+            return t.deserialize(v, sf);
+        }
+        data_value operator()(const set_type_impl& t) {
+            return t.deserialize(v, sf);
+        }
```
avikivity, line 1824: Why do you need these three special cases? Can't you write `[sf, v] (const auto& t) { return t.deserialize(v, sf); }`?

Looks like the only thing the visitor does apart from that is to abort on a bad type.

avikivity:
Looks very good, very easy to review. I left some minor comments.

cql3/maps.cc
denesb: I think long-term we should examine why do we need these serialize-deserialize methods at all. We need to be able to do 3 things with our values:
* de/serialize to CQL wire format: for most types this is the same as the in-memory format
* de/serialize to on-disk format: I suspect for most types the same applies as before
* compare types

Neither of these should require deserializing a fragmented buffer belonging to a UTF8 value into a `sstring` C++ type, then re-serializing it into another format or comparing it to another `sstring`. We should drop the intermediate format.

One more uses case is pretty printing and genuine conversion to C++ types in tests. Neither of these is on any hot path.

types.cc
michoecho: Collections have a two-argument version of deserialize (`collection_type_impl::deserialize`) which allows specifying a `cql_serialization_format`. This used to be a virtual method, but since we want to template it, we need to make it non-virtual and dispatch the call manually, with an abstract_type visitor. Just like we do in `abstract_type::deserialize`. (We could also write a `switch` over the types ourselves. Maybe we could help the compiler with optimization that way). 

You can't `[sf, v] (const auto& t) { return t.deserialize(v, sf); }` because a visitor needs to match all types, and non-collection types don't have a `deserialize` with two arguments.

types.cc
michoecho: I will do that, but this code is *very* unreachable (a new collection type would need to be added) so it doesn't really matter.

types.cc
michoecho: > The existing code will survive it.

What do you mean? Calling `front()` on an empty `bytes_view` (aka `std::basic_string_view`) is undefined behaviour.

types.cc
michoecho: Wow, yet another concept syntax. I didn't know that was a thing.

> If you like

 I don't know. Should I like?

utils/fragment_range.hh
```diff
@@ -198,3 +198,16 @@ decltype(auto) with_linearized(const View& v, Function&& fn)
         return fn(linearized(v));
     }
 }
+
+class single_fragmented_view {
+    bytes_view _view;
+public:
+    using fragment_type = bytes_view;
+    explicit single_fragmented_view(bytes_view bv) : _view(bv) {}
```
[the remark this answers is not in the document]
michoecho: I made it explicit specifically so that it isn't seamless. It will make it easier to find potential linearizations and/or uses of old `bytes_view` versions of stuff.

types.cc
```diff
@@ -1899,9 +1899,24 @@ static T deserialize_value(const integer_type_impl<T>& t, bytes_view v) {
     return read_simple_exactly<T>(v);
 }
 
-static sstring deserialize_value(const string_type_impl&, bytes_view v) {
+template<FragmentedView View>
+sstring deserialize_value(const string_type_impl&, View v) {
     // FIXME: validation?
-    return sstring(reinterpret_cast<const char*>(v.begin()), v.size());
+    sstring buf(sstring::initialized_later(), v.size_bytes());
```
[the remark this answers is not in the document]
michoecho: True. But it's an independent issue for another pull request.

utils/fragment_range.hh
```diff
@@ -173,3 +173,28 @@ concept FragmentedMutableView = requires (T view) {
...
+template<FragmentedView View, typename Function>
+requires (!FragmentRange<View>) && requires (Function fn, bytes_view bv) {
+    fn(bv);
+}
+decltype(auto) with_linearized(const View& v, Function&& fn)
+{
+    if (v.size_bytes() == v.current_fragment().size()) [[likely]] {
+        return fn(v.current_fragment());
+    } else {
+        return fn(linearized(v));
```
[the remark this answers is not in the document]
michoecho: I don't think so. In C++, temporaries stay alive until the full containing expression is evaluated, no?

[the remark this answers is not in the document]
denesb: I don't think so. Here, the compiler has no way to know that `bytes_view` depends on the lifetime of `bytes`. I think only the value immediately passed to the function is lifetime-extended. But an experiment should be easy enough to confirm/deny.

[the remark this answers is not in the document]
michoecho: After https://en.cppreference.com/w/cpp/language/lifetime:
> All temporary objects are destroyed as the last step in evaluating the full-expression that (lexically) contains the point where they were created, and if multiple temporary objects were created, they are destroyed in the order opposite to the order of creation.

In other words, for:
```
struct A {
    A() { std::cout << "A "; }
    ~A() { std::cout << "~A "; }
};
struct B {
    B() { std::cout << "B "; }
    ~B() { std::cout << "~B "; }
    operator A() { return A(); }
};
void f(A) {
    std::cout << "f "; 
}
int main() { f(B()); }
```
the only possible output is `B A f ~A ~B `.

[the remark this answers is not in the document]
michoecho:
```
int& f(int&& x) { return &x; }
int main() {
    int *a;
    a = &f(0);
    std::cout << *a;
}
```
is undefined behaviour, but
```
int& f(int&& x) { return &x; }
int main() {
    int *a;
    a = &f(0),
    std::cout << *a;
}
```
is perfectly good C++.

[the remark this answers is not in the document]
denesb: We are safe then. I learned something new today.

types.cc
avikivity: > Should I like?

Good question. One the one hand, it has less fluff. On the other hand, it hides the fact that it is a template. On the feet, we may learn to recognize the syntax in time, and perhaps the fact that it is a template is not so important.

types.cc
avikivity: You are right, I withdraw the comment.

types.cc
avikivity: I see. Maybe they should have the two-argument form. But even if we agree they should, let's leave it for later.

michoecho:
Post-review changes:
- Adressed the minor comments in https://github.com/scylladb/scylla/pull/7692#discussion_r530246766, https://github.com/scylladb/scylla/pull/7692#discussion_r530266440 and https://github.com/scylladb/scylla/pull/7692#discussion_r530880687.
- Corrected a comment.
- Documented the preconditions of FragmentedView methods.
- Added an optimization for single-fragment views (c990869), as suggested in https://github.com/scylladb/scylla/pull/7692#discussion_r530259586.

utils/fragment_range.hh
michoecho: Done.

types.hh
```diff
@@ -1108,6 +1108,22 @@ typename Type::value_type deserialize_value(Type& t, bytes_view v) {
     return t.deserialize_value(v);
 }
 
+template<FragmentedView View>
+void read_fragmented(View& v, size_t n, bytes::value_type* out) {
+    assert(n <= v.size_bytes());
```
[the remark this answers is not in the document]
michoecho: Done.

cql3/maps.cc
michoecho: @avikivity Done in c990869. Please take a look.

michoecho:
I have removed linearizations from validate() (https://github.com/michoecho/scylla/compare/types_remove_linearization...michoecho:types_remove_linearization_2) and moved on to serialize(). I wonder, should I push the validate() changes here for review, or make a separate PR?

avikivity:
@michoecho this one is fat enough, let's get it in

avikivity:
@denesb if you have no further comments, I can merge.

michoecho:
There's a lot of very small commits here. Should I do some squashing before merge?

avikivity:
I love small commits!

michoecho:
New change: I forgot to push the single-fragment optimization for the collection (two-argument) form of deserialize() earlier. It's the same optimization as for the single-argument form.

denesb:
>     * Added an optimization for single-fragment views ([c990869](https://github.com/scylladb/scylla/commit/c990869c8ea66e24c96d978bfc1c38fc1c74b3b2)), as suggested in [#7692 (comment)](https://github.com/scylladb/scylla/pull/7692#discussion_r530259586).

I don't see how this resolves the problem. All the different overloads, including the `single_fragment_view` uses the same generic implementation, I didn't find any `single_fragment_view` specific overload taking advantage of the fact that we have single fragment. I can see that we have `bytes_view` variants of the various `read_*` methods, but I don't see when a `single_fragment_view` is converted to `bytes_view`. What am I missing?

michoecho:
@denesb It's not about `read_*`, but about the `FragmentedView` methods. `prefix()`, `remove_prefix()` and `remove_current()` are trivial for `single_fragmented_view`, but complex for other `FragmentedView` types. That's why it's an optimization.
`single_fragmented_view` _is_ a `bytes_view` underneath, so `read_*` functions don't do any extra work compared to their `bytes_view` counterparts. (With the exception of `read_fragmented`, which I used to read IPv6 addresses, which previously were read by simple `memcpy`. It probably deserves a special overload. I will add it in a moment.)
Besides that, we are now passing 2 words (as opposed to 4 words of multi-fragment views) through optimization barriers.

denesb:
I see, thanks for explaining.

denesb (approved):
LGTM.

michoecho:
Minor changes: https://github.com/scylladb/scylla/compare/4953df5..588bebb
- Removed the bounds check from `read_fragmented`, because it's always called after they have already been checked. Added a comment documenting that.
- Made `read_simple_exactly` not modify its argument, to match the behaviour of the `bytes_view` version exactly (though it doesn't matter).
- Added a specialization for `read_fragmented(single_fragmented_view)`, because the generic version does more work than necessary for single-fragment views.

michoecho:
Fixed a linkage mistake I just made: https://github.com/scylladb/scylla/compare/588bebb..e504bba

michoecho:
@avikivity I haven't modified the PR description (which has just become the merge commit message in next) after review fixes, so it contains outdated commit hashes. Is that fine?

avikivity:
It's not the end of the world. In the future avoid using commit hashes in descriptions, they're too volatile until actually committed.

michoecho:
Added a missing explicit instantiation for `collection_type_impl::deserialize(single_fragmented_view) and collection_type_impl::deserialize(temporary_fragmented_buffer::view)`. (Its lack caused `release` builds to fail, because the implicit instantiations were optimized away.)
Removed commit hashes from the PR description.

PR #7699, opened 2020-11-25
Don't use secondary indices for multi-column restrictions

### LINKED ISSUE #7680: Multi-column restrictions deemed invalid with a secondary index
The following CQL works fine on Cassandra:
```cql
create table t (p int, c1 int, c2 int, primary key(p, c1, c2));
create index i1 on t(c1);
insert into t(p, c1, c2) values (1, 11, 21);
insert into t(p, c1, c2) values (2, 12, 22);
select c1 from t where (c1,c2)=(11,21) allow filtering;
```
but throws invalid_request_exception on Scylla:
`exceptions::invalid_request_exception (Single value can be obtained from single-column restrictions only)`

dekimir:
Fix #7680 by never using secondary index for multi-column restrictions.

Modify expr::is_supported_by() to handle multi-column correctly.

Tests: unit (dev)

### DIFF
--- a/cql3/restrictions/statement_restrictions.cc
+++ b/cql3/restrictions/statement_restrictions.cc
@@ -230,7 +230,8 @@ statement_restrictions::statement_restrictions(database& db,
     process_clustering_columns_restrictions(select_a_collection, for_view, allow_filtering);
 
     // Covers indexes on the first clustering column (among others).
-    if (_is_key_range && _has_queriable_ck_index) {
+    if (_is_key_range && _has_queriable_ck_index &&
+        !dynamic_pointer_cast<multi_column_restriction>(_clustering_columns_restrictions)) {
         _uses_secondary_indexing = true;
     }
 
--- a/cql3/expr/expression.cc
+++ b/cql3/expr/expression.cc
@@ -776,9 +776,11 @@ bool is_supported_by(const expression& expr, const secondary_index::index& idx) 
                             return idx.supports_expression(*col.col, oper.op);
                         },
                         [&] (const std::vector<column_value>& cvs) {
-                            return boost::algorithm::any_of(cvs, [&] (const column_value& c) {
-                                return idx.supports_expression(*c.col, oper.op);
-                            });
+                            if (cvs.size() == 1) {
+                                return idx.supports_expression(*cvs[0].col, oper.op);
+                            }
+                            // We don't use index table for multi-column restrictions, as it cannot avoid filtering.
+                            return false;
                         },
                         [&] (const token&) { return false; },
                     }, oper.lhs);
--- a/test/boost/secondary_index_test.cc
+++ b/test/boost/secondary_index_test.cc
@@ -1589,3 +1589,29 @@ SEASTAR_TEST_CASE(test_filtering_indexed_column) {
         });
     });
 }
+
+namespace {
+
+auto I(int32_t x) { return int32_type->decompose(x); }
+
+} // anonymous namespace
+
+SEASTAR_TEST_CASE(indexed_multicolumn_where) {
+    return do_with_cql_env_thread([] (auto& e) {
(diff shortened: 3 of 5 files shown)


psarna (approved):

cql3/restrictions/statement_restrictions.cc
```diff
@@ -230,7 +230,8 @@ statement_restrictions::statement_restrictions(database& db,
     process_clustering_columns_restrictions(select_a_collection, for_view, allow_filtering);
 
     // Covers indexes on the first clustering column (among others).
-    if (_is_key_range && _has_queriable_ck_index) {
+    if (_is_key_range && _has_queriable_ck_index &&
+        !dynamic_pointer_cast<multi_column_restriction>(_clustering_columns_restrictions)) {
         _uses_secondary_indexing = true;
     }
```
avikivity, line 236: On Cassandra, does  it use the index with multi-column restrictions, or does it give up and filter?

Since (a, b) = (1, 2) can be rewritten as a = 1 AND b = 2, an index on either a or b should work.

dekimir: > On Cassandra, does it use the index with multi-column restrictions, or does it give up and filter?

I can't answer that without research, but why does it matter?  Are Cassandra indices sufficiently similar to ours to make that answer applicable?
 
> Since (a, b) = (1, 2) can be rewritten as a = 1 AND b = 2, an index on either a or b should work.

How?  What index-table query would give you exactly the right rows, without further filtering?

nyh: There are some specific cases which can be done without filtering. For example, you can do `a = 1 AND k = 2`, where `a` is an indexed column and `k` is the partition key, without filtering. Or, unrelated to indexes, you can do `k = 1 AND c = 2` for the partition and clustering key. If (k, c) = (1, 2) is just another syntax for the same thing, why should it require filtering?

dekimir: > There are some specific cases which can be done without filtering.

In all such cases the base table can be queried without filtering, too (unless I'm missing something).  And that will always be faster.

> Or, unrelated to indexes, you can do `k = 1 AND c = 2` for the partition and clustering key. If (k, c) = (1, 2)

[Only clustering columns](https://sourcegraph.com/github.com/scylladb/scylla/-/blob/cql3/multi_column_relation.hh#L230:50) can participate in a multi-column restriction; this query is invalid.

avikivity: > > On Cassandra, does it use the index with multi-column restrictions, or does it give up and filter?
> 
> I can't answer that without research, but why does it matter? Are Cassandra indices sufficiently similar to ours to make that answer applicable?
> 
> > Since (a, b) = (1, 2) can be rewritten as a = 1 AND b = 2, an index on either a or b should work.
> 
> How? What index-table query would give you exactly the right rows, without further filtering?

You would need further filtering. But presumably, if you have an index on a and you then filter for b, that is better than a full scan filtering on both.

dekimir: > if you have an index on a and you then filter for b, that is better than a full scan filtering on both.

There is neither full scan nor filtering, though -- if `(a,b)=?` is allowed at all, then `(a,b)` is a clustering prefix.

avikivity: Are tuple restrictions rejected early if they are not a clustering prefix?

dekimir: > Are tuple restrictions rejected early if they are not a clustering prefix?

Yes: see [link](https://sourcegraph.com/github.com/scylladb/scylla/-/blob/cql3/multi_column_relation.hh#L230:50) above.

avikivity: Ok.

PR #7705, opened 2020-11-26
install.sh: apply sysctl.d files on non-packaging installation

### LINKED ISSUE #7702: install.sh does not apply sysctl configuration
install.sh copies files to /etc/sysctl.d, but does not apply them, so they do not take effect until reboot.

Usually, this is fine, since rpm/deb will take care of that. But non-packaging installs need it.

syuu1228:
We don't apply sysctl.d files on non-packaging installation, apply them
just like rpm/deb taking care of that.

Fixes #7702

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -423,6 +423,10 @@ elif ! $packaging; then
     chown -R scylla:scylla $rdata
     chown -R scylla:scylla $rhkdata
 
+    for file in dist/common/sysctl.d/*.conf; do
+        bn=$(basename "$file")
+        sysctl -p "$rusr"/lib/sysctl.d/"$bn"
+    done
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -423,6 +423,10 @@ elif ! $packaging; then
     chown -R scylla:scylla $rdata
     chown -R scylla:scylla $rhkdata
 
+    for file in dist/common/sysctl.d/*.conf; do
+        bn=$(basename "$file")
+        sysctl -p "$rusr"/lib/sysctl.d/"$bn"
+    done
```
avikivity, line 429: Maybe it should be in post_install.sh? Or maybe we should merge post_install.sh here.

PR #7719, opened 2020-11-29
test/cql-pytest: tests for string validation (UTF-8 and ASCII)

nyh:
The first two patches in this series are small improvements to cql-pytest to prepare for the third and main patch. This third patch adds cql-pytest tests which check that we fail CQL queries that try to inject non-ASCII and non-UTF-8 strings for ascii and text columns, respectively.

The tests do not discover any unknown bug in Scylla, however, they do show that Scylla is more strict in its definition of "valid UTF-8" compared to Cassandra.

### DIFF
--- a/test/cql-pytest/conftest.py
+++ b/test/cql-pytest/conftest.py
@@ -69,5 +69,16 @@ def test_keyspace(cql):
     yield name
     cql.execute("DROP KEYSPACE " + name)
 
+# The "scylla_only" fixture can be used by tests for Scylla-only features,
+# which do not exist on Apache Cassandra. A test using this fixture will be
+# skipped if running with "run-cassandra".
+@pytest.fixture(scope="session")
+def scylla_only(cql):
+    # We recognize Scylla by checking if there is any system table whose name
+    # contains the word "scylla":
+    names = [row.table_name for row in cql.execute("SELECT * FROM system_schema.tables WHERE keyspace_name = 'system'")]
+    if not any('scylla' in name for name in names):
+        pytest.skip('Scylla-only test skipped')
+
 # TODO: use new_test_table and "yield from" to make shared test_table
 # fixtures with some common schemas.
--- a/test/cql-pytest/test_validation.py
+++ b/test/cql-pytest/test_validation.py
@@ -1,0 +1,368 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 ScyllaDB
+#
+# This file is part of Scylla.
+#
+# Scylla is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Scylla is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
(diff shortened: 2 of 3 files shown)


test/cql-pytest/conftest.py
```diff
@@ -69,5 +69,16 @@ def test_keyspace(cql):
     yield name
     cql.execute("DROP KEYSPACE " + name)
 
+# The "scylla_only" fixture can be used by tests for Scylla-only features,
```
yarongilor, line 72: i'm not too familiar with this pytest implementation, but doesn't it have a logging of the tested version at all? meaning, like Dtest has a scylla_version parameter or a log print for that.

nyh: The idea of cql-pytest is that it uses **CQL** (i.e., the Python Cassandra driver) to communicate with the server - that can be either Scylla of any version, Cassandra, one node or a cluster of 20 nodes, etc. When we run Scylla (`test/cql-pytest/run` does this automatically), Scylla's log contains whatever it usually prints out (including the version), and this is available in Jenkins for example; But the test itself does not have access to this log - just to CQL requests. That's intentional.

test/cql-pytest/test_validation.py
```diff
@@ -0,0 +1,368 @@
...
+    finally:
+        cassandra.cqltypes.UTF8Type.serialize = orig_serialize
+
+# 3. We can also insert the bad UTF-8 as part of the request string itself.
+#    This will make the entire request string invalid UTF-8, not just the
+#    value to be inserted, so Scylla and Cassandra should complain that the
+#    entire request is bad - not just the inserted value.
+# FIXME: this test is INCOMPLETE! It's very hard to get Python to output an
+# illegal UTF-8 string in this case, and I gave up. Unlike the prepared-
+# statement case above where it was easy to find and monkey-patch the
```
avikivity, line 215: This monkey-patch sounds like a time bomb, exploding when the driver implementation changes.

test/cql-pytest/test_validation.py
```diff
@@ -0,0 +1,368 @@
...
+        cql.execute(f"INSERT INTO {table1} (k, t) VALUES (1, '{s}')")
+        results = list(cql.execute(f"SELECT k, t FROM {table1} WHERE k=1"))
+        assert len(results) == 1
+        assert results[0].k == 1 and results[0].t == s
+    # FIXME: Need to figure out the appropriate monkey-patching or other
+    # trick to make the following work (i.e., pass the invalid string to
+    # the server, and let the server - not the driver - fail.
+    # for b in bad_utf8:
+    #  print(b)
+    #  cql.execute("INSERT INTO {} (k, t) VALUES (1, '{}')".format(table1, b.decode(errors='surrogateescape')))
```
avikivity, line 231: Possible trick: hijack the connection between server and client using a proxy, send values such as `XXX replace me XXX`, and replace them in the proxy.

test/cql-pytest/test_validation.py
nyh: This is true, but it's not really a dangerous bomb - perhaps a stink bomb :-) the most likely consequence of such a driver change would be that the test will fail, and at that point we can either quickly fix the monkey-patching or mark the test "skip".

test/cql-pytest/test_validation.py
nyh: Yes, this is possible, and the proxy can even be in Python (the proxy can use bytes instead of unicode strings) so it can be part of the test to run this proxy and connect to it. Python even has an existing way to encode invalid UTF-8 - that "surrogateescape" thing - so we don't necessarily need to invent our own "XXX" things.
Sounds like a very interesting exercise :-)

Issue #7395, opened 2020-10-12
scylla_setup fails in non-interactive mode


cogitaria-admin:
*Installation details*

Scylla version (or git commit hash): 4.1

Cluster size:
OS (RHEL/CentOS/Ubuntu/AWS AMI): 1

I ran scylla_setup in interactive mode on the first node of a multi-node cluster.  On completion I received this message:

```
ScyllaDB setup finished.
scylla_setup accepts command line arguments as well! For easily provisioning in a similar environment than this, type:

    scylla_setup --no-raid-setup --no-enable-service --no-ntp-setup \
                 --no-coredump-setup --no-sysconfig-setup \
                 --no-version-check --no-cpuscaling-setup
```
The suggested command **fails** with a usage message every time. There is no indication what is wrong or missing.  Just to reiterate, the failing command is the one suggested by scylla_setup itself, following interactive installation. 

```
usage: scylla_setup [-h] (--disks DISKS | --no-raid-setup) [--nic NIC]
                    [--ntp-domain NTP_DOMAIN] [--ami] [--setup-nic-and-disks]
                    [--developer-mode] [--no-ec2-check] [--no-kernel-check]
                    [--no-verify-package] [--no-enable-service]
                    ... MORE DETAILS ...
```

syuu1228:
I'm not able to reproduce the error on 4.1.0, 4.1.9 and 4.4-dev.
However, I found another pattern that could not run:
```
ScyllaDB setup finished.
scylla_setup accepts command line arguments as well! For easily provisioning in a similar environment than this, type:

    scylla_setup --no-raid-setup --no-verify-package --io-setup

Also, to avoid the time-consuming I/O tuning you can add --no-io-setup and copy the contents of /etc/scylla.d/io*
Only do that if you are moving the files into machines with the exact same hardware
```
This is because --io-setup expects one argument, but the message doesn't include it.
I guess original issue has similar problem but I don't find what it is.

syuu1228:
Okay I found the reason of this.
scylla_setup requires to set --nic even --no-sysconfig-setup, but when we skipping sysconfig_setup we don't specify NIC, so suggesting argument and actual behavior doesn't matches.

see: https://github.com/scylladb/scylla/blob/master/dist/common/scripts/scylla_setup#L295

io_setup issue is another bug, I'll create new issue.

syuu1228:
Previous comment was wrong, so deleted it

syuu1228:
[...]

PR #7724, opened 2020-11-30
scylla_setup: fix wrong command suggestion

syuu1228:
scylla_setup command suggestion does not shows an argument of --io-setup,
because we mistakely stores bool value on it (recognized as 'store_true').
We always need to print '--io-setup X' on the suggestion instead.

Also, --nic is currently ignored by command suggestion, need to print it just like other options.

Related #7395


dist/common/scripts/scylla_setup
```diff
@@ -299,7 +299,7 @@ if __name__ == '__main__':
                     sys.exit(1)
 
     disks = args.disks
-    nic = args.nic
+    args.nic
```
avikivity, line 302: What does this do?

syuu1228: Oh sorry, that was probably mistakenly removed while editing the code, reverted the line

PR #7727, opened 2020-12-01
debian pkg: Make deb packages explicitly depend on versioned components

### LINKED ISSUE #5514: Scylla Ubuntu meta package doesn't specify versions of its dependencies
*Installation details*
Scylla version (or git commit hash): 3.1.0-0.20191012.9c3cdded9-1
Cluster size: N/A
OS (RHEL/CentOS/Ubuntu/AWS AMI): Ubuntu 18.04

*Problem*
Installing the 3.1.0 version of Scylla using the `scylla` metapackage will install more recent 3.1.x versions of scylla.

First, use `apt` to install version 3.1.0 of scylla:
```
ubuntu@dev-mike-scylla001.us-e:/etc/apt/sources.list.d$ sudo apt install scylla=3.1.0-0.20191012.9c3cdded9-1
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following additional packages will be installed:
  scylla-conf scylla-jmx scylla-kernel-conf scylla-python3 scylla-server scylla-tools scylla-tools-core
The following NEW packages will be installed:
  scylla scylla-conf scylla-jmx scylla-kernel-conf scylla-python3 scylla-server scylla-tools scylla-tools-core
```
`apt` will then install the most recent 3.1.x version of Scylla:
```
Preparing to unpack .../0-scylla-conf_3.1.2-0.20191127.9d78d848e-1_amd64.deb ...
Unpacking scylla-conf (3.1.2-0.20191127.9d78d848e-1) ...
Selecting previously unselected package scylla-python3.
Preparing to unpack .../1-scylla-python3_3.7.2-0.20191127.9d78d848e-1_amd64.deb ...
Unpacking scylla-python3 (3.7.2-0.20191127.9d78d848e-1) ...
Selecting previously unselected package scylla-server.
Preparing to unpack .../2-scylla-server_3.1.2-0.20191127.9d78d848e-1_amd64.deb ...
Unpacking scylla-server (3.1.2-0.20191127.9d78d848e-1) ...
Selecting previously unselected package scylla-jmx.
...
...
```

This was a problem for us because we spun up a new Scylla node which then started alerting because it was on a different version from the rest of the fleet. It seems like the meta package should be declaring the correct versions of its dependencies as 3.1.0 here.

eliransin:
Up until now, Scylla's debian packages dependencies versions were
unspecified. This was due to a technical difficulty to determine
the version of the dependent upon packages (such as scylla-python3
or scylla-jmx). Now, when those packages are also built as part of
this repo and are built with a version identical to the server package
itself we can depend all of our packages with explicit versions.
The motivation for this change is that if a user tries to install
a specific Scylla version by installing a specific meta package,
it will silently drag in the latest components instead of the ones
of the requested versions.
The expected change in behavior is that after this change an attempt
to install a metapackage with version which is not the latest will fail
with an explicit error hinting the user what other packages of the same
version should be explicitly included in the command line.

Fixes #5514

### DIFF
--- a/dist/debian/control.template
+++ b/dist/debian/control.template
@@ -17,7 +17,7 @@ Conflicts: %{product}-server (<< 1.1)
 
 Package: %{product}-server
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}, %{product}-conf, %{product}-python3
+Depends: ${shlibs:Depends}, ${misc:Depends}, %{product}-conf (= ${binary:Version}), %{product}-python3 (= ${binary:Version})
 Description: Scylla database server binaries 
  Scylla is a highly scalable, eventually consistent, distributed,
  partitioned row DB.
@@ -42,7 +42,11 @@ Description: Scylla kernel tuning configuration
 Package: %{product}
 Section: metapackages
 Architecture: any
-Depends: %{product}-server, %{product}-jmx, %{product}-tools, %{product}-tools-core, %{product}-kernel-conf
+Depends: %{product}-server (= ${binary:Version})
+ , %{product}-jmx (= ${binary:Version})
+ , %{product}-tools (= ${binary:Version})
+ , %{product}-tools-core (= ${binary:Version})
+ , %{product}-kernel-conf (= ${binary:Version})
 Description: Scylla database metapackage
  Scylla is a highly scalable, eventually consistent, distributed,
  partitioned row DB.


avikivity:
Looks good.

eliransin:
@syuu1228 can you have a look?
Thanks,

eliransin:
@syuu1228  / @penberg  ping

eliransin:
@penberg / @syuu1228 if you are unavailable to review this, who can review?

PR #7730, opened 2020-12-01
Commitlog optional sync wait with timeout + lowered/configurable disk/buffer alignment (v3)

elcallio:
Makes `force_sync` optionally contain a timeout value, which, if present, will ensure the periodic flush is done no later than this, and wait for said flush to happen. 

Also make disk/buffer alignment be taken from actual files when creating/recycling them. Iff we are using fully overwritten files (currently restricted to the o_dsync mode), we can also use the potentially smaller read disk alignment, since we are not appending. 
Disk alignment can also be configured via config object if user has better knowledge on how we should do this. 

v2:
    * Use overwrite alignment if doing only overwrite
    * Ensure to adjust actual alignment if/when doing file wrapping
 v3:
    * Kill alignment config param. Useless and unsafe.

### DIFF
--- a/db/commitlog/commitlog.cc
+++ b/db/commitlog/commitlog.cc
@@ -275,6 +275,7 @@ public:
     using timeout_exception_factory = request_controller_timeout_exception_factory;
 
     basic_semaphore<timeout_exception_factory> _flush_semaphore;
+    basic_semaphore<timeout_exception_factory> _sync_semaphore;
 
     seastar::metrics::metric_groups _metrics;
 
@@ -394,6 +395,17 @@ public:
         if (!_shutdown) {
             _timer.arm(std::chrono::milliseconds(cfg.commitlog_sync_period_in_ms + extra));
         }
+    }
+
+    future<> force_sync(time_point latest) {
+        if (_timer.get_timeout() < latest) {
+            return make_ready_future<>();
+        }        
+        
+        _timer.cancel();
+        _timer.arm(latest);
+
+        return _sync_semaphore.wait();
     }
 
     std::vector<sstring> get_active_names() const;
@@ -904,6 +916,12 @@ public:
         });
     }
 
+    future<> force_sync(time_point latest, replay_position waitfor) {
+        return _segment_manager->force_sync(latest).then([this, waitfor] {
+            return _pending_ops.wait_for_pending(waitfor);
+        });
+    }
+
     /**
      * Add a "mutation" to the segment.
      */
@@ -977,6 +995,8 @@ public:
             write<uint32_t>(out, mecrc->checksum());
         }
 
+        replay_position high;
+
         for (size_t entry = 0; entry < writer->num_entries; ++entry) {
             replay_position rp(_desc.id, position());
             auto id = writer->id(entry);
@@ -1007,6 +1027,7 @@ public:
                 mecrc->process(checksum);
             }
 
+            high = rp;
             writer->result(entry, std::move(h));
         }
 
@@ -1018,8 +1039,10 @@ public:
         ++_segment_manager->totals.allocation_count;
         ++_num_allocs;
 
-        if (_segment_manager->cfg.mode == sync_mode::BATCH || writer->sync) {
+        if (_segment_manager->cfg.mode == sync_mode::BATCH || writer->sync == force_sync::yes) {
             return batch_cycle(timeout).discard_result();
+        } else if (writer->sync) {
+            return force_sync(*writer->sync, high);
         } else {
             // If this buffer alone is too big, potentially bigger than the maximum allowed size,
             // then no other request will be allowed in to force the cycle()ing of this buffer. We
@@ -1145,6 +1168,7 @@ db::commitlog::segment_manager::segment_manager(config c)
         ? size_t(std::ceil(*cfg.commitlog_flush_threshold_in_mb / double(smp::count))) * 1024 * 1024 
         : (max_disk_size - (max_disk_size > (max_size/2) ? (max_size/2) : 0)))
     , _flush_semaphore(cfg.max_active_flushes)
+    , _sync_semaphore(0)
     // That is enough concurrency to allow for our largest mutation (max_mutation_size), plus
     // an existing in-flight buffer. Since we'll force the cycling() of any buffer that is bigger
     // than default_size at the end of the allocation, that allows for every valid mutation to
@@ -1868,9 +1892,12 @@ future<> db::commitlog::segment_manager::clear() {
  * Called by timer in periodic mode.
  */
 void db::commitlog::segment_manager::sync() {
-    for (auto s : _segments) {
-        (void)s->sync(); // we do not care about waiting...
-    }
+    // we do not care about waiting...
+    (void)parallel_for_each(_segments, [](sseg_ptr s) {
+        return s->sync().discard_result(); 
+    }).then([this] {
+        _sync_semaphore.signal(_sync_semaphore.waiters());
+    });
 }
 
 void db::commitlog::segment_manager::on_timer() {
@@ -1881,6 +1908,7 @@ void db::commitlog::segment_manager::on_timer() {
         if (cfg.mode != sync_mode::BATCH) {
             sync();
         }
+
         // IFF a new segment was put in use since last we checked, and we're
         // above threshold, request flush.
         if (_new_counter > 0) {
--- a/db/commitlog/commitlog.cc
+++ b/db/commitlog/commitlog.cc
@@ -414,8 +414,8 @@ public:
 
     using buffer_type = fragmented_temporary_buffer;
 
-    buffer_type acquire_buffer(size_t s);
-    temporary_buffer<char> allocate_single_buffer(size_t);
+    buffer_type acquire_buffer(size_t s, size_t align);
+    temporary_buffer<char> allocate_single_buffer(size_t, size_t);
 
     future<std::vector<descriptor>> list_descriptors(sstring dir);
 
@@ -447,6 +447,7 @@ private:
     future<> _reserve_replenisher;
     seastar::gate _gate;
     uint64_t _new_counter = 0;
+    std::optional<size_t> _disk_write_alignment;
 };
 
 template<typename T>
@@ -516,6 +517,8 @@ class db::commitlog::segment : public enable_shared_from_this<segment>, public c
     uint64_t _flush_pos = 0;
     uint64_t _size_on_disk = 0;
 
+    size_t _alignment;
+
     bool _closed = false;
     // Not the same as _closed since files can be reused
     bool _closed_file = false;
@@ -572,14 +575,14 @@ public:
     // The commit log (chained) sync marker/header size in bytes (int: length + int: checksum [segmentId, position])
     static constexpr size_t sync_marker_size = 2 * sizeof(uint32_t);
 
-    static constexpr size_t alignment = 4096;
     // TODO : tune initial / default size
-    static constexpr size_t default_size = align_up<size_t>(128 * 1024, alignment);
+    static constexpr size_t default_size = 128 * 1024;
 
-    segment(::shared_ptr<segment_manager> m, descriptor&& d, file&& f, uint64_t initial_disk_size)
+    segment(::shared_ptr<segment_manager> m, descriptor&& d, file&& f, uint64_t initial_disk_size, size_t alignment)
             : _segment_manager(std::move(m)), _desc(std::move(d)), _file(std::move(f)),
         _file_name(_segment_manager->cfg.commit_log_location + "/" + _desc.filename()), 
         _size_on_disk(initial_disk_size),
+        _alignment(alignment),
         _sync_time(clock_type::now()), _pending_ops(true) // want exception propagation
     {
         ++_segment_manager->totals.segments_created;
@@ -747,10 +750,10 @@ public:
             overhead += descriptor_header_size;
         }
 
-        auto a = align_up(s + overhead, alignment);
+        auto a = align_up(s + overhead, _alignment);
         auto k = std::max(a, default_size);
 
-        _buffer = _segment_manager->acquire_buffer(k);
+        _buffer = _segment_manager->acquire_buffer(k, _alignment);
         _buffer_ostream = _buffer.get_ostream();
         auto out = _buffer_ostream.write_substream(overhead);
         out.fill('\0', overhead);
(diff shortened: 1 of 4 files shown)


db/commitlog/commitlog.cc
```diff
@@ -342,6 +343,21 @@ class db::commitlog::segment_manager : public ::enable_shared_from_this<segment_
         }
     }
 
+    future<> force_sync(time_point latest) {
+        if (_timer.get_timeout() < latest) {
+            return make_ready_future<>();
+        }
+        if (_timer.armed()) {
+            _timer.cancel();
+        }
```
avikivity, line 352: canceling an unarmed timer is safe.

db/commitlog/commitlog.cc
```diff
@@ -1683,6 +1710,9 @@ void db::commitlog::segment_manager::on_timer() {
         if (cfg.mode != sync_mode::BATCH) {
             sync();
         }
+
+        _sync_semaphore.signal(_sync_semaphore.waiters());
+
```
avikivity, line 1723: What if different waiters are waiting for different times?

db/commitlog/commitlog_entry.hh
```diff
@@ -37,9 +37,35 @@ public:
...
+    }
+    bool operator!=(const force_sync& v) const {
+        return !(*this == v);
+    }
+
+    static const force_sync no;
+    static const force_sync yes;
+};
+}
+
```
avikivity, line 65: Maybe this can be simplified: 

  time_point::min() -> must sync immediately
  time::point(1234) -> must sync by time 1234
  time_point::max() -> totally relaxed (but of course periodic mode will sync it earlier)

It makes sense to do this if it works out naturally, not if we have to add a bunch of ifs everywhere.

avikivity: I see that it's already halfway like this - force_sync::yes is time_point{0}. And maybe max() isn't the same as force_sync::no, because force_sync::no doesn't wait for the sync.

avikivity:
Looks reasonable, will await other reviews.

db/commitlog/commitlog.cc
elcallio: They will be woken up at the earliest time someone forces flush to happen. My assumption is that the wait time is a max latency we are willing to add, but if things move along before that we are happy.

db/commitlog/commitlog_entry.hh
elcallio: Indeed.

db/commitlog/commitlog.cc
avikivity: Yes, makes sense.

avikivity:
@gleb-cloudius please review.

elcallio:
ping?

gleb-cloudius (approved):

db/commitlog/commitlog.cc
```diff
@@ -1340,6 +1345,9 @@ future<db::commitlog::segment_manager::sseg_ptr> db::commitlog::segment_manager:
                 : make_ready_future<uint64_t>(0)
                 ;
 
+            // if we always overwrite, it should be fine to use this alignment...
```
gleb-cloudius, line 1348: Why is this?

elcallio: According to @avikivity, if the file is pre-written (we only overwrite), we can be allowed to write using the smaller alignment (512 - which should be its own constant - but...). 
Note that the initial write (zeros) still use 4096.

gleb-cloudius:
On Sun, Jan 10, 2021 at 11:54:46PM -0800, Calle Wilund wrote:
> @elcallio commented on this pull request.
> 
> 
> 
> > @@ -1340,6 +1345,9 @@ future<db::commitlog::segment_manager::sseg_ptr> db::commitlog::segment_manager:
>                  : make_ready_future<uint64_t>(0)
>                  ;
>  
> +            // if we always overwrite, it should be fine to use this alignment...
> 
> According to @avikivity, if the file is pre-written (we only overwrite), we can be allowed to write using the smaller alignment (512 - which should be its own constant - but...). 
I wonder what is the logic behind that. Also how universal it is.
If it is FS specific may be another parameter overwrite_alignment is
needed (not for this patch of course).

> Note that the initial write (zeros) still use 4096. 
> 
Yes, I saw that.

--
			Gleb.

db/commitlog/commitlog.cc
avikivity: This ended up incorrect. We need disk_write_dma_alignment still.

elcallio: How unfortunate. Glö.

avikivity: I'm patching Seastar to supply a lower d_w_d_a when possible.

elcallio: removed.

kostja:
* Calle Wilund <notifications@github.com> [20/12/01 22:01]:

We will be taking this to a test to see if it provides a
noticeable improvement in LWT performance.

-- 
https://scylladb.com

elcallio:
I will settle for measurable.

alecco:
For LWT this doesn't help much. Testing cqlstress-lwt-example.yaml (with STCS).

If we set force_sync 10ms wait the test crawls 10x slower and the DB ends up with timeout errors.

If we set force_sync 1ms wait the test runs with similar performance but there is no significant reduction of commitlog writes.

Reducing the underlying storage block size to 1KB does reduce the size. Here are the scylla_commitlog_bytes_written stats for master, master with 1KB (hardcoded commitlog alignment 1KB), and two runs of 7730 with 1KB (auto-detect) and 1ms wait.

```
metrics_master_4k_n_10000000:               211.806052352 - 100%
metrics_master_1k_n_10000000:               161.721262080 -  76%
metrics_7730_1k_n_10000000_wait_1ms:        161.602186240 -  76%
metrics_7730_1k_n_10000000_wait_1ms_b:      161.786242048 -  76%
```

On the other hand, having storage block size auto-detection is a good thing.

elcallio:
Updated to use overwrite alignment + update on file wrapping (extension).

elcallio:
le ping?

db/commitlog/commitlog.cc
```diff
@@ -396,6 +397,21 @@ class db::commitlog::segment_manager : public ::enable_shared_from_this<segment_
...
 
+    future<> force_sync(time_point latest) {
+        if (_timer.get_timeout() < latest) {
+            return make_ready_future<>();
+        }        
+        
+        _timer.cancel();
+        
+        if (clock_type::now() < latest) {
+            on_timer();
```
avikivity, line 408: No need for this, if the timer is already expired by the time you arm it, it will fire next reactor poll. I think it's better to have fewer paths into the timer callback.

db/commitlog/commitlog.cc
```diff
@@ -1881,6 +1909,9 @@ void db::commitlog::segment_manager::on_timer() {
         if (cfg.mode != sync_mode::BATCH) {
             sync();
         }
+
+        _sync_semaphore.signal(_sync_semaphore.waiters());
+
```
avikivity, line 1925: Adding to the conversation from Dec 2020: I think it's better for the timer to wake a sleeping coroutine (or do_until()) loop. This gets rid of the gate, and serializes timer invocations. The main service can wait for that coroutine to complete and so avoid use-after-frees. For a future cleanup of course.

db/commitlog/commitlog.cc
```diff
@@ -1881,6 +1909,9 @@ void db::commitlog::segment_manager::on_timer() {
         if (cfg.mode != sync_mode::BATCH) {
             sync();
         }
+
+        _sync_semaphore.signal(_sync_semaphore.waiters());
```
avikivity, line 1924: I don't understand how you can release _sync_semaphore without waiting on some from sync().

db/commitlog/commitlog_entry.cc
```diff
@@ -36,6 +36,9 @@
 
 #include <seastar/core/simple-stream.hh>
 
+const detail::force_sync detail::force_sync::no;
+const detail::force_sync detail::force_sync::yes(db::timeout_clock::time_point{});
```
avikivity, line 40: time_point::min()? it's unlikely that 0 will be after the current time, but better to make it clear that we want yes to be the earliest time possible.

We might as well use the periodic timeout instead of no, and so fold periodic mode into this. We have too many paths.

db/commitlog/commitlog.cc
```diff
@@ -1020,6 +1045,8 @@ class db::commitlog::segment : public enable_shared_from_this<segment>, public c
 
         if (_segment_manager->cfg.mode == sync_mode::BATCH || writer->sync) {
             return batch_cycle(timeout).discard_result();
+        } else if (writer->sync) {
+            return force_sync(*writer->sync, high);
```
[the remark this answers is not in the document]
elcallio: Hm, must be a rebase error. Should be 
`_segment_manager->cfg.mode == sync_mode::BATCH || writer->sync == force_sync::yes`

test/boost/commitlog_test.cc
```diff
@@ -205,6 +205,29 @@ SEASTAR_TEST_CASE(test_commitlog_new_segment){
     });
 }
 
+SEASTAR_TEST_CASE(test_commitlog_new_segment_512_align){
+    commitlog::config cfg;
+    cfg.commitlog_segment_size_in_mb = 1;
+    cfg.disk_write_alignment = 512;
```
avikivity, line 211: This may fail on a test system that requires larger alignment (e.g. sector size = 4096).

db/commitlog/commitlog.cc
elcallio: Depends on whether we assume this is signalled when sync is initiated or done. But I see your point. Can change it to be signalled on sync completion.

test/boost/commitlog_test.cc
elcallio: True. Should probably change the code in segment alloc to use a max of config and segment alignment. Otoh, makes the config parameter rather pointless. So maybe just kill it?

elcallio: Killed it.

elcallio:
Updated.

db/commitlog/commitlog.cc
elcallio: Done

db/commitlog/commitlog.cc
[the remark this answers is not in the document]
elcallio: Done

db/commitlog/commitlog.cc
elcallio: Removed

db/commitlog/commitlog_entry.cc
elcallio: min is probably better, though I am pretty sure it is 0 in this (and most) case(s). Don't understand "use the periodic timeout" part? Fold what how?
Too many paths, yes. This particular part of the code is yet another bandaid in our somewhat disorganized attempts to decide what we actually need out of the module to make it work with all the changing requirements. Would be great to eventually simplify. But the "no" case is literally equivalent to the old "false" case, so I don't really see that it adds any "new" complexity. The non yes/no does somewhat, but fairly benignly imho.

avikivity: Let's see what sync modes we have now:
1. sync now no matter what
2. sync when we have 128k worth of buffer, or 10s elapsed, whichever came sooner
3. sync when we have min_alignment, or writer->sync elapsed

I was thinking of folding 2 and 3, and specify 1 in terms of 2/3 (timeout == time_point::min(), so always elapsed).

db/commitlog/commitlog_entry.hh
```diff
@@ -37,9 +37,35 @@ public:
...
+    bool operator==(const force_sync& v) const {
+        return std::equal_to<opt_type>{}(*this, v);
+    }
+    bool operator!=(const force_sync& v) const {
+        return !(*this == v);
+    }
+
+    static const force_sync no;
+    static const force_sync yes;
```
avikivity, line 62: I don't think we need the optional, it just adds a lot of branches. We can use time_point::min() to force immediate sync, time_point::max() not to force any time-based sync, and a point in time to force sync at that point. Moreover we can use the timer everywhere with that, we don't have to check everywhere for min/max.

db/commitlog/commitlog_entry.cc
elcallio: Problem perhaps is that (1) is the batch mode, which has its own weird rules, and does not use/force the timer task. D you mean to move this into it simply forcing the timer to run? Does that not potentially add latency for the batch case?
Moreover, batch has its own rules, and when in place, does never even use the timer path. So should it remain such?

I am somewhat unsure on the term "fold" here. (2) and (3) are just two ways the timer/perioding sync is issued, except in (2) it is also done by the allocating code path directly. Is your suggestion to change _all_ paths that need to sync things to make the timer run? 
Since we only use one segment at a time, this is should be ok:ish, we won't really force any premature sync/flush-to-disk on segments (if we do, they are already non-allocating, and won't add any wasted space because of it - just ensure disk flushing).

db/commitlog/commitlog_entry.hh
elcallio: How does it add branches? You would have the same amount of branches to check "is min", "is max" etc. It is just semantics.
And we can't very well use the timer with "max" (as in arm it at), because that is not what we meant... Comparing to next trigger time, yes.

elcallio:
Too speed things up, I cherry-picked the disk alignment change into a separate PR, #8935. The timeout parts are perhaps less urgent, since afaik LWT team decided they did not really need it right now?

db/commitlog/commitlog_entry.hh
avikivity: I mean not having checks for is_min/is_max at all.

Why can't we arm the timer at time_point::max()? It arms a time_point that will never be reached (at least while we're alive), so it's the same as no_timeout. Similarly a timer at time_point::min() will expire immediately (or at least, at the next poll period).

elcallio: Because that is not what "max" (or min, or timeout) means in the context. It means "not before, but respect periodicity". 
So either way we at the very least need to do a `std::min(timeout, active_timeout)`.

bhalevy:
Is this PR still applicable or can it be closed?

mykaul:
> Is this PR still applicable or can it be closed?

@elcallio  ?

elcallio:
Since apparently not needed, can be closed.

PR #7739, opened 2020-12-02
gossip: Added SNITCH_NAME to `application_state`

### LINKED ISSUE #6832: Do not allow a node to join with a different snitch

jul-stas:
Snitch name needs to be exchanged within cluster once, on shadow
round, so joining nodes cannot use wrong snitch. The snitch names
are compared on bootstrap and on normal node start.

If the cluster already used mixed snitches, the upgrade to this
version will fail. In this case customer needs to add a node with
correct snitch for every node with the wrong snitch, then put
down the nodes with the wrong snitch and only then do the upgrade.

Fixes #6832

### DIFF
--- a/gms/gossiper.cc
+++ b/gms/gossiper.cc
@@ -1757,7 +1757,8 @@ future<> gossiper::do_shadow_round(std::unordered_set<gms::inet_address> nodes, 
             gms::application_state::STATUS,
             gms::application_state::HOST_ID,
             gms::application_state::TOKENS,
-            gms::application_state::SUPPORTED_FEATURES}};
+            gms::application_state::SUPPORTED_FEATURES,
+            gms::application_state::SNITCH_NAME}};
         logger.info("Gossip shadow round started with nodes={}", nodes);
         std::unordered_set<gms::inet_address> nodes_talked;
         size_t nodes_down = 0;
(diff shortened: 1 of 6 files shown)


gms/gossiper.cc
```diff
@@ -2326,6 +2327,20 @@ void gossiper::check_knows_remote_features(std::set<std::string_view>& local_fea
     }
 }
 
+void gossiper::check_snitch_name_matches(const std::unordered_set<gms::inet_address>& initial_contact_nodes, const std::unordered_set<gms::inet_address>& loaded_endpoints) {
+    for (const auto& contact_point : boost::range::join(initial_contact_nodes, loaded_endpoints)) {
```
asias, line 2331: We need to check all the nodes in endpoint_state_map. For example, when a new node joins, the initial_contact_nodes can be a single node and loaded_endpoints is empty. endpoint_state_map contains all the nodes in the cluster. We should check them all.

jul-stas: I updated it to use `endpoint_state_map`; @asias please have a look if this is what you asked for.

gms/gossiper.cc
```diff
@@ -2326,6 +2327,20 @@ void gossiper::check_knows_remote_features(std::set<std::string_view>& local_fea
...
 }
 
+void gossiper::check_snitch_name_matches() {
+    for (const auto& [address, state] : endpoint_state_map) {
+        const auto remote_snitch_name = state.get_application_state_ptr(application_state::SNITCH_NAME);
+        if (!remote_snitch_name) {
+            continue;
+        }
+
+        const auto& my_snitch_name = locator::i_endpoint_snitch::get_local_snitch_ptr()->get_name();
```
asias, line 2337: nit: it can be moved outside the loop.

gms/gossiper.cc
asias: I had one minor comment. Otherwise looks good to me. Thanks.

gms/gossiper.cc
jul-stas: Oh, sure. Fixed.

PR #7765, opened 2020-12-07
tools: toolchain: add node_exporter

### LINKED ISSUE #2190: Node_exporter is not part of Scylla packaging, failing to install in a connection-less environment
Scylla version (or git commit hash): Scylla 1.6.1
OS (RHEL/CentOS/Ubuntu/AWS AMI): Centos 7.3

Installing Scylla in a connection-less environment (no connection to the outside world), Scylla_setup invokes the script: node_exporter_install
which try to pull from github the relevant prometheus/node_exporter package:
curl -L https://github.com/prometheus/node_exporter/releases/download/$version/node_exporter-$version.linux-amd64.tar.gz -o $dir/node_exporter-$version.linux-amd64.tar.gz

and fails. 
It will be useful to package the node_exporter with the other scylla packages.

syuu1228:
tools: toolchain: add node_exporter

Download node_exporter in frozen image to prepare adding node_exporter
to relocatable pacakge.

Related #2190

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 #
 # This file is open source software, licensed to you under the terms
 # of the Apache License, Version 2.0 (the "License").  See the NOTICE file
@@ -91,6 +91,7 @@ fedora_packages=(
     fakeroot
     file
     dpkg-dev
+    curl
 )
 
 fedora_python3_packages=(
@@ -140,18 +141,54 @@ arch_packages=(
     thrift
 )
 
+NODE_EXPORTER_VERSION=1.0.1
+declare -A NODE_EXPORTER_CHECKSUM=(
+    ["x86_64"]=3369b76cd2b0ba678b6d618deab320e565c3d93ccb5c2a0d5db51a53857768ae
+    ["aarch64"]=017514906922fcc4b7d727655690787faed0562bc7a17aa9f72b0651cb1b47fb
+    ["s390x"]=2f22d1ce18969017fb32dbd285a264adf3da6252eec05f03f105cf638ec0bb06
+)
+declare -A NODE_EXPORTER_ARCH=(
+    ["x86_64"]=amd64
+    ["aarch64"]=arm64
+    ["s390x"]=s390x
+)
+NODE_EXPORTER_DIR=/opt/scylladb/dependencies
+
+node_exporter_filename() {
+    echo "node_exporter-$NODE_EXPORTER_VERSION.linux-${NODE_EXPORTER_ARCH["$(arch)"]}.tar.gz"
+}
+
+node_exporter_fullpath() {
+    echo "$NODE_EXPORTER_DIR/$(node_exporter_filename)"
+}
+
+node_exporter_checksum() {
+    sha256sum "$(node_exporter_fullpath)" | while read -r sum _; do [[ "$sum" == "${NODE_EXPORTER_CHECKSUM["$(arch)"]}" ]]; done
+}
+
+node_exporter_url() {
+    echo "https://github.com/prometheus/node_exporter/releases/download/v$NODE_EXPORTER_VERSION/$(node_exporter_filename)"
+}
+
+
 print_usage() {
     echo "Usage: install-dependencies.sh [OPTION]..."
     echo ""
     echo "  --print-python3-runtime-packages Print required python3 packages for Scylla"
+    echo "  --print-node-exporter-filename Print node_exporter filename"
     exit 1
 }
 
 PRINT_PYTHON3=false
+PRINT_NODE_EXPORTER=false
 while [ $# -gt 0 ]; do
     case "$1" in
         "--print-python3-runtime-packages")
             PRINT_PYTHON3=true
+            shift 1
+            ;;
+        "--print-node-exporter-filename")
+            PRINT_NODE_EXPORTER=true
             shift 1
             ;;
          *)
@@ -166,6 +203,11 @@ if $PRINT_PYTHON3; then
         exit 1
     fi
     echo "${fedora_python3_packages[@]}"
+    exit 0
+fi
+
+if $PRINT_NODE_EXPORTER; then
+    node_exporter_fullpath
     exit 0
 fi
 
@@ -194,6 +236,17 @@ elif [ "$ID" = "fedora" ]; then
     fi
     yum install -y "${fedora_packages[@]}" "${fedora_python3_packages[@]}"
     pip3 install cassandra-driver
+
+    if [ -f "$(node_exporter_fullpath)" ] && node_exporter_checksum; then
+        echo "$(node_exporter_filename) already exists, skipping download"
+    else
+        mkdir -p "$NODE_EXPORTER_DIR"
+        curl -fSL -o "$(node_exporter_fullpath)" "$(node_exporter_url)"
+        if ! node_exporter_checksum; then
+            echo "$(node_exporter_filename) download failed"
+            exit 1
+        fi
+    fi
 elif [ "$ID" = "centos" ]; then
     yum install -y "${centos_packages[@]}"
     echo -e "Configure example:\n\tpython3.4 ./configure.py --enable-dpdk --mode=release --static-boost --compiler=/opt/scylladb/bin/g++-7.3 --python python3.4 --ldflag=-Wl,-rpath=/opt/scylladb/lib64 --cflags=-I/opt/scylladb/include --with-antlr3=/opt/scylladb/bin/antlr3"


scripts/node_exporter_download.py
```diff
@@ -0,0 +1,43 @@
...
+
+VERSION='1.0.1'
+DEST_DIR='{}/../build/node_exporter'.format(os.path.dirname(__file__))
+
+if __name__ == '__main__':
+    if os.path.exists(DEST_DIR):
+        shutil.rmtree(DEST_DIR)
+    os.makedirs(DEST_DIR)
+    r = requests.get(f'https://github.com/prometheus/node_exporter/releases/download/v{VERSION}/node_exporter-{VERSION}.linux-amd64.tar.gz', timeout=5)
+    bo = io.BytesIO(r.content)
```
avikivity, line 38: I don't know that I want a rule that uses the network in our build system. Yes, we have them in jmx and java, but let's not add more.

syuu1228: Hmm, I thought building from source can be the solution for this, but I found that node_exporter and prometheus build system are actually using network to fetch go dependencies (like our java project)
We may have to add prebuilt binary to this repo if we really want to make it completely offline.

avikivity: Or we may require the user to fetch and install node_exporter themselves.

syuu1228: How is download node_exporter on dbuild docker image building time, instead of source code building time?
As we already do `pip3 install cassandra-driver` for python module, we also can `go get  github.com/prometheus/node_exporter@v1.0.1` to install node_exporter on dbuild Docker image (stored on ~/go).
Once it installed, we can just copy it on building time, no need network.

Additionally, we may able to do same thing for Java, by resolving all maven dependencies (stored on ~/.m2).
referenced: https://stackoverflow.com/questions/42208442/maven-docker-cache-dependencies

syuu1228: How about following change to add node_exporter on dbuild image:
```
diff --git a/install-dependencies.sh b/install-dependencies.sh
index 62c84c821..7ff717235 100755
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -91,6 +91,7 @@ fedora_packages=(
     fakeroot
     file
     dpkg-dev
+    golang
 )
 
 fedora_python3_packages=(
@@ -194,6 +195,7 @@ elif [ "$ID" = "fedora" ]; then
     fi
     yum install -y "${fedora_packages[@]}" "${fedora_python3_packages[@]}"
     pip3 install cassandra-driver
+    GO111MODULE=on go get github.com/prometheus/node_exporter@v1.0.1
 elif [ "$ID" = "centos" ]; then
     yum install -y "${centos_packages[@]}"
     echo -e "Configure example:\n\tpython3.4 ./configure.py --enable-dpdk --mode=release --static-boost --compiler=/opt/scylladb/bin/g++-7.3 --python python3.4 --ldflag=-Wl,-rpath=/opt/scylladb/lib64 --cflags=-I/opt/scylladb/include --with-antlr3=/opt/scylladb/bin/antlr3"
diff --git a/tools/toolchain/Dockerfile b/tools/toolchain/Dockerfile
index 9f5782d31..7ac1a8ff3 100644
--- a/tools/toolchain/Dockerfile
+++ b/tools/toolchain/Dockerfile
@@ -6,10 +6,11 @@ RUN dnf -y install 'dnf-command(copr)' \
     && dnf -y install ccache \
     && dnf -y install gdb \
     && dnf -y install devscripts debhelper fakeroot file rpm-build \
-    && /bin/bash -x ./install-dependencies.sh && dnf -y update && dnf clean all \
+    && env GOPATH=/go /bin/bash -x ./install-dependencies.sh && dnf -y update && dnf clean all \
     && echo 'ALL ALL=(ALL:ALL) NOPASSWD: ALL' >> /etc/sudoers \
     && cp system-auth /etc/pam.d \
     && echo 'Defaults !requiretty' >> /etc/sudoers
 RUN mkdir -p /root/.m2/repository
 ENV JAVA8_HOME=/usr/lib/jvm/java-1.8.0-openjdk
+ENV GOPATH=/go
 CMD /bin/bash
```

avikivity: It means the build won't work outside dbuild (unless you run the same command).

We could move it to install-dependencies.sh (and then we can just download the package in install-dependencies.sh and store it in /opt/scylladb/dependencies or something).

avikivity: I guess that's similar to what we do with python-cassandra-driver.

syuu1228: > It means the build won't work outside dbuild (unless you run the same command).

Right, and I'm trying to avoid breaking build outside of dbuild, something like this:
`go get` will download the package on `$GOPATH` if it's defined, otherwise it will download to `$HOME/go`.
So we can define both pattern in create-relocatable-package.py:
```
if os.getenv('GOPATH'):
    node_exporter = '{}/bin/node_exporter'.format(os.getenv('GOPATH'))
else:
    node_exporter = os.path.expanduser('~/go/bin/node_exporter')
ar.reloc_add(node_exporter, arcname='node_exporter')
```

On this way, we can support all these three patterns:
 - In dbuild, GOPATH=/go
 - Outside of dbuild, GOPATH defined
 - Outside of dbuild, GOPATH not defined, installed to ~/go

Note that why I defined `ENV GOPATH=/go` in Dockerfile is, because install-dependencies.sh will run in root, but building package could run in non-root user, so root and non-root have to reference same directory.

avikivity: I don't want separate paths for dbuild and outside dbuild. But I think we can just curl the binary package and store it somewhere, we don't need to build from source.

Same as install-dependencies.sh doesn't build clang from source, it just gets a binary package. If one day we get node_exporter in Fedora, we can switch it to rpm.

syuu1228: Oh sorry, I was misunderstanding that `go get` just downloading source&binary, if it's compiling we should use just download binary tar.gz using curl instead.
How about something like this:
```
diff --git a/install-dependencies.sh b/install-dependencies.sh
index 62c84c821..4573dbe7b 100755
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -91,6 +91,7 @@ fedora_packages=(
     fakeroot
     file
     dpkg-dev
+    curl
 )
 
 fedora_python3_packages=(
@@ -140,6 +141,8 @@ arch_packages=(
     thrift
 )
 
+NODE_EXPORTER_VERSION=1.0.1
+
 print_usage() {
     echo "Usage: install-dependencies.sh [OPTION]..."
     echo ""
@@ -194,6 +197,10 @@ elif [ "$ID" = "fedora" ]; then
     fi
     yum install -y "${fedora_packages[@]}" "${fedora_python3_packages[@]}"
     pip3 install cassandra-driver
+    curl -fSL -o build/node_exporter.tar.gz https://github.com/prometheus/node_exporter/releases/download/v$NODE_EXPORTER_VERSION/node_exporter-$NODE_EXPORTER_VERSION.linux-amd64.tar.gz
+    tar -C build -xpf build/node_exporter.tar.gz
+    rm -rf build/node_exporter
+    mv build/node_exporter-$NODE_EXPORTER_VERSION.linux-amd64 build/node_exporter
 elif [ "$ID" = "centos" ]; then
     yum install -y "${centos_packages[@]}"
     echo -e "Configure example:\n\tpython3.4 ./configure.py --enable-dpdk --mode=release --static-boost --compiler=/opt/scylladb/bin/g++-7.3 --python python3.4 --ldflag=-Wl,-rpath=/opt/scylladb/lib64 --cflags=-I/opt/scylladb/include --with-antlr3=/opt/scylladb/bin/antlr3"
```

avikivity: We can't use build/ in install-dependencies.sh, it doesn't exist yet. . Also, we don't know which directory the user will use after ./install-dependencies is run (I have two scylla directories on my machine). We can store it in /opt/scylladb/dependencies.

syuu1228:
@avikivity Pushed new version of the patch, splited it frozen image part and packaging part.
I think we cannot add packaging part until frozen image updated, otherwise build will break since there is no node_exporter image in frozen image yet.

avikivity:
Yes.

syuu1228:
Ah seems like I'm breaking something, please don't merge this until I fixed it

install-dependencies.sh
```diff
@@ -140,20 +141,50 @@ arch_packages=(
...
 )
 
+NODE_EXPORTER_VERSION=1.0.1
+declare -A NODE_EXPORTER_CHECKSUM=(
+    ["x86_64"]=3369b76cd2b0ba678b6d618deab320e565c3d93ccb5c2a0d5db51a53857768ae
+    ["aarch64"]=017514906922fcc4b7d727655690787faed0562bc7a17aa9f72b0651cb1b47fb
+)
+declare -A NODE_EXPORTER_ARCH=(
+    ["x86_64"]=amd64
+    ["aarch64"]=arm64
```
avikivity, line 151: Please also add s390x, I plan to add it to the toolchain on the next build.

syuu1228:
@avikivity The bug fixed (I mistakenly didn't committed node_exporter_url(), added now), also added s390x arch.

syuu1228:
Sent PR for packaging part: #7796

install-dependencies.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
```
avikivity, line 1: I think this will break, since we have calls to tools/java/install-dependencies.sh which don't exist.

avikivity: It does not break.

PR #7766, opened 2020-12-07
api: Add cardinality to toppartitions results

### LINKED ISSUE #4089: toppartitions API should return cardinality
The API for toppartitions should also return the cardinality.
As explained in the documentation:
the number of unique operations in the sample set

StarostaGit:
This change enhances the toppartitions api to also return
the cardinality of the read and write sample sets. It now uses
the size() method of space_saving_top_k class, counting the unique
operations in the sampled set for up to the given capacity.

Fixes #4089

### DIFF
--- a/api/api-doc/column_family.json
+++ b/api/api-doc/column_family.json
@@ -2925,12 +2925,20 @@
          "id":"toppartitions_query_results",
          "description":"nodetool toppartitions query results",
          "properties":{
+            "read_cardinality":{
+               "type":"long",
+               "description":"Number of the unique operations in the sample set"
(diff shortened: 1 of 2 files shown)


StarostaGit:
Now that we have this on the api side, we should change nodetool to use what's returned, instead of counting the cardinality itself. I'll make an issue for that and follow up

StarostaGit:
Didn't commit one file accidentally, now it's all there

avikivity:
Ugh, I assumed the new variables were already in IDL and didn't bother to check.

PR #7767, opened 2020-12-07
Reinstate [[nodiscard]] support

avikivity:
The switch to clang disabled the clang-specific -Wunused-value
since it generated some harmless warnings. Unfortunately, that also
prevent [[nodiscard]] violations from warning.
    
Fix by clearing all instances of the warning (including [[nodiscard]]
violations that crept in while it was disabled) and reinstating the warning.


denesb (approved):

avikivity:
@penberg ping

PR #7770, opened 2020-12-07
types: don't linearize in validate()

michoecho:
A sequel to #7692.

This series gets rid of linearization when validating collections and tuple types. (Other types were already validated without linearizing).
The necessary helpers for reading from fragmented buffers were introduced in #7692. All this series does is put them to use in `validate()`.

Refs: #6138

### DIFF
--- a/types.cc
+++ b/types.cc
@@ -1586,9 +1586,13 @@ struct validate_visitor {
 };
 }
 
-void abstract_type::validate(const fragmented_temporary_buffer::view& view, cql_serialization_format sf) const {
-    visit(*this, validate_visitor<fragmented_temporary_buffer::view>{view, sf});
+template <FragmentedView View>
+void abstract_type::validate(const View& view, cql_serialization_format sf) const {
+    visit(*this, validate_visitor<View>{view, sf});
 }
+// Explicit instantiation.
+template void abstract_type::validate<>(const single_fragmented_view&, cql_serialization_format) const;
+template void abstract_type::validate<>(const fragmented_temporary_buffer::view&, cql_serialization_format) const;
 
 void abstract_type::validate(bytes_view v, cql_serialization_format sf) const {
     visit(*this, validate_visitor<single_fragmented_view>{single_fragmented_view(v), sf});
--- a/types.cc
+++ b/types.cc
@@ -1413,14 +1413,14 @@ static void validate_aux(const tuple_type_impl& t, bytes_view v, cql_serializati
 }
 
 namespace {
-template <typename UnderlyingView, typename FragmentRangeView>
+template <FragmentedView View>
 struct validate_visitor {
-    UnderlyingView underlying;
-    FragmentRangeView v;
-
+    const View& v;
     cql_serialization_format sf;
 
-    void operator()(const reversed_type_impl& t) { return t.underlying_type()->validate(underlying, sf); }
+    void operator()(const reversed_type_impl& t) {
+        visit(*t.underlying_type(), validate_visitor<View>{v, sf});
+    }
     void operator()(const abstract_type&) {}
     template <typename T> void operator()(const integer_type_impl<T>& t) {
         if (v.empty()) {
@@ -1447,9 +1447,9 @@ struct validate_visitor {
         }
     }
     void operator()(const ascii_type_impl&) {
-        // ASCII can be validated without linearization
-        for (auto& frag : v) {
-            if (!utils::ascii::validate(frag)) {
+        // ASCII can be validated independently for each fragment
+        for (View fv = v; fv.size_bytes() > 0; fv.remove_current()) {
+            if (!utils::ascii::validate(fv.current_fragment())) {
                 throw marshal_exception("Validation failed - non-ASCII character in an ASCII string");
             }
         }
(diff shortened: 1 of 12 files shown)


avikivity:
Very good. I'd like the same single-fragment optimization trick so that if a validation needs to call read_simple several times, it uses the simple version using the single fragment view rather than the multi-fragment view.

types.cc
```diff
@@ -1423,23 +1423,23 @@ struct validate_visitor {
     void operator()(const reversed_type_impl& t) { return t.underlying_type()->validate(underlying, sf); }
     void operator()(const abstract_type&) {}
     template <typename T> void operator()(const integer_type_impl<T>& t) {
-        if (v.empty()) {
+        if (v.size_bytes() == 0) {
```
denesb, line 1429: Why not add `empty()` to `FragmentView` instead?

types.cc
```diff
@@ -1401,16 +1401,15 @@ tuple_type_impl::get_instance(std::vector<data_type> types) {
...
-        if (*vi) {
-            (*ti)->validate(**vi, sf);
+    while (ti != t.all_types().end() && v.size_bytes()) {
+        std::optional<View> e = read_tuple_element(v);
+        if (e) {
+            (*ti)->validate(*e, sf);
         }
         ++ti;
-        ++vi;
     }
```
denesb, line 1415: Unrelated: we should check here that there is neither types nor bytes left.

types.cc
```diff
@@ -1401,16 +1401,15 @@ tuple_type_impl::get_instance(std::vector<data_type> types) {
     return intern::get_instance(std::move(types));
 }
 
-static void validate_aux(const tuple_type_impl& t, bytes_view v, cql_serialization_format sf) {
+template <FragmentedView View>
+static void validate_aux(const tuple_type_impl& t, View v, cql_serialization_format sf) {
     auto ti = t.all_types().begin();
-    auto vi = tuple_deserializing_iterator::start(v);
-    auto end = tuple_deserializing_iterator::finish(v);
```
denesb, line 1405: Can `tuple_deserializing_iterator` be removed now?

types.cc
```diff
@@ -1328,15 +1328,17 @@ list_type_impl::is_value_compatible_with_frozen(const collection_type_impl& prev
...
-static void validate_aux(const list_type_impl& t, bytes_view v, cql_serialization_format sf) {
+template <FragmentedView View>
+static void validate_aux(const list_type_impl& t, View v, cql_serialization_format sf) {
     auto nr = read_collection_size(v, sf);
     for (int i = 0; i != nr; ++i) {
         t.get_elements_type()->validate(read_collection_value(v, sf), sf);
     }
-    if (!v.empty()) {
+    if (v.size_bytes()) {
+        auto hex = with_linearized(v, [] (bytes_view bv) { return to_hex(bv); });
```
denesb, line 1338: This might trigger a `std::bad_alloc` before we could throw `marshal_exception`. I think it is a bad idea to include the remains as it can be arbitrarily large. This is a pre-existing problem as we linearized before as well, on a higher level, so it can be done in a follow-up.

denesb (commented):
So good to see all these linearizations gone!

types.cc
avikivity: Yes, I suppose we can include a few bytes before the problem area, bytes at the problem area, and a position indicator. We should be generous so that most inputs will be fully included, but no so generous that a 10MB map<> causes bad_allocs.

michoecho:
Added single-fragment special cases for collections and strings.

types.cc
michoecho: Not yet. However, there are only two other uses in the codebase, and they can be trivially replaced with `read_tuple_element()`, so we can remove it soon. But since those uses are unrelated, let's do it in another patch.

types.cc
michoecho: Should I do that now? Since this is a refactoring/optimization series, maybe we shouldn't try to improve the semantics of verification here.

types.cc
michoecho: API orthogonality was my rationale. Whenever you have both `empty()` and `size_bytes()` available, you have one more decision to make in your life.
But if you think `empty()` is more natural or readable, just say a word and I will switch to it.

michoecho:
If you have a better name for this helper than `with_simplified`, do tell.

types.cc
denesb: If you want to, but it is also fine to send as a follow-up patch.

types.cc
denesb: ok

types.cc
denesb: Most containers have it, even though you can argue it is redundant if they have `size()`. I think it is more natural to read, and you avoid churn in the code too.

michoecho:
@avikivity Please don't requeue this series yet, I want to address https://github.com/scylladb/scylla/pull/7770#discussion_r540097331 first.

types.cc
avikivity: I thought the same, but what would empty() be? no fragments or size_bytes() == 0?

I guess the latter makes more sense.

michoecho: Is there a difference?

michoecho:
Changes:
- addressed https://github.com/scylladb/scylla/pull/7770#discussion_r540097331 by adding `empty()` to `FragmentedView`.
- pushed fix for a null-pointer-arithmetic undefined behaviour bug (introduced in #7692) discovered by UBSAN.

types.cc
michoecho: Done. Reverted back to `empty()`.

michoecho:
Okay, I think this PR is ready now.
https://github.com/scylladb/scylla/pull/7770#discussion_r538212426 was adressed, and https://github.com/scylladb/scylla/pull/7770#discussion_r538222728, https://github.com/scylladb/scylla/pull/7770#discussion_r538223429, https://github.com/scylladb/scylla/pull/7770#discussion_r538225302 are unrelated, and so left for another patch.

PR #7771, opened 2020-12-07
types: don't linearize in serialize_for_cql()

michoecho:
A sequel to #7692.

This series gets rid of linearization in `serialize_for_cql`, which serializes collections and user types from `collection_mutation_view` to CQL. We switch from `bytes` to `bytes_ostream` as the intermediate buffer type.

The only user of of `serialize_for_cql` immediately copies the result to another `bytes_ostream`. We could avoid some copies and allocations by writing to the final `bytes_ostream` directly, but it's currently hidden behind a template.

Before this series, `serialize_for_cql_aux()` delegated the actual writing to `collection_type_impl::pack` and `tuple_type_impl::build_value`, by passing them an intermediate `vector`. After this patch, the writing is done directly in `serialize_for_cql_aux()`. Pros: we avoid the overhead of creating an intermediate vector, without bloating the source code (because creating that intermediate vector requires just as much code as serializing the values right away). Cons: we duplicate the CQL collection format knowledge contained in `collection_type_impl::pack` and `tuple_type_impl::build_value`.

Refs: #6138

### DIFF
--- a/types.cc
+++ b/types.cc
@@ -3123,10 +3123,10 @@ static bytes_ostream serialize_for_cql_aux(const user_type_impl& type, collectio
     return out;
 }
 
-bytes serialize_for_cql(const abstract_type& type, collection_mutation_view v, cql_serialization_format sf) {
+bytes_ostream serialize_for_cql(const abstract_type& type, collection_mutation_view v, cql_serialization_format sf) {
     assert(type.is_multi_cell());
 
-    return linearized(v.with_deserialized(type, [&] (collection_mutation_view_description mv) {
+    return v.with_deserialized(type, [&] (collection_mutation_view_description mv) {
         return visit(type, make_visitor(
             [&] (const map_type_impl& ctype) { return serialize_for_cql_aux(ctype, std::move(mv), sf); },
             [&] (const set_type_impl& ctype) { return serialize_for_cql_aux(ctype, std::move(mv), sf); },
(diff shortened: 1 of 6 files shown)


avikivity:
We may want to standardize on managed_bytes in the future as our standard fragmented bytes type. If we do, then we can morph bytes_ostream to be a builder for managed_bytes.

types.cc
```diff
@@ -694,6 +694,32 @@ void write_collection_value(bytes::iterator& out, cql_serialization_format sf, b
     out = std::copy_n(val_bytes.begin(), val_bytes.size(), out);
 }
 
+// Passing the wrong integer type to a generic serialization function is a particularly
+// easy mistake to do, so we want to disable template parameter deduction here.
+// Hence std::type_identity.
```
denesb, line 699: Neat

types.cc
```diff
@@ -694,6 +694,32 @@ void write_collection_value(bytes::iterator& out, cql_serialization_format sf, b
...
+template<typename T>
+void write_simple(bytes_ostream& out, std::type_identity_t<T> val) {
+    auto val_be = net::hton(val);
+    auto val_ptr = reinterpret_cast<const bytes::value_type*>(&val_be);
+    out.write(bytes_view(val_ptr, sizeof(T)));
+}
+
+void write_collection_value(bytes_ostream& out, cql_serialization_format sf, data::value_view val) {
```
denesb, line 707: It is also possible to template `write_collection_value` on `OutputIterator` and support both `bytes::iterator` and `bytes_ostream::output_iterator`. Maybe not worth it is slower than this solution.

denesb (approved):

types.cc
avikivity: I expect the other write_collection_value() will just die.

PR #7784, opened 2020-12-11
utils: fragmented_temporary_buffer: don't add to null pointer

michoecho:
When fragmented_temporary_buffer::view is created from a bytes_view,
_current is null. In that case, in remove_current(), null pointer offset
happens, and ubsan complains. Fix that.


michoecho:
Nevermind. #7770 was dequeued, so I will close this and push the commit there.

PR #7785, opened 2020-12-11
chore: Separate Scylla install specific commands

hrvolapeter:
Use case: building for aarch64 will completely fail install scylla step. This makes it easier to figure out the install specific step which should be replaced


avikivity:
Hi, before we can accept contributions, we need a signed contributor's license agreement.

https://www.scylladb.com/open-source-nosql-database/contributor-agreement/

mykaul:
@hrvolapeter - will you be updating this PR, signing the CLA, etc?

PR #7807, opened 2020-12-17
Azure Ls v2 local disk setup

### LINKED ISSUE #8325: dist/common/scripts/scylla_util.py doesn't support Azure images
in https://github.com/scylladb/scylla-machine-image/issues/54 we need to add support to Azure image. 

I have created scylla Azure image and it fail with the following error:
```
azureuser@yaronkaikovTest:/opt/scylladb/scripts$ journalctl -xe
-- Defined-By: systemd
-- Support: http://www.ubuntu.com/support
-- 
-- A start job for unit scylla-image-setup.service has begun execution.
-- 
-- The job identifier is 1063.
Mar 21 12:56:34 yaronkaikovTest systemd-udevd[224]: Network interface NamePolicy= disabled on kernel command line, ignoring.
Mar 21 12:56:37 yaronkaikovTest scylla_image_setup[2244]: Traceback (most recent call last):
Mar 21 12:56:37 yaronkaikovTest scylla_image_setup[2244]:   File "/opt/scylladb/scylla-machine-image/libexec/scylla_image_setup", line 25, in <module>
Mar 21 12:56:37 yaronkaikovTest scylla_image_setup[2244]:     cloud_instance = get_cloud_instance()
Mar 21 12:56:37 yaronkaikovTest scylla_image_setup[2244]:   File "/opt/scylladb/scripts/scylla_util.py", line 565, in get_cloud_instance
Mar 21 12:56:37 yaronkaikovTest scylla_image_setup[2244]:     raise Exception("Unknown cloud provider! Only AWS/GCP supported.")
Mar 21 12:56:37 yaronkaikovTest scylla_image_setup[2244]: Exception: Unknown cloud provider! Only AWS/GCP supported.
Mar 21 12:56:37 yaronkaikovTest systemd[1]: scylla-image-setup.service: Main process exited, code=exited, status=1/FAILURE
-- Subject: Unit process exited
-- Defined-By: systemd
-- Support: http://www.ubuntu.com/support
```

tarzanek:
fixes #8325

The iotune tests happened on Centos 8.2 both with stock and elrepo kernel, using Scylla 4.3 rc3

results are in https://docs.google.com/spreadsheets/d/1_uYq8UxY47XF5jreetrpleykLPqNGjfPXIirvTPh6rk/edit#gid=1101336711

### DIFF
--- a/dist/common/scripts/scylla_io_setup
+++ b/dist/common/scripts/scylla_io_setup
@@ -272,12 +272,71 @@ if __name__ == "__main__":
                     disk_properties["write_bandwidth"] = 4663037952
                     #below is google, above is our measured
                     #disk_properties["write_bandwidth"] = 4680 * mbs
-                properties_file = open(etcdir() + "/scylla.d/io_properties.yaml", "w")
-                yaml.dump({"disks": [disk_properties]}, properties_file,  default_flow_style=False)
-                ioconf = open(etcdir() + "/scylla.d/io.conf", "w")
-                ioconf.write("SEASTAR_IO=\"--io-properties-file={}\"\n".format(properties_file.name))
+                if "read_iops" in disk_properties:
+                    properties_file = open(etcdir() + "/scylla.d/io_properties.yaml", "w")
+                    yaml.dump({"disks": [disk_properties]}, properties_file,  default_flow_style=False)
+                    ioconf = open(etcdir() + "/scylla.d/io.conf", "w")
+                    ioconf.write("SEASTAR_IO=\"--io-properties-file={}\"\n".format(properties_file.name))
+                else:
+                    logging.error(
+                        'Did not detect number of disks in Google Cloud instance setup for auto local disk tuning, running manual iotune.')
+                    run_iotune()
             else:
-                logging.error('This is not a recommended Google Cloud instance setup for auto tuning, running manual iotune.')
+                logging.error(
+                    'This is not a recommended Google Cloud instance setup for auto local disk tuning, running manual iotune.')
+                run_iotune()
+        elif azure_instance().is_azure_instance():
+            idata = azure_instance()
+            if idata.is_recommended_instance():
+                disk_properties = {}
+                disk_properties["mountpoint"] = datadir()
+                nr_disks = idata.nvmeDiskCount
+                # below is based on https://docs.microsoft.com/en-us/azure/virtual-machines/lsv2-series
+                # note that scylla iotune might measure more, this is Azure recommended
+                # since write properties are not defined, they come from our iotune tests
+                mbs = 1024*1024
+                if nr_disks == 1:
+                    disk_properties["read_iops"] = 400000
+                    disk_properties["read_bandwidth"] = 2000 * mbs
+                    disk_properties["write_iops"] = 271696
+                    disk_properties["write_bandwidth"] = 1314 * mbs
+                elif nr_disks == 2:
+                    disk_properties["read_iops"] = 800000
+                    disk_properties["read_bandwidth"] = 4000 * mbs
+                    disk_properties["write_iops"] = 552434
+                    disk_properties["write_bandwidth"] = 2478 * mbs
+                elif nr_disks == 4:
+                    disk_properties["read_iops"] = 1500000
+                    disk_properties["read_bandwidth"] = 8000 * mbs
+                    disk_properties["write_iops"] = 1105063
+                    disk_properties["write_bandwidth"] = 4948 * mbs
+                elif nr_disks == 6:
+                    disk_properties["read_iops"] = 2200000
+                    disk_properties["read_bandwidth"] = 14000 * mbs
+                    disk_properties["write_iops"] = 1616847
+                    disk_properties["write_bandwidth"] = 7892 * mbs
+                elif nr_disks == 8:
+                    disk_properties["read_iops"] = 2900000
+                    disk_properties["read_bandwidth"] = 16000 * mbs
+                    disk_properties["write_iops"] = 2208081
+                    disk_properties["write_bandwidth"] = 9694 * mbs
+                elif nr_disks == 10:
+                    disk_properties["read_iops"] = 3800000
+                    disk_properties["read_bandwidth"] = 20000 * mbs
+                    disk_properties["write_iops"] = 2546511
+                    disk_properties["write_bandwidth"] = 11998 * mbs
+                if "read_iops" in disk_properties:
+                    properties_file = open(etcdir() + "/scylla.d/io_properties.yaml", "w")
+                    yaml.dump({"disks": [disk_properties]}, properties_file,  default_flow_style=False)
+                    ioconf = open(etcdir() + "/scylla.d/io.conf", "w")
+                    ioconf.write("SEASTAR_IO=\"--io-properties-file={}\"\n".format(properties_file.name))
+                else:
+                    logging.error(
+                        'Did not detect number of disks in Azure Cloud instance setup for auto local disk tuning, running manual iotune.')
+                    run_iotune()
+            else:
+                logging.error(
+                    'This is not a recommended Azure Cloud instance setup for auto local disk tuning, running manual iotune.')
                 run_iotune()
         else:
             run_iotune()
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -17,6 +17,7 @@
 
 import configparser
 import io
+import json
 import logging
 import os
 import re
@@ -29,6 +30,7 @@ import urllib.parse
 import urllib.request
 import yaml
 import psutil
+import socket
 import sys
 from pathlib import Path, PurePath
 from subprocess import run, DEVNULL
@@ -137,7 +139,6 @@ class gcp_instance:
     @staticmethod
     def is_gce_instance():
         """Check if it's GCE instance via DNS lookup to metadata server."""
-        import socket
         try:
             addrlist = socket.getaddrinfo('metadata.google.internal', 80)
         except socket.gaierror:
@@ -193,7 +194,6 @@ class gcp_instance:
 
     def __get_nvme_disks_from_metadata(self):
         """get list of nvme disks from metadata server"""
-        import json
         try:
             disksREST=self.__instance_metadata("disks", True)
             disksobj=json.loads(disksREST)
@@ -351,6 +351,172 @@ class gcp_instance:
         except urllib.error.HTTPError:  # empty user-data
             return ""
 
+
+class azure_instance:
+    """Describe several aspects of the current Azure instance"""
+
+    EPHEMERAL = "ephemeral"
+    ROOT = "root"
+    GETTING_STARTED_URL = "http://www.scylladb.com/doc/getting-started-azure/"
+    META_DATA_BASE_URL = "http://169.254.169.254/1.0/meta-data/instance/"
+    ENDPOINT_SNITCH = "GossipingPropertyFileSnitch"
+
+    def __init__(self):
+        self.__type = None
+        self.__cpu = None
+        self.__location = None
+        self.__zone = None
+        self.__memoryGB = None
+        self.__nvmeDiskCount = None
+        self.__firstNvmeSize = None
+        self.__osDisks = None
+
+    @staticmethod
+    def is_azure_instance():
+        """Check if it's Azure instance via DNS lookup to metadata server."""
+        try:
+            addrlist = socket.getaddrinfo('metadata.azure.internal', 80)
+        except socket.gaierror:
+            return False
+        return True
+
+# as per https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service?tabs=windows#supported-api-versions
+    API_VERSION = "?api-version=2021-01-01"
+
+    def __instance_metadata(self, path):
+        """query Azure metadata server"""
+        return curl(self.META_DATA_BASE_URL + path + self.API_VERSION + "&format=text")
+
+    def is_in_root_devs(self, x, root_devs):
+        for root_dev in root_devs:
+            if root_dev.startswith(os.path.join("/dev/", x)):
+                return True
+        return False
+
+    def _non_root_nvmes(self):
+        """get list of nvme disks from os, filter away if one of them is root"""
+        nvme_re = re.compile(r"nvme\d+n\d+$")
+
+        root_dev_candidates = [x for x in psutil.disk_partitions() if x.mountpoint == "/"]
+        if len(root_dev_candidates) != 1:
+            raise Exception("found more than one disk mounted at root ".format(root_dev_candidates))
+
+        root_devs = [x.device for x in root_dev_candidates]
+
+        nvmes_present = list(filter(nvme_re.match, os.listdir("/dev")))
+        return {self.ROOT: root_devs, self.EPHEMERAL: [x for x in nvmes_present if not self.is_in_root_devs(x, root_devs)]}
+
+    @property
+    def os_disks(self):
+        """populate disks from /dev/ and root mountpoint"""
+        if self.__osDisks is None:
+            __osDisks = {}
+            nvmes_present = self._non_root_nvmes()
+            for k, v in nvmes_present.items():
+                __osDisks[k] = v
+            self.__osDisks = __osDisks
+        return self.__osDisks
+
+    def getEphemeralOsDisks(self):
+        """return just transient disks"""
+        return self.os_disks[self.EPHEMERAL]
+
+    @property
+    def nvmeDiskCount(self):
+        """get # of nvme disks available for scylla raid"""
+        if self.__nvmeDiskCount is None:
+            try:
+                ephemeral_disks = self.getEphemeralOsDisks()
+                count_os_disks = len(ephemeral_disks)
+            except Exception as e:
+                print("Problem when parsing disks from OS:")
+                print(e)
+                count_os_disks = 0
+            count_metadata_nvme_disks = self.__get_nvme_disks_count_from_metadata()
+            self.__nvmeDiskCount = count_os_disks if count_os_disks < count_metadata_nvme_disks else count_metadata_nvme_disks
+        return self.__nvmeDiskCount
+
+    instanceToDiskCount = {
+        "L8s": 1,
+        "L16s": 2,
+        "L32s": 4,
+        "L48s": 6,
+        "L64s": 8,
+        "L80s": 10
+    }
+
+    def __get_nvme_disks_count_from_metadata(self):
+        #storageProfile in VM metadata lacks the number of NVMEs, it's hardcoded based on VM type
+        return self.instanceToDiskCount.get(self.instance_class(), 0)
+
+    @property
+    def instancelocation(self):
+        """return the location of this instance, e.g. eastus"""
+        if self.__location is None:
+            self.__location = self.__instance_metadata("location")
+        return self.__location
+
+    @property
+    def instancezone(self):
+        """return the zone of this instance, e.g. 1"""
+        if self.__zone is None:
+            self.__zone = self.__instance_metadata("zone")
+        return self.__zone
+
+    @property
+    def instancetype(self):
+        """return the type of this instance, e.g. Standard_L8s_v2"""
+        if self.__type is None:
+            self.__type = self.__instance_metadata("vmSize")
+        return self.__type
+
+    @property
+    def cpu(self):
+        """return the # of cpus of this instance"""
+        if self.__cpu is None:
+            self.__cpu = psutil.cpu_count()
+        return self.__cpu
+
+    @property
+    def memoryGB(self):
+        """return the size of memory in GB of this instance"""
+        if self.__memoryGB is None:
+            self.__memoryGB = psutil.virtual_memory().total/1024/1024/1024
+        return self.__memoryGB
+
+    def instance_version(self):
+        """Returns the size of the instance we are running in. i.e.: v2"""
+        instancetypesplit = self.instancetype.split("_")
+        return instancetypesplit[2] if len(instancetypesplit) > 2 else ""
+
+    def instance_purpose(self):
+        """Returns the class of the instance we are running in. i.e.: Standard"""
+        return self.instancetype.split("_")[0]
+
+    def instance_class(self):
+        """Returns the purpose of the instance we are running in. i.e.: L8s"""
+        return self.instancetype.split("_")[1]
+
+    def is_unsupported_instance(self):
+        """Returns if this instance type belongs to unsupported ones for nvmes"""
+        return False
+
+    def is_supported_instance(self):
+        """Returns if this instance type belongs to supported ones for nvmes"""
+        if self.instance_class() in list(self.instanceToDiskCount.keys()):
+            return True
+        return False
+
+    def is_recommended_instance_size(self):
+        """if this instance has at least 2 cpus, it has a recommended size"""
+        if int(self.instance_size()) > 1:
+            return True
+        return False
+
+    def is_recommended_instance(self):
+        if self.is_recommended_instance_size() and not self.is_unsupported_instance() and self.is_supported_instance():
+            return True
+        return False
 
 class aws_instance:
     """Describe several aspects of the current AWS instance"""


dist/common/scripts/scylla_util.py
```diff
@@ -343,6 +343,183 @@ def user_data(self):
...
+        """Returns the purpose of the instance we are running in. i.e.: L8s"""
+        return self.instancetype.split("_")[1]
+
+    def is_unsupported_instance(self):
+        """Returns if this instance type belongs to unsupported ones for nvmes"""
+        return False
+
+    def is_supported_instance(self):
+        """Returns if this instance type belongs to supported ones for nvmes"""
+        if self.instance_class() in ['L8s', 'L16s', 'L32s' ,'L48s','L64s','L80s']:
```
tarzanek, line 483: this could optionally use the keyset of instanceToDiskCount

tarzanek:
@bentsi could you have a look please?

slivne:
I don't understand how the numbers align with the official specs MSFT Azure https://docs.microsoft.com/en-us/azure/virtual-machines/lsv2-series posted - but reviewers will ask the questions I am sure

tarzanek:
@slivne bursting applies to all other disks than nvme, nvme has a column there in above link, I don't think bursting is used for it
Azure VMs unlike other VMs always get a data folder attached with persistent storage and I'd expect bursting is used there and for OS disk.
see https://docs.microsoft.com/en-us/azure/virtual-machines/disk-bursting
and
https://azure.microsoft.com/en-us/blog/better-performance-with-bursting-enhancement-on-azure-disks/
none of them talks about nvmes

tarzanek:
also if you check the google doc, our tests for reads were always a bit higher, so I've set them to what is the read SLA in MS specs
for write I guess we will need to go with smallest of my few attempts
I tried diskxplorer one one instance too and it matched the expected number of queues (defaults that scylla assigns based on cpus), so I didn't run it and I rely just on 4.3 iotune(with Pavels fixes) for write iops and bandwidth

dist/common/scripts/scylla_util.py
```diff
@@ -343,6 +343,183 @@ def user_data(self):
...
+        self.__cpu = None
+        self.__memoryGB = None
+        self.__nvmeDiskCount = None
+        self.__firstNvmeSize = None
+        self.__osDisks = None
+
+    @staticmethod
+    def is_azure_instance():
+        """Check if it's Azure instance via DNS lookup to metadata server."""
+        import socket
```
bentsi, line 366: imports should be on top of the file

dist/common/scripts/scylla_util.py
```diff
@@ -343,6 +343,183 @@ def user_data(self):
...
+    def __init__(self):
+        self.__type = None
+        self.__cpu = None
+        self.__memoryGB = None
+        self.__nvmeDiskCount = None
+        self.__firstNvmeSize = None
+        self.__osDisks = None
+
+    @staticmethod
+    def is_azure_instance():
```
bentsi, line 364: looks like a copy-paste from `is_gce_instance` with small changes, generalize by moving the code into separate function or create a base class that all `*_instance` classes will inherit from.

avikivity:
I'm not reviewing a commit message without an explanation. Phrase the subject as an action done to the code, not a random noun.

tarzanek:
rebased, split commits, tried to write the message in form what was done to code
@avikivity can you review if you will have a spare minute?

@bentsi I'd like to refactor all 3 classes we have (aws, gcp, azure) in a separate commit, I don't want the refactoring to be part of this commit if possible

dist/common/scripts/scylla_io_setup
```diff
@@ -277,7 +277,54 @@ if __name__ == "__main__":
                 ioconf = open(etcdir() + "/scylla.d/io.conf", "w")
                 ioconf.write("SEASTAR_IO=\"--io-properties-file={}\"\n".format(properties_file.name))
             else:
-                logging.error('This is not a recommended Google Cloud instance setup for auto tuning, running manual iotune.')
+                logging.error('This is not a recommended Google Cloud instance setup for auto local disk tuning, running manual iotune.')
+                run_iotune()
+        elif azure_instance().is_azure_instance():
+            idata = gcp_instance()
```
yaronkaikov, line 283: This should be change to `azure_instance()` i guess.

dist/common/scripts/scylla_util.py
```diff
@@ -351,6 +351,183 @@ def user_data(self):
...
             return ""
 
+class azure_instance:
+    """Describe several aspects of the current Azure instance"""
+
+    EPHEMERAL = "ephemeral"
+    ROOT = "root"
+    GETTING_STARTED_URL = "http://www.scylladb.com/doc/getting-started-azure/"
+    META_DATA_BASE_URL = "http://169.254.169.254/1.0/meta-data/instance/"
+    ENDPOINT_SNITCH = "GossipingPropertyFileSnitch"
```
yaronkaikov, line 361: Do we have the property file? in https://github.com/scylladb/scylla/pull/8361/files#diff-9fe4dbe99d9ee03a478c9b25175e06bbd4cef9303c0ab1b9bdeea6e52d284cd4R141 we use simpleSnitch which doesn't require one

dist/common/scripts/scylla_util.py
```diff
@@ -351,6 +351,183 @@ def user_data(self):
...
+            return False
+        for res in addrlist:
+            af, socktype, proto, canonname, sa = res
+            if af == socket.AF_INET:
+                addr, port = sa
+                if addr == "169.254.169.254":
+                    return True
+        return False
+
+    API_VERSION="?api-version=2020-06-01"
```
yaronkaikov, line 386: I think API latest version is 2021-01-01, maybe we should use it?

yaronkaikov (changes requested):

dist/common/scripts/scylla_io_setup
tarzanek: fixed

dist/common/scripts/scylla_util.py
tarzanek: yes, we can definitely upgrade, since we use the basic common functionality, updating

dist/common/scripts/scylla_util.py
tarzanek: this is a very good comment
so we will need to add Azure Snitch to Scylla too!

tarzanek: I don't think we shall use SimpleSnitch, it's not production ready snitch
so until Azure Snitch is in Scylla whoever uses this snitch later in image, will need to create the property file

tarzanek:
thank you for review @yaronkaikov ! :-)
So with the Snitch it's a prob, I wouldn't use SimpleSnitch
is it possible image will generate the property file?
What GCP+EC2 are doing is just getting the region + zone and then using it as dc and rack , there is eventually a postifx/suffix added ( https://docs.scylladb.com/operating-scylla/system-configuration/snitch/#googlecloudsnitch )
so I guess we can add methods to get this info out of metadata for Azure and you will use it later on, how does that sound until we have an AzureSnitch in scylla cpp code base?

tarzanek:
e.g. sample GCP cluster with GCP snitch, suffix is _eu:
region is europe_west1
zones are b,c,d

$ nodetool status
Datacenter: europe-west1_eu
===========================
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address       Load       Tokens       Owns    Host ID                               Rack
UN  192.168.1.208  5.63 TB    256          ?       a787247a-69c3-42f3-c55f-da53135cd789  b
UN  192.168.1.209  5.54 TB    256          ?       cbe1fbb1-c309-4375-de0d-dfc7ae03d665  c
UN  192.168.1.210  5 TB       256          ?       4bed803b-ac5b-4722-d86c-d1f9ddb6fbef  d

tarzanek:
in my notes I have
```
# curl -H Metadata:true --noproxy "*" "http://169.254.169.254/metadata/instance/compute?api-version=2017-04-02"
# {"location":"eastus2","name":"mlxperfTest","offer":"CentOS","osType":"Linux","platformFaultDomain":"0","platformUpdateDomain":"0","publisher":"OpenLogic","sku":"8_1-gen2","version":"8.1.2020062401","vmId":"ebacd2f3-58ec-41ea-952e-8c7d51f84abf","vmSize":"Standard_L8s_v2"}[
```

so location can be used, however I am not sure how Azure is picking zones in the region now

tarzanek:
once the zone logic is known, it would be easy to add it to cpp similar to https://github.com/scylladb/scylla/blob/master/locator/gce_snitch.hh

(looking at https://github.com/apache/cassandra/tree/trunk/src/java/org/apache/cassandra/locator they don't have an Azure snitch)

tarzanek:
but then I found https://issues.apache.org/jira/browse/CASSANDRA-9387
and he does use the zone:

https://github.com/yoshw/cassandra/commit/7101f3ad8f622612248a866fca37a8864f51943c

dist/common/scripts/scylla_util.py
yaronkaikov: @tarzanek There is a template file already in Scylla repo - `conf/cassandra-rackdc.properties` - so we can actually edit this file with the required information

yaronkaikov (approved):

yaronkaikov:
> once the zone logic is known, it would be easy to add it to cpp similar to https://github.com/scylladb/scylla/blob/master/locator/gce_snitch.hh
> 
> (looking at https://github.com/apache/cassandra/tree/trunk/src/java/org/apache/cassandra/locator they don't have an Azure snitch)

@tarzanek when running the command `curl -H Metadata:true --noproxy "*" "http://169.254.169.254/metadata/instance?api-version=2020-06-01"`
I get the output 
```
{"compute":{"azEnvironment":"AzurePublicCloud","customData":"","evictionPolicy":"","isHostCompatibilityLayerVm":"false","licenseType":"","location":"eastus","name":"test","offer":"","osProfile":{"adminUsername":"scyllaadm","computerName":"test","disablePasswordAuthentication":"true"},"osType":"Linux","placementGroupId":"","plan":{"name":"","product":"","publisher":""},"platformFaultDomain":"0","platformUpdateDomain":"0","priority":"","provider":"Microsoft.Compute","publicKeys":[{"keyData":"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQCb/uCXTsL4iH/zOfcbOhEma+9d\r\nKRE+Q6fAHMfxhMG8hWuqVxMQMcGTyFgx/hXGBaB5kGIwScR4+Tyt0OuNLu9y2AbH\r\n5u3KXQVraED4UOQO8onAMk4ID2hkW1k46VDLLL/hJJC6ffdGYYx93nEny8mlgmFL\r\nVMnVc1tWThc2lWd68WKgTEDtkCcxSgSnEkMpFOos8/swcIMVdYNNpXlFZEm6i/ln\r\no1jRV5+yYydzK9nQsExFDaFNenHc3T86rAenTfCz+xn0f4VqBwd5sakDaGqJfAAQ\r\nxLmOqGQxyBOeoPSCUb+gDkYCE0Av1vLgGBSn16vcd+MNfetEw89x+aDg+NIV4ply\r\nvBXfXVzAI6jgOqsb0ZokN2AExhA/DOjQO2TXTCw0m+GjaLJj/pSshTl8eYt3Nyil\r\niKKnR7WKVH7i9W4SwtjiQp2JjQajc0mIeVSEAhpa3h9BFszzZjGfiAiJruCQTADl\r\nWNnbN0xq558HMgHa68Z6itlW+gRv1OK1mn25+XM= generated-by-azure\r\n","path":"/home/scyllaadm/.ssh/authorized_keys"}],"publisher":"","resourceGroupName":"scylla-images","resourceId":"/subscriptions/6c268694-47ab-43ab-b306-3c5514bc4112/resourceGroups/scylla-images/providers/Microsoft.Compute/virtualMachines/test","securityProfile":{"secureBootEnabled":"false","virtualTpmEnabled":"false"},"sku":"","storageProfile":{"dataDisks":[],"imageReference":{"id":"/subscriptions/6c268694-47ab-43ab-b306-3c5514bc4112/resourceGroups/scylla-images/providers/Microsoft.Compute/images/ScyllaDB-4.6.dev-0.20210420.dc7beec382-1","offer":"","publisher":"","sku":"","version":""},"osDisk":{"caching":"ReadWrite","createOption":"FromImage","diffDiskSettings":{"option":""},"diskSizeGB":"30","encryptionSettings":{"enabled":"false"},"image":{"uri":""},"managedDisk":{"id":"/subscriptions/6c268694-47ab-43ab-b306-3c5514bc4112/resourceGroups/scylla-images/providers/Microsoft.Compute/disks/test_disk1_300fb68b293643d6876dc8da1c1b11e6","storageAccountType":"Premium_LRS"},"name":"test_disk1_300fb68b293643d6876dc8da1c1b11e6","osType":"Linux","vhd":{"uri":""},"writeAcceleratorEnabled":"false"}},"subscriptionId":"6c268694-47ab-43ab-b306-3c5514bc4112","tags":"","tagsList":[],"userData":"","version":"","vmId":"65260cdc-7abd-4540-87c2-37972c3d2a62","vmScaleSetName":"","vmSize":"Standard_L8s_v2","zone":"1"},"network":{"interface":[{"ipv4":{"ipAddress":[{"privateIpAddress":"10.1.0.42","publicIpAddress":""}],"subnet":[{"address":"10.1.0.0","prefix":"24"}]},"ipv6":{"ipAddress":[]},"macAddress":"0022481E11CB"}]}}
```
We can use **"location":"eastus"** and **"zone":"1"**

yaronkaikov:
@tarzanek Please add a fixes tag to https://github.com/scylladb/scylla/issues/8325

tarzanek:
added both methods for location and zone and tag into the commit

dist/common/scripts/scylla_util.py
```diff
@@ -351,6 +351,199 @@ def user_data(self):
...
+    def get_file_size_by_seek(filename):
+        "Get the file size by seeking at end"
+        fd= os.open(filename, os.O_RDONLY)
+        try:
+            return os.lseek(fd, 0, os.SEEK_END)
+        finally:
+            os.close(fd)
+
+    # as per https://docs.microsoft.com/en-us/azure/virtual-machines/lsv2-series
+    AZURE_NVME_DISK_SIZE_2020=1966
```
avikivity, line 527: Can't we query it from the disk itself?

dist/common/scripts/scylla_util.py
```diff
@@ -351,6 +351,199 @@ def user_data(self):
...
+        try:
+            addrlist = socket.getaddrinfo('metadata.azure.internal', 80)
+        except socket.gaierror:
+            return False
+        for res in addrlist:
+            af, socktype, proto, canonname, sa = res
+            if af == socket.AF_INET:
+                addr, port = sa
+                if addr == "169.254.169.254":
+                    return True
```
avikivity, line 385: Why this check? It seems unnecessary, and potentially breaking for someone trying to emulate the API.

dist/common/scripts/scylla_util.py
```diff
@@ -351,6 +351,199 @@ def user_data(self):
...
+        """get list of nvme disks from os, filter away if one of them is root"""
+        nvme_re = re.compile(r"nvme\d+n\d+$")
+
+        root_dev_candidates = [x for x in psutil.disk_partitions() if x.mountpoint == "/"]
+        if len(root_dev_candidates) != 1:
+            raise Exception("found more than one disk mounted at root ".format(root_dev_candidates))
+
+        root_dev = root_dev_candidates[0].device
+        # if root_dev.startswith("/dev/mapper"):
+        #     raise Exception("mapper used for root, not checking if nvme is used ".format(root_dev))
```
avikivity, line 405: What's this comment?

dist/common/scripts/scylla_util.py
```diff
@@ -351,6 +351,199 @@ def user_data(self):
...
+    def is_recommended_instance_size(self):
+        """if this instance has at least 2 cpus, it has a recommended size"""
+        if int(self.instance_size()) > 1:
+            return True
+        return False
+
+    @staticmethod
+    def get_file_size_by_seek(filename):
+        "Get the file size by seeking at end"
+        fd= os.open(filename, os.O_RDONLY)
```
avikivity, line 520: coding style: space before "="

dist/common/scripts/scylla_util.py
```diff
@@ -351,6 +351,199 @@ def user_data(self):
...
+    @property
+    def firstNvmeSize(self):
+        """return the size of first non root NVME disk in GB"""
+        if self.__firstNvmeSize is None:
+            ephemeral_disks = self.getEphemeralOsDisks()
+            firstDisk = ephemeral_disks[0]
+            firstDiskSize = self.get_file_size_by_seek(os.path.join("/dev/", firstDisk))
+            firstDiskSizeGB = firstDiskSize/1024/1024/1024
+            if firstDiskSizeGB >= self.AZURE_NVME_DISK_SIZE_2020:
+                self.__firstNvmeSize = firstDiskSizeGB
```
avikivity, line 538: Why this check? Why not trust the disk size you measured?

dist/common/scripts/scylla_util.py
```diff
@@ -351,6 +351,199 @@ def user_data(self):
...
+            else:
+                raise Exception("First nvme is smaller than lowest expected size. ".format(firstDisk))
```
avikivity, line 540: And why this check? It seems unnecessarily restrictive.

dist/common/scripts/scylla_io_setup
```diff
@@ -277,7 +277,54 @@ if __name__ == "__main__":
...
+                elif nr_disks == 8:
+                    disk_properties["read_iops"] = 2900000
+                    disk_properties["read_bandwidth"] = 16000 * mbs
+                    disk_properties["write_iops"] = 2208081
+                    disk_properties["write_bandwidth"] = 9694 * mbs
+                elif nr_disks == 10:
+                    disk_properties["read_iops"] = 3800000
+                    disk_properties["read_bandwidth"] = 20000 * mbs
+                    disk_properties["write_iops"] = 2546511
+                    disk_properties["write_bandwidth"] = 11998 * mbs
```
avikivity, line 321: It would be nicer to make this table a dict:

{ 1: [400000, 2000, 271698, 1314],
  2: [...
}

If you want.

dist/common/scripts/scylla_io_setup
```diff
@@ -277,7 +277,54 @@ if __name__ == "__main__":
...
+                properties_file = open(etcdir() + "/scylla.d/io_properties.yaml", "w")
```
avikivity, line 322: Let's have an `else` branch so we don't crash if we don't match anything.

dist/common/scripts/scylla_util.py
tarzanek: fixed

dist/common/scripts/scylla_util.py
tarzanek: we do query it, this is a sanity check that we're not below this baseline with detected nvme disk

dist/common/scripts/scylla_util.py
tarzanek: removed, sorry,
I am actually tempted to replace all this method with https://github.com/scylladb/scylla/pull/8444
but it's not merged yet

dist/common/scripts/scylla_util.py
tarzanek: I can just drop it, since it is not needed for now

dist/common/scripts/scylla_util.py
tarzanek: I can just drop it, since it is not needed for now

tarzanek: and I fixed formatting for all code (auto-format)
yes ... again :-)

dist/common/scripts/scylla_util.py
tarzanek: dropping this, not needed

dist/common/scripts/scylla_util.py
tarzanek: replaced it with method from 8444

dist/common/scripts/scylla_util.py
tarzanek: this is a good question
the idea if I remember it right was to make sure it's an internet socket and the IP resolved is the one we expect (and that is set by MS Azure folks (and it basically the same for all clouds, but dns name is different))

how do you expect this to be @avikivity to fit your idea? (since I don't feel like relaxing it, resp. don't know how to relax this check in smarter way, sorry :-( )

dist/common/scripts/scylla_io_setup
tarzanek: added + added for GCP too

dist/common/scripts/scylla_io_setup
tarzanek: can I leave this for refactoring that I promised to Bentsi?

dist/common/scripts/scylla_io_setup
tarzanek: it needs to be added to AWS, but I'd like to defer that to a different PR

dist/common/scripts/scylla_util.py
tarzanek: note that this DNS is in local /etc/hosts , so the request shouldn't really involve network

avikivity: By relaxing the check I meant removing the whole thing. Why do you care if it's an inet address or what the address is? If you can connect to that dns name, that's enough.

dist/common/scripts/scylla_io_setup
avikivity: Yes ("If you want" above).

dist/common/scripts/scylla_util.py
tarzanek: okay, removing it then

dist/common/scripts/scylla_io_setup
avikivity: Sure (would be better to defer GCP too)

dist/common/scripts/scylla_util.py
avikivity: It doesn't look removed.

dist/common/scripts/scylla_util.py
```diff
@@ -378,13 +378,7 @@ def is_azure_instance():
...
         except socket.gaierror:
             return False
-        for res in addrlist:
-            af, socktype, proto, canonname, sa = res
-            if af == socket.AF_INET:
-                addr, port = sa
-                if addr == "169.254.169.254":
-                    return True
-        return False
+        return True
```
avikivity, line 381: Oh, it's removed here. Don't introduce something in patch 1 and remove it in patch 3, unless your goal is to annoy the reviewer.

dist/common/scripts/scylla_util.py
```diff
@@ -427,6 +421,80 @@ def getEphemeralOsDisks(self):
...
+#         "storageAccountType": "Standard_LRS"
+#       },
+#       "name": "exampleosdiskname",
+#       "osType": "Linux",
+#       "vhd": {
+#         "uri": ""
+#       },
+#       "writeAcceleratorEnabled": "false"
+#     }
+# }
```
avikivity, line 496: Please explain what this comment means so somebody looking at it in 4 months doesn't have to guess.

dist/common/scripts/scylla_util.py
```diff
@@ -518,6 +586,37 @@ def is_recommended_instance_size(self):
...
+
+    # @staticmethod
+    # def get_file_size_by_seek(filename):
+    #     "Get the file size by seeking at end"
+    #     fd= os.open(filename, os.O_RDONLY)
+    #     try:
+    #         return os.lseek(fd, 0, os.SEEK_END)
+    #     finally:
+    #         os.close(fd)
+
```
avikivity, line 598: What's this comment?

dist/common/scripts/scylla_util.py
tarzanek: will squash, sorry, was trying to make the change after review obvious

dist/common/scripts/scylla_util.py
tarzanek: removing, it was just a placeholder for the refactoring

dist/common/scripts/scylla_util.py
tarzanek: removing, it was just a placeholder for the refactoring

dist/common/scripts/scylla_util.py
tarzanek: will remove

tarzanek:
rewritten the commits, squashed the removal of check for dns reply
removed comments that didn't belong to the last commit
and removed the placeholder comments for refactoring
and since it's 4.5. I used the --force to push :-D

dist/common/scripts/scylla_util.py
tarzanek: refactoring will happen in next PR

PR #7809, opened 2020-12-18
installs.sh: Add seastar scripts to $PATH

### LINKED ISSUE #6731: seastar-cpu-map.sh missing from PATH
*Installation details*
Scylla version (or git commit hash): 4.0
Cluster size: 1

When trying to run seastar-cpu-map.sh on a Scylla docker instance, version 4.1, as per [this doc.](https://docs.scylladb.com/kb/map-cpu/) The script is not found on the $PATH.
It does exist on the server:
[root@89d8cbcdff24 /]# find . -name "seastar*.*"
./usr/lib/scylla/seastar-cpu-map.sh
./opt/scylladb/scripts/seastar-cpu-map.sh

penberg:
Add seastar script to the SBINFILES variable, which is used to create
symbolic links to scripts so that they appear in $PATH.

Fixes #6731

### DIFF
--- a/dist/debian/debian/scylla-server.install
+++ b/dist/debian/debian/scylla-server.install
@@ -12,6 +12,7 @@ usr/bin/scyllatop
 usr/sbin/scylla*
 usr/sbin/node_exporter_install
 usr/sbin/node_health_check
+usr/sbin/seastar-cpu-map.sh
 opt/scylladb/scripts/*
 opt/scylladb/swagger-ui/dist/*
 opt/scylladb/api/api-doc/*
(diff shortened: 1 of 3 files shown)


avikivity:
Alternative idea: add /etc/profile.d/scylla.sh that adjusts PATH to include /opt/scylladb/bin. This way we avoid polluting /usr/bin.

Downside is that it doesn't show up until the next login.

penberg:
@avikivity `seastar-cpu-map.sh` is unfortunately not in `/opt/scylladb/bin`, but in `/opt/scylladb/scripts`, which is why we have all these symlinks in the first place...

avikivity:
We could add symlinks in /opt/scylladb/bin too.

PR #7810, opened 2020-12-18
clustering_order_reader_merger: fix the 0 readers case

kbr-:
With 0 readers the merger would produce a `partition_end` fragment
when it should immediately return `end_of_stream` instead.


raphaelsc (approved):
LGTM

denesb (approved):

PR #7818, opened 2020-12-21
docs: add organization

dgarcia360:
@lauranovich @tzach This PR adds moves divides the dev docs into separate folders according to the organization plan.

## Before merging ...
The repo maintainer should activate GitHub pages:

1. Open the repository Settings, and scroll down to the "GitHub Pages" section.
2. Select `gh-pages` branch.
3. Wait a couple of minutes, it might take a while until GitHub applies the changes. If everything goes well, you will see the docs published under https://scylladb.github.io/scylla

## How to test changes

1. Run the command ``make preview`` from the ``docs`` directory.
3. Check the terminal where you have executed the previous command. It should not raise warnings.
3. Open in a new browser tab http://127.0.0.1:5500/ to see the new proposed organization.

## Next Steps

 In a subsequent iteration, @lauranovich and I will submit another PR moving all the [Wiki contents](https://github.com/scylladb/scylla/wiki) to the docs folder.


avikivity:
@dgarcia360 I applied the github configuration change, but the page does not render.

dgarcia360:
Thanks, @avikivity it could take a while. The page seems to be rendering now: http://scylladb.github.io/scylla/

image

nyh:
I would have liked (maybe I missed it??), before this patch moving everything around, some discussion on the motivation of these moves, and the subdivisions being proposed here.

I guess the name "design-notes" is the new name of we used to call *developer*-oriented documents, right? Can't we pick a shorter name? Maybe just "design" or "development" or "dev"?
What does the very general name "guides" supposed to mean?
What determined that logging.md, tracing.md and api_v2.md belongs in "guides" but metrics.md belongs in design-notes?
Why is cdc.md in design-notes while alternator is in a separate top-level directory?

What worries me most is that after you move things out of the new design-notes, we will start to remove stuff that isn't user-oriented enough. If we do this, I think it will be a big mistake.

nyh:
> What determined that logging.md, tracing.md and api_v2.md belongs in "guides"
>...
> What worries me most is that after you move things out of the new design-notes, we will start to remove stuff that isn't user-oriented enough. If we do this, I think it will be a big mistake.

Let me explain what I meant here in a bit more detail. If you look at logging.md, you'll see it doesn't describe just how a user enables or configures logging. It also mentions few (too few...) implementation details like mentioning Seastar, and it also mentions multiple ways of changing the logging level - configuration options, curl, and nodetool. I'm very worried that as soon as this document is declared "user facing", some of this information will promptly be removed. I think this would be a mistake. Developers need this information, even if we don't recommend that users use it.

**I ask anyone involved in this document reorganization to think ten times before removing any information from documents**. This is what - let's be honest - killed our wiki and caused developers to stop writing there. Let's try not to kill the docs/ directory.

dgarcia360:
Thanks @nyh for sharing your concerns.

> I would have liked (maybe I missed it??), before this patch moving everything around, some discussion on the motivation of these moves, and the subdivisions being proposed here.

I was asked to add Sphinx to this repository. There is currently no clear definition of what goes where, so we proposed the following organization:

* **Design notes** - These documents are informative, describing how new features work for other contributors and exploratory research.
* **Guides** - These documents are problem / solution-oriented. Right now they are mainly instructions on how to build and test the Scylla codebase. (Goal-oriented documents)
* **Contribute** - These documents are guidelines on how to contribute and maintain the project, but they don't talk about the codebase.

In a second iteration, we plan to migrate also all the contents from the Wiki to this documentation folder. It might be difficult to find information in 100s of documents placed in the same directory, so an initial division might be beneficial. I will appreciate any suggestions on how to organize the docs in a way that works better for all the devs/contributors.

> I guess the name "design-notes" is the new name of we used to call developer-oriented documents, right? Can't we pick a shorter name? Maybe just "design" or "development" or "dev"?

I agree with calling the folder just "design". I'd not use ``dev`` since there is the possibility ``scylla-docs`` ends merged in this repository. If this happens, these docs will be classified into ``docs/dev`` folder while keeping the subfolder structure proposed.
 
 > What determined that logging.md, tracing.md and api_v2.md belongs in "guides" but metrics.md belongs in design-notes?

After reading the doc between the lines, I feel it could fit there. Feel free to change them to another directory if fits better under guides.
 
 > Why is cdc.md in design-notes while alternator is in a separate top-level directory?

They were classified under ``design-notes`` originally but @lauranovich asked to not move the alternator docs during this iteration. We are evaluating if we keep them there, or integrate them into the user docs. In any case, these docs will not be removed from this repo, but they will add a redirection notice at the top of the file without removing its contents.

> What worries me most is that after you move things out of the new design-notes, we will start to remove stuff that isn't user-oriented enough

This PR is not removing anything. The intention is not to curate or remove any of the docs, but try to add some organization to these files.

> as soon as this document is declared "user facing"

The plan does not consider removing any doc.  If a document from this folder is eligible to be user-facing, it can be adapted copied, and adapted in a separate file under ``scylla-docs``. Nevertheless,  the original notes could be kept as they in this folder with a redirection notice on top.

nyh:
> In a second iteration, we plan to migrate also all the contents from the Wiki to this documentation folder. It might be difficult to find information in 100s of documents placed in the same directory, so an initial division might be beneficial. I will appreciate any suggestions on how to organize the docs in a way that works better for all the devs/contributors.

I think we should have held this discussion before the movement was committed. But we can always do further moves incrementally. That is, until we start to delete information (which was my main worry).

> After reading the doc between the lines, I feel it could fit there. Feel free to change them to another directory if fits better under guides.

I think the proper process would have been:
1. Discuss and review a new directory structure.
2. Start moving documents into the new structure. These movements also get reviewed, and there may be arguments about specific files (e.g., logging.md) and which directory they really belong to.

Somehow we seemed to have skipped this process - or I just missed it? (because @avikivity did commit it, so I guess he felt it went through the proper process).

> They were classified under `design-notes` originally but @lauranovich asked to not move the alternator docs during this iteration. We are evaluating if we keep them there, or integrate them into the user docs before moving them. In any case, these docs will not be removed from this repo, but they will add a redirection notice at the top of the file without removing its contents.

If you change any of the pre-existing files, please give their original authors ample time to review these changes, and don't bunch changes to many unrelated files in one patch.

> This PR is not removing anything. The intention is not to curate or remove any of the docs, but try to add some organization to these files.

I see. I thought that some of the new directories were already meant to be presented to users, as-is.  If it's **just** reorganization of the existing developers' documents, without any of these directories being declared more user-oriented than others, then I have less objections. But then I'm just not sure about the new directory structure. It looks almost random to me. From now on, when I'm looking for information about (say) logging, I will just need to search for each recursively because the new directory structure doesn't tell me where to look. But I guess we can improve this structure later, it's not a disaster (again, unless we start deleting material).

PR #7827, opened 2020-12-23
docs: how to run scylla with monitoring

sitano:
add a doc to Docker Hub description on how to run Scylla with monitoring.

Related PR in scylla-monitoring, but can't be reused https://github.com/scylladb/scylla-monitoring/pull/1181.

### DIFF
--- a/docs/guides/docker-hub.md
+++ b/docs/guides/docker-hub.md
@@ -429,6 +429,58 @@ Example value:
 
 * [Best practices for running Scylla on docker](http://docs.scylladb.com/procedures/best_practices_scylla_on_docker/)
 
+## Scylla Monitoring
+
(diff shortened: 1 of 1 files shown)


sitano:
We have discussed with Tzach that I need to rework it to use default scylla-monitoring start-all.sh script if it possible for this doc section.

tzach:
@amnonh please comment
I'm worried we are duplicating info from the monitoring stack into this project

sitano:
@tzach it's just a description of how to run monitoring manually with docker and prometheus, without implying scylla-monitoring. Nothing special. Mostly it's a squeeze of Scylla-monitoring scripts that demonstrate the simple approach.

tzach:
mentioning Scylla Monitoring in docker hub page make sense, but IMHO we should not copy content from the monitoring Stack, rather link to it.
@annastuchlik can you please send a new PR?

PR #7833, opened 2020-12-25
leveled_compaction_strategy: fix boundary of maximum sstable level

amoskong:
The MAX_LEVELS is the levels count, but sstable level (index) starts
from 0. So the maximum and valid level is MAX_LEVELS - 1.

Signed-off-by: Amos Kong <amos@scylladb.com>


raphaelsc (approved):
LGTM thanks for catching this off-by-one

PR #7846, opened 2020-12-29
Introduce load and stream feature

### LINKED ISSUE #7831: Load and stream arbitrary sstables
This feature extends the nodetool refresh to allow loading arbitrary sstables that do not belong to a node into the cluster.  It loads the sstables from disk and calculates the owning nodes of the data and streams to the owners automatically.

From example, say the old cluster has 6 nodes and the new cluster has 3 nodes. We can copy the sstables from the old cluster to any of the new nodes and trigger the load and stream process.  

This can make restores and migrations much easier.

asias:
storage_service: Introduce load_and_stream

=== Introduction ===

This feature extends the nodetool refresh to allow loading arbitrary sstables
that do not belong to a node into the cluster. It loads the sstables from disk
and calculates the owning nodes of the data and streams to the owners
automatically.

From example, say the old cluster has 6 nodes and the new cluster has 3 nodes.
We can copy the sstables from the old cluster to any of the new nodes and
trigger the load and stream process.

This can make restores and migrations much easier.

=== Performance ===

I managed to get 40MB/s per shard on my build machine.
CPU: AMD Ryzen 7 1800X Eight-Core Processor
DISK: Samsung SSD 970 PRO 512GB

Assume 1TB sstables per node, each shard can do 40MB/s, each node has 32
shards, we can finish the load and stream 1TB of data in 13 mins on each
node.

1TB / 40 MB per shard * 32 shard / 60 s = 13 mins

=== Tests ===

backup_restore_tests.py:TestBackupRestore.load_and_stream_to_new_cluster_test
which creates a cluster with 4 nodes and inserts data, then use
load_and_stream to restore to a 2 nodes cluster.

=== Usage ===

curl -X POST "http://{ip}:10000/storage_service/sstables/{keyspace}?cf={table}&load_and_stream=true

=== Notes ===

Btw, with the old nodetool refresh, the node will not pick up the data
that does not belong to this node but it will not delete it either. One
has to run nodetool cleanup to remove those data manually which is a
surprise to me and probably to users as well. With load and stream, the
process will delete the sstables once it finishes stream, so no nodetool
cleanup is needed.

The name of this feature load and stream follows load and store in CPU world.

Fixes #7831

### DIFF
--- a/api/api-doc/storage_service.json
+++ b/api/api-doc/storage_service.json
@@ -1764,6 +1764,22 @@
                      "allowMultiple":false,
                      "type":"string",
                      "paramType":"query"
+                  },
+                  {
+                     "name":"load_and_stream",
+                     "description":"Load the sstables and stream to all replica nodes that owns the data",
+                     "required":false,
+                     "allowMultiple":false,
+                     "type":"string",
+                     "paramType":"query"
+                  },
+                  {
+                     "name":"primary_replica_only",
+                     "description":"Load the sstables and stream to primary replica node that owns the data. Repair is needed after the load and stream process",
+                     "required":false,
+                     "allowMultiple":false,
+                     "type":"string",
+                     "paramType":"query"
                   }
                ]
             }
--- a/distributed_loader.cc
+++ b/distributed_loader.cc
@@ -152,14 +152,14 @@ future<> distributed_loader::verify_owner_and_mode(fs::path path) {
 };
 
 future<>
-distributed_loader::process_sstable_dir(sharded<sstables::sstable_directory>& dir) {
+distributed_loader::process_sstable_dir(sharded<sstables::sstable_directory>& dir, bool sort_sstables_according_to_owner) {
     return dir.invoke_on(0, [] (const sstables::sstable_directory& d) {
         return distributed_loader::verify_owner_and_mode(d.sstable_dir());
-    }).then([&dir] {
-      return dir.invoke_on_all([&dir] (sstables::sstable_directory& d) {
+    }).then([&dir, sort_sstables_according_to_owner] {
+      return dir.invoke_on_all([&dir, sort_sstables_according_to_owner] (sstables::sstable_directory& d) {
         // Supposed to be called with the node either down or on behalf of maintenance tasks
         // like nodetool refresh
-        return d.process_sstable_dir(service::get_local_streaming_priority()).then([&dir, &d] {
+        return d.process_sstable_dir(service::get_local_streaming_priority(), sort_sstables_according_to_owner).then([&dir, &d] {
             return d.move_foreign_sstables(dir);
         });
       });
@@ -447,6 +447,40 @@ distributed_loader::process_upload_dir(distributed<database>& db, distributed<db
         }, size_t(0), std::plus<size_t>()).get0();
 
         dblog.info("Loaded {} SSTables into {}", loaded, datadir.native());
+    });
+}
+
+future<std::tuple<utils::UUID, std::vector<std::vector<sstables::shared_sstable>>>>
+distributed_loader::get_sstables_from_upload_dir(distributed<database>& db, sstring ks, sstring cf) {
+    return seastar::async([&db, ks = std::move(ks), cf = std::move(cf)] {
+        global_column_family_ptr global_table(db, ks, cf);
+        sharded<sstables::sstable_directory> directory;
+        auto table_id = global_table->schema()->id();
+        auto upload = fs::path(global_table->dir()) / "upload";
+
+        directory.start(upload, db.local().get_config().initial_sstable_loading_concurrency(), std::ref(db.local().get_sharded_sst_dir_semaphore()),
+            sstables::sstable_directory::need_mutate_level::yes,
+            sstables::sstable_directory::lack_of_toc_fatal::no,
+            sstables::sstable_directory::enable_dangerous_direct_import_of_cassandra_counters(db.local().get_config().enable_dangerous_direct_import_of_cassandra_counters()),
+            sstables::sstable_directory::allow_loading_materialized_view::no,
+            [&global_table] (fs::path dir, int64_t gen, sstables::sstable_version_types v, sstables::sstable_format_types f) {
+                return global_table->make_sstable(dir.native(), gen, v, f, &error_handler_gen_for_upload_dir);
+
+        }).get();
+
+        auto stop = defer([&directory] {
+            directory.stop().get();
+        });
+
+        std::vector<std::vector<sstables::shared_sstable>> sstables_on_shards(smp::count);
+        lock_table(directory, db, ks, cf).get();
+        bool sort_sstables_according_to_owner = false;
+        process_sstable_dir(directory, sort_sstables_according_to_owner).get();
+        directory.invoke_on_all([&sstables_on_shards] (sstables::sstable_directory& d) mutable {
+            sstables_on_shards[this_shard_id()] = d.get_unsorted_sstables();
+        }).get();
+
+        return std::make_tuple(table_id, sstables_on_shards);
     });
 }
 
--- a/service/storage_service.cc
+++ b/service/storage_service.cc
@@ -82,6 +82,7 @@
 #include "repair/repair.hh"
 #include "service/priority_manager.hh"
 #include "utils/generation-number.hh"
+#include <seastar/core/coroutine.hh>
 
 using token = dht::token;
 using UUID = utils::UUID;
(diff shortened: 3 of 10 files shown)


dorlaor:
Really nice! How's your setup look like with regard to the other
receiving nodes? Is the bottleneck the 10Gbps network?

Also, will the node automatically delete the data that was loaded?
What happens in case there are errors in the middle?

asias:
> Really nice! How's your setup look like with regard to the other receiving nodes? Is the bottleneck the 10Gbps network? 

The test was running on a single machine. 

> Also, will the node automatically delete the data that was loaded? What happens in case there are errors in the middle?

Yes, it  will remove the sstables processed gradually. Currently, we process 16 sstables in batch. The 16 sstables will be removed once the load and stream is finished.  In case of error,  unprocessed sstables will be kept in the upload directory, so we can load and stream again later.

distributed_loader.cc
```diff
@@ -450,6 +451,42 @@ distributed_loader::process_upload_dir(distributed<database>& db, distributed<db
     });
 }
 
+future<std::tuple<utils::UUID, std::vector<std::vector<sstables::shared_sstable>>>>
+distributed_loader::get_sstables_from_upload_dir(distributed<database>& db, sstring ks, sstring cf) {
+    seastar::thread_attributes attr;
+    attr.sched_group = db.local().get_streaming_scheduling_group();
```
avikivity, line 457: IIRC, all REST API calls run under the streaming group anyway. So this isn't needed.

distributed_loader.cc
```diff
@@ -450,6 +451,42 @@ distributed_loader::process_upload_dir(distributed<database>& db, distributed<db
...
+        std::vector<std::vector<sstables::shared_sstable>> sstables_on_shards(smp::count);
+        lock_table(directory, db, ks, cf).get();
+        process_sstable_dir(directory).get();
+        directory.invoke_on_all([&sstables_on_shards] (sstables::sstable_directory& d) mutable {
+            sstables_on_shards[this_shard_id()] = d.get_sstables_load_and_stream();
+        }).get();
+
+        return std::make_tuple(table_id, sstables_on_shards);
+    });
+}
```
avikivity, line 488: Consider using a coroutine instead of a thread.

sstables/sstable_directory.cc
```diff
@@ -122,6 +124,11 @@ sstable_directory::process_descriptor(sstables::entry_descriptor desc, const ::i
             return make_ready_future<>();
         }
     }).then([sst, this] {
+        if (_load_and_stream) {
```
avikivity, line 127: This yet-another-boolean-flag is ugly. How about changing process_descriptor() to always collect sstables in a vector, and then adding an extra pass to do the per-shard sorting in a new function, which the existing users would call?

Then the shard-sorting is opt-in by whoever wants it.

api/api-doc/storage_service.json
```diff
@@ -1756,6 +1756,22 @@
                      "allowMultiple":false,
                      "type":"string",
                      "paramType":"query"
+                  },
+                  {
+                     "name":"load_and_stream",
```
avikivity, line 1761: Should we make it the default? Or even the only option? I don't see why the existing option is useful.

service/storage_service.cc
```diff
@@ -2658,20 +2658,207 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
+        for (;;) {
+            std::optional<std::tuple<int32_t>> status_opt = _source().get0();
+            if (!status_opt) {
+                slogger.debug("Finished reading source from node={}", _node);
+                break;
+            }
+            auto status = std::get<0>(*status_opt);
+            slogger.debug("Got error code from node={}, status={}", _node, status);
+            if (status == -1) {
+                _error_from_peer = true;
```
avikivity, line 2686: This bad error handling in rpc stream is really sad.

service/storage_service.cc
```diff
@@ -2658,20 +2658,207 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
+    }
+    size_t num_bytes_sent() {
+        return _num_bytes_sent;
+    }
+};
+
+future<> storage_service::load_and_stream(sstring ks_name, sstring cf_name,
+        utils::UUID table_id, std::vector<sstables::shared_sstable> sstables, bool primary_replica_only) {
+    seastar::thread_attributes attr;
+    attr.sched_group = _db.local().get_streaming_scheduling_group();
```
avikivity, line 2723: Again not needed. And you can use a coroutine.

service/storage_service.cc
```diff
@@ -2658,20 +2658,207 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
+    });
+}
+
+// For more details, see the commends on column_family::load_new_sstables
+// All the global operations are going to happen here, and just the reloading happens
+// in there.
+future<> storage_service::load_new_sstables(sstring ks_name, sstring cf_name,
+        bool load_and_stream, bool primary_replica_only) {
+    seastar::thread_attributes attr;
+    attr.sched_group = _db.local().get_streaming_scheduling_group();
```
avikivity, line 2834: again not needed.

service/storage_service.cc
```diff
@@ -2658,20 +2658,207 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
+            }
+        } catch (...) {
+            slogger.warn("Done loading new SSTables for keyspace={}, table={}, load_and_stream={}, primary_replica_only={}, status=failed: {}",
+                    ks_name, cf_name, load_and_stream, primary_replica_only, std::current_exception());
+            _loading_new_sstables = false;
+            throw;
+        }
+        slogger.warn("Done loading new SSTables for keyspace={}, table={}, load_and_stream={}, primary_replica_only={}, status=successed",
+                ks_name, cf_name, load_and_stream, primary_replica_only);
         _loading_new_sstables = false;
```
avikivity, line 2862: This will be leaked if we throw.

service/storage_service.cc
```diff
@@ -2658,20 +2658,207 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
+            nr_sst_current += sst_processed.size();
+            auto reader = table.make_streaming_reader(s, full_partition_range, sst_set);
 
-    return distributed_loader::process_upload_dir(_db, _sys_dist_ks, _view_update_generator, ks_name, cf_name).finally([this, ks_name, cf_name] {
-        slogger.info("Done loading new SSTables for {}.{}", ks_name, cf_name);
+            try {
+                netw::messaging_service& ms = _messaging.local();
+                for (;;) {
+                    auto mf = reader(db::no_timeout).get0();
+                    if (mf) {
```
avikivity, line 2768: Nicer: 

```c++
    while (auto mf = reader(db::no_timeout).get0()) {
```

or

```c++
    while (auto mf = co_await reader(db::no_timeout)) {
```

service/storage_service.cc
```diff
@@ -2658,20 +2658,207 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
+                        bool is_partition_start = mf->is_partition_start();
+                        if (is_partition_start) {
+                            ++num_partitions_processed;
+                            auto& start = mf->as_partition_start();
+                            current_dk = make_lw_shared<dht::decorated_key>(start.key());
```
avikivity, line 2773: Why make_lw_shared?

avikivity: Also, why copy it?

service/storage_service.cc
```diff
@@ -2658,20 +2658,207 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
+                        num_bytes_read += fmf.representation().size();
+                        parallel_for_each(current_targets, [&metas, &fmf, is_partition_start] (const gms::inet_address& node) {
+                            return metas.at(node).send(fmf, is_partition_start);
+                        }).get();
+                    } else {
+                        for (auto& [node, meta] : metas) {
+                            meta.finish();
+                        }
+                        break;
+                    }
```
avikivity, line 2800: This should be out of the loop.

service/storage_service.cc
```diff
@@ -2658,20 +2658,207 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
+                for (auto& [node, meta] : metas) {
+                    meta.error();
+                }
+                throw;
+            }
+
+            for (auto& sst : sst_processed) {
+                for (auto& f : sst->component_filenames()) {
+                    slogger.debug("load_and_stream: ops_uuid={}, ks={}, table={}, remove sst={}", ops_uuid, ks_name, cf_name, f);
+                    remove_file(f);
```
avikivity, line 2812: sst->unlink()

service/storage_service.cc
```diff
@@ -2658,20 +2658,207 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
+        } else {
+            _loading_new_sstables = true;
+        }
+        slogger.info("Loading new SSTables for keyspace={}, table={}, load_and_stream={}, primary_replica_only={}",
+                ks_name, cf_name, load_and_stream, primary_replica_only);
+        try {
+            if (load_and_stream) {
+                utils::UUID table_id;
+                std::vector<std::vector<sstables::shared_sstable>> sstables_on_shards;
+                std::tie(table_id, sstables_on_shards) = distributed_loader::get_sstables_from_upload_dir(_db, ks_name, cf_name).get0();
```
avikivity, line 2847: Why not structured bindings?

service/storage_service.cc
avikivity: Oh, there's a check above.

service/storage_service.cc
```diff
@@ -2658,20 +2658,207 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
+                            auto& start = mf->as_partition_start();
+                            current_dk = make_lw_shared<dht::decorated_key>(start.key());
+
+                            current_targets = rs.get_natural_endpoints(current_dk->token());
+                            if (primary_replica_only && current_targets.size() > 1) {
+                                current_targets.resize(1);
+                            }
+                            slogger.trace("load_and_stream: ops_uuid={}, current_dk={}, current_targets={}", ops_uuid,
+                                    current_dk->token(), current_targets);
+                            for (auto& node : current_targets) {
```
avikivity, line 2781: What happens if we stream to ourself?

service/storage_service.cc
```diff
@@ -2658,20 +2658,207 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
+        try {
+            if (load_and_stream) {
+                utils::UUID table_id;
+                std::vector<std::vector<sstables::shared_sstable>> sstables_on_shards;
+                std::tie(table_id, sstables_on_shards) = distributed_loader::get_sstables_from_upload_dir(_db, ks_name, cf_name).get0();
+                service::get_storage_service().invoke_on_all([&sstables_on_shards, ks_name, cf_name, table_id, primary_replica_only] (storage_service& ss) mutable {
+                    return ss.load_and_stream(ks_name, cf_name, table_id, std::move(sstables_on_shards[this_shard_id()]), primary_replica_only);
+                }).get();
+            } else {
+                distributed_loader::process_upload_dir(_db, _sys_dist_ks, _view_update_generator, ks_name, cf_name).get();
```
avikivity, line 2852: How is view update generation handled? By the current streaming server code?

avikivity:
Looks good. The streaming target will reshard if it gets partitions for the wrong shards?

Please test with an unsharded sstable.

service/storage_service.cc
asias: Like rebuild or replace,  the receiver does not generate view update, the view table should be loaded and streamed directly.

service/storage_service.cc
asias: Currently, it goes to the rpc like remote node. We can optimize the local write to avoid the rpc. I plan to do the optimization later.

service/storage_service.cc
asias: It does not compile on clang for some reason. It compiles on gcc.

service/storage_service.cc
asias: Cool.

api/api-doc/storage_service.json
asias: I think it is better to leave the default as it is now, not to surprise user. We do have user and procedures to use the old nodetool refresh.

distributed_loader.cc
asias: I will try to have some fun.

distributed_loader.cc
asias: It does not matter, it is called from storage_service::load_new_sstables which is already under streaming group. I will drop this. Thanks.

sstables/sstable_directory.cc
asias: I am not confident about the changes in sstables and distributed_loader.  I will think about it.

service/storage_service.cc
asias: Yes, I agree.

service/storage_service.cc
asias: Here, I actually want to make it explicit. Because we may want to run it under a different group other than streaming. For example, repair and load_and_stream run in parallel. We may want to make load_and_stream have even lower priority.

service/storage_service.cc
asias: I think I simplified the code. Now I can get rid of the shared ptr.

service/storage_service.cc
asias: It can be outside the loop. It is also natural to handle the !mf path directly. With while (auto mf = co_await reader(db::no_timeout)), it is better to do it outside the loop.

asias:
> Looks good. The streaming target will reshard if it gets partitions for the wrong shards?

We do not even reshard on the receiver. It uses the same code of streaming which uses the multi shard writer to send the parititon to correct shard. 

> 
> Please test with an unsharded sstable.

Can you explain more? What do you want to test?

api/api-doc/storage_service.json
avikivity: Hmm. We could have a new nodetool command, `nodetool import`.

Perhaps we can even have a REST API that accepts an sstable as input (and pushes it through HTTP POST). We can then issue `nodetool upload /path/to/sstable` and get the sstable streamed through http and then through this mechanism.

service/storage_service.cc
avikivity: storage_service doesn't deal with scheduling groups now. I'd like to keep it this way and not spread out the information about what runs where all over the source base. Right now it is inheritance based and main makes most of the decisions.

service/storage_service.cc
avikivity: Ah, you probably capture it in a lambda afterwards. http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1091r3.html

service/storage_service.cc
avikivity: ok.

avikivity: (it's not bad, just want to be sure it works)

service/storage_service.cc
avikivity: But this isn't how it works with regular nodetool refresh.

I don't know if it's good or bad, if users are expected to keep backups of view tables or not.

service/storage_service.cc
avikivity: > > Looks good. The streaming target will reshard if it gets partitions for the wrong shards?
> 
> We do not even reshard on the receiver. It uses the same code of streaming which uses the multi shard writer to send the parititon to correct shard.

That's what I meant.

> 
> > Please test with an unsharded sstable.
> 
> Can you explain more? What do you want to test?


Create an sstable with smp::count = 1, then load it into a cluster with smp::count > 1, verify it is split correctly.

avikivity: (I don't know why this reply got attached to this code chunk)

service/storage_service.cc
asias: Ok. removed all attr.

service/storage_service.cc
asias: done

service/storage_service.cc
asias: done.

service/storage_service.cc
asias: OK.

service/storage_service.cc
asias: Yes, capturing in a lambda is what I did here. And it was a surprise to me and wasted me some until I gave it up ...

api/api-doc/storage_service.json
asias: This sounds neat. However, to upload TB of sstables in production, I would rather use rsync instead of HTTP POST.

asias:
> Looks good. The streaming target will reshard if it gets partitions for the wrong shards?
> 
> Please test with an unsharded sstable.

I tested and worked in my new dtest for load and stream.

sstables/sstable_directory.cc
asias: I tried. sstable_directory::process_descriptor only processes one sstable. So, we need to pass a flag to distributed_loader::process_sstable_dir and sstable_directory::process_sstable_dir everywhere to instruct to call one of or both of sstable_directory::process_descriptor_collect_only() and sstable_directory::process_descriptor_sort_only().

I think it is more complicated and even more ugly than this version.

service/storage_service.cc
asias: It is bad if rpc compression is turned on. It is a waste to compress and uncompress locally.  I agree it is not urgent.

asias:
Changes in v2 (b34c5b89d9068e1b48158d85924813d1f3dd7506):
- Switch to coroutine
- Drop seastar::thread_attributes for streaming group
- Use while (auto mf = co_await reader(db::no_timeout)
- Move meta.finish outside the looop
- Use unlink to remove sstables files
- Test restore to a cluster with different shard count
- Improve error handling
- Rebase to latest master

api/api-doc/storage_service.json
avikivity: POST should be able to stream, and it has the advantage of using our I/O scheduler to write, so we don't overwhelm the disk.

sstables/sstable_directory.cc
avikivity: Think of it from the point of view of someone maintaining sstable_directory. What does `sstables::sstable_directory::load_and_stream::yes` mean? It doesn't mean that it will load and stream sstables. It can only be understood by understanding its caller.

It's bad design for a function to behave differently depending on the name of the caller (it's okay to have flags, but the flag can't mean "called from X".

My suggestion is to change process_descriptor to avoid the complicated sorting it does not:

```
future<>
sstable_directory::process_descriptor(sstables::entry_descriptor desc, const ::io_priority_class& iop) {
    if (desc.version > _max_version_seen) {
        _max_version_seen = desc.version;
    }

    auto sst = _sstable_object_from_existing_sstable(_sstable_dir, desc.generation, desc.version, desc.format);
    return sst->load(iop).then([this, sst] {
        validate(sst);
        if (_need_mutate_level) {
            dirlog.trace("Mutating {} to level 0\n", sst->get_filename());
            return sst->mutate_sstable_level(0);
        } else {
            return make_ready_future<>();
        }
    }).then([sst, this] {
        return sst->get_open_info().then([sst, this] (sstables::foreign_sstable_open_info info) {
            auto shards = sst->get_shards_for_this_sstable();
            if (shards.size() == 1) {
                if (shards[0] == this_shard_id()) {
                    dirlog.trace("{} identified as a local unshared SSTable", sst->get_filename());
                    _unshared_local_sstables.push_back(sst);
                } else {
                    dirlog.trace("{} identified as a remote unshared SSTable", sst->get_filename());
                    _unshared_remote_sstables[shards[0]].push_back(std::move(info));
                }
            } else {
                dirlog.trace("{} identified as a shared SSTable", sst->get_filename());
                _shared_sstable_info.push_back(std::move(info));
            }
            return make_ready_future<>();
        });
    });
}
```

We can drop the second continuation and replace it with something like:

```c++
    }).then([sst, this] {
        _unsorted_sstables.push_back(sst);
    });
```

And then call a new function (or add a new flag sort_sstables_according_to_owner) to pick up _unsorted_sstables and apply the shard sorting step.

api/api-doc/storage_service.json
asias: User can set a bandwidth limitation to rsync to limit the impact. What kind of tool the user use to POST + stream on the sender side?

sstables/sstable_directory.cc
asias: > Think of it from the point of view of someone maintaining sstable_directory. What does sstables::sstable_directory::load_and_stream::yes mean? It doesn't mean that it will load and stream sstables. It can only be understood by understanding its caller.

I agree it is hard to figure out what does this flag mean.  It is better to name the flag sstables::sstable_directory::collect_sstables_only::yes.

asias: > My suggestion is to change process_descriptor to avoid the complicated sorting it does not:

I think I got your idea to split the sort part out, but this means we need to carry a flag everywhere in other places.

asias:
Changes in v3 b139fa334e082f91e1dc4f8aea155d5a996df867:
 - Drop the load_and_stream flag passed to sstable_directory constructor
- Split the sort part out of sstable_directory::process_descriptor
- Rename get_sstables_load_and_stream to get_unsorted_sstables

sstables/sstable_directory.cc
asias: @avikivity I pushed v3 to split out the "sort sstables" as you suggested. Let's make some progress.

api/api-doc/storage_service.json
avikivity: curl, or an updated nodetool

service/storage_service.cc
```diff
@@ -2773,22 +2774,245 @@ void storage_service::add_expire_time_if_found(inet_address endpoint, int64_t ex
...
-    return distributed_loader::process_upload_dir(_db, _sys_dist_ks, _view_update_generator, ks_name, cf_name).finally([this, ks_name, cf_name] {
-        slogger.info("Done loading new SSTables for {}.{}", ks_name, cf_name);
+    slogger.info("Loading new SSTables for keyspace={}, table={}, load_and_stream={}, primary_replica_only={}",
+            ks_name, cf_name, load_and_stream, primary_replica_only);
+    try {
+        if (load_and_stream) {
+            utils::UUID table_id;
+            std::vector<std::vector<sstables::shared_sstable>> sstables_on_shards;
+            std::tie(table_id, sstables_on_shards) = co_await distributed_loader::get_sstables_from_upload_dir(_db, ks_name, cf_name);
+            co_await service::get_storage_service().invoke_on_all([&sstables_on_shards, ks_name, cf_name, table_id, primary_replica_only] (storage_service& ss) mutable -> future<> {
```
avikivity, line 3000: storage_service is a `peering_sharded_service`, so you can use container() instead of service::get_storage_service().

asias: OK

asias: Done. Fixed in v4.

asias:
Changes in v4 4d32d0317248d7c84ba91a16bc3252b2c8d98428:
- Use container instead of service::get_storage_service

avikivity:
I'll wait until #7820 gets promoted and queue.

PR #7858, opened 2021-01-04
Support unofficial distributions

syuu1228:
Since we introduced relocatable package and offline installer, scylla binary itself can run almost any distributions.
However, setup scripts are not designed to run in unsupported distributions, it causes error on such environment.
This PR adds minimal support to run offline installation on unsupported distributions, tested on SLES, Arch Linux and Gentoo.

### DIFF
--- a/dist/common/scripts/scylla_cpuscaling_setup
+++ b/dist/common/scripts/scylla_cpuscaling_setup
@@ -42,7 +42,7 @@ if __name__ == '__main__':
         cpufreq = systemd_unit('cpufrequtils.service')
         cpufreq.enable()
         cpufreq.restart()
-    if is_gentoo():
+    elif is_gentoo():
         if not shutil.which('cpupower'):
             emerge_install('sys-power/cpupower')
         with open('/etc/conf.d/cpupower') as f:
@@ -53,7 +53,7 @@ if __name__ == '__main__':
         cpufreq = systemd_unit('cpupower-frequency-set.service')
         cpufreq.enable()
         cpufreq.restart()
-    if is_arch():
+    elif is_arch():
         if not shutil.which('cpupower'):
             pkg_install('cpupower')
         cfg = sysconfig_parser('/etc/default/cpupower')
@@ -62,7 +62,7 @@ if __name__ == '__main__':
         cpupwr = systemd_unit('cpupower.service')
         cpupwr.enable()
         cpupwr.restart()
-    if is_amzn2():
+    elif is_amzn2():
         if not shutil.which('cpupower'):
             yum_install('cpupowerutils')
         cfg = sysconfig_parser('/etc/sysconfig/scylla-cpupower')
@@ -100,3 +100,6 @@ WantedBy=multi-user.target
         cpupwr = systemd_unit('cpupower.service')
         cpupwr.enable()
         cpupwr.restart()
+    else:
+        print("Unsupported distribution, skipping setup..")
+        sys.exit(0)
--- a/dist/common/scripts/scylla_cpuscaling_setup
+++ b/dist/common/scripts/scylla_cpuscaling_setup
@@ -56,6 +56,15 @@ if __name__ == '__main__':
         cpufreq = systemd_unit('cpupower-frequency-set.service')
         cpufreq.enable()
         cpufreq.restart()
+    if is_arch():
+        if not shutil.which('cpupower'):
+            pkg_install('cpupower')
+        cfg = sysconfig_parser('/etc/default/cpupower')
+        cfg.set('governor', 'performance')
+        cfg.commit()
+        cpupwr = systemd_unit('cpupower.service')
+        cpupwr.enable()
+        cpupwr.restart()
     if is_amzn2():
         if not shutil.which('cpupower'):
             yum_install('cpupowerutils')
--- a/dist/common/scripts/scylla_cpuscaling_setup
+++ b/dist/common/scripts/scylla_cpuscaling_setup
@@ -26,6 +26,22 @@ import shlex
 import distro
 from scylla_util import *
 
+UNIT_DATA= '''
+[Unit]
+Description=Scylla cpupower service
+After=syslog.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+EnvironmentFile=/etc/sysconfig/scylla-cpupower
+ExecStart=/usr/bin/cpupower $CPUPOWER_START_OPTS
+ExecStop=/usr/bin/cpupower $CPUPOWER_STOP_OPTS
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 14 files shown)


install.sh
```diff
@@ -174,6 +174,11 @@ check_usermode_support() {
     [ -n "$user" ]
 }
 
+if ! $packaging && [ ! -d /run/systemd/system/ ]; then
+    echo "systemd does not detected, unsupported distribution."
```
penberg, line 178: s/systemd does not/systemd is not/

dist/common/scripts/scylla_util.py
```diff
@@ -73,6 +73,11 @@ def datadir_p():
 def scyllabindir_p():
     return scylladir_p() / 'bin'
 
+def sysconfdir_p():
+    p = scylladir_p() / 'SCYLLA-SYSCONFDIR-FILE'
```
penberg, line 77: The `SCYLLA-SYSCONFDIR-FILE` is bit odd, I think. Would a symbolic link do the trick? @avikivity what do you think?

unified/install.sh
```diff
@@ -83,6 +83,11 @@ while [ $# -gt 0 ]; do
     esac
 done
 
+if [ ! -d /run/systemd/system/ ]; then
+    echo "systemd does not detected, unsupported distribution."
```
penberg, line 87: s/systemd does not/systemd is not/

penberg (commented):
Overall, looks good time, just few minor comments.

dist/common/scripts/scylla_util.py
avikivity: It breaks compatibility with existing installations. We must be able to load existing /etc/sysconfig/scylla-server files.

Let's move the default location to /etc/scylla.d. See http://0pointer.de/blog/projects/on-etc-sysinit.html.

We already have 

```
EnvironmentFile=/etc/sysconfig/scylla-server
EnvironmentFile=/etc/scylla.d/*.conf
```

Let's change that to 

```
EnvironmentFile=-/etc/sysconfig/scylla-server
EnvironmentFile=-/etc/default/scylla-server
EnvironmentFile=/etc/scylla.d/*.conf
```

This way /etc/sysconfig and /etc/default are still supported, but optional. scylla_setup can write to files in /etc/scylla.d (it's simpler since each component can write to a different file).

dist/common/scripts/scylla_cpuscaling_setup
```diff
@@ -56,6 +56,15 @@ if __name__ == '__main__':
         cpufreq = systemd_unit('cpupower-frequency-set.service')
         cpufreq.enable()
         cpufreq.restart()
+    if is_arch():
+        if not shutil.which('cpupower'):
+            pkg_error_exit('cpupower')
```
avikivity, line 74: Can't we use pkg_install here? and consolidate with existing code?

avikivity:
A suggestion (can come later): have package name translation pkg_xlat[pkg_name][os] so that package names that are slightly different between OSes can be referred to with a single name. For example

    pkg_xlat['cpupowerutils]['debian'] = 'cpufrequtils'

pkg_install can check for a name translation before installing.

dist/common/scripts/scylla_util.py
syuu1228: This does not trying to change sysconfig directory path, what the patch trying to do is allow python script to use --sysconfdir parameter passed to install.sh.

On our systemd unit file we generate sysconfdir.conf to customize EnvironmentFile directive as follows:
```
if [ "$sysconfdir" != "/etc/sysconfig" ]; then
    cat << EOS > "$retc"/systemd/system/scylla-server.service.d/sysconfdir.conf
[Service]
EnvironmentFile=
EnvironmentFile=$sysconfdir/scylla-server
EnvironmentFile=/etc/scylla.d/*.conf
EOS
fi
```

Unlike systemd unit file, current python script is hardcoded the path, so it's inconsistent:
```
if is_redhat_variant():
    cfg = sysconfig_parser('/etc/sysconfig/scylla-server')
else:
    cfg = sysconfig_parser('/etc/default/scylla-server')
```

If user specified non-standard path to install.sh, python script will fail (on Gentoo, sysconfig directory is /etc/conf.d)
The patch is adding a text file to store specified sysconfig directory path (SCYLLA-SYSCONFDIR-FILE), and read it from python script.

But I realized we don't really need the file, we can directly generate a simple python module file in install.sh instead, something like this:
```
 cat << EOS > "$rprefix"/scripts/scylla_sysconfdir.py
SYSCONFDIR="$sysconfdir"
EOS
```
Then use it from scylla_util.py:
```
--- a/dist/common/scripts/scylla_util.py
+++ b/dist/common/scripts/scylla_util.py
@@ -34,6 +34,7 @@ from pathlib import Path, PurePath
 from subprocess import run, DEVNULL
 
 import distro
+from scylla_sysconfdir import SYSCONFDIR
 
 
 def scriptsdir_p():
@@ -74,9 +75,7 @@ def scyllabindir_p():
     return scylladir_p() / 'bin'
 
 def sysconfdir_p():
-    p = scylladir_p() / 'SCYLLA-SYSCONFDIR-FILE'
-    with p.open('r', encoding='utf-8') as f:
-        return Path(f.read().strip())
+    return Path(SYSCONFDIR)
 
 def scriptsdir():
     return str(scriptsdir_p())
```


Moving sysconfig file to /etc/scylla.d sounds nice, easier to handle the file from script, but it may cause issue on upgrading from previous versions, I think it's safer to keep current path.

install.sh
syuu1228: Fixed the message.

dist/common/scripts/scylla_util.py
syuu1228: Pushed suggested code (generating python module in install.sh)

unified/install.sh
syuu1228: Fixed message

dist/common/scripts/scylla_cpuscaling_setup
syuu1228: Replaced with pkg_install(). It calls pkg_error_exit() on on unsupported distribution anyway, so result is same.

syuu1228:
> A suggestion (can come later): have package name translation pkg_xlat[pkg_name][os] so that package names that are slightly different between OSes can be referred to with a single name. For example
> 
> ```
> pkg_xlat['cpupowerutils]['debian'] = 'cpufrequtils'
> ```
> 
> pkg_install can check for a name translation before installing.

Added package name translation.

syuu1228:
@avikivity BTW, can we add /usr/bin/power as a relocatable command?
I think it's better to add for offline installation, and also make setup script simpler.

avikivity:
What is /usr/bin/power?

PR #7870, opened 2021-01-04
cql3: Fix unset values in collection operations

### LINKED ISSUE #7740: CQL prepared statements incomplete support for "unset" values
In Cassandra issue https://issues.apache.org/jira/browse/CASSANDRA-7304, an idea was raised to add to prepared statement bindings the notion of "unset" values, which are distinct from "null" values: Updating with a null value is tantamount to deleting the cell, while updating with an "unset" value is simply ignored. This "unset" feature makes it possible to use the same prepared statement in more scenarios instead of creating multiple prepared statement for different combinations of columns. 

This feature was added in Cassandra commit 48f644686b48357354f16c74b02b6d2c450a8c2d in April 2015, and was first released  in Cassandra 2.2.0 and the v4 binary protocol, with the following line in the release notes:
``` 
* Distinguish between null and unset in protocol v4 (CASSANDRA-7304)
```
The way to use it in the Python driver (for example) is through a special UNSET_VALUE value, as  documented in https://docs.datastax.com/en/developer/python-driver/3.24/api/cassandra/query/#module-cassandra.query-module-data

We did add support for unset values in Scylla back in January 2017 - see commit 9fb2f31 by @penberg. However, the Cassandra unit tests for this feature -  `cassandra_tests/validation/entities/collection_test.py`::`testMapWithUnsetValues`, and also `testListWithUnsetValues` and `testSetWithUnsetValues` (translated to cql-pytest as part of #7722) fail, and show that we got some of the details in the implementation wrong. At least when collections are involved.

In some of these tests, the UNSET_VALUE behaves like an ordinary null and causes the cell to be deleted, instead of just not doing anything to it. In the last test, `testSetWithUnsetValues`, the behavior is even worse: One of the operations in that test, `UPDATE %s SET s = s - ? WHERE k = 10"` bound with `UNSET_VALUE`, **crashes** Scylla, with the following assertion failure:
```
test_scylla: cql3/sets.cc:327: virtual void cql3::sets::discarder::execute(mutation &, const clustering_key_prefix
 &, const cql3::update_parameters &): Assertion `svalue' failed.
```
This is potentially dangerous - it can be used to DoS Scylla, so I'm upgrading this issue to "bug".

dekimir:
Fix handling of unset values so ported Cassandra tests can pass.

Fixes #7740.

Tests: unit (dev)

### DIFF
--- a/cql3/constants.hh
+++ b/cql3/constants.hh
@@ -192,8 +192,11 @@ public:
 
         virtual ::shared_ptr<terminal> bind(const query_options& options) override {
             auto bytes = bind_and_get(options);
-            if (!bytes) {
+            if (bytes.is_null()) {
                 return ::shared_ptr<terminal>{};
+            }
+            if (bytes.is_unset_value()) {
+                return UNSET_VALUE;
             }
             return ::make_shared<constants::value>(std::move(cql3::raw_value::make_value(to_bytes(*bytes))));
         }
--- a/cql3/expr/expression.cc
+++ b/cql3/expr/expression.cc
@@ -27,6 +27,7 @@
 #include <fmt/ostream.h>
 #include <unordered_map>
 
+#include "cql3/constants.hh"
 #include "cql3/lists.hh"
 #include "cql3/tuples.hh"
 #include "index/secondary_index_manager.hh"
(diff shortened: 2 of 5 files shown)


nyh (approved):
I'll commit, because it's a quick-and-dirty fix for a crash, but I'm fairly sure this is the right fix and will need to change again... UNSET_VALUE needs special treatment, not the same treatment as null, and this is why the tests continue to fail after this patch.

nyh:
I was about to commit this patch, but then noticed that after your patch, `testSetWithUnsetValues` no longer fails!
So it was wrong to replace the skip with a xfail - it passes! So please drop the "xfail".

But on second thought - the fact that this test passes now suggests that maybe (?) your trivial fix is the **correct** fix and fixes everything. If this the case, why do it only for sets? If you do the same trivial fix also for the maps and lists, you'll make all the tests in `test/cql-pytest/cassandra_tests/validation/entities/collections_test.py` with comments referring to #7740 pass, and you can also close #7740. You'll solve a whole issue instead of just one example case.

dekimir:
Just pushed v2, PTAL.
- fixed UNSET handling for all collections
- enabled all the tests
- marked #7740 as fixed
As usual, there are likely remnant bugs untriggered by tests.

nyh (approved):

bhalevy:
Is this PR all merged and can be closed?

dekimir:
Yes. It should have been closed automatically but didn't. 🙁

PR #7879, opened 2021-01-05
Make commitlog disk limit a hard limit.

elcallio:
Refs #6148

Commitlog disk limit was previously a "soft" limit, in that we allowed allocating new segments, even if we were over
disk usage max. This would also cause us sometimes to create new segments and delete old ones, if badly timed in 
needing and releasing segments, in turn causing useless disk IO for pre-allocation/zeroing. 

This patch set does:
* Make limit a hard limit. If we have disk usage > max, we wait for delete or recycle.
* Make flush threshold configurable. Default is ask for flush when over 50% usage. (We do not wait for results)
* Make flush "partial". We flush X% of the used space (used - thres/2), and make the rp limit accordingly. This means we will try to clear the N oldest segments, not all. I.e. "lighter" flush. Of course, if the CL is wholly dominated by a single CF, this will not really help much. But when > 1 cf is used, it means we can skip those not having unflushed data < req rp.
* Force more eager flush/recycle if we're out of segments

Note: flush threshold is not exposed in scylla config (yet). Because I am unsure of wording, and even if it should.
Note: testing is sparse, esp. in regard to latency/timeouts added in high usage scenarios. While I can fairly easily provoke "stalls" (i.e. forced waiting for segments to free up) with simple C-S, it is hard to say exactly where in a more sane config (I set my limits looow) latencies will start accumulating.

v2: 
* Changed update of flush pos to be max of done/prev

v3:
* Use shared promise for keeping track of waiting for disk deletions instead of semaphore. 
* Change flush size to be only data amount over usage threshold

### DIFF
--- a/db/commitlog/commitlog.cc
+++ b/db/commitlog/commitlog.cc
@@ -362,7 +362,7 @@ public:
         _flush_handlers.erase(id);
     }
 
-    void flush_segments(bool = false);
+    void flush_segments(uint64_t size_to_remove);
 
 private:
     future<> clear_reserve_segments();
@@ -1246,7 +1246,7 @@ void db::commitlog::segment_manager::create_counters(const sstring& metrics_cate
     });
 }
 
-void db::commitlog::segment_manager::flush_segments(bool force) {
+void db::commitlog::segment_manager::flush_segments(uint64_t size_to_remove) {
     if (_segments.empty()) {
         return;
     }
@@ -1259,8 +1259,20 @@ void db::commitlog::segment_manager::flush_segments(bool force) {
 
     // But if all segments are closed or we force-flush,
     // include all.
-    if (force || !active->is_still_allocating()) {
+    if (!active->is_still_allocating()) {
         high = replay_position(high.id + 1, 0);
+    }
+
+    auto n = size_to_remove;
+
+    if (size_to_remove != 0) {
+        for (auto& s : _segments) {
+            if (n <= s->_size_on_disk) {
+                high = replay_position(s->_desc.id, db::position_type(s->_size_on_disk));
+                break;
+            }
+            n -= s->_size_on_disk;
+        }
     }
 
     // Now get a set of used CF ids:
@@ -1271,7 +1283,7 @@ void db::commitlog::segment_manager::flush_segments(bool force) {
         }
     });
 
-    clogger.debug("Flushing ({}) to {}", force, high);
+    clogger.debug("Flushing ({} MB) to {}", size_to_remove/(1024*1024), high);
 
     // For each CF id: for each callback c: call c(id, high)
     for (auto& f : callbacks) {
@@ -1695,8 +1707,8 @@ void db::commitlog::segment_manager::on_timer() {
             auto cur = totals.active_size_on_disk;
             if (max != 0 && cur >= max) {
                 _new_counter = 0;
-                clogger.debug("Used size on disk {} MB exceeds local maximum {} MB", cur / (1024 * 1024), max / (1024 * 1024));
-                flush_segments();
+                clogger.debug("Used size on disk {} MB exceeds local threshold {} MB", cur / (1024 * 1024), max / (1024 * 1024));
+                flush_segments(cur - max);
             }
         }
         return do_pending_deletes();
--- a/db/commitlog/commitlog.cc
+++ b/db/commitlog/commitlog.cc
@@ -1427,11 +1427,19 @@ future<db::commitlog::segment_manager::sseg_ptr> db::commitlog::segment_manager:
 
     ++_new_counter;
 
-    // don't increase reserve count if we are at max, or we would go over disk limit. 
-    if (_reserve_segments.empty() && (_reserve_segments.max_size() < cfg.max_reserve_segments) && (totals.total_size_on_disk + max_size) <= max_disk_size) {
-        _reserve_segments.set_max_size(_reserve_segments.max_size() + 1);
-        clogger.debug("Increased segment reserve count to {}", _reserve_segments.max_size());
+    if (_reserve_segments.empty()) {        
+        // don't increase reserve count if we are at max, or we would go over disk limit. 
+        if (_reserve_segments.max_size() < cfg.max_reserve_segments && (totals.total_size_on_disk + max_size) <= max_disk_size) {
+            _reserve_segments.set_max_size(_reserve_segments.max_size() + 1);
+            clogger.debug("Increased segment reserve count to {}", _reserve_segments.max_size());
+        }
+        // if we have no reserve and we're above/at limits, make background task a little more eager.
+        if (!_shutdown && totals.total_size_on_disk >= disk_usage_threshold) {
+            _timer.cancel();
+            _timer.arm(std::chrono::milliseconds(0));
+        }
     }
+
     return _reserve_segments.pop_eventually().then([this] (auto s) {
         _segments.push_back(std::move(s));
         _segments.back()->reset_sync_time();
--- a/db/commitlog/commitlog.cc
+++ b/db/commitlog/commitlog.cc
@@ -233,6 +233,7 @@ public:
     using request_controller_type = basic_semaphore<timeout_exception_factory, db::timeout_clock>;
     using request_controller_units = semaphore_units<timeout_exception_factory, db::timeout_clock>;
     request_controller_type _request_controller;
+    shared_promise<> _disk_deletions;
 
     std::optional<shared_future<with_clock<db::timeout_clock>>> _segment_allocating;
     std::unordered_map<sstring, descriptor> _files_to_delete;
@@ -372,7 +373,7 @@ private:
     segment_id_type _ids = 0;
     std::vector<sseg_ptr> _segments;
     queue<sseg_ptr> _reserve_segments;
-    std::deque<sstring> _recycled_segments;
+    queue<sstring> _recycled_segments;
     std::unordered_map<flush_handler_id, flush_handler> _flush_handlers;
     flush_handler_id _flush_ids = 0;
     replay_position _flush_position;
@@ -1060,6 +1061,7 @@ db::commitlog::segment_manager::segment_manager(config c)
     // always be admitted for processing.
     , _request_controller(max_request_controller_units(), request_controller_timeout_exception_factory{})
     , _reserve_segments(1)
+    , _recycled_segments(std::numeric_limits<size_t>::max())
     , _reserve_replenisher(make_ready_future<>())
 {
     assert(max_size > 0);
@@ -1396,8 +1398,7 @@ future<db::commitlog::segment_manager::sseg_ptr> db::commitlog::segment_manager:
     }
 
     if (!_recycled_segments.empty()) {
-        auto src = std::move(_recycled_segments.front());
-        _recycled_segments.pop_front();
+        auto src = _recycled_segments.pop();
         // Note: we have to do the rename here to ensure
         // proper descriptor id order. If we renamed in the delete call
         // that recycled the file we could potentially have
@@ -1405,6 +1406,14 @@ future<db::commitlog::segment_manager::sseg_ptr> db::commitlog::segment_manager:
         clogger.debug("Using recycled segment file {} -> {}", src, dst);
         return rename_file(std::move(src), dst).then([this, d = std::move(d), dst = std::move(dst), flags] () mutable {
             return allocate_segment_ex(std::move(d), std::move(dst), flags);
+        });
+    }
+
+    if (max_disk_size != 0 && totals.total_size_on_disk >= max_disk_size) {
+        clogger.debug("Disk usage ({} MB) exceeds maximum ({} MB) - allocation will wait...", totals.total_size_on_disk/(1024*1024), max_disk_size/(1024*1024));
+        auto f = cfg.reuse_segments ? _recycled_segments.not_empty() :  _disk_deletions.get_shared_future();
+        return f.then([this] {
+            return allocate_segment();
         });
     }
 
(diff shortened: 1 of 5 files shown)


table.cc
```diff
@@ -1262,9 +1262,14 @@ future<std::unordered_map<sstring, table::snapshot_details>> table::get_snapshot
...
 
-future<> table::flush() {
+future<> table::flush(std::optional<db::replay_position> pos) {
+    if (pos && *pos < _flush_rp) {
+        return make_ready_future<>();
+    }
     auto op = _pending_flushes_phaser.start();
-    return _memtables->request_flush().then([op = std::move(op)] {});
+    return _memtables->request_flush().then([this, op = std::move(op), fp = _highest_rp] {
+        _flush_rp = fp;
```
avikivity, line 1271: What happens if another flush happens in parallel, and the completions are reordered? Should we merge fp into _flush_rp instead of assigning it?

avikivity: s/.then/co_await/?

db/commitlog/commitlog.cc
```diff
@@ -1259,10 +1259,22 @@ void db::commitlog::segment_manager::flush_segments(bool force) {
...
     // include all.
-    if (force || !active->is_still_allocating()) {
+    if (!active->is_still_allocating()) {
         high = replay_position(high.id + 1, 0);
     }
 
+    auto n = size_to_remove;
+
+    if (size_to_remove != 0) {
+        for (auto& s : _segments) {
```
avikivity, line 1272: Won't this get called again and again? Or is it serialized against the previous call of flush_segments?

avikivity: I'm thinking we should consider only the last segment where we left off, towards the segment that marks the boundary where we want to flush.

db/commitlog/commitlog.cc
```diff
@@ -1696,7 +1708,7 @@ void db::commitlog::segment_manager::on_timer() {
             if (max != 0 && cur >= max) {
                 _new_counter = 0;
                 clogger.debug("Used size on disk {} MB exceeds local maximum {} MB", cur / (1024 * 1024), max / (1024 * 1024));
-                flush_segments();
+                flush_segments(cur - max/2);
```
avikivity, line 1739: Why not cur - max? since max is the threshold.

btw, I don't see why it should be in the timer. As soon as a segment is completed, we can consider it against the threshold. The timer only makes it less predictable.

db/commitlog/commitlog.cc
```diff
@@ -1428,11 +1428,19 @@ future<db::commitlog::segment_manager::sseg_ptr> db::commitlog::segment_manager:
...
+        // don't increase reserve count if we are at max, or we would go over disk limit. 
+        if (_reserve_segments.max_size() < cfg.max_reserve_segments && (totals.total_size_on_disk + max_size) <= max_disk_size) {
+            _reserve_segments.set_max_size(_reserve_segments.max_size() + 1);
+            clogger.debug("Increased segment reserve count to {}", _reserve_segments.max_size());
+        }
+        // if we have no reserve and we're above/at limits, make background task a little more eager.
+        if (!_shutdown && totals.total_size_on_disk >= disk_usage_threshold) {
+            _timer.cancel();
+            _timer.arm(std::chrono::milliseconds(0));
+        }
```
avikivity, line 1441: I don't know if this is needed, especially if we move the logic to flush from a timer to segment close.

Are we preallocating far enough ahead? Perhaps we should make preallocation completely independent of the rest of commitlog; launch a fiber to preallocate until we reach our limit, then stop.

db/commitlog/commitlog.cc
elcallio: Once a segment is actually ready, it will be removed from this list. We will loop here every time until that time, but otoh we don't have to keep another state to keep track of relation "size we want to free" and "amount requested flushed". The list is usually ~10 or less, so should not be very expensive.

db/commitlog/commitlog.cc
elcallio: We can add another eager re-time of the background on segment finish, but I think it is better to keep the code paths separate. 
As for the size thingamajing, my reasoning is that requesting only cur-max would make us end up here more often - better to try to get a little larger portion, esp if we find we can tune up the threshold to something like 2/3 of actual max, which might be better, assuming flush can keep up.

db/commitlog/commitlog.cc
elcallio: Preallocation is done "on-demand" as soon as reserve is empty. This is to try to enforce flushing of memtables to allow this to finish.

table.cc
elcallio: But you are right. I sort of assumed the above serialized memtable flushes, but should not assume. Should be max(rp, curr)

db/commitlog/commitlog.cc
```diff
@@ -1609,7 +1626,9 @@ future<> db::commitlog::segment_manager::delete_file(const sstring& filename) {
     return seastar::file_size(filename).then([this, filename](uint64_t size) {
         clogger.debug("Deleting segment file {}", filename);
         return commit_io_check(&seastar::remove_file, filename).then([this, size] {
-            totals.total_size_on_disk -= size;  
+            clogger.trace("Reclaimed {} MB", size/(1024*1024));
+            totals.total_size_on_disk -= size;
+            _disk_deletions.signal(size);
```
gleb-cloudius, line 1639: The use of `_disk_deletions` look suspicious to me. We wait only when we exceed the limit, but always signal on delete. So if we delete while under the limit we may allocate when we over it.

db/commitlog/commitlog.cc
avikivity: I think we should launch preallocation during startup (unless, perhaps, --developer-mode). This is to avoid stalls when inserting the first terabyte of data.

What about dropping that timer? It might simplify things.

db/commitlog/commitlog.cc
avikivity: Ah, hysteresis. But doesn't that take us to 1/4 of the maximum? Since the threshold is around 1/2 of commitlog? I think that is too eager.

db/commitlog/commitlog.cc
avikivity: Why is it 10 or less?

db/commitlog/commitlog.cc
elcallio: Maybe. Worth tuning. But can change initial value.

db/commitlog/commitlog.cc
elcallio: Usually (in servers I run). It should be in the order of 10:s, since commitlog space is divided by shards, and we typically scale by sharding, not by throwing the same level of memory/disk to one shard. 
Point being, it is not a massive list. Esp not if we manage to get memtables to disk in an orderly fashion.

db/commitlog/commitlog.cc
elcallio: We do. It is part of startup. 
Timer just says we want to encourage memtable flushes now please.

db/commitlog/commitlog.cc
elcallio: Yeah, you are right. Should replace with a shard future (or try to keep the semaphore synced with alloc/delete files - but that would complicate exception handling something fierce...)

db/commitlog/commitlog.cc
avikivity: i3 family has 7GB/shard. That's 218 segments. We can envision servers with more memory/shard.

elcallio: Even so, assuming we flush cur-max, i.e. less than half the size, you will loop through ~100 elems and do a subtract+check. Not expensive. 
The alternative is trying to keep state here. Which is problematic, because there is no absolute correlation between disk space used, segments made free by flush requests and segment order here. In fact, the database could at any time by it self elect to release CF/rp:s so that a segment on the "other" side of your clock handle is dropped, and we move back to acceptable disk footprint. But that does not mean that when we next want to remove X bytes should start from a state marker set lower. 
I still think simply collecting the X lowest RP segments making up the N bytes we want to remove each time is the cleanest way to do it. Can maybe shortcut it by keeping track of how much we've currently asked to remove and including this in the trigger so we don't repeatedly request the same flushes here.

elcallio: But I don't even think that is really necessary, since we already restrict the flush requests to when we've created new segments since the last - i.e. we've actually increased footprint, not just still are over it.

avikivity: How often to we iterate over all of these segments?

I'm for keeping this state, but after a major refactoring (and coroutinization) of the code. It's quite gnarly.

elcallio: Every time we've created a new segment since last check, and are/remain over threshold. Which means at worst, assuming instigated memtable flush did not help (yet), and an incoming data rate at 32MB/shard would mean once every second. 
Less with lower data rate.

avikivity: Not thrilled with it. Iterating over hundreds of segments can cause a small stall (probably won't exceed our task quota). But we can improve it after some refactoring makes it easier.

gleb-cloudius:
I see that there were changes to the PR, but there is no changelog. May be changelog is the thing of the past with this new and enhanced github technology, but I cannot figure our how to see what changed since the previous review. What button should I press?

elcallio:
Mea culpa. Did not update the blurb. 
Since it is forced-pushed you can't really see the diff between iterations. 
Though technically, you would not see the difference really in a v2 patch mail either.

gleb-cloudius:
> Though technically, you would not see the difference really in a v2 patch mail either.

Sure, but in mail I complain about missing changelog right away :)

elcallio:
Don't be so hard on yourself. You complained pretty promptly here too! :-D

db/commitlog/commitlog.cc
```diff
@@ -1427,11 +1427,19 @@ future<db::commitlog::segment_manager::sseg_ptr> db::commitlog::segment_manager:
...
-        _reserve_segments.set_max_size(_reserve_segments.max_size() + 1);
-        clogger.debug("Increased segment reserve count to {}", _reserve_segments.max_size());
+    if (_reserve_segments.empty()) {        
+        // don't increase reserve count if we are at max, or we would go over disk limit. 
+        if (_reserve_segments.max_size() < cfg.max_reserve_segments && (totals.total_size_on_disk + max_size) <= max_disk_size) {
+            _reserve_segments.set_max_size(_reserve_segments.max_size() + 1);
+            clogger.debug("Increased segment reserve count to {}", _reserve_segments.max_size());
+        }
+        // if we have no reserve and we're above/at limits, make background task a little more eager.
+        if (!_shutdown && totals.total_size_on_disk >= disk_usage_threshold) {
```
kostja, line 1437: total_size_on_disk is inaccurate. It is incremented when we put a file to files_to_delete list, but at this point the file is not deleted yet and may be recycled. The following code explicitly re-arms the timer, but it's logic is inaccurate - we may re-arm the timer even though we will end up doing more work on disk, or may re-arm the timer too late, because total_size_on_disk doesn't include the size of reserved segments and the size of segments in the deletion queue (files_to_delete). 
This maybe kind-of works, but there is no guarantee we don't end up deleting the segments, or don't end up waiting for flushes for too long.

I'd start these series from building up accurate metrics for the total size on disk, which includes everything, and designing a mechanism which aligns the speed of flushes with the speed of use of segments. Simply kicking in the flushes when we're at 75% of segment size limit may be too much or too little, depending on the workload.

kostja (changes requested):

db/commitlog/commitlog.cc
elcallio: What are you talking about? total_size_on_disk is decremented in one place: (after actual file deletion), and incremented in two: when we allocate disk space (in allocate_segment_ex, we add space written by prealloc, and in segment::flush we add space _added_ (if any) by flush).

elcallio: And it very much includes the size of reserve. If they hold space.

kostja: total_size_on_disk is decremented in ~segment(), where the segment is added to files_to_delete in function add_file_to_delete(). At this point the file is still taking up space. 
Later on we're looking at the list of files in deciding to maybe not delete the file, and put it into reserved_segments. This happens in delete_segments. At this point we do not increment-back the total size on disk. So the total size on disk doesn't account for reserved segments and it is inaccurate, because it is decremented before a file is deleted *or* added to the list of reserved segments.

elcallio: No, it is not. I don't know which version of the code you are looking at, but it certainly is not this one.

kostja: I was looking at scylla-dev/master. Somehow it's kept up to date, but it doesn't contain some of the patches. When were your previous patches merged?

elcallio: I have no idea what state scylla-dev is in. These patches are based on scylla master. And are self-contained as such.

slivne:
@avikivity - what is left to get this code in - alejo and kostja tests showed this can improve performance ?

avikivity:
I guess I need to re-review it.

avikivity:
As does @gleb-cloudius

table.cc
```diff
@@ -1262,9 +1262,14 @@ future<std::unordered_map<sstring, table::snapshot_details>> table::get_snapshot
     });
 }
 
-future<> table::flush() {
+future<> table::flush(std::optional<db::replay_position> pos) {
+    if (pos && *pos < _flush_rp) {
```
avikivity, line 1266: Shouldn't this be `<=`? So that two flushes at the same position get coalesced?

Maybe I'm mistaken about replay_position numbering.

elcallio: Technically I guess it should. Though it should not happen. This is to deal with out-of-ordering, in which case things will be non-equal, but potentially lower than threshold.

db/commitlog/commitlog.cc
```diff
@@ -1533,14 +1542,21 @@ future<> db::commitlog::segment_manager::clear_reserve_segments() {
...
+    auto i = tmp.begin();
+    auto e = tmp.end();
 
     return parallel_for_each(i, e, [this](const sstring& filename) {
         clogger.debug("Deleting recycled segment file {}", filename);
         return delete_file(filename);
-    }).finally([this, re = std::move(re)] {
+    }).finally([this, tmp = std::move(tmp)] {
         return do_pending_deletes();
     });
```
avikivity, line 1569: Not sure why we are deleting those recycled segments. Won't we need to recreate them on the next startup?

Note, this isn't a comment on the patch but on the surrounding code.

avikivity:
Waiting for @gleb-cloudius review.

db/commitlog/commitlog.cc
elcallio: This is a sort of round-about mechanism, but the do_pending_deletes will go to delete_segments which will in fact probably just add most of these files to a new recycled list. The only point with this exercise is that we go through disk footprint checks one final time, only keeping the files that are under the limit. So in the new world where we uphold this limit, the above re-recycling should be pretty pointless.

gleb-cloudius (approved):

PR #7898, opened 2021-01-09
row_cache: linearize key in cache_entry::do_read()

### LINKED ISSUE #7897: crash in row cache if a partition key is larger than 13k
cache_entry::do_read() is missing a with_linearized_managed_bytes(). This can cause a crash if a partition key is larger than 13k.

avikivity:
do_read() does not linearize cache_entry::_key; this can cause a crash
with keys larger than 13k.

Fixes #7897.

### DIFF
--- a/row_cache.cc
+++ b/row_cache.cc
@@ -1263,7 +1263,9 @@ flat_mutation_reader cache_entry::read(row_cache& rc, read_context& reader, row_
 // Assumes reader is in the corresponding partition
 flat_mutation_reader cache_entry::do_read(row_cache& rc, read_context& reader) {
     auto snp = _pe.read(rc._tracker.region(), rc._tracker.cleaner(), _schema, &rc._tracker, reader.phase());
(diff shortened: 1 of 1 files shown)


row_cache.cc
```diff
@@ -1263,7 +1263,9 @@ flat_mutation_reader cache_entry::read(row_cache& rc, read_context& reader, row_
 // Assumes reader is in the corresponding partition
 flat_mutation_reader cache_entry::do_read(row_cache& rc, read_context& reader) {
     auto snp = _pe.read(rc._tracker.region(), rc._tracker.cleaner(), _schema, &rc._tracker, reader.phase());
-    auto ckr = query::clustering_key_filter_ranges::get_ranges(*_schema, reader.slice(), _key.key());
+    auto ckr = with_linearized_managed_bytes([&] {
+        return query::clustering_key_filter_ranges::get_ranges(*_schema, reader.slice(), _key.key());
```
tgrabiec, line 1267: Both slice and _key should be allocated in the standard allocator, how come the keys are fragmented?

tgrabiec (approved):

PR #7903, opened 2021-01-11
redis: support large redis message

### LINKED ISSUE #7273: redis 'exists' command fails with lots of keys

*Installation details*
Scylla version (or git commit hash): 6438c7cc2de49fe6c48bedd81a7b75e36ec00a7f
Cluster size: 1
OS (RHEL/CentOS/Ubuntu/AWS AMI): Ubuntu

When trying to add a test case for the 'exists' command using lots of keys, I found that it was passing with redis but failing with scylla, returning always the same (wrong) number of keys. This number varies depending on the size of the keys, so it seems linked with the size of the network packet, or something of fixed size.

The test:
```
def test_exists_lots_of_keys(redis_host, redis_port):
    r = connect(redis_host, redis_port)
    keys = []
    for i in range(0, 30):
        k = random_string(11)
        v = random_string(10)
        r.set(k, v)
        keys.append(k)
    assert r.exists(*keys) == len(keys)
```

Failing with:
```
>       assert r.exists(*keys) == len(keys)
E       AssertionError: assert 27 == 30
E        +  where 27 = <bound method Redis.exists of Redis<ConnectionPool<Connection<host=127.0.0.1,port=6973,db=0>>>>(*['zmmwvzovtad', 'ssyddizbwtd', 'wxuhojewfcj', 'tgayovfytgx', 'kotpccwkzls', 'ahbcaequbec', ...])
E        +    where <bound method Redis.exists of Redis<ConnectionPool<Connection<host=127.0.0.1,port=6973,db=0>>>> = Redis<ConnectionPool<Connection<host=127.0.0.1,port=6973,db=0>>>.exists
E        +  and   30 = len(['zmmwvzovtad', 'ssyddizbwtd', 'wxuhojewfcj', 'tgayovfytgx', 'kotpccwkzls', 'ahbcaequbec', ...])

```

If I increase the size of the keys, it fails with a smaller number of existing keys and vice versa.

syuu1228:
If the message is larger than current buffer size, we need to consume
more data until we reach to tail of the message.
To do so, we need to return nullptr when it's not on the tail.

Fixes #7273

### DIFF
--- a/redis/protocol_parser.rl
+++ b/redis/protocol_parser.rl
@@ -41,19 +41,19 @@ action mark {
 
 action start_blob {
     g.mark_start(p);
-    _size_left = _arg_size;
+    _bytes_left = _bytes_count;
 }
 
 action start_command {
     g.mark_start(p);
-    _size_left = _arg_size;
+    _bytes_left = _bytes_count;
 }
 
 action advance_blob {
-    auto len = std::min(static_cast<uint32_t>(pe - p), _size_left);
-    _size_left -= len;
+    auto len = std::min(static_cast<uint32_t>(pe - p), _bytes_left);
+    _bytes_left -= len;
     p += len;
-    if (_size_left == 0) {
+    if (_bytes_left == 0) {
       _req._args.push_back(str());
       p--;
       fret;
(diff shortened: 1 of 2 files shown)


nyh:
Unless I'm misunderstanding it, this patch seems to me the wrong solution: Whatever is the best buffer size (I don't know), isn't it important that we support correct parsing of the input even if it cross over a buffer boundary? Doesn't this patch just delay the inevitable, and it will still fail just with a larger message?

My guess is that `redis_protocol_parser` is buggy in its handling of protocol pieces which cross buffer boundaries, and should be fixed.

syuu1228:
@nyh Fix PR to continue consuming after retuning from parse() method, not extending buffer size.

Referenced seastar httpd parser impl: https://github.com/scylladb/seastar/blob/master/src/http/request_parser.rl

nyh (approved):
Excellent, that looks much better. Thanks.

PR #7910, opened 2021-01-12
cql3: Use correct comparator in timeuuid min/max

### LINKED ISSUE #7729: min/max aggregate functions are broken for timeuuid
CQL test case:

```
create table tuuidt (a int, b timeuuid, primary key (a,b));
insert into tuuidt (a, b) values (0, 13814000-1dd2-11ff-8080-808080808080);
insert into tuuidt (a, b) values (0, 6b1b3620-33fd-11eb-8080-808080808080);
select b, todate(b) from tuuidt where a = 0;
select todate(min(b)) from tuuidt where a = 0;
select todate(max(b)) from tuuidt where a = 0;
```

CQL output:

```
cqlsh:test1> create table tuuidt (a int, b timeuuid, primary key (a,b));
cqlsh:test1> insert into tuuidt (a, b) values (0, 13814000-1dd2-11ff-8080-808080808080);
cqlsh:test1> insert into tuuidt (a, b) values (0, 6b1b3620-33fd-11eb-8080-808080808080);
cqlsh:test1> select b, todate(b) from tuuidt where a = 0;

 b                                    | system.todate(b)
--------------------------------------+------------------
 6b1b3620-33fd-11eb-8080-808080808080 |       2020-12-01
 13814000-1dd2-11ff-8080-808080808080 |       2038-09-06

(2 rows)
cqlsh:test1> select todate(min(b)) from tuuidt where a = 0;

 system.todate(system.min(b))
------------------------------
                   2038-09-06

(1 rows)
cqlsh:test1> select todate(max(b)) from tuuidt where a = 0;

 system.todate(system.max(b))
------------------------------
                   2020-12-01
```

Please note that the clustering-key-order retrieval produces a correct order, since uses a different comparator, while min/max functions use UUID::operator<, which is binary compare, so produce an incorrect order.

Cassandra has no bug:
```
cassandra@cqlsh:test> create table tuuidt (a int, b timeuuid, primary key (a,b));
cassandra@cqlsh:test> insert into tuuidt (a, b) values (0, 13814000-1dd2-11ff-8080-808080808080);
cassandra@cqlsh:test> insert into tuuidt (a, b) values (0, 6b1b3620-33fd-11eb-8080-808080808080);
cassandra@cqlsh:test> select b, todate(b) from tuuidt where a = 0;

 b                                    | system.todate(b)
--------------------------------------+------------------
 6b1b3620-33fd-11eb-8080-808080808080 |       2020-12-01
 13814000-1dd2-11ff-8080-808080808080 |       2038-09-06

(2 rows)
cassandra@cqlsh:test> select todate(min(b)) from tuuidt where a = 0;

 system.todate(system.min(b))
------------------------------
                   2020-12-01

(1 rows)
cassandra@cqlsh:test> select todate(max(b)) from tuuidt where a = 0;

 system.todate(system.max(b))
------------------------------
                   2038-09-06

```

dekimir:
The min/max aggregators use aggregate_type_for comparators, and the
aggregate_type_for<timeuuid> is regular uuid.  But that yields wrong
results; timeuuids should be compared as timestamps.

Fix it by changing aggregate_type_for<timeuuid> from uuid to timeuuid,
so aggregators can distinguish betwen the two.  Then specialize the
aggregation utilities for timeuuid.

Add a cql-pytest and change some unit tests, which relied on naive
uuid comparators.

Fixes #7729.

Tests: unit (dev, debug)

### DIFF
--- a/cql3/functions/aggregate_fcts.cc
+++ b/cql3/functions/aggregate_fcts.cc
@@ -219,7 +219,7 @@ struct aggregate_type_for<simple_date_native_type> {
 
 template<>
 struct aggregate_type_for<timeuuid_native_type> {
-    using type = timeuuid_native_type::primary_type;
+    using type = timeuuid_native_type;
 };
 
 template<>
@@ -227,6 +227,7 @@ struct aggregate_type_for<time_native_type> {
     using type = time_native_type::primary_type;
 };
 
+// WARNING: never invoke this on temporary values; it will return a dangling reference.
 template <typename Type>
 const Type& max_wrapper(const Type& t1, const Type& t2) {
     using std::max;
@@ -239,6 +240,10 @@ inline const net::inet_address& max_wrapper(const net::inet_address& t1, const n
             (t1.in_family() == family::INET || t2.in_family() == family::INET)
             ? sizeof(::in_addr) : sizeof(::in6_addr);
     return std::memcmp(t1.data(), t2.data(), len) >= 0 ? t1 : t2;
+}
+
+inline const timeuuid_native_type& max_wrapper(const timeuuid_native_type& t1, const timeuuid_native_type& t2) {
+    return t1.uuid.timestamp() > t2.uuid.timestamp() ? t1 : t2;
 }
 
 template <typename Type>
@@ -323,6 +328,7 @@ make_max_function() {
     return make_shared<max_function_for<Type>>();
 }
 
+// WARNING: never invoke this on temporary values; it will return a dangling reference.
 template <typename Type>
 const Type& min_wrapper(const Type& t1, const Type& t2) {
     using std::min;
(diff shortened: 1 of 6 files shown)


cql3/functions/aggregate_fcts.cc
```diff
@@ -219,14 +219,15 @@ struct aggregate_type_for<simple_date_native_type> {
...
-    using type = timeuuid_native_type::primary_type;
+    using type = timeuuid_native_type;
 };
 
 template<>
 struct aggregate_type_for<time_native_type> {
     using type = time_native_type::primary_type;
 };
 
+// WARNING: never invoke this on temporary values; it will return a dangling reference.
```
avikivity, line 230: We can enforce it: add

```c++
template <typename Type1, typename Type2>
void max_wrapper(Type1&& t1, Type&& t2) = delete;
```

can be a nice follow-up.

dekimir: Won't that prevent invoking max_wrapper on lvalues?
https://godbolt.org/z/46E678

If it were easy to prevent exactly the bad scenario, wouldn't std::max have it already?

avikivity: Actually, it may be correct as long as you don't capture the result by reference. C++ extends the lifetime of temporaries until the end of the expression, so `auto m = std::max(temp1(), temp2());` should work (but `auto& m = std::max(temp1(), temp2());` would not).

dekimir: You're right.  I was aiming to replicate [this warning](https://en.cppreference.com/w/cpp/algorithm/max#Notes) (see "Notes" section), but I didn't capture it precisely.

PR #7917, opened 2021-01-13
api: remove potential large allocation in /column_family/ GET request…

### LINKED ISSUE #7916: api: a large allocation occurs in /column_family/ GET request handler
As detected in #6376, in the 5000 tables cluster test, a large (>1MB) allocation occurs when responding to a /column_family/ GET request, because we're creating a vector which contains names and keyspaces of all column families, and the response is generated by creating a string with all data from the vector. The function for creating this vector is here https://github.com/scylladb/scylla/blob/6cfc949e6217ac987dbed6b53c10e399cb5cf61e/api/column_family.cc#L313-L323

wmitros:
… handler

The reply to a /column_family/ GET request contains info about all
column families. Currently, all this info is stored in a single
string when replying, and this string may require a big allocation
when there are many column families.
To avoid that allocation, instead of a single string, use a
body_writer function, which writes chunks of the message content
to the output stream.

Fixes #7916

Signed-off-by: Wojciech Mitros <wojciech.mitros@scylladb.com>

### DIFF
--- a/api/column_family.cc
+++ b/api/column_family.cc
@@ -310,7 +310,7 @@ void set_column_family(http_context& ctx, routes& r) {
         return res;
     });
 
-    cf::get_column_family.set(r, [&ctx] (const_req req){
+    cf::get_column_family.set(r, [&ctx] (std::unique_ptr<request> req){
             vector<cf::column_family_info> res;
             for (auto i: ctx.db.local().get_column_families_mapping()) {
                 cf::column_family_info info;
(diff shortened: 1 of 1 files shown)


api/column_family.cc
```diff
@@ -310,7 +310,7 @@ void set_column_family(http_context& ctx, routes& r) {
         return res;
     });
 
-    cf::get_column_family.set(r, [&ctx] (const_req req){
+    cf::get_column_family.set(r, [&ctx] (std::unique_ptr<request> req){
             vector<cf::column_family_info> res;
```
avikivity, line 314: I think this can be large too. Maybe it should be a chunked_vector.

wmitros: `sizeof(cf::column_family_info)` returns 224, so with 5000 tables it is in fact over 1MB. It can't be a chunked_vector though because column_family_info's move constructor isn't noexcept, but I think we can just use an std::list instead. Also, we can't use the `stream_object` method because it uses the json formatter from seastar which only accepts std vectors and maps, but I see that we have `stream_range_as_array` for that case.
Should I create a new PR for this change or is it OK to add it here?

PR #7921, opened 2021-01-13
docker: remove sshd from the image

sitano:
implicit revert of 63222932630c67faafaf162e54c3c8b60859aee9

sshd previosly was used by the scylla manager 1.0.
new version does not need it. there is no point of
having it currently. it also confuses everyone.

Signed-off-by: Ivan Prisyazhnyy <ivan@scylladb.com>


fruch:
@penberg @sitano 

This change is breaking the tests for docker artifacts,  
https://jenkins.scylladb.com/view/nexts/job/scylla-master/job/artifacts/job/artifacts-docker-test/

they happen to be communicating via this sshd..

avikivity:
What are they communicating?

fruch:
> What are they communicating?

This is the way the those test are build in SCT, they how they start/stop scylla, and any other command it need to run locally on the node.

I guess it can be changed, or even the test can add the sshd back if needed, but it won't happen on it's own, we'll need time to adapt the code for one of the two options. (working with `docker exec` or installing ssh on our own)

avikivity:
I'll revert it for now.

fruch:
thanks @avikivity 

I've raise an issue in SCT, so we'll handle it, and then the can be reintroduced:
https://github.com/scylladb/scylla-cluster-tests/issues/3143

sitano:
this PR is waiting for https://github.com/scylladb/scylla-cluster-tests/issues/3143

fruch:
@sitano I hope we don't have other users counting on it in some other places, we aren't aware of.
It's a bummer when you get such surprises, and need to handle them unplanned.

we'll need to communicate it clearly on the release notes (once it would be out)

sitano:
@fruch yeah, sorry for that. I did not know there is a dependency in SCT. I don't know if there are users who use sshd in our docker image. if our users still use an old manager then they will not be able to upgrade Scylla docker image. I suppose we can warn everyone that in this release there will be no sshd dependency anymore.

penberg:
`Dockerfile` no longer exists as of aa7c40ba50f0132c8a83a40936ec08917fb9c527 so closing this.

PR #7922, opened 2021-01-13
scylla_ntp_setup: use chrony on all distributions

syuu1228:
This is v2 of #7229, fixed an error on Ubuntu 16.04.

----

To simplify scylla_ntp_setup, use chrony on all distributions.


avikivity:
It has extra patches in.

syuu1228:
@avikivity rebased.

PR #7923, opened 2021-01-13
docker: optimize build process

sitano:
fixes minor build script issues:

- make `docker build -t scylla .` to work by default by allowing
correct default version

- args must be used later so that when you change them it does not wipe
out all previously built image layers

- by default install `scylla` package instead of scylla-$VERSION package
to allow installing the latest version out of the repo by default. So
both commands will work:

    $ docker build -t scylla .
    $ docker build --build-arg VERSION=4.4.dev -t scylla .

- yum update to be cached as another prepare layer before actually
installing Scylla. that allows reusing cached layers when you will be
willing to change package version. useful especially for local builds.

- disable yum fastestmirror detect plugin to minimize scylla package
install time after full clean up.

Note. VERSION flag is used by the CI to build custom images.

### DIFF
--- a/dist/docker/redhat/Dockerfile
+++ b/dist/docker/redhat/Dockerfile
@@ -4,10 +4,6 @@ MAINTAINER Avi Kivity <avi@cloudius-systems.com>
 
 ENV container docker
 
-# The SCYLLA_REPO_URL argument specifies the URL to the RPM repository this Docker image uses to install Scylla. The default value is the Scylla's unstable RPM repository, which contains the daily build.
-ARG SCYLLA_REPO_URL=http://downloads.scylladb.com/rpm/unstable/centos/master/latest/scylla.repo
-ARG VERSION=4.6.dev
-
 ADD scylla_bashrc /scylla_bashrc
 
 # Scylla configuration:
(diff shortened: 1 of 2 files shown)


dist/docker/redhat/Dockerfile
```diff
@@ -4,10 +4,6 @@ MAINTAINER Avi Kivity <avi@cloudius-systems.com>
 
 ENV container docker
 
-# The SCYLLA_REPO_URL argument specifies the URL to the RPM repository this Docker image uses to install Scylla. The default value is the Scylla's unstable RPM repository, which contains the daily build.
-ARG SCYLLA_REPO_URL=http://downloads.scylladb.com/rpm/unstable/centos/master/latest/scylla.repo
-ARG VERSION=666.development
```
penberg, line 9: You cannot just drop `VERSION`. We added it on purpose in commit de19496ff7a1aeeb22096006ef2309270e9255e2 to ensure that the Jenkins pipelines can make sure the Docker image has the exact same version we're promoting.

penberg (changes requested):

dist/docker/redhat/Dockerfile
sitano: thank you for highlighting this! I turned back VERSION parameter with default value = '' so that both default and specific version will work.

sitano:
Now it must work with both:

    $ docker build -t scylla .
    $ docker build --build-arg VERSION=4.4.dev -t scylla .

However, I don't know how to test it with our CI

sitano:
https://github.com/scylladb/scylla/pull/7921 was reverted for a while to await completion from SCT. This one also must wait for it then.

avikivity:
If you cache yum update, won't that bloat the image?

sitano:
@avikivity yes, `yum update` will bloat the image. thats why we are having there `yum clean all` right after.

penberg:
@sitano Can you please rebase this? Please drop the ssh-removal commit too.

sitano:
@penberg I rebased. But currently our `http://downloads.scylladb.com/rpm/unstable/centos/master/latest/scylla.repo` returns something invalid so I can't build it:

```
curl http://downloads.scylladb.com/rpm/unstable/centos/master/latest/scylla.repo
<html>
  <head>
    <title>ScyllaDB Downloads</title>
</head>
<body>
  <div id="listing"></div>

<script src="http://ajax.googleapis.com/ajax/libs/jquery/1.9.0/jquery.min.js"></script>
<script type="text/javascript">
  var BUCKET_URL = 'http://downloads.scylladb.com.s3.amazonaws.com';
</script>
<script src="/scripts/list.js"></script>
</body>
</html>
```

Shall we fix it anyhow in this patch?

dist/docker/redhat/Dockerfile
```diff
@@ -33,12 +29,22 @@ ADD scyllasetup.py /scyllasetup.py
...
-RUN curl $SCYLLA_REPO_URL -o /etc/yum.repos.d/scylla.repo && \
-    yum -y install epel-release && \
+
+# Prepare layer for actual software install
+RUN sed -i -e 's/enabled=1/enabled=0/' /etc/yum/pluginconf.d/fastestmirror.conf && \
+    yum -y install epel-release hostname supervisor openssh-server openssh-clients rsyslog && \
     yum -y clean expire-cache && \
     yum -y update && \
-    yum -y install scylla-$VERSION hostname supervisor openssh-server openssh-clients rsyslog && \
+    yum clean all
```
avikivity, line 38: I think this `yum clean all` here is correct, but it disagrees with what the changelog says.

I think it's the changelog that should be fixed, not the code.

avikivity:
We should consider using the unified tarball instead of packages. This would allow us to base the image on alpine linux or another small base, and to avoid using the repo for the build. We'd just take the already-built tarball and install it in the container.

penberg:
@sitano The `Dockerfile` no longer exists as of aa7c40ba50f0132c8a83a40936ec08917fb9c527 so I am closing this pull request.

PR #7939, opened 2021-01-20
dbuild: avoid --pids-limit with podman and cgroupsv1

### LINKED ISSUE #7938: dbuild fails with an error on older kernels (without cgroupsv2)
```console
$ ./tools/toolchain/dbuild 
Error: container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: process_linux.go:422: setting cgroup config for procHooks process caused: cannot set pids limit: container could not join or create cgroup: OCI runtime error
```

avikivity:
Podman doesn't correctly support --pids-limit with cgroupsv1. Some
versions ignore it, and some versions reject the option.

To avoid the error, don't supply --pids-limit if cgroupsv2 is not
available (detected by its presence in /proc/filesystems). The user
is required to configure the pids limit in
/etc/containers/containers.conf.

Fixes #7938.

### DIFF
--- a/tools/toolchain/dbuild
+++ b/tools/toolchain/dbuild
@@ -146,12 +146,24 @@ if [ -z "$is_podman" ]; then
        "${group_args[@]}"
        -v /etc/passwd:/etc/passwd:ro
        -v /etc/group:/etc/group:ro
+       --pids-limit -1
        )
 else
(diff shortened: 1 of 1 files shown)


avikivity:
Note: this will make the arm machine build again.

nyh (approved):

PR #7941, opened 2021-01-20
dist: drop /etc/security/limits.d/scylla.conf

### LINKED ISSUE #7925: Scylla doesn't use /etc/security/limits.d/scylla.conf

Scylla version 4.2.1-0.20201108.4fb8ebccff with build-id d19a95fd85a8e7df2928ccd8c729941f263e8adc
Cluster size: 4 Datacenters (3 x 3 nodes 1 x 7 nodes)
OS (RHEL/CentOS/Ubuntu/AWS AMI): 18.04.5 LTS (GNU/Linux 4.15.0-128-generic x86_64)

Platform (physical/VM/cloud instance type/docker): Physical
Hardware: sockets= 2 cores=8 hyperthreading= 32 memory= 128 GB

So while dealing with a different issue with hitting a "too many open files" issue.  I ran into this.

the /etc/security/limits.d/scylla.conf
is configured like this on all of our servers.
`scylla  -  core     unlimited

scylla  -  memlock  unlimited

scylla  -  nofile   200000

scylla  -  as       unlimited

scylla  -  nproc    8096
`

But looking at /proc/PROCID/limits it was showing nofile at 800000   I finally dug into it more.  It appears that it is being statically set in the /etc/init/scylla-server.conf  config file 

`setuid scylla

setgid scylla

limit core unlimited unlimited

limit memlock unlimited unlimited

limit nofile 800000 800000

limit as unlimited unlimited

limit nproc 8096 8096

chdir /var/lib/scylla

env HOME=/var/lib/scylla
`

so more informative.  it would be good to correct the documentation to point to the correct place to configure this.  I imagine it was done on purpose to only give the DB process itself the different restrictions vs anything that is running as scylla.  the JMX process also does it's own thing.

syuu1228:
Drop limits.d conf file, since we don't use it.
We set these parameters via systemd unit file instead.

Fixes #7925

### DIFF
--- a/dist/common/limits.d/scylla.conf
+++ b/dist/common/limits.d/scylla.conf
@@ -1,5 +1,0 @@
-scylla  -  core     unlimited
-scylla  -  memlock  unlimited
-scylla  -  nofile   200000
-scylla  -  as       unlimited
-scylla  -  nproc    8096
--- a/dist/common/systemd/scylla-server.service
+++ b/dist/common/systemd/scylla-server.service
@@ -7,6 +7,7 @@ Wants=scylla-housekeeping-daily.timer
 [Service]
 PermissionsStartOnly=true
 Type=notify
+LimitCORE=infinity
 LimitMEMLOCK=infinity
 LimitNOFILE=800000
 LimitAS=infinity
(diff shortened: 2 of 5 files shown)


dist/common/limits.d/scylla.conf
```diff
@@ -1,5 +0,0 @@
-scylla  -  core     unlimited
```
avikivity, line 1: I don't see an equivalent to this setting.

syuu1228: @avikivity Added LimitCORE=infinity to scylla-server.service.

PR #7943, opened 2021-01-21
allocation_strategy: set preferred max contiguous allocation to 128k …

michoecho:
…for standard allocations

Now that managed_bytes and its users do not assume that a managed_bytes
instance allocated using standard_allocation_strategy is non-fragmented,
we can set the preferred max contiguous allocation to 128k. This causes
managed_bytes to fragment instances that are larger than this size.

Note that managed_bytes is the only user.

### DIFF
--- a/utils/allocation_strategy.hh
+++ b/utils/allocation_strategy.hh
@@ -185,6 +185,9 @@ public:
 
 class standard_allocation_strategy : public allocation_strategy {
 public:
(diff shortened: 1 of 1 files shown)


bhalevy:
@avikivity should this be backported to 4.4 following 60f5ec3644f8961105b3b0db219c1c230f416e1a?

avikivity:
Let's see if there is any fallout.

PR #7946, opened 2021-01-21
multishard_combining_reader: do not use `smp::count`

### LINKED ISSUE #7945: `multishard_combining_reader` uses `smp::count` in one place instead of `_sharder.shard_count()`
`multishard_combining_reader` currently only works under the assumption that every table uses the same sharder configured using the node's number of shards. But we could potentially specify a different sharder for a chosen table, e.g. one that puts everything on shard 0. Then this assumption will be broken and the reader causes a segfault. The fix should be easy - simply replace that one use of `smp::count` with `_sharder.shard_count()`.

kbr-:
`multishard_combining_reader` currently only works under the assumption
that every table uses the same sharder configured using the node's number
of shards. But we could potentially specify a different sharder for a chosen table,
e.g. one that puts everything on shard 0.
Then this assumption will be broken and the reader causes a segfault.

Fixes #7945.

### DIFF
--- a/mutation_reader.cc
+++ b/mutation_reader.cc
@@ -1828,7 +1828,7 @@ void multishard_combining_reader::on_partition_range_change(const dht::partition
         boost::push_heap(_shard_selection_min_heap);
     };
 
-    for (auto shard = _current_shard + 1; shard < smp::count; ++shard) {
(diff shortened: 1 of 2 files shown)


denesb (approved):
LGTM, thanks for the fix @kbr-!

PR #7947, opened 2021-01-21
schema_tables: put schema tables on shard 0

kbr-:
We use a custom sharder for all schema tables: every table under
the `system_schema` keyspace, plus `system.scylla_table_schema_history`.
This sharder puts all data on shard 0.

To achieve this, we hardcode the sharder in initial schema object
definitions. Furthermore - since the sharder is not stored inside schema
mutations yet - whenever we deserialize schema objects from mutations,
we modify the sharder based on the schema's keyspace and table names.

A regression test is added to ensure no one forgets to set the special
sharder for newly added schema tables. This test assumes that all newly
added schema tables will end up in the `system_schema` keyspace (other
tables may go unnoticed, unfortunately).

---

Everything seems to magically work:
1. tracing confirms that each desired table (everything under `system_schema` + `system.scylla_table_schema_history`) has data only under shard 0 (performing `select * from`). I opened a bunch of `cqlsh` sessions which were connected to different shards - each routed the query to shard 0, as desired.
```
                                                       Execute CQL3 query | 2021-01-21 16:43:13.878000 | 127.0.0.1 |              0 | 127.0.0.1
                                            Parsing a statement [shard 1] | 2021-01-21 16:43:13.878038 | 127.0.0.1 |              3 | 127.0.0.1
                                         Processing a statement [shard 1] | 2021-01-21 16:43:13.878386 | 127.0.0.1 |            351 | 127.0.0.1
                                    read_data: querying locally [shard 1] | 2021-01-21 16:43:13.878685 | 127.0.0.1 |            650 | 127.0.0.1
                        Start querying token range (-inf, +inf) [shard 1] | 2021-01-21 16:43:13.878704 | 127.0.0.1 |            670 | 127.0.0.1
                              Creating shard reader on shard: 0 [shard 0] | 2021-01-21 16:43:13.879179 | 127.0.0.1 |              8 | 127.0.0.1
 Scanning cache for range (-inf, +inf) and slice {(-inf, +inf)} [shard 0] | 2021-01-21 16:43:13.879589 | 127.0.0.1 |            430 | 127.0.0.1
                                               Querying is done [shard 1] | 2021-01-21 16:43:13.902964 | 127.0.0.1 |          24929 | 127.0.0.1
                           Done processing - preparing a result [shard 1] | 2021-01-21 16:43:13.916814 | 127.0.0.1 |          38780 | 127.0.0.1
                                                         Request complete | 2021-01-21 16:43:13.918379 | 127.0.0.1 |          40379 | 127.0.0.1
```
2. rolling upgrade works, even downgrade
3. data in these tables is not lost between upgrades/downgrades (created bunch of keyspaces and tables - they still show up in the schema tables after such upgrade/downgrade and I can normally use them)

Does not require fixing https://github.com/scylladb/scylla/issues/7642 (mutations don't have info about the sharder - we simply "fix" the schema after deserializing it, same trick as with "extra durability")

### DIFF
--- a/schema.cc
+++ b/schema.cc
@@ -934,6 +934,11 @@ schema_builder& schema_builder::with_sharder(unsigned shard_count, unsigned shar
     return *this;
 }
 
+schema_builder& schema_builder::with_null_sharder() {
+    _raw._sharder = get_sharder(1, 0);
+    return *this;
+}
+
 schema_builder::schema_builder(std::string_view ks_name, std::string_view cf_name,
         std::optional<utils::UUID> id, data_type rct)
         : _raw(id ? *id : utils::UUID_gen::get_time_UUID())
--- a/db/schema_tables.cc
+++ b/db/schema_tables.cc
@@ -271,6 +271,7 @@ schema_ptr keyspaces() {
         ));
         builder.set_gc_grace_seconds(schema_gc_grace);
         builder.with_version(generate_schema_version(builder.uuid()));
+        builder.with_null_sharder();
         return builder.build();
     }();
     return schema;
@@ -312,6 +313,7 @@ schema_ptr tables() {
         ));
         builder.set_gc_grace_seconds(schema_gc_grace);
         builder.with_version(generate_schema_version(builder.uuid()));
+        builder.with_null_sharder();
         return builder.build();
     }();
     return schema;
@@ -340,6 +342,7 @@ schema_ptr scylla_tables(schema_features features) {
             offset += 2;
         }
         sb.with_version(generate_schema_version(id, offset));
+        sb.with_null_sharder();
         return sb.build();
     };
     static thread_local schema_ptr schemas[2][2] = { {make(false, false), make(false, true)}, {make(true, false), make(true, true)} };
@@ -381,6 +384,7 @@ static schema_ptr columns_schema(const char* columns_table_name) {
         ));
     builder.set_gc_grace_seconds(schema_gc_grace);
     builder.with_version(generate_schema_version(builder.uuid()));
+    builder.with_null_sharder();
     return builder.build();
 }
 schema_ptr columns() {
@@ -416,6 +420,7 @@ static schema_ptr computed_columns_schema(const char* columns_table_name) {
         ));
     builder.set_gc_grace_seconds(schema_gc_grace);
     builder.with_version(generate_schema_version(builder.uuid()));
+    builder.with_null_sharder();
     return builder.build();
 }
 
@@ -445,6 +450,7 @@ schema_ptr dropped_columns() {
         ));
         builder.set_gc_grace_seconds(schema_gc_grace);
         builder.with_version(generate_schema_version(builder.uuid()));
+        builder.with_null_sharder();
         return builder.build();
     }();
     return schema;
@@ -470,6 +476,7 @@ schema_ptr triggers() {
         ));
         builder.set_gc_grace_seconds(schema_gc_grace);
         builder.with_version(generate_schema_version(builder.uuid()));
+        builder.with_null_sharder();
         return builder.build();
     }();
     return schema;
@@ -514,6 +521,7 @@ schema_ptr views() {
         ));
         builder.set_gc_grace_seconds(schema_gc_grace);
         builder.with_version(generate_schema_version(builder.uuid()));
+        builder.with_null_sharder();
         return builder.build();
     }();
     return schema;
@@ -540,6 +548,7 @@ schema_ptr indexes() {
         ));
         builder.set_gc_grace_seconds(schema_gc_grace);
         builder.with_version(generate_schema_version(builder.uuid()));
+        builder.with_null_sharder();
         return builder.build();
     }();
     return schema;
@@ -566,6 +575,7 @@ schema_ptr types() {
         ));
         builder.set_gc_grace_seconds(schema_gc_grace);
         builder.with_version(generate_schema_version(builder.uuid()));
+        builder.with_null_sharder();
         return builder.build();
     }();
     return schema;
@@ -595,6 +605,7 @@ schema_ptr functions() {
         ));
         builder.set_gc_grace_seconds(schema_gc_grace);
         builder.with_version(generate_schema_version(builder.uuid()));
+        builder.with_null_sharder();
         return builder.build();
     }();
     return schema;
@@ -624,6 +635,7 @@ schema_ptr aggregates() {
         ));
         builder.set_gc_grace_seconds(schema_gc_grace);
         builder.with_version(generate_schema_version(builder.uuid()));
+        builder.with_null_sharder();
         return builder.build();
     }();
     return schema;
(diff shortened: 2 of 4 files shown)


denesb:
Resharding is only needed if shard numbers change in such a way that any range R that used to belong to a single shard, now belongs to 2 or more shards. If you change shards in a compatible way, no resharding is needed, in other words, we don't need to do anything when ranges are merged. For example changing the shard count from any N->N/2 doesn't trigger resharding. Also, changing from any N->1 doesn't trigger resharding either.

denesb:
> Also how do you even reshard if for resharding you need to be able to interpret mutations - but to do that, you need the tables' schema - but to do that, you need to access the schema tables, which are not yet resharded?

You can read data that has to be resharded, you just have to filter out data you don't own. I think we still do this automatically for shared sstables (that contain data owned by more than 1 shard).

avikivity:
> Resharding is only needed if shard numbers change in such a way that any range R that used to belong to a single shard, now belongs to 2 or more shards. If you change shards in a compatible way, no resharding is needed, in other words, we don't need to do anything when ranges are merged. For example changing the shard count from any N->N/2 doesn't trigger resharding. Also, changing from any N->1 doesn't trigger resharding either.


Yes. It's still a good idea that the "which shards own this sstable" question is answered by the schema's sharder. Likely true, but worth verifying.

avikivity:
Impressively simple. Good job @haaawk for laying the groundwork.

avikivity:
I queued the first patch.

schema.cc
```diff
@@ -934,6 +934,11 @@ schema_builder& schema_builder::with_sharder(unsigned shard_count, unsigned shar
     return *this;
 }
 
+schema_builder& schema_builder::with_null_sharder() {
+    _raw._sharder = get_sharder(1, 0);
```
tgrabiec, line 938: We have to also fix https://github.com/scylladb/scylla/issues/7642, right?

tgrabiec: We don't, I see why (and you mention it too)

tgrabiec (approved):

kbr-:
> Resharding is only needed if shard numbers change in such a way that any range R that used to belong to a single shard, now belongs to 2 or more shards. If you change shards in a compatible way, no resharding is needed, in other words, we don't need to do anything when ranges are merged. For example changing the shard count from any N->N/2 doesn't trigger resharding. Also, changing from any N->1 doesn't trigger resharding either.

Thanks @denesb. Indeed, I looked more carefully now, and it does reshard all schema tables when I create a bunch of keyspaces/tables *with this patch applied* (including the case where I bootstrap a fresh Scylla with this patch) and then downgrade.
It doesn't reshard on upgrade, as expected according to your explanation.

---

@avikivity 

> It's still a good idea that the "which shards own this sstable" question is answered by the schema's sharder. Likely true, but worth verifying.

Trying to answer this question as best as I could, and to convince myself that there are no issues lurking in the shadows if we change the system schema tables to use the custom null sharder, I performed a bit of an investigation.

There are two pieces of information that each sstable has assigned that are related to sharding:
- a list of token ranges - on disk and in memory. Each partition in the sstable is supposed to be in one of those token ranges.
- a list of shard numbers - in memory. When we restart, we derive this from the list of token ranges above. However, we also assign this when we create the sstable in the first place. In that case, we derive the list of token ranges from the list of shard numbers (so it's the other way around).

I was reading the code in order to check that these two pieces are correctly calculated and the invariant associated with token ranges is maintained, even if we use a custom sharder.

There are two execution paths we need to consider:
1. when we create a new sstable,
2. when we restart and load sstables from disk.

Ad 1: When we create an sstable, we always assign *a single shard number* to the sstable (so the list of shard numbers is a singleton). Then from this single shard number we derive the list of token ranges, as explained further below.

There are two cases of creating new sstables:
1. When we flush a memtable, or when we perform any compaction that is not a resharding compaction, we assign *the current shard* (`this_shard_id()`). This is correct because all partitions that will be written to the newly created sstable belong to the current shard:
- for a memtable flush, a shard flushes only its own memtables. It assumes that all partitions in this memtable belong to it. That's true, since writes are always routed to the right shard using the `dht::shard_of` function, and (after @haaawk's changes) that function always uses the schema's sharder.
- for a compaction that is not a resharding compaction, a shard compacts *only its own unshared sstables*, i.e. sstables such that all partitions they contain belong to this shard; at this point we assume that existing sstables are assigned to shards correctly, which would be the inductive hypothesis if we turned all this discussion into an inductive proof.

2. When we perform a resharding compaction, we go through all partitions in an sstable and assign them shards using the `shard_of` function (using the schema's sharder). These partitions are bucketed according to shards, and from each bucket we create a new sstable, assigning it the shard associated with the bucket. Hence these new sstables are assigned correct shards as well. See `shard_based_splitting_writer` and `resharding_compaction::create_compaction_writer`.

There are actually scenarios which I didn't look at in detail which also create new sstables: streaming and repair. But fortunately for our case, these don't apply - we only change the sharder for local tables for now. Further investigation will need to be performed there when we decide to use distributed custom sharders.

After a shard is assigned to a new sstable - which we determined to be assigned correctly - we calculate the token ranges. This is done by taking the smallest and the largest token in the sstable, making a range out of it - [smallest, largest] - and then intersecting that range with the assigned shard. This truncates the large into a bunch of smaller subranges. Because we previosuly guaranteed that all partitions belong to the assigned shard, and also all partitions belong to the [smallest, largest] token range, and the intersecting code uses the schema's sharder as well, it follows that all partitions belong to these obtained smaller subranges. Therefore the token ranges are calculated correctly, at least according to the sharder currently being used by the schema.

Ad 2: When we load an sstable from disk, we look at the list of token ranges - we don't have a list of shards yet. We assume that each partition in the sstable belongs to one of these ranges (which, again, would be the inductive hypothesis if turned this into an inductive proof). However, even though each such range belonged to a single shard - and all of them to the same shard - when the node was previously shut down, it might not be true now. Thus we intersect each range with each shard and see if there are non-empty intersections with 2 shards or more. If so, we perform resharding, which creates new sstables with correctly assigned shards as explained above, and then removes this sstable, restoring our invariants. Otherwise, this sstable is unshared, so we assign it the one shard that owns this sstable. The intersecting code uses the schema's sharder, so the shard is correctly assigned.

QED.

kbr-:
I also ran a full dtest-release run on Jenkins: https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/1078/
it timed out, and a few dtests failed, but:
- most of the dtests that failed are failing on current master
- the other 3 failed due to a timeout - the test code had trouble connecting to the nodes - but I ran these dtests locally and they passed. Must have been a Jenkins machine hiccup.

kbr-:
v2:
- added a regression test that tries to ensure no one forgets assigning the custom sharder to new schema tables
- updated commit message and PR cover letter
- rebased

avikivity:
I think the important point is that there is no global shard_of() that does not accept a schema. It's still possible to create a dht::sharder, but the only sharder constructors I see are in schema.cc, in repair (which isn't applicable to schema tables), and in tests.

PR #7951, opened 2021-01-22
test: mutation_test: fix initialization order bug with thread local storage

avikivity:
test_cell_external_memory_usage uses with_allocator() to observe how some
types allocate memory. However, compiler reordering (observed with clang 11
on aarch64) can move the various thread-local CQL type object initialization
into the with_allocator() scope; so any managed object allocated as part of
this initialization also gets measured, and the test fails. The code movement
is legal, as far as I can tell.

Fix this by initializing the type object early; use an atomic_thread_fence
as an optimization barrier so the compiler doesn't eliminate the or move
the early initialization.

### DIFF
--- a/test/boost/mutation_test.cc
+++ b/test/boost/mutation_test.cc
@@ -1980,9 +1980,19 @@ SEASTAR_THREAD_TEST_CASE(test_cell_equals) {
     BOOST_REQUIRE(!c4.equals(*bytes_type, c1));
 }
 
+// Global to avoid elimination by the compiler; see below for use
+thread_local data_type force_type_thread_local_init_evaluation [[gnu::used]];
+
 SEASTAR_THREAD_TEST_CASE(test_cell_external_memory_usage) {
     measuring_allocator alloc;
 
+    // Force evaluation of int32_type and all the other types. This is so
+    // the compiler doesn't reorder it into the with_allocator section, below,
+    // and any managed_bytes instances creates during that operation interfere
+    // with the measurements.
+    force_type_thread_local_init_evaluation = int32_type;
+    // optimization barrier
+    std::atomic_signal_fence(std::memory_order_seq_cst);
 
     auto test_live_atomic_cell = [&] (data_type dt, bytes_view bv) {
         with_allocator(alloc, [&] {


avikivity:
Update: made the global variable thread_local to avoid a debug mode assert failure when a lw_shared_ptr is destroyed on the wrong shard.

test/boost/mutation_test.cc
```diff
@@ -1980,9 +1980,19 @@ SEASTAR_THREAD_TEST_CASE(test_cell_equals) {
     BOOST_REQUIRE(!c4.equals(*bytes_type, c1));
 }
 
+// Global to avoid elimination by the compiler; see below for use
+thread_local data_type force_type_thread_local_init_evaluation [[gnu::used]];
+
 SEASTAR_THREAD_TEST_CASE(test_cell_external_memory_usage) {
     measuring_allocator alloc;
 
+    // Force evaluation of int33_type and all the other types. This is so
```
nyh, line 1989: int32 :-)

test/boost/mutation_test.cc
```diff
@@ -1980,9 +1980,19 @@ SEASTAR_THREAD_TEST_CASE(test_cell_equals) {
...
+    // the compiler doesn't reorder it into the with_allocator section, below,
+    // and any managed_bytes instances creates during that operation interfere
+    // with the measurements.
+    force_type_thread_local_init_evaluation = int32_type;
```
nyh, line 1993: just curious, wouldn't a simple
```c++
(void) int2_type;
```
work too? Is that variable really necessary just to avoid optimization?

nyh (approved):
Looks good but I just have a question whether it could be simpler

test/boost/mutation_test.cc
avikivity: off-by-one

test/boost/mutation_test.cc
avikivity: The compiler can optimize your code to nothing.

The variable isn't needed to avoid optimization, it's needed to avoid code motion. The sequence "assign to global variable; signal fence" cannot be reordered, since the compiler can't be sure a signal won't examine the variable after the fence.

avikivity:
v2: fix off-by-one in comment

tgrabiec (approved):

Issue #3612, opened 2018-07-19
Nodetool cfstats in LCS is not shard aware


slivne:
```
nodetool cfstats 
.
.

 SSTables in each level: [49/4, 285/10, 2967/100, 9461/1000]
```

the values do not make any sense

avikivity:
Why open a duplicate issue?

amnonh:
I summarize here my take on the issue, so it would remain for future discussions.

The nodedool gets all of its information from the scylla-jmx using jmx, to do that it uses MBean (predefined java interface objects).
The scylla-jmx uses the Rest API to get information from Scylla.

We can (and would) change any of this components but should do so in a way that will not break compatibility for users of the MBeans.

We need to either propagate the number of shards or the maximal number of sstables per level from scylla to the nodetool.

We don't want to change the current MBean, so we have those options:
* Manage to piggyback on an existing MBean call
* Add our own MBean, in which we can put whatever functionality we want.

tzach:
> * Manage to piggyback on an existing MBean call
> * Add our own MBean, in which we can put whatever functionality we want.

seem like the path of least resistance

> * Add it as a jmx-metric, metrics have their way to dynamically register with a unified interface,
> so a new metrics can be created with no need to define an MBEan

Are we using jmx-metric? if not, we should not introduce it now.
The long-term solution is moving away from JMX, so we should do our best not to introduce new JMX base tech into the mix.

amnonh:
We cannot avoid using jmx as long as we are using the current nodetool, even if we are going to replace nodetool in the future with something that would use our API directly, there are still some who uses jmx directly, so we should not modify existing MBean.

I think adding a shard/smp metrics is the simplest, but I'll see how the code is going to look first.

amnonh:
I have an end-to-end solution that I'm trying to push gradualy:
Scylla
Scylla API would expose the number of shards using:
http://localhost:10000/system/shards

JMX
A new metric in the storage service metrics (Storage service acts as Cassandra's miscellaneous drawer for anything that does not have a specific location)

Nodetool
NodeProbe would have a wrapper to get the number of shards
TableStats would multiply the number of sstables by the number of shards.

Note: after this change, the behavior of nodetool tablestats would be different.
Nodetool show the n/m (n sstables out of maximum m per level) **only** if n> m.
if n <m it would show n.

up until now, we had many cases that n is greater than m, but this will rarely be the case anymore

amnonh:
[...]

PR #7995, opened 2021-01-31
API: column_family per_level, add max per level

amnonh:
Sometimes it's useful not to know the maximum sstable per level not only
the number of sstable per level.

Specifically, it is useful for nodetool tablestats

This patch adds an optional parameter to GET
/column_family/sstables/per_level/ API.

If add_max is set to true, the returned array would contain an
additional entry per level with the maximum value.

Following changes are required in the JMX to support it.

Relates to #3612


mykaul:
@amnonh - still relevant?

PR #8005, opened 2021-02-01
dist/redhat: stop using systemd macros, call systemctl directly

syuu1228:
This is v2 of #5605, since it's dequeued.

----

Fedora version of systemd macros does not work correctly on CentOS7,
since CentOS7 does not support "file trigger" feature.
To fix the issue we need to stop using systemd macros, call systemctl
directly.

See scylladb/scylla-jmx#94

### DIFF
--- a/dist/redhat/scylla.spec
+++ b/dist/redhat/scylla.spec
@@ -78,13 +78,18 @@ getent passwd scylla || /usr/sbin/useradd -g scylla -s /sbin/nologin -r -d %{_sh
 %post server
 /opt/scylladb/scripts/scylla_post_install.sh
 
-%systemd_post scylla-server.service
+if [ $1 -eq 1 ] ; then
+    /usr/bin/systemctl preset scylla-server.service ||:
+fi
 
 %preun server
-%systemd_preun scylla-server.service
+if [ $1 -eq 0 ] ; then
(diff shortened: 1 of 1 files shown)


dist/redhat/systemd.inc
```diff
@@ -0,0 +1,19 @@
...
+%undefine systemd_post
+%define systemd_post() %{expand:%systemd_daemon_reload}
+
+%undefine systemd_preun
+%define systemd_preun() \
+if [ $1 -eq 0 ] && [ -x /usr/bin/systemctl ]; then \
+    /usr/bin/systemctl disable %1 ||: \
+    /usr/bin/systemctl stop %1 ||: \
+fi \
+%{nil}
```
avikivity, line 16: Please avoid copied LGPL code, instead insert the commands directly in the %post, %postun, %preun, etc. sections.

syuu1228: Well it's re-implemented from scratch after you mentioned LGPL issue: https://github.com/scylladb/scylla/pull/5605#pullrequestreview-345154690

avikivity: There's no need for a .inc file. You can just add the commands to the scriptlet.

syuu1228: Dropped .inc file, replaced %systemd_* macros with systemctl commands.

PR #8007, opened 2021-02-01
API: Fix aggregation in column_familiy

### LINKED ISSUE #4513: Nodetool tablestats reports inconsistent bloom filter size

*Installation details*
Scylla version (or git commit hash): 3.0.6
Cluster size: 6 nodes
OS (RHEL/CentOS/Ubuntu/AWS AMI): CentOS 7.5.1804

We tried to make a BF_size(num_partitions) plot (https://plot.ly/~Bychin/42.embed) out of `nodetool tablestats` output and it looks like BF size stays ~const with the number of partitions constantly growing, while I would expect  it to constantly grow. According to our research in dev 1B partitions would require ~600MB BF with 10%FP. `ls` of bloom filters in the data directory reports something closer to 600MB rather than 60MB.

We have TWCS, writing ~300M new unique partitions every week, never expiring old data. To me it looks like `nodetool tablestats` reports size of something close to BF of one time window.

amnonh:
Few method in column_familiy API were doing the aggregation wrong,
specifically, bloom filter disk size.

The issue is not always visible, it happens when there are multiple
filter files per shard.

Fixes #4513

### DIFF
--- a/api/column_family.cc
+++ b/api/column_family.cc
@@ -656,7 +656,7 @@ void set_column_family(http_context& ctx, routes& r) {
     cf::get_bloom_filter_disk_space_used.set(r, [&ctx] (std::unique_ptr<request> req) {
         return map_reduce_cf(ctx, req->param["name"], uint64_t(0), [] (column_family& cf) {
             return std::accumulate(cf.get_sstables()->begin(), cf.get_sstables()->end(), uint64_t(0), [](uint64_t s, auto& sst) {
-                return sst->filter_size();
+                return s + sst->filter_size();
             });
         }, std::plus<uint64_t>());
     });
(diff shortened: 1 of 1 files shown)


bhalevy (approved):

api/column_family.cc
```diff
@@ -656,47 +656,47 @@ void set_column_family(http_context& ctx, routes& r) {
     cf::get_bloom_filter_disk_space_used.set(r, [&ctx] (std::unique_ptr<request> req) {
         return map_reduce_cf(ctx, req->param["name"], uint64_t(0), [] (column_family& cf) {
             return std::accumulate(cf.get_sstables()->begin(), cf.get_sstables()->end(), uint64_t(0), [](uint64_t s, auto& sst) {
-                return sst->filter_size();
+                return s + sst->filter_size();;
```
penberg, line 659: Double semicolon (";") at the end of the line.

penberg (changes requested):
Looks good, minor nit.

api/column_family.cc
amnonh: oops,

@penberg I've forced push a fix that remove it

PR #8011, opened 2021-02-01
Fix mixed cluster schema sync

eliransin:
When a table is altered in a mixed cluster by a node with a more recent version, the request can
fail if there is a difference in schema_features between the two versions.
This miniset handles the two problems that prevents the sync.

### DIFF
--- a/db/schema_tables.cc
+++ b/db/schema_tables.cc
@@ -783,6 +783,13 @@ future<std::vector<canonical_mutation>> convert_schema_to_mutations(distributed<
 
 std::vector<mutation>
 adjust_schema_for_schema_features(std::vector<mutation> schema, schema_features features) {
+    //Don't send the `computed_columns` table mutations to nodes that doesn't know it.
+    if (!features.contains(schema_feature::COMPUTED_COLUMNS)) {
+        schema.erase(std::remove_if(schema.begin(), schema.end(), [] (const mutation& m) {
+            return m.schema()->cf_name() == COMPUTED_COLUMNS;
+        }) , schema.end());
+    }
+
     for (auto& m : schema) {
         m = redact_columns_for_missing_features(m, features);
     }
--- a/db/schema_tables.cc
+++ b/db/schema_tables.cc
@@ -700,7 +700,9 @@ redact_columns_for_missing_features(mutation m, schema_features features) {
         return std::move(m);
     }
     slogger.debug("adjusting schema_tables mutation due to possible in-progress cluster upgrade");
-    m.upgrade(scylla_tables(features));
+    // The global schema ptr make sure it will be registered in the schema registry.
+    global_schema_ptr redacted_schema{scylla_tables(features)};
+    m.upgrade(redacted_schema);
     return std::move(m);
 }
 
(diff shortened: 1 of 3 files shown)


eliransin:
/cc @slivne  this is the fix for scylladb/scylla-enterprise#1612

@avikivity / @tgrabiec I wonder about the right way to merge this:
The first patch is generic and it is relevant for whenever we add columns to `scylla_tables`
But the second one is specific to the transition from 2019 to 2020. However it can be relevant as a boylerplate code for when
we add more schema tables.
What do you think? How should I split / submit this patchset  ( assuming you agree with the patches themselves of course)

db/schema_tables.cc
```diff
@@ -700,7 +700,9 @@ redact_columns_for_missing_features(mutation m, schema_features features) {
         return std::move(m);
     }
     slogger.debug("adjusting schema_tables mutation due to possible in-progress cluster upgrade");
-    m.upgrade(scylla_tables(features));
+    // The global schema ptr make sure it will be registered in the schema registry.
+    global_schema_ptr redacted_schema{scylla_tables(features)};
+    m.upgrade(redacted_schema);
     return std::move(m);
```
avikivity, line 706: Will it stay alive long enough?

avikivity: global_schema_ptr doesn't extend the life of a schema indefinitely.

eliransin: It will last until next reboot. The schema_ptr that it is registering is stored in a static thread local array in the scylla_tables function.

avikivity: Ah! yes.

eliransin:
@tgrabiec can you review given that you are the code owner and you are the only one that really gets the full picture in those schema issues.
Thanks,

tgrabiec (approved):

eliransin:
I seam to have broken next with this PR. Please revert and I'll debug it
offline.

psarna:
dequeued from /next

eliransin:
@tgrabiec - I had a bug which I fixed. It was twofold:
1.  the previous version of 7660550 had a stupid bug that resulted in an empty mutation when the `COMPUTED_COLUMNS` feature was on, which resulted in nodes that can't get to schema agreement since the schema merge was always with empty mutation vector in a homogenic cluster.
2. There was no schema recalculation trigger upon enabling the 'COMPUTED_COLUMN' feature. This I believe could have caused a lot of problem also not in the context of this PR.

psarna:
Was the change tested with (a subset of) the failed dtests from last failed /next runs?

eliransin:
@psarna it passed all of them but @tgrabiec had a change request for 3179739 , he asked why don't we use remove_if,
the problem with remove_if is that it (for reason that I now search) crashes the enterprise 2020.1 when backported.
This is probably related to the move constructor or the way those mutations are built, but I still don't know how it is related.
On OSS it runs perfectly fine.

eliransin:
found it - need to erase after remove if :facepalm:
will push a fix.

eliransin:
@tgrabiec  addressed the remove_if comment.

tgrabiec (approved):

Issue #7961, opened 2021-01-25
large allocation in untyped_result_set reading cdc generations (seastar_memory - oversized allocation: 2023424 bytes. This is non-fatal, but could lead to latency and/or fragmentation issues)


yakirgb:
This is Scylla's bug tracker, to be used for reporting bugs only.
If you have a question about Scylla, and not a bug, please ask it in
our mailing-list at scylladb-dev@googlegroups.com or in our slack channel.

- [x] I have read the disclaimer above, and I am reporting a suspected malfunction in Scylla.

*Installation details*
Scylla version: 4.3
Cluster size: 10 nodes / 38 shards / 250GB usage
OS: CentOS 7 , kernel 5

Platform: physical
Hardware: cores=38 memory=125GB
Disks: 1 NVME of 3TB

Getting errors of during startup of:
```
-- Logs begin at Thu 2020-10-22 10:50:24 UTC, end at Mon 2021-01-25 10:33:05 UTC. --
Jan 25 10:26:45 r1.tab.com scylla[31756]:  [shard 0] seastar_memory - oversized allocation: 2023424 bytes. This is non-fatal, but could lead to latency and/or fragmentation issues. Please report: ...
```
Attached full backtrace to ticket.

Thank you, Yakir.

yakirgb:
Full backtrace:
```
-- Logs begin at Thu 2020-10-22 10:50:24 UTC, end at Mon 2021-01-25 10:33:05 UTC. --
Jan 25 10:26:45 r1.tab.com scylla[31756]:  [shard 0] seastar_memory - oversized allocation: 2023424 bytes. This is non-fatal, but could lead to latency and/or fragmentation issues. Please report: at    0x2fb56dd
                                                                    0x2fb5930
                                                                    0x2fb5d99
                                                                    0x2b05bc7
                                                                    0x2b07d4b
                                                                    0x1a23501
                                                                    0x1a25201
                                                                    0x1a26676
                                                                    0x1a28e97
                                                                    0x1a2b399
                                                                    0x1a228bf
                                                                    0x19cc165
                                                                    0x2b57137
                                                                    0x2b573ee
                                                                    0x2b96cd5
                                                                    0x2aead5e
                                                                    0x2aeb7ee
                                                                    0xd4c162
                                                                    /opt/scylladb/libreloc/libc.so.6+0x27041
                                                                    0xc2c98d
                                                                    --------
[...]

PR #8015, opened 2021-02-02
Untyped_result_set: make non-copying and fragment retaining

### LINKED ISSUE #8014: Make untyped_result_set non-copying and retain fragments
This type was not intending for what it is used for. (Meant for small, internal data reads). 
But now it is the center of CDC. 

Still, the type does deep copy+linearization of source data -> potentially very large allocs.

We should try to make it a non-copying proxy to (fragmented) backing storage.

elcallio:
Refs #7961
Fixes #8014

The "untyped_result_set" object was created for small, internal access to cql-stored metadata.
It is nowadays used for rather more than that (cdc). 
This has the potential of mixing badly with the fact that the type does deep copying of data
and linearizes all (not to mention handles multiple rows rather inefficiently).

Instead of doing a deep copy of input, we keep assume ownership and build
rows of the views therein, potentially retaining fragmented data as-is
avoiding premature linearization.

Note that this is not all sugar and flowers though. Any data access will
by nature be more expensive, and the view collections we create are
potentially just as expensive as copying for small cells.

Otoh, it allows writing code using this that avoids data copying,
depending on destination.

v2:
* Fixed wrong collection reserved in visitor
* Changed row index from shared ptr to ref
* Moved typedef
* Removed non-existing constructors
* Added const ref to index build
* Fixed raft usage after rebase

v3:
* Changed shared_ptr to unique

### DIFF
--- a/cql3/untyped_result_set.cc
+++ b/cql3/untyped_result_set.cc
@@ -47,58 +47,118 @@
 #include "result_set.hh"
 #include "transport/messages/result_message.hh"
 
-cql3::untyped_result_set_row::untyped_result_set_row(const map_t& data)
-    : _data(data)
+cql3::untyped_result_set_row::untyped_result_set_row(const index_map& index, const cql3::metadata& metadata, data_views data)
+    : _name_to_index(index)
+    , _metadata(metadata)
+    , _data(std::move(data))
 {}
 
-cql3::untyped_result_set_row::untyped_result_set_row(const std::vector<lw_shared_ptr<column_specification>>& columns, std::vector<bytes_opt> data)
-: _columns(columns)
-, _data([&columns, data = std::move(data)] () mutable {
-    map_t tmp;
-    std::transform(columns.begin(), columns.end(), data.begin(), std::inserter(tmp, tmp.end()), [](lw_shared_ptr<column_specification> c, bytes_opt& d) {
-       return std::make_pair<sstring, bytes_opt>(c->name->to_string(), std::move(d));
-    });
-    return tmp;
-}())
-{}
+size_t cql3::untyped_result_set_row::index(const std::string_view& name) const {
+    auto i = _name_to_index.find(name);
+    return i != _name_to_index.end() ? i->second : std::numeric_limits<size_t>::max();
+}
 
 bool cql3::untyped_result_set_row::has(std::string_view name) const {
-    auto i = _data.find(name);
-    return i != _data.end() && i->second;
+    auto i = index(name);
+    if (i < _data.size()) {
+        return !std::holds_alternative<std::monostate>(_data.at(i));
+    }
+    return false;
+}
+
+cql3::untyped_result_set_row::view_type cql3::untyped_result_set_row::get_view(std::string_view name) const {
+    return std::visit(make_visitor(
+        [](std::monostate) -> view_type { throw std::bad_variant_access(); },
+        [](const view_type& v) -> view_type { return v; },
+        [](const bytes& b) -> view_type { return view_type(b); }
+    ), _data.at(index(name)));
+}
+
+const std::vector<lw_shared_ptr<cql3::column_specification>>& cql3::untyped_result_set_row::get_columns() const {
+    return _metadata.get_names();
 }
 
 using cql_transport::messages::result_message;
 
-cql3::untyped_result_set::untyped_result_set(const cql3::result_set& rs) {
-    auto& cn = rs.get_metadata().get_names();
-    for (auto& r : rs.rows()) {
-        // r is const ref. TODO: make this more efficient by either wrapping result set
-        // or adding modifying accessors to it.
-        _rows.emplace_back(cn, r);
-    }
+cql3::untyped_result_set::index_map_ptr cql3::untyped_result_set::make_index(const cql3::metadata& metadata) {
+    auto map = std::make_unique<untyped_result_set_row::index_map>();
+    auto& names = metadata.get_names();
+    size_t i = 0;
+    std::transform(names.begin(), names.end(), std::inserter(*map, map->end()), [&](const lw_shared_ptr<column_specification>& c) mutable {
+        return std::make_pair<std::string_view, size_t>(c->name->text(), i++);
+    });
+    return map;
 }
 
+struct cql3::untyped_result_set::visitor {
+    rows_type& rows;
+    const cql3::metadata& meta;
+    const untyped_result_set_row::index_map& index;
+    untyped_result_set_row::data_views tmp;
+
+    visitor(rows_type& r, const cql3::metadata& m, const untyped_result_set_row::index_map& i)
+        : rows(r)
+        , meta(m)
+        , index(i)
+    {}
+
+    void start_row() {
+        tmp.reserve(index.size());
+    }
+    void accept_value(std::optional<query::result_bytes_view>&& v) {
+        if (v) {
+            tmp.emplace_back(std::move(*v));
+        } else {
+            tmp.emplace_back(std::monostate{});
+        } 
+    }
+    // somewhat weird dispatch, but when visiting directly via
+    // result_generator, pk:s will be temporary - and sent 
+    // as views, not opt_views. So we can dispatch on this and 
+    // simply copy the temporaries.
+    void accept_value(const query::result_bytes_view& v) {
+        tmp.emplace_back(v.linearize());
+    }
+    void end_row() {
+        rows.emplace_back(untyped_result_set_row(index, meta, std::exchange(tmp, {})));
+    }
+};
+
 cql3::untyped_result_set::untyped_result_set(::shared_ptr<result_message> msg)
-    : _rows([msg]{
-    class visitor : public result_message::visitor_base {
+    : _storage(msg)
+{
+    class msg_visitor : public result_message::visitor_base {
     public:
-        std::optional<untyped_result_set> res;
+        const cql3::result* res = nullptr;
         void visit(const result_message::rows& rmrs) override {
-            const auto& rs = rmrs.rs();
-            const auto& set = rs.result_set();
-            res.emplace(set); // construct untyped_result_set by const ref.
+            res = &rmrs.rs();
         }
     };
-    visitor v;
+    msg_visitor v;
     if (msg != nullptr) {
         msg->accept(v);
     }
     if (v.res) {
-        return std::move(v.res->_rows);
+        auto& metadata = v.res->get_metadata();
+        _index = make_index(metadata);
+        v.res->visit(visitor{_rows, metadata, *_index});
     }
-    return rows_type{};
-}())
-{}
+}
+
+cql3::untyped_result_set::untyped_result_set(const schema& s, foreign_ptr<lw_shared_ptr<query::result>> qr, const cql3::selection::selection& selection, const query::partition_slice& slice)
+    : _storage(std::make_tuple(std::move(qr), selection.get_result_metadata()))
+{
+    auto& qt = std::get<qr_tuple>(_storage);
+    auto& qres = std::get<0>(qt);
+    auto& metadata = *std::get<1>(qt);
+
+    _index = make_index(metadata);
+    visitor v{_rows, metadata, *_index};
+    result_generator::query_result_visitor<visitor> vv(s, v, selection);
+    query::result_view::consume(*qres, slice, vv);
+}
+
+cql3::untyped_result_set::~untyped_result_set() = default;
 
 const cql3::untyped_result_set_row& cql3::untyped_result_set::one() const {
     if (_rows.size() != 1) {
--- a/cql3/untyped_result_set.hh
+++ b/cql3/untyped_result_set.hh
@@ -41,36 +41,60 @@
  */
 #include <unordered_map>
 #include <optional>
+#include <seastar/core/sharded.hh>
 #include "bytes.hh"
 #include "types.hh"
 #include "types/map.hh"
 #include "types/list.hh"
 #include "types/set.hh"
+#include "serializer.hh"
+#include "bytes_ostream.hh"
 #include "transport/messages/result_message_base.hh"
 #include "column_specification.hh"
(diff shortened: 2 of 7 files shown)


cql3/untyped_result_set.hh
```diff
@@ -41,36 +41,61 @@
...
+    using opt_view_type = std::optional<view_type>;
+    using view_holder = std::variant<std::monostate, view_type, bytes>;
 private:
-    const std::vector<lw_shared_ptr<column_specification>> _columns;
-    using map_t = flat_hash_map<sstring, bytes_opt>;
-    const map_t _data;
+    friend class untyped_result_set;
+    using index_map = std::unordered_map<std::string_view, size_t>;
+    using data_views = std::vector<view_holder>;
+    using index_map_ptr = ::shared_ptr<index_map>;
```
haaawk, line 83: could this be a lw_shared_ptr instead of ::shared_ptr?

elcallio: I guess. The types are known, so... Could even be a ref + unique_ptr in parent object.

cql3/untyped_result_set.cc
```diff
@@ -47,58 +47,118 @@
...
+    }
+    return false;
+}
+
+cql3::untyped_result_set_row::view_type cql3::untyped_result_set_row::get_view(std::string_view name) const {
+    return std::visit(make_visitor(
+        [](std::monostate) -> view_type { throw std::bad_variant_access(); },
+        [](const view_type& v) -> view_type { return v; },
+        [](const bytes& b) -> view_type { return view_type(b); }
+    ), _data.at(index(name)));
```
haaawk, line 74: I never know whether we should be using at() and pay for safety or [] and avoid checks. @avikivity what's your take on this?

cql3/untyped_result_set.cc
```diff
@@ -47,58 +47,118 @@
...
-    for (auto& r : rs.rows()) {
-        // r is const ref. TODO: make this more efficient by either wrapping result set
-        // or adding modifying accessors to it.
-        _rows.emplace_back(cn, r);
-    }
+cql3::untyped_result_set_row::index_map_ptr cql3::untyped_result_set::make_index(const cql3::metadata& metadata) {
+    auto map = ::make_shared<untyped_result_set_row::index_map>();
+    auto& names = metadata.get_names();
+    size_t i = 0;
+    std::transform(names.begin(), names.end(), std::inserter(*map, map->end()), [&](lw_shared_ptr<column_specification> c) mutable {
```
haaawk, line 87: nit: I think const lw_shared_ptr<column_specification>& would work as well.

cql3/untyped_result_set.hh
```diff
@@ -188,8 +217,19 @@ public:
         return _rows.back();
     }
 private:
+    using qr_tuple = std::tuple<foreign_ptr<lw_shared_ptr<query::result>>, shared_ptr<const cql3::metadata>>;
+    using storage = std::variant<std::monostate
+        , ::shared_ptr<cql_transport::messages::result_message>
+        , qr_tuple
+    >;
+    struct visitor;
```
haaawk, line 225: Why do we have to expose this visitor in the header? Wouldn't it be enough to make it internal to .cc file?

cql3/untyped_result_set.hh
```diff
@@ -188,8 +217,19 @@ public:
...
+
+    storage _storage;
+    untyped_result_set_row::index_map_ptr _index;
     rows_type _rows;
+
     untyped_result_set() = default;
+    static untyped_result_set_row::index_map_ptr make_index(const cql3::metadata&);
```
haaawk, line 232: I know it is a matter of taste but maybe it would be better to hide this function inside .cc file and not expose it in the header since it's only used internally?

cql3/untyped_result_set.hh
```diff
@@ -147,23 +172,27 @@ public:
...
-    typedef std::vector<row> rows_type;
+    using rows_type = std::vector<row>;
     using const_iterator = rows_type::const_iterator;
     using iterator = rows_type::const_iterator;
 
     untyped_result_set(::shared_ptr<cql_transport::messages::result_message>);
-    untyped_result_set(const cql3::result_set&);
+    untyped_result_set(const schema&, foreign_ptr<lw_shared_ptr<query::result>>, const cql3::selection::selection&, const query::partition_slice&);
+    untyped_result_set(cql3::result_set&&);
+    untyped_result_set(cql3::result&&);
```
haaawk, line 193: Where are those two constructors implemented? I couldn't find their definitions. Are they used at all?

cql3/untyped_result_set.hh
elcallio: namespacing + private types + friend relationship.

cql3/untyped_result_set.hh
elcallio: same as above.

cql3/untyped_result_set.cc
elcallio: When things are not in hot path, I prefer the checked accessors, just to make sure we catch any borked code.

cql3/untyped_result_set.hh
elcallio: They are not supposed to be there. Sorry, forgot to prune...

cql3/untyped_result_set.hh
haaawk: I don't follow. Could you elaborate a bit more for weaker audience ;).

cql3/untyped_result_set.cc
```diff
@@ -47,58 +47,118 @@
...
+    untyped_result_set_row::data_views tmp;
+
+    visitor(rows_type& r, const cql3::metadata& m, untyped_result_set_row::index_map_ptr i)
+        : rows(r)
+        , meta(m)
+        , index(i)
+    {}
+
+    void start_row() {
+        rows.reserve(index->size());
```
haaawk, line 106: Shouldn't this be tmp.reserve(...)?

cql3/untyped_result_set.cc
```diff
@@ -47,58 +47,118 @@
...
+        if (v) {
+            tmp.emplace_back(std::move(*v));
+        } else {
+            tmp.emplace_back(std::monostate{});
+        } 
+    }
+    // somewhat weird dispatch, but when visiting directly via
+    // result_generator, pk:s will be temporary - and sent 
+    // as views, not opt_views. So we can dispatch on this and 
+    // simply copy the temporaries.
```
haaawk, line 118: Would it be possible to move from those temporaries instead of copying from them?

elcallio: Maybe. Not sure if that violates the contract.

cql3/untyped_result_set.cc
elcallio: Very much so. Me so sloppy.

haaawk (approved):
LGTM

cql3/untyped_result_set.hh
elcallio: The visitor/static below use types and constructors that I made private in the row type (because they should be internal). untyped_result_set can see them because it is a friend. Top-level routines/types cannot. Can make more elaborate friend declarations, but to me at least it seems easier to just do these harmless (and private) declarations instead. 
Also, I prefer nested types to file-local ones, if for nothing else so for scoping of concern.

haaawk: Got you. I was looking at a wrong friendship relation (result_generator <-> untyped_result_set) and couldn't figure out why it matters :)

types/listlike_partial_deserializing_iterator.hh
```diff
@@ -28,6 +28,12 @@
 int read_collection_size(bytes_view& in, cql_serialization_format sf);
 bytes_view read_collection_value(bytes_view& in, cql_serialization_format sf);
 
+template <FragmentedView View>
+int read_collection_size(View& in, cql_serialization_format);
+
+template <FragmentedView View>
+View read_collection_value(View& in, cql_serialization_format);
+
```
avikivity, line 36: How does it work? These are templates, if they are defined in a source file, then linkage may break if used with the wrong types. If they are defined in a header file, then this is unneeded.

Is it just needed as a forward declaration?

cql3/untyped_result_set.cc
avikivity: Agree with @elcallio. These less-hot code paths are also more error-prone.

cql3/untyped_result_set.hh
avikivity: I agree about nesting, but the #include dependency addition is unfortunate. Maybe hide the variant behind a unique_ptr? It's just a placeholder after all, no?

Also, std::any? probably doesn't work due to copyablity.

types/listlike_partial_deserializing_iterator.hh
elcallio: These templates are explicitly instantiated in types.cc, so magically enough it works (no compile/linkage errors, even without an extern template decl - same as collection type deserialize templates).

cql3/untyped_result_set.hh
elcallio: I felt that yet another level of indirection was perhaps pushing it... Not happy about the extra indirection - really with the foreign_ptr was not so an lw_shared_ptr... Even tried to work around the forward declaration aspect there, but it just becomes terrible. 
std::any does not work because of move-only types.

types/listlike_partial_deserializing_iterator.hh
avikivity: ok.

elcallio:
Updated and rebased
v2:
* Fixed wrong collection reserved in visitor
* Changed row index from shared ptr to ref
* Moved typedef
* Removed non-existing constructors
* Added const ref to index build
* Fixed raft usage after rebase

cql3/untyped_result_set.cc
elcallio: No, you cannot, since the args are views, not the actual temporaries. And changing that in turn would break other code, which I guess is fixable, but probably not worth it... At least not for now.

cql3/untyped_result_set.cc
haaawk: Isn't this code executed at least once for each request? I thought that whatever gets executed 'per request' is already a hot path, no? I understand that we have a more hot path when something gets executed 'per sstable' but still once per request can be 1-5M times / s, no?

elcallio: That is not a hot path. An inner loop is a hot path.

haaawk: It seems that every person has its own definition of a hot path :). @tgrabiec was always telling me 'per request' operation is a hot path and I sort of agree considering that we're aiming for 5M ops/s

elcallio: That depends on the context. For a bounds check it is not. For a major allocation or prime number calculations it might be.

cql3/untyped_result_set.hh
```diff
@@ -188,8 +214,20 @@ public:
...
+    using index_map_ptr = ::lw_shared_ptr<untyped_result_set_row::index_map>;
+    using qr_tuple = std::tuple<foreign_ptr<lw_shared_ptr<query::result>>, shared_ptr<const cql3::metadata>>;
+    using storage = std::variant<std::monostate
+        , ::shared_ptr<cql_transport::messages::result_message>
+        , qr_tuple
+    >;
+    struct visitor;
+
+    storage _storage;
+    index_map_ptr _index;
```
kbr-, line 226: Does this have to be a pointer? If I understand correctly, both `untyped_result_set::visitor` and `untyped_result_set_row` lifetimes are contained in `untyped_result_set`'s lifetime, so `untyped_result_set` could just store an `index_map` directly and these guys would have a reference to it.

kbr- (approved):
one nit/question, otherwise LGTM

cql3/untyped_result_set.hh
elcallio: I want both types to be movable. Having the index stored inside untyped_result_set wound break refs in the row objects.

kbr-: Then `unique_ptr`?

elcallio: I guess it can be. now that we made the refs in rows actual refs. Not a huge deal though.

elcallio: changed.

cql3/untyped_result_set.cc
```diff
@@ -47,58 +47,118 @@
...
+    std::transform(names.begin(), names.end(), std::inserter(*map, map->end()), [&](const lw_shared_ptr<column_specification>& c) mutable {
+        return std::make_pair<std::string_view, size_t>(c->name->text(), i++);
+    });
+    return map;
 }
 
+struct cql3::untyped_result_set::visitor {
+    rows_type& rows;
+    const cql3::metadata& meta;
+    const index_map_ptr& index;
```
kbr-, line 96: Now moving `untyped_result_set` will cause a dangling reference, no?
I had something different in mind, i.e. `const index_map&`, which is a reference to whatever the `unique_ptr` points to (and that something doesn't move when we move the owning `untyped_result_set`)

kbr-: Ok nvm it's just the `visitor`, `untyped_result_set` doesn't move during the lifetime of the `visitor` so there's no problem.
But still, why `ptr&` instead of just `&`

elcallio: Smallest possible change. And avoid using long type reference.

kbr-: k

elcallio:
ping?

kbr- (approved):

elcallio:
ping?

slivne:
@avikivity kamil and piotrj blessed this - and calle addressed your comments - ping

avikivity:
Dequeued, unit tests don't compile.

elcallio:
Rebased. Fixed implicit linkage error.

PR #8018, opened 2021-02-02
Split CDC streams table partitions into clustered rows 

### LINKED ISSUE #7993: Improve CDC stream description (user-facing) tables
In order to query CDC log tables properly, users need to know the set of stream IDs. In Scylla 4.3 this is done by querying the `system_distributed.cdc_streams_descriptions` table, which has the following schema:
```
CREATE TABLE system_distributed.cdc_streams_descriptions (
    time timestamp PRIMARY KEY,
    expired timestamp,
    streams frozen<set<blob>>
) ...
```
The `time` column denotes the timestamp of a CDC generation and `streams` is the set of all stream IDs in that generation (the `expired` column is not relevant in this discussion).

Unfortunately, the set of streams under each timestamp can get quite large. Its size is O(N * S * V), where N is the number of nodes in the cluster, S is the maximum number of shards over the set of nodes, and V is the number of vnodes in the token ring per node. As an example, with 100 nodes, 64 shards, and 256 tokens per node, there would be around 1,6M stream IDs and this set (including metadata to store it in serialized form) would weigh about 32MB.

Querying tables containing such large collections is "hard". Scylla doesn't do it efficiently: it performs a large continuous allocation when reading the set and processes it without yielding, causing a reactor stall.

We need to use a better schema for this table that uses many clustered rows of small size.

Proposed schema
---

I propose the following schema:
```
CREATE TABLE system_distributed.cdc_streams_descriptions_v2 (
    time timestamp,
    stream_id blob,
    expired timestamp static,
    ok boolean static,
    PRIMARY KEY (time, stream_id)
) ...
```

First, observe that the name has changed. This is to make upgrade easy: if the old table was already being used, it stays and won't conflict with the new table. The user will be able to remove the old table after upgrade if they so desire.

We use `stream_id` as the clustering key, so each stream ID is kept in a separate row. Thus the rows have a small, constant size.

A new boolean static column `ok` is added. This is required to handle failures when updating the table, as described in the next section.

We can consider an alternative schema:
```
CREATE TABLE system_distributed.cdc_streams_descriptions_v2 (
    time timestamp,
    vnode_end bigint,
    expired timestamp static,
    ok boolean static,
    streams set<blob>,
    PRIMARY KEY (time, vnode_end)
) ...
```

This groups stream IDs according to the token ring vnodes that were used to generate them. We would be still using collections, but they are O(S), hence of reasonable size (and each partition would have O(N*V) rows).

This schema could also make the table potentially easier to update since we need to insert fewer rows in order to insert a whole generation of streams.

Updating the table
---

Today the table is updated as follows:
1. A node notices a new CDC generation (through gossip).
2. The node queries the `time` column of `cdc_streams_descriptions` to check if the generation is already inserted (it may have been, and likely has been, inserted by another node; but all nodes do the check for fault tolerance).
3. If not, it retrieves the generation (how exactly this is done is not relevant to the discussion; today it queries the `cdc_generation_descriptions` table and we want to change that - see #7985 - but it's an orthogonal problem).
4. Then it computes the list of streams and inserts it to the table using a single `INSERT` statement.

The node keeps retrying steps 2-4 if one of them fails, until it inserts the generation or notices in step 2 that it's already inserted.
Note that multiple nodes racing to insert the same generation is not a problem since the operation is idempotent.

Eventually at least one node should manage to insert the streams. There is one scenario in which the list of streams won't be inserted: a new generation would need to be created and then all nodes would have to crash/restart before any of them manages to insert the old one; after restart they would not remember about inserting the old generation any more but focus only on the new one.

This scenario could by ruled out by having the nodes scan all existing generations on restart (and seeing if any needs to be inserted into the streams table), but the chances of that scenario happening seem to be miniscule. However, to be safe, we can use this opportunity to improve that part as well.

With the new table, steps 2 and 4 must be changed.

In step 2, instead of querying `time`, the node would query the `ok` column. The reason is that the partition may exist, but the generation may not have been inserted fully (since now it consists of multiple rows, which may be inserted multiple steps, as described below). However, the new solution ensures that if `ok` is `true`, it means that the whole generation was inserted.

In step 4 we can no longer use a single insert, as we're dealing with multiple rows. A potential option may be to use a single batch statement, but I'm not sure if Scylla can handle such large batches. For example, consider the cluster of 100 nodes described previously. With the first schema proposal we need to insert ~1,6M rows (~N * V * S) and with the second we need to insert ~25K rows (~N * V).

Perhaps we can use an internal mutation API to create the mutation directly - so we don't need to go through CQL - and then the number of rows shouldn't matter that much. Currently we're issuing a single CQL query to insert the entire generation. But I'm not sure whether Scylla can easily handle sending a 30MB mutation or not. I'm guessing such mutations do appear during large queries.

Another option is to insert the list of streams in multiple CQL queries, grouping the rows into batches and performing a round-trip for each batch.

Finally, after inserting all the rows, the node must update the `ok` column to denote that the streams have been fully inserted. The clients, and other nodes considering if they should insert the streams themselves, can then check this column to make the right decisions.

Upgrading
---

We don't necessarily need to do any special handling for the upgrade procedure. The upgraded node can start using the new table right away (note that it's a different table since it has a different name). Thus it will insert the streams of the latest generation (the one with the highest timestamp) into the new table. Hence the latest generation will be in both the old and the new table (assuming that the old table existed; if it didn't, i.e. we're jumping versions, there's nothing to consider). All further generations will be inserted to the new table, unless a new generation is created during rolling upgrade, in which it will be inserted to the old table by old nodes and to the new table by new nodes (that should never happen, but if it does, it's harmless).

It may potentially be a problem that generation other than the latest one operating during upgrade will not be present in the new table (only in the old). For example, if the clients are using only the newer table (which they eventually should), they won't observe the previous generations. This can be probably handled by the clients themselves in most cases, e.g. by consuming the data in previous generations, and only then performing the upgrade (so all unconsumed data is in generations that were created after the upgrade or in the latest generation operating during the upgrade, which is also inserted into the new table). Otherwise, if required, we could:
1. provide a tool that rewrites streams from the old table to the new table,
2. or implement the previously mentioned improvement, i.e. have nodes scan all generations on restart and insert them to the streams table if necessary; in this case, the old generations would be automatically inserted as well.

But if we implement this new table and backport to 4.3 quickly, it's unlikely that any client will run into this problem.

---

cc @haaawk @slivne @avikivity @elcallio @nyh @piodul @avelanarius

kbr-:
Until now, the lists of streams in the `cdc_streams_descriptions` table
for a given generation were stored in a single collection. This solution
has multiple problems when dealing with large clusters (which produce
large lists of streams):
1. large allocations
2. reactor stalls
3. mutations too large to even fit in commitlog segments

This commit changes the schema of the table as described in issue #7993.
The streams are grouped according to token ranges, each token range
being represented by a separate clustering row. Rows are inserted in
reasonably large batches for efficiency.

The table is renamed to enable easy upgrade. On upgrade, the latest CDC
generation's list of streams will be (re-)inserted into the new table.

Yet another table is added: one that contains only the generation
timestamps clustered in a single partition. This makes it easy for CDC
clients to learn about new generations. It also enables an elegant
two-phase insertion procedure of the generation description: first we
insert the streams; only after ensuring that a quorum of replicas
contains them, we insert the timestamp. Thus, if any client observes a
timestamp in the timestamps table (even using a ONE query),
it means that a quorum of replicas must contain the list of streams.

---

Nodes automatically ensure that the latest CDC generation's list of
streams is present in the streams description table. When a new
generation appears, we only need to update the table for this
generation; old generations are already inserted.

However, we've changed the description table (from
`cdc_streams_descriptions` to `cdc_streams_descriptions_v2`). The
existing mechanism only ensures that the latest generation appears in
the new description table. We add an additional procedure that
rewrites the older generations as well, if we find that it is necessary
to do so (i.e. when some CDC log tables may contain data in these
generations).

### DIFF
--- a/alternator/streams.cc
+++ b/alternator/streams.cc
@@ -499,19 +499,11 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
     // TODO: creation time
 
     auto normal_token_owners = _proxy.get_token_metadata_ptr()->count_normal_token_owners();
-    // cannot really "resume" query, must iterate all data. because we cannot query neither "time" (pk) > something,
-    // or on expired...
-    // TODO: maybe add secondary index to topology table to enable this?
-    return _sdks.cdc_get_versioned_streams({ normal_token_owners }).then([this, &db, schema, shard_start, limit, ret = std::move(ret), stream_desc = std::move(stream_desc), ttl](std::map<db_clock::time_point, cdc::streams_version> topologies) mutable {
 
-        // filter out cdc generations older than the table or now() - cdc::ttl (typically dynamodb_streams_max_window - 24h)
-        auto low_ts = std::max(as_timepoint(schema->id()), db_clock::now() - ttl);
+    // filter out cdc generations older than the table or now() - cdc::ttl (typically dynamodb_streams_max_window - 24h)
+    auto low_ts = std::max(as_timepoint(schema->id()), db_clock::now() - ttl);
 
-        auto i = topologies.lower_bound(low_ts);
-        // need first gen _intersecting_ the timestamp.
-        if (i != topologies.begin()) {
-            i = std::prev(i);
-        }
+    return _sdks.cdc_get_versioned_streams(low_ts, { normal_token_owners }).then([this, &db, shard_start, limit, ret = std::move(ret), stream_desc = std::move(stream_desc)] (std::map<db_clock::time_point, cdc::streams_version> topologies) mutable {
 
         auto e = topologies.end();
         auto prev = e;
@@ -519,9 +511,7 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
 
         std::optional<shard_id> last;
 
-        // i is now at the youngest generation we include. make a mark of it.
-        auto first = i;
-
+        auto i = topologies.begin();
         // if we're a paged query, skip to the generation where we left of.
         if (shard_start) {
             i = topologies.find(shard_start->time);
@@ -547,7 +537,7 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
         };
 
         // need a prev even if we are skipping stuff
-        if (i != first) {
+        if (i != topologies.begin()) {
             prev = std::prev(i);
         }
 
--- a/cdc/generation.cc
+++ b/cdc/generation.cc
@@ -547,6 +547,11 @@ future<> maybe_rewrite_streams_descriptions(
         co_return;
     }
 
+    if (db.get_config().cdc_dont_rewrite_streams()) {
+        cdc_log.warn("Stream rewriting disabled. Manual administrator intervention may be required...");
+        co_return;
+    }
+
     // For each CDC log table get the TTL setting (from CDC options) and the table's creation time
     std::vector<time_and_ttl> times_and_ttls;
     for (auto& [_, cf] : db.get_column_families()) {
--- a/cdc/generation.cc
+++ b/cdc/generation.cc
@@ -382,7 +382,7 @@ static void do_update_streams_description(
         db::system_distributed_keyspace& sys_dist_ks,
         db::system_distributed_keyspace::context ctx) {
     if (sys_dist_ks.cdc_desc_exists(streams_ts, ctx).get0()) {
-        cdc_log.debug("update_streams_description: description of generation {} already inserted", streams_ts);
+        cdc_log.info("Generation {}: streams description table already updated.", streams_ts);
         return;
     }
 
@@ -393,14 +393,7 @@ static void do_update_streams_description(
         throw std::runtime_error(format("could not find streams data for timestamp {}", streams_ts));
     }
 
-    std::set<cdc::stream_id> streams_set;
-    for (auto& entry: topo->entries()) {
-        streams_set.insert(entry.streams.begin(), entry.streams.end());
-    }
-
-    std::vector<cdc::stream_id> streams_vec(streams_set.begin(), streams_set.end());
-
-    sys_dist_ks.create_cdc_desc(streams_ts, streams_vec, ctx).get();
+    sys_dist_ks.create_cdc_desc(streams_ts, *topo, ctx).get();
     cdc_log.info("CDC description table successfully updated with generation {}.", streams_ts);
 }
 
--- a/cdc/generation.cc
+++ b/cdc/generation.cc
@@ -28,6 +28,7 @@
 
 #include "keys.hh"
 #include "schema_builder.hh"
+#include "database.hh"
 #include "db/config.hh"
 #include "db/system_keyspace.hh"
 #include "db/system_distributed_keyspace.hh"
@@ -38,6 +39,7 @@
 #include "gms/gossiper.hh"
 
 #include "cdc/generation.hh"
+#include "cdc/cdc_options.hh"
 
 extern logging::logger cdc_log;
 
@@ -426,6 +428,173 @@ void update_streams_description(
             }
         });
     }
+}
+
+static db_clock::time_point as_timepoint(const utils::UUID& uuid) {
+    return db_clock::time_point{std::chrono::milliseconds(utils::UUID_gen::get_adjusted_timestamp(uuid))};
+}
+
+static future<std::vector<db_clock::time_point>> get_cdc_desc_v1_timestamps(
+        db::system_distributed_keyspace& sys_dist_ks,
+        abort_source& abort_src,
+        const noncopyable_function<unsigned()>& get_num_token_owners) {
+    while (true) {
+        try {
+            co_return co_await sys_dist_ks.get_cdc_desc_v1_timestamps({ get_num_token_owners() });
+        } catch (...) {
+            cdc_log.warn(
+                    "Failed to retrieve generation timestamps for rewriting: {}. Retrying in 60s.",
+                    std::current_exception());
+        }
+        co_await sleep_abortable(std::chrono::seconds(60), abort_src);
+    }
+}
+
+// Contains a CDC log table's creation time (extracted from its schema's id)
+// and its CDC TTL setting.
+struct time_and_ttl {
+    db_clock::time_point creation_time;
+    int ttl;
+};
+
+/*
+ * See `maybe_rewrite_streams_descriptions`.
+ * This is the long-running-in-the-background part of that function.
+ * It returns the timestamp of the last rewritten generation (if any).
+ */
+static future<std::optional<db_clock::time_point>> rewrite_streams_descriptions(
+        std::vector<time_and_ttl> times_and_ttls,
+        shared_ptr<db::system_distributed_keyspace> sys_dist_ks,
+        noncopyable_function<unsigned()> get_num_token_owners,
+        abort_source& abort_src) {
+    cdc_log.info("Retrieving generation timestamps for rewriting...");
+    auto tss = co_await get_cdc_desc_v1_timestamps(*sys_dist_ks, abort_src, get_num_token_owners);
+    cdc_log.info("Generation timestamps retrieved.");
+
+    // Find first generation timestamp such that some CDC log table may contain data before this timestamp.
+    // This predicate is monotonic w.r.t the timestamps.
+    auto now = db_clock::now();
+    std::sort(tss.begin(), tss.end());
+    auto first = std::partition_point(tss.begin(), tss.end(), [&] (db_clock::time_point ts) {
+        // partition_point finds first element that does *not* satisfy the predicate.
+        return std::none_of(times_and_ttls.begin(), times_and_ttls.end(),
+                [&] (const time_and_ttl& tat) {
+            // In this CDC log table there are no entries older than the table's creation time
+            // or (now - the table's ttl). We subtract 10s to account for some possible clock drift.
+            // If ttl is set to 0 then entries in this table never expire. In that case we look
+            // only at the table's creation time.
+            auto no_entries_older_than =
+                (tat.ttl == 0 ? tat.creation_time : std::max(tat.creation_time, now - std::chrono::seconds(tat.ttl)))
+                    - std::chrono::seconds(10);
+            return no_entries_older_than < ts;
+        });
+    });
+
+    // Find first generation timestamp such that some CDC log table may contain data in this generation.
+    // This and all later generations need to be written to the new streams table.
+    if (first != tss.begin()) {
+        --first;
+    }
+
+    if (first == tss.end()) {
+        cdc_log.info("No generations to rewrite.");
+        co_return std::nullopt;
+    }
+
+    cdc_log.info("First generation to rewrite: {}", *first);
+
+    bool each_success = true;
+    co_await max_concurrent_for_each(first, tss.end(), 10, [&] (db_clock::time_point ts) -> future<> {
+        while (true) {
+            try {
+                co_return co_await do_update_streams_description(ts, *sys_dist_ks, { get_num_token_owners() });
+            } catch (const no_generation_data_exception& e) {
+                cdc_log.error("Failed to rewrite streams for generation {}: {}. Giving up.", ts, e);
+                each_success = false;
+                co_return;
+            } catch (...) {
+                cdc_log.warn("Failed to rewrite streams for generation {}: {}. Retrying in 60s.", ts, std::current_exception());
+            }
+            co_await sleep_abortable(std::chrono::seconds(60), abort_src);
+        }
+    });
+
+    if (each_success) {
+        cdc_log.info("Rewriting stream tables finished successfully.");
+    } else {
+        cdc_log.info("Rewriting stream tables finished, but some generations could not be rewritten (check the logs).");
+    }
+
+    if (first != tss.end()) {
+        co_return *std::prev(tss.end());
+    }
+
+    co_return std::nullopt;
+}
+
+future<> maybe_rewrite_streams_descriptions(
+        const database& db,
+        shared_ptr<db::system_distributed_keyspace> sys_dist_ks,
+        noncopyable_function<unsigned()> get_num_token_owners,
+        abort_source& abort_src) {
+    if (!db.has_schema(sys_dist_ks->NAME, sys_dist_ks->CDC_DESC_V1)) {
+        // This cluster never went through a Scylla version which used this table
+        // or the user deleted the table. Nothing to do.
+        co_return;
+    }
+
+    if (co_await db::system_keyspace::cdc_is_rewritten()) {
+        co_return;
+    }
+
+    // For each CDC log table get the TTL setting (from CDC options) and the table's creation time
+    std::vector<time_and_ttl> times_and_ttls;
+    for (auto& [_, cf] : db.get_column_families()) {
+        auto& s = *cf->schema();
+        auto base = cdc::get_base_table(db, s.ks_name(), s.cf_name());
+        if (!base) {
+            // Not a CDC log table.
+            continue;
+        }
+        auto& cdc_opts = base->cdc_options();
+        if (!cdc_opts.enabled()) {
+            // This table is named like a CDC log table but it's not one.
+            continue;
+        }
+
+        times_and_ttls.push_back(time_and_ttl{as_timepoint(s.id()), cdc_opts.ttl()});
+    }
+
+    if (times_and_ttls.empty()) {
+        // There's no point in rewriting old generations' streams (they don't contain any data).
+        cdc_log.info("No CDC log tables present, not rewriting stream tables.");
+        co_return co_await db::system_keyspace::cdc_set_rewritten(std::nullopt);
+    }
+
+    // It's safe to discard this future: the coroutine keeps system_distributed_keyspace alive
+    // and the abort source's lifetime extends the lifetime of any other service.
+    (void)(([_times_and_ttls = std::move(times_and_ttls), _sys_dist_ks = std::move(sys_dist_ks),
+                _get_num_token_owners = std::move(get_num_token_owners), &_abort_src = abort_src] () mutable -> future<> {
+        auto times_and_ttls = std::move(_times_and_ttls);
+        auto sys_dist_ks = std::move(_sys_dist_ks);
+        auto get_num_token_owners = std::move(_get_num_token_owners);
+        auto& abort_src = _abort_src;
+
+        // This code is racing with node startup. At this point, we're most likely still waiting for gossip to settle
+        // and some nodes that are UP may still be marked as DOWN by us.
+        // Let's sleep a bit to increase the chance that the first attempt at rewriting succeeds (it's still ok if
+        // it doesn't - we'll retry - but it's nice if we succeed without any warnings).
+        co_await sleep_abortable(std::chrono::seconds(10), abort_src);
+
+        cdc_log.info("Rewriting stream tables in the background...");
+        auto last_rewritten = co_await rewrite_streams_descriptions(
+                std::move(times_and_ttls),
+                std::move(sys_dist_ks),
+                std::move(get_num_token_owners),
+                abort_src);
+
+        co_await db::system_keyspace::cdc_set_rewritten(last_rewritten);
+    })());
 }
 
 } // namespace cdc
--- a/cdc/generation.cc
+++ b/cdc/generation.cc
@@ -390,7 +390,7 @@ static future<> do_update_streams_description(
 
     auto topo = co_await sys_dist_ks.read_cdc_topology_description(streams_ts, ctx);
     if (!topo) {
-        throw std::runtime_error(format("could not find streams data for timestamp {}", streams_ts));
+        throw no_generation_data_exception(streams_ts);
     }
 
     co_await sys_dist_ks.create_cdc_desc(streams_ts, *topo, ctx);
--- a/cdc/generation.cc
+++ b/cdc/generation.cc
@@ -22,8 +22,9 @@
 #include <boost/type.hpp>
 #include <random>
 #include <unordered_set>
-#include <seastar/core/sleep.hh>
 #include <algorithm>
+#include <seastar/core/sleep.hh>
+#include <seastar/core/coroutine.hh>
 
 #include "keys.hh"
 #include "schema_builder.hh"
@@ -376,24 +377,23 @@ std::optional<db_clock::time_point> get_streams_timestamp_for(const gms::inet_ad
     return gms::versioned_value::cdc_streams_timestamp_from_string(streams_ts_string);
 }
 
-// Run inside seastar::async context.
-static void do_update_streams_description(
+static future<> do_update_streams_description(
         db_clock::time_point streams_ts,
         db::system_distributed_keyspace& sys_dist_ks,
         db::system_distributed_keyspace::context ctx) {
-    if (sys_dist_ks.cdc_desc_exists(streams_ts, ctx).get0()) {
+    if (co_await sys_dist_ks.cdc_desc_exists(streams_ts, ctx)) {
         cdc_log.info("Generation {}: streams description table already updated.", streams_ts);
-        return;
+        co_return;
     }
 
     // We might race with another node also inserting the description, but that's ok. It's an idempotent operation.
 
-    auto topo = sys_dist_ks.read_cdc_topology_description(streams_ts, ctx).get0();
+    auto topo = co_await sys_dist_ks.read_cdc_topology_description(streams_ts, ctx);
     if (!topo) {
         throw std::runtime_error(format("could not find streams data for timestamp {}", streams_ts));
     }
 
-    sys_dist_ks.create_cdc_desc(streams_ts, *topo, ctx).get();
+    co_await sys_dist_ks.create_cdc_desc(streams_ts, *topo, ctx);
     cdc_log.info("CDC description table successfully updated with generation {}.", streams_ts);
 }
 
@@ -403,7 +403,7 @@ void update_streams_description(
         noncopyable_function<unsigned()> get_num_token_owners,
         abort_source& abort_src) {
     try {
-        do_update_streams_description(streams_ts, *sys_dist_ks, { get_num_token_owners() });
+        do_update_streams_description(streams_ts, *sys_dist_ks, { get_num_token_owners() }).get();
     } catch(...) {
         cdc_log.warn(
             "Could not update CDC description table with generation {}: {}. Will retry in the background.",
@@ -416,7 +416,7 @@ void update_streams_description(
             while (true) {
                 sleep_abortable(std::chrono::seconds(60), abort_src).get();
                 try {
-                    do_update_streams_description(streams_ts, *sys_dist_ks, { get_num_token_owners() });
+                    do_update_streams_description(streams_ts, *sys_dist_ks, { get_num_token_owners() }).get();
                     return;
                 } catch (...) {
                     cdc_log.warn(
--- a/cdc/generation.hh
+++ b/cdc/generation.hh
@@ -131,6 +131,13 @@ public:
     {}
 };
 
+class no_generation_data_exception : public std::runtime_error {
+public:
+    no_generation_data_exception(db_clock::time_point generation_ts)
+        : std::runtime_error(format("could not find generation data for timestamp {}", generation_ts))
+    {}
+};
+
 /* Should be called when we're restarting and we noticed that we didn't save any streams timestamp in our local tables,
  * which means that we're probably upgrading from a non-CDC/old CDC version (another reason could be
  * that there's a bug, or the user messed with our local tables).
--- a/cdc/generation.hh
+++ b/cdc/generation.hh
@@ -199,4 +199,15 @@ void update_streams_description(
         noncopyable_function<unsigned()> get_num_token_owners,
         abort_source&);
 
+/* Part of the upgrade procedure. Useful in case where the version of Scylla that we're upgrading from
+ * used the "cdc_streams_descriptions" table. This procedure ensures that the new "cdc_streams_descriptions_v2"
+ * table contains streams of all generations that were present in the old table and may still contain data
+ * (i.e. there exist CDC log tables that may contain rows with partition keys being the stream IDs from
+ * these generations). */
+future<> maybe_rewrite_streams_descriptions(
+        const database&,
+        shared_ptr<db::system_distributed_keyspace>,
+        noncopyable_function<unsigned()> get_num_token_owners,
+        abort_source&);
+
 } // namespace cdc
--- a/cdc/generation.hh
+++ b/cdc/generation.hh
@@ -41,6 +41,7 @@
 #include "db_clock.hh"
 #include "dht/token.hh"
 #include "locator/token_metadata.hh"
+#include "utils/chunked_vector.hh"
 
 namespace seastar {
     class abort_source;
@@ -121,10 +122,10 @@ public:
  */ 
 class streams_version {
 public:
-    std::vector<stream_id> streams;
+    utils::chunked_vector<stream_id> streams;
     db_clock::time_point timestamp;
 
-    streams_version(std::vector<stream_id> s, db_clock::time_point ts)
+    streams_version(utils::chunked_vector<stream_id> s, db_clock::time_point ts)
         : streams(std::move(s))
         , timestamp(ts)
     {}
--- a/cdc/generation.hh
+++ b/cdc/generation.hh
@@ -123,12 +123,10 @@ class streams_version {
 public:
     std::vector<stream_id> streams;
     db_clock::time_point timestamp;
-    std::optional<db_clock::time_point> expired;
 
-    streams_version(std::vector<stream_id> s, db_clock::time_point ts, std::optional<db_clock::time_point> exp)
+    streams_version(std::vector<stream_id> s, db_clock::time_point ts)
         : streams(std::move(s))
         , timestamp(ts)
-        , expired(std::move(exp))
     {}
 };
 
--- a/cql3/query_processor.cc
+++ b/cql3/query_processor.cc
@@ -668,10 +668,14 @@ struct internal_query_state {
     bool more_results = true;
 };
 
-::shared_ptr<internal_query_state> query_processor::create_paged_state(const sstring& query_string,
-        const std::initializer_list<data_value>& values, int32_t page_size) {
+::shared_ptr<internal_query_state> query_processor::create_paged_state(
+        const sstring& query_string,
+        db::consistency_level cl,
+        const timeout_config& timeout_config,
+        const std::initializer_list<data_value>& values,
+        int32_t page_size) {
     auto p = prepare_internal(query_string);
-    auto opts = make_internal_options(p, values, db::consistency_level::ONE, infinite_timeout_config, page_size);
+    auto opts = make_internal_options(p, values, cl, timeout_config, page_size);
     ::shared_ptr<internal_query_state> res = ::make_shared<internal_query_state>(
             internal_query_state{
                     query_string,
@@ -935,17 +939,20 @@ bool query_processor::migration_subscriber::should_invalidate(
     return statement->depends_on_keyspace(ks_name) && (!cf_name || statement->depends_on_column_family(*cf_name));
 }
 
-future<> query_processor::query(
+future<> query_processor::query_internal(
         const sstring& query_string,
+        db::consistency_level cl,
+        const timeout_config& timeout_config,
         const std::initializer_list<data_value>& values,
+        int32_t page_size,
         noncopyable_function<future<stop_iteration>(const cql3::untyped_result_set_row&)>&& f) {
-    return for_each_cql_result(create_paged_state(query_string, values), std::move(f));
+    return for_each_cql_result(create_paged_state(query_string, cl, timeout_config, values, page_size), std::move(f));
 }
 
-future<> query_processor::query(
+future<> query_processor::query_internal(
         const sstring& query_string,
         noncopyable_function<future<stop_iteration>(const cql3::untyped_result_set_row&)>&& f) {
-    return for_each_cql_result(create_paged_state(query_string, {}), std::move(f));
+    return query_internal(query_string, db::consistency_level::ONE, infinite_timeout_config, {}, 1000, std::move(f));
 }
 
 }
--- a/db/system_distributed_keyspace.cc
+++ b/db/system_distributed_keyspace.cc
@@ -406,7 +406,7 @@ system_distributed_keyspace::cdc_get_versioned_streams(context ctx) {
 
         for (auto& row : *cql_result) {
             auto ts = row.get_as<db_clock::time_point>("time");
-            std::vector<cdc::stream_id> ids;
+            utils::chunked_vector<cdc::stream_id> ids;
             row.get_list_data<bytes>("streams", std::back_inserter(ids)); 
             result.emplace(ts, cdc::streams_version(std::move(ids), ts));
         }
--- a/db/system_distributed_keyspace.cc
+++ b/db/system_distributed_keyspace.cc
@@ -36,12 +36,13 @@
 #include <seastar/core/seastar.hh>
 #include <seastar/core/shared_ptr.hh>
 #include <seastar/core/coroutine.hh>
+#include <seastar/core/future-util.hh>
 
 #include <boost/range/adaptor/transformed.hpp>
 
 #include <optional>
 #include <vector>
-#include <optional>
+#include <set>
 
 extern logging::logger cdc_log;
 
@@ -92,12 +93,31 @@ schema_ptr cdc_generations() {
 /* A user-facing table providing identifiers of the streams used in CDC generations. */
 schema_ptr cdc_desc() {
     thread_local auto schema = [] {
-        auto id = generate_legacy_id(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_DESC);
-        return schema_builder(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_DESC, {id})
+        auto id = generate_legacy_id(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_DESC_V2);
+        return schema_builder(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_DESC_V2, {id})
                 /* The timestamp of this CDC generation. */
                 .with_column("time", timestamp_type, column_kind::partition_key)
-                /* The set of stream identifiers used in this CDC generation. */
+                /* For convenience, the list of stream IDs in this generation is split into token ranges
+                 * which the stream IDs were mapped to (by the partitioner) when the generation was created.  */
+                .with_column("range_end", long_type, column_kind::clustering_key)
+                /* The set of stream identifiers used in this CDC generation for the token range
+                 * ending on `range_end`. */
                 .with_column("streams", cdc_streams_set_type)
+                .with_version(system_keyspace::generate_schema_version(id))
+                .build();
+    }();
+    return schema;
+}
+
+/* A user-facing table providing CDC generation timestamps. */
+schema_ptr cdc_timestamps() {
+    thread_local auto schema = [] {
+        auto id = generate_legacy_id(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_TIMESTAMPS);
+        return schema_builder(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_TIMESTAMPS, {id})
+                /* This is a single-partition table. The partition key is always "timestamps". */
+                .with_column("key", utf8_type, column_kind::partition_key)
+                /* The timestamp of this CDC generation. */
+                .with_column("time", reversed_type_impl::get_instance(timestamp_type), column_kind::clustering_key)
                 /* Expiration time of this CDC generation (or null if not expired). */
                 .with_column("expired", timestamp_type)
                 .with_version(system_keyspace::generate_schema_version(id))
@@ -106,11 +126,14 @@ schema_ptr cdc_desc() {
     return schema;
 }
 
+static const sstring CDC_TIMESTAMPS_KEY = "timestamps";
+
 static std::vector<schema_ptr> all_tables() {
     return {
         view_build_status(),
         cdc_generations(),
         cdc_desc(),
+        cdc_timestamps(),
     };
 }
 
@@ -329,40 +352,70 @@ system_distributed_keyspace::expire_cdc_topology_description(
             false).discard_result();
 }
 
-static set_type_impl::native_type prepare_cdc_streams(const std::vector<cdc::stream_id>& streams) {
-    set_type_impl::native_type ret;
-    for (auto& s: streams) {
-        ret.push_back(data_value(s.to_bytes()));
-    }
-    return ret;
-}
-
-static future<mutation> get_cdc_streams_descriptions_mutation(
+static future<std::vector<mutation>> get_cdc_streams_descriptions_v2_mutation(
         const database& db,
         db_clock::time_point time,
-        const std::vector<cdc::stream_id>& streams) {
-    auto s = db.find_schema(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_DESC);
-    mutation m(s, partition_key::from_singular(*s, time));
-    m.set_cell(clustering_key::make_empty(), to_bytes("streams"),
-            make_set_value(cdc_streams_set_type, prepare_cdc_streams(streams)), api::new_timestamp());
-    co_return m;
+        const cdc::topology_description& desc) {
+    auto s = db.find_schema(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_DESC_V2);
+
+    auto ts = api::new_timestamp();
+    std::vector<mutation> res;
+    res.emplace_back(s, partition_key::from_singular(*s, time));
+    size_t size_estimate = 0;
+    for (auto& e : desc.entries()) {
+        // We want to keep each mutation below ~1 MB.
+        if (size_estimate >= 1000 * 1000) {
+            res.emplace_back(s, partition_key::from_singular(*s, time));
+            size_estimate = 0;
+        }
+
+        set_type_impl::native_type streams;
+        streams.reserve(e.streams.size());
+        for (auto& stream : e.streams) {
+            streams.push_back(data_value(stream.to_bytes()));
+        }
+
+        // We estimate 20 bytes per stream ID.
+        // Stream IDs themselves weigh 16 bytes each (2 * sizeof(int64_t))
+        // but there's metadata to be taken into account.
+        // It has been verified experimentally that 20 bytes per stream ID is a good estimate.
+        size_estimate += e.streams.size() * 20;
+        res.back().set_cell(clustering_key::from_singular(*s, dht::token::to_int64(e.token_range_end)),
+                to_bytes("streams"), make_set_value(cdc_streams_set_type, std::move(streams)), ts);
+
+        co_await make_ready_future<>(); // maybe yield
+    }
+
+    co_return res;
 }
 
 future<>
 system_distributed_keyspace::create_cdc_desc(
         db_clock::time_point time,
-        const std::vector<cdc::stream_id>& streams,
+        const cdc::topology_description& desc,
         context ctx) {
     using namespace std::chrono_literals;
-    auto m = co_await get_cdc_streams_descriptions_mutation(_qp.db(), time, streams);
-    co_return co_await _sp.mutate(
+
+    auto ms = co_await get_cdc_streams_descriptions_v2_mutation(_qp.db(), time, desc);
+    co_await max_concurrent_for_each(ms, 20, [&] (mutation& m) -> future<> {
+        // We use the storage_proxy::mutate API since CQL is not the best for handling large batches.
+        co_await _sp.mutate(
             { std::move(m) },
             quorum_if_many(ctx.num_token_owners),
             db::timeout_clock::now() + 10s,
             nullptr, // trace_state
             empty_service_permit(),
             false // raw_counters
-    );
+        );
+    });
+
+    // Commit the description.
+    co_await _qp.execute_internal(
+            format("INSERT INTO {}.{} (key, time) VALUES (?, ?)", NAME, CDC_TIMESTAMPS),
+            quorum_if_many(ctx.num_token_owners),
+            internal_distributed_timeout_config,
+            { CDC_TIMESTAMPS_KEY, time },
+            false).discard_result();
 }
 
 future<>
@@ -371,7 +424,7 @@ system_distributed_keyspace::expire_cdc_desc(
         db_clock::time_point expiration_time,
         context ctx) {
     return _qp.execute_internal(
-            format("UPDATE {}.{} SET expired = ? WHERE time = ?", NAME, CDC_DESC),
+            format("UPDATE {}.{} SET expired = ? WHERE time = ?", NAME, CDC_TIMESTAMPS),
             quorum_if_many(ctx.num_token_owners),
             internal_distributed_timeout_config,
             { expiration_time, streams_ts },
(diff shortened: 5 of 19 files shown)


kbr-:
I performed some experiments to compare solution that inserts the rows in batches vs one which doesn't.

Assumptions:
- ~20ms round-trip latency
- <= 100 write concurrency
- ~60MB of data (200 node cluster, 64 shards each)

Results:
- no batching: takes 10s to update the description table
- batching into ~1MB mutations: takes 1s

kbr-:
There's one thing that troubles me, which should probably be resolved before merging this...

Right now, it's quite easy for a client to check for new CDC generations using the streams table: scan the table, querying only the `time` column (which is a partition key column); then filter out timestamps that the client already knows. For any new timestamps, fetch the rest of the data (the `description` column) using a single-partition query.
With the new schema where the generation data may consist of a large number of clustering rows, something weird is happening. I tried the following query:
```
select time, ok from system_distributed.cdc_streams_descriptions_v2 per partition limit 1;
```
I query the `ok` static column to filter out generations that were not fully inserted. I use `per partition limit 1` because I don't need more than 1 row for each partition.

Unfortunately, this query results in many empty pages in-between different partitions (if the partitions are large). I'm guessing that's because when the coordinator contacts replicas, replicas must send entire partitions even if we use `per partition limit`; otherwise, we wouldn't be able to merge the mutations. But that means checking for new generations will be quite inefficient. I wonder if there's a simple way to fix this.

elcallio:
Am I missing something? Where does old stream ids get migrated? I mean, obviously the new table will contain equivalent data for the cluster as the last active generation, but what of the ones before? Maybe not a prio, but if you had a gen-switch less than 24h before updating nodes in cluster, will you not start having issues? Or do we push it to clients to also read the old table? (perhaps more sane)...

kbr-:
@elcallio your question is addressed at the end of #7993. AFAIU we want to backport this change to 4.3 as quickly as possible. So hopefully no-one will actually require any migration. But if they do, there are ways, e.g. an external tool that rewrites from old table to new, or have client read both tables as you said.

denesb:
> Unfortunately, this query results in many empty pages in-between different partitions (if the partitions are large). I'm guessing that's because when the coordinator contacts replicas, replicas must send entire partitions even if we use `per partition limit`; otherwise, we wouldn't be able to merge the mutations. But that means checking for new generations will be quite inefficient. I wonder if there's a simple way to fix this.

Per partition limit is implemented on the replica, what is happening is probably that there are many partitions that have 0 rows for this query, yet they will still be present in the results, which is what the empty pages you see are for.

kbr-:
> Per partition limit is implemented on the replica, what is happening is probably that there are many partitions that have 0 rows for this query, yet they will still be present in the results, which is what the empty pages you see are for.

But that can't be it, because in this test I did every partition had the same number of rows. And for this particular query every partition had exactly 1 row.

The schema was
```
CREATE TABLE system_distributed.cdc_streams_descriptions_v2 (
    time timestamp,
    vnode_end bigint,
    expired timestamp static,
    ok boolean static,
    streams set<blob>,
    PRIMARY KEY (time, vnode_end)
) ...
```

and the query was
```
select time, ok from system_distributed.cdc_streams_descriptions_v2 per partition limit 1;
```

denesb:
Try enabling tracing and see what is actually happening.

kbr-:
Well here's an example...
[cqlshout.txt](https://github.com/scylladb/scylla/files/5918542/cqlshout.txt)
(there were 3 pages; the file doesn't contain stdin)

I don't think I understand what's happening from this... why was the range `[{3997650101456510944, pk{0008000001775db25dd5}}, {4057785162919698611, end}]` queried 3 times?

denesb:
This looks like anormal paged range scan to me. Said range is not queried 3 times, it simply spans 2 pages.

denesb:
Unfortunately we don't have anything in tracing from the pager, so we don't know which pages are empty (and why).

kbr-:
v2:
- added a separate table for generation timestamps, as described in https://github.com/scylladb/scylla/issues/7993#issuecomment-772509689
- added a rewrite procedure, ensuring that old generations which are still needed get written into the new streams table, as described in https://github.com/scylladb/scylla/issues/7993#issuecomment-772500383

kbr-:
@tgrabiec could you take a quick look at this comment, maybe you have some ideas?

The PR adds a new distribtued table in `system_distributed`, and upgraded nodes immediately attempt to read from it. Schema propagation races with the read; the read is usually faster. I work around this by forcing the table to be synced on the replicas that are being read from by sending an empty mutation first... but maybe there's a better way?
```
    // Reading from this table on a freshly upgraded node that is the first to announce the CDC_TIMESTAMPS
    // schema would most likely result in replicas refusing to return data, telling the node that they can't
    // find the schema. Indeed, it takes some time for the nodes to synchronize their schema:
    // for some reason, `announce_new_column_family`, which is called on `system_distributed_keyspace` initialization,
    // doesn't reach the other nodes immediately (perhaps because of the moment it's called; the node isn't fully
    // started yet).
    //
    // This problem doesn't occur on writes since writes enforce schema pull if the receiving replica
    // notices that the write comes from an unknown schema, but it does occur on reads.
    //
    // Hence we work around it with a hack: we send a mutation with an empty partition to force our replicas
    // to pull the schema.
    //
    // This is not strictly necessary; the code that calls this function does it in a retry loop
    // so eventually, after the schema gets pulled, the read would succeed.
    // Still, the errors are also unnecessary and if we can get rid of them - let's do it.
    //
    // FIXME: find a more elegant way to deal with this ``problem''.
```

kbr-:
I need to perform some more manual testing, but in the meantime the change is ready to be reviewed.

kbr-:
Note: backporting this will require backporting 33ee07a9d8aa17a79a5d43037feb7200469a3f19 first

tgrabiec:
> @tgrabiec could you take a quick look at this comment, maybe you have some ideas?
> 
> The PR adds a new distribtued table in `system_distributed`, and upgraded nodes immediately attempt to read from it. Schema propagation races with the read; the read is usually faster. I work around this by forcing the table to be synced on the replicas that are being read from by sending an empty mutation first... but maybe there's a better way?
> 
> ```
>     // Reading from this table on a freshly upgraded node that is the first to announce the CDC_TIMESTAMPS
>     // schema would most likely result in replicas refusing to return data, telling the node that they can't
>     // find the schema. Indeed, it takes some time for the nodes to synchronize their schema:
>     // for some reason, `announce_new_column_family`, which is called on `system_distributed_keyspace` initialization,
>     // doesn't reach the other nodes immediately (perhaps because of the moment it's called; the node isn't fully
>     // started yet).
>     //
>     // This problem doesn't occur on writes since writes enforce schema pull if the receiving replica
>     // notices that the write comes from an unknown schema, but it does occur on reads.
>     //
>     // Hence we work around it with a hack: we send a mutation with an empty partition to force our replicas
>     // to pull the schema.
>     //
>     // This is not strictly necessary; the code that calls this function does it in a retry loop
>     // so eventually, after the schema gets pulled, the read would succeed.
>     // Still, the errors are also unnecessary and if we can get rid of them - let's do it.
>     //
>     // FIXME: find a more elegant way to deal with this ``problem''.
> ```

The proper fix for this is to pull the schema on reads as well. It's a one-liner change.

tgrabiec:
Posted "migration_manager: Propagate schema changes with reads like we do
on writes"

kbr-:
Thanks @tgrabiec.
Unfortunately, for now, I must leave this workaround in place: for the proper fix to work, the upgrade would have to be performed in two steps (first so all replicas apply the fix, then this PR). But I guess it will be possible to remove it after one major release.

db/system_distributed_keyspace.cc
```diff
@@ -117,9 +118,10 @@ bool system_distributed_keyspace::is_extra_durable(const sstring& cf_name) {
     return cf_name == CDC_TOPOLOGY_DESCRIPTION;
 }
 
-system_distributed_keyspace::system_distributed_keyspace(cql3::query_processor& qp, service::migration_manager& mm)
+system_distributed_keyspace::system_distributed_keyspace(cql3::query_processor& qp, service::migration_manager& mm, service::storage_proxy& sp)
         : _qp(qp)
```
haaawk, line 145: Is query_processor still needed when we're switching to lower level api?

db/system_distributed_keyspace.cc
```diff
@@ -335,17 +337,33 @@ static set_type_impl::native_type prepare_cdc_streams(const std::vector<cdc::str
...
+
 future<>
 system_distributed_keyspace::create_cdc_desc(
         db_clock::time_point time,
         const std::vector<cdc::stream_id>& streams,
         context ctx) {
-    return _qp.execute_internal(
-            format("INSERT INTO {}.{} (time, streams) VALUES (?,?)", NAME, CDC_DESC),
+    using namespace std::chrono_literals;
+    auto m = co_await get_mutation(_qp.db(), time, streams);
```
haaawk: I have to admit that my coroutine knowledge is quite poor but why do we need co_await for a call that never blocks?

db/system_distributed_keyspace.cc
```diff
@@ -335,17 +337,33 @@ static set_type_impl::native_type prepare_cdc_streams(const std::vector<cdc::str
...
+    co_return co_await _sp.mutate(
+            { std::move(m) },
             quorum_if_many(ctx.num_token_owners),
-            internal_distributed_timeout_config,
-            { time, make_set_value(cdc_streams_set_type, prepare_cdc_streams(streams)) },
-            false).discard_result();
+            db::timeout_clock::now() + 10s,
```
haaawk, line 408: why +10s?

db/system_distributed_keyspace.cc
```diff
@@ -407,10 +407,9 @@ system_distributed_keyspace::cdc_get_versioned_streams(context ctx) {
 
         for (auto& row : *cql_result) {
             auto ts = row.get_as<db_clock::time_point>("time");
-            auto exp = row.get_opt<db_clock::time_point>("expired");
```
haaawk, line 391: Should we remove "expired" from the schema as well? I might remember incorrectly but I think @elcallio had something to do with that field.

alternator/streams.cc
```diff
@@ -499,29 +499,19 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
     // TODO: creation time
 
     auto normal_token_owners = _proxy.get_token_metadata_ptr()->count_normal_token_owners();
-    // cannot really "resume" query, must iterate all data. because we cannot query neither "time" (pk) > something,
-    // or on expired...
-    // TODO: maybe add secondary index to topology table to enable this?
-    return _sdks.cdc_get_versioned_streams({ normal_token_owners }).then([this, &db, schema, shard_start, limit, ret = std::move(ret), stream_desc = std::move(stream_desc), ttl](std::map<db_clock::time_point, cdc::streams_version> topologies) mutable {
 
-        // filter out cdc generations older than the table or now() - cdc::ttl (typically dynamodb_streams_max_window - 24h)
-        auto low_ts = std::max(as_timepoint(schema->id()), db_clock::now() - ttl);
```
haaawk, line 508: I'm a bit confused. Was this code dead because of return in line 505?

db/system_distributed_keyspace.cc
```diff
@@ -96,8 +97,27 @@ schema_ptr cdc_desc() {
         return schema_builder(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_DESC, {id})
                 /* The timestamp of this CDC generation. */
                 .with_column("time", timestamp_type, column_kind::partition_key)
-                /* The set of stream identifiers used in this CDC generation. */
+                /* For convenience, the list of stream IDs in this generation is split into token ranges
```
haaawk, line 100: s/token ranges/vnodes?

db/system_distributed_keyspace.cc
```diff
@@ -96,8 +97,27 @@ schema_ptr cdc_desc() {
...
+                 * which the stream IDs were mapped to (by the partitioner) when the generation was created.  */
+                .with_column("range_end", long_type, column_kind::clustering_key)
```
haaawk, line 102: s/range_end/vnode_end?

db/system_distributed_keyspace.cc
```diff
@@ -329,41 +352,73 @@ system_distributed_keyspace::expire_cdc_topology_description(
...
+
+    auto ts = api::new_timestamp();
+    std::vector<mutation> res;
+    res.emplace_back(s, partition_key::from_singular(*s, time));
+    size_t size_estimate = 0;
+    for (auto& e : desc.entries()) {
+        // We want to keep each mutation below ~1 MB.
+        if (size_estimate >= 1000 * 1000) {
+            res.emplace_back(s, partition_key::from_singular(*s, time));
+            size_estimate = 0;
```
haaawk, line 370: Wouldn't it be better to do:
```
while (size_estimate > 1000 * 1000) {
  res.emplace_back(s, partition_key::from_singular(*s, time));
  size_estimate -= 1000 * 1000;
}
```

?

db/system_distributed_keyspace.cc
```diff
@@ -329,41 +352,73 @@ system_distributed_keyspace::expire_cdc_topology_description(
...
+
+        // We estimate 20 bytes per stream ID.
+        // Stream IDs themselves weigh 16 bytes each (2 * sizeof(int64_t))
+        // but there's metadata to be taken into account.
+        // It has been verified experimentally that 20 bytes per stream ID is a good estimate.
+        size_estimate += e.streams.size() * 20;
+        res.back().set_cell(clustering_key::from_singular(*s, dht::token::to_int64(e.token_range_end)),
+                to_bytes("streams"), make_set_value(cdc_streams_set_type, std::move(streams)), ts);
+
+        co_await make_ready_future<>(); // maybe yield
```
haaawk, line 387: Isn't yielding at every vector element too much? Wouldn't it be better to yield at every 100th element or sth?

db/system_distributed_keyspace.cc
```diff
@@ -383,37 +438,94 @@ future<bool>
...
+    for (auto& row : *timestamps_cql) {
+        timestamps.push_back(row.get_as<db_clock::time_point>("time"));
+    }
+
+    // `time` is the table's clustering key, so the results are already sorted
+    auto first = std::lower_bound(timestamps.begin(), timestamps.end(), not_older_than);
+    // need first gen _intersecting_ the timestamp.
+    if (first != timestamps.begin()) {
+        --first;
+    }
```
haaawk, line 507: Since timestamps are clustering keys in the table wouldn't it be possible to just use WHERE time >= not_older_than in the CQL above?

db/system_distributed_keyspace.cc
```diff
@@ -383,37 +438,94 @@ future<bool>
...
+    co_await parallel_for_each(first, timestamps.end(), [this, &ctx, &result] (db_clock::time_point ts) -> future<> {
+        auto streams_cql = co_await _qp.execute_internal(
+                format("SELECT streams FROM {}.{} WHERE time = ?", NAME, CDC_DESC),
+                quorum_if_many(ctx.num_token_owners),
+                internal_distributed_timeout_config,
+                { ts },
+                false);
+
+        // FIXME: get rid of large allocations
+        std::vector<cdc::stream_id> ids;
```
haaawk, line 519: ids.reserve(streams_cql->size())?

db/system_distributed_keyspace.cc
```diff
@@ -383,37 +438,94 @@ future<bool>
...
+        for (auto& row : *streams_cql) {
+            row.get_list_data<bytes>("streams", std::back_inserter(ids));
+            co_await make_ready_future<>(); // maybe yield
```
haaawk, line 522: Maybe not yield at every element?

haaawk (commented):
LGTM

cdc/generation.cc
```diff
@@ -374,4 +376,130 @@ void update_streams_description(
     }
 }
 
+static db_clock::time_point as_timepoint(const utils::UUID& uuid) {
```
haaawk, line 379: Would it make sense to make this a member function of UUID class?

cdc/generation.cc
```diff
@@ -374,4 +376,130 @@ void update_streams_description(
...
+    return db_clock::time_point{std::chrono::milliseconds(utils::UUID_gen::get_adjusted_timestamp(uuid))};
+}
+
+static future<std::vector<db_clock::time_point>> get_cdc_desc_v1_timestamps(
```
haaawk, line 383: This function seems familiar. Is it a copy of another place that does the same thing?

cdc/generation.cc
```diff
@@ -374,4 +376,130 @@ void update_streams_description(
...
+        // or the user deleted the table. Nothing to do.
+        co_return;
+    }
+
+    if (co_await db::system_keyspace::cdc_is_rewritten()) {
+        co_return;
+    }
+
+    // For each CDC log table get the TTL setting (from CDC options) and the table's creation time
+    std::vector<std::pair<db_clock::time_point, int>> times_and_ttls;
```
haaawk, line 467: It would be more readable if we had a dedicated type instead of a pair of unnamed elements.

cdc/generation.cc
```diff
@@ -374,4 +376,130 @@ void update_streams_description(
...
+        } catch (...) {
+            cdc_log.warn(
+                    "Failed to retrieve generation timestamps for rewriting: {}. Retrying in 60s.",
+                    std::current_exception());
+        }
+        co_await sleep_abortable(std::chrono::seconds(60), abort_src);
+    }
+}
+
+static future<std::optional<db_clock::time_point>> rewrite_streams_descriptions(
```
haaawk, line 399: A comment saying what's the returned value of this function would be appreciated.

cdc/generation.cc
```diff
@@ -374,4 +376,130 @@ void update_streams_description(
...
+            continue;
+        }
+        auto& cdc_opts = base->cdc_options();
+        if (!cdc_opts.enabled()) {
+            // This table is named like a CDC log table but it's not one.
+            continue;
+        }
+
+        times_and_ttls.emplace_back(as_timepoint(s.id()), cdc_opts.ttl());
+    }
```
haaawk, line 482: For a cluster with large number of tables this can cause reactor stalls.

cdc/generation.cc
```diff
@@ -374,4 +376,130 @@ void update_streams_description(
...
+            // Not a CDC log table.
+            continue;
+        }
+        auto& cdc_opts = base->cdc_options();
+        if (!cdc_opts.enabled()) {
+            // This table is named like a CDC log table but it's not one.
+            continue;
+        }
+
+        times_and_ttls.emplace_back(as_timepoint(s.id()), cdc_opts.ttl());
```
haaawk, line 481: Would it make sense to do here the filtering that's currently done in `rewrite_streams_descriptions `? Number of CDC table is expected to be relatively low and we're iterating through them anyway. It seems that collecting all CDC tables in a vector and then removing them makes little sense when we can check before addition, no?

db/system_keyspace.cc
```diff
@@ -1574,6 +1574,19 @@ future<> update_cdc_streams_timestamp(db_clock::time_point tp) {
...
+future<> cdc_set_rewritten(std::optional<db_clock::time_point> tp) {
+    if (tp) {
+        return qctx->execute_cql(
+                format("INSERT INTO system.{} (key, streams_timestamp) VALUES (?, ?)", v3::CDC_LOCAL),
+                sstring(v3::CDC_LOCAL), *tp).discard_result();
+    } else {
+        // Insert just the row marker.
+        return qctx->execute_cql(
+                format("INSERT INTO system.{} (key) VALUES (?)", v3::CDC_LOCAL),
+                sstring(v3::CDC_LOCAL)).discard_result();
```
haaawk, line 1586: Why do we have to insert anything in this case?

haaawk: Looking at `cdc_is_rewritten` it seems that not inserting anything and inserting just row marker makes no difference, no?

service/storage_service.cc
```diff
@@ -749,7 +749,8 @@ void storage_service::handle_cdc_generation(std::optional<db_clock::time_point>
         return;
     }
 
-    if (!db::system_keyspace::bootstrap_complete() || !_sys_dist_ks.local_is_initialized()) {
+    if (!db::system_keyspace::bootstrap_complete() || !_sys_dist_ks.local_is_initialized()
+            || !_sys_dist_ks.local().started()) {
         // We still haven't finished the startup process.
```
avikivity, line 762: What if it's locally started but on some remote shard still in progress?

Really, these dependencies should be expressed statically rather than computed dynamically. Of course that's beyond the scope of this series.

db/system_distributed_keyspace.cc
```diff
@@ -335,17 +337,33 @@ static set_type_impl::native_type prepare_cdc_streams(const std::vector<cdc::str
...
+static future<mutation> get_mutation(
+        const database& db,
+        db_clock::time_point time,
+        const std::vector<cdc::stream_id>& streams) {
+    auto s = db.find_schema(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_DESC);
+    mutation m(s, partition_key::from_singular(*s, time));
+    m.set_cell(clustering_key::make_empty(), to_bytes("streams"),
+            make_set_value(cdc_streams_set_type, prepare_cdc_streams(streams)), api::new_timestamp());
+    co_return m;
+}
```
avikivity, line 391: Why is this a coroutine? It looks like it's a single continuation.

db/system_distributed_keyspace.cc
avikivity: Removing things from schemas requires extreme care.

alternator/streams.cc
```diff
@@ -499,29 +499,19 @@ future<executor::request_return_type> executor::describe_stream(client_state& cl
...
+    // filter out cdc generations older than the table or now() - cdc::ttl (typically dynamodb_streams_max_window - 24h)
+    auto low_ts = std::max(as_timepoint(schema->id()), db_clock::now() - ttl);
 
-        auto i = topologies.lower_bound(low_ts);
-        // need first gen _intersecting_ the timestamp.
-        if (i != topologies.begin()) {
-            i = std::prev(i);
-        }
+    return _sdks.cdc_get_versioned_streams(low_ts, { normal_token_owners }).then([this, &db, shard_start, limit, ret = std::move(ret), stream_desc = std::move(stream_desc)] (std::map<db_clock::time_point, cdc::streams_version> topologies) mutable {
```
avikivity, line 506: Why does a change to the internals of cdc require a change to alternator streams? I guess the reason is that there was no abstraction? But it doesn't query the table directly.

db/system_distributed_keyspace.cc
avikivity: Yielding (in this way) does nothing if not needed by the scheduler.

db/system_distributed_keyspace.cc
avikivity: Ok, in a following patch it is a coroutine.

db/system_distributed_keyspace.cc
```diff
@@ -96,8 +97,27 @@ schema_ptr cdc_desc() {
...
+
+/* A user-facing table providing CDC generation timestamps. */
+schema_ptr cdc_timestamps() {
+    thread_local auto schema = [] {
+        auto id = generate_legacy_id(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_TIMESTAMPS);
+        return schema_builder(system_distributed_keyspace::NAME, system_distributed_keyspace::CDC_TIMESTAMPS, {id})
+                /* This is a single-partition table. The partition key is always "timestamps". */
+                .with_column("key", utf8_type, column_kind::partition_key)
+                /* The timestamp of this CDC generation. */
+                .with_column("time", timestamp_type, column_kind::clustering_key)
```
avikivity, line 120: Sort it descending? So you don't have to scan everything if you have some of the rows already?

db/system_distributed_keyspace.cc
```diff
@@ -96,8 +97,27 @@ schema_ptr cdc_desc() {
...
                 .with_column("time", timestamp_type, column_kind::partition_key)
-                /* The set of stream identifiers used in this CDC generation. */
+                /* For convenience, the list of stream IDs in this generation is split into token ranges
+                 * which the stream IDs were mapped to (by the partitioner) when the generation was created.  */
+                .with_column("range_end", long_type, column_kind::clustering_key)
+                /* The set of stream identifiers used in this CDC generation for the token range
+                 * ending on `range_end`. */
                 .with_column("streams", cdc_streams_set_type)
+                .with_version(system_keyspace::generate_schema_version(id))
+                .build();
```
avikivity, line 107: It looks like you're changing the schema of CDC_DESC. That can't be right.

If you changed the contents of system_distributed_keyspace::CDC_DESC, that's just confusing.

db/system_distributed_keyspace.cc
```diff
@@ -329,41 +352,73 @@ system_distributed_keyspace::expire_cdc_topology_description(
...
+        const cdc::topology_description& desc,
         context ctx) {
     using namespace std::chrono_literals;
-    auto m = co_await get_mutation(_qp.db(), time, streams);
-    co_return co_await _sp.mutate(
+
+    auto ms = co_await get_mutations(_qp.db(), time, desc);
+    // The number of mutations will be around 30 in a 100-node cluster with 64 shards per node,
+    // so in practice limiting concurrency here is not necessary, but just to be safe...
+    co_await max_concurrent_for_each(ms, 100, [&] (mutation& m) -> future<> {
```
avikivity, line 403: This requires 100MB just for this operation. It can stress a smaller shard.

I'd limit it to 20MB.

db/system_distributed_keyspace.cc
```diff
@@ -383,37 +438,94 @@ future<bool>
...
+
+    // `time` is the table's clustering key, so the results are already sorted
+    auto first = std::lower_bound(timestamps.begin(), timestamps.end(), not_older_than);
+    // need first gen _intersecting_ the timestamp.
+    if (first != timestamps.begin()) {
+        --first;
+    }
+
+    std::map<db_clock::time_point, cdc::streams_version> result;
+    co_await parallel_for_each(first, timestamps.end(), [this, &ctx, &result] (db_clock::time_point ts) -> future<> {
```
avikivity, line 510: What is the expected concurrency here?

db/system_distributed_keyspace.cc
avikivity: utils::chunked_vector will help the FIXME

db/system_distributed_keyspace.hh
```diff
@@ -55,8 +55,11 @@ public:
     /* Nodes use this table to communicate new CDC stream generations to other nodes. */
     static constexpr auto CDC_TOPOLOGY_DESCRIPTION = "cdc_generation_descriptions";
 
-    /* This table is used by CDC clients to learn about avaliable CDC streams. */
-    static constexpr auto CDC_DESC = "cdc_streams_descriptions";
+    /* This table is used by CDC clients to learn about available CDC streams. */
+    static constexpr auto CDC_DESC = "cdc_streams_descriptions_v2";
+
```
avikivity, line 60: That can be super confusing for someone debugging the code and looking at the wrong version. Not to mention upgrade support.

cdc/generation.cc
```diff
@@ -322,23 +323,23 @@ std::optional<db_clock::time_point> get_streams_timestamp_for(const gms::inet_ad
 }
 
 // Run inside seastar::async context.
-static void do_update_streams_description(
```
avikivity, line 325: Comment now wrong.

avikivity:
Looks reasonable, please add something to docs/ that explains the old schema and the rewriting process (since the old schema no longer appears in code).

My biggest concern is how this is tested.

db/system_distributed_keyspace.cc
kbr-: Not all functions switch to the lower level api

db/system_distributed_keyspace.cc
kbr-: `get_mutation` returns a `future`, we want to wait on this future. We don't want to discard it.
Perhaps I don't fully understand the question

db/system_distributed_keyspace.cc
kbr-: Everything else in `system_distributed_keyspace` uses `10s`

db/system_distributed_keyspace.cc
kbr-: Dunno, these ranges are vnodes when they are created - true, but after a topology change - not necessary (for old generations)

db/system_distributed_keyspace.cc
kbr-: see above

db/system_distributed_keyspace.cc
kbr-: that would only find generations which started after `not_older_than`, but we want all generations which operated during `not_older_than` orlater. So one more generation potentially compared to what `time >= not_older_than` would return

db/system_distributed_keyspace.cc
kbr-: it doesn't matter

cdc/generation.cc
kbr-: No, it's original, the one and only.
What do you mean?

cdc/generation.cc
kbr-: Don't the comments and variable names make it obvious already?
I could do
```
using time_and_ttl = std::pair<...>;
std::vector<time_and_ttl> times_and_ttls;
```
but I don't really see the point

cdc/generation.cc
kbr-: How many tables can a cluster really have? 1k? 10k? 100k?

cdc/generation.cc
kbr-: To do the filtering we need to get all generation timestamps from the old table, and that's a potentially heavy / long operation, hence I moved it into the background.

db/system_keyspace.cc
kbr-:
```
     // We don't care about the actual timestamp; it's additional information for debugging purposes. 
```

kbr-: so you're right, we don't need the timestamp - but maybe it'll turn out to be useful

service/storage_service.cc
kbr-: We only need the local instance (shard 0).
Btw. the other instances don't do anything anyway (see `system_distributed_keyspace::start`, the first `if`)

db/system_distributed_keyspace.cc
haaawk: Ok.

db/system_distributed_keyspace.cc
haaawk: but get_mutation never blocks so there's no need to wait. The correct answer is that in the following commits you make get_mutation do some calls that could block.

db/system_distributed_keyspace.cc
haaawk: Ok. What's the meaning of it? Timeout?

db/system_distributed_keyspace.cc
haaawk: Fair enough - it is true for old generations.

db/system_distributed_keyspace.cc
haaawk: I see. Makes sense.

db/system_distributed_keyspace.cc
haaawk: You care to explain why it does not matter?

cdc/generation.cc
haaawk: I must have seen thing retry loop in other contexts. Maybe it's worth having a util.

alternator/streams.cc
kbr-: This change is not strictly required, but the new implementation of the description table allows much more efficient implementation of this part in alternator streams code.
Previously this code would fetch **all** generations (lots of data) and then filter them out according to timestamps.
Now we fetch timestamps first with low cost using the new timestamps table, filter the timestamps out, and then query only the generations whose timestamps are left

cdc/generation.cc
haaawk: Comment doesn't follow the object. o.ttl is much more readable than o.second.
I meant a proper object not an alias.

db/system_distributed_keyspace.cc
kbr-: What additional utility does sorting it descending give? One can just query `time > X` if they want generations only newer than `X` (that's what clients want to know usually)

db/system_distributed_keyspace.cc
kbr-: You mean I should also change the variable name to e.g. `CDC_DESC_V2`?

cdc/generation.cc
haaawk: I think we had users with 10k tables before.
This should already be enough to cause problems.
We had problems related to the number of tables in the past.

db/system_keyspace.cc
haaawk: I meant why do we need `else` here? The behaviour of `cdc_is_written` is the same no matter if we insert the marker or not.

db/system_distributed_keyspace.cc
kbr-: It can get quite large, I'll limit it

cdc/generation.cc
haaawk: ok.

db/system_distributed_keyspace.cc
kbr-: So the correct question is "why `get_mutation` returns a future"
which you answered correctly indeed.

db/system_distributed_keyspace.cc
kbr-: Yes. Previously it was expressed using `internal_distributed_timeout_config`

db/system_distributed_keyspace.cc
kbr-: As Avi explained, yielding in this way does nothing if not needed by the scheduler.
Even if it did, this is not on any hot path, and it's a potentially long operation anyway
also, the rows can be quite "large" too (with 64 shards they will be ~1KB)

cdc/generation.cc
kbr-: Maybe, maybe not necessarily
I'd say it's simple enough that an abstraction would only introduce unnecessary complexity with disputable gains (code deduplication is not a goal on its own)

cdc/generation.cc
kbr-: Processing a mutation fragment by `combined_reader` - which is a far more complex operation than what we're doing here - takes 100ns.
So processing 10k such fragments would take 100ns *10k = 1ms.
Now consider processing 10k tables by this loop, it'll take significantly less than 1ms.

I don't think this loop will cause reactor stalls.
Even if it did, they would be short, and this operation happens once per node, so I don't think it matters.

db/system_keyspace.cc
kbr-: Ah, I see.

Not inserting anything vs inserting just the row maker **does make a difference**.
If we don't insert anything, `SELECT key FROM system.cdc_local WHERE key = rewritten` will return 0 rows
if we insert the row marker, it will return 1 row

cdc/generation.cc
haaawk: ok.

db/system_keyspace.cc
haaawk: But for `cdc_is_written ` it makes no difference whether there's no row or 1 row with no key column -> `!msg->empty() && msg->one().has("key")`. Is there another user of this table?

kbr-: Well, first of all, `set_rewritten` has a bug, it should use the `rewritten` key, not the `cdc_local` key... So the current code would rewrite on each restart, and it would mess up the `system.cdc_local` table by setting the `cdc_local` key (which contains the current generation timestamp) to a wrong value potentially.

Now, after I fix the bug, then:
- if there's a row with `key = rewritten` containing just the row marker, then `!msg->empty() && msg->one().has("key")` will return **true** - because `key` is the partition key, so the row will show up like this:
```
cqlsh> select * from system.cdc_local ;

 key       | streams_timestamp
-----------+---------------------------------
 cdc_local | 2021-02-04 17:56:34.119000+0000
 rewritten |                            null
```
- if there is no such row, then the check will be `false`.

kbr-: There is no such thing as "row with no `key` column" because `key` is the partition key

haaawk: > There is no such thing as "row with no `key` column" because `key` is the partition key

Why do we check `msg->one().has("key")` then?

cdc/generation.cc
kbr-: Actually no, it's too concrete to make it a generic method of that class
(it refers to `db_clock`, which UUID knows nothing of)

haaawk: It just returns db_clock::time_point so in the include it will be enough to declare it. It seems that db_clock is a fundamental class in our database so I don't think it is a problem that it's concrete. 

I can see that alternator/streams.cc has exactly the same function. Have you copy pasted it?

kbr-:
v3:
- use `chunked_vector` for `cdc::streams_version::streams`
- use `time_and_ttl` struct instead of `pair<time_point, int>` in (`maybe_`)`rewrite_streams_descriptions`
- comment `rewrite_streams_descriptions`
- fix `db::system_keyspace::cdc_set_rewritten`: it used the wrong key (`"cdc_local"`, when it should use `"rewritten"`)
- rename `CDC_DESC` to `CDC_DESC_V2`
- update docs/design-notes/cdc.md to describe the new tables
- reduce concurrency limit in `create_cdc_desc` to 20
- add a missing note about `system.cdc_local` in cdc.md
- describe the upgrading procedure in cdc.md
- limit concurrency in `cdc_get_versioned_streams`
- call `db::system_keyspace::cdc_set_rewritten` on a bootstrapping node. We only want upgrading nodes to perform rewriting, not newly joining ones.
- remove outdated comment above `do_update_streams_description`
- added a little sleep at the start of the rewriting background task. At this point, we're most likely still waiting for gossip to settle and some nodes that are UP may still be marked as DOWN by us. We sleep to increase the chance that the first attempt at rewriting succeeds (it's still ok if it doesn't - we'll retry - but it's nice if we succeed without any warnings).

See diff: https://github.com/scylladb/scylla/compare/421f120eaa5dbde5c7dee4853b5e646143a4f32b..872b7bb068ad09fc1ceb92ff97f6b45648080412

> My biggest concern is how this is tested.

The upgrade/rewrite procedure I'm testing manually. As for the new tables / new schema, there  are some dtests that test the stream description table which I must update; also @haaawk said he'll update the Java CDC library / replicator so I can run the SCT replication test with the new schema.

cdc/generation.cc
kbr-: > It just returns db_clock::time_point so in the include it will be enough to declare it. It seems that db_clock is a fundamental class in our database so I don't think it is a problem that it's concrete.

Yeah but it's a UUID library that doesn't know about a database. I'd like to keep it clean

> I can see that alternator/streams.cc has exactly the same function. Have you copy pasted it?

Yes, I was hoping you won't notice

db/system_keyspace.cc
kbr-: You're right, it's not needed. But it also doesn't hurt. So I left it :P

kbr-:
v4:
- added a simple CQL pytest for testing the streams description table.

kbr-:
I made the following experiment on a single-node cluster:
1. Insert 50 generations into `cdc_streams_descriptions`, each of size ~6MB (20 * 256 * 64 stream IDs)
2. Perform `select time from system_distributed.cdc_streams_descriptions ;` with paging off

Result:
```
storage_proxy - Exception when communicating with 127.0.0.1, to read from system_distributed.cdc_streams_descriptions: std::runtime_error (Memory usage of unpaged query exceeds hard limit of 104857600 (configured via max_memory_for_unlimited_query_hard_limit))
```

It looks like the query must be paged... but I wonder if that will be enough...

db/system_distributed_keyspace.cc
avikivity: Well, time > X can be slightly more efficient with descending sort, though in many cases it will be the same.

If there is no promoted index, then time > X with descending time is just reading rows from the beginning of the partition. With ascending time, you read and discard rows until you reach time > X.

If there is a promoted index, we'll seek it in both cases.

db/system_distributed_keyspace.cc
avikivity: Yes.

avikivity:
> I made the following experiment on a single-node cluster:
> 
> 1. Insert 50 generations into `cdc_streams_descriptions`, each of size ~6MB (20 * 256 * 64 stream IDs)
> 2. Perform `select time from system_distributed.cdc_streams_descriptions ;` with paging off
> 
> Result:
> 
> ```
> storage_proxy - Exception when communicating with 127.0.0.1, to read from system_distributed.cdc_streams_descriptions: std::runtime_error (Memory usage of unpaged query exceeds hard limit of 104857600 (configured via max_memory_for_unlimited_query_hard_limit))
> ```
> 
> This means that the current rewriting code won't work if there are sufficiently many generations to rewrite.
> It looks like the query must be paged... but I wonder if that will be enough...

Paging will not work since it's just one row. But note that system queries are unlimited (except by the amount of memory in the shard).

kbr-:
> Paging will not work since it's just one row. But note that system queries are unlimited (except by the amount of memory in the shard).

I'm doing a full range scan here to retrieve the timestamps of all generations.
This experiment was performed using `cqlsh`. With `paging off`, this is the error I got. With `paging on`, this error didn't appear; I had each timestamp in a separate page.

And yeah, `execute_internal` managed to perform that query, although with a huge amount of allocation warnings etc. comming from `scanning_and_populating_reader`.
I'll try to find or write a better API for performing internal queries in a paged fashion.

kbr-:
v5:
- if we detect that a generation is not present in the internal table, we don't rewrite it (we need the internal info to do the rewrite). That's why the `no_generation_data_exception` type was introduced, which we catch in the rewrite process and act accordingly
- additional logging to better monitor the process
- after retrieving timestamps, check if there are any generations to rewrite; if there are none, finish. Previously the code assumed that there's always at least one generation, and would enter undefined behavior otherwise. Actually, there always should be at least one generation on rewrite, but the check does not hurt and may help if we forgot some edge case
- add a config option to disable streams rewriting. The rewriting process is potentially long, expensive, and prone to failure. Due to #8061 it may consume a lot of memory. It may keep failing endlessly (and retrying), straining the cluster. This flag is added as a backdoor for potential future use by admins / field engineers. I don't expect it to ever be used, but it won't hurt to have it.
- fix use-after-free of `abort_src`; the variable was a lambda capture, and the lambda was a coroutine; the captured variable gets destroyed after the first yield. The code now makes a copy of the variable into the coroutine
- improve API for querying tables internally with paging using `query_processor`. The previous API was not powerful enough for the needs of this PR. See the relevant commit for details
- use a paged query to retrieve generation timestamps from the old streams table, using the improved API
- increase the timeout for the query to 60s (even on a single-node cluster, with 50 large generations, the query took 3s due to #8061)
- limit permissions for modifying the new timestamps table, update the cdc-metadata-tables permission test to accomodate for the new schema and the new timestamps table

I also did a number of experiments / tests with the upgrade procedure, e.g. to check how the rewrite process will handle a large number of large generations. With the paged query it looks like it works (no memory limits are hit), but it does take some time.

In any case, I think the PR is ready for re-review.

kbr-:
@elcallio could you review the part with `cdc_get_versioned_streams` and inside alternator? Unless of course you want to review the whole PR, which would be even better :P

cdc/generation.cc
elcallio: Doesn't the "move UUID to use std::chrono" series sort of invite a templated variant of this that can be UUID-local?

elcallio: Seems like a thing outside scope of this PR though.

elcallio (approved):
No real issues on my end. lgtm.

cdc/generation.hh
```diff
@@ -124,12 +124,10 @@ class streams_version {
...
     std::vector<stream_id> streams;
     db_clock::time_point timestamp;
-    std::optional<db_clock::time_point> expired;
 
-    streams_version(std::vector<stream_id> s, db_clock::time_point ts, std::optional<db_clock::time_point> exp)
+    streams_version(std::vector<stream_id> s, db_clock::time_point ts)
         : streams(std::move(s))
         , timestamp(ts)
-        , expired(std::move(exp))
     {}
```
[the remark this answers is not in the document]
kbr-: Out of laziness, it made the later commits easier to implement.
Also the alternator/streams code uses an `expired` variable which is defined using generation timestamps and has no direct relationship to this struct's expired field, so removing it also removes some confusion imo.

cdc/generation.cc
kbr-: Yeah, maybe some kind of template would do (parameterized using `ratio` and maybe something else). But meh, I'll let someone else do this :P

kbr-:
@haaawk @avikivity ping

db/system_distributed_keyspace.hh
```diff
@@ -73,6 +75,8 @@ public:
     future<> start();
     future<> stop();
 
+    bool started() { return _started; }
```
haaawk, line 92: bool started() const

db/system_distributed_keyspace.cc
```diff
@@ -335,17 +337,33 @@ static set_type_impl::native_type prepare_cdc_streams(const std::vector<cdc::str
     return ret;
 }
 
+// cdc_streams_descriptions mutation
```
haaawk: Maybe name the function get_cdc_streams_descriptions_mutation instead of adding a comment?

db/system_distributed_keyspace.cc
haaawk: with descending order we can fetch the latest generation with LIMIT 1

cql3/query_processor.cc
```diff
@@ -668,10 +668,14 @@ struct internal_query_state {
     bool more_results = true;
```
haaawk, line 668: All those 4 listed changes should be in separate commit each.

kbr-: I don't think there's any value in doing that (I'm for splitting into small commits in general, but in that particular case it would do more harm than good IMO)

docs/design-notes/cdc.md
```diff
@@ -211,5 +211,40 @@ When nodes learn about a CDC generation through gossip, they race to update thes
...
+6. For each timestamp that's left:
+6.1 if it's already present in the new table, skip it (we check this by querying `cdc_generation_timestamps`
+6.2 fetch the generation (by querying `cdc_generation_descriptions`)
+6.3 insert the generation's streams into the new table
+7. Insert a row with `key = "rewritten"` into `system.cdc_local`.
+
+Note that every node will perform this procedure on upgrade, but there's a high chance that only one of them actually proceeds all the way to step 6.2 if upgrade is performed correctly, i.e. in a rolling fashion (nodes are restarted one-by-one).
+
+In order to prevent new nodes to do the rewriting (we only want upgrading nodes to do it), we insert the `key = "rewritten"` row on bootstrap as well, before we start this procedure (so the node won't pass the second check).
```
haaawk, line 247: @kbr- do I understand correctly that the addition of even a single new node will cause rewrite to be abandonned/stopped?

db/system_keyspace.cc
```diff
@@ -1646,6 +1661,14 @@ future<std::optional<db_clock::time_point>> get_saved_cdc_streams_timestamp() {
     });
 }
 
+future<bool> cdc_is_rewritten() {
+    // We don't care about the actual timestamp; it's additional information for debugging purposes.
```
haaawk, line 1665: Is this comment still valid?

docs/design-notes/cdc.md
kbr-: It means that the new node won't attempt the rewrite (assuming it's bootstrapped with the new version).
Only nodes which upgrade from pre-this-commit to post-this-commit attempt rewriting.

db/system_keyspace.cc
```diff
@@ -1646,6 +1661,14 @@ future<std::optional<db_clock::time_point>> get_saved_cdc_streams_timestamp() {
...
+    return qctx->execute_cql(format("SELECT key FROM system.{} WHERE key = ?", v3::CDC_LOCAL), "rewritten")
+            .then([] (::shared_ptr<cql3::untyped_result_set> msg) {
+        return !msg->empty() && msg->one().has("key");
```
haaawk, line 1668: Is it possible that !msg->empty() and !msg->one().has("key")?

db/system_keyspace.cc
kbr-: Yes

db/system_keyspace.cc
kbr-: No

db/system_keyspace.cc
```diff
@@ -1646,6 +1661,14 @@ future<std::optional<db_clock::time_point>> get_saved_cdc_streams_timestamp() {
     });
 }
 
+future<bool> cdc_is_rewritten() {
+    // We don't care about the actual timestamp; it's additional information for debugging purposes.
+    return qctx->execute_cql(format("SELECT key FROM system.{} WHERE key = ?", v3::CDC_LOCAL), "rewritten")
```
haaawk, line 1666: s/"rewritten"/CDC_REWRITTEN_KEY?

cdc/generation.cc
```diff
@@ -373,4 +375,171 @@ void update_streams_description(
...
+    }
+
+    // It's safe to discard this future: the coroutine keeps system_distributed_keyspace alive
+    // and the abort source's lifetime extends the lifetime of any other service.
+    (void)(([_times_and_ttls = std::move(times_and_ttls), _sys_dist_ks = std::move(sys_dist_ks),
+                _get_num_token_owners = std::move(get_num_token_owners), &_abort_src = abort_src] () mutable -> future<> {
+        auto times_and_ttls = std::move(_times_and_ttls);
+        auto sys_dist_ks = std::move(_sys_dist_ks);
+        auto get_num_token_owners = std::move(_get_num_token_owners);
+        auto& abort_src = _abort_src;
```
haaawk, line 531: Why are you moving those objects to local variables to later move them again? Can't you just move straight from the lambda in lines 535-539?

haaawk (commented):
LGTM in general. I left some comments. I think @avikivity idea to reverse sort the timestamps table is worth implementing.

cql3/query_processor.cc
haaawk: It would be easier to review. Bundling together unrelated changes makes more harm.

docs/design-notes/cdc.md
haaawk: ok. I forgot it's a local table that affects only single node.

db/system_keyspace.cc
haaawk: Can we drop the msg->one().has("key") check then?

cql3/query_processor.cc
kbr-: Isn't this change sufficiently small and "obvious"?

db/system_keyspace.cc
kbr-: Yes, if you really want

cdc/generation.cc
kbr-: The lambda-capture variables are not local variables, there are struct/class fields (the struct/class being the lambda's internal implementation).
The coroutine - the lambda's body - is that struct's `operator()`.
Unfortunately, on the first coroutine suspension, the struct/class gets freed and accessing these variables is a use-after-free. I need variables inside the coroutine frame (which gets preserved across suspensions) - that's what the definitions here are for.

This is how I understand it, perhaps I got it wrong - @avikivity knows what's going on

cdc/generation.cc
```diff
@@ -373,4 +375,171 @@ void update_streams_description(
...
+
+        // This code is racing with node startup. At this point, we're most likely still waiting for gossip to settle
+        // and some nodes that are UP may still be marked as DOWN by us.
+        // Let's sleep a bit to increase the chance that the first attempt at rewriting succeeds (it's still ok if
+        // it doesn't - we'll retry - but it's nice if we succeed without any warnings).
+        co_await sleep_abortable(std::chrono::seconds(10), abort_src);
+
+        cdc_log.info("Rewriting stream tables in the background...");
+        auto last_rewritten = co_await rewrite_streams_descriptions(
+                std::move(times_and_ttls),
```
kbr-, line 541: @haaawk using `_times_and_ttls` here (for example) would be a use-after-free

cdc/generation.cc
```diff
@@ -373,4 +375,171 @@ void update_streams_description(
...
+
+    if (times_and_ttls.empty()) {
+        // There's no point in rewriting old generations' streams (they don't contain any data).
+        cdc_log.info("No CDC log tables present, not rewriting stream tables.");
+        co_return co_await db::system_keyspace::cdc_set_rewritten(std::nullopt);
+    }
+
+    // It's safe to discard this future: the coroutine keeps system_distributed_keyspace alive
+    // and the abort source's lifetime extends the lifetime of any other service.
+    (void)(([_times_and_ttls = std::move(times_and_ttls), _sys_dist_ks = std::move(sys_dist_ks),
```
avikivity, line 526: It may be safe, but exceptions will not be logged nicely (there will be that exceptional future ignore). Is there a reason not to await this future?

A few seconds slowdown during startup are well worth getting good exceptions, and being sure that the rewrite is done and over.

cdc/generation.cc
```diff
@@ -492,6 +492,11 @@ future<> maybe_rewrite_streams_descriptions(
         co_return;
     }
 
+    if (db.get_config().cdc_dont_rewrite_streams()) {
+        cdc_log.warn("Stream rewriting disabled. Manual administrator intervention may be required...");
```
avikivity, line 496: Very good.

avikivity:
Very good. I left an obligatory comment on the dropped future, but it's minor.

I do think we should sort descending. We've just seen what a bad schema can do. This is minor in comparison, but recovering from a schema mistake is difficult and disrupts everyone - users, drivers, test, and people who write the translation code.

cdc/generation.cc
kbr-: The running time of this task is unbounded - for example, suppose we've shut down the entire cluster, and now we're restarting nodes one by one.
This task won't be able to finish until all nodes are restarted (in the worst case) because it performs a full scan of the streams table.

As for exceptions, all expected exceptions are catched. For example: if nodes are unavailable and the scanning query fails, we log the exception using `cdc_log` and retry after a minute (see `get_cdc_desc_v1_timestamps`). There should be no exceptions flying out of this function

kbr-: Ahh, one exception is the sleep aborted exception coming from `sleep_abortable`. But meh, they only happen on shutdown, I'm not sure if catching them is worth it.

haaawk:
So the last thing left is reverse order of timestamps @kbr- ?

cdc/generation.cc
elcallio: That is correct, but you could still just move the values to your stack frame without going through the lambda move, since you call the lambda immediately anyway (nothing s lot between call and reaching the init-moves above)

kbr-:
v7: reverse order of timestamps, see diff: https://github.com/scylladb/scylla/compare/c9aa10f844920fc7478ef70a3a4aa91bfbe16a5e..e42b1e1485f81eeafbb7fddbd42424382096608f

kbr-:
rebased

cdc/generation.cc
kbr-: @elcallio you mean by passing them as parameters to the lambda?
```
(void)(([] (int x) {
    // yield
    // do something with x
})(42));
```

avikivity:
Since it was based on next and next was edited, I had to dequeue it and can't requeue it.

Please rebase it on master (and don't base on next in the future, it's unstable).

kbr-:
Ok, I'll rebase onto master after @haaawk's patch gets merged into master, since these PRs conflict

kbr-:
@avikivity the pull request is reopened as #8116

Issue #7341, opened 2020-10-06
cassandra-stress starts to timeout and fail after ~20min with GCE local disks


fruch:
*Installation details*
Scylla version (or git commit hash): 4.1.7 (4.1.7-0.20200918.2251a1c577)
Cluster size: 4
OS (RHEL/CentOS/Ubuntu/AWS AMI): Centos8/Ubuntu1804

*Hardware details (for performance issues)*
Platform (physical/VM/cloud instance type/docker): GCE (n1-highmem-16)
Hardware: sockets= cores= hyperthreading= memory=
Disks: (SSD/HDD, count) Nvme 8/6 disks

## Summary

Trying to fill up a GCE setup with 1Tb of data, for testing backup to Google Storage. 

This is the stress command we are using:
```
cassandra-stress write cl=QUORUM n=1100200300 -schema 'replication(factor=3) compaction(strategy=LeveledCompactionStrategy)' -port jmx=6868 -mode cql3 native -rate threads=150 -col 'size=FIXED(200) n=FIXED(5)' -pop seq=1..1100200300
```

After ~20min we start getting timeouts that fails the load with the following errors:
```
com.datastax.driver.core.exceptions.WriteTimeoutException: Cassandra timeout during SIMPLE write query at consistency QUORUM (2 replica were required but only 1 acknowledged the write)
```

## Logs (Centos8)
* **db-cluster** - https://cloudius-jenkins-test.s3.amazonaws.com/519a5474-7368-4a4d-8ac2-485540923422/20201005_113703/db-cluster-519a5474.zip
* **loader-set** - https://cloudius-jenkins-test.s3.amazonaws.com/519a5474-7368-4a4d-8ac2-485540923422/20201005_113703/loader-set-519a5474.zip
* **monitor-set** - https://cloudius-jenkins-test.s3.amazonaws.com/519a5474-7368-4a4d-8ac2-485540923422/20201005_113703/monitor-set-519a5474.zip
* **sct-runner** - https://cloudius-jenkins-test.s3.amazonaws.com/519a5474-7368-4a4d-8ac2-485540923422/20201005_113703/sct-runner-519a5474.zip
*  **monitor** - `hydra investigate show-monitor 519a5474-7368-4a4d-8ac2-48554092342`

## Further thing we tried
we tried this also with ubuntu1804, suspecting maybe it's a kernel related issue, but got the similar results

[...]

PR #8019, opened 2021-02-02
scylla_io_setup did not configure pre tuned gce instances correctly

slivne:
scylla_io_setup condition for nr_disks was using the bitwise operator
(&) instead of logical and operator (and) causing the io_properties
files to have incorrect values

Fixes: #7341

Signed-off-by: Shlomi Livne <shlomi@scylladb.com>

### DIFF
--- a/dist/common/scripts/scylla_io_setup
+++ b/dist/common/scripts/scylla_io_setup
@@ -244,12 +244,12 @@ if __name__ == "__main__":
                 # and https://cloud.google.com/compute/docs/disks/local-ssd#nvme
                 # note that scylla iotune might measure more, this is GCP recommended
                 mbs=1024*1024
-                if nr_disks >= 1 & nr_disks < 4:
+                if nr_disks >= 1 and nr_disks < 4:
                     disk_properties["read_iops"] = 170000 * nr_disks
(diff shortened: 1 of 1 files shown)


slivne:
Prior to the fix the generated file for a 6 disk gce instance n1-standard-16 is 

```
 cat /etc/scylla.d/io_properties.yaml 
disks:
- mountpoint: /var/lib/scylla
  read_bandwidth: 4152360960
  read_iops: 1020000
  write_bandwidth: 2202009600
  write_iops: 540000
```

After this fix the generated file is

```
cat /etc/scylla.d/io_properties.yaml 
disks:
- mountpoint: /var/lib/scylla
  read_bandwidth: 2778726400
  read_iops: 680000
  write_bandwidth: 1468006400
  write_iops: 360000
```

tarzanek:
ok && instead of & should have been used, thank you for the fix @slivne

tarzanek (approved):

avikivity:
Backported to 4.3, 4.4. GCP support was not there in 4.2.

tzach:
@slivne are you sure this fix #7341 ?
If so can we close it?

slivne:
Ill be sure when wa say it did

I think its fine to leave it open till they do  as its a burning issue.

PR #8024, opened 2021-02-02
tools: toolchain: add simplified procedure for creating dbuild images

avikivity:
The current procedure for building images is complicated, as it
requires access to x86_64, aarch64, and s390x machines. Add an alternative
procedure that is fully automated, as it relies on emulation on a single
machine.

It is slow, but requires less attention.


avikivity:
Some bugs fixed, tested.

PR #8033, opened 2021-02-03
test: perf_fast_forward: Quiesce sstable readers between test cases

tgrabiec:
Currently, sstable readers which go out of scope close file readers in the background.
This can leave, for example, file read-aheads active after the read is done. Those will interfere
with the next test case and be reported under the next test case, incorrectly.

Fix by joining with those background operations between test cases.

Tests: [dev] perf_fast_forward

### DIFF
--- a/sstables/data_consume_context.hh
+++ b/sstables/data_consume_context.hh
@@ -27,6 +27,7 @@
 #include <seastar/core/future.hh>
 #include <seastar/util/optimized_optional.hh>
 
+#include "db/background.hh"
 #include "shared_sstable.hh"
 #include "row.hh"
 #include "sstables.hh"
@@ -124,8 +125,7 @@ public:
     ~data_consume_context() {
         if (_ctx) {
             auto f = _ctx->close();
-            //FIXME: discarded future.
-            (void)f.handle_exception([ctx = std::move(_ctx), sst = std::move(_sst)](auto) {});
+            (void)f.handle_exception([ctx = std::move(_ctx), sst = std::move(_sst), op = db::start_background_job()](auto) {});
         }
     }
 
--- a/database.cc
+++ b/database.cc
@@ -28,6 +28,7 @@
 #include "db/system_keyspace.hh"
 #include "db/commitlog/commitlog.hh"
 #include "db/config.hh"
+#include "db/background.hh"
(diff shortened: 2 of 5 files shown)


sstables/data_consume_context.hh
```diff
@@ -124,8 +125,7 @@ public:
     ~data_consume_context() {
         if (_ctx) {
             auto f = _ctx->close();
-            //FIXME: discarded future.
-            (void)f.handle_exception([ctx = std::move(_ctx), sst = std::move(_sst)](auto) {});
+            (void)f.handle_exception([ctx = std::move(_ctx), sst = std::move(_sst), op = db::start_background_job()](auto) {});
```
avikivity, line 128: What about moving this to sstable_manager? It already tracks the background close of the sstable itself.

tgrabiec: I imagined that we will want to track all background operations, not just those which originate in the sstable layer. So the mechanism shouldn't be sstable specific. 

But rather, we should add db::start_background_job() to `sstable_manager::deactivate(sstable*)`.

avikivity: Different users will have different meanings for "all". Of course we have a single user now, so it's not yet a problem.

What about approaching it from another direction? @bhalevy is working on adding close() to readers, so they won't go out of scope in normal operation.

avikivity: (not go out of scope while having work to do)

tgrabiec: There is at least one user which needs "all" to mean every background operation which is triggered by the database layer. We can have more fine-grained tracking at different layers, but we will still have to combine them into the meaning expected by tests. We can split the tracking later if necessary.

Once @bhalevy's work goes in, it could replace some of these, but not all. It won't cover the sstable manager part, for example.

tgrabiec: Bu the sstable manager part won't be hit by this test, so this patch will become unnecessary.

mykaul:
@tgrabiec - do we wish to pursue this, or close?

tgrabiec:
I think it's still relevant

bhalevy:
> I think it's still relevant

please rebase and let's push it forward if it's still relevant

